PR #171, opened 2014-08-29
Add getnetworkinfo support to btcctl.

dajohi:


stevenroose:
Why was this closed?

dajohi:
Looks like it was the old code before it was rewritten.  btcctl does support getnetworkinfo still.

PR #198, opened 2014-12-22
Limit the number of addresses announced.

dajohi:
This matches the Bitcoin Core commit https://github.com/bitcoin/bitcoin/commit/12a49cac0a561ada277e93549cae26a3123a6023

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -9,6 +9,7 @@ import (
 	"container/list"
 	"fmt"
 	"io"
+	prand "math/rand"
 	"net"
 	"strconv"
 	"sync"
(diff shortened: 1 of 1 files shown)


davecgh:
Package names should be all lowercase per effective Go guidelines.  Also, please rebase this to the latest master.

After that, looks good.

peer.go
```diff
@@ -1145,21 +1169,18 @@ func (p *peer) pushAddrMsg(addresses []*btcwire.NetAddress) error {
...
+	if numAdded > 0 {
+		for _, na := range msg.AddrList {
+			// Add address to known addresses for this peer.
+			p.knownAddresses[addrmgr.NetAddressKey(na)] = struct{}{}
+		}
 
-	// Send message with remaining addresses if needed.
-	if numAdded%btcwire.MaxAddrPerMsg != 0 {
 		p.QueueMessage(msg, nil)
+		// NOTE: This needs to be a new address message and not
```
davecgh: This NOTE is no longer needed since there is only a single message now.  It was there before since multiple messages were being generated.

PR #203, opened 2015-01-01
Fixed DeletePortMapping documentation typo

Beldur:


davecgh:
Thanks.  I'm surprised the linter didn't catch that.

jrick:
It's because it was a method for an unexported type, which golint ignores.

PR #204, opened 2015-01-03
Let help message be printed by go-flags

Beldur:
go-flags can print help message with the `flags.PrintErrors` Flag. So there is no need to print in manually and check for error type.

### DIFF
--- a/config.go
+++ b/config.go
@@ -324,13 +324,10 @@ func loadConfig() (*config, []string, error) {
 	// help message error can be ignored here since they will be caught by
 	// the final parse below.
 	preCfg := cfg
-	preParser := newConfigParser(&preCfg, &serviceOpts, flags.HelpFlag)
+	preParser := newConfigParser(&preCfg, &serviceOpts, flags.HelpFlag|flags.PrintErrors)
 	_, err := preParser.Parse()
 	if err != nil {
-		if e, ok := err.(*flags.Error); ok && e.Type == flags.ErrHelp {
-			fmt.Fprintln(os.Stderr, err)
-			return nil, nil, err
-		}
+		return nil, nil, err
 	}
 
 	// Show the version and exit if the version flag was specified.
--- a/config.go
+++ b/config.go
@@ -319,21 +319,26 @@ func loadConfig() (*config, []string, error) {
 	// Service options which are only added on Windows.
 	serviceOpts := serviceOptions{}
 
+	// Create appName and usageMessage for later use
+	appName := filepath.Base(os.Args[0])
+	appName = strings.TrimSuffix(appName, filepath.Ext(appName))
+	usageMessage := fmt.Sprintf("Use %s -h to show usage", appName)
+
 	// Pre-parse the command line options to see if an alternative config
 	// file or the version flag was specified.  Any errors aside from the
 	// help message error can be ignored here since they will be caught by
 	// the final parse below.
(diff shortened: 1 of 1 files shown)


davecgh:
Thanks for the pr, but this is a no from me.  The current logic is very specific to make cli errors and file parse error consistent.

For example, the current functionality:

Invalid CLI option:
```bash
$ ./btcd --boguscliopt
unknown flag `boguscliopt'
Use btcd -h to show usage
```

Invalid option in the config file:
```bash
$ ./btcd
Error parsing config file: unknown option: bogusconfigfileopt
Use btcd -h to show usage
```

However with this pull request:

Invalid CLI option:
```bash
$ ./btcd --boguscliopt
unknown flag `boguscliopt'
```

Invalid option in the config file:
```bash
$ ./btcd
Error parsing config file: unknown option: bogusconfigfileopt
Use btcd -h to show usage
```

Beldur:
Made it consistent with second command line argument parsing.

davecgh:
This latest modifies the behavior it in another way.  Currently, if there is a is a bad CLI option *and* an invalid entry in the config file, the invalid entry in the config file will be shown, whereas with this pr the invalid CLI option will be shown.

**Current master with invalid CLI option and invalid config file entry**
```bash
$ ./btcd --boguscliopt
Error parsing config file: unknown option: bogusconfigfileopt
Use btcd -h to show usage
``` 

**This pull request with invalid CLI option and invalid config file entry**
```bash
$ ./btcd --boguscli
unknown flag `boguscli'
Use btcd -h to show usage
``` 

Perhaps we could argue the merits of one way over another, but I really don't see any benefit to changing this code since it's already well tested, works as intended, and is easy to follow.  This latest version still manually checks the error type which goes against the stated intention of this pull request.

I'm certainly not against changing code that fixes bugs, improves consistency, or otherwise has benefit, but I currently don't see any upside to making this change.  I don't feel like it makes the code any easier to read, and both attempts at changing it have resulted in changing the current error handling behavior.

I'll leave it open to get the opinion of others (@jrick, @dajohi, etc), but my current recommendation is to leave this this section as it is unless there is a good reason to change it.

jrick:
It works well as is and is well tested, and the code is understandable, so I'm in favor of not changing this.

Beldur:
Are you sure you get this result?

    $ ./btcd --boguscliopt
    Error parsing config file: unknown option: bogusconfigfileopt
    Use btcd -h to show usage

The error message makes no sense as 1. there should not be a config file being parsed 2. bogusconfigfileopt != boguscliopt

Anyway, the change is not very important and in favour of the well tested current behavior I will close.

davecgh:
@Beldur: On current master, yes that is the result.

The config file is parsed before the CLI options so that the CLI options take precedence and override anything specified in the config file.  Therefore a parse error in the config file is seen before an error in the CLI options.  That is the reason for the CLI preparse to begin with as otherwise options like -h and -V, which should function regardless of whether there are any errors in the config file, would not work since the parse error would be seen first.

Another issue in regards to the change that was proposed in this pr is that the -h (--help) option intentionally puts the currently configured values into parenthesis after the argument description.  Those values are clearly modifiable by the config file and thus it requires parsing the config file first before displaying help.  That is the reason the preparse error is ignored when it's a help error, so the config file can be parsed for properly showing the current values in the help output.

PR #205, opened 2015-01-04
Add support for an optional address-based transaction index

Roasbeef:
This PR adds support for an optional `addrindex`. This PR depends on both conformal/btcdb#11 and conformal/btcjson#33. Additionally, this PR is the final pull request needed towards completion of conformal/btcd#190 (will also update `btcrpcclient` once `btcjson2` is in).

With the addition of this PR, `btcd` gains an additional service, the `addrIndexer`. Factoring in new the optional `--addrindex` flag, there are three possible operating modes w.r.t the address-based transaction index:
  * `btcd` is fully synced, `addrindex` is activated `(chaintip > addrtip)`:
     * In this mode, the `addrIndexer` enters "catch up" mode. While within this mode, the address index will be concurrently built up starting from the genesis block to the current best main chain height. 
  * `btcd` is fully synced/syncing, `addrindex` is actived `(chaintip == addrtip)`:
     * At this stage the address index is fully caught up with the best height. This mode is characterized by a single worker serially processing indexing jobs passed to it by the `blockmanager` once it receives a notification that a new block has been connected to the best chain. 
  * `btcd` is fully synced/syncing, `addrindex` is *not* activated:
     * The `addrIndexer` won't be initialized and started. Instead `btcd` will respond to incoming `searchrawtransaction` RPC's based off the stale database index. 
  
`btcjson` support for the new `searchrawtransaction` RPC command has been added in: conformal/btcjson#33. This PR continues the RPC work by fully implementing the call within the `btcd` server, and also adding it to `btcctl`. It's worth nothing that unlike the address-based index implemented in a fork of `bitcoind`, `searchrawtransaction` as implemented will also search within the mempool for transactions matching the query. Transactions pulled from the mempool are differentiated from tx's within the database in their JSON response: they don't have a `"confirmations"` field etc. The `mempool` has been modified, adding an in memory address-based index which is created and maintained if `addrindex` is activated. 

NOTE: Support for deleting the entire address index has been added to `btcdb` in conformal/btcdb#11, but isn't yet exposed. Should this be exposed as a separate config parameter? Something like `--dropaddrindex`.

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -53,8 +53,8 @@ type invMsg struct {
 	peer *peer
 }
 
-// blockMsg packages a bitcoin block message and the peer it came from together
-// so the block handler has access to that information.
+// headersMsg packages a bitcoin headers message and the peer it came from
+// together so the block handler has access to that information.
 type headersMsg struct {
 	headers *wire.MsgHeaders
 	peer    *peer
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -166,9 +166,9 @@ type blockManager struct {
 	blockChain        *blockchain.BlockChain
 	requestedTxns     map[wire.ShaHash]struct{}
 	requestedBlocks   map[wire.ShaHash]struct{}
+	progressLogger    *blockProgressLogger
 	receivedLogBlocks int64
 	receivedLogTx     int64
-	lastBlockLogTime  time.Time
 	processingReqs    bool
 	syncPeer          *peer
 	msgChan           chan interface{}
@@ -436,41 +436,6 @@ func (b *blockManager) handleDonePeerMsg(peers *list.List, p *peer) {
 	}
 }
 
-// logBlockHeight logs a new block height as an information message to show
-// progress to the user.  In order to prevent spam, it limits logging to one
-// message every 10 seconds with duration and totals included.
-func (b *blockManager) logBlockHeight(block *btcutil.Block) {
-	b.receivedLogBlocks++
-	b.receivedLogTx += int64(len(block.MsgBlock().Transactions))
-
-	now := time.Now()
-	duration := now.Sub(b.lastBlockLogTime)
-	if duration < time.Second*10 {
-		return
-	}
-
-	// Truncate the duration to 10s of milliseconds.
-	durationMillis := int64(duration / time.Millisecond)
-	tDuration := 10 * time.Millisecond * time.Duration(durationMillis/10)
-
-	// Log information about new block height.
-	blockStr := "blocks"
-	if b.receivedLogBlocks == 1 {
-		blockStr = "block"
-	}
-	txStr := "transactions"
-	if b.receivedLogTx == 1 {
-		txStr = "transaction"
-	}
-	bmgrLog.Infof("Processed %d %s in the last %s (%d %s, height %d, %s)",
-		b.receivedLogBlocks, blockStr, tDuration, b.receivedLogTx,
-		txStr, block.Height(), block.MsgBlock().Header.Timestamp)
-
-	b.receivedLogBlocks = 0
-	b.receivedLogTx = 0
-	b.lastBlockLogTime = now
-}
-
 // handleTxMsg handles transaction messages from all peers.
 func (b *blockManager) handleTxMsg(tmsg *txMsg) {
 	// NOTE:  BitcoinJ, and possibly other wallets, don't follow the spec of
@@ -628,7 +593,7 @@ func (b *blockManager) handleBlockMsg(bmsg *blockMsg) {
 		// When the block is not an orphan, log information about it and
 		// update the chain state.
 
-		b.logBlockHeight(bmsg.block)
+		b.progressLogger.LogBlockHeight(bmsg.block)
 
 		// Query the db for the latest best block since the block
 		// that was processed could be on a side chain or have caused
@@ -834,7 +799,7 @@ func (b *blockManager) handleHeadersMsg(hmsg *headersMsg) {
 		b.headerList.Remove(b.headerList.Front())
 		bmgrLog.Infof("Received %v block headers: Fetching blocks",
 			b.headerList.Len())
-		b.lastBlockLogTime = time.Now()
+		b.progressLogger.SetLastLogTime(time.Now())
 		b.fetchHeaderBlocks()
 		return
 	}
@@ -1167,6 +1132,12 @@ func (b *blockManager) handleNotifyMsg(notification *blockchain.Notification) {
 			r.ntfnMgr.NotifyBlockConnected(block)
 		}
 
+		// If we're maintaing the address index, and it is up to date
+		// then update it based off this new block.
+		if cfg.AddrIndex && b.server.addrIndexer.IsCaughtUp() {
+			b.server.addrIndexer.UpdateAddressIndex(block)
+		}
+
 	// A block has been disconnected from the main block chain.
 	case blockchain.NTBlockDisconnected:
 		block, ok := notification.Data.(*btcutil.Block)
@@ -1344,14 +1315,15 @@ func newBlockManager(s *server) (*blockManager, error) {
 	}
 
 	bm := blockManager{
-		server:           s,
-		requestedTxns:    make(map[wire.ShaHash]struct{}),
-		requestedBlocks:  make(map[wire.ShaHash]struct{}),
-		lastBlockLogTime: time.Now(),
-		msgChan:          make(chan interface{}, cfg.MaxPeers*3),
-		headerList:       list.New(),
-		quit:             make(chan struct{}),
+		server:          s,
+		requestedTxns:   make(map[wire.ShaHash]struct{}),
+		requestedBlocks: make(map[wire.ShaHash]struct{}),
+		progressLogger:  newBlockProgressLogger("Processed", bmgrLog),
+		msgChan:         make(chan interface{}, cfg.MaxPeers*3),
+		headerList:      list.New(),
+		quit:            make(chan struct{}),
 	}
+	bm.progressLogger = newBlockProgressLogger("Processed", bmgrLog)
 	bm.blockChain = blockchain.New(s.db, s.netParams, bm.handleNotifyMsg)
 	bm.blockChain.DisableCheckpoints(cfg.DisableCheckpoints)
 	if !cfg.DisableCheckpoints {
--- a/btcd.go
+++ b/btcd.go
@@ -81,6 +81,17 @@ func btcdMain(serverChan chan<- *server) error {
 	}
 	defer db.Close()
 
+	if cfg.DropAddrIndex {
+		btcdLog.Info("Deleting entire addrindex.")
+		err := db.DeleteAddrIndex()
+		if err != nil {
+			btcdLog.Errorf("Unable to delete the addrindex: %v", err)
+			return err
+		}
+		btcdLog.Info("Successfully deleted addrindex, exiting")
+		return nil
+	}
+
 	// Ensure the database is sync'd and closed on Ctrl+C.
 	addInterruptHandler(func() {
 		btcdLog.Infof("Gracefully shutting down the database...")
--- a/chainindexer.go
+++ b/chainindexer.go
@@ -1,0 +1,487 @@
+// Copyright (c) 2013-2014 Conformal Systems LLC.
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package main
+
+import (
+	"container/heap"
+	"runtime"
+	"sync"
+	"sync/atomic"
+
+	"github.com/btcsuite/btcd/blockchain"
+	"github.com/btcsuite/btcd/database"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+	"golang.org/x/crypto/ripemd160"
+)
+
+type indexState int
+
+const (
+	// Our two operating modes.
+
+	// We go into "CatchUp" mode when, on boot, the current best
+	// chain height is greater than the last block we've indexed.
+	// "CatchUp" mode is characterized by several concurrent worker
+	// goroutines indexing blocks organized by a manager goroutine.
+	// When in "CatchUp" mode, incoming requests to index newly solved
+	// blocks are backed up for later processing. Once we've finished
+	// catching up, we process these queued jobs, and then enter into
+	// "maintainence" mode.
+	indexCatchUp indexState = iota
+	// When in "maintainence" mode, we have a single worker serially
+	// processing incoming jobs to index newly solved blocks.
+	indexMaintain
+)
+
+// Limit the number of goroutines that concurrently
+// build the index to catch up based on the number
+// of processor cores.  This help ensure the system
+// stays reasonably responsive under heavy load.
+var numCatchUpWorkers = runtime.NumCPU() * 3
+
+// indexBlockMsg packages a request to have the addresses of a block indexed.
+type indexBlockMsg struct {
+	blk  *btcutil.Block
+	done chan struct{}
+}
+
+// writeIndexReq represents a request to have a completed address index
+// committed to the database.
+type writeIndexReq struct {
+	blk       *btcutil.Block
+	addrIndex database.BlockAddrIndex
+}
+
+// addrIndexer provides a concurrent service for indexing the transactions of
+// target blocks based on the addresses involved in the transaction.
+type addrIndexer struct {
+	server          *server
+	started         int32
+	shutdown        int32
+	state           indexState
+	quit            chan struct{}
+	wg              sync.WaitGroup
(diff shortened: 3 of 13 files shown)


davecgh:
Glad to see this is nearing completion.  I'll help test it and give it a good review over the next few days.  I'll be able to tell as well during testing, but, out of curiosity, how much does it increase the database size?

Roasbeef:
I have a new local node syncing the mainnet chain while catching up the address index, it's connected to an HDD so it's pretty slow. Can see for the mainnet once it's finished, but it took the testnet chain from 1.5GB to 2GB fwiw.

blockmanager.go
```diff
@@ -1341,14 +1310,14 @@ func newBlockManager(s *server) (*blockManager, error) {
...
-		quit:             make(chan struct{}),
-	}
+		server:          s,
+		requestedTxns:   make(map[btcwire.ShaHash]struct{}),
+		requestedBlocks: make(map[btcwire.ShaHash]struct{}),
+		msgChan:         make(chan interface{}, cfg.MaxPeers*3),
+		headerList:      list.New(),
+		quit:            make(chan struct{}),
+	}
+	bm.progressLogger = newBlockProgressLogger("Processed")
```
davecgh: This can created inside the initialization struct since it doesn't require a reference to the block manager itself.

mempool.go
```diff
@@ -610,12 +611,55 @@ func (mp *txMemPool) removeTransaction(tx *btcutil.Tx) {
...
 			delete(mp.outpoints, txIn.PreviousOutPoint)
 		}
 		delete(mp.pool, *txHash)
 		mp.lastUpdated = time.Now()
 	}
+
+}
+
+// removeTransactionFromAddrIndex removes the passed transaction from our
+// address based index.
```
davecgh: This appear to require the mutex.  Please keep consistent with the other functions and document the mutex requirements.  For example 

```Go
//
// This function MUST be called with the mempool lock held (for writes).
```

davecgh:
> $ golint .
> rpcserver.go:3003:9: if block ends with a return statement, so drop this else and outdent its block

davecgh:
In regards to invoking `searchrawtransaction` with address indexing disabled, I think it would make more sense to return an error rather than returning results based on stale data where most likely aren't right.  It seems dangerous to serve potentially invalid data.

I'd like to hear from @adamkrellenstein, @ouziel-slam,  @carryforward since they have expressed interest in this functionality and will be making use of it.

mempool.go
Roasbeef: Done. I've added similar documentation to related functions that were added to the mempool in this PR.

blockmanager.go
Roasbeef: Fixed.

Roasbeef:
I agree about not serving stale data. Pursuing this further, should there also be similar behavior when the index is activated but not yet caught up? As in while the indexer is ripping through past blocks, all `searchrawtransaction` requests should yield an error response?

davecgh:
Yes, I personally feel like it should error until it is caught up as well.

davecgh:
Since the memdb backend does not actually implement the bits necessary to support the address index, please add a check in `config.go` for when both `--dbtype=memdb` and `--addrindex` is set to display an error the combination isn't supported and exit.

carryforward:
For the general case, and for the next few years, yes, failing on incomplete sounds good.

I can imagine our application benefiting from accessing stale data, but ours is a special case.  We can address this a few years down the line.

adamkrellenstein:
Sounds good to me. I don't have any opinion on `--dropaddrindex`.

Do we want mempool mempool transactions just to not have the `confirmations` field, or to have it and set it to `0`? I like the latter way better.

davecgh:
I agree with confirmations = 0 for mempool transactions.  That is more consistent with other calls as well.

Roasbeef:
Pushed a few commits fixing the following:
- [x] A `--dbtype=memdb` and `--addrindex` combination results in a displayed error and abort on startup.
- [x] If the `addrindex` isn't yet caught up, all `searchrawtransaction` requests fail with an error. 
- [x] Transactions returned from the mempool now have `Confirmations: 0` in the JSON response, instead of omitting the field altogether. This required a [change to `btcjson`](https://github.com/conformal/btcjson/pull/37) since the `Confirmations` field had an `omitempty` struct tag. It appears `getrawtransaction` also wasn't showing transactions pulled from the mempool as `Confirmations: 0`.
- [x] The new `blockProgressLogger` struct now also takes a subsystem's logger, so subsequent logs will show up in the logs under the correct subsystem.
- [x] A bug discovered when the addrIndexer was racing against a newly syncing node. The manager goroutine and the writer could possibly get out of sync regarding the current indexing "state". The manager would think we we're caught up, but the writer had a different view. This is fixed by the manager enforcing state changes, and the writer being promoted to a permanent goroutine.

Roasbeef:
For the curious, the size of database with a fully synced-up node, and current address index is: `40GB`

davecgh:
This need to be rebased and updated for the `util/*` -> `cmd/*` move.

davecgh:
Can you please add the new option to the `sample-btcd.conf` file and `doc.go`?

Roasbeef:
Pushed some commits doing the following: 
   * I've added a `--dropaddrindex` flag, which when enabled deletes the entire addrindex and then exits. 
   * The new flags now show up in `doc.go` and `sample-btcd.conf`.
   * Fixed a bug in the indexWriter than caused starvation halting the indexing process if a re-org happened with the indexer was in "maintenance" mode.

btcd.go
```diff
@@ -81,6 +81,17 @@ func btcdMain(serverChan chan<- *server) error {
 	}
 	defer db.Close()
 
+	if cfg.DropAddrIndex {
+		btcdLog.Infof("Deleting entire addrindex.")
```
davecgh: `Info` will suffice here since there are no arguments.

btcd.go
```diff
@@ -81,6 +81,17 @@ func btcdMain(serverChan chan<- *server) error {
...
+		err := db.DeleteAddrIndex()
+		if err != nil {
+			btcdLog.Errorf("Unable to delete the addrindex: %v", err)
+			return err
+		}
+		btcdLog.Infof("Successfully deleted addrindex, exiting")
```
davecgh: Same

davecgh:
There is a new logging subsystem now (`ADXR`), so the `debuglevel` section in `docs/json_rpc_api.md` needs to be updated to include the new subsystem.

davecgh:
I hit this while testing:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x10 pc=0x4eb2bd]

goroutine 51 [running]:
github.com/btcsuite/btcutil.(*Block).Bytes(0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
        D:/Projects/mygo/src/github.com/btcsuite/btcutil/block.go:53 +0x6d
github.com/btcsuite/btcutil.(*Block).TxLoc(0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
        D:/Projects/mygo/src/github.com/btcsuite/btcutil/block.go:174 +0x76
main.(*addrIndexer).indexBlockAddrs(0xc083618a80, 0x0, 0x0)
        D:/Projects/mygo/src/github.com/btcsuite/btcd/chainindexer.go:428 +0x63
main.(*addrIndexer).indexCatchUpWorker(0xc083618a80, 0xc0838f3ec0, 0xc082ed0020, 0xc0838f2600)
        D:/Projects/mygo/src/github.com/btcsuite/btcd/chainindexer.go:378 +0xcd
created by main.(*addrIndexer).indexManager
        D:/Projects/mygo/src/github.com/btcsuite/btcd/chainindexer.go:172 +0x3da
```

chainindexer.go
```diff
@@ -0,0 +1,457 @@
...
+		addrIndex[indexKey] = append(addrIndex[indexKey], locInBlock)
+	}
+	return nil
+}
+
+// indexBlockAddrs returns a populated index of the all the transactions in the
+// passed block based on the addresses involved in each transaction.
+func (a *addrIndexer) indexBlockAddrs(blk *btcutil.Block) btcdb.BlockAddrIndex {
+	addrIndex := make(btcdb.BlockAddrIndex)
+	txLocs, _ := blk.TxLoc()
```
davecgh: The error is being ignored here.  That is bad!

chainindexer.go
```diff
@@ -0,0 +1,457 @@
...
+	close(workerQuit)
+	a.wg.Done()
+}
+
+// indexCatchUpWorker indexes the transactions of previously validated and
+// stored blocks.
+// NOTE: Must be run as a goroutine
+func (a *addrIndexer) indexCatchUpWorker(workChan chan *indexBlockMsg,
+	wg *sync.WaitGroup, quit chan struct{}) {
+	wg.Add(1)
```
davecgh: A `WaitGroup` must always be incremented before launching the goroutine which it is tracking.  Otherwise the main thread could hit the `.Done()` call before it ever increments it to begin with.

Roasbeef:
Updated the section in `docs/json_rpc_api.md` and also created a PR to add the new subsystem to the `debuglevel` command options in `btcjson`.

btcd.go
Roasbeef: Switched over.

btcd.go
Roasbeef: Fixed.

chainindexer.go
Roasbeef: Ah, of course. The `WaitGroup` is now incremented before launching the goroutine.

chainindexer.go
```diff
@@ -0,0 +1,457 @@
...
+	}
+	close(workerQuit)
+	a.wg.Done()
+}
+
+// indexCatchUpWorker indexes the transactions of previously validated and
+// stored blocks.
+// NOTE: Must be run as a goroutine
+func (a *addrIndexer) indexCatchUpWorker(workChan chan *indexBlockMsg,
+	wg *sync.WaitGroup, quit chan struct{}) {
```
davecgh: Since the `WaitGroup` is now proper incremented before calling this function, this function should no longer take it as a parameter.

Roasbeef: The `WaitGroup` still needs to be passed in since graceful exits of the workers are handled by the `indexManager`. This is a separate `WaitGroup` independent of the regular struct-wide WG for the `AddrIndexer` so we can signal our workers to exit after we've caught up and wait for them to finish up their final task(s).

davecgh: Ah, yeah.  You're right.  I was thinking the `workerWg`was on the `addrIndexer` so you'd just do `a.workerWg.Done()` like the rest of the code.

sample-btcd.conf
```diff
@@ -254,3 +254,11 @@
 ; be disabled if this option is not specified.  The profile information can be
 ; accessed at http://localhost:<profileport>/debug/pprof once running.
 ; profile=6061
+
```
davecgh: What do you think about moving this before the mining section?  It's not a big deal, but I think it fits a little better just after the RPC server section since it requires RPC and is probably more likely to be used over the mining and debug bits.

docs/json_rpc_api.md
```diff
@@ -562,7 +562,7 @@ The following is an overview of the RPC methods which are implemented by btcd, b
 |Description|Dynamically changes the debug logging level.<br />The levelspec can either a debug level or of the form `<subsystem>=<level>,<subsystem2>=<level2>,...`<br />The valid debug levels are `trace`, `debug`, `info`, `warn`, `error`, and `critical`.<br />The valid subsystems are `AMGR`, `BCDB`, `BMGR`, `BTCD`, `CHAN`, `DISC`, `PEER`, `RPCS`, `SCRP`, `SRVR`, and `TXMP`.<br />Additionally, the special keyword `show` can be used to get a list of the available subsystems.|
```
davecgh: Looks like ADXR is needed here too.

Roasbeef: Ah, overlooked that, nice catch. Fixed.

chainindexer.go
Roasbeef: All errors are now addressed in latest commit.

sample-btcd.conf
Roasbeef: Agreed, moved to live after the RPC section.

davecgh:
@Roasbeef The btcdb merge is about to take place, so please rebase this and update it for the new location after pr#258 is merged.

The change is a simple `github.com/btcsuite/btcdb` -> `github.com/btcsuite/btcd/database` and of course update all `btcdb.` references to `database.`.

davecgh:
Also, speaking of `docs/json_rpc_api.md`, I noticed the new `searchrawtransaction` RPC needs to be documented under the btcd extensions section as well.

Roasbeef:
Documented the new RPC under the `btcd extensions` section in the docs, and updated for the new `btcdb -> database` paths.

davecgh:
As an update to this, my testing has been going well thus far.  I've tried almost all of the combinations now and everything has been working as expected.

davecgh:
This needs a rebase/update for `blockchain` and `txscript`.

chainindexer.go
```diff
@@ -0,0 +1,497 @@
...
+// indexManager creates, and oversees worker index goroutines.
+// indexManager is the main goroutine for the addresses indexer.
+// It creates, and oversees worker goroutines to index incoming blocks, with
+// the exact behavior depending on the current index state
+// (catch up, vs maintain). Completion of catch-up mode is always proceeded by
+// a gracefull transition into "maintain" mode.
+// NOTE: Must be run as a goroutine.
+func (a *addrIndexer) indexManager() {
+	if a.state == indexCatchUp {
+		adxrLog.Infof("Building up address index from height: %v to %v.",
```
davecgh: Any reason for the `:` in there?  Looks like a relic before you added the top-end range.

sample-btcd.conf
```diff
@@ -199,6 +199,13 @@
 ; the default).
 ; notls=1
 
+; ------------------------------------------------------------------------------
+; Optional Transaction Indexes
+; ------------------------------------------------------------------------------
+; Build and maintain a full address-based transaction index.
+; addrindex=1
+; Delete the entire address index on start up, then exit.
```
davecgh: Nitpick: All of the other options have a blank line between them with blank lines separating the groupings.  Should probably be consistent here.

davecgh:
I've tested this pretty extensively and everything looks good.  Once the rebase and minor nits are addressed, this is ready for merge.

chainindexer.go
Roasbeef: Heh, yeah you're right, was left over before I changed the logging message. Removed the colon.

sample-btcd.conf
Roasbeef: Added the blank link for consistency.

Roasbeef:
Addressed final review comments. Also rebased and squashed commits. Initially attempted to re-order and squash during rebase, but it got messy rather quickly with conflict resolution. Opted to squash into a single commit, updating the commit message.

davecgh:
This is no longer working as of the latest rebase and squash.

> $ btcd --addrindex
> 16:13:42 2015-02-01 [INF] BTCD: Block database loaded with block height 11111
> ...
> 16:13:42 2015-02-01 [INF] ADXR: Building up address index from height 0 to 11111.
> ...
> 16:13:42 2015-02-01 [ERR] ADXR: Unable to get current address index tip: address index hasn't been built up yet
> 16:13:42 2015-02-01 [WRN] SRVR: Server shutting down

Roasbeef:
Should be fixed now with latest rebase and squash. Was erroneously treating the error which is returned from `FetchAddrIndexTip` as critical from the writer goroutine. It only returns an error if the index hasn't been built up, so it was bailing out even before the first write.

ouziel-slama:
I'am currently testing with Counterparty. A little detail: in the `addrindex` branch of bitcoind `searchrawtransactions` have an "s" at the end.

Roasbeef:
@ouziel-slama just pushed a commit fixing the pluralisation. However, if you want to test with this new change, you'll also need to checkout and install btcsuite/btcjson#40 since this commit is dependent on that PR.

ouziel-slama:
@Roasbeef, thank you!

ouziel-slama:
Another detail: `getrawtransaction` accepts only integer (0 or 1) for the second optional boolean parameter `verbose`, but `searchrawtransaction` accepts only boolean parameter.

davecgh:
@ouziel-slama: Thanks.  There are actually several instances where some commands take a bool while other take an "int bool" (0 or 1).  The inconsistency drives me insane!  However, in this case, the `searchrawtransactions` command in BC pr also treats verbose as an integer, so I agree it should be changed to match for compatibility..

Roasbeef:
Just pushed a new commit, the `verbose` parameter now type-checks to an `int` instead of a `bool` in order to maintain compatibility with bitcoind.

Roasbeef:
Rebased, and updated paths for the `btcsuite/btcwire` -> `btcd/wire` merge.

PR #206, opened 2015-01-06
Add a mempool janitor service.

dajohi:
The mempool janitor service runs in its own goroutine scanning
the mempool every 24 hours removing any transactions older than
3 days.

Idea based on Jeff Garzik's mempool janitor for Bitcoin Core:
https://github.com/bitcoin/bitcoin/pull/3753

### DIFF
--- a/mempool.go
+++ b/mempool.go
@@ -76,6 +76,14 @@ const (
 	// and as a base for calculating minimum required fees for larger
 	// transactions.  This value is in Satoshi/1000 bytes.
 	minTxRelayFee = 1000
+
+	// janitorScanInterval is the interval of time the janitor uses
+	// between scans to purge old transactions from the mempool.
+	janitorScanInterval = time.Hour * 24
+
+	// janitorMaxAge is the maximum duration a transaction is allowed
+	// to exist in the mempool.
+	janitorMaxAge = time.Hour * 24 * 3
 )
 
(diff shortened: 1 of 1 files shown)


mempool.go
```diff
@@ -1234,14 +1242,36 @@ func (mp *txMemPool) LastUpdated() time.Time {
...
+// janitorScanInterval removing any transactions that are older
+// than the duration specified by janitorMaxAge.
+func (mp *txMemPool) janitorHandler() {
+	for {
+		time.Sleep(janitorScanInterval)
+
+		mp.Lock()
+		for _, desc := range mp.pool {
+			if time.Since(desc.Added) > janitorMaxAge {
+				mp.removeTransaction(desc.Tx)
```
jrick: I'm worried about this modifying mp.pool as it's being iterated over.  I believe that if a deletion cause a reallocation (to compact the hash table), we may miss iterating over some transactions.

davecgh:
Please reference the Bitcoin core commit which this comes from in the pull request and commit message to give credit where it is due.

davecgh:
After discussing this, we've decided to close it.  This is not a statement as to whether or not the idea is sound, rather due to a lack of interest in the feature as well as how trivial it is for remote nodes to bypass the intended functionality.

PR #207, opened 2015-01-07
Improve double spend error strings.

jrick:
The mempool's MaybeAcceptTransaction methods have also been modified
to return a slice of transaction hashes referenced by the transaction
inputs which are unknown (totally spent or never seen).  While this is
currently used to include the first hash in a ProcessTransaction error
message if inserting orphans is not allowed, it may also be used in
the future to request orphan transactions from peers.

### DIFF
--- a/mempool.go
+++ b/mempool.go
@@ -741,8 +741,9 @@ func (txD *TxDesc) CurrentPriority(txStore btcchain.TxStore, nextBlockHeight int
 func (mp *txMemPool) checkPoolDoubleSpend(tx *btcutil.Tx) error {
 	for _, txIn := range tx.MsgTx().TxIn {
 		if txR, exists := mp.outpoints[txIn.PreviousOutPoint]; exists {
-			str := fmt.Sprintf("transaction %v in the pool "+
-				"already spends the same coins", txR.Sha())
+			str := fmt.Sprintf("output %v already spent by "+
+				"transaction %v in the memory pool",
+				txIn.PreviousOutPoint, txR.Sha())
 			return txRuleError(btcwire.RejectDuplicate, str)
 		}
 	}
@@ -799,10 +800,7 @@ func (mp *txMemPool) FetchTransaction(txHash *btcwire.ShaHash) (*btcutil.Tx, err
 // more details.
 //
 // This function MUST be called with the mempool lock held (for writes).
-func (mp *txMemPool) maybeAcceptTransaction(tx *btcutil.Tx, isOrphan *bool, isNew, rateLimit bool) error {
-	if isOrphan != nil {
-		*isOrphan = false
-	}
+func (mp *txMemPool) maybeAcceptTransaction(tx *btcutil.Tx, isNew, rateLimit bool) ([]*btcwire.ShaHash, error) {
 	txHash := tx.Sha()
 
 	// Don't accept the transaction if it already exists in the pool.  This
@@ -810,7 +808,7 @@ func (mp *txMemPool) maybeAcceptTransaction(tx *btcutil.Tx, isOrphan *bool, isNe
 	// be a quick check to weed out duplicates.
 	if mp.haveTransaction(txHash) {
 		str := fmt.Sprintf("already have transaction %v", txHash)
-		return txRuleError(btcwire.RejectDuplicate, str)
+		return nil, txRuleError(btcwire.RejectDuplicate, str)
 	}
 
 	// Perform preliminary sanity checks on the transaction.  This makes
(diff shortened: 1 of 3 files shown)


mempool.go
```diff
@@ -799,10 +800,9 @@ func (mp *txMemPool) FetchTransaction(txHash *btcwire.ShaHash) (*btcutil.Tx, err
 // more details.
 //
 // This function MUST be called with the mempool lock held (for writes).
-func (mp *txMemPool) maybeAcceptTransaction(tx *btcutil.Tx, isOrphan *bool, isNew, rateLimit bool) error {
-	if isOrphan != nil {
-		*isOrphan = false
-	}
+func (mp *txMemPool) maybeAcceptTransaction(tx *btcutil.Tx,
+	isNew, rateLimit bool) (missingParents []*btcwire.ShaHash, err error) {
```
davecgh: No named returns, please.

mempool.go
```diff
@@ -1040,28 +1043,33 @@ func (mp *txMemPool) maybeAcceptTransaction(tx *btcutil.Tx, isOrphan *bool, isNe
...
+//
+// If the transaction is an orphan (missing parent transactions), the
+// transaction is NOT added to the orphan pool, but each unknown referenced
+// parent is returned.  Use ProcessTransaction instead if new orphans should
+// be added to the orphan pool.
 //
 // This function is safe for concurrent access.
-func (mp *txMemPool) MaybeAcceptTransaction(tx *btcutil.Tx, isOrphan *bool, isNew, rateLimit bool) error {
+func (mp *txMemPool) MaybeAcceptTransaction(tx *btcutil.Tx,
+	isNew, rateLimit bool) (missingParents []*btcwire.ShaHash, err error) {
```
davecgh: Same here.

mempool.go
jrick: I'll update the patch to not use these so it's consistent with the rest of the btcd code, but really all of the issues we've had with named returns are not with the named return values themselves, but rather using bare return statements (which I did not add).  Named returns without bare return statements only help to document the function's behavior.

davecgh: I did notice that it was not doing bare returns and agree that they're the culprit, but I don't even want to open the door for their usage at all.  If they aren't even in the declaration, there is no way to inadvertently forget to return specific values in one of the error paths.  The documentation is easy enough to put in the comment.

jrick:
Bumped the maximum allowed rune length of logged reject messages to 250, as the longer error strings introduced by this PR exceed the previous max of 200.

davecgh:
This also appears to change the orphan handling without mentioning it in the commit message.

mempool.go
```diff
@@ -1040,28 +1042,31 @@ func (mp *txMemPool) maybeAcceptTransaction(tx *btcutil.Tx, isOrphan *bool, isNe
...
 
-	return mp.maybeAcceptTransaction(tx, isOrphan, isNew, rateLimit)
+	return mp.maybeAcceptTransaction(tx, isNew, rateLimit)
 }
 
 // processOrphans determines if there are any orphans which depend on the passed
-// transaction hash (they are no longer orphans if true) and potentially accepts
-// them.  It repeats the process for the newly accepted transactions (to detect
+// transaction hash (it is possible that they are no longer orphans) and
+// potentially accepts them to the memory pool.  It repeats the process for the newly accepted transactions (to detect
```
davecgh: Looks like an enter is missing here.

mempool.go
```diff
@@ -1089,26 +1094,39 @@ func (mp *txMemPool) processOrphans(hash *btcwire.ShaHash) error {
...
 				return err
 			}
-
-			if !isOrphan {
-				// Generate the inventory vector and relay it.
-				iv := btcwire.NewInvVect(btcwire.InvTypeTx, tx.Sha())
-				mp.server.RelayInventory(iv)
-			} else {
-				mp.removeOrphan(orphanHash)
+			if len(missingParents) != 0 {
```
davecgh: The behavior has been changed here.  I'll have to review it more closely to see the implications, but I wanted to point out that this is a change in behavior.  Previously, it removed the orphan from the pool and pushed the orphan hash on to the process hashes list.  Now, it is doing neither of those things.

jrick: You're right it's a change in behavior.  The orphan was already removed from the orphan pool above (line 1111 in this diff).  It it not added back by the maybeAcceptTransaction (and I added additional comments to document this behavior), so an extra delete is unnecessary.  Additionally, processOrphans is documented to be called on a newly accepted transaction, but in this case the transaction was purged rather than accepted.

However, because the previous behavior of processOrphans was to also remove any orphan transactions which remained orphans, with this change we lose the behavior of recursively removing dependant orphan transactions on the first removed one.  That should be fixed, but I don't think processOrphans is the correct way to handle this (there is no reason to call maybeAcceptTransaction to determine if they are an orphan, because we know at least one parent is missing), so that functionality should be re-added into another function.

davecgh:
As discussed on IRC, please split the behavior change to another pull request  The error string changes are uncontroversial and can go in quickly, but we'll need to test the orphan changes thoroughly and examine all the implications of the behavior change.

jrick:
Orphan eviction behavior should be exactly the same as on master now, but I added plenty of comments explaining why I think master is iffy and how I think it should be done.  Will submit another PR with those changes for review separately.

davecgh:
OK.  We'll want to address the comments regarding orphan behavior separately.

PR #212, opened 2015-01-08
Switch peer.requestQueue from linked list to slice.

jrick:
Only two operations are performed with this data structure: adding to
the back and removing from the front.  Because middle inserts and
deletions are never needed, a linked list results in overall worse
performance due to an extra allocation for each element's node, worse
cache locality, and the runtime cost of boxing/unboxing each item
during accesses.

On top of the performance gains, a slice is more type safe as it is a
true generic data structure making it is impossible to insert or
access an element with the wrong type.

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -930,7 +930,7 @@ func (b *blockManager) handleInvMsg(imsg *invMsg) {
 		}
 		if !haveInv {
 			// Add it to the request queue.
-			imsg.peer.requestQueue.PushBack(iv)
+			imsg.peer.requestQueue = append(imsg.peer.requestQueue, iv)
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


jrick:
list.List is used in a couple other places in the peer code.  If this change is OK, I can include the others too, or save them for a later commit.

davecgh:
We'll want to run this for a while with the profiler to make sure it doesn't introduce any memory leaks. It reads fine, but I'm always extra careful about this scheme because, once upon a time, it lead to a memory lead back in the early Go days

jrick:
FWIW, we're already using slices as queues this way for websocket notifications.  If there are any runtime leaks due to this, they'd already be seen on master.

davecgh:
I've been running this for a while and keeping an eye on memory consumption on both the OS and the profile.  Everything looks good.

OK

PR #213, opened 2015-01-08
Correct RPC time handling for getblocktemplate.

davecgh:
This pull request modifies the getblocktemplate RPC to correctly handle the curtime field.

Currently, the RPC server will refuse to serve a block template if the current time is before the minimum allowed time for a block even though the real generated block template already accounted for it.

This consists of three changes:
- Remove the unnecessary and incorrect check in the RPC invocation for the current time against the min required time since it is already handled properly by the block template generation.
- Expose the network time source to the RPC work state and use the adjusted time for checking against the maximum allowed time instead of  the current time.
- Set the returned RPC result curtime field to the time of the generated block header.


dajohi:
OK

jrick:
ok

PR #217, opened 2015-01-11
Don't treat nulldata tx outs as nonstandard dust.

davecgh:
This pull request modifies the memory pool standard transaction dust check to exclude transactions with nulldata outputs so they may be considered standard.


jrick:
ok

PR #245, opened 2015-01-18
Move all wiki docs into docs folder of the repo.

davecgh:
This allows the docs to be updated via pull requests and makes it easier to keep it in sync as a part of normal code updates.


jcvernaleo:
I think this is a much better way to handle the docs.  Keeping them separate does not encourage good documentation or collaboration.

PR #249, opened 2015-01-21
Always ping the remote peer.

dajohi:
This commit unconditionally sends a ping to the remote peer every
pingInterval time (currently 2 minutes).  Additionally, if the remote
peer's protocol version is greater than btcwire.BIP0031Version, the
remote peer must send a pong in response to the ping.  A timer is
setup for such peers and will disconnect the peer if a response is
not received in pingTimeoutMinutes (currently 1 minute).

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -50,9 +50,12 @@ const (
 	// we time out a peer.
 	idleTimeoutMinutes = 5
 
-	// pingTimeoutMinutes is the number of minutes since we last sent a
-	// message requiring a reply before we will ping a host.
-	pingTimeoutMinutes = 2
+	// pingInterval is the interval in which a ping is sent to the peer
+	pingInterval = 2 * time.Minute
+
+	// pingTimeoutMinutes is the number of minutes the server will wait
+	// for a pong before disconnecting the peer.
+	pingTimeoutMinutes = 1 * time.Minute
 )
 
 var (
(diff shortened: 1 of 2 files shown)


peer.go
```diff
@@ -50,9 +50,12 @@ const (
...
 
-	// pingTimeoutMinutes is the number of minutes since we last sent a
-	// message requiring a reply before we will ping a host.
-	pingTimeoutMinutes = 2
+	// pingInterval is the interval in which a ping is sent to the peer
+	pingInterval = 2 * time.Minute
+
+	// pingTimeoutMinutes is the number of minutes the server will wait
+	// for a pong before disconnecting the peer.
+	pingTimeoutMinutes = 1 * time.Minute
```
davecgh: Perhaps call this `pongTimeoutMinutes` since you're waiting on a pong?

peer.go
```diff
@@ -1688,76 +1695,59 @@ cleanup:
...
 	for {
 		select {
+		case <-pingTicker.C:
+			nonce, err := btcwire.RandomUint64()
+			if err != nil {
+				peerLog.Errorf("Failed to generate nonce for ping to %s: %v",
+					p, err)
+				continue
+			}
+			p.QueueMessage(btcwire.NewMsgPing(nonce), nil)
```
davecgh: Minor nitpick, but the other case has a space before it.  Please be consistent.

peer.go
```diff
@@ -1245,10 +1249,12 @@ func (p *peer) handlePongMsg(msg *btcwire.MsgPong) {
 	// infrequently enough that if they overlap we would have timed out
 	// the peer.
 	if p.protocolVersion > btcwire.BIP0031Version &&
-		p.lastPingNonce != 0 && msg.Nonce == p.lastPingNonce {
-		p.lastPingMicros = time.Now().Sub(p.lastPingTime).Nanoseconds()
-		p.lastPingMicros /= 1000 // convert to usec.
-		p.lastPingNonce = 0
+		msg.Nonce == p.lastPingNonce {
+		if p.pingTimer != nil {
```
davecgh: I'm fairly certain this is a race.  Both the `inHandler` and `outHandler` goroutines are accessing the `pingTimer` field on the peer without any synchronization to protect access.

peer.go
```diff
@@ -1245,10 +1249,12 @@ func (p *peer) handlePongMsg(msg *btcwire.MsgPong) {
 	// infrequently enough that if they overlap we would have timed out
 	// the peer.
 	if p.protocolVersion > btcwire.BIP0031Version &&
```
davecgh: I personally think a simple check for `p.lastPingNonce != 0` would be better than the protocol version check here.  The pong message didn't exist until after that protocol version and thus this function would never even be invoked on older protocol versions anyways.

PR #252, opened 2015-01-26
Intentionally ignore alert messages.

dajohi:
Currently, the reference client bans peers that send alerts not signed
with its key.  We could verify against their key, but since the
reference client developers are currently unwilling to support other
implementations' alert messages, we will not relay theirs.

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -1468,7 +1468,14 @@ out:
 			p.handlePongMsg(msg)
 
 		case *btcwire.MsgAlert:
-			p.server.BroadcastMessage(msg, p)
+			// Intentionally ignore alert messages.
+			//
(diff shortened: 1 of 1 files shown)


davecgh:
While I don't really much care for this solution and would much rather have a setup where the current `alert` message only used for consensus-related update messages and involves a `network consensus key` which all implementations would care about, this seems reasonable to do for now until a better solution is implemented.

A possible solution is for us to simply invent a new message type (perhaps `btcdalert`?) which will simply be ignored by other implementations and relayed amongst btcd nodes.  Granted that solution isn't the best case scenario since it won't be able to relay through all nodes, but it is something that would work and doesn't have any negative consequences on the network.

Issue #231, opened 2015-01-16
deadlock in orphan processing with SPV clients connected.


dajohi:
in processOrphans, the mempool is locked already.

```go
                        if len(missingParents) == 0 {
                                // Generate and relay the inventory vector for the
                                // newly accepted transaction.
                                iv := btcwire.NewInvVect(btcwire.InvTypeTx, tx.Sha())
                                mp.server.RelayInventory(iv)
                        } else {
```

RelayInventory contains this chunk:

```go
                         if p.filter.IsLoaded() {
                                tx, err := s.txMemPool.FetchTransaction(&iv.Hash)
                                if err != nil {
                                        peerLog.Warnf("Attempt to relay tx %s "+
                                                "that is not in the memory pool",
                                                iv.Hash)
                                        return
                                }

                                if !p.filter.MatchTxAndUpdate(tx) {
                                        return
                                }
                        }

```

The deadlock happens when we call FetchTransaction.  The mempool is already locked, so we wait for it to be released, which will never happen since processOrphans is waiting for RelayInventory to return.

Since we do not care about a return value, you could simply do:
```go
-                               mp.server.RelayInventory(iv)
+                               go mp.server.RelayInventory(iv)
```

but not sure if that is optimal.

davecgh:
As discussed on IRC, I agree that this is an issue.  However, the proposed solution has the issue that it loses ordering of the inventory.  This can be problematic when you have transactions that depend on other transactions that are to be relayed as well.   This could easily lead to remote peer rejecting the transaction because it doesn't know about its dependency transaction yet depending on its particular orphan rules.

davecgh:
@Roasbeef will be taking a look at this.

Roasbeef:
FWIW, if Go had recursive-locks, then this deadlock would never arise...

However, I've come up with, what I think may be a viable solution. Rather than having the main `peerHandler` attempt to fetch and match TX's for relay to SPV peers, what about having a dedicated SPV transaction relay handler? This would eliminate the random deadlock than can arise when we get a circular chain of calls from: `mp.processTransaction`->`mp.server.RelayInventory`->`mp.FetchTransaction`, occasionally causing a nasty deadlock if the mempool blocks on the buffered relay channel while the server is blocking on `FetchTransaction`. This approach should also maintain inventory ordering for peers. 

Here's a diff of a quick PoC: 

```diff
diff --git a/server.go b/server.go
index 3bf3e58..064bd99 100644
--- a/server.go
[...]

PR #262, opened 2015-01-29
Relay inv msgs now include underlying data object

Roasbeef:
When an inv is to be sent to the server for relaying, the sender already has access to the underlying data. So instead of requiring the relay to look up the data by hash, the data is now coupled in the relay request message.

Previously, forcing the relay to look up the data itself led to an occasional deadlock between the `mempool` and `server` as detailed in: #231. This PR purpotes to eliminate any possiblity for deadlock by implementing [davecgh's proposed solution](https://github.com/btcsuite/btcd/issues/231#issuecomment-71876668).

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -1131,7 +1131,7 @@ func (b *blockManager) handleNotifyMsg(notification *btcchain.Notification) {
 
 		// Generate the inventory vector and relay it.
 		iv := btcwire.NewInvVect(btcwire.InvTypeBlock, hash)
-		b.server.RelayInventory(iv)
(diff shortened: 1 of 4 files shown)


davecgh:
Thanks @Roasbeef.  This reads well.  I'd like to have @dajohi test it since he is the original issue author.

jrick:
I'll load this up and see if I run into anything, but agree that the code looks good.

jrick:
ok

davecgh:
@Roasbeef please rebase this to the latest master.  It's good to go.

Roasbeef:
Cool, just rebased to latest master.

PR #265, opened 2015-01-29
typo:using_bootrap_dat.md -> using_bootstrap_dat.md

multikatt:
Typo in filename, this should fix that.


davecgh:
Nice catch.  Thanks.

PR #269, opened 2015-01-31
Use goleveldb Has() API.

dajohi:
This change converts the leveldb database's ExistsSha() and
ExistsTxSha to use the goleveldb API.  Has() only returns if
the key exists and does not need to read the entire value into
memory resulting in less disk i/o and much less GC.


davecgh:
OK.

PR #272, opened 2015-02-04
txscript: Add new flags ScriptVerifyStrictEncoding and ScriptVerifyDerSignatures.

dajohi:
The ScriptVerifyStrictEncoding flag enforces signature scripts
and public keys to follow the strict encoding requirements.

This flag mimics Bitcoin Core's SCRIPT_VERIFY_STRICTENC flag and
brings the Bitcoin Core test scripts up to date.


txscript/script.go
```diff
@@ -328,6 +341,109 @@ func IsPushOnlyScript(script []byte) bool {
 	return isPushOnly(pops)
 }
 
+// IsStrictHashType returns whether or not the passed hashtype adheres to
+// the strict encoding requirements.
+func IsStrictHashType(hashType SigHashType) bool {
```
davecgh: This should not be exported.

txscript/script.go
```diff
@@ -328,6 +341,109 @@ func IsPushOnlyScript(script []byte) bool {
...
+	sigHashType := hashType & ^SigHashAnyOneCanPay
+	if sigHashType < SigHashAll || sigHashType > SigHashSingle {
+		return false
+	}
+	return true
+}
+
+// IsStrictPubKey returns whether or not the passed public key adheres to
+// the strict encoding requirements.
+func IsStrictPubKey(pubKey []byte) bool {
```
davecgh: This should not be exported.

txscript/script.go
```diff
@@ -328,6 +341,109 @@ func IsPushOnlyScript(script []byte) bool {
...
+	} else if len(pubKey) == 65 && pubKey[0] == 0x04 {
+		// Uncompressed
+		return true
+	}
+	return false
+}
+
+// IsStrictSignature returns whether or not the passed signature adheres to
+// the strict encoding requirements.
+func IsStrictSignature(sig []byte) bool {
```
davecgh: This should not be exported.

txscript/opcode.go
```diff
@@ -1918,53 +1907,94 @@ func opcodeCheckMultiSig(op *parsedOpcode, s *Script) error {
...
-				pubKeys[curPk], err =
-					btcec.ParsePubKey(pubKeyStrings[curPk],
-						btcec.S256())
-				if err != nil {
-					continue
+	for _, signature := range signatures {
+		script = removeOpcodeByData(script, signature)
+	}
+
+	if len(signatures) != numSignatures {
```
davecgh: This is not the correct behavior.

txscript/opcode.go
```diff
@@ -1918,53 +1907,94 @@ func opcodeCheckMultiSig(op *parsedOpcode, s *Script) error {
...
+			break done
+		}
+
+		// Get the hashtype
+		hashType := SigHashType(rawSignature[len(rawSignature)-1])
+
+		// Trim off the hashtype
+		signature := rawSignature[:len(rawSignature)-1]
+
+		if s.verifyStrictEncoding {
```
davecgh: I'd much rather there be a `checkSignatureEncoding` and `checkPubKeyEncoding` which more closely mirrors BC here (and in `opcodeCheckSig` too) rather than checking the flags in multiple places.  It's much easier to ensure consensus if the same naming and logic is followed.  Also, there is less chance of forgetting to check for a condition in multiple places when it is centralized into the `checkX` functions.  For example, adding the Low S checks will only have to be done in `checkSignatureEncoding`, after checking for the flag,  instead of multiple places.

davecgh:
Commit message "This flag mimics -> These flags mimic".

davecgh:
I pushed a commit with some optimizations for the signature checking to a branch of the same name on my github repo.  Feel free to steal them and squash them into this (after reviewing of course).

txscript/opcode.go
```diff
@@ -1914,56 +1913,102 @@ func opcodeCheckMultiSig(op *parsedOpcode, s *Script) error {
...
-			s.dstack.PushBool(false)
-			return nil
+
+		// Parse the pubkey.
+		parsedPubKey, err := btcec.ParsePubKey(pubKey, btcec.S256())
+		if err != nil {
+			continue
+		}
+
+		hash := calcScriptHash(script, SigHashType(hashType), &s.tx,
```
davecgh: `hashType` is already a `SigHashType`.  No need to cast it again.

davecgh:
OK

davecgh:
I should also note that we've discussed the fact the der flag can be removed now and a few other minor things, but those are planned for separate commits.

jrick:
I've looked over this diff carefully and believe that it does not introduce any consensus bugs.

PR #282, opened 2015-02-06
[btcec] Add generation code for the endomorphism vectors.

davecgh:
This pull request adds code which generates the linearly independent vectors used by the secp256k1 endomorphism code.  These values are hard-coded into the curve already, but having the code used to generate them is handy should any future curves be added which can also make use of the same class of endomorphism.


davecgh:
This was reviewed and blessed by @jrick offline.

PR #283, opened 2015-02-06
[btcec] Don't ignore errs setting consts in btcec init.

davecgh:
This pull request converts the initialization of the constants to use a function which panics on error instead of just ignoring the error.  This is acceptable since they are hard-coded constants and should never fail.


davecgh:
This was reviewed and blessed by @jrick offline.

PR #284, opened 2015-02-06
[btcec] Move the btcec benchmarks into the main package.

davecgh:
The benchmarks are still only compiled when running 'go test' so this has no effect on regular usage.  This is being done because benchmarks often need access to internal state.

Normal tests are kept in a separate package since they also serve to exercise the public API, and by intentionally making it more difficult to reach into the internals, it helps ensure the public API is sanely usable.

Since the benchmarks can now access the internals directly, this commit also removes the functions which exposed the internals to the test package from internal_test.go which were only used by the benchmarks.

Also, it removes a duplicate benchmark.

### DIFF
--- a/btcec/bench_test.go
+++ b/btcec/bench_test.go
@@ -2,26 +2,22 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-package btcec_test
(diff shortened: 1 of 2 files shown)


jrick:
ok

PR #285, opened 2015-02-06
[btcec] Consistency and general cleanup in btcec. 

davecgh:
This pull request contains various modifications for code and comment consistency in the btcec package:
- Call out references at the top and reference them by their identifier in the other comments
- Remove a TODO that no longer applies
- Add comments to the fields in the KoblitzCurve struct and reorder them slightly
- Make comments wrap to 80
- Cleanup code that was far exceeding col 80 (only function declarations typically do this)
- Extend block comments to use as much of the 80 cols as available
- Add a bit more explanation in a couple of places
- Update copyright year on secp256k1.go
- Fix a couple of typos in the comments

### DIFF
--- a/btcec/btcec.go
+++ b/btcec/btcec.go
@@ -9,6 +9,8 @@ package btcec
 // References:
 //   [SECG]: Recommended Elliptic Curve Domain Parameters
 //     http://www.secg.org/sec2-v2.pdf
+//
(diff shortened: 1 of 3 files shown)


davecgh:
The size depends on the curve, so I'd rather it be a field on the curve.  I know the package is a bit too secp256k1-centric at the moment, but I'd prefer not to push it further that direction.

jrick:
Everything else looks ok.

PR #292, opened 2015-02-09
txscript: Add example for manully signing a txout.

davecgh:
This pull request adds a new example to the txscript package that demonstrates creating a new transaction which redeems funds and signing the referenced transaction output the SignTxOutput function.


jrick:
ok

Issue #138, opened 2014-06-11
error message 'No sync peer candidates available' when 7 are available


carryforward:
I have observed an interesting log message which may be indicative of a bug.  When btcd connects to the first node, it starts syncing to it.  It then connected to 7 other nodes. All notes are from outbound connections.  It gets all the blocks then loses connectivity to the first node.  It complains that it cannot sync the blockchain, because there are no peer candidates available, even though it is connected to 7 others.  Only after it connects to a new node does it claims to restart syncing.

This is running from the latest commits as of today.

Feel free to close this if it is expected behavior.

```
user@computer:~$ btcd
18:32:07 2014-06-10 [INF] BTCD: Version 0.8.0-beta
18:32:07 2014-06-10 [INF] BTCD: Loading block database from '/home/user/.btcd/data/mainnet/blocks_leveldb'
18:32:07 2014-06-10 [INF] BTCD: Block database loaded with block height 305063
18:32:07 2014-06-10 [INF] BMGR: Generating initial block node index.  This may take a while...
18:32:09 2014-06-10 [INF] BMGR: Block index generation complete
18:32:09 2014-06-10 [INF] RPCS: RPC server listening on 127.0.0.1:8334
18:32:09 2014-06-10 [INF] SRVR: Server listening on [::]:8333
18:32:09 2014-06-10 [INF] SRVR: Server listening on 0.0.0.0:8333
18:32:09 2014-06-10 [INF] AMGR: Loaded 12004 addresses from '/home/user/.btcd/data/mainnet/peers.json'
18:32:43 2014-06-10 [INF] BMGR: New valid peer 74.103.125.183:8333 (outbound) (/Satoshi:0.8.3/)
18:32:43 2014-06-10 [INF] BMGR: Syncing to block height 305175 from peer 74.103.125.183:8333
18:32:44 2014-06-10 [INF] BMGR: New valid peer 204.27.61.162:8333 (outbound) (/Satoshi:0.9.1/)
18:32:44 2014-06-10 [INF] BMGR: New valid peer 74.194.16.86:8333 (outbound) (/Satoshi:0.9.1/)
18:32:44 2014-06-10 [INF] BMGR: New valid peer 62.45.21.149:8333 (outbound) (/Satoshi:0.9.1/)
18:32:44 2014-06-10 [INF] BMGR: New valid peer 70.185.114.205:8333 (outbound) (/Satoshi:0.8.6/)
[...]

PR #296, opened 2015-02-11
Dynamically update block heights of peers

Roasbeef:
In order to avoid prior situations of [stalled syncs due to outdated peer height data](https://github.com/btcsuite/btcd/issues/138), we now update block heights of peers in real-time as we learn of their announced blocks.

Updates happen when:
   * A peer sends us an orphan block. We update based on the height embedded in the scriptSig for the coinbase tx (BIP 0034).
   * When a peer sends us an inv for a block we already know of.
   * When peers announce new blocks. Subsequent announcements that lost the announcement race are recognized and peer 
     heights are updated accordingly

Additionally, the `getpeerinfo` command has been modified to include both the starting height, and current height of connected peers:
 ```	
......
"startingheight": 322820,
"currentheight": 322837,
```

NOTE: This PR is dependent on: btcsuite/btcjson#43

### CODE REVIEW

File: blockchain/validate.go [hunk 1]
```diff
@@ -574,19 +576,27 @@ func checkSerializedHeight(coinbaseTx *btcutil.Tx, wantHeight int64) error {
...
+
+	return serializedHeight, nil
+}
+
+// checkSerializedHeight checks if the signature script in the passed
+// transaction starts with the serialized block height of wantHeight.
+func checkSerializedHeight(coinbaseTx *btcutil.Tx, wantHeight int64) error {
+	if serializedHeight, err := ExtractCoinbaseHeight(coinbaseTx); err != nil {
+		return err
+	} else if int64(serializedHeight) != wantHeight {
```
davecgh: Please avoid doing else blocks on scoped ifs as it's harder to read and isn't used anywhere else in the code base.

```Go
serializedHeight, err := ExtractCoinbaseHeight(coinbaseTx)
if err != nil {
	return err
}

if int64(serializedHeight) != wantHeight {
...
```

blockmanager.go
```diff
@@ -579,8 +579,33 @@ func (b *blockManager) handleBlockMsg(bmsg *blockMsg) {
...
+	// recognition of an orphan, we also use this information to update
+	// the block heights over other peers who's invs may have been ignored
+	// if we are actively syncing while the chain is not yet current or
+	// who may have lost the lock announcment race.
+	var heightUpdate int32
+	var blkShaUpdate *wire.ShaHash
+
 	// Request the parents for the orphan block from the peer that sent it.
 	if isOrphan {
+		// TODO(roasbeef): Only check ver 2+?
```
davecgh: Yes, for sure.

peer.go
```diff
@@ -209,6 +211,29 @@ func (p *peer) isKnownInventory(invVect *wire.InvVect) bool {
...
 }
 
+// UpdateLastBlockHeight updates the last known block for the peer. It is safe
+// for concurrent access.
+func (p *peer) UpdateLastBlockHeight(newHeight int32) {
+	p.StatsMtx.Lock()
+	defer p.StatsMtx.Unlock()
+
+	//if newHeight > p.lastBlock {
+	peerLog.Infof("Updating last block height of peer %v from %v to %v",
```
davecgh: I know this is here for debug, but adding this as a reminder to not have it as Infof.

peer.go
```diff
@@ -209,6 +211,29 @@ func (p *peer) isKnownInventory(invVect *wire.InvVect) bool {
...
+	//}
+}
+
+// UpdateLastAnnouncedBlock updates meta-data about the last block sha this
+// peer is known to have announced. It is safe for concurrent access.
+func (p *peer) UpdateLastAnnouncedBlock(blkSha *wire.ShaHash) {
+	p.StatsMtx.Lock()
+	defer p.StatsMtx.Unlock()
+
+	peerLog.Debugf("Updating last blk for peer %v, %v", p.addr, blkSha)
```
davecgh: This is probably a Tracef.

Roasbeef: Indeed, demoted to a `Tracef`.

peer.go
Roasbeef: Yeah it was a left over logging statement for tracing through execution. I've demoted it a `Tracef` log statement.

blockmanager.go
Roasbeef: Done.

blockchain/validate.go
Roasbeef: Noted. Change this usage and another introduced in the `blockmanager`.

blockmanager.go
```diff
@@ -579,8 +579,39 @@ func (b *blockManager) handleBlockMsg(bmsg *blockMsg) {
...
+	var blkShaUpdate *wire.ShaHash
+
 	// Request the parents for the orphan block from the peer that sent it.
 	if isOrphan {
+		// We've just received an orphan block from a peer. In order
+		// to update the height of the peer, we try to extract the
+		// block height from the scriptSig of the coinbase transaction.
+		// Extraction is only attempted if the block's version is
+		// high enough (ver 2+).
+		if blockchain.ShouldHaveSerializedBlockHeight(&bmsg.block.MsgBlock().Header) {
```
davecgh: This is definitely going way past the 80 mark.  A few chars here and there I don't say too much about, but this one is pushing 100.

Roasbeef: Fixed.

Roasbeef:
Squashed and rebased to bring in latest repo merges from master.

blockmanager.go
```diff
@@ -873,6 +913,36 @@ func (b *blockManager) handleInvMsg(imsg *invMsg) {
...
+		return
+	}
+
+	// If our chain is current and a peer announces a block we already
+	// know of, then update their current block height.
+	if lastBlock != -1 && b.current() {
+		exsits, err := b.server.db.ExistsSha(&invVects[lastBlock].Hash)
+		if err == nil && exsits {
+			blkHeight, err := b.server.db.FetchBlockHeightBySha(&invVects[lastBlock].Hash)
+			if err != nil || blkHeight == 0 {
```
jrick: Why the check for 0?  If it is 0 for some reason (without err=nil) it will print something ugly like "<nil>" when it logs this below.

blockmanager.go
```diff
@@ -873,6 +913,36 @@ func (b *blockManager) handleInvMsg(imsg *invMsg) {
...
+	// Ignore invs from peers that aren't the sync if we are not current.
+	// Helps prevent fetching a mass of orphans.
+	if imsg.peer != b.syncPeer && !b.current() {
+		return
+	}
+
+	// If our chain is current and a peer announces a block we already
+	// know of, then update their current block height.
+	if lastBlock != -1 && b.current() {
+		exsits, err := b.server.db.ExistsSha(&invVects[lastBlock].Hash)
```
jrick: typo

blockchain/validate.go
```diff
@@ -556,16 +565,18 @@ func CheckBlockSanity(block *btcutil.Block, powLimit *big.Int, timeSource Median
...
 }
 
-// checkSerializedHeight checks if the signature script in the passed
-// transaction starts with the serialized block height of wantHeight.
-func checkSerializedHeight(coinbaseTx *btcutil.Tx, wantHeight int64) error {
+// ExtractCoinbaseHeight attempts to extract the height of the block
+// from the scriptSig of a coinbase transaction. Coinbase heights
+// are only present in blocks of version 2 or later. This was added as part of
+// BIP0034.
+func ExtractCoinbaseHeight(coinbaseTx *btcutil.Tx) (uint64, error) {
```
jrick: This should probably return a signed int (I'd argue for int32, but that ship has sailed).

blockmanager.go
```diff
@@ -618,6 +654,16 @@ func (b *blockManager) handleBlockMsg(bmsg *blockMsg) {
 		}
 	}
 
+	// Update the block height for this peer. But only send a message to
+	// the server for updating peer heights if this is an orphan or our
+	// chain is "current". This avoid sending a spammy amount of messages
+	// if we're syncing the chain from scratch.
+	if blkShaUpdate != nil && heightUpdate != 0 {
+		bmsg.peer.UpdateLastBlockHeight(int32(heightUpdate))
```
jrick: This int32 cast is unneeded.

server.go
```diff
@@ -185,6 +199,25 @@ func (p *peerState) forAllPeers(closure func(p *peer)) {
 	p.forAllOutboundPeers(closure)
 }
 
+// handleUpdatePeerHeight updates the heights of all peers who were known to
+// announce a block we recently accepted.
+func (s *server) handleUpdatePeerHeight(state *peerState, umsg updatePeerHeightMsg) {
+	state.forAllPeers(func(p *peer) {
```
jrick: Why must we iterate all peers looking for block heights to update, when only a single peer needs this update?

blockmanager.go
Roasbeef: Agreed that the check for 0 is unnecessary, and can result in an unhelpful error log similar to a prior case in involving the `addrindex`. I was previously also checking for zero since in the case of an error, `db. FetchBlockHeightBySha` will return `0` for the height *and* and a non-nil error: 

```go
func (db *LevelDb) getBlkLoc(sha *wire.ShaHash) (int64, error) {
	key := shaBlkToKey(sha)

	data, err := db.lDb.Get(key, db.ro)
	if err != nil {
		if err == leveldb.ErrNotFound {
			err = database.ErrBlockShaMissing
		}
		return 0, err
	}
...
```

However, since the cases are coupled together it's sufficient to only check for a non-nil error. I've updated this fragment to only check for the error.

blockmanager.go
Roasbeef: Fixed.

blockchain/validate.go
Roasbeef: Good point, changed to instead return an `int64`. Choosing an `int64` since in `checkSerializedHeight` the height to be compared to is an `int64`. I agree we should be using an `int32`, but I'd rather maintain consistency with the rest of the codebase.  Within `blockchain` and `database` block heights are represented as an `int64`

blockmanager.go
Roasbeef: Fixed to remove the unnecessary cast.

server.go
Roasbeef: The source peer, whose new block we just accepted has already had its height updated at this point. The purpose of the function is to update the heights of all other peers who lost the block announcement race. I've pluralized 'Height' in the function to further clarify its behavior.

server.go
```diff
@@ -185,6 +199,25 @@ func (p *peerState) forAllPeers(closure func(p *peer)) {
...
+		if p == umsg.originPeer {
+			return
+		}
+
+		// If the peer has recently announced a block, and this block
+		// matches our newly accepted block, then update their block
+		// height.
+		if p.lastAnnouncedBlock != nil && bytes.Equal(
+			p.lastAnnouncedBlock.Bytes(), umsg.newSha.Bytes()) {
+			p.UpdateLastBlockHeight(umsg.newHeight)
```
jrick: I think lastAnnouncedBlock should be nilled here, after the peer's height is updated.

Roasbeef: Great suggestion, will skip the additional check if the peer hasn't announced a block since the last round. `lastAnnouncedBlock` now gets nilled after a peer's height has been updated. 

Nice catch, the read to `lastAnnouncedBlock` is now protected by the peer's mutex.

davecgh:
This is conflicting with master currently.  Please rebase and resolve the merge conflicts along with @jrick's comment.  I'd like to get this in today for the upcoming release.

Roasbeef:
Rebased to latest master, addressed the final comment, and squashed down into a single commit.

davecgh:
I've run into a few issues with this branch and I don't have time right now to chase them down, so I'm going to refrain from putting this in the release.

davecgh:
Alright, so I manged to track down the issues I was having and they are not related to this branch after all.

davecgh:
This PR is ready after rebase.

Roasbeef:
Done, build is all green :+1:

PR #298, opened 2015-02-12
btcec: Optimize pre-computed table load.

### LINKED ISSUE #297: secp256k1.go appears to be causing some errors for Google App Engine
The long []byte array in the file appears to be causing some issues for Google App Engine - https://groups.google.com/forum/#!topic/google-appengine-go/YRsH3cWdql8 .

Is that data necessary for the library to function, or does it simply speed the process up?

davecgh:
This pull request modifies the pre-computed table used to optimize the secp256k1 scalar multiplication to a string instead of a byte slice.  This change makes the compile more efficient since the Go compiler internally
represents bytes slices inefficiently.

This reduces the memory needed to compile btcec to 3MB versus the previous 40MB before this change.

In addition, it modifies the code which loads the pre-computed table to deserialize directly into the table instead of into locals that are then copied.

Fixes #297.


jrick:
Why not use the `io.Reader` returned by `base64.NewDecoder` with a passed in `strings.NewReader(secp256k1BytePoints)`?  That should avoid the need to create a copy of the entire string as a byte slice each startup.

davecgh:
Done.

PR #299, opened 2015-02-12
blockchain: Just fetch the height

dajohi:
We only care about the height, so just fetch the height instead of
the entire block.


davecgh:
OK

PR #300, opened 2015-02-13
txscript: Add new flag ScriptVerifyMinimalData

dajohi:
The ScriptVerifyMinimalData enforces that all push operations use the
minimal data push required.  This is part of BIP0062.

This commit mimics Bitcoin Core commit
698c6abb25c1fbbc7fa4ba46b60e9f17d97332ef


davecgh:
I'm reviewing this currently, but one thing that stands out is that `PeekInt` also needs the same treatment.

txscript/script.go
```diff
@@ -653,6 +658,10 @@ const (
 	// to compily with the DER format.
 	ScriptVerifyDERSignatures
 
+	// ScriptVerifyMinimalData defines that signatures must use the smallest
+	// push operator.
+	ScriptVerifyMinimalData
```
davecgh: Please call out that this is BIP0062 rules 3 and 4.

txscript/opcode.go
```diff
@@ -963,6 +998,14 @@ func (pop *parsedOpcode) exec(s *Script) error {
 	if s.condStack[0] != OpCondTrue && !pop.conditional() {
 		return nil
 	}
+
+	if s.dstack.verifyMinimalData && s.condStack[0] == OpCondTrue &&
```
davecgh: Please comment this.  Something like:

```Go
// Ensure all executed data push opcodes use the minimal encoding when the
// minimal data verification is set.
```

txscript/stack.go
```diff
@@ -132,6 +133,25 @@ func (s *Stack) PopInt() (*big.Int, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	if s.verifyMinimalData && len(so) > 0 {
```
davecgh: This code is identical, so please refactor it into a function that takes the byte array as a parameter.

davecgh:
Looks good now. 

OK.

PR #302, opened 2015-02-14
txscript: Remove ScriptCanonicalSignatures

dajohi:
Remove ScriptCanonicalSignatures and use the new
ScriptVerifyDERSignatures flag.  The ScriptVerifyDERSignatures
flag accomplishes the same functionality.


davecgh:
OK

PR #306, opened 2015-02-18
Simplify chain configuration.

dajohi:


blockchain/chain.go
```diff
@@ -610,10 +610,12 @@ func (b *BlockChain) pruneBlockNodes() error {
 
 // isMajorityVersion determines if a previous number of blocks in the chain
 // starting with startNode are at least the minimum passed version.
-func (b *BlockChain) isMajorityVersion(minVer int32, startNode *blockNode, numRequired, numToCheck uint64) bool {
+func (b *BlockChain) isMajorityVersion(minVer int32, startNode *blockNode,
+	numRequired uint64) bool {
```
davecgh: Please either unwrap the function signature, or add a blank line after it to be consistent with the rest of the code.

blockchain/accept.go
```diff
@@ -111,9 +111,7 @@ func (b *BlockChain) maybeAcceptBlock(block *btcutil.Block, flags BehaviorFlags)
 		// upgraded.  This is part of BIP0034.
 		if blockHeader.Version < 2 {
 			if b.isMajorityVersion(2, prevNode,
-				b.chainParams.BlockV1RejectNumRequired,
-				b.chainParams.BlockV1RejectNumToCheck) {
-
```
davecgh: Please don't nuke this extra line.  It's there for visual readability since the indentation is the same on the wrapped if.

chaincfg/params.go
```diff
@@ -176,18 +174,15 @@ var RegressionNetParams = Params{
...
+	// Enforce current block version once majority of the network has
+	// upgraded.
 	// 51% (51 / 100)
-	// This is part of BIP0034.
-	CoinbaseBlockHeightNumRequired: 51,
-	CoinbaseBlockHeightNumToCheck:  100,
+	// Reject previous block versions once a majority of the network has
+	// upgraded.
+	// 75% (75 / 100)
+	BlockEnforceNumRequired: 51,
```
davecgh: Since the regression network is shared with BC, these should be updated to match.

davecgh:
After those comments are addressed, this will be good to go.

davecgh:
OK

PR #311, opened 2015-02-19
cmd/btcctl: Rewrite btcctl to use the new features of btcjson.

### LINKED ISSUE #305: btcctl command help: decoderawtransaction misnamed
This not a big deal but from ./btcctl
`
Usage:
  btcctl [OPTIONS]
...
Commands:
...
	debuglevel <levelspec>
	decoderawtransaction <txhash>
	decodescript <hex>
`
It should be` decoderawtransaction <hex>`, not `decoderawtransaction <txhash>`

davecgh:
This pull request contains what is essentially a complete rewrite of the btcctl utility to make use of the new features provided by the latest version btcjson and improve several things along the way.  The following summarizes the changes:

- The supported commands and handling now come directly from btcjson, so it is no longer necessary to manually add new commands.  Once a command has been registered with btcjson, it will automatically become usable by btcctl complete with full error handling (once it is re-compiled of  course)
- Rather than dumping the entire list of commands on every error, the user now must specifically request the list of commands via the -l option
- The list of commands is now categorized by chain and wallet and alphabetized
- The help flag now only shows the help options instead of also dumping all of the commands
- The error display on valid commands with invalid parameters has been greatly improved to show the specific parameter number, reason, and error code
- When a valid command is specified with invalid parameter, only the usage for that specific command is shown now
- It is now possible to use a SOCKS5 proxy for connection
- The output of commands has been improved in the following ways:
  - Strings on commands such as `getbestblockhash` no longer have quotes wrapped around them
  - Fields that are integers no longer show in scientific notation when they are large (timestamps for example)

This closes #305 as a side effect.

### DIFF
--- a/cmd/btcctl/httpclient.go
+++ b/cmd/btcctl/httpclient.go
@@ -1,0 +1,128 @@
+package main
+
+import (
+	"bytes"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"net/http"
+
+	"github.com/btcsuite/btcd/btcjson/v2/btcjson"
+	"github.com/btcsuite/go-socks/socks"
+)
+
+// newHTTPClient returns a new HTTP client that is configured according to the
+// proxy and TLS settings in the associated connection configuration.
+func newHTTPClient(cfg *config) (*http.Client, error) {
+	// Configure proxy if needed.
+	var dial func(network, addr string) (net.Conn, error)
+	if cfg.Proxy != "" {
+		proxy := &socks.Proxy{
+			Addr:     cfg.Proxy,
+			Username: cfg.ProxyUser,
+			Password: cfg.ProxyPass,
+		}
+		dial = func(network, addr string) (net.Conn, error) {
+			c, err := proxy.Dial(network, addr)
+			if err != nil {
+				return nil, err
+			}
+			return c, nil
+		}
+	}
+
+	// Configure TLS if needed.
+	var tlsConfig *tls.Config
+	if !cfg.NoTLS && cfg.RPCCert != "" {
+		pem, err := ioutil.ReadFile(cfg.RPCCert)
+		if err != nil {
+			return nil, err
+		}
+
+		pool := x509.NewCertPool()
+		pool.AppendCertsFromPEM(pem)
+		tlsConfig = &tls.Config{
+			RootCAs:            pool,
+			InsecureSkipVerify: cfg.TLSSkipVerify,
+		}
+	}
+
+	// Create and return the new HTTP client potentially configured with a
+	// proxy and TLS.
+	client := http.Client{
+		Transport: &http.Transport{
+			Dial:            dial,
+			TLSClientConfig: tlsConfig,
+		},
+	}
+	return &client, nil
+}
+
+// sendPostRequest sends the marshalled JSON-RPC command using HTTP-POST mode
+// to the server described in the passed config struct.  It also attempts to
+// unmarshal the response as a JSON-RPC response and returns either the result
+// field or the error field depending on whether or not there is an error.
+func sendPostRequest(marshalledJSON []byte, cfg *config) ([]byte, error) {
+	// Generate a request to the configured RPC server.
+	protocol := "http"
+	if !cfg.NoTLS {
+		protocol = "https"
+	}
+	url := protocol + "://" + cfg.RPCServer
+	bodyReader := bytes.NewReader(marshalledJSON)
+	httpRequest, err := http.NewRequest("POST", url, bodyReader)
+	if err != nil {
+		return nil, err
+	}
+	httpRequest.Close = true
+	httpRequest.Header.Set("Content-Type", "application/json")
+
+	// Configure basic access authorization.
+	httpRequest.SetBasicAuth(cfg.RPCUser, cfg.RPCPassword)
+
+	// Create the new HTTP client that is configured according to the user-
+	// specified options and submit the request.
+	httpClient, err := newHTTPClient(cfg)
+	if err != nil {
+		return nil, err
+	}
+	httpResponse, err := httpClient.Do(httpRequest)
+	if err != nil {
+		return nil, err
+	}
+
+	// Read the raw bytes and close the response.
+	respBytes, err := ioutil.ReadAll(httpResponse.Body)
+	httpResponse.Body.Close()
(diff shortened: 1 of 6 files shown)


davecgh:
Any feedback on this?

cmd/btcctl/httpclient.go
```diff
@@ -0,0 +1,132 @@
...
+			Dial:            dial,
+			TLSClientConfig: tlsConfig,
+		},
+	}
+	return &client, nil
+}
+
+// sendPostRequest sends the marshalled JSON-RPC command using HTTP-POST mode
+// to the server described in the passed config struct.  It also attempts to
+// unmarshals the response as a JSON-RPC response and returns either the result
```
Roasbeef: unmarshals -> unmarshal

Roasbeef:
Unsure if I have something incorrectly configured package-wise on my side, but when I attempt to pass the optional `verbose` parameter to `searchrawtransactions` (for both cases), I get the following error: 

```bash
➜  ~  btcctl searchrawtransactions mj2TT1WakMWYQHu9Ap6jNq6BqUXd4VPYvv false
-32700: second optional parameter 'verbose' must be an int: json: cannot unmarshal bool into Go value of type int
➜  ~  btcctl searchrawtransactions mj2TT1WakMWYQHu9Ap6jNq6BqUXd4VPYvv true
-32700: second optional parameter 'verbose' must be an int: json: cannot unmarshal bool into Go value of type int
```

However, when I try out other commands that take a `bool` for an optional `verbose` parameter, they parse and return successfully.

cmd/btcctl/httpclient.go
```diff
@@ -0,0 +1,132 @@
...
+			c, err := proxy.Dial(network, addr)
+			if err != nil {
+				fmt.Println("Proxy error", err)
+				return nil, err
+			}
+			fmt.Println("proxy dial complete")
+			return c, nil
+		}
+
+		// TODO(davec): Get the proxy working.
```
Roasbeef: Friendly reminder to address the `TODO`, and commented out sections below and above before merging. However, looking at the function assignment above, it looks as if you've already done this. `dial` gets initialized to a version which dials the `SOCKS5` proxy as an intermediary step.

davecgh:
Recall that `searchrawtransactions` was changed to take an int, not a bool.  I would much rather it take a bool, but that is not consistent with the BC branch and it has to be an int to be compatible.

cmd/btcctl/httpclient.go
davecgh: Thanks, I did forget to finish that.

Roasbeef:
Right, I remember the switch. But when I pass in integer values for `verbose` I get the same error message. 

```bash
➜  ~  btcctl searchrawtransactions 2MwWtjYAEupM5YRiSEKDn11SkYeCiEyLfoX 1
-32700: second optional parameter 'verbose' must be an int: json: cannot unmarshal bool into Go value of type int
➜  ~  btcctl searchrawtransactions 2MwWtjYAEupM5YRiSEKDn11SkYeCiEyLfoX 0
-32700: second optional parameter 'verbose' must be an int: json: cannot unmarshal bool into Go value of type int
``` 

After digging a bit closer, it looks like this error is a result of `rpcserver.go` still using legacy `btcjsonv1` methods when parsing commands: 
```go
func jsonRPCRead(w http.ResponseWriter, r *http.Request, s *rpcServer) {
....
	var reply btcjson.Reply
	cmd, jsonErr := parseCmd(body)
	if cmd != nil {
		// Unmarshaling at least a valid JSON-RPC message succeeded.
		// Use the provided id for errors.
		id := cmd.Id()
		reply.Id = &id
	}
	if jsonErr != nil {
		reply.Error = jsonErr
....
```

Which leads to:
```go 
func parseCmd(b []byte) (btcjson.Cmd, *btcjson.Error) {
	cmd, err := btcjson.ParseMarshaledCmd(b) <--
	if err != nil {
		jsonErr, ok := err.(btcjson.Error)
		if !ok {
			jsonErr = btcjson.Error{
				Code:    btcjson.ErrParse.Code,
				Message: err.Error(),
			}
		}
		return cmd, &jsonErr
	}
.....
```

I believe the  error i'm seeing is the result of `btcjson.ParseMarshaledCmd` failing.

davecgh:
Hmmm.  It works properly for me when using `0` and `1`.

```bash
$ btcctl --testnet searchrawtransactions 2MwWtjYAEupM5YRiSEKDn11SkYeCiEyLfoX 1
[
  {
```

davecgh:
Ok, I know what it is.  I already fixed the command on my btcjsonv2 integration branch, but the version of btcjsonv2 in master still has the `SearchRawTransactionsCmd` as a bool instead of an int.

https://github.com/btcsuite/btcd/blob/cb7c24141aee94f82adb57d6dd0c9fa2b45c1043/v2/btcjson/chainsvrcmds.go#L516

davecgh:
I've pulled that change into this branch now, so if you rebuild `btcd` and `btcctl` on this branch, that will be resolved.

Roasbeef:
Nice, yeah everything works properly on my end now :+1:

cmd/btcctl/httpclient.go
```diff
@@ -0,0 +1,132 @@
...
+	respBytes, err := ioutil.ReadAll(httpResponse.Body)
+	httpResponse.Body.Close()
+	if err != nil {
+		err = fmt.Errorf("error reading json reply: %v", err)
+		return nil, err
+	}
+
+	// Handle unsuccessful HTTP responses
+	if httpResponse.StatusCode < 200 || httpResponse.StatusCode >= 300 {
+		return nil, fmt.Errorf("%s", respBytes)
```
jrick: It would be good to include the error code in the returned error.

jrick:
Couple things:

1. Why is listCommands defined in config.go? Doesn't look like it touches any config bits.
2. Converting the response bytes to a string before printing the result seems a little unnecessary, but performance isn't a concern here.  Still, it stood out as odd to me.

davecgh:
1. `listCommands` is defined in that file since it's the only place it's called from (in `loadConfig`)
2. I think it reads cleaner converting it to a string, but I'll change it over.

cmd/btcctl/httpclient.go
davecgh: Standard practice is for the HTTP to return the error in the body.  It might be worthwhile to check for an empty response and print it out in that case.

jrick:
ok

PR #312, opened 2015-02-23
Reject free/low-fee transactions with insufficient priority.

dajohi:
By default, have the mempool reject free and low-fee transactions that
have insufficient priority to be mined in the next block.

Addtionally, add a new configuration option, -norelaypriority, to
disable the check.

### DIFF
--- a/mempool.go
+++ b/mempool.go
@@ -1179,6 +1179,24 @@ func (mp *txMemPool) maybeAcceptTransaction(tx *btcutil.Tx, isNew, rateLimit boo
 		return nil, txRuleError(wire.RejectInsufficientFee, str)
 	}
 
+	// Require that free transactions have sufficient priority to be mined
+	// in the next block.
+	if !cfg.NoRelayPriority && txFee < minFee {
+		txD := &TxDesc{
+			Tx:     tx,
+			Added:  time.Now(),
+			Height: curHeight,
+			Fee:    txFee,
+		}
+		currentPriority := txD.CurrentPriority(txStore, nextBlockHeight)
+		if currentPriority <= minHighPriority {
+			str := fmt.Sprintf("transaction %v has insufficient "+
+				"priority (%g <= %g)", txHash,
+				currentPriority, minHighPriority)
+			return nil, txRuleError(wire.RejectInsufficientFee, str)
+		}
+	}
+
 	// Free-to-relay transactions are rate limited here to prevent
 	// penny-flooding with tiny transactions as a form of attack.
 	if rateLimit && txFee < minFee {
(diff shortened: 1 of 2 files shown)


dajohi:
This mimics bitcoin core.
https://github.com/bitcoin/bitcoin/blob/master/src/main.cpp#L1037-L1040

mempool.go
```diff
@@ -1132,6 +1132,25 @@ func (mp *txMemPool) maybeAcceptTransaction(tx *btcutil.Tx, isNew, rateLimit boo
...
+			Added:  time.Now(),
+			Height: curHeight,
+			Fee:    txFee,
+		}
+		if txFee < minFee &&
+			txD.CurrentPriority(txStore, nextBlockHeight) <
+				minHighPriority {
+
+			str := fmt.Sprintf("transaction %v has insufficient "+
+				"priority ", txHash)
```
jrick: Extra space at the end

mempool.go
```diff
@@ -1132,6 +1132,25 @@ func (mp *txMemPool) maybeAcceptTransaction(tx *btcutil.Tx, isNew, rateLimit boo
...
+	// Require that free transactions have sufficient priority to be mined
+	// in the next block.
+	if !cfg.NoRelayPriority {
+		txD := &TxDesc{
+			Tx:     tx,
+			Added:  time.Now(),
+			Height: curHeight,
+			Fee:    txFee,
+		}
+		if txFee < minFee &&
```
jrick: I'd move this first part of the check into the if statement above, to avoid creating the TxDesc when this is false.

mempool.go
```diff
@@ -1132,6 +1132,22 @@ func (mp *txMemPool) maybeAcceptTransaction(tx *btcutil.Tx, isNew, rateLimit boo
...
+	// in the next block.
+	if !cfg.NoRelayPriority && txFee < minFee {
+		txD := &TxDesc{
+			Tx:     tx,
+			Added:  time.Now(),
+			Height: curHeight,
+			Fee:    txFee,
+		}
+		if txD.CurrentPriority(txStore, nextBlockHeight) <= minHighPriority {
+			str := fmt.Sprintf("transaction %v has insufficient "+
```
davecgh: I think the error message should include the current priority it has and the required priority.

mempool.go
```diff
@@ -1132,6 +1132,22 @@ func (mp *txMemPool) maybeAcceptTransaction(tx *btcutil.Tx, isNew, rateLimit boo
...
+	// Require that free transactions have sufficient priority to be mined
+	// in the next block.
+	if !cfg.NoRelayPriority && txFee < minFee {
+		txD := &TxDesc{
+			Tx:     tx,
+			Added:  time.Now(),
+			Height: curHeight,
+			Fee:    txFee,
+		}
+		if txD.CurrentPriority(txStore, nextBlockHeight) <= minHighPriority {
```
davecgh: Should this be <?  It should be acceptable with the minimum.

davecgh:
Once the error bit is addressed and this is rebased, it's good to go.

mempool.go
jrick: The < in the error message should also be a <=, to match the program logic.

davecgh: Right.

davecgh:
OK

PR #313, opened 2015-02-23
Correct documentation for searchrawtransactions

Roasbeef:
`searchrawtransactions` can return multiple transactions. Previously the documentation erroneously had return examples showing only a single tx for each case. 
  * The documentation previously presented the output for the `non-verbose` case as a single json string, instead of an array of json strings. 
  * Similarly, for the `verbose` mode, the documentation detailed the output as a single json object, instead of an array of json objects.


davecgh:
OK

PR #314, opened 2015-02-24
rpcserver: Convert to make use of new btcjson.

davecgh:
This pull request converts the RPC server over to use the new features available in the latest version of btcjson and improve a few things along the way. This following summarizes the changes:

- All btcjson imports have been updated to the latest package version
- The help has been significantly improved
  - Invoking help with no command specified now provides an alphabetized list of all supported commands along with one-line usage
  - The help for each command is automatically generated and provides much more explicit information such as the type of each parameter, whether or not it's optional or required, etc
  - The websocket-specific commands are now provided when accessing the help when connected via websockets
  - Help has been added for all websocket-specific commands and is only accessible when connected via websockets
- The error returns and handling of both the standard and websocket handlers has been made consistent
- All RPC errors have been converted to the new RPCError type
- Various variables have been renamed for consistency
- Several RPC errors have been improved
- The commands that are marked as unimplemented have been moved into the
  separate map where they belong
- Several comments have been improved
- An unnecessary check has been removed from the createrawtransaction handler
- The command parsing has been restructured a bit to pave the way for JSON-RPC 2.0 batching support
- The order of several functions has been udpated to more closely match the ordering used by the rest of the base base such that functions are typically defined before they are used and near their call site
- When invalid HTTP Basic Access Authentication details are provided to the websocket endpoint, return a WWW-Authenticate header just like the non-websocket endpoint
- When a connection to the websocket endpoint fails to properly upgrade, return a 400 Bad Request HTTP error status code

### DIFF
--- a/rpcwebsocket.go
+++ b/rpcwebsocket.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2013-2014 Conformal Systems LLC.
+// Copyright (c) 2013-2015 Conformal Systems LLC.
 // Use of this source code is governed by an ISC
 // license that can be found in the LICENSE file.
 
@@ -19,8 +19,7 @@ import (
 
 	"golang.org/x/crypto/ripemd160"
 
-	"github.com/btcsuite/btcd/btcjson"
-	"github.com/btcsuite/btcd/btcjson/btcws"
+	"github.com/btcsuite/btcd/btcjson/v2/btcjson"
 	"github.com/btcsuite/btcd/database"
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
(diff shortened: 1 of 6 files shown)


jrick:
Could we generate the help strings in a 'go generate' step, instead of at init each time?

davecgh:
They aren't generated at init.  They're only generated when requested, and then cached.

jrick:
Well, the map is still created at init. :)

rpcwebsocket.go
```diff
@@ -478,31 +480,36 @@ func (m *wsNotificationManager) notifyForNewTx(clients map[chan struct{}]*wsClie
...
+			if marshalledJSONVerbose != nil {
+				wsc.QueueNotification(marshalledJSONVerbose)
+				continue
+			}
 
+			net := m.server.server.chainParams
+			rawTx, err := createTxRawResult(net, txShaStr, mtx, nil,
+				0, nil)
+			if err != nil {
+				return
```
jrick: This is just shuffled code, but shouldn't this error be logged?

rpcwebsocket.go
```diff
@@ -1376,34 +1374,80 @@ func newWebsocketClient(server *rpcServer, conn *websocket.Conn,
...
-		ntfnChan:      make(chan []byte, 1),      // nonblocking sync
-		asyncChan:     make(chan btcjson.Cmd, 1), // nonblocking sync
+		ntfnChan:      make(chan []byte, 1),        // nonblocking sync
+		asyncChan:     make(chan *parsedRPCCmd, 1), // nonblocking sync
 		sendChan:      make(chan wsResponse, websocketSendBufferSize),
 		quit:          make(chan struct{}),
 	}
 }
 
+// handleWebsocketHelp implement the help command for websocket connections.
```
Roasbeef: implement -> implements

davecgh:
The map is referenced at run-time in order to dynamically generate the help as needed.  We can't nil it.  As mentioned before creating code to use `go generate` is not a bad optimization, but it won't make it into this pr.

rpcwebsocket.go
davecgh: Fixed.

davecgh:
All review items have been addressed.

davecgh:
Rebased and updated to latest master.

davecgh:
This was merged, but not marked properly by github.

PR #172, opened 2014-08-29
Add config option to disable bloom filtering.

dajohi:

PR #317, opened 2015-02-25
Add new option --nobloom.

dajohi:
Add a new configuration option --nobloom that disables bloom
filtering (BIP0037) support.  Bloom filtering is very CPU
intensive, so a user might wish to specify this option to lower
CPU utilization at the cost of increased bandwidth usage.  However,
this option also increases both CPU utilization and bandwidth usage
for SPV clients.

refs #172

### DIFF
--- a/peer.go
+++ b/peer.go
@@ -592,10 +592,31 @@ func (p *peer) pushMerkleBlockMsg(sha *wire.ShaHash, doneChan, waitChan chan str
 		return err
 	}
 
-	// Generate a merkle block by filtering the requested block according
-	// to the filter for the peer and fetch any matched transactions from
-	// the database.
-	merkle, matchedHashes := bloom.NewMerkleBlock(blk, p.filter)
+	var merkle *wire.MsgMerkleBlock
+	var matchedHashes []*wire.ShaHash
+	if cfg.NoBloom {
+		numTx := len(blk.Transactions())
+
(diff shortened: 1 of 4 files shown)


peer.go
```diff
@@ -554,10 +554,29 @@ func (p *peer) pushMerkleBlockMsg(sha *wire.ShaHash, doneChan, waitChan chan str
...
+		// original block.
+		header := blk.MsgBlock().Header
+		numTxs := len(blk.Transactions())
+
+		merkle = wire.NewMsgMerkleBlock(&header)
+		merkle.Transactions = uint32(numTxs)
+		for _, tx := range blk.Transactions() {
+			merkle.Hashes = append(merkle.Hashes, tx.Sha())
+		}
+		merkle.Flags = bytes.Repeat([]byte{0x01}, numTxs)
```
davecgh: This is not correct.  That Flags parameter of a merkle block is a bitset.  This should therefore choosing the correct number of bytes based on the number of bits needed to represent each transaction and setting trailing zeros in the final byte for transactions which are not present.

davecgh:
This isn't ready yet since the merkle block being served currently has invalid flags.

davecgh:
I'd rather wait until `SFNodeBloom` service bit is in place along with the protocol version bump and morph this pull request into working properly with that.

dajohi:
Close in favor of #499.

PR #319, opened 2015-02-25
blockchain: Determine script flags sooner.

davecgh:
This pull request moves the definition of the flags which are needed to check transaction scripts higher up the call stack to pave the way for adding support for v3 blocks.  While here, also spruce up a couple of sections.

There are no functional changes in this pull request.


jrick:
ok

PR #322, opened 2015-02-26
Use ScriptVerifyMinimalData

dajohi:
Additionally, drop HasCanoncialPushes as ScriptVerifyMinimalData
offers more complete checks.


davecgh:
Since the internal function for `hasCanonicalPushes` still exists, please don't remove the tests for it.  Instead, you can export the internal function only when the tests are running as `TstHasCanonicalPushes` and call that instead.  You'll probably have to do the same with `parseScript` since the current `HasCanonicalPushes` calls it and it's not exported.

davecgh:
Looks good now.  OK

PR #325, opened 2015-02-26
txscript: export StandardVerifyFlags

dajohi:
By exporting StandardVerifyFlags, clients can ensure they create
transactions that btcd will accept into its mempool.

This flag doesn't belong in txscript.  It belongs in a
policy package.  However, this is currently the least worse place.


davecgh:
OK

PR #326, opened 2015-02-27
btcjson: Fix a bug in btcjson v2

dajohi:
btcjson v2 switched an optional field to mandatory which resulted
in a nil deference.  This switches it back to optional.


jrick:
ok

PR #331, opened 2015-02-28
txscript: Increase maximum allowed nulldata bytes

dajohi:
This change increases the maximum allowed bytes allowed in pushed
data to be considered a nulldata transaction.  This matches the current
value the reference implementation uses by default.


txscript/script.go
```diff
@@ -303,7 +307,8 @@ func isMultiSig(pops []parsedOpcode) bool {
 // false otherwise.
 func isNullData(pops []parsedOpcode) bool {
 	// A nulldata transaction is either a single OP_RETURN or an
-	// OP_RETURN SMALLDATA (where SMALLDATA is a push data up to 40 bytes).
+	// OP_RETURN SMALLDATA (where SMALLDATA is a push data up to
+	// maxDataCarrierSize bytes.
```
davecgh: Looks like this is missing the `)` before the period.

davecgh:
OK once the minor ) typo is fixed.

PR #332, opened 2015-02-28
Use network adjusted time for finalized tx check.

davecgh:
This pull request modifies finalized transaction check used by the memory pool and block templates to use the network adjusted time instead of the unadjusted local time.  This helps keep the transactions accepted to the
memory pool, and hence allowed to relay, more consistent across nodes.


dajohi:
OK

PR #334, opened 2015-03-01
rpcserver: Use default values for getnetworkhashps.

davecgh:
When the fields in the command for the getnetworkhashps RPC don't have the fields set, use the intended default values.

Since the btcjson package sets these fields to the default values when a command is unmarshaled from the wire, this typically isn't necessary. However, when the RPC server calls the handler internally with optional command fields set to nil, as is the case in getmininginfo, the defaults need to be set as well.


jrick:
ok

PR #335, opened 2015-03-01
Add id and timeoffset to getpeerinfo.

dajohi:


btcjson/jsonresults.go
```diff
@@ -137,6 +137,7 @@ type GetNetworkInfoResult struct {
 
 // GetPeerInfoResult models the data returned from the getpeerinfo command.
 type GetPeerInfoResult struct {
+	Id             string  `json:"addr"`
```
davecgh: This won't be used, so it probably doesn't matter, but `json:"id"`.

peer.go
```diff
@@ -402,6 +408,12 @@ func (p *peer) handleVersionMsg(msg *wire.MsgVersion) {
 	// Set the remote peer's user agent.
 	p.userAgent = msg.UserAgent
 
+	// Set the peer's time offset.
+	p.timeOffset = msg.Timestamp.Unix() - time.Now().Unix()
+
+	// Set the peer's ID.
+	p.id = atomic.AddInt32(&nodeCount, 1)
```
davecgh: If you're going to use atomic, the variable must be to aligned in the struct.  Thus, it needs to be moved up after `disconnect` before the `net.Conn` interface.  Also, it needs a comment after it clearly stating it's used by atomic so people know it can't be moved.

davecgh:
This is missing help for the new fields as indicated by the build failure.
```
rpcserverhelp_test.go:44: Failed to generate help for method 'getpeerinfo': getpeerinforesult-timeoffset
```

Also, I think `Id` needs to be `ID` to keep lint happy.

btcjson/jsonresults.go
```diff
@@ -144,14 +145,14 @@ type GetPeerInfoResult struct {
...
+	ConnTime       int64   `json:"conntime"`
+	TimeOffset     int64   `json:"timeoffset"`
 	PingTime       float64 `json:"pingtime"`
 	PingWait       float64 `json:"pingwait,omitempty"`
-	ConnTime       int64   `json:"conntime"`
 	Version        uint32  `json:"version"`
 	SubVer         string  `json:"subver"`
 	Inbound        bool    `json:"inbound"`
 	StartingHeight int32   `json:"startingheight"`
-	CurrentHeight  int32   `json:"currentheight,omitempty"`
```
davecgh: I don't think we should be losing this field.

btcjson/v2/btcjson/chainsvrresults.go
```diff
@@ -143,14 +144,14 @@ type GetPeerInfoResult struct {
...
+	ConnTime       int64   `json:"conntime"`
+	TimeOffset     int64   `json:"timeoffset"`
 	PingTime       float64 `json:"pingtime"`
 	PingWait       float64 `json:"pingwait,omitempty"`
-	ConnTime       int64   `json:"conntime"`
 	Version        uint32  `json:"version"`
 	SubVer         string  `json:"subver"`
 	Inbound        bool    `json:"inbound"`
 	StartingHeight int32   `json:"startingheight"`
-	CurrentHeight  int32   `json:"currentheight,omitempty"`
```
davecgh: Nor here.

davecgh:
Looks good after you bring the `CurrentHeight` field back and rebase to latest master.

davecgh:
OK

PR #336, opened 2015-03-01
Add new 'dropnode' RPC command

### LINKED ISSUE #79: need new rpc command .. addnode <ip> remove .. does not remove from dns list as well as persistent list
<pre>
104939 &lt;@davec&gt; ok so it's only looking in peers that we added
104950 &lt;@davec&gt; actually I think it's persistent peers in general
104952 &lt; todd&gt; the getpeerinfo rpc command can list all peers regardless of rpc list or dns list
105011 &lt; todd&gt; but the rpc 'addnode &lt;ip&gt; remove' cannot remove something from the dns list, only the rpc list
105036 &lt;@davec&gt; right    case delNodeMsg:
105036 &lt; todd&gt; so from what you said, 's/rpc list/persistent list/'
105036 &lt;@davec&gt;         found := false
105037 &lt;@davec&gt;         for e := state.persistentPeers.Front(); e != nil; e = e.Next() {
105044 &lt;@davec&gt; yep, that's wrong
105048 &lt;@davec&gt; ticket please
...
111339 &lt;@davec&gt; todd: oga just informed me that addnode is only *supposed* to work on persistent peers to be compatible with bitcoind
111410 &lt;@davec&gt; I rather agree that isn't very smart, but I guess we'll just provide an extension command that does it
111432 &lt;@davec&gt; not being able to remove any connection doesn't make sense

</pre>

Roasbeef:
This PR adds new JSON-RPC command: `dropnode`. 

This aims to fix #79 by adding a new command that can disconnect any (inbound or outbound) non-persistent peer.

### DIFF
--- a/server.go
+++ b/server.go
@@ -465,9 +465,9 @@ func (s *server) handleQuery(querymsg interface{}, state *peerState) {
 		}
 
 	case dropNodeMsg:
-		// Check inbound peers. We pass an empty callback since we don't
+		// Check inbound peers. We pass a nil callback since we don't
 		// require any additional actions on disconnect for inbound peers.
-		found := disconnectPeer(state.peers, msg.addr, func(p *peer) {})
+		found := disconnectPeer(state.peers, msg.addr, nil)
 		if found {
 			msg.reply <- nil
 			return
@@ -500,16 +500,19 @@ func (s *server) handleQuery(querymsg interface{}, state *peerState) {
 
 // disconnectPeer attempts to drop the connection of a tageted peer in the
 // passed peer list. This function returns true on success and false if the
-// peer is unable to be located. If the peer is found, the passed callback
-// function `whenFound' called with the peer as the argument before it is
-// removed from the peerList and is disconnected from the server
+// peer is unable to be located. If the peer is found, and the passed callback:
+// `whenFound' isn't nil, we call it with the peer as the argument before
+// it is removed from the peerList, and is disconnected from the server.
 func disconnectPeer(peerList *list.List, targetAddr string, whenFound func(*peer)) bool {
 	for e := peerList.Front(); e != nil; e = e.Next() {
 		peer := e.Value.(*peer)
 		if peer.addr == targetAddr {
+			if whenFound != nil {
+				whenFound(peer)
+			}
+
 			// This is ok because we are not continuing
 			// to iterate so won't corrupt the loop.
-			whenFound(peer)
 			peerList.Remove(e)
 			peer.Disconnect()
 			return true
--- a/server.go
+++ b/server.go
@@ -371,6 +371,11 @@ type delNodeMsg struct {
 	reply chan error
 }
 
+type dropNodeMsg struct {
+	addr  string
+	reply chan error
+}
+
 type getAddedNodesMsg struct {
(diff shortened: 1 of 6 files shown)


server.go
```diff
@@ -469,6 +464,28 @@ func (s *server) handleQuery(querymsg interface{}, state *peerState) {
 			msg.reply <- errors.New("peer not found")
 		}
 
+	case dropNodeMsg:
+		// Check inbound peers. We pass an empty callback since we don't
+		// require any additional actions on disconnect for inbound peers.
+		found := disconnectPeer(state.peers, msg.addr, func(p *peer) {})
```
davecgh: I'd suggest passing nil for the function here and adding a nil check in the `disconnectPeer` function.  This will avoid creating a closure which is much more expensive than a nil check.

server.go
```diff
@@ -481,6 +498,26 @@ func (s *server) handleQuery(querymsg interface{}, state *peerState) {
...
+// peer is unable to be located. If the peer is found, the passed callback
+// function `whenFound' called with the peer as the argument before it is
+// removed from the peerList and is disconnected from the server
+func disconnectPeer(peerList *list.List, targetAddr string, whenFound func(*peer)) bool {
+	for e := peerList.Front(); e != nil; e = e.Next() {
+		peer := e.Value.(*peer)
+		if peer.addr == targetAddr {
+			// This is ok because we are not continuing
+			// to iterate so won't corrupt the loop.
+			whenFound(peer)
```
davecgh: Per the comment above, you could do a nil check here:

```Go
if whenFound != nil {
	whenFound(peer)
}
```

server.go
Roasbeef: Great suggestion, I've changed this fragment to pass a nil function instead.

server.go
Roasbeef: Done.

dajohi:
This PR gets a firm NO from me.

Not in favor of yet another command specific command.  The 'addnode' command is already a bad enough name.

I am in favor of making a new 'node' command to control nodes such as:

btcctl node remove ID
btcctl node connect IP [perm|temp]
btcctl node disconnect ID

Then in the future, i'd like a 
btcctl node ID debuglevel ...

dajohi:
And if others disagree with me, then i ask not to make yet another top level command.  Add it as a subcommand to addnode.

Roasbeef:
What do you mean by: "command specific command"? Are you opposed to this PR strictly simply because it introduces a new command? Or do you feel that the feature it provides is not useful? 

What if it was altered a bit to introduce a new subcommand under `addnode`? In this case, what should the new sub-command be named? Since `remove` already exists as a subcommand, any similar synonym could potentially cause confusion due to redundancy/ambiguity (which is why I created a new command in the first place). 

I like your suggestion of introducing a custom, new command giving full control of peer connectivity (persistent vs non, etc.). `addnode` is pretty much a mess naming and semantic wise. With the proposed `node` command, what's the difference between `disconnect` and `remove` in that instance?

dajohi:
I like the idea.  I am solely opposed because of the name. :)  I would like to organize new commands into subcommands.

Roasbeef:
Closing in favor of #341

PR #341, opened 2015-03-05
Add new 'node' RPC command

### LINKED ISSUE #79: need new rpc command .. addnode <ip> remove .. does not remove from dns list as well as persistent list

Roasbeef:
This PR adds new JSON-RPC command: `node`. `node` allows a node operator full control over the peers their node is connect to.
A call to the new `node` RPC can take the following forms:
  * `btcctl node connect <addr> perm` (will attempt connect to a peer and mark it was permanent) 
  * `btcctl node connect <addr> temp` (will attempt to connect to a peer)
  * `btcctl node disconnect <addr or ID>` (will attempt to remove an outgoing or incoming peer)
  * `btcctl node remove <addr or ID>` (will attempt to remove a permanent peer)

This PR replaces #336. 
This aims to fix #79 by adding a new command that can disconnect any (inbound or outbound) non-persistent peer.

### DIFF
--- a/btcjson/v2/btcjson/btcdextcmds.go
+++ b/btcjson/v2/btcjson/btcdextcmds.go
@@ -7,6 +7,42 @@
 
 package btcjson
 
+// NodeSubCmd defines the type used in the addnode JSON-RPC command for the
+// sub command field.
+type NodeSubCmd string
+
+const (
+	// NConnect indicates the specified host that should be connected to.
+	NConnect NodeSubCmd = "connect"
+
+	// NRemove indicates the specified peer that should be removed as a
+	// persistent peer.
+	NRemove NodeSubCmd = "remove"
+
+	// NDisconnect indicates the specified peer should be disonnected.
+	NDisconnect NodeSubCmd = "disconnect"
+)
+
+// NodeCmd defines the dropnode JSON-RPC command.
+type NodeCmd struct {
+	SubCmd        NodeSubCmd `jsonrpcusage:"\"connect|remove|disconnect\""`
+	Target        string
+	ConnectSubCmd *string `jsonrpcusage:"\"perm|temp\""`
+}
+
+// NewNodeCmd returns a new instance which can be used to issue a `node`
+// JSON-RPC command.
+//
+// The parameters which are pointers indicate they are optional.  Passing nil
+// for optional parameters will use the default value.
+func NewNodeCmd(subCmd NodeSubCmd, target string, connectSubCmd *string) *NodeCmd {
+	return &NodeCmd{
+		SubCmd:        subCmd,
+		Target:        target,
+		ConnectSubCmd: connectSubCmd,
+	}
+}
+
 // DebugLevelCmd defines the debuglevel JSON-RPC command.  This command is not a
 // standard Bitcoin command.  It is an extension for btcd.
 type DebugLevelCmd struct {
@@ -45,6 +81,7 @@ func init() {
 	flags := UsageFlag(0)
 
 	MustRegisterCmd("debuglevel", (*DebugLevelCmd)(nil), flags)
+	MustRegisterCmd("node", (*NodeCmd)(nil), flags)
 	MustRegisterCmd("getbestblock", (*GetBestBlockCmd)(nil), flags)
 	MustRegisterCmd("getcurrentnet", (*GetCurrentNetCmd)(nil), flags)
 }
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -155,6 +155,7 @@ var rpcHandlersBeforeInit = map[string]commandHandler{
 	"gettxout":              handleGetTxOut,
 	"getwork":               handleGetWork,
 	"help":                  handleHelp,
+	"node":                  handleNode,
 	"ping":                  handlePing,
 	"searchrawtransactions": handleSearchRawTransactions,
 	"sendrawtransaction":    handleSendRawTransaction,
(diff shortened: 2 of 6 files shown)


davecgh:
@dajohi Does this one address your concerns with the original PR?

dajohi:
Now that getpeerinfo contains ID, disconnect should take an ID.

davecgh:
@dajohi: I'd suggest it should take either address or ID.

The address already works with "addr:port" to be specific enough to identify a given peer.  When "addr" with no port is provided, it appends the default port for the network and attempts to remove that node.

So, I think ideally the current logic would stay, but perhaps `disconnect`/`remove` should additionally allow the "addr" to be an ID.  It would be a simple check in the code I think.   First, check if it's a valid IP address (with optional port), and if so, use the current semantics.  If not, attempt to parse it as an unsigned integer and, if successful, treat it as an ID.  Otherwise, invalid address or id error.

btcjson/v2/btcjson/btcdextcmds.go
```diff
@@ -7,6 +7,42 @@
...
+
+	// NDisconnect indicates the specified peer should be disonnected.
+	NDisconnect NodeSubCmd = "disconnect"
+)
+
+// NodeCmd defines the dropnode JSON-RPC command.
+type NodeCmd struct {
+	SubCmd        NodeSubCmd `jsonrpcusage:"\"connect|remove|disconnect\""`
+	Addr          string
+	ConnectSubCmd *string `jsonrpcusage:"\"perm|non-perm\""`
```
davecgh: What's the default?

rpcserver.go
```diff
@@ -344,6 +345,36 @@ func handleAddNode(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (in
...
+
+	addr := normalizeAddress(c.Addr, activeNetParams.DefaultPort)
+	var err error
+	switch c.SubCmd {
+	case "disconnect":
+		err = s.server.DisconnectNode(addr)
+	case "remove":
+		err = s.server.RemoveNode(addr)
+	case "connect":
+		switch *c.ConnectSubCmd {
```
davecgh: This will panic if the optional parameter isn't specified currently.

I'd suggest something like:

```Go
subCmd := "non-perm"
if c.ConnectSubCmd != nil {
	subCmd = *c.ConnectSubCmd
}
switch subCmd {
...
}
```

rpcserver.go
```diff
@@ -344,6 +345,36 @@ func handleAddNode(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (in
...
+		case "perm", "non-perm":
+			err = s.server.ConnectNode(addr, *c.ConnectSubCmd == "perm")
+		default:
+			err = errors.New("invalid subcommand for node connect")
```
davecgh: I know that addnode does this too, but really it should be changed too.  The problem here is this will result in an internal error which is not accurate for an invalid parameter.  It really should be doing something like:

```Go
err = &btcjson.RPCError{
        Code: ErrRPCInvalidParameter,
        Message: "invalid subcommand for node connect",
    }
```

rpcserver.go
```diff
@@ -344,6 +345,36 @@ func handleAddNode(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (in
...
+		}
+	default:
+		err = errors.New("invalid subcommand for node")
```
davecgh: The same is true here with the error.  This should be:

```Go
err = &btcjson.RPCError{
        Code: ErrRPCInvalidParameter,
        Message: "invalid subcommand for node",
    }
```

rpcserver.go
```diff
@@ -344,6 +345,36 @@ func handleAddNode(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (in
...
+	}
+
+	if err != nil {
+		return nil, internalRPCError(err.Error(), "")
```
davecgh: This is not really accurate either since many cases, such as an address that doesn't exist, are not internal errors.  They are simply `ErrRPCInvalidParameter`.  However, the returned errors from each of the functions above needs to be examined to determine what the type of errors that can be returned are to react accordingly.

dajohi:
Can we change no-perm to temp?

dajohi:
@davecgh If i have 2 connections open to ip:port, should disconnect ip:port disconnect the first, or all?

davecgh:
I'd say all of them ideally.

btcjson/v2/btcjson/btcdextcmds.go
Roasbeef: Yeah the intent is to require the caller to specify perm/non-perm. The handler code in `rpcserver.go` currently detects if the optional parameter is missing, and returns the appropriate error if so. 

EDIT: non-perm is now `temp`

rpcserver.go
Roasbeef: Fixed, adopted the first method.

rpcserver.go
Roasbeef: Noted, I've changed this fragment to return the proper error, and additionally altered `addnode` to follow the same behavior for returned errors.

rpcserver.go
Roasbeef: Done.

rpcserver.go
Roasbeef: Fixed, changed `addnode` as well.

Roasbeef:
Recently pushed a commit changing the following: 
  * `node disconnect/remove` can now operate on both `ip:port` or a valid `node id`. 
  * for `node connect`, `non-perm` has been changed to `temp`,  `temp` is now the default. 

@dajohi by "2 connections",  do you mean one inbound connection and outbound connection to the same `ip:port`?

dajohi:
@Roasbeef No, I don't think that is possible.   I am talking about > 1 outbound connections to same ip:port.

Roasbeef:
`node disconnect <target>` now disconnects all matching targets rather than just the first one. I've also updated the command help and RPC documentation to reflect this new behavior.

davecgh:
Thanks for the updates.  I'll get the latest changes reviewed.

davecgh:
A couple of things:

* Calling `node remove existing_id_that_is_non_perm` results in `-8: peer not found`.  While I understand that's because the node is not a permanent one and can't be removed, I think the error should be more descriptive there.  Something like: "non-permanent peers can't be removed - use disconnect".

Not necessary for this PR, but food for thought:
* I think it would made sense to have something like `node info (connected|added)` which returns the same information as `getpeerinfo` for connected peers, and `getaddednodeinfo` for added/perm peers.  We'll want to keep `getpeerinfo` and `getaddednodeinfo` for compatibility, but it feels a little odd to do `./btcctl node connect/disconnect/remove` but have to use `/btcctl getpeerinfo` and `/btcctl getaddednodeinfo` to see the results.  Would like @dajohi feedback on this point.

davecgh:
More testing and I ran into this:

> $ ./btcctl node remove 10.1.2.3
> -8: invalid address or node ID
> $ ./btcctl node remove 10.1.2.3:8333

That is to say it didn't remove the node by IP without a port.

Also I noticed that there is no way to see the id for a perm peer that is not connected.

davecgh:
Probably also related to the previous issue:

> $ ./btcctl node connect 10.1.2.3 perm
> $ ./btcctl getaddednodeinfo true
> [
>  {
>    "addednode": "10.1.2.3:8333",
>    "connected": false,
>    
>  }
> ]
> $ ./btcctl node connect 10.1.2.3 perm
> -8: peer already connected
> $ ./btcctl node connect 10.1.2.3
> $

I suspect that should error out because the node already exists as a perm peer even though it's not connected yet.

Roasbeef:
Pushed some commits fixing the following:
  * error messages for `remove/disconnect` has been improved. The user is now pointed to the proper
    sub-command if we detect they might have made a mistake discerning a peer as `perm` or `temp`. 
  * A bug in parsing nodeId vs IP has been fixed. `ip` and `ip:port` should now work correctly for 
    `remove/disconnect`.
  * Trying to connect a peer that's already added as a permanent peer now returns a new error message.

Roasbeef:
Regarding not being able to show the ID of a permanent peer that isn't currently connected. 

It seems that this is the current implemented behavior of `getpeerinfo`, if the peer isn't connected, it's skipped: 
```go
	case getPeerInfoMsg:
		syncPeer := s.blockManager.SyncPeer()
		infos := make([]*btcjson.GetPeerInfoResult, 0, state.peers.Len())
		state.forAllPeers(func(p *peer) {
			if !p.Connected() {
				return
			}
```

davecgh:
Correct, `getpeerinfo` only shows connected peers.  What I was getting at is I think the `getaddednodeinfo` response needs to include the peer ID.  That would of course then transfer over into the `node info` responses discussed above.

davecgh:
I believe this is ready for merge.  There are a few points brought up that can be improved upon, but what is there works properly and reads well.

Please squash and rebase it.

Roasbeef:
Rebased and squashed :+1:

server.go
```diff
@@ -475,16 +491,20 @@ func (s *server) handleQuery(querymsg interface{}, state *peerState) {
 		msg.reply <- infos
 
 	case addNodeMsg:
+	case connectNodeMsg:
```
tuxcanfly: I think this breaks the existing `addNodeMsg`, `delNodeMsg` cases because they are no longer being handled. It would probably make sense to keep these cases as they were for compatibility. If not, we could deprecate `addnode` RPC command with a note about using `node` instead.

Roasbeef: Yeah you're right. Trying to use the `addnode` sub-commands, `del` or `add` causes the `btcctl` caller to hang indefinitely waiting for a reply since the case stacking essentially made the commands into a no-op. 

I agree that we should mark `addnode` as deprecated and point users to the new `node` command instead. But for now, 've created as a temporary PR fix: btcsuite/btcd#398

Issue #122, opened 2014-04-26
Allow to stop watching for transaction notifications


shesek:
After issuing `notifyreceived` / `notifyspent`, there's no way to stop watching for those notifications. I'm working on something that only cares about the first transaction being sent to an address, at which point it should stop watching it.

Would it make sense for you to add "un-notify" commands?

PR #342, opened 2015-03-05
Allow clients to cancel transaction notifications

Roasbeef:
This PR aims to address #122 by providing JSON-RPC methods allowing clients to cancel received and spent notifications. 

Note: before this can be properly integrated into `btcwallet`, `btcrpcclient` will need to be updated to use `btcjsonv2`.

### DIFF
--- a/btcjson/chainsvrwscmds.go
+++ b/btcjson/chainsvrwscmds.go
@@ -31,6 +31,15 @@ func NewNotifyBlocksCmd() *NotifyBlocksCmd {
 	return &NotifyBlocksCmd{}
 }
 
+// StopNotifyBlocksCmd defines the stopnotifyblocks JSON-RPC command.
+type StopNotifyBlocksCmd struct{}
+
+// NewStopNotifyBlocksCmd returns a new instance which can be used to issue a
+// stopnotifyblocks JSON-RPC command.
+func NewStopNotifyBlocksCmd() *StopNotifyBlocksCmd {
+	return &StopNotifyBlocksCmd{}
+}
+
 // NotifyNewTransactionsCmd defines the notifynewtransactions JSON-RPC command.
 type NotifyNewTransactionsCmd struct {
 	Verbose *bool `jsonrpcdefault:"false"`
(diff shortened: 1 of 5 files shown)


davecgh:
Thanks.  I'll review this within the next few days.

btcjson/v2/btcjson/chainsvrwscmds.go
```diff
@@ -93,6 +93,32 @@ func NewNotifySpentCmd(outPoints []OutPoint) *NotifySpentCmd {
...
+// NewStopNotifySpentCmd returns a new instance which can be used to issue a
+// stopnotifyspent JSON-RPC command.
+func NewStopNotifySpentCmd(outPoints []OutPoint) *StopNotifySpentCmd {
+	return &StopNotifySpentCmd{
+		OutPoints: outPoints,
+	}
+}
+
+// StopNotifyReceivedCmd defines the stopnotifyreceived JSON-RPC command.
+type StopNotifyReceivedCmd struct {
```
davecgh: Minor nitpick, but all of the other command definitions alphabetized, so I think this should be defined before `type StopNotifySpentCmd struct` to be consistent.

davecgh:
I need to test this out still, but aside from my minor nitpick, this reads well.

btcjson/v2/btcjson/chainsvrwscmds.go
Roasbeef: Fixed.

davecgh:
I noticed this includes `stopnotifyspent` and `stopnotifyreceived`, but it lacks stop for the other 2 notification registrations.  In particular, `stopnotifyblocks` and `stopnotifynewtransactions`.

davecgh:
I've tested this and everything looks good.  The code reads well too.  Just need to squash and rebase it and it's good to go.

OK

davecgh:
Oh, I just realized I was a bit hasty.  While the code is correct and works properly, I noticed the documentation for the JSON-RPC API wasn't updatd.  The new commands need to be added to [docs/json_rpc_api.md](https://github.com/btcsuite/btcd/blob/master/docs/json_rpc_api.md#Notifications).

davecgh:
This is just waiting on the documentation updates.

Roasbeef:
The documentation has been updated with the latest commit.

davecgh:
Thanks for the udpates.  OK after rebase and squash.

Roasbeef:
Rebased and squashed :+1:

davecgh:
Thanks, but it's not against the latest master!

PR #343, opened 2015-03-10
wire: Add func to get pkscript locs from a tx.

davecgh:
This pull request provides a new function named PkScriptLocs on the MsgTx type which can be used to efficiently retrieve a list of offsets for the public key scripts for the serialized form of the transaction.

This is useful for certain applications which store fully serialized transactions and want to be able to quickly index into the serialized transaction to extract a give public key script directly thereby avoiding the need to deserialize the entire transaction.


jrick:
ok

PR #344, opened 2015-03-10
blockchain: Provide new IsCoinBaseTx function.

davecgh:
This pull request adds a new function to the blockchain package named IsCoinBaseTx which performs the same function as IsCoinBase except it takes raw wire transactions as opposed to the higher level util transactions.

While here, it also adds a file for benchmarks along with a couple of benchmarks for the IsCoinBase and IsCoinBaseTx functions.

Finally, the function was very slightly optimized:

```
BenchmarkIsCoinBaseOld  100000000  10.7 ns/op  0 B/op  0 allocs/op
BenchmarkIsCoinBaseNew  200000000  6.05 ns/op  0 B/op  0 allocs/op
```

### DIFF
--- a/blockchain/bench_test.go
+++ b/blockchain/bench_test.go
@@ -1,0 +1,32 @@
+// Copyright (c) 2015 Conformal Systems LLC.
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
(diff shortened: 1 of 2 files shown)


jrick:
ok

PR #345, opened 2015-03-10
Process orphans on block acceptance.

dajohi:
For every transaction in a newly accepted block, process the orphan
pool moving now no longer orphan transactions to the mempool.

Previously, no longer orphan transactions would remain in the orphan
pool.

### DIFF
--- a/mempool.go
+++ b/mempool.go
@@ -391,6 +391,7 @@ func calcMinRequiredTxRelayFee(serializedSize int64) int64 {
 
 // removeOrphan is the internal function which implements the public
 // RemoveOrphan.  See the comment for RemoveOrphan for more details.
+//
 // This function MUST be called with the mempool lock held (for writes).
 func (mp *txMemPool) removeOrphan(txHash *wire.ShaHash) {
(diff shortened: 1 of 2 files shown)


mempool.go
```diff
@@ -1261,12 +1263,23 @@ func (mp *txMemPool) MaybeAcceptTransaction(tx *btcutil.Tx, isNew, rateLimit boo
...
 
-// processOrphans determines if there are any orphans which depend on the passed
+// ProcessOrphans determines if there are any orphans which depend on the passed
 // transaction hash (it is possible that they are no longer orphans) and
 // potentially accepts them to the memory pool.  It repeats the process for the
 // newly accepted transactions (to detect further orphans which may no longer be
 // orphans) until there are no more.
 //
+// This function is safe for concurrent access.
+func (mp *txMemPool) ProcessOrphans(hash *wire.ShaHash) error {
```
davecgh: Define this after the thing is calls, please.

davecgh:
The concept looks sound to me.

davecgh:
OK

PR #347, opened 2015-03-11
txscript:  Add new flag ScriptVerifyCleanStack

dajohi:
The ScriptVerifyCleanStack flag requires that only a single
stack element remains after evaluation and that when interpreted
as a bool, it must be true.  This is BIP0062, rule 6.

This mimics Bitcoin Core commit b6e03cc59208305681745ad06f2056ffe6690597


txscript/script.go
```diff
@@ -755,23 +769,29 @@ func (s *Script) Execute() (err error) {
...
 
-	return s.CheckErrorCondition()
+	return s.CheckErrorCondition(true)
 }
 
 // CheckErrorCondition returns nil if the running script has ended and was
 // successful, leaving a a true boolean on the stack. An error otherwise,
 // including if the script has not finished.
-func (s *Script) CheckErrorCondition() (err error) {
+func (s *Script) CheckErrorCondition(finalScript bool) (err error) {
```
davecgh: The named return can be removed.

davecgh:
This should add the new flag to the standard verify flags.

txscript/script.go
```diff
@@ -624,6 +630,12 @@ const (
 	// executed.
 	ScriptDiscourageUpgradableNops
 
+	// ScriptVerifyCleanStack defines that the stack must contain only
+	// one stack element after evaluation and that the element must be
+	// true if interpreted as a boolean.
```
davecgh: Please call out that this is Rule 6 of BIP0062.

txscript/script.go
```diff
@@ -624,6 +630,12 @@ const (
...
+	// This flag should never be used without the ScriptBip16 flag.
```
davecgh: Since this flag should never be used without the ScriptBip16 flag, the `NewScript` function should return an error when it's called with this flag set without also having the ScriptBip16 flag set.

txscript/script.go
```diff
@@ -715,6 +727,13 @@ func NewScript(scriptSig []byte, scriptPubKey []byte, txidx int, tx *wire.MsgTx,
 		m.dstack.verifyMinimalData = true
 		m.astack.verifyMinimalData = true
 	}
+	if flags&ScriptVerifyCleanStack == ScriptVerifyCleanStack {
+		if flags&ScriptBip16 != ScriptBip16 {
+			return nil, fmt.Errorf("the ScriptVerifyCleanStack " +
```
davecgh: Please make this an exported error defined near the top where the others are.

davecgh:
Also, this needs a new test for the invalid flag combination error

txscript/script_test.go
```diff
@@ -4816,3 +4816,52 @@ func TestIsPushOnlyScript(t *testing.T) {
...
+			},
+		},
+		LockTime: 0,
+	}
+	pkScript := []byte{txscript.OP_NOP}
+
+	for i, test := range tests {
+		_, err := txscript.NewScript(tx.TxIn[0].SignatureScript,
+			pkScript, 0, tx, test)
+		if err == nil {
```
davecgh: Please check for the specific error:

```Go
if err != txscript.ErrInvalidFlags {
	t.Fatalf("TestInvalidFlagCombinations #%d unexpected "+
		"error %v", i, err)
}
```

davecgh:
OK

PR #348, opened 2015-03-13
btcctl: Support - argument to read from stdin.

davecgh:
This pull request modifies the argument handling for btcctl to treat a parameter that is a single dash as an indicator to read that paramter from stdin instead.

This change allows commands, such as the submitblock, to accept data piped from stdin for any parameter.  This, in turn, allows large arguments, such as blocks, which can often be too big for a single argument due to Operating System limitations to be submitted by putting them into a file and redirecting stdin.

For example:
```bash
$ btcctl submitblock - <block.hex
$ cat block.hex | btcctl submitblock -

$ btcctl sendrawtransaction - <tx.hex
$ cat tx.hex | btcctl sendrawtransaction -
```

### DIFF
--- a/cmd/btcctl/btcctl.go
+++ b/cmd/btcctl/btcctl.go
@@ -1,9 +1,11 @@
 package main
 
 import (
+	"bufio"
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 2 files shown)


jrick:
Instead of special casing this, why not use `-` to denote reading from stdin?  Then this could be used for other requests as well, and the argument read in from stdin wouldn't always need to be the very first argument for the RPC.

It will also be easier to document, instead of special casing the help text for `submitblock`.

davecgh:
@jrick: I like that idea.  I'll make the changes.

davecgh:
I've updated this to accept '-' for arguments instead.  The PR title and description have been updated as well.

jrick:
ok

PR #349, opened 2015-03-13
txscript: Add new flag ScriptVerifyLowS

dajohi:
The ScriptVerifyLowS flag defines that script signatures must
comply with the DER format as well as have an S value less than
or equal to the half order.


txscript/script.go
```diff
@@ -367,6 +377,44 @@ func (s *Script) checkHashTypeEncoding(hashType SigHashType) error {
 	return nil
 }
 
+func (s *Script) checkLowDERSignature(sig []byte) error {
```
davecgh: Missing a comment here.

davecgh: After reviewing more, I personally don't think this function should exist at all.  Rather, the `checkSignatureEncoding` function should check the flag at the appropriate place and do the check there:

```Go
if s.verifyLowS { /* do the checks and return error if no good */ }
```

davecgh:
OK

txscript/script.go
```diff
@@ -164,6 +170,9 @@ var ErrUnsupportedAddress = errors.New("unsupported address type")
 // This timestamp corresponds to Sun Apr 1 00:00:00 UTC 2012.
 var Bip16Activation = time.Unix(1333238400, 0)
 
+// curve halforder, used to tame ECDSA malleability (see BIP0062)
+var halforder = new(big.Int).Rsh(new(big.Int).Set(btcec.S256().N), 1)
```
davecgh: The `btcec.S256().N` parameter is already a `big.Int`.  No need to create the extra one here:

var halforder = new(big.Int).Rsh(btcec.S256().N, 1)

davecgh:
OK

PR #350, opened 2015-03-14
Ignore getaddr requests from outbound peers.

dajohi:
This change reduces fingerprinting attacks and
mimics Bitcoin Core commit
dca799e1db6e319fdd47e0bfdb038eab0efabb85


davecgh:
OK

PR #351, opened 2015-03-15
Do a little addrindex database cleanup.

dajohi:
- Use explicit index values for byte slices
- Fix a bug in FetchTxsForAddr that allocated an extra 10 bytes
  for each address index
- Add missing iterator release in error path
- Check for iterator errors.


davecgh:
OK

PR #352, opened 2015-03-17
Reduce fingerprinting.

dajohi:
This change reduces fingerprinting via timestamps in addr messages.
Previously, the last seen time for an address was updated when
certain protocol commands were received.  Now, the last seen time
is set when the peer disconnects if the peer had sent a verack
message and was connected for more than 20 minutes.

This mimics Bitcoin Core commit:
9c2737901b5203f267d21d728019d64b46f1d9f3

Also, add additional sanity checking before updating the peer's
timestamp.  These include:

 - Do not mark a peer as connected if we never received
   a version message.
 - Disconnect a peer for sending a verack before btcd
   sent a version
 - Disconnect a peer for sending multiple verack's

### DIFF
--- a/addrmgr/addrmanager.go
+++ b/addrmgr/addrmanager.go
@@ -862,12 +862,12 @@ func (a *AddrManager) Good(addr *wire.NetAddress) {
 	if ka == nil {
 		return
 	}
+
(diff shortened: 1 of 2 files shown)


davecgh:
OK

Issue #346, opened 2015-03-11
"searchrawtransactions" returns duplicates of transactions


aakselrod:
While btcd was running, I broadcast two transactions: ba1eecb28b363f40ee8ad4796989c8eb3d552e8ab85c146fac00d161147dbf1d and 271c05315790682c964d9ab29d3e84cc79a8f0d80bf6acf638186ce36ea780ae (not through the same btcd I'm querying).

After the transactions got confirmed, I am seeing something like the blockquote below. It looks like transactions may not be getting removed from `txMemPool.addrindex` when they're confirmed in a block, because the duplicate 0-confirm transactions go away after restarting btcd.

If I get a chance soon, I will try and see why this is happening/try to fix it.

> $ ./btcctl -u x -P xx --testnet searchrawtransactions "miSTbp6DUSA3v46j3bTT6gKV1JQS21rEqN"
> [
>   {
>     "hex": "01000000011dbf7d1461d100ac6f145cb88a2e553debc8896979d48aee403f368bb2ec1eba010000006a473044022015052ca8ad3593300048cd16efd2ecc99ea6d2a860562cfd341f7859a79696900220687091479b412d864df8c5003581db0a49f755f51b1f7dfb0940d8cd9409dfa9012102d4e5142c293fff6ca2d6eb134ab1a01e782889d2a913b063da6829756e096e91ffffffff0240787d01000000001976a914200ec2d84e94db949d8bcd472463d9280c04579b88ac00ed6831000000001976a914875d60b7ce1fd2eb5778d9c5c89031873832728c88ac00000000",
>     "txid": "271c05315790682c964d9ab29d3e84cc79a8f0d80bf6acf638186ce36ea780ae",
>     "version": 1,
>     "locktime": 0,
>     "vin": [
>       {
>         "txid": "ba1eecb28b363f40ee8ad4796989c8eb3d552e8ab85c146fac00d161147dbf1d",
>         "vout": 1,
>         "scriptSig": {
>           "asm": "3044022015052ca8ad3593300048cd16efd2ecc99ea6d2a860562cfd341f7859a79696900220687091479b412d864df8c5003581db0a49f755f51b1f7dfb0940d8cd9409dfa901 02d4e5142c293fff6ca2d6eb134ab1a01e782889d2a913b063da6829756e096e91",
[...]

Issue #303, opened 2015-02-17
--addrindex doesn't work on preprocessed blocks


manan19:
Generating address index with --addrindex option succeeds when processing blocks from scratch.

But generating the address index on a fully synced node throws the following errors -
```
11:40:46 2015-02-17 [ERR] ADXR: Couldn't get referenced txOut (cf2ab4e635099dfc8f2c0438c2d7eb6fdbc57ce3c7d497869e1466f03f872d61:0): <nil>
```

To reproduce bug, drop the address index from a fully synced node and then try to rengenerate.

```
$ btcd --testnet --dropaddrindex --rpclisten=127.0.0.1:18334 --listen=127.0.0.1:18555
11:38:37 2015-02-17 [INF] BTCD: Version 0.9.0-beta
11:38:37 2015-02-17 [INF] BTCD: Loading block database from '/Users/manan/Library/Application Support/Btcd/data/testnet/blocks_leveldb'
11:38:37 2015-02-17 [INF] BTCD: Block database loaded with block height 323335
11:38:37 2015-02-17 [INF] BTCD: Deleting entire addrindex.
11:39:12 2015-02-17 [INF] BTCD: Successfully deleted addrindex, exiting

$ btcd --testnet --addrindex --rpclisten=127.0.0.1:18334 --listen=127.0.0.1:18555
11:40:32 2015-02-17 [INF] BTCD: Version 0.9.0-beta
11:40:32 2015-02-17 [INF] BTCD: Loading block database from '/Users/manan/Library/Application Support/Btcd/data/testnet/blocks_leveldb'
11:40:32 2015-02-17 [INF] BTCD: Block database loaded with block height 323335
11:40:32 2015-02-17 [WRN] AMGR: Skipping bound address: address 127.0.0.1 is not routable
11:40:32 2015-02-17 [INF] BMGR: Generating initial block node index.  This may take a while...
11:40:33 2015-02-17 [INF] BMGR: Block index generation complete
11:40:33 2015-02-17 [INF] SRVR: Server listening on 127.0.0.1:18555
11:40:33 2015-02-17 [INF] ADXR: Building up address index from height 0 to 323335.
11:40:33 2015-02-17 [INF] RPCS: RPC server listening on 127.0.0.1:18334
11:40:33 2015-02-17 [INF] AMGR: Loaded 1742 addresses from file '/Users/manan/Library/Application Support/Btcd/data/testnet/peers.json'
[...]

PR #354, opened 2015-03-19
Addrindex fixes

aakselrod:
This fixes issue #346 and makes it so that transactions are returned in order. In order to do the latter, the integers in the address index are stored as big endian rather than little endian; this requires an index rebuild and the pull request does that automatically. Unfortunately, there is a bug exposed by the index rebuild which causes a shutdown (see issue #303), which I'll be working on fixing next.

### DIFF
--- a/database/ldb/block.go
+++ b/database/ldb/block.go
@@ -286,6 +286,28 @@ func (db *LevelDb) NewestSha() (rsha *wire.ShaHash, rblkid int64, err error) {
 	return &sha, db.lastBlkIdx, nil
 }
 
+// checkAddrIndexVersion returns an error if the address index version stored
+// in the database is less than the current version, or if it doesn't exist.
+// This function is used on startup to signal OpenDB to drop the address index
+// if it's in an old, incompatible format.
+func (db *LevelDb) checkAddrIndexVersion() error {
+	db.dbLock.Lock()
+	defer db.dbLock.Unlock()
+
+	data, err := db.lDb.Get(addrIndexVersionKey, db.ro)
+	if err != nil {
+		return database.ErrAddrIndexDoesNotExist
+	}
+
+	indexVersion := binary.LittleEndian.Uint16(data)
+
+	if indexVersion != uint16(addrIndexCurrentVersion) {
+		return database.ErrAddrIndexDoesNotExist
+	}
+
+	return nil
+}
+
 // fetchAddrIndexTip returns the last block height and block sha to be indexed.
 // Meta-data about the address tip is currently cached in memory, and will be
 // updated accordingly by functions that modify the state. This function is
--- a/database/ldb/leveldb.go
+++ b/database/ldb/leveldb.go
@@ -80,6 +80,9 @@ func parseArgs(funcName string, args ...interface{}) (string, error) {
 	return dbPath, nil
 }
 
+// CurrentDBVersion is the database version.
+var CurrentDBVersion int32 = 1
+
 // OpenDB opens an existing database for use.
 func OpenDB(args ...interface{}) (database.Db, error) {
 	dbpath, err := parseArgs("OpenDB", args...)
@@ -141,10 +144,20 @@ blocknarrow:
 		}
 	}
 
+	log.Infof("Checking address index")
+
 	// Load the last block whose transactions have been indexed by address.
 	if sha, idx, err := ldb.fetchAddrIndexTip(); err == nil {
-		ldb.lastAddrIndexBlkSha = *sha
-		ldb.lastAddrIndexBlkIdx = idx
+		if err = ldb.checkAddrIndexVersion(); err == nil {
+			ldb.lastAddrIndexBlkSha = *sha
+			ldb.lastAddrIndexBlkIdx = idx
+			log.Infof("Address index good, continuing")
+		} else {
+			log.Infof("Address index in old, incompatible format, dropping...")
+			ldb.deleteOldAddrIndex()
+			ldb.DeleteAddrIndex()
+			log.Infof("Old, incompatible address index dropped and can now be rebuilt")
+		}
 	} else {
 		ldb.lastAddrIndexBlkIdx = -1
 	}
@@ -155,9 +168,6 @@ blocknarrow:
 
 	return db, nil
 }
-
-// CurrentDBVersion is the database version.
-var CurrentDBVersion int32 = 1
 
 func openDB(dbpath string, create bool) (pbdb database.Db, err error) {
 	var db LevelDb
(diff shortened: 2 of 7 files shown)


database/ldb/block.go
```diff
@@ -286,6 +286,28 @@ func (db *LevelDb) NewestSha() (rsha *wire.ShaHash, rblkid int64, err error) {
 	return &sha, db.lastBlkIdx, nil
 }
 
+// checkAddrIndexVersion returns an error if the address index version stored
+// in the database is less than the current version, or if it doesn't exist.
+// This function is used on startup to signal OpenDB to drop the address index
+// if it's in an old, incompatible format.
+func (db *LevelDb) checkAddrIndexVersion() (err error) {
```
davecgh: Please avoid using named errors:

```Go
+func (db *LevelDb) checkAddrIndexVersion() error {
```

aakselrod: Will change this in next commit.

database/ldb/tx.go
```diff
@@ -35,6 +35,14 @@ var addrIndexMetaDataKey = []byte("addrindex")
 // iterators.
 var addrIndexKeyPrefix = []byte("a-")
 
+// Address index version is required to drop/rebuild address index if version
+// is older than current as the format of the index may have changed. This is
+// true when going from no version to version 1 as the address index is stored
+// as big endian in version 1 and little endian in the original code. Version
+// is stored as two bytes, little endian (to match all the code but the index).
+var addrIndexVersionKey = []byte("addrindexversion")
+var addrIndexCurrentVersion = uint16(1)
```
davecgh: This should probably be a `const`.

aakselrod: ACK, will update this as well.

rpcserver.go
```diff
@@ -2620,17 +2611,31 @@ func handleSearchRawTransactions(s *rpcServer, cmd interface{}, closeChan <-chan
...
-		// Slice off any possible overflow.
-		addressTxs = addressTxs[:numRequested]
-	} else {
-		// Otherwise, we'll also take a look into the database.
-		dbTxs, err := s.server.db.FetchTxsForAddr(addr, numToSkip,
-			numRequested-len(addressTxs))
-		if err == nil && len(dbTxs) != 0 {
-			for _, txReply := range dbTxs {
+
+	// While it's more efficient to check the mempool for relevent transactions
```
davecgh: relevent -> relevant

aakselrod: Whoops, I cut and pasted the old comment's misspelling and didn't notice. Will fix.

rpcserver.go
```diff
@@ -2620,17 +2611,31 @@ func handleSearchRawTransactions(s *rpcServer, cmd interface{}, closeChan <-chan
...
+	// first, we want to return results in order of occurrence/dependency so
+	// we'll check the mempool only if there aren't enough results returned
+	// by the database.
+	dbTxs, err := s.server.db.FetchTxsForAddr(addr, numToSkip,
+		numRequested-len(addressTxs))
+	if err == nil && len(dbTxs) != 0 {
```
davecgh: There is no reason to check for `len(dbTxs) != 0` here and it is slightly less efficient to do so since for loop already does so meaning it is duplicated work.

rpcserver.go
```diff
@@ -2620,17 +2611,31 @@ func handleSearchRawTransactions(s *rpcServer, cmd interface{}, closeChan <-chan
...
+			addressTxs = append(addressTxs, txReply)
+		}
+	}
+
+	// This code (and txMemPool.FilterTransactionsByAddress()) doesn't sort by
+	// dependency. This might be something we want to do in the future when we
+	// return results for the client's convenience, or leave it to the client.
+	if len(addressTxs) < numRequested {
+		memPoolTxs, err := s.server.txMemPool.FilterTransactionsByAddress(addr)
+		if err == nil && len(memPoolTxs) != 0 {
```
davecgh: Same here.  No reason to check len again since the for loop already does it.

rpcserver.go
aakselrod: ACK, will fix in next commit.

rpcserver.go
aakselrod: ACK here as well, will fix in next commit.

database/ldb/block.go
```diff
@@ -286,6 +286,28 @@ func (db *LevelDb) NewestSha() (rsha *wire.ShaHash, rblkid int64, err error) {
...
+func (db *LevelDb) checkAddrIndexVersion() error {
+	db.dbLock.Lock()
+	defer db.dbLock.Unlock()
+
+	data, err := db.lDb.Get(addrIndexVersionKey, db.ro)
+	if err != nil {
+		return database.ErrAddrIndexDoesNotExist
+	}
+
+	indexVersion := binary.LittleEndian.Uint16(data[0:])
```
Roasbeef: Minor nitpick: no slicing is needed here. You can directly pass the byte slice: 
```go
indexVersion := binary.LittleEndian.Uint16(data)
```

database/ldb/leveldb.go
```diff
@@ -143,8 +143,13 @@ blocknarrow:
...
 	// Load the last block whose transactions have been indexed by address.
 	if sha, idx, err := ldb.fetchAddrIndexTip(); err == nil {
-		ldb.lastAddrIndexBlkSha = *sha
-		ldb.lastAddrIndexBlkIdx = idx
+		if err = ldb.checkAddrIndexVersion(); err == nil {
+			ldb.lastAddrIndexBlkSha = *sha
+			ldb.lastAddrIndexBlkIdx = idx
+		} else {
+			ldb.deleteOldAddrIndex()
+			ldb.DeleteAddrIndex()
```
Roasbeef: I think it'd be nice to throw a logging statement in here detailing the detection and deletion of the deprecated `addrindex` version. Otherwise a node operator might be a bit confused if `btcd` comes up and hangs for a bit while deleting the old index without any informational output.

database/ldb/tx.go
```diff
@@ -552,9 +571,65 @@ func (db *LevelDb) DeleteAddrIndex() error {
...
+	iter := db.lDb.NewIterator(bytesPrefix([]byte("a-")), db.ro)
+	numInBatch := 0
+	for iter.Next() {
+		key := iter.Key()
+		// With a 24-bit index key prefix, 1 in every 2^24 keys is a collision.
+		// We check the length to make sure we only delete address index keys.
+		// We also check the last two bytes to make sure the suffix doesn't
+		// match other types of index that are 34 bytes long.
+		if len(key) == 34 && (key[33] != byte(120) ||
+			(key[32] != byte(116) && key[32] != byte(115))) {
```
Roasbeef: To make this a bit more readable, you may want to consider changing if statement to something like:
```go
if len(key) == 34 && !(bytes.HasSuffix(key, []byte("sx")) || bytes.HasSuffix(key, []byte("tx")))
```

davecgh: I agree though I would use a global variable for each suffix to avoid a ton of copies.  Since a string is immutable, go has to make a copy each time you do `[]byte("tx" )` or `[]byte("tx" )` which is hard on the garbage collector.

jrick: Not quite true, the compiler "sees" `[]byte{"literal"}` as `[]byte{'l', 'i', 't', 'e', 'r', 'a', 'l'}`.  Only when converting from a string variable is there a copy.

jrick: Totally on board with declaring these as globals though.  Even though there's no copy from a string variable, it has to create the slice each time (otherwise, what happens if bytes from the byte literal were changed in functions it was passed to?).

database/ldb/block.go
aakselrod: ACK, fixing in next version of pull request.

database/ldb/leveldb.go
aakselrod: ACK, fixing in next version of pull request.

database/ldb/tx.go
aakselrod: I'll declare these as globals and replace them where they're used.

aakselrod: Sorry, I assume by "global" you mean "package level" variables - I'm adding these in btcd/database/ldb/leveldb.go right above the functions where they're most commonly used.

davecgh: Yes, package level.

davecgh:
This look good to me now.  OK.

dajohi:
OK

PR #359, opened 2015-03-30
Fix race from direct access of FetchTransactionStore

cjepson:
Because FetchTransactionStore in GetBlockTemplate occasionally accesses
the internal blockchain memory structure while it is being read or modified,
a race can occur. To prevent this, FetchTransactionStore is instead
routed through the internal channel for blockchain requests.

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -107,6 +107,20 @@ type processBlockResponse struct {
 	err      error
 }
 
+// fetchTransactionStoreResponse is a response sent to the reply channel of a
+// fetchTransactionStoreMsg.
+type fetchTransactionStoreResponse struct {
+	TxStore blockchain.TxStore
+	err     error
+}
+
+// fetchTransactionStoreMsg is a message type to be sent across the message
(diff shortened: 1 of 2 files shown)


blockmanager.go
```diff
@@ -1301,6 +1322,15 @@ func (b *blockManager) CalcNextRequiredDifficulty(timestamp time.Time) (uint32,
 	return response.difficulty, response.err
 }
 
+// FetchTransactionStore makes use of FetchTransactionStore on an internal
+// instance of a block chain. It is safe for concurrent access.
+func (b *blockManager) FetchTransactionStore(tx *btcutil.Tx) (blockchain.TxStore, error) {
+	reply := make(chan fetchTransactionStoreResponse, 1)
+	b.msgChan <- fetchTransactionStoreMsg{tx: tx}
```
davecgh: This needs to specify the reply channel or it will hang.

davecgh:
This reads correctly now and I've been running with the fix for a while now with no issues.

OK.

Issue #358, opened 2015-03-30
[btcec] Deterministic signatures according to RFC6979


oleganza:
Looks like Go's crypto package does not provide [RFC6979](https://tools.ietf.org/html/rfc6979)-compliant ECDSA signatures. I think we should have that in btcd.

Reference implementations:

* [C in Trezor](https://github.com/trezor/trezor-crypto/blob/9fea8f8ab377dc514e40c6fd1f7c89a74c1d8dc6/ecdsa.c#L256)
* [C in libsecp256k1](https://github.com/bitcoin/secp256k1/blob/d5b53aa81816214aa7c9e1095ff6ee1d4c13f9d2/src/hash_impl.h#L205)
* [Objective-C](https://github.com/oleganza/CoreBitcoin/blob/09eff31fad0597c0d4022be76ecf2b497f4ac161/CoreBitcoin/BTCKey.m#L267)
* [Python](https://github.com/warner/python-ecdsa/blob/4461bf3d29453454945b2cd9e666df306e60d81c/ecdsa/rfc6979.py#L54)

zmanian:
There is a go implementation that works with the Bitcoin curve here.

https://github.com/codahale/rfc6979

Strongly in favor of having deterministic signatures be the default for btcd

zmanian:
Needs some refactoring to match the coding style in btcd.

davecgh:
This is a duplicate of btcwallet#146 and I agree this is something we want.

However, I do want to point out though that it's not something btcd itself has to worry about since it doesn't do any signature generation (only verification) and a deterministic signature is a valid signature, so it will verify as expected.  All signing is done by btcwallet, so that is why the issue over there is the valid issue.

All of that said, the deterministic signing will likely need to go into the btcec package which is in this repository so I'll leave this issue open here as well.

Thanks for the reference implementation links.

oleganza:
Hey guys, please see the pull request and move discussion there. 

Thanks!

PR #360, opened 2015-03-31
Deterministic signatures according to RFC6979

oleganza:
See https://github.com/btcsuite/btcd/issues/358 for discussion.

The test vectors are the same as in [Trezor](https://github.com/trezor/trezor-crypto/blob/9fea8f8ab377dc514e40c6fd1f7c89a74c1d8dc6/tests.c#L432-L453) and [CoreBitcoin](https://github.com/oleganza/CoreBitcoin/blob/e93dd71207861b5bf044415db5fa72405e7d8fbc/CoreBitcoin/BTCKey%2BTests.m#L23-L49) implementations and cover both the raw nonce (k) and the complete DER-encoded signature (which must have canonical S).

### DIFF
--- a/btcec/example_test.go
+++ b/btcec/example_test.go
@@ -34,16 +34,14 @@ func Example_signMessage() {
 	}
 
 	// Serialize and display the signature.
-	//
-	// NOTE: This is commented out for the example since the signature
-	// produced uses random numbers and therefore will always be different.
-	//fmt.Printf("Serialized Signature: %x\n", signature.Serialize())
+	fmt.Printf("Serialized Signature: %x\n", signature.Serialize())
 
 	// Verify the signature for the message using the public key.
 	verified := signature.Verify(messageHash, pubKey)
 	fmt.Printf("Signature Verified? %v\n", verified)
 
 	// Output:
+	// Serialized Signature: 304402201008e236fa8cd0f25df4482dddbb622e8a8b26ef0ba731719458de3ccd93805b022032f8ebe514ba5f672466eba334639282616bb3c2f0ab09998037513d1f9e3d6d
 	// Signature Verified? true
 }
 
@@ -68,6 +66,7 @@ func Example_verifySignature() {
 	sigBytes, err := hex.DecodeString("30450220090ebfb3690a0ff115bb1b38b" +
 		"8b323a667b7653454f1bccb06d4bbdca42c2079022100ec95778b51e707" +
 		"1cb1205f8bde9af6592fc978b0452dafe599481c46d6b2e479")
+
 	if err != nil {
 		fmt.Println(err)
 		return
--- a/btcec/internal_test.go
+++ b/btcec/internal_test.go
@@ -69,3 +69,8 @@ func (curve *KoblitzCurve) TstDoubleJacobian(x1, y1, z1, x3, y3, z3 *fieldVal) {
 func NewFieldVal() *fieldVal {
 	return new(fieldVal)
 }
+
+// TstNonceRFC6979 makes the nonceRFC6979 function available to the test package.
+func TstNonceRFC6979(privkey *big.Int, hash []byte) *big.Int {
+	return nonceRFC6979(privkey, hash)
+}
--- a/btcec/privkey.go
+++ b/btcec/privkey.go
@@ -53,14 +53,12 @@ func (p *PrivateKey) ToECDSA() *ecdsa.PrivateKey {
 	return (*ecdsa.PrivateKey)(p)
 }
 
-// Sign wraps ecdsa.Sign to sign the provided hash (which should be the result
-// of hashing a larger message) using the private key.
+// Sign generates an ECDSA signature for the provided hash (which should be the result
+// of hashing a larger message) using the private key. Produced signature
+// is deterministic (same message and same key yield the same signature) and canonical
+// in accordance with RFC6979 and BIP0062.
 func (p *PrivateKey) Sign(hash []byte) (*Signature, error) {
-	r, s, err := ecdsa.Sign(rand.Reader, p.ToECDSA(), hash)
-	if err != nil {
-		return nil, err
-	}
-	return &Signature{R: r, S: s}, nil
+	return signRFC6979(p, hash)
 }
 
 // PrivKeyBytesLen defines the length in bytes of a serialized private key.
--- a/btcec/signature.go
+++ b/btcec/signature.go
@@ -5,11 +5,16 @@
 package btcec
 
 import (
+	"bytes"
 	"crypto/ecdsa"
 	"crypto/elliptic"
+	"crypto/hmac"
 	"errors"
 	"fmt"
+	"hash"
 	"math/big"
+
+	"github.com/btcsuite/fastsha256"
 )
 
 // Errors returned by canonicalPadding.
@@ -24,10 +29,17 @@ type Signature struct {
 	S *big.Int
 }
 
-// curve order and halforder, used to tame ECDSA malleability (see BIP-0062)
 var (
+	// Curve order and halforder, used to tame ECDSA malleability (see BIP-0062)
 	order     = new(big.Int).Set(S256().N)
 	halforder = new(big.Int).Rsh(order, 1)
+
+	// Used in RFC6979 implementation when testing the nonce for correctness
+	one = big.NewInt(1)
+
+	// oneInitializer is used to fill a byte slice with byte 0x01.  It is provided
+	// here to avoid the need to create it multiple times.
+	oneInitializer = []byte{0x01}
 )
 
 // Serialize returns the ECDSA signature in the more strict DER format.  Note
@@ -395,4 +407,126 @@ func RecoverCompact(curve *KoblitzCurve, signature,
 	}
 
 	return key, ((signature[0] - 27) & 4) == 4, nil
+}
+
+// signRFC6979 generates a deterministic ECDSA signature according to RFC 6979 and BIP 62.
+func signRFC6979(privateKey *PrivateKey, hash []byte) (*Signature, error) {
+
+	privkey := privateKey.ToECDSA()
+	N := order
+	k := nonceRFC6979(privkey.D, hash)
+	inv := new(big.Int).ModInverse(k, N)
+	r, _ := privkey.Curve.ScalarBaseMult(k.Bytes())
+	if r.Cmp(N) == 1 {
+		r.Sub(r, N)
+	}
+
+	if r.Sign() == 0 {
+		return nil, errors.New("calculated R is zero")
+	}
+
+	e := hashToInt(hash, privkey.Curve)
+	s := new(big.Int).Mul(privkey.D, r)
+	s.Add(s, e)
+	s.Mul(s, inv)
+	s.Mod(s, N)
+
+	if s.Cmp(halforder) == 1 {
+		s.Sub(N, s)
+	}
+	if s.Sign() == 0 {
+		return nil, errors.New("calculated S is zero")
+	}
+	return &Signature{R: r, S: s}, nil
+}
+
+// nonceRFC6979 generates an ECDSA nonce (`k`) deterministically according to RFC 6979.
+// It takes a 32-byte hash as an input and returns 32-byte nonce to be used in ECDSA algorithm.
+func nonceRFC6979(privkey *big.Int, hash []byte) *big.Int {
+
+	curve := S256()
+	q := curve.Params().N
+	x := privkey
+	alg := fastsha256.New
+
+	qlen := q.BitLen()
+	holen := alg().Size()
+	rolen := (qlen + 7) >> 3
+	bx := append(int2octets(x, rolen), bits2octets(hash, curve, rolen)...)
+
+	// Step B
+	v := bytes.Repeat(oneInitializer, holen)
+
+	// Step C (Go zeroes the all allocated memory)
+	k := make([]byte, holen)
+
+	// Step D
+	k = mac(alg, k, append(append(v, 0x00), bx...))
+
+	// Step E
+	v = mac(alg, k, v)
+
+	// Step F
+	k = mac(alg, k, append(append(v, 0x01), bx...))
+
+	// Step G
+	v = mac(alg, k, v)
+
+	// Step H
+	for {
+		// Step H1
+		var t []byte
+
+		// Step H2
+		for len(t)*8 < qlen {
+			v = mac(alg, k, v)
+			t = append(t, v...)
+		}
+
+		// Step H3
+		secret := hashToInt(t, curve)
+		if secret.Cmp(one) >= 0 && secret.Cmp(q) < 0 {
+			return secret
+		}
+		k = mac(alg, k, append(v, 0x00))
+		v = mac(alg, k, v)
+	}
+}
+
+// mac returns an HMAC of the given key and message.
+func mac(alg func() hash.Hash, k, m []byte) []byte {
+	h := hmac.New(alg, k)
+	h.Write(m)
+	return h.Sum(nil)
+}
+
+// https://tools.ietf.org/html/rfc6979#section-2.3.3
+func int2octets(v *big.Int, rolen int) []byte {
+	out := v.Bytes()
+
+	// left pad with zeros if it's too short
+	if len(out) < rolen {
+		out2 := make([]byte, rolen)
+		copy(out2[rolen-len(out):], out)
+		return out2
+	}
+
+	// drop most significant bytes if it's too long
+	if len(out) > rolen {
+		out2 := make([]byte, rolen)
+		copy(out2, out[len(out)-rolen:])
+		return out2
+	}
+
+	return out
+}
+
+// https://tools.ietf.org/html/rfc6979#section-2.3.4
+func bits2octets(in []byte, curve elliptic.Curve, rolen int) []byte {
+	z1 := hashToInt(in, curve)
+	z2 := new(big.Int).Sub(z1, curve.Params().N)
+	if z2.Sign() < 0 {
+		return int2octets(z1, rolen)
+	}
+	return int2octets(z2, rolen)
 }
--- a/btcec/signature_test.go
+++ b/btcec/signature_test.go
@@ -7,11 +7,13 @@ package btcec_test
 import (
 	"bytes"
 	"crypto/rand"
+	"encoding/hex"
 	"fmt"
 	"math/big"
 	"testing"
 
 	"github.com/btcsuite/btcd/btcec"
+	"github.com/btcsuite/fastsha256"
 )
 
 type signatureTest struct {
@@ -19,6 +21,19 @@ type signatureTest struct {
 	sig     []byte
 	der     bool
 	isValid bool
+}
+
+// decodeHex decodes the passed hex string and returns the resulting bytes.  It
+// panics if an error occurs.  This is only used in the tests as a helper since
+// the only way it can fail is if there is an error in the test source code.
+func decodeHex(hexStr string) []byte {
+	b, err := hex.DecodeString(hexStr)
+	if err != nil {
+		panic("invalid hex string in test source: err " + err.Error() +
+			", hex: " + hexStr)
+	}
+
+	return b
 }
 
 var signatureTests = []signatureTest{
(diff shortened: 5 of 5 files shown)


davecgh:
Thanks for the pull request.  I'm going to do a quick review now for general things, but I plan to dig into the RFC a little later and do a very careful analysis since this is obviously extremely important to get right given signature issues can easily lead to stolen funds.

btcec/privkey.go
```diff
@@ -56,11 +56,7 @@ func (p *PrivateKey) ToECDSA() *ecdsa.PrivateKey {
 // Sign wraps ecdsa.Sign to sign the provided hash (which should be the result
```
davecgh: Please update the comment to reflect reality here.

btcec/signature.go
```diff
@@ -396,3 +401,134 @@ func RecoverCompact(curve *KoblitzCurve, signature,
...
+	qlen := q.BitLen()
+	holen := alg().Size()
+	rolen := (qlen + 7) >> 3
+	bx := append(int2octets(x, rolen), bits2octets(hash, q, qlen, rolen)...)
+
+	// Step B
+	v := bytes.Repeat([]byte{0x01}, holen)
+
+	// Step C
+	k := bytes.Repeat([]byte{0x00}, holen)
```
davecgh: This can be optimized since Go promises to automatically zeroes all allocated memory using the much faster `runtime·memclr`.

```Go
// Step C.  Go promises to automatically zero all allocated memory.
k := make([]byte, holen)
```

oleganza:
@davecgh thanks, updated the code according to your remarks.

dajohi:
this project will need to be forked.

btcec/signature_test.go
```diff
@@ -491,3 +493,88 @@ func TestSignCompact(t *testing.T) {
...
+		t.Errorf("Signature for message '%s' is incorrect: %s (expected %s)", msg, hex.EncodeToString(actualSigBytes), hex.EncodeToString(sigBytes))
+	}
+}
+
+func TestRFC6979(t *testing.T) {
+	// Test vectors matching Trezor and CoreBitcoin implementations.
+	// - https://github.com/trezor/trezor-crypto/blob/9fea8f8ab377dc514e40c6fd1f7c89a74c1d8dc6/tests.c#L432-L453
+	// - https://github.com/oleganza/CoreBitcoin/blob/e93dd71207861b5bf044415db5fa72405e7d8fbc/CoreBitcoin/BTCKey%2BTests.m#L23-L49
+	verifyRFC6979Nonce(
+		"cca9fbcc1b41e5a95d369eaa6ddcff73b61a4efaa279cfc6567e8daa39cbaf50",
```
davecgh: Because these are hard coded strings in the source code, I'd prefer all of these use a function like the following `decodeHex` which panics if any errors occur rather than ignoring them.

```Go
// decodeHex converts the passed hex string into a big integer pointer and will
// panic is there is an error.  This is only provided for the hard-coded
// constants so errors in the source code can bet detected. It will only (and
// must only) be called for initialization purposes.
func decodeHex(hexStr string) []byte {
	hexBytes, err := hex.DecodeString(hexStr)
	if err != nil {
		panic("invalid hex in source file: " + hexStr)
	}
	return hexBytes
}
```

btcec/signature.go
```diff
@@ -396,3 +401,134 @@ func RecoverCompact(curve *KoblitzCurve, signature,
...
+	x := new(big.Int).SetBytes(privkey)
+	alg := fastsha256.New
+
+	qlen := q.BitLen()
+	holen := alg().Size()
+	rolen := (qlen + 7) >> 3
+	bx := append(int2octets(x, rolen), bits2octets(hash, q, qlen, rolen)...)
+
+	// Step B
+	v := bytes.Repeat([]byte{0x01}, holen)
```
davecgh: I'd suggest using a package-level variable like:

```Go
// oneInitializer is used to fill a byte slice with byte 0x01.  It is provided
// here to avoid the need to create it multiple times.
var oneInitializer = []byte{0x01}
```

Then using it here:

```Go
// Step B.
v := bytes.Repeat(oneInitializer, holen)
```

The reason is that the current code will create a new byte slice for what is essentially a constant value on every call to the function.  In general, generating extra garbage can quickly lead to really slowing Go programs down, so I like to make an effort to avoid extra allocations wherever possible.

btcec/signature.go
```diff
@@ -396,3 +401,134 @@ func RecoverCompact(curve *KoblitzCurve, signature,
 
 	return key, ((signature[0] - 27) & 4) == 4, nil
 }
+
+var one = big.NewInt(1)
```
davecgh: Please declare this in the existing variable declaration block where order and halforder are declared.

btcec/signature_test.go
```diff
@@ -491,3 +493,88 @@ func TestSignCompact(t *testing.T) {
...
+	if !bytes.Equal(actualNonce, nonce) {
+		t.Errorf("Nonce for message '%s' is incorrect: %s (expected %s)", msg, hex.EncodeToString(actualNonce), hex.EncodeToString(nonce))
+	}
+}
+
+func verifyRFC6979Signature(hexKey string, msg string, hexSignature string, t *testing.T) {
+	pkBytes, _ := hex.DecodeString(hexKey)
+	privKey, _ := btcec.PrivKeyFromBytes(btcec.S256(), pkBytes)
+	sigBytes, _ := hex.DecodeString(hexSignature)
+	hasher := fastsha256.New()
```
davecgh: This can be simplified to:

``` Go
hash := fastsha256.Sum256([]byte(msg))
```

btcec/signature.go
```diff
@@ -396,3 +401,134 @@ func RecoverCompact(curve *KoblitzCurve, signature,
...
+	if s.Cmp(halforder) == 1 {
+		s = new(big.Int).Sub(order, s)
+	}
+
+	return &Signature{R: r, S: s}, nil
+}
+
+// NonceRFC6979 generates a ECDSA nonce (`k`) deterministically according to RFC 6979.
+// It takes a 32-byte hash as an input and returns 32-byte nonce to be used in ECDSA algorithm.
+func NonceRFC6979(privkey []byte, hash []byte) *big.Int {
```
davecgh: Is there any reason this is exported?  I see it's called in the tests, but if that's the only reason for it, it should be exported via the `internal_test.go` so it's only exported during the tests instead of all the time.

oleganza: No particular reason. I didn't know there's `internal_test.go` for exporting to tests only.

btcec/signature_test.go
```diff
@@ -491,3 +506,88 @@ func TestSignCompact(t *testing.T) {
...
+	if !bytes.Equal(actualSigBytes, sigBytes) {
+		t.Errorf("Signature for message '%s' is incorrect: %s (expected %s)", msg, hex.EncodeToString(actualSigBytes), hex.EncodeToString(sigBytes))
+	}
+}
+
+func TestRFC6979(t *testing.T) {
+	// Test vectors matching Trezor and CoreBitcoin implementations.
+	// - https://github.com/trezor/trezor-crypto/blob/9fea8f8ab377dc514e40c6fd1f7c89a74c1d8dc6/tests.c#L432-L453
+	// - https://github.com/oleganza/CoreBitcoin/blob/e93dd71207861b5bf044415db5fa72405e7d8fbc/CoreBitcoin/BTCKey%2BTests.m#L23-L49
+	verifyRFC6979Nonce(
```
davecgh: We've been trying to move everything over to table driven tests to be more consistent in the code.  Would you mind updating the test code to the following (after adding all the vectors in).  Note that I also simplified a couple of things such as using `%x` in the errors instead of manually encoding the bytes.  Also, I changed the error check on sign to a `t.Errorf` instead of panic since that isn't a hard coded value.

With this code the `verifyRFC6979Nonce` and `verifyRFC6979Signature` functions aren't needed.

```Go
func TestRFC6979(t *testing.T) {
	// Test vectors matching Trezor and CoreBitcoin implementations.
	// - https://github.com/trezor/trezor-crypto/blob/9fea8f8ab377dc514e40c6fd1f7c89a74c1d8dc6/tests.c#L432-L453
	// - https://github.com/oleganza/CoreBitcoin/blob/e93dd71207861b5bf044415db5fa72405e7d8fbc/CoreBitcoin/BTCKey%2BTests.m#L23-L49
	tests := []struct {
		key       string
		msg       string
		nonce     string
		signature string
	}{
		{
			"cca9fbcc1b41e5a95d369eaa6ddcff73b61a4efaa279cfc6567e8daa39cbaf50",
			"sample",
			"2df40ca70e639d89528a6b670d9d48d9165fdc0febc0974056bdce192b8e16a3",
			"3045022100af340daf02cc15c8d5d08d7735dfe6b98a474ed373bdb5fbecf7571be52b384202205009fb27f37034a9b24b707b7c6b79ca23ddef9e25f7282e8a797efe53a8f124",
		},
		// The rest....
	}

	for i, test := range tests {
		pkBytes := decodeHex(test.key)
		hash := fastsha256.Sum256([]byte(test.msg))

		// Ensure deterministically generated nonce is the expected
		// value.
		gotNonce := btcec.TstNonceRFC6979(pkBytes, hash[:]).Bytes()
		wantNonce := decodeHex(test.nonce)
		if !bytes.Equal(gotNonce, wantNonce) {
			t.Errorf("NonceRFC6979 #%d (%s): Nonce is incorrect: "+
				"%x (expected %x)", i, test.msg, gotNonce,
				wantNonce)
			continue
		}

		// Ensure deterministically generated signature is the expected
		// value.
		privKey, _ := btcec.PrivKeyFromBytes(btcec.S256(), pkBytes)
		gotSig, err := privKey.Sign(hash[:])
		if err != nil {
			t.Errorf("Sign #%d (%s): unexpected error: %v", i,
				test.msg, err)
			continue
		}
		gotSigBytes := gotSig.Serialize()
		wantSigBytes := decodeHex(test.signature)
		if !bytes.Equal(gotSigBytes, wantSigBytes) {
			t.Errorf("Sign #%d (%s): mismatched signature: %x "+
				"(expected %x)", i, test.msg, gotSigBytes,
				wantSigBytes)
			continue
		}
	}
}
```

davecgh:
Alright that's it for now.  I'll do an in-depth review of the algorithm against the RFC later tonight.

btcec/signature.go
```diff
@@ -396,3 +408,130 @@ func RecoverCompact(curve *KoblitzCurve, signature,
 
 	return key, ((signature[0] - 27) & 4) == 4, nil
 }
+
+// SignRFC6979 generates a deterministic ECDSA signature according to RFC 6979 and BIP 62.
+func SignRFC6979(privateKey *PrivateKey, hash []byte) (*Signature, error) {
```
davecgh: Probably don't need to export this since it's scoped to the private key.

oleganza:
@davecgh do you know what's wrong with `./goclean.sh existing with status 1` that makes Travis fail?

davecgh:
Looks like it needs a gofmt on `signature_test.go`.

oleganza:
Thanks. The error was not very helpful. Lets see if it works.

davecgh:
Yeah, I noticed that as well.  I just made #361 to address that.

davecgh:
Thanks for your work on this!  It's looking good.

I imagine it's right given the tests pass, but as I said previously, I still plan to give the algorithm a through review later tonight after I finish up with some other things I'm working on.

oleganza:
Thanks for the comments, I appreciate your feedback.

davecgh:
Oh, now that we have deterministic signature, I think it would be worth updating the `SignMessage` example in `example_test.go`.  IIRC it has the print of the serialized signature commented out because it's random.  With a deterministic signature, it can be printed and checked with the `// Output: ` line.

btcec/signature.go
```diff
@@ -396,3 +408,130 @@ func RecoverCompact(curve *KoblitzCurve, signature,
...
+	}
+
+	e := hashToInt(hash, privkey.Curve)
+	s := new(big.Int).Mul(privkey.D, r)
+	s.Add(s, e)
+	s.Mul(s, inv)
+	s.Mod(s, N)
+
+	if s.Cmp(halforder) == 1 {
+		s = new(big.Int).Sub(order, s)
```
davecgh: The extra allocation can be avoided here.

```Go
s.Sub(order, s)
```

That is `s = order-S`.

btcec/signature.go
```diff
@@ -396,3 +408,130 @@ func RecoverCompact(curve *KoblitzCurve, signature,
...
+	h.Write(m)
+	return h.Sum(buf[:0])
+}
+
+// https://tools.ietf.org/html/rfc6979#section-2.3.2
+func bits2int(in []byte, qlen int) *big.Int {
+	vlen := len(in) * 8
+	v := new(big.Int).SetBytes(in)
+	if vlen > qlen {
+		v = new(big.Int).Rsh(v, uint(vlen-qlen))
```
davecgh: Can avoid the allocation here too.

```Go
v.Rsh(v, uint(vlen-qlen))
```

btcec/signature.go
```diff
@@ -396,3 +408,130 @@ func RecoverCompact(curve *KoblitzCurve, signature,
...
+
+// mac returns an HMAC of the given key and message.
+func mac(alg func() hash.Hash, k, m, buf []byte) []byte {
+	h := hmac.New(alg, k)
+	h.Write(m)
+	return h.Sum(buf[:0])
+}
+
+// https://tools.ietf.org/html/rfc6979#section-2.3.2
+func bits2int(in []byte, qlen int) *big.Int {
```
davecgh: Also, this function appears to do the exact same thing as `hashToInt`.  They should probably be reconciled.

btcec/signature.go
```diff
@@ -396,3 +408,130 @@ func RecoverCompact(curve *KoblitzCurve, signature,
...
+		if secret.Cmp(one) >= 0 && secret.Cmp(q) < 0 {
+			return secret
+		}
+		k = mac(alg, k, append(v, 0x00), k)
+		v = mac(alg, k, v, v)
+	}
+}
+
+// mac returns an HMAC of the given key and message.
+func mac(alg func() hash.Hash, k, m, buf []byte) []byte {
```
davecgh: There is no need for the extra buf parameter here.  Just pass nil into `h.Sum(nil)`.

btcec/internal_test.go
```diff
@@ -69,3 +69,9 @@ func (curve *KoblitzCurve) TstDoubleJacobian(x1, y1, z1, x3, y3, z3 *fieldVal) {
 func NewFieldVal() *fieldVal {
 	return new(fieldVal)
 }
+
+// NonceRFC6979 generates a ECDSA nonce (`k`) deterministically according to RFC 6979.
```
davecgh: `TstNonceRFC6979 makes the nonceRFC6979 function available to the test package.`

btcec/signature.go
```diff
@@ -396,3 +408,130 @@ func RecoverCompact(curve *KoblitzCurve, signature,
...
+	s.Mod(s, N)
+
+	if s.Cmp(halforder) == 1 {
+		s = new(big.Int).Sub(order, s)
+	}
+
+	return &Signature{R: r, S: s}, nil
+}
+
+// NonceRFC6979 generates a ECDSA nonce (`k`) deterministically according to RFC 6979.
```
davecgh: `nonceRFC6979`

davecgh:
Alright, I've finished carefully comparing this against the RFC and everything looks accurate.  I would like to point a few things that can be improved in the future:

* There are various allocations happening through the use of append that could be avoided by creating appropriately size byte slices and copying the relevant data into them.
* The Sign function takes an already hashed message while the signing code assumes the hash function is sha256.  This is a currently a valid assumption, however care will need to be taken should another hash function come into use.
* This code assumes that the curve is secp256k1 while this package is intended to be able to work with any curve.  There are a few other areas where this is the case however, so I don't think it would be fair to hold up this PR due to this.
* There are several optimization opportunities where the conversions between `big.Int`s and `[]byte` could be avoided.

All of that said, I see no reason any of those points should hold up this PR since it works properly and provides highly useful functionality.

@oleganza  However, I would like to see the the other comments addressed.

davecgh:
@jimmysong Can you review and verify the algorithms conform to the RFC as well?

jimmysong:
@davecgh Looking at the RFC spec now. Will have review the code soon.

oleganza:
@davecgh thanks! It's my first time I write in Go. Sorry for some obvious things like unnecessary allocations. Thanks for correcting those.

btcec/signature.go
```diff
@@ -396,3 +408,130 @@ func RecoverCompact(curve *KoblitzCurve, signature,
...
+
+// signRFC6979 generates a deterministic ECDSA signature according to RFC 6979 and BIP 62.
+func signRFC6979(privateKey *PrivateKey, hash []byte) (*Signature, error) {
+
+	privkey := privateKey.ToECDSA()
+	N := order
+	k := nonceRFC6979(privkey.D.Bytes(), hash)
+	inv := new(big.Int).ModInverse(k, N)
+	r, _ := privkey.Curve.ScalarBaseMult(k.Bytes())
+	r.Mod(r, N)
```
jimmysong: Tiny nit here, this is probably just a bit faster:

if r > N {
  r = r - N
}

btcec/signature.go
```diff
@@ -396,3 +408,130 @@ func RecoverCompact(curve *KoblitzCurve, signature,
...
+
+	if r.Sign() == 0 {
+		return nil, errors.New("calculated R is zero")
+	}
+
+	e := hashToInt(hash, privkey.Curve)
+	s := new(big.Int).Mul(privkey.D, r)
+	s.Add(s, e)
+	s.Mul(s, inv)
+	s.Mod(s, N)
```
jimmysong: Per signature spec, you should be checking if s is zero just as you do to see if r is zero:

http://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm

Honestly, this is not likely, but if you're checking r, you should check s.

btcec/signature.go
```diff
@@ -396,3 +408,130 @@ func RecoverCompact(curve *KoblitzCurve, signature,
...
+
+	if s.Cmp(halforder) == 1 {
```
jimmysong: I'm probably missing something here, but is this part (subtracting s from n if s > n/2) in the spec for rfc6979? If so, can someone point me to it?

oleganza: RFC6979 does not define S. It only defines the `k` nonce. I'm just making sure that S is BIP62-compliant from the very creation (it's also canonicalized in `Serialize` method which is useful when one wants to re-encode existing "bad" signature).

btcec/signature.go
```diff
@@ -396,3 +408,130 @@ func RecoverCompact(curve *KoblitzCurve, signature,
...
+		v = new(big.Int).Rsh(v, uint(vlen-qlen))
+	}
+	return v
+}
+
+// https://tools.ietf.org/html/rfc6979#section-2.3.3
+func int2octets(v *big.Int, rolen int) []byte {
+	out := v.Bytes()
+
+	// pad with zeros if it's too short
```
jimmysong: "left pad with zeros"

btcec/signature.go
davecgh: It's not in the spec.  It's part of BIP0062 and is used to generate signatures which can't be malleated.  Without this you can take a signature, perform the same transform without knowing the private key, and it will remain valid.  That doesn't break security, but it does mean the hash of the transaction it's signing is different, thus causing malleability.

davecgh: https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki#Low_S_values_in_signatures

btcec/signature.go
```diff
@@ -396,3 +408,130 @@ func RecoverCompact(curve *KoblitzCurve, signature,
...
+	// Step G
+	v = mac(alg, k, v, v)
+
+	// Step H
+	for {
+		// Step H1
+		var t []byte
+
+		// Step H2
+		for len(t) < qlen/8 {
```
jimmysong: shouldn't this be:

    for len(t) * 8 < qlen {

?

btcec/signature.go
jimmysong: Thanks, I'm fine with this, then.

btcec/signature.go
davecgh: It doesn't matter here because qlen = 256, hence no truncation happens on 256/8 = 32, but yes, to be proper with any qlen, it should be.

jimmysong:
@oleganza @davecgh I've checked everything and made my comments. A couple of nits and the only major thing is step H2. But even there, it'll go through fine if qlen is a multiple of 8, which it is in secp256k1.

davecgh:
@jimmysong Thanks for taking the time to look it over!

oleganza:
Addressed all the comments. Please double-check.

oleganza:
@davecgh @jimmysong Should I squash and rebase to make the merge neat?

jimmysong:
@oleganza I would check for s being zero, but that's about it.

davecgh:
@oleganza Yes, I'd like a single commit after the final OKs.  Please hold off until then though as it's much easier to review the incremental changes during the review process that address items as separate commits.

btcec/signature.go
```diff
@@ -396,3 +408,125 @@ func RecoverCompact(curve *KoblitzCurve, signature,
...
+	e := hashToInt(hash, privkey.Curve)
+	s := new(big.Int).Mul(privkey.D, r)
+	s.Add(s, e)
+	s.Mul(s, inv)
+	s.Mod(s, N)
+
+	if s.Cmp(halforder) == 1 {
+		s.Sub(N, s)
+	}
+	if r.Sign() == 0 {
```
davecgh: `s.Sign`

davecgh:
All updates look good except the typo in commit 489d98e2789349572c9ffc01ca3319fe1d3e6fd0, which I commented on (`s.Sign`).

btcec/example_test.go
```diff
@@ -37,13 +37,14 @@ func Example_signMessage() {
 	//
```
davecgh: This NOTE can be removed now since it's no longer true with you change.

oleganza:
Oops, thanks. Waiting for the final OK.

davecgh:
Looks good.  OK.

PR #363, opened 2015-04-01
Create limited RPC user.

aakselrod:
The limited user is specified with the --rpclimituser and
--rpclimitpass options (or the equivalent in the config file).
The config struct and loadConfig() are updated to take the
new options into account.  The limited user can have neither
the same username nor the same password as the admin user.

The package-level rpcLimit map in rpcserver.go specifies
the RPC commands accessible by limited users. This map
includes both HTTP/S and websocket commands.

The checkAuth function gets a new return parameter to
signify whether the user is authorized to change server
state. The result is passed to the jsonRPCRead function and
to the WebsocketHandler function in rpcwebsocket.go.

The wsClient struct is updated with an "isAdmin" field
signifying that the client is authorized to change server
state, written by WebsocketHandler and handleMessage.
The handleMessage function also checks the field to
allow or disallow an RPC call.

The following documentation files are updated:
- doc.go
- sample-btcd.conf
- docs/README.md
- docs/json_rpc_api.md
- docs/configure_rpc_server_listen_interfaces.md


davecgh:
I haven't reviewed the code yet, but I did notice there are several documentation updates that will be needed that I don't see here:

* The `sample-btcd.conf` file to include the new options with appropriate comments
* The `doc.go` file to include the new options (this allows it to show up properly on [GoDoc](http://godoc.org/github.com/btcsuite/btcd)
* [docs/README.md Controlling and Querying section](https://github.com/btcsuite/btcd/blob/master/docs/README.md#BtcctlConfig)
* [docs/json_rpc_api.md Authentication section](https://github.com/btcsuite/btcd/blob/master/docs/json_rpc_api.md#Authentication)
* [docs/configure_rpc_server_listen_interfaces.md](https://github.com/btcsuite/btcd/blob/master/docs/configure_rpc_server_listen_interfaces.md)

rpcserver.go
```diff
@@ -223,6 +223,47 @@ var rpcUnimplemented = map[string]struct{}{
...
+	"getgenerate":           struct{}{},
+	"gethashespersec":       struct{}{},
+	"getinfo":               struct{}{},
+	"getmininginfo":         struct{}{},
+	"getnettotals":          struct{}{},
+	"getnetworkhashps":      struct{}{},
+	"getrawmempool":         struct{}{},
+	"getrawtransaction":     struct{}{},
+	"gettxout":              struct{}{},
+	"ping":                  struct{}{},
```
davecgh: This one isn't read-only.

rpcserver.go
```diff
@@ -3218,7 +3276,8 @@ func createMarshalledReply(id, result interface{}, replyErr error) ([]byte, erro
 }
 
 // jsonRPCRead handles reading and responding to RPC messages.
-func (s *rpcServer) jsonRPCRead(w http.ResponseWriter, r *http.Request) {
+func (s *rpcServer) jsonRPCRead(w http.ResponseWriter, r *http.Request,
+	maywrite bool) {
```
davecgh: Minor nit: Please try to stick with camelCase for consistency.  i.e `mayWrite`.

rpcwebsocket.go
```diff
@@ -975,6 +981,16 @@ func (c *wsClient) handleMessage(msg []byte) {
 		return
 	}
 
+	// Check if the user is read-only and disconnect client if unauthorized
+	if !c.maywrite {
+		if _, ok := rpcReadOnly[request.Method]; !ok {
+			rpcsLog.Warnf("Read-only websocket client %s tried to access "+
+				"method that changes server's state", c.addr)
+			c.Disconnect()
```
davecgh: This handling is not consistent with the HTTP path which just returns an error.  I think both paths should be consistent, and I'd prefer an error be returned like it is in the HTTP path over disconnected them with no further information.

The reason disconnects are done above is because that path only happens for unauthenticated users who must send the first message as authenticate.

davecgh:
@aakselrod:  Thanks for the PR.  It looks good overall.  I think we'll want to debate the specific command list.  I think `ping` is uncontested since it clearly isn't a read-only command, but I know @dajohi has mentioned some others which are technically read-only, but might be worth leaving out to enhance privacy.

rpcserver.go
```diff
@@ -3437,6 +3501,16 @@ func newRPCServer(listenAddrs []string, s *server) (*rpcServer, error) {
...
 		quit:         make(chan int),
 	}
+	if cfg.RPCUser != "" && cfg.RPCPass != "" {
+		login := cfg.RPCUser + ":" + cfg.RPCPass
+		auth := "Basic " + base64.StdEncoding.EncodeToString([]byte(login))
+		rpc.authsha = fastsha256.Sum256([]byte(auth))
+	}
+	if cfg.RPCLimitUser != "" && cfg.RPCLimitPass != "" {
+		login := cfg.RPCLimitUser + ":" + cfg.RPCLimitPass
+		roauth := "Basic " + base64.StdEncoding.EncodeToString([]byte(login))
```
davecgh: Not a big deal, but `roauth` should probably just be `auth` here as well since it's not read-only anymore.

aakselrod: Changed to `auth` from `roauth`.

sample-btcd.conf
```diff
@@ -144,14 +144,18 @@
...
+; NOTE: The RPC server is disabled by default if rpcuser AND rpcpass, or
+; rpclimituser AND rpclimitpass, are not specified.
 ; ------------------------------------------------------------------------------
 
-; Secure the RPC API by specifying the username and password.  You must specify
-; both or the RPC server will be disabled.
-; rpcuser=whatever_username_you_want
+; Secure the RPC API by specifying the username and password.  You can also
+; specify a limited username and password.  You must specify at least one
+; full set of credentials - read-only or read-write - or the RPC server will
```
davecgh: read-only -> limited

davecgh: read-write -> admin

sample-btcd.conf
```diff
@@ -144,14 +144,18 @@
...
+; be disabled.
+; rpcuser=whatever_readwrite_username_you_want
```
davecgh: readwrite -> admin

sample-btcd.conf
aakselrod: Fixed.

sample-btcd.conf
aakselrod: Fixed.

sample-btcd.conf
```diff
@@ -144,14 +144,18 @@
...
 ; ------------------------------------------------------------------------------
 
-; Secure the RPC API by specifying the username and password.  You must specify
-; both or the RPC server will be disabled.
-; rpcuser=whatever_username_you_want
+; Secure the RPC API by specifying the username and password.  You can also
+; specify a limited username and password.  You must specify at least one
+; full set of credentials - limited or admin - or the RPC server will
+; be disabled.
+; rpcuser=whatever_readwrite_username_you_want
```
davecgh: whatever_readwrite_username_you_want -> whatever_admin_username_you_want

aakselrod: Fixed. :)

davecgh:
Alright, everything looks good!  If you can squash/rebase it, I'll merge it.

davecgh:
I've tested the following:

- Specifying the same username for limited/admin user causes an error on start as expected
- Specifying the same password for limited/admin user causes an error on start as expected
- The RPC server is disabled if neither set of credentials is provided
- The RPC server is enabled if only admin credentials are provided
- The RPC server is enabled if only limited credentials are provided
- The RPC server is enabled if both sets of credentials are provided
- Attempting to issue limited RPC commands against the limited credentials returns an error as expected
- Attempting to mix/match usernames and passwords fails credential checks as intended

PR #366, opened 2015-04-04
Minor hashing-related optimizations.

davecgh:
This pull request contains three classes of optimizations:
 - Reduce the number of unnecessary hash copies
 - Improve the performance of the DoubleSha256 function
 - A couple of minor optimizations of the ShaHash functions

The first class is a result of the Bytes function on a ShaHash making a copy of the bytes before returning them.  It really should have been named CloneBytes, but that would break the API now.

To address this, a comment has been added to the function which explicitly calls out the copy behavior.  In addition, all call sites of .Bytes on a ShaHash in the code base have been updated to simply slice the array when a copy is not needed.  This saves a significant amount of data copying.

The second optimization modifies the DoubleSha256 function to directly use fastsha256.Sum256 instead of the hasher interface.  This reduces the number of allocations needed.  A benchmark for the function has been added as well.
```
old: BenchmarkDoubleSha256  500000   3691 ns/op   192 B/op   3 allocs/op
new: BenchmarkDoubleSha256  500000   3081 ns/op    32 B/op   1 allocs/op
```
The final optimizations are for the ShaHash IsEqual and SetBytes functions which have been modified to make use of the fact the type is an array and remove an unneeded subslice.

### DIFF
--- a/blockchain/difficulty.go
+++ b/blockchain/difficulty.go
@@ -57,15 +57,14 @@ var (
 // perform math comparisons.
 func ShaHashToBig(hash *wire.ShaHash) *big.Int {
 	// A ShaHash is in little-endian, but the big package wants the bytes
-	// in big-endian.  Reverse them.  ShaHash.Bytes makes a copy, so it
-	// is safe to modify the returned buffer.
-	buf := hash.Bytes()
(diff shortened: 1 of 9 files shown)


jrick:
ok

PR #367, opened 2015-04-04
wire: Add new DoubleSha256SH function.

davecgh:
This pull request adds a new function which is similar to the DoubleSha256 function except it returns a ShaHash copy instead of a byte slice.  It also adds a new benchmark for it.

This can be a slight optimization in certain cases where the caller ultimately wants a ShaHash since it can avoid a heap allocation and additional copy to convert the result to a ShaHash (the function simply performs a type cast against the returned array which is not possible against a []byte).
```
existing: DoubleSha256     500000   3081 ns/op   32 B/op   1 allocs/op
     new: DoubleSha256SH   500000   2939 ns/op    0 B/op   0 allocs/op
```
The hashing functions for blocks and transactions have also been updated to make use of the new function since they directly return the ShaHash. The transaction change in particular is quite useful since transactions are frequently hashed and this change allows all of those hashes to avoid an additional heap allocation.

### DIFF
--- a/blockchain/merkle.go
+++ b/blockchain/merkle.go
@@ -34,11 +34,8 @@ func HashMerkleBranches(left *wire.ShaHash, right *wire.ShaHash) *wire.ShaHash {
 	copy(sha[:wire.HashSize], left[:])
 	copy(sha[wire.HashSize:], right[:])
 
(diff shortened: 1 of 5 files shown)


jrick:
ok after rebase.

PR #371, opened 2015-04-07
Create new rpctest packaged for rpc driven integration testing. 

### LINKED ISSUE #116: Create an RPC test harness
Currently the RPC server bits have no test coverage.  This needs to be rectified.

Roasbeef:
This PR addresses #116 by creating a new package (`rpctest`) which provides functionality for writing automated black-box tests to exercise the RPC interface. 

An instance of a `rpctest` consists of an active `btcd` process running in (typically) `--simnet` mode, a `btcrpcclient` instance connected to said node, and finally an embedded in-memory wallet instance (the `memWallet`) which manages any created coinbase outputs created by the mining `btcd` node. 

As part of the `SetUp` process for an RPC test, a test author can optionally opt to have a test blockchain created. The second argument to `SetUp` dictates the number of mature coinbase outputs desired. The `btcd` process will then be directed to generate a test chain of length: `100 + numMatureOutputs`. 

The embedded `memWallet` instance acts as a minimal, simple wallet for each `Harness` instance. The `memWallet` itself is a BIP 32 HD wallet capable of creating new addresses, creating fully signed transactions, creating+broadcasting a transaction paying to an arbitrary set of outputs, and querying the currently confirmed balance. 

In order to extend the test change with blocks containing arbitrary transactions, one can use the `Generate` rpc call via the exposed `btcrpcclient `connected to the active `btcd` node. Additionally, the `Harness` also exposes a secondary block generation API allowing callers to create blocks with a set of hand-selected transactions, and an arbitrary `BlockVersion` or `Timestamp`. 

After execution of test logic `TearDown` should be called, allowing the test instance to clean up created temporary directories, and shut down the running processes. 

Running multiple concurrent `rpctest.Harness` instances is supported in order to allow for test authors to exercise complex scenarios. As a result, the primary interface to create, and initialize an `rpctest.Harness` instance is concurrent safe, with shared package level private global variables protected by a `sync.Mutex`. 

To demo the new package I've written (basic) RPC tests for the following commands: 
   * `getbestblock`
   * `getblockhash`
   * `getblockcount`

The above tests are housed in a new file: `rpcserver_test.go` which utilizes `rpctest.Harness`, along with the `testing` package's new [`TestMain`](https://golang.org/pkg/testing/#hdr-Main) function. `TestMain` is utilized in order to maintain a higher degree of control over the test suite initialization, and tear down. At this point all created rpctests are housed within a single file, however as the package see more usage within `btcd`, we may opt to further split up generated tests within an internal package with an appropriate folder structure. 

Implements and closes #116

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/371)

### DIFF
--- a/rpcserver_test.go
+++ b/rpcserver_test.go
@@ -1,0 +1,133 @@
+// Copyright (c) 2016 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/rpctest"
+)
+
+func testGetBestBlock(r *rpctest.Harness, t *testing.T) {
+	_, prevbestHeight, err := r.Node.GetBestBlock()
+	if err != nil {
+		t.Fatalf("Call to `getbestblock` failed: %v", err)
+	}
+
+	// Create a new block connecting to the current tip.
+	generatedBlockHashes, err := r.Node.Generate(1)
+	if err != nil {
+		t.Fatalf("Unable to generate block: %v", err)
+	}
+
+	bestHash, bestHeight, err := r.Node.GetBestBlock()
+	if err != nil {
+		t.Fatalf("Call to `getbestblock` failed: %v", err)
+	}
+
+	// Hash should be the same as the newly submitted block.
+	if !bytes.Equal(bestHash[:], generatedBlockHashes[0][:]) {
+		t.Fatalf("Block hashes do not match. Returned hash %v, wanted "+
+			"hash %v", bestHash, generatedBlockHashes[0][:])
+	}
+
+	// Block height should now reflect newest height.
+	if bestHeight != prevbestHeight+1 {
+		t.Fatalf("Block heights do not match. Got %v, wanted %v",
+			bestHeight, prevbestHeight+1)
+	}
+}
+
+func testGetBlockCount(r *rpctest.Harness, t *testing.T) {
+	// Save the current count.
+	currentCount, err := r.Node.GetBlockCount()
+	if err != nil {
+		t.Fatalf("Unable to get block count: %v", err)
+	}
+
+	if _, err := r.Node.Generate(1); err != nil {
+		t.Fatalf("Unable to generate block: %v", err)
+	}
+
+	// Count should have increased by one.
+	newCount, err := r.Node.GetBlockCount()
+	if err != nil {
+		t.Fatalf("Unable to get block count: %v", err)
+	}
+	if newCount != currentCount+1 {
+		t.Fatalf("Block count incorrect. Got %v should be %v",
+			newCount, currentCount+1)
+	}
+}
+
+func testGetBlockHash(r *rpctest.Harness, t *testing.T) {
+	// Create a new block connecting to the current tip.
+	generatedBlockHashes, err := r.Node.Generate(1)
+	if err != nil {
+		t.Fatalf("Unable to generate block: %v", err)
+	}
+
+	info, err := r.Node.GetInfo()
+	if err != nil {
+		t.Fatalf("call to getinfo cailed: %v", err)
+	}
+
+	blockHash, err := r.Node.GetBlockHash(int64(info.Blocks))
+	if err != nil {
+		t.Fatalf("Call to `getblockhash` failed: %v", err)
+	}
+
+	// Block hashes should match newly created block.
+	if !bytes.Equal(generatedBlockHashes[0][:], blockHash[:]) {
+		t.Fatalf("Block hashes do not match. Returned hash %v, wanted "+
+			"hash %v", blockHash, generatedBlockHashes[0][:])
+	}
+}
+
+var rpcTestCases = []rpctest.HarnessTestCase{
+	testGetBestBlock,
+	testGetBlockCount,
+	testGetBlockHash,
+}
+
+var primaryHarness *rpctest.Harness
+
+func TestMain(m *testing.M) {
+	var err error
+	primaryHarness, err = rpctest.New(&chaincfg.SimNetParams, nil, nil)
+	if err != nil {
+		fmt.Println("unable to create primary harness: ", err)
+		os.Exit(1)
+	}
+
+	// Initialize the primary mining node with a chain of length 125,
+	// providing 25 mature coinbases to allow spending from for testing
+	// purposes.
+	if err = primaryHarness.SetUp(true, 25); err != nil {
+		fmt.Println("unable to setup test chain: ", err)
+		os.Exit(1)
+	}
+
+	exitCode := m.Run()
+
+	// Clean up the primary harness created above. This includes removing
+	// all temporary directories, and shutting down any created processes.
+	if err := primaryHarness.TearDown(); err != nil {
+		fmt.Println("unable to setup test chain: ", err)
+		os.Exit(1)
+	}
+
+	os.Exit(exitCode)
+}
+
+func TestRpcServer(t *testing.T) {
+	for _, testCase := range rpcTestCases {
+		testCase(primaryHarness, t)
+	}
+}
--- a/rpctest/README.md
+++ b/rpctest/README.md
@@ -1,0 +1,33 @@
+rpctest
+=======
+
+[![Build Status](http://img.shields.io/travis/btcsuite/btcd.svg)]
+(https://travis-ci.org/btcsuite/btcd) [![ISC License]
+(http://img.shields.io/badge/license-ISC-blue.svg)](http://copyfree.org)
+[![GoDoc](https://img.shields.io/badge/godoc-reference-blue.svg)]
+(http://godoc.org/github.com/btcsuite/btcd/rpctest)
+
+Package rpctest provides a btcd-specific RPC testing harness crafting and
+executing integration tests by driving a `btcd` instance via the `RPC`
+interface. Each instance of an active harness comes equipped with a simple
+in-memory HD wallet capable of properly syncing to the generated chain,
+creating new addresses, and crafting fully signed transactions paying to an
+arbitrary set of outputs. 
+
+This package was designed specifically to act as an RPC testing harness for
+`btcd`. However, the constructs presented are general enough to be adapted to
+any project wishing to programmatically drive a `btcd` instance of its
+systems/integration tests. 
+
+## Installation and Updating
+
+```bash
+$ go get -u github.com/btcsuite/btcd/rpctest
+```
+
+## License
+
+
+Package rpctest is licensed under the [copyfree](http://copyfree.org) ISC
+License.
+
--- a/rpctest/blockgen.go
+++ b/rpctest/blockgen.go
@@ -1,0 +1,173 @@
+// Copyright (c) 2016 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package rpctest
+
+import (
+	"errors"
+	"math"
+	"math/big"
+	"runtime"
+	"time"
+
+	"github.com/btcsuite/btcd/blockchain"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+)
+
+// solveBlock attempts to find a nonce which makes the passed block header hash
+// to a value less than the target difficulty. When a successful solution is
+// found true is returned and the nonce field of the passed header is updated
+// with the solution. False is returned if no solution exists.
+func solveBlock(header *wire.BlockHeader, targetDifficulty *big.Int) bool {
+	// sbResult is used by the solver goroutines to send results.
+	type sbResult struct {
+		found bool
+		nonce uint32
+	}
+
+	// solver accepts a block header and a nonce range to test. It is
+	// intended to be run as a goroutine.
+	quit := make(chan bool)
+	results := make(chan sbResult)
+	solver := func(hdr wire.BlockHeader, startNonce, stopNonce uint32) {
+		// We need to modify the nonce field of the header, so make sure
+		// we work with a copy of the original header.
+		for i := startNonce; i >= startNonce && i <= stopNonce; i++ {
+			select {
+			case <-quit:
+				return
+			default:
+				hdr.Nonce = i
+				hash := hdr.BlockHash()
+				if blockchain.HashToBig(&hash).Cmp(targetDifficulty) <= 0 {
+					results <- sbResult{true, i}
+					return
+				}
+			}
+		}
+		results <- sbResult{false, 0}
+	}
+
+	startNonce := uint32(0)
+	stopNonce := uint32(math.MaxUint32)
+	numCores := uint32(runtime.NumCPU())
+	noncesPerCore := (stopNonce - startNonce) / numCores
+	for i := uint32(0); i < numCores; i++ {
+		rangeStart := startNonce + (noncesPerCore * i)
+		rangeStop := startNonce + (noncesPerCore * (i + 1)) - 1
+		if i == numCores-1 {
+			rangeStop = stopNonce
+		}
+		go solver(*header, rangeStart, rangeStop)
+	}
+	for i := uint32(0); i < numCores; i++ {
+		result := <-results
+		if result.found {
+			close(quit)
+			header.Nonce = result.nonce
+			return true
+		}
+	}
+
+	return false
+}
+
+// standardCoinbaseScript returns a standard script suitable for use as the
+// signature script of the coinbase transaction of a new block. In particular,
+// it starts with the block height that is required by version 2 blocks.
+func standardCoinbaseScript(nextBlockHeight int32, extraNonce uint64) ([]byte, error) {
+	return txscript.NewScriptBuilder().AddInt64(int64(nextBlockHeight)).
+		AddInt64(int64(extraNonce)).Script()
+}
+
+// createCoinbaseTx returns a coinbase transaction paying an appropriate
+// subsidy based on the passed block height to the provided address.
+func createCoinbaseTx(coinbaseScript []byte, nextBlockHeight int32,
+	addr btcutil.Address, net *chaincfg.Params) (*btcutil.Tx, error) {
+
+	// Create the script to pay to the provided payment address.
+	pkScript, err := txscript.PayToAddrScript(addr)
+	if err != nil {
+		return nil, err
+	}
+
+	tx := wire.NewMsgTx()
+	tx.AddTxIn(&wire.TxIn{
+		// Coinbase transactions have no inputs, so previous outpoint is
+		// zero hash and max index.
+		PreviousOutPoint: *wire.NewOutPoint(&chainhash.Hash{},
+			wire.MaxPrevOutIndex),
+		SignatureScript: coinbaseScript,
+		Sequence:        wire.MaxTxInSequenceNum,
+	})
+	tx.AddTxOut(&wire.TxOut{
+		Value:    blockchain.CalcBlockSubsidy(nextBlockHeight, net),
+		PkScript: pkScript,
+	})
+	return btcutil.NewTx(tx), nil
+}
+
+// createBlock creates a new block building from the previous block.
+func createBlock(prevBlock *btcutil.Block, inclusionTxs []*btcutil.Tx,
+	blockVersion int32, blockTime time.Time,
+	miningAddr btcutil.Address, net *chaincfg.Params) (*btcutil.Block, error) {
+
+	prevHash := prevBlock.Hash()
+	blockHeight := prevBlock.Height() + 1
+
+	// If a target block time was specified, then use that as the header's
+	// timestamp. Otherwise, add one second to the previous block unless
+	// it's the genesis block in which case use the current time.
+	var ts time.Time
+	switch {
+	case !blockTime.IsZero():
+		ts = blockTime
+	default:
+		ts = prevBlock.MsgBlock().Header.Timestamp.Add(time.Second)
+	}
+
+	extraNonce := uint64(0)
+	coinbaseScript, err := standardCoinbaseScript(blockHeight, extraNonce)
+	if err != nil {
+		return nil, err
+	}
+	coinbaseTx, err := createCoinbaseTx(coinbaseScript, blockHeight,
+		miningAddr, net)
+	if err != nil {
+		return nil, err
+	}
+
+	// Create a new block ready to be solved.
+	blockTxns := []*btcutil.Tx{coinbaseTx}
+	if inclusionTxs != nil {
+		blockTxns = append(blockTxns, inclusionTxs...)
+	}
+	merkles := blockchain.BuildMerkleTreeStore(blockTxns)
+	var block wire.MsgBlock
+	block.Header = wire.BlockHeader{
+		Version:    blockVersion,
+		PrevBlock:  *prevHash,
+		MerkleRoot: *merkles[len(merkles)-1],
+		Timestamp:  ts,
+		Bits:       net.PowLimitBits,
+	}
+	for _, tx := range blockTxns {
+		if err := block.AddTransaction(tx.MsgTx()); err != nil {
+			return nil, err
+		}
+	}
+
+	found := solveBlock(&block.Header, net.PowLimit)
+	if !found {
+		return nil, errors.New("Unable to solve block")
+	}
+
+	utilBlock := btcutil.NewBlock(&block)
+	utilBlock.SetHeight(blockHeight)
+	return utilBlock, nil
+}
--- a/rpctest/doc.go
+++ b/rpctest/doc.go
@@ -1,0 +1,12 @@
+// Package rpctest provides a btcd-specific RPC testing harness crafting and
+// executing integration tests by driving a `btcd` instance via the `RPC`
+// interface. Each instance of an active harness comes equipped with a simple
+// in-memory HD wallet capable of properly syncing to the generated chain,
+// creating new addresses, and crafting fully signed transactions paying to an
+// arbitrary set of outputs.
+//
+// This package was designed specifically to act as an RPC testing harness for
+// `btcd`. However, the constructs presented are general enough to be adapted to
+// any project wishing to programmatically drive a `btcd` instance of its
+// systems/integration tests.
+package rpctest
--- a/rpctest/memwallet.go
+++ b/rpctest/memwallet.go
@@ -1,0 +1,543 @@
+// Copyright (c) 2016 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package rpctest
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/btcsuite/btcd/blockchain"
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcrpcclient"
+	"github.com/btcsuite/btcutil"
+	"github.com/btcsuite/btcutil/hdkeychain"
+)
+
+var (
+	// hdSeed is the BIP 32 seed used by the memWallet to initialize it's
+	// HD root key. This value is hard coded in order to ensure
+	// deterministic behavior across test runs.
+	hdSeed = [chainhash.HashSize]byte{
+		0x79, 0xa6, 0x1a, 0xdb, 0xc6, 0xe5, 0xa2, 0xe1,
+		0x39, 0xd2, 0x71, 0x3a, 0x54, 0x6e, 0xc7, 0xc8,
+		0x75, 0x63, 0x2e, 0x75, 0xf1, 0xdf, 0x9c, 0x3f,
+		0xa6, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+	}
+)
+
+// utxo represents an unspent output spendable by the memWallet. The maturity
+// height of the transaction is recorded in order to properly observe the
+// maturity period of direct coinbase outputs.
+type utxo struct {
+	pkScript       []byte
+	value          btcutil.Amount
+	keyIndex       uint32
+	maturityHeight int32
+	isLocked       bool
+}
+
+// isMature returns true if the target utxo is considered "mature" at the
+// passed block height. Otherwise, false is returned.
+func (u *utxo) isMature(height int32) bool {
+	return height >= u.maturityHeight
+}
+
+// chainUpdate encapsulates an update to the current main chain. This struct is
+// used to sync up the memWallet each time a new block is connected to the main
+// chain.
+type chainUpdate struct {
+	blockHash   *chainhash.Hash
+	blockHeight int32
+}
+
+// undoEntry is functionally the opposite of a chainUpdate. An undoEntry is
+// created for each new block received, then stored in a log in order to
+// properly handle block re-orgs.
+type undoEntry struct {
+	utxosDestroyed map[wire.OutPoint]*utxo
+	utxosCreated   []wire.OutPoint
+}
+
+// memWallet is a simple in-memory wallet whose purpose is to provide basic
+// wallet functionality to the harness. The wallet uses a hard-coded HD key
+// hierarchy which promotes reproducibility between harness test runs.
+type memWallet struct {
+	coinbaseKey  *btcec.PrivateKey
+	coinbaseAddr btcutil.Address
+
+	// hdRoot is the root master private key for the wallet.
+	hdRoot *hdkeychain.ExtendedKey
+
+	// hdIndex is the next available key index offset from the hdRoot.
+	hdIndex uint32
+
+	// currentHeight is the latest height the wallet is known to be synced
+	// to.
+	currentHeight int32
+
+	// addrs tracks all addresses belonging to the wallet. The addresses
+	// are indexed by their keypath from the hdRoot.
+	addrs map[uint32]btcutil.Address
+
+	// utxos is the set of utxos spendable by the wallet.
+	utxos map[wire.OutPoint]*utxo
+
+	// reorgJournal is a map storing an undo entry for each new block
+	// received. Once a block is disconnected, the undo entry for the
+	// particular height is evaluated, thereby rewinding the effect of the
+	// disconnected block on the wallet's set of spendable utxos.
+	reorgJournal map[int32]*undoEntry
+
+	chainUpdates      []*chainUpdate
+	chainUpdateSignal chan struct{}
+	chainMtx          sync.Mutex
+
+	net *chaincfg.Params
+
+	rpc *btcrpcclient.Client
+
+	sync.RWMutex
+}
+
+// newMemWallet creates and returns a fully initialized instance of the
+// memWallet given a particular blockchain's parameters.
+func newMemWallet(net *chaincfg.Params, harnessID uint32) (*memWallet, error) {
+	// The wallet's final HD seed is: hdSeed || harnessID. This method
+	// ensures that each harness instance uses a deterministic root seed
+	// based on its harness ID.
+	var harnessHDSeed [chainhash.HashSize + 4]byte
+	copy(harnessHDSeed[:], hdSeed[:])
+	binary.BigEndian.PutUint32(harnessHDSeed[:chainhash.HashSize], harnessID)
+
+	hdRoot, err := hdkeychain.NewMaster(harnessHDSeed[:], net)
+	if err != nil {
+		return nil, nil
+	}
+
+	// The first child key from the hd root is reserved as the coinbase
+	// generation address.
+	coinbaseChild, err := hdRoot.Child(0)
+	if err != nil {
+		return nil, err
+	}
+	coinbaseKey, err := coinbaseChild.ECPrivKey()
+	if err != nil {
+		return nil, err
+	}
+	coinbaseAddr, err := keyToAddr(coinbaseKey, net)
+	if err != nil {
+		return nil, err
+	}
+
+	// Track the coinbase generation address to ensure we properly track
+	// newly generated bitcoin we can spend.
+	addrs := make(map[uint32]btcutil.Address)
+	addrs[0] = coinbaseAddr
+
+	return &memWallet{
+		net:               net,
+		coinbaseKey:       coinbaseKey,
+		coinbaseAddr:      coinbaseAddr,
+		hdIndex:           1,
+		hdRoot:            hdRoot,
+		addrs:             addrs,
+		utxos:             make(map[wire.OutPoint]*utxo),
+		chainUpdateSignal: make(chan struct{}),
+		reorgJournal:      make(map[int32]*undoEntry),
+	}, nil
+}
+
+// Start launches all goroutines required for the wallet to function properly.
+func (m *memWallet) Start() {
+	go m.chainSyncer()
+}
+
+// SyncedHeight returns the height the wallet is known to be synced to.
+//
+// This function is safe for concurrent access.
+func (m *memWallet) SyncedHeight() int32 {
+	m.RLock()
+	defer m.RUnlock()
+	return m.currentHeight
+}
+
+// SetRPCClient saves the passed rpc connection to btcd as the wallet's
+// personal rpc connection.
+func (m *memWallet) SetRPCClient(rpcClient *btcrpcclient.Client) {
+	m.rpc = rpcClient
+}
+
+// IngestBlock is a call-back which is to be triggered each time a new block is
+// connected to the main chain. Ingesting a block updates the wallet's internal
+// utxo state based on the outputs created and destroyed within each block.
+func (m *memWallet) IngestBlock(blockHash *chainhash.Hash, height int32, t time.Time) {
+	// Append this new chain update to the end of the queue of new chain
+	// updates.
+	m.chainMtx.Lock()
+	m.chainUpdates = append(m.chainUpdates, &chainUpdate{blockHash, height})
+	m.chainMtx.Unlock()
+
+	// Launch a goroutine to signal the chainSyncer that a new update is
+	// available. We do this in a new goroutine in order to avoid blocking
+	// the main loop of the rpc client.
+	go func() {
+		m.chainUpdateSignal <- struct{}{}
+	}()
+}
+
+// chainSyncer is a goroutine dedicated to processing new blocks in order to
+// keep the wallet's utxo state up to date.
+//
+// NOTE: This MUST be run as a goroutine.
+func (m *memWallet) chainSyncer() {
+	var update *chainUpdate
+
+	for range m.chainUpdateSignal {
+		// A new update is available, so pop the new chain update from
+		// the front of the update queue.
+		m.chainMtx.Lock()
+		update = m.chainUpdates[0]
+		m.chainUpdates[0] = nil // Set to nil to prevent GC leak.
+		m.chainUpdates = m.chainUpdates[1:]
+		m.chainMtx.Unlock()
+
+		// Fetch the new block so we can process it shortly below.
+		block, err := m.rpc.GetBlock(update.blockHash)
+		if err != nil {
+			return
+		}
+
+		// Update the latest synced height, then process each
+		// transaction in the block creating and destroying utxos
+		// within the wallet as a result.
+		m.Lock()
+		m.currentHeight = update.blockHeight
+		undo := &undoEntry{
+			utxosDestroyed: make(map[wire.OutPoint]*utxo),
+		}
+		for _, tx := range block.Transactions() {
+			mtx := tx.MsgTx()
+			isCoinbase := blockchain.IsCoinBaseTx(mtx)
+
+			m.evalOutputs(mtx.TxOut, tx.Hash(), isCoinbase, undo)
+			m.evalInputs(mtx.TxIn, undo)
+		}
+
+		// Finally, record the undo entry for this block so we can
+		// properly update our internal state in response to the block
+		// being re-org'd from the main chain.
+		m.reorgJournal[update.blockHeight] = undo
+		m.Unlock()
+	}
+}
+
+// evalOutputs evaluates each of the passed outputs, creating a new matching
+// utxo within the wallet if we're able to spend the output.
+func (m *memWallet) evalOutputs(outputs []*wire.TxOut, txHash *chainhash.Hash,
+	isCoinbase bool, undo *undoEntry) {
+
+	for i, output := range outputs {
+		pkScript := output.PkScript
+
+		// Scan all the addresses we currently control to see if the
+		// output is paying to us.
+		for keyIndex, addr := range m.addrs {
+			pkHash := addr.ScriptAddress()
+			if !bytes.Contains(pkScript, pkHash) {
+				continue
+			}
+
+			// If this is a coinbase output, then we mark the
+			// maturity height at the proper block height in the
+			// future.
+			var maturityHeight int32
+			if isCoinbase {
+				maturityHeight = m.currentHeight + int32(m.net.CoinbaseMaturity)
+			}
+
+			op := wire.OutPoint{Hash: *txHash, Index: uint32(i)}
+			m.utxos[op] = &utxo{
+				value:          btcutil.Amount(output.Value),
+				keyIndex:       keyIndex,
+				maturityHeight: maturityHeight,
+				pkScript:       pkScript,
+			}
+			undo.utxosCreated = append(undo.utxosCreated, op)
+		}
+	}
+}
+
+// evalInputs scans all the passed inputs, destroying any utxos within the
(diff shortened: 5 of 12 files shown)


davecgh:
Thanks.  I haven't reviewed this yet, but from the description, I will say I'd really prefer this does not depend on btcsim.

davecgh:
I generally like the overall idea here.  Nice work!  Obviously, as you've already noted, there are various TODOs and unfinished doco which I didn't comment on since it's still a WIP.

One thing I suspect would be nicer though is rather than launching a new btcd instance for every test by default, I'd probably do majority of the stuff in `rpcserver_test.go` using a single entry point and setup which each individual test issuing the RPC calls and comparing to what the state should be after the initial setup.

I like the fact you made it possible to launch multiple, as there will definitely be some cases where this is needed, however, I don't think more than one is needed for the vast majority.

The main reason I bring this up is related to how long it will take for tests to complete.  As the number of RPC server tests grow, if each one is spinning up a new instance, it could easily get to the point it takes a long time to run the tests.  Travis has a limit on how long tests can run, so this could lead to erroneous failure reporting.

davecgh:
There is now a `generate` RPC which can be used to create an arbitrary number of blocks.  It might be useful to use it where applicable in this as well.

davecgh:
@Roasbeef I believe you mentioned in IRC you had updated this.  I see it's failing Travis at the moment.  Is it ready for review aside from the the fact it can't find the executable on Travis?  We'll obviously need to solve that.

Roasbeef:
Hey @davecgh, yeah, it seems Travis is unable to find the executable while running the tests involving `rpctest`. Although I haven't looked into it extensively, I suspect there'll be an easy fix. 

There are still a few outstanding TODO's w.r.t this PR I haven't quite gotten to yet. However, they're pretty minor (comments, docs, some more test cases), so I'd say this PR is ready for a preliminary review.

I've been using this package to write [integration tests within `lnd`](https://github.com/LightningNetwork/lnd/blob/master/lnwallet/wallet_test.go). So far, I've found writing tests leveraging `rpctest` to be rather streamlined.

davecgh:
Thanks for the update.

davecgh:
Can we get this working on Travis?  It can't be merged as long as Travis is failing as it would cause every PR from then on to no longer pass tests.

rpctest/README.md
```diff
@@ -0,0 +1,38 @@
+rpctest
+=======
+
+[![Build Status](https://travis-ci.org/btcsuite/btcd.png?branch=master)]
+(https://travis-ci.org/btcsuite/btcd)
+
+Package rpctests provides a btcd-specific RPC testing harness for excersing 
+regressions via the RPC interface.
+...TODO
```
davecgh: Looks like a missed TODO here!

rpctest/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+You can also view the documentation locally once the package is installed with
+the `godoc` tool by running `godoc -http=":6060"` and pointing your browser to
+http://localhost:6060/pkg/github.com/btcsuite/btcd/rpctest
+
+
+## Installation
+
+```bash
+$ go get github.com/btcsuite/btcd/rpctest
```
davecgh: We've been trying to switch over to using the `-u` flag since it work for update as well as initial install.

`go get -u github.com/btcsuite/btcd/rpctest`

rpctest/README.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+## Usage
+
+...TODO
```
davecgh: TODO here as well.

rpctest/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## License
+
+Package coinset is licensed under the [copyfree](http://copyfree.org) ISC
```
davecgh: `rpctest`

rpctest/utils.go
```diff
@@ -0,0 +1,172 @@
...
+
+		delete(testInstances, harness.testNodeDir)
+	}
+
+	return nil
+}
+
+// ActiveHarnesses returns a slice of all currently active test harnesses. A
+// test harness if considered "active" if it has been created, but not yet
+// teared down.
```
davecgh: torn down.

davecgh:
This has a lot of test failures and it quite slow.  I suspect the slowness has to do with what I mentioned before about launching a new node for every new test.  Ideally, there would be a single setup that starts the node so that the tests use a single entry point and setup with each individual test issuing the RPC calls and comparing to what the state should be after the initial setup.

```
$ go test -v
=== RUN   TestSetUp
--- FAIL: TestSetUp (10.57s)
        rpc_harness_test.go:60: Create test datadir was not deleted.
=== RUN   TestCoinbaseSpend
--- PASS: TestCoinbaseSpend (10.34s)
=== RUN   TestConnectNode
--- FAIL: TestTearDownAll (21.43s)
        rpc_harness_test.go:186: unable to teardown all harnesses: remove D:\msys64\tmp\rpctest-4765005391\simnet\wallet.db: The process cannot access the file because it is being used by another process.
--- PASS: TestConnectNode (20.38s)
=== RUN   TestTearDownAll
--- FAIL: TestTearDownAll (20.83s)
        rpc_harness_test.go:181: unable to complete rpctest setup: connection timedout
=== RUN   TestActiveHarnesses
--- PASS: TestActiveHarnesses (0.11s)
=== RUN   TestJoinMempools
FAIL
--- PASS: TestJoinMempools (21.06s)
=== RUN   TestJoinBlocks
--- PASS: TestJoinBlocks (19.34s)
FAIL
exit status 1
FAIL    github.com/btcsuite/btcd/rpctest        104.805s
```

rpctest/rpc_harness.go
```diff
@@ -0,0 +1,360 @@
...
+// Harness fully encapsulates an active btcd process, along with an embdedded
+// btcwallet in order to provide a unified platform for creating rpc driven
+// integration tests involving btcd. The active btcd node will typically be
+// run in simnet mode in order to allow for easy generation of test blockchains.
+// Additionally, a special method is provided which allows on to easily generate
+// coinbase spends. The active btcd process if fully managed by Harness, which
+// handles the necessary initialization, and teardown of the process along with
+// any temporary directories created as a result. Multiple Harness instances may
+// be run concurrently, in order to allow for testing complex scenarios involving
+// multuple nodes.
```
davecgh: multiple

rpctest/rpc_harness.go
```diff
@@ -0,0 +1,360 @@
...
+
+	// Used to protest concurrent access to above declared variables.
+	harnessStateMtx sync.RWMutex
+)
+
+// Harness fully encapsulates an active btcd process, along with an embdedded
+// btcwallet in order to provide a unified platform for creating rpc driven
+// integration tests involving btcd. The active btcd node will typically be
+// run in simnet mode in order to allow for easy generation of test blockchains.
+// Additionally, a special method is provided which allows on to easily generate
```
davecgh: allows on**e**

rpctest/rpc_harness.go
```diff
@@ -0,0 +1,360 @@
...
+// coinbase spends. The active btcd process if fully managed by Harness, which
```
davecgh: process i**s**

rpctest/utils.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	return nil
+}
+
+// ConnectNode establishes a new peer-to-peer connection from nodeA to nodeB.
+// The connection made is flagged as persistent, therefore in the case of
+// disconnects, nodeA will attempt to reestablish a connection to nodeB.
+func ConnectNode(from *Harness, to *Harness) error {
+	var err error
```
davecgh: This isn't needed since it's defined as a part of the `GetPeerInfo` below.

davecgh:
Is the wallet really needed here?  I think everything could be done using `hdkeychain` and it would remove a huge dependency on wallet and help speed things up as well.

rpctest/README.md
```diff
@@ -4,9 +4,9 @@ rpctest
 [![Build Status](https://travis-ci.org/btcsuite/btcd.png?branch=master)]
 (https://travis-ci.org/btcsuite/btcd)
 
-Package rpctests provides a btcd-specific RPC testing harness for excersing 
-regressions via the RPC interface.
-...TODO
+Package rpctests provides a btcd-specific RPC testing harness crafting and executing integration tests by driving a `btcd` instance via the `RPC` interface. Each instance of a active harness comes equipped with a simple in-memory HD wallet capable of properly syncing to the generated chain, creating new addreses, and crafting fully signed transactions paying to an abitrary set of outputs. 
+
+This package was designed specifically to act as an RPC testing harness for `btcd`. However, the constructs presented are general enoguh to be adapted to any project wishing to programatially drive a `btcd` instance of its systems/integration tests.
```
davecgh: enoguh -> enough

Roasbeef:
I've just pushed some changes which should push this PR much closer (if not already there) to completon. 

I've ripped out `btcwallet` in favor of a simple in-memory wallet, the `memWallet`. Additionally I've re-instated the older `GenerateAndSubmitBlocks` method with additional support for custom block versions and timestamps. 

Finally the `rpc_harness_test.go` file has been refactored a bit to emulate the style in `rpcserver_test.go`. Namely, it attempts to utilize a single `rpctest.Harness` throughout the testing run which along with the `memWallet` addition speeds up the tests quite a bit.

davecgh:
Please add the new required packages to `glide.yaml` as well.  I realize the lock file picked them up, but they should be specified in the yaml too.

```diff
diff --git a/glide.yaml b/glide.yaml
index 5298a27..69cfe0f 100644
--- a/glide.yaml
+++ b/glide.yaml
@@ -6,6 +6,8 @@ import:
   - .
   - bloom
   - base58
+  - hdkeychain
+- package: github.com/btcsuite/btcrpcclient
 - package: github.com/btcsuite/fastsha256
 - package: github.com/btcsuite/go-flags
 - package: github.com/btcsuite/go-socks
```

Roasbeef:
Sure, I just pushed commit updating `glide.yaml` 👍

rpctest/README.md
```diff
@@ -0,0 +1,33 @@
+rpctest
+=======
+
+[![Build Status](https://travis-ci.org/btcsuite/btcd.png?branch=master)]
+(https://travis-ci.org/btcsuite/btcd)
+
+Package rpctests provides a btcd-specific RPC testing harness crafting and executing integration tests by driving a `btcd` instance via the `RPC` interface. Each instance of a active harness comes equipped with a simple in-memory HD wallet capable of properly syncing to the generated chain, creating new addresses, and crafting fully signed transactions paying to an arbitrary set of outputs.
```
davecgh: of *an* active

rpctest/README.md
```diff
@@ -0,0 +1,33 @@
...
+=======
+
+[![Build Status](https://travis-ci.org/btcsuite/btcd.png?branch=master)]
+(https://travis-ci.org/btcsuite/btcd)
+
+Package rpctests provides a btcd-specific RPC testing harness crafting and executing integration tests by driving a `btcd` instance via the `RPC` interface. Each instance of a active harness comes equipped with a simple in-memory HD wallet capable of properly syncing to the generated chain, creating new addresses, and crafting fully signed transactions paying to an arbitrary set of outputs. 
+
+This package was designed specifically to act as an RPC testing harness for `btcd`. However, the constructs presented are general enough to be adapted to any project wishing to programmatically drive a `btcd` instance of its systems/integration tests. 
+
+## Documentation
```
davecgh: Would you make this match the new style with the rest of the project?  Namely this section was removed in favor a doc badge at the top.  Also they all have a badge for the license too (although the license section below wasn't removed).

```
 [![ISC License]
(http://img.shields.io/badge/license-ISC-blue.svg)](http://copyfree.org)
[![GoDoc](https://img.shields.io/badge/godoc-reference-blue.svg)]
(http://godoc.org/github.com/btcsuite/btcd/rpctest)
```

rpctest/README.md
```diff
@@ -0,0 +1,33 @@
...
+Full `go doc` style documentation for the project can be viewed online without
+installing this package by using the GoDoc site
+[here](http://godoc.org/github.com/btcsuite/btcd/rpctest).
+
+You can also view the documentation locally once the package is installed with
+the `godoc` tool by running `godoc -http=":6060"` and pointing your browser to
+http://localhost:6060/pkg/github.com/btcsuite/btcd/rpctest
+
+
+## Installation
```
davecgh: Consistenty with other packages:

```## Installation and Updating```

File: rpcserver_test.go [hunk 14]
```diff
@@ -0,0 +1,129 @@
+package main
```
davecgh: Needs a license header.

rpctest/blockgen.go
```diff
@@ -0,0 +1,175 @@
+// Copyright (c) 2016 Conformal Systems LLC.
```
davecgh:
```
// Copyright (c) 2016 The btcsuite developers
```

rpctest/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package rpctest ....
```
davecgh: Looks like you forgot to complete this!

rpctest/memwallet.go
```diff
@@ -0,0 +1,517 @@
...
+
+	sync.RWMutex
+}
+
+// newMemWallet creates and returns a fully initialized instance of the
+// memWallet given a particular blockchain's parameters.
+func newMemWallet(net *chaincfg.Params, harnessID uint32) (*memWallet, error) {
+	// The wallet's final HD seed is: hdSeed || harnessID. This method
+	// ensures that each harness instance uses a deterministic root seed
+	// based on it's harness ID.
```
davecgh: it's -> its

rpctest/memwallet.go
```diff
@@ -0,0 +1,517 @@
...
+
+		balance += utxo.value
+	}
+
+	return balance
+}
+
+// keyToAddr maps the passed private to corresponding p2pkh address.
+func keyToAddr(key *btcec.PrivateKey, net *chaincfg.Params) (btcutil.Address, error) {
+	keyHash := btcutil.Hash160(key.PubKey().SerializeCompressed())
```
davecgh: This doesn't really matter much, but should this just create an `AddressPubKey` and allow it to return the `AddressPubKeyHash` in order to avoid knowing the internal mechanics as much?

```Go
pubKeyAddr, err := btcutil.NewAddressPubKey(key.PubKey().SerializeCompressed())
if err != nil {
	return nil, err
}
return pubKeyAddr.AddressPubKeyHash()

File: rpctest/memwallet.go [hunk 19]
```diff
@@ -0,0 +1,517 @@
...
+		update = m.chainUpdates[0]
+		m.chainUpdates[0] = nil // Set to nil to prevent GC leak.
+		m.chainUpdates = m.chainUpdates[1:]
+
+		m.chainMtx.Unlock()
+
+		// Fetch the new block so we can process it shortly below.
+		block, err := m.rpc.GetBlock(update.blockHash)
+		if err != nil {
+			return
```
davecgh: I want to say this should probably somehow be signaled, but I think it won't matter since the intent of this is testing and failure here will result in an invalid wallet state which means it won't match the expected test state and will lead to a test failure anyways.

rpctest/memwallet.go
```diff
@@ -0,0 +1,517 @@
...
+
+// chainSyncer is a goroutine dedicated to processing new blocks in order to
+// keep the wallet's utxo state up to date.
+//
+// NOTE: This MUST be run as a goroutine.
+func (m *memWallet) chainSyncer() {
+	var update *chainUpdate
+
+	for range m.chainUpdateSignal {
+		m.chainMtx.Lock()
```
davecgh: Mind wrapping this around the critical section closer?  I find it easier to see the lock taken and released when it's closer.

```Go
// A new update is available, so pop the new chain update from
// the front of the update queue.
m.chainMtx.Lock()
update = m.chainUpdates[0]
m.chainUpdates[0] = nil // Set to nil to prevent GC leak.
m.chainUpdates = m.chainUpdates[1:]
m.chainMtx.Unlock()
```

rpctest/memwallet.go
```diff
@@ -0,0 +1,517 @@
...
+	isCoinbase bool, undo *undoEntry) {
+
+	for i, output := range outputs {
+		pkScript := output.PkScript
+
+		// Scan all the addresses we currently control to see if the
+		// output is paying to us.
+		for keyIndex, addr := range m.addrs {
+			pkHash := addr.ScriptAddress()
+			if !bytes.Contains(pkScript, pkHash) {
```
davecgh: This should work fine since the harness is controlling all the created blocks and transactions, but I'm sure you know this isn't really reliable in general.

rpctest/memwallet.go
```diff
@@ -0,0 +1,517 @@
...
+	// selection shortly below.
+	var outputAmt btcutil.Amount
+	for _, output := range outputs {
+		outputAmt += btcutil.Amount(output.Value)
+		tx.AddTxOut(output)
+	}
+
+	// Attempt to fund the transaction with spendable utxos.
+	if err := m.fundTx(tx, outputAmt, btcutil.Amount(feeRate)); err != nil {
+		return nil, nil
```
davecgh: `nil, err`

rpctest/memwallet.go
```diff
@@ -0,0 +1,517 @@
...
+			return nil, err
+		}
+
+		txIn.SignatureScript = sigScript
+
+		spentOutputs = append(spentOutputs, outPoint)
+	}
+
+	for _, utxo := range spentOutputs {
+		delete(m.utxos, utxo)
```
davecgh: This seems like it can leak utxos in the event of failure of sending the tx (such as in the case of a test that intentionally crafts transactions that should fail).

Maybe they just be marked as unspendable and rely on the block connection code above to actual remove them?  The funding code would obviously not consider them if they're marked unspendable.

From there you could have some type of function to invalidate/unwind the pending transaction which would mark all of its spent utxos as unspent again.

That would allow the code to restore the pending transaction's spent txos to test double spend scenarios and the like.

rpctest/node.go
```diff
@@ -0,0 +1,290 @@
...
+		// --rpcconnect
+		args = append(args, fmt.Sprintf("--rpcconnect=%s", n.rpcConnect))
+	}
+	// --rpccert
+	args = append(args, fmt.Sprintf("--rpccert=%s", n.certFile))
+	// --rpckey
+	args = append(args, fmt.Sprintf("--rpckey=%s", n.keyFile))
+	// --txindex
+	args = append(args, "--txindex")
+	if n.dataDir != "" {
```
davecgh: Probably should add `--addrindex` too.  Even though nothing uses it yet, it will ensure the `searchrawtransactions` RPC can be tested.

rpctest/rpc_harness_test.go
```diff
@@ -0,0 +1,454 @@
...
+			os.Exit(1)
+		}
+	}
+
+	os.Exit(exitCode)
+}
+
+func TestHarness(t *testing.T) {
+	// We should have 1250 BTC of mature unspendable outputs.
+	expectedBalance := btcutil.Amount(1250 * btcutil.SatoshiPerBitcoin)
```
davecgh: I'd suggest `numMatureOutputs * 50` so this is a little less fragile if the constant is changed (obviously that will be wrong if `numMatureOutputs > mainHarness.ActiveNet.SubsidyReductionInterval`, but I don't foresee that happening).

rpctest/rpc_harness.go
```diff
@@ -0,0 +1,395 @@
...
+	miningAddr := fmt.Sprintf("--miningaddr=%s", wallet.coinbaseAddr)
+	extraArgs = append(extraArgs, miningAddr)
+
+	config, err := newConfig("rpctest", certFile, keyFile, extraArgs)
+	if err != nil {
+		return nil, err
+	}
+
+	// Generate p2p+rpc listening addresses.
+	p2p, rpc := generateListeningAddresses()
```
davecgh: Avoid the locals?

```Go
config.listen, config.rpcListen = generateListeningAddresses()
```

rpctest/rpc_harness.go
```diff
@@ -0,0 +1,395 @@
...
+	}
+
+	// If a handler for the OnBlockConnected/OnBlockDisconnected callback
+	// has already been set, then we create a wrapper callback which
+	// executes both the currently registered callback, and the mem
+	// wallet's callback.
+	if handlers.OnBlockConnected != nil {
+		obc := handlers.OnBlockConnected
+		handlers.OnBlockConnected = func(hash *chainhash.Hash, height int32, t time.Time) {
+			obc(hash, height, t)
```
davecgh: Wouldn't the callback expect the wallet to already be updated at this point?  In other words, I think `wallet.ingestBlock` should be called before the user callback.

rpctest/rpc_harness.go
```diff
@@ -0,0 +1,395 @@
...
+		}
+	} else {
+		// Otherwise, we can claim the callback ourselves.
+		handlers.OnBlockConnected = wallet.ingestBlock
+	}
+	if handlers.OnBlockDisconnected != nil {
+		obd := handlers.OnBlockConnected
+		handlers.OnBlockDisconnected = func(hash *chainhash.Hash, height int32, t time.Time) {
+			obd(hash, height, t)
+			wallet.unwindBlock(hash, height, t)
```
davecgh: Same here.

rpctest/rpc_harness.go
```diff
@@ -0,0 +1,395 @@
...
+	config.rpcListen = rpc
+
+	// Create the testing node bounded to the simnet.
+	node, err := newNode(config, nodeTestData)
+	if err != nil {
+		return nil, err
+	}
+
+	nodeNum := numTestInstances
+	numTestInstances++
```
davecgh: The globals should be protected by a mutex since ultimately there will likely be multiple goroutines running the tests with more than instance and thus concurrent calls to `New`.

rpctest/rpc_harness.go
```diff
@@ -0,0 +1,395 @@
...
+
+	return h, nil
+}
+
+// SetUp initializes the rpc test state. Initialization includes: starting up a
+// simnet node, creating a websockets client and connecting to the started
+// node, and finally: optionally generating and submitting a testchain with a
+// configurable number of mature coinbase outputs coinbase outputs.
+func (h *Harness) SetUp(createTestChain bool, numMatureOutputs uint32) error {
+	var err error
```
davecgh: This can be removed.  Everything else below beside the first two assign via `:=`.

rpctest/memwallet.go
```diff
@@ -0,0 +1,517 @@
...
+		hdRoot:            hdRoot,
+		addrs:             addrs,
+		utxos:             make(map[wire.OutPoint]*utxo),
+		chainUpdateSignal: make(chan struct{}),
+		reorgJournal:      make(map[int32]*undoEntry),
+	}, nil
+}
+
+// start launches all goroutines required for the wallet to function properly.
+func (m *memWallet) start() {
```
davecgh: I'd recommend "exporting" all of the functions on the type that are intended to be called external to the type.  The type itself is not exported, so callers outside of the package won't be call them, but exporting them makes it more clear which functions are intended to be called within the package.

However, `chainSyncer`, for example, is never meant to be called by anything except the `memWallet` type itself, so it would remain unexported.

The described style is what is typically used throughout the codebase (there is probably some legacy code that doesn't follow it, but that is the goal).

rpctest/rpc_harness.go
```diff
@@ -0,0 +1,395 @@
...
+	for i := 0; i < h.maxConnRetries; i++ {
+		if client, err = btcrpcclient.New(&rpcConf, h.handlers); err != nil {
+			time.Sleep(time.Duration(i) * 50 * time.Millisecond)
+			continue
+		}
+		break
+	}
+
+	if client == nil {
+		return fmt.Errorf("connection timedout")
```
davecgh: timeout

rpctest/rpc_harness.go
```diff
@@ -0,0 +1,395 @@
...
+	if err = h.node.start(); err != nil {
+		return err
+	}
+	if err := h.connectRPCClient(); err != nil {
+		return err
+	}
+
+	h.wallet.start()
+
+	// Ensure the wallet properly dispatches our registered call-back for
```
davecgh: s/the wallet/btcd/

rpctest/rpc_harness.go
```diff
@@ -0,0 +1,395 @@
...
+		}
+	}
+
+	// Block until the wallet has fully synced up to the tip of the main
+	// chain.
+	_, height, err := h.Node.GetBestBlock()
+	if err != nil {
+		return err
+	}
+	ticker := time.NewTicker(time.Millisecond * 100)
```
davecgh: Since this is only for test code, I suppose this is good enough, but ideally there should be a way to get a channel out of the wallet that is signalled when `ingestWallet` gets done.  Then this code could also have a timeout.

```Go
select {
case <-channelFromWallet: /* ... */
case <-time.After(time.Minute): /* error and stuff */
}

File: rpctest/rpc_harness.go [hunk 35]
```diff
@@ -0,0 +1,395 @@
...
+				break out
+			}
+		}
+	}
+
+	return nil
+}
+
+// TearDown stops the running rpc test instance. All created processes are
+// killed, and temporary directories removed.
```
davecgh: Should it be made clear that `SetUp` and `TearDown` have to be called from the same goroutine since they aren't concurrent safe?  That is almost always going to be the case anyways, but since this is intended to be usable by others as well, that's important information.

davecgh: After looking further it looks like the entire harness really needs to be run in the same goroutine, so probably should just call out in `New` that while you can call New concurrently, the returned harness is not concurrent safe.

rpctest/rpc_harness.go
```diff
@@ -0,0 +1,395 @@
...
+//
+// This function is safe for concurrent access.
+func (h *Harness) ConfirmedBalance() btcutil.Amount {
+	return h.wallet.ConfirmedBalance()
+}
+
+// SendOutputs creates, signs, and finally broadcasts a transaction spending
+// the harness' available mature coinbase outputs creating new outputs
+// according to targetOutputs.
+func (h *Harness) SendOutputs(targetOutputs []*wire.TxOut,
```
davecgh:
```Go
//
```

rpctest/rpc_harness.go
```diff
@@ -0,0 +1,395 @@
...
+// wallet.
+//
+// This function is safe for concurrent access.
+func (h *Harness) ConfirmedBalance() btcutil.Amount {
+	return h.wallet.ConfirmedBalance()
+}
+
+// SendOutputs creates, signs, and finally broadcasts a transaction spending
+// the harness' available mature coinbase outputs creating new outputs
+// according to targetOutputs.
```
davecgh:
```Go
//
```

rpctest/rpc_harness.go
```diff
@@ -0,0 +1,395 @@
...
+// GenerateAndSubmitBlock creates a block whose contents include the passed
+// transactions and submits it to the running simnet node. For generating
+// blocks with only a coinbase tx, callers can simply pass nil instead of
+// transactions to be mined. Additionally, a custom block version can be set by
+// the caller. A blockVersion of -1 indicates that the current default block
+// version should be used.
+//
+// This function is safe for concurrent access.
+func (h *Harness) GenerateAndSubmitBlock(txns []*btcutil.Tx, blockVersion int32,
+	blockTime *time.Time) (*btcutil.Block, error) {
```
davecgh: `time.Time` [should not use a pointer per the doco](https://golang.org/pkg/time/#Time).

rpctest/blockgen.go
```diff
@@ -0,0 +1,175 @@
...
+	tx.AddTxOut(&wire.TxOut{
+		Value:    blockchain.CalcBlockSubsidy(nextBlockHeight, net),
+		PkScript: pkScript,
+	})
+	return btcutil.NewTx(tx), nil
+}
+
+// createBlock creates a new block building from the previous block.
+func createBlock(prevBlock *btcutil.Block, inclusionTxs []*btcutil.Tx,
+	blockVersion int32, blockTime *time.Time,
```
davecgh: `time.Time` [should not use a pointer per the doco](https://golang.org/pkg/time/#Time).

rpctest/rpc_harness_test.go
```diff
@@ -0,0 +1,454 @@
...
+		t.Fatalf("unable to generate single block: %v", err)
+	}
+
+	block, err := r.Node.GetBlock(blockHashes[0])
+	if err != nil {
+		t.Fatalf("unable to get block: %v", err)
+	}
+
+	minedTx := block.Transactions()[1]
+	txSha := minedTx.Hash()
```
davecgh: Please use `Hash` instead of `Sha` throughout.  A recent commit scrubbed the code case of them and I'd like to avoid reintroducing it.

rpctest/rpc_harness_test.go
```diff
@@ -0,0 +1,454 @@
...
+	if err != nil {
+		t.Fatalf("unable to generate single block: %v", err)
+	}
+
+	block, err := r.Node.GetBlock(blockHashes[0])
+	if err != nil {
+		t.Fatalf("unable to get block: %v", err)
+	}
+
+	minedTx := block.Transactions()[1]
```
davecgh: This will panic if the test didn't work as expected.  Should be a `if len(block.Transactions) < 2` { t.Fatalf(...) }`.

rpctest/rpc_harness_test.go
```diff
@@ -0,0 +1,454 @@
...
+	txSha := minedTx.Hash()
+	if !bytes.Equal(txid[:], txSha[:]) {
```
davecgh: The hash is an array so there is no need to `bytes.Equal`.  Just do `if *expectedTxHash == *txHash`.

rpctest/rpc_harness_test.go
```diff
@@ -0,0 +1,454 @@
...
+		t.Fatalf("txid's don't match, %v vs %v", txSha, txid)
+	}
+
+	// Next, generate a spend much greater than the block reward.
+	genSpend(btcutil.Amount(500 * btcutil.SatoshiPerBitcoin))
+}
```
davecgh: Looks like the results are never checked here.

rpctest/rpc_harness_test.go
```diff
@@ -0,0 +1,454 @@
...
+	}
+}
+
+func testConnectNode(r *Harness, t *testing.T) {
+	// Create a fresh test harnesses.
+	harness, err := New(&chaincfg.SimNetParams, nil, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := harness.SetUp(true, 25); err != nil {
```
davecgh: Can save some time by avoiding creating the 125 blocks since they aren't used.

rpctest/rpc_harness.go
```diff
@@ -0,0 +1,395 @@
...
+
+	// defaultRPCPort is the initial rpc port which will be used by the
+	// first created rpc harnesses to listen on for incoming rpc
+	// connections. Subsequent allocated ports for future rpc harness
+	// instances will be monotonically increasing even numbers calculated
+	// as such: defaultP2pPort + (2 * harness.nodeNum).
+	defaultRPCPort = 18556
+
+	// testInstances is a private package-level slice used to keep track of
+	// allvactive test harnesses. This global can be used to perform
```
davecgh: s/allvactive/all active/

rpctest/rpc_harness.go
```diff
@@ -0,0 +1,395 @@
...
+		rpc = net.JoinHostPort(localhost,
+			strconv.Itoa(defaultRPCPort+(2*numTestInstances)))
+	}
+
+	return p2p, rpc
+}
+
+func init() {
+	// Create the testInstances map once the package has been imported.
+	testInstances = make(map[string]*Harness)
```
davecgh: Since it's a `var`, you can you just create it in the declaration.

rpctest/rpc_harness_test.go
```diff
@@ -0,0 +1,454 @@
...
+	}
+
+	// The global testInstances map should now be fully purged with no
+	// active test harnesses remaining.
+	activeHarnesses := ActiveHarnesses()
+	if len(activeHarnesses) != 0 {
+		t.Fatalf("test harnesses still active after TestTearDownAll()")
+	}
+
+	for _, harness := range activeHarnesses {
```
davecgh: This was just checked for `!= 0` above and the test fails if not, so this will not do anything as there is nothing to range.  I think you meant to get a local copy before calling `TearDownAll` above and use that.

rpctest/rpc_harness_test.go
```diff
@@ -0,0 +1,454 @@
...
+	// This select case should fall through to the default as the goroutine
+	// should be blocked on the JoinNodes calls.
+	select {
+	case <-poolsSynced:
+		t.Fatalf("mempools detected as synced yet harness1 has a new tx")
+	default:
+	}
+
+	// Establish an outbound connection from harness1 to harness2. After
+	// the initial handshake both node should exchange inventory resulting
```
davecgh: both nodes

rpctest/rpc_harness_test.go
```diff
@@ -0,0 +1,454 @@
...
+	// in a synced mempool.
+	if err := ConnectNode(r, harness); err != nil {
+		t.Fatalf("unable to connect harnesses: %v", err)
+	}
+
+	// Select once again with a special timeout case after 1 minute. The
+	// goroutine above should now be blocked on sending into the unbuffered
+	// channel. The send should immediately suceed. In order to avoid the
```
davecgh: succeed

rpctest/rpc_harness_test.go
```diff
@@ -0,0 +1,454 @@
...
+	select {
+	case <-poolsSynced:
+		// fall through
+	case <-time.After(time.Minute):
+		t.Fatalf("block heights never detected as synced")
+	}
+
+}
+
+// TODO(roasbeef): revisit...
```
davecgh: Reminder

rpctest/rpc_harness_test.go
```diff
@@ -0,0 +1,454 @@
...
+	blockVersion := block.MsgBlock().Header.Version
+	if blockVersion != wire.BlockVersion {
+		t.Fatalf("block version is not default: expected %v, got %v",
+			wire.BlockVersion, blockVersion)
+	}
+
+	// Next generate a block with a "non-standard" block version along with
+	// time stamp a minute after the previous block's timestamp.
+	timestamp := block.MsgBlock().Header.Timestamp.Add(time.Minute)
+	targetBlockVersion := int32(1337)
```
davecgh: With versionbits this could eventually become something standard.  How about 0?

rpctest/rpc_harness_test.go
```diff
@@ -0,0 +1,454 @@
...
+	if err != nil {
+		t.Fatalf("unable to execute getinfo on node: %v", err)
+	}
+	expectedChainHeight := numMatureOutputs + uint32(mainHarness.ActiveNet.CoinbaseMaturity)
+	if uint32(nodeInfo.Blocks) != expectedChainHeight {
+		t.Errorf("Chain height is %v, should be %v",
+			nodeInfo.Blocks, expectedChainHeight)
+	}
+
+	for _, testCase := range harnessTestCases {
```
davecgh: I'd just call them all directly here.  The slice isn't used anywhere else.

rpctest/utils.go
```diff
@@ -0,0 +1,170 @@
...
+			}
+		}
+
+		blocksMatch = true
+	}
+
+	return nil
+}
+
+// ConnectNode establishes a new peer-to-peer connection from nodeA to nodeB.
```
davecgh: Looks like the variables were renamed from `nodeA` and `nodeB` and the comment wasn't updated.

rpctest/utils.go
```diff
@@ -0,0 +1,170 @@
...
+func TearDownAll() error {
+	harnessStateMtx.Lock()
+	defer harnessStateMtx.Unlock()
+
+	for _, harness := range testInstances {
+		if err := harness.TearDown(); err != nil {
+			return err
+		}
+
+		delete(testInstances, harness.testNodeDir)
```
davecgh: This shouldn't be here.  It's deleted in `harness.TearDown`.

davecgh:
Alright, I'm done with the review.  I'm really impressed overall!  This is extremely well done as far as general ease of use, flexibility, and overall code quality.  This is going to greatly increase the ability to test not only the RPC interface, but also various multi-node policies and other integration style tests.

I especially like the `Harness` framework and the utility functions that it provides as well as the `rpctest` utility functions for connecting together multiple harnesses (`ConnectNode`) and waiting until they are synced (blocks or mempool) via `JoinNodes`.  It might be useful to make the `JoinType` type bit flags so the caller can wait on both if desired, but it's also not that difficult for them to just call `JoinNodes` twice with each type currently.

I also really like that the `rpctest` package is usable externally so other services can make use of it to perform integration testing on their product(s).

rpctest/utils.go
Roasbeef: Very true, this line is redundant. I've removed it in a new commit.

rpctest/utils.go
Roasbeef: Correct. I've updated the comments.

rpctest/rpc_harness_test.go
Roasbeef: Ehh, but it keeps this function compact rather than growing linearly each time a new test case is added. I prefer to retain the usage of the slice.

rpctest/rpc_harness_test.go
Roasbeef: That's a stale TODO left over from a `git stash pop`. Removed.

rpctest/rpc_harness_test.go
Roasbeef: Fixed.

rpctest/rpc_harness_test.go
Roasbeef: Fixed.

rpctest/rpc_harness_test.go
Roasbeef: Nice catch. Updated to grab a local copy before the call to `TearDownAll`.

rpctest/rpc_harness.go
Roasbeef: Very true, also saves a few lines to boot. Updated to create it in-line with the declaration.

rpctest/rpc_harness.go
Roasbeef: Fixed.

rpctest/rpc_harness_test.go
Roasbeef: Good point, changed the test to not create any blocks at all.

rpctest/rpc_harness_test.go
Roasbeef: Hmm I must've moved on before completing this test. I've updated the test to also check the results as done above.

rpctest/rpc_harness_test.go
Roasbeef: Fixed.

rpctest/rpc_harness_test.go
Roasbeef: Excellent catch. I've updated the test to avoid a possible panic scenario.

rpctest/rpc_harness_test.go
Roasbeef: Sure, I've updated this code an any other instances of `Sha` instead of `Hash` throughout this PR.

rpctest/blockgen.go
Roasbeef: Fixed.

rpctest/rpc_harness.go
Roasbeef: Ah, I wasn't aware of the `IsZero` method on `time.Time`. 

I've updated the related code and modified the documentation to reflect this change.

rpctest/rpc_harness.go
Roasbeef: Added.

rpctest/rpc_harness.go
Roasbeef: Yeah I think it should. I've updated the documentation for `SetUp` and `TearDown` to include this information.

rpctest/rpc_harness.go
Roasbeef: Fixed.

rpctest/rpc_harness.go
Roasbeef: Fixed.

rpctest/memwallet.go
Roasbeef: Agreed, I think that's a great style to adhere to in general, and I typically do so within my own projects outside of btcd. 

I've updated all the functions meant for "public" consumption within the `memWallet` to be capitalized.

rpctest/rpc_harness.go
Roasbeef: Removed.

rpctest/rpc_harness.go
Roasbeef: Fixed.

rpctest/rpc_harness.go
Roasbeef: Good observation, I've swapped the execution order of the callbacks.

rpctest/rpc_harness.go
Roasbeef: Done.

rpctest/rpc_harness_test.go
Roasbeef: Great suggestion, I've updated the line to base the `expectedBalance` off of `numMatureOutputs` instead.

rpctest/node.go
Roasbeef: Sure, I've added the `--addrindex` flag also.

rpctest/memwallet.go
Roasbeef: Good point. 

I've modified the `utxoStruct` to add a `isLocked` field which denotes that the output has recently been used to construct a fully signed transaction, yet not broadcasted. A new method has been added to "unlock" any previously locked outputs due to funding a transaction. The new function is currently called `FreeLockedOutputs`, I'm open to any suggestion w.r.t to the name or its arguments 😄 

I've added a new test under `rpc_harness_test.go` which exercises the new behavior. Additionally the documentation for `CreateTransaction` has been updated to reflect the new behavior.

rpctest/memwallet.go
Roasbeef: Ooohh, nice catch!

rpctest/memwallet.go
Roasbeef: Sure. I always go back and forth myself as to which style I prefer w.r.t spacing around critical sections.

rpctest/memwallet.go
Roasbeef: Hmm yeah initially I considered passing in an instance of `testing.T` would which fail the test if an error was detected. But then I realized the package may not always be used in conjunction with a `testing.T`. 

I ended up just silently failing here, since erroneous behavior should lead to a failure in the test suite as you pointed out.

rpctest/memwallet.go
Roasbeef: Yeah that's how I originally had it in an earlier version, but then flip flopped on which path to create a `p2pkh` address I preferred. 

I've updated the function to return the `AddressPubKeyHash` address via an instance of a `AddressPubKey` address, similar to your code fragment.

rpctest/memwallet.go
Roasbeef: Fixed.

rpctest/blockgen.go
Roasbeef: Fixed.

rpcserver_test.go
Roasbeef: Fixed.

rpctest/README.md
Roasbeef: Fixed.

rpctest/README.md
Roasbeef: Sure, I've updated this section. Let me know if it's still mismatched w.r.t to the rest of the project.

rpctest/README.md
Roasbeef: Fixed.

Roasbeef:
Thanks! I'll be very happy to see this PR finally merged as it's over a year old at this point! 

> I also really like that the rpctest package is usable externally so other services and projects can make use of it to perform integration testing on their product(s).

Agreed. I've been using this package heavily within [`lnd`](https://github.com/LightningNetwork/lnd) over the past few months to test various behaviors within the daemon.

rpctest/rpc_harness_test.go
davecgh: Fair enough.  It's either here or the slice that grows linearly each time a new test cases is added.  Pretty much the same to me.  Since you prefer the slice, let's leave it.

rpctest/README.md
```diff
@@ -4,13 +4,23 @@ rpctest
...
+
+This package was designed specifically to act as an RPC testing harness for
+`btcd`. However, the constructs presented are general enough to be adapted to
+any project wishing to programmatically drive a `btcd` instance of its
+systems/integration tests. 
 
 ## Documentation
 
-[![GoDoc](https://godoc.org/github.com/btcsuite/btcd/rpctest?status.png)]
+[![ISC License]
```
davecgh: Please see the [blockchain README.md](https://github.com/btcsuite/btcd/blob/763f731c5c97ab2f659f0c7dd3e6a8f48c723daa/blockchain/README.md) for example.  Notice how the entire `## Documentation` section is gone and these are at the top along with the build status.

rpctest/blockgen.go
```diff
@@ -125,8 +125,8 @@ func createBlock(prevBlock *btcutil.Block, inclusionTxs []*btcutil.Tx,
 	// it's the genesis block in which case use the current time.
 	var ts time.Time
 	switch {
-	case blockTime != nil:
-		ts = *blockTime
+	case !blockTime.IsZero():
+		ts = blockTime
 	case blockHeight == 0:
```
davecgh: Shouldn't this be impossible?  You always have to build off at least the genesis block which is at height and given `blockHeight` is `prevBlock.Height() + 1`, that means it must always be at least 1.

rpctest/memwallet.go
```diff
@@ -481,16 +479,38 @@ func (m *memWallet) CreateTransaction(outputs []*wire.TxOut, feeRate btcutil.Amo
 
 		txIn.SignatureScript = sigScript
 
-		spentOutputs = append(spentOutputs, outPoint)
+		spentOutputs = append(spentOutputs, utxo)
 	}
 
+	// As these outputs are now being spent by this newly created
+	// transaction, mark the outputs are "locked". This action ensures
+	// these outputs won't be doubled spent by any subsequent transactions.
```
davecgh: s/doubled/double/

rpctest/memwallet.go
```diff
@@ -481,16 +479,38 @@ func (m *memWallet) CreateTransaction(outputs []*wire.TxOut, feeRate btcutil.Amo
...
 	}
 
 	return tx, nil
 }
 
+// FreeLockedOutputs unlocks any outputs which were previously locked due to
+// being selected to fund a transaction via the CreateTransaction method.
+//
+// This function is safe for concurrent access.
+func (m *memWallet) FreeLockedOutputs(inputs []*wire.TxIn) {
```
davecgh: Bikeshedding on the name, but how about `UnlockOutputs` ?

rpctest/rpc_harness.go
```diff
@@ -317,6 +326,15 @@ func (h *Harness) CreateTransaction(targetOutputs []*wire.TxOut,
 	return h.wallet.CreateTransaction(targetOutputs, feeRate)
 }
 
+// FreeLockedOutputs unlocks any outputs which were previously marked as
+// unspendabe due to being selected to fund a transaction via the
+// CreateTransaction method.
+//
+// This function is safe for concurrent access.
+func (h *Harness) FreeLockedOutputs(inputs []*wire.TxIn) {
```
davecgh: Same deal here if the name of the wallet method is changed to `UnlockOutputs`.  And the docs above then could say `are locked in order to avoid potential ... MUST be unlocked via a call to UnlockOutputs ...`

rpctest/memwallet.go
Roasbeef: Fixed.

rpctest/rpc_harness.go
Roasbeef: Done.

rpctest/memwallet.go
Roasbeef: Done. Performed a search-and-replace throughout the PR.

rpctest/README.md
Roasbeef: Modified to match the rest of the project.

rpctest/blockgen.go
Roasbeef: Hmm, yeah you're correct. Seems this was left over from the original gist
this was pulled from. I've removed this case entirely.

rpctest/README.md
```diff
@@ -16,28 +19,15 @@ This package was designed specifically to act as an RPC testing harness for
...
 
 ```bash
 $ go get -u github.com/btcsuite/btcd/rpctest
 ```
 
 ## License
-Package rpctest is licensed under the [copyfree](http://copyfree.org) ISC
+
+
+Package blockchain is licensed under the [copyfree](http://copyfree.org) ISC
```
davecgh: `blockchain` -> `rpctest`

Roasbeef: Left over from my copy-pasta :)

Fixed.

davecgh:
Everything looks good.  Please squash it all down and I'll merge it!  Thanks for all your on this.

Roasbeef:
Thanks! This package will be an excellent addition to the package. Happy to finally see it completed and merged. 

Squashed waaay down 👍

davecgh:
OK

PR #375, opened 2015-04-09
btcec: Added ECDH and encryption/decryption support

ishbir:
Added ECDH and encryption/decryption support based on RFC 4753, RFC 5903 and ANSI X9.63 section 5.8.1.

The format of encrypted data matches what Pyelliptic expects. It's to ensure that the exchange of data among between existing systems (PyBitmessage and any Go implementation) is possible.


davecgh:
Thanks for the PR.  I'll probably not be able to give this a proper review until early next week.

btcec/README.md
```diff
@@ -47,6 +47,16 @@ $ go get github.com/btcsuite/btcd/btcec
   Demonstrates verifying a secp256k1 signature against a public key that is
   first parsed from raw bytes.  The signature is also parsed from raw bytes.
 
+* [Encryption]
+  (http://godoc.org/github.com/monetas/bmd/bmec#example-package--EncryptMessage)
```
davecgh: `btcsuite/btcec`

btcec/README.md
```diff
@@ -47,6 +47,16 @@ $ go get github.com/btcsuite/btcd/btcec
...
+  Demonstrates encrypting a message for a public key that is first parsed from
+  raw bytes, then decrypting it using the corresponding private key.
+
+* [Decryption]
+  (http://godoc.org/github.com/monetas/bmd/bmec#example-package--DecryptMessage)
```
davecgh: `btcsuite/btcec`

btcec/ciphering.go
```diff
@@ -0,0 +1,192 @@
+package btcec
```
davecgh: Missing the license header

btcec/ciphering_test.go
```diff
@@ -0,0 +1,172 @@
+package btcec_test
```
davecgh: Missing the license header

davecgh:
Test failure:
```
--- FAIL: Example_encryptMessage (0.01s)
got: pubkey isn't on secp256k1 curve
want: test message
```

btcec/ciphering.go
```diff
@@ -0,0 +1,196 @@
+// Copyright (c) 2013-2014 Conformal Systems LLC.
```
davecgh: New file.  2015.

btcec/ciphering_test.go
```diff
@@ -0,0 +1,176 @@
+// Copyright (c) 2013-2014 Conformal Systems LLC.
```
davecgh: 2015 here as well.

ishbir:
I saw the failure but was unable to replicate it. Try rebuilding via Travis?

davecgh:
I've run the tests several times locally and not seeing it either.  The rebuild on Travis worked, but given that it's a constant value and there are no random numbers involved, it's quite worrying.  I'm going to setup a test loop on this locally and run it ~10,000 times just to make sure I can't reproduce it.  I suspect it was a hardware issue on the Travis build machine.

davecgh:
I've hit the failure a few time locally, so it's definitely a bug.  Here is a script you can run as well:

First generate the test binary with:
```sh
$ go test -c -o btcectest
```

Then create and run the following script to continuously run the specific test until it fails:

```sh
$ cat test_loop.sh
#!/bin/sh

for i in {0..10000}; do
        echo "Test iteration $i"
        btcectest -test.run=Example_encryptMessage
        if [ $? != 0 ]; then
                echo "Test $i failed!"
                exit
        fi
done
```

In addition, I added a prefix to each of the `fmt.Printn` in the example and can confirm the error is coming out of the call to `plaintext, err := btcec.Decrypt(privKey, ciphertext)`.

davecgh:
Another point of feedback is please remove the named returns.  They are not allowed in any new code as they have historically led to subtle bugs and return garbage values versus the zero value for the other parameters.

davecgh:
As an additional data point for you, I had it print the pubkey when the error case is hit:

```
pubkey: 04fd3ddd327638538b25ba8e1b6efec2c5276fefa7ec9a6d097afa52d040939137cabe029880e4333ab5a8a7c5c4892e111f5d0a877e9f113bfbd7788fbcd52f00
```
Calculating the left and right sides of the equation using the x, and y values in that pubkey:
```
x = fd3ddd327638538b25ba8e1b6efec2c5276fefa7ec9a6d097afa52d040939137
y = cabe029880e4333ab5a8a7c5c4892e111f5d0a877e9f113bfbd7788fbcd52f00
x^3 + 7 (mod p)= a2ac4e67855a6907d6aa6009e8e3f6058fcdef94ab5aa48a77ea84b393860d47
y^2 (mod p) = 4e67855a6907d6aa6009e8e3f6058fcdef94ab5aa48a77ea84b436320fb3d26c
```
That indeed is not a valid point on the secp256k1 curve.

ishbir:
Thank you for the feedback. I will take a look at the issue later today.

ishbir:
Encryption isn't deterministic as it generates an ephemeral private key before generating shared secret and then uses that for encryption using AES-256-CBC. The error was that I had naively assumed that length of X.Bytes() and Y.Bytes() would always be 32. I changed the code, ran the test 10,000x and it works every time now. Other issues that you pointed out have also been fixed.

Thanks for the assistance. Appreciate it. :)

davecgh:
Great.  I'm glad you were able to track it down.  As I stated before, I won't be able to give this a proper review (verification of all of the math and algorithms) until next week.

ishbir:
Any updates?

davecgh:
Sorry for the delay.  I haven't forgot about it, but since it's non-trivial review, I haven't got to it yet.

davecgh:
As an update, I'm planning to review this over the weekend.

ishbir:
Okay.

btcec/ciphering.go
davecgh: We've recently relicensed everything "The btcsuite authors".  Please update the license accordingly.

btcec/ciphering_test.go
davecgh: We've recently relicensed everything "The btcsuite authors".  Please update the license accordingly.

btcec/ciphering.go
```diff
@@ -0,0 +1,194 @@
...
+	paddedIn := addPKCSPadding(in)
+	out := make([]byte, aes.BlockSize+70+len(paddedIn)+sha256.Size)
+	iv := out[:aes.BlockSize]
+	if _, err = io.ReadFull(rand.Reader, iv); err != nil {
+		return nil, err
+	}
+	// start writing public key
+	pb := ephemeral.PubKey().SerializeUncompressed()
+	// curve and X length: 0x02CA = 714, 0x20 = 32
+	copy(out[aes.BlockSize:aes.BlockSize+4], []byte{0x02, 0xCA, 0x00, 0x20})
```
davecgh: It would probably make since to define this as a package-level constant global (have to use var) rather than having to create this each time.

```Go
var whateverName = [4]byte{0x02, 0xCA, 0x00, 0x20}
...
	copy(out[aes.BlockSize:aes.BlockSize+4], whateverName[:])
...
```

Also, I'd rather this section using a rolling offset variable as opposed to ever increasing addition in the slice offsets.  Something like the following:

```Go
offset := aes.BlockSize
copy(out[offset:offset+4], whateverName[:])
offset += 4

// X
copy(out[offset:offset+32], pb[1:33])
offset += 32

// Y length.  32 bytes.
copy(out[offset:offset+2], whateverName[2:4])
offset += 2

// Y
copy(out[offset:offset+32], pb[33:])
```

btcec/ciphering.go
```diff
@@ -0,0 +1,194 @@
...
+func Decrypt(priv *PrivateKey, in []byte) ([]byte, error) {
+	if len(in) < aes.BlockSize+70+aes.BlockSize+sha256.Size {
+		return nil, errInputTooShort
+	}
+
+	// read iv
+	iv := in[:aes.BlockSize]
+
+	// start reading pubkey
+	if binary.BigEndian.Uint16(in[aes.BlockSize:aes.BlockSize+2]) != 714 {
```
davecgh: Same deal in this section here.  Rolling offsets are easier to read and update without making mistakes.

ishbir:
Done.

ishbir:
Any updates @davecgh?

davecgh:
Please rebase it to the latest master.  I'm going to give it another review pass tonight, but I believe this is ready to be merged after rebase.

ishbir:
Done.

davecgh:
Alright.  Please update the license headers and I'll get this merged.  It's good to go after that.

ishbir:
Oh right. Sorry, didn't see the comment. Done.

davecgh:
Thanks.

OK

PR #377, opened 2015-04-12
Added tests for addrmgr.KnownAddress

DanielKrawisz:
I have made tests for knownaddresses.go. This meant making an internal_test.go file to expose chance() and isBad(). I have tested both functions in accordance with the comments given for each of them. Both functions turned out to be buggy and required fixes. 

A comment for chance() says "Very recent attempts are less likely to be retried." However, the if clause below only affects c if the last attempt was NOT very recent, or greater than 10 minutes. The way this function works now is that very recent attempts are MORE likely to be retried! 

The next comment for chance() says "Failed attempts deprioritise." However, the for loop says "for i := ka.attempts; i < 0; i++" However, ka.attempts is supposed to be zero or greater. It will not ordinarily be negative. This means that the loop never actually executes. 

Now, as for isBad(), the comment says "isBad returns true if the address in question has not been tried in the last minute and meets one of the following criteria:
2) It hasn't been seen in over a month"
However, the line of code that checks for this checks whether the address is LESS than a month old and then returns true if it is. This marks the address as bad for the opposite reason it should! 

The tests I made cover all branches of chance() and isBad().

### DIFF
--- a/addrmgr/internal_test.go
+++ b/addrmgr/internal_test.go
@@ -1,0 +1,25 @@
+// Copyright (c) 2013-2015 Conformal Systems LLC.
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package addrmgr
+
+import (
+	"time"
+
+	"github.com/btcsuite/btcd/wire"
+)
+
+func TstKnownAddressIsBad(ka *KnownAddress) bool {
+	return ka.isBad()
+}
+
+func TstKnownAddressChance(ka *KnownAddress) float64 {
+	return ka.chance()
+}
+
+func TstNewKnownAddress(na *wire.NetAddress, attempts int,
+	lastattempt, lastsuccess time.Time, tried bool, refs int) *KnownAddress {
+	return &KnownAddress{na: na, attempts: attempts, lastattempt: lastattempt,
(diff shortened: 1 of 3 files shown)


addrmgr/internal_test.go
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2013-2014 Conformal Systems LLC.
```
davecgh: New file.  2015

davecgh:
@DanielKrawisz.  Nice catches.  Please rebase this on top of #376 (after you rebase that one on top of master, so I can do fast-forward merges.

DanielKrawisz:
Ok, I'm pretty sure I did what you want! I'm still learning the finer points of git.

davecgh:
@DanielKrawisz:  Not sure if you noticed, but Travis has some test failures after the rebase.

davecgh:
@DanielKrawisz Any updates on this?

DanielKrawisz:
Sorry, I didn't notice your message! I'll try again.

DanielKrawisz:
There, I didn't notice that the chance function had been redefined. The tests pass now.

davecgh:
Alright thanks.  Please rebase over master and squash it to a single commit.

DanielKrawisz:
Ok, there!

PR #380, opened 2015-04-16
database: Major redesign of database package.

### LINKED ISSUE #199: power outage and corrupted db
My database is on an external drive which was (accidentally) suddenly unplugged while btcd was running. Now I get this error when making a particular `getrawtransaction` call. What can I do to repair the database?

```
08:24:51 2014-12-24 [WRN] BCDB: unable to decode tx block 335644 00000000000000001259f32bc7942a729547e66fad32525c3803bc83720d86e4 txoff 545 txlen 192
08:24:51 2014-12-24 [ERR] RPCS: Error fetching tx: unexpected EOF
```

The API returns `{'code': -5, 'message': 'No information available about transaction'}`.

### LINKED ISSUE #201: Database seems to extremely rarely miss transactions that were added
We had a report on IRC today where a user is missing a transaction in the database that should be there since the previous block was accepted without error.  In particular it is transaction  57e99c4ce5c36c8927759d1f90a472bc88d8d50300a170b4e2d38376674a38b7 from block 00000000000000001117831bc2f8eae21759f9e5e70b0857d79d58a1866add1 (height 334691).

It is most likely a case of database corruption, but I've asked him to post the database LOG for review.

### LINKED ISSUE #247: Very slow block processing
I'm running btcd in Ubuntu Linux and it is syncing really slow.

I've mentioned this in the chat channel and most people suggested it was due to IO limitations. However, I'm monitoring file IO now and, although it spikes every now and then, there are often very long periods of no file IO at all.

I've just launched btcd after several days of being offline and this is my log output:
https://gist.github.com/stevenroose/d4b88d6dc0e1e5425b34

You can see the current block height when launched: "Syncing to block height 339784".

I have processing rates of about 1-4 blocks a minute.

### LINKED ISSUE #255: [database] Support multiple (non-best) chains, concurrent reads on different forks.
<a href="https://github.com/jrick"></img></a> **Issue by [jrick](https://github.com/jrick)**
_Thursday Jul 17, 2014 at 04:34 GMT_
_Originally opened as https://github.com/btcsuite/btcdb/issues/6_

----

The database currently only saves the best chain.  If a fork extends beyond the current best chain, the reorged blocks will be removed from the database.  This makes the database unsafe for multiple callers who concurrently expect their given block range to be valid.

For example, when btcd must rescan some range of blocks, and the tip of the range is removed from the best chain due to a new best block,   db lookups will begin failing for missing the block for some hash, or worse (if only heights are used), lookups will silently continue with no indication that previously processed blocks may not be on the same fork.

A better solution would be to use an iterator which allows walking  through some range of blocks in a fork.  Even if the best chain is switched, these blocks can still be iterated over.  After the iteration is finished, the caller can finish (it returned results about the chain at the time the request was made) or a switch back to the main chain and/or best block can be made if more blocks were attached.

### LINKED ISSUE #256: [database] BlockShaMissing should include the missing block hash
<a href="https://github.com/jrick"></img></a> **Issue by [jrick](https://github.com/jrick)**
_Friday Aug 08, 2014 at 15:34 GMT_
_Originally opened as https://github.com/btcsuite/btcdb/issues/7_

----

The BlockShaMissing error is not very descriptive when printed, due to the missing block sha not being included in the error string.  To fix this, BlockShaMissing should be changed to a type rather than just a variable.  For example:

```Go
type BlockShaMissing btcwire.ShaHash

func (e *BlockShaMissing) Error() string {
        return "block " + (*btcwire.ShaHash)(e).String() + " missing"
}
```

### LINKED ISSUE #257: [database] Remove Db.DropAfterBlockBySha and Db.RollbackClose
<a href="https://github.com/jongillham"></img></a> **Issue by [jongillham](https://github.com/jongillham)**
_Friday Oct 24, 2014 at 10:57 GMT_
_Originally opened as https://github.com/btcsuite/btcdb/issues/9_

----

I would like to remove [Db.DropAfterBlockBySha](https://github.com/conformal/btcdb/blob/master/db.go#L40) and [Db.RollbackClose](https://github.com/conformal/btcdb/blob/master/db.go#L111) as it prohibits several NoSQL datastores from using btcd due to their lax transaction/atomicity guarantees and no concept of rollback.

I just wanted to start a discussion on how feasible this would be and what the implications are before I dive in and break everything.

davecgh:
[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/380)

This pull request contains a complete redesign and rewrite of the database package that approaches things in a vastly different manner than the previous version.  This is the first part of several stages that will be needed to ultimately make use of this new package.

Some of the reasons for this were discussed in #255, however a quick summary is as follows:

- The previous database could only contain blocks on the main chain and reorgs required deleting the blocks from the database.  This made it impossible to store orphans and could make external RPC calls for information about blocks during the middle of a reorg fail.
- The previous database interface forced a high level of bitcoin-specific intelligence such as spend tracking into each backend driver.
- The aforementioned point led to making it difficult to implement new backend drivers due to the need to repeat a lot of non-trivial logic which is better handled at a higher layer, such as the blockchain package.
- The old database stored all blocks in leveldb.  This made it extremely inefficient to do things such as lookup headers and individual transactions since the entire block had to be loaded from leveldb (which entails it doing data copies) to get access.
- The vast majority of database activity after the initial block download is read activity, however leveldb, as its name implies, is optimized for leveled write performance at the expense of read performance.

In order to address all of these concerns, and others not mentioned, the database interface has been redesigned as follows:

- Two main categories of functionality are provided: block storage and metadata storage
- All block storage and metadata storage are done via read-only and read-write MVCC transactions with both manual and managed modes
  - Support for multiple concurrent readers and a single writer
  - Readers use a snapshot and therefore are not blocked by the writer
- Some key properties of the block storage and retrieval API:
  - It is generic and does NOT contain additional bitcoin logic such spend tracking and block linking
  - Provides access to the raw serialized bytes so deserialization is not forced for callers that don't need it
  - Support for fetching headers via independent functions which allows implementations to provide significant optimizations
  - Ability to efficiently retrieve arbitrary regions of blocks (transactions, scripts, etc)
- A rich metadata storage API is provided:
  - Key/value with arbitrary data
  - Support for buckets and nested buckets
  - Bucket iteration through a couple of different mechanisms
  - Cursors for efficient and direct key seeking
- Supports registration of backend database implementations
- Comprehensive test coverage
- Provides strong documentation with example usage

This pull request also contains an implementation of the previously discussed interface named ffldb (flat file plus leveldb metadata backend).  Here is a quick overview:

- Highly optimized for read performance with consistent write performance regardless of database size
- All blocks are stored in flat files on the file system
- Bulk block region fetching is optimized to perform linear reads which improves performance on spindle disks
- Anti-corruption mechanisms:
  - Flat files contain full block checksums to quickly an easily detect database corruption without needing to do expensive merkle root calculations
  - Metadata checksums
  - Open reconciliation
- Extensive test coverage:
  - Comprehensive blackbox interface testing
  - Whitebox testing which uses intimate knowledge to exercise uncommon failure paths such as deleting files out from under the database
  - Corruption tests (replacing random data in the files)

In addition, there is a new tool provided under the new database directory named `dbtool` which provides a few basic commands for testing the database.  It is designed around commands, so it could be useful to expand on in the future.

Finally, this commit addresses the following issues:

- Adds support for and therefore closes #255
- Fixes #199
- Fixes #201
- Implements and closes #256
- Obsoletes and closes #257
- Closes #247 once the required chain and btcd modifications are in place
  to make use of this new code

There are several things that need to happen before this PR can be merged and ultimately used:
- [x] The blockchain package needs to be redesigned to pick up the functionality the new database is losing (spend tracking, chain linking, transaction index) (#491)
- [x] The blockchain package needs to be made safe for concurrency (#491)
- [x] btcd needs to be updated to use the redesigned blockchain and database packages (#491)
- [x] A database cache is needed to increase throughput


davecgh:
Rebased and updated for `btcutil.Block.Sha()` API change.

database2/ffboltdb/blockio.go
```diff
@@ -0,0 +1,746 @@
...
+	// openBlockFiles houses the open file handles for existing block files
+	// which have been opened read-only along with an individual RWMutex.
+	// This scheme allows multiple concurrent readers to the same file while
+	// preventing the file from closed out from under them.
+	//
+	// lruMutex protects concurrent access to the least recently used list
+	// and lookup map.
+	//
+	// openBlocksLRU tracks how the open files are refenced by pushing the
+	// least recently used files to end of the list.  When a file needs to
```
jrick: s/end/beginning/ ?

database2/ffboltdb/blockio.go
```diff
@@ -0,0 +1,746 @@
...
+	//
+	// None of the mutexes are required to be locked at the same time, and
+	// often aren't.  However, if they are to be locked simultaneously, they
+	// MUST be locked in the order previously specified.
+	//
+	// Due to the high performance and multi-read concurrency requirements,
+	// write locks should only be held for the minimum time necessary.
+	obfMutex         sync.RWMutex
+	lruMutex         sync.Mutex
+	openBlocksLRU    *list.List
```
jrick: Can you comment the type saved by the list?

database2/ffboltdb/blockio.go
davecgh: It's right.  Most recently used goes to the beginning, least recently used to the end.

database2/ffboltdb/blockio.go
davecgh: Yep.

database2/ffboltdb/blockio.go
davecgh: However I'll reword it in terms of the most recently used since I can see how that might be worded in a confusing fashion.

database2/ffboltdb/blockio.go
```diff
@@ -0,0 +1,746 @@
...
+	// The following fields are related to the flat files which hold the
+	// actual blocks.   The number of open files is limited by maxOpenFiles.
+	//
+	// obfMutex protects concurrent access to the openBlockFiles map.  It is
+	// a RWMutex so multiple readers can simultaneously access open files.
+	//
+	// openBlockFiles houses the open file handles for existing block files
+	// which have been opened read-only along with an individual RWMutex.
+	// This scheme allows multiple concurrent readers to the same file while
+	// preventing the file from closed out from under them.
```
davecgh: from closed -> from being closed

database2/ffboltdb/blockio.go
```diff
@@ -0,0 +1,746 @@
...
+	// Open the appropriate file as read-only.
+	filePath := blockFilePath(s.basePath, fileNum)
+	file, err := os.Open(filePath)
+	if err != nil {
+		return nil, makeDbErr(database.ErrDriverSpecific, err.Error(),
+			err)
+	}
+	blockFile := &lockableFile{file: file}
+
+	// Close the least recently used file is the file exceeds the max
```
jrick: s/is/if/ ?

database2/ffboltdb/blockio.go
```diff
@@ -0,0 +1,746 @@
...
+	// Bitcoin network.
+	origOffset := wc.curOffset
+	hasher := crc32.New(castagnoli)
+	var scratch [4]byte
+	byteOrder.PutUint32(scratch[:], uint32(s.network))
+	n, err := wc.curFile.file.WriteAt(scratch[:], int64(wc.curOffset))
+	wc.curOffset += uint32(n)
+	if err != nil {
+		str := fmt.Sprintf("failed to write network to file %d at "+
+			"offset %d: %v", wc.curFileNum, wc.curOffset, err)
```
jrick: If the current offset was incremented then the offset reported in the error message is not the offset where this data was being written to.  Subtract n here?

database2/ffboltdb/blockio.go
```diff
@@ -0,0 +1,746 @@
...
+		return blockLocation{}, makeDbErr(database.ErrDriverSpecific,
+			str, err)
+	}
+	_, _ = hasher.Write(scratch[:])
+
+	// Block length.
+	byteOrder.PutUint32(scratch[:], blockLen)
+	n, err = wc.curFile.file.WriteAt(scratch[:], int64(wc.curOffset))
+	wc.curOffset += uint32(n)
```
jrick: Since this pattern of writing to the file, and then incrementing the cursor offset, and then possibly returning the error is done in several places, would it make sense to abstract this out to a method of the writer cursor?

database2/ffboltdb/blockio.go
```diff
@@ -0,0 +1,746 @@
...
+
+	// The network associated with the block must match the current active
+	// network, otherwise somebody probably put the block files for the
+	// wrong network in the directory.
+	serializedNet := byteOrder.Uint32(serializedData[:4])
+	if serializedNet != uint32(s.network) {
+		str := fmt.Sprintf("block data for block %s is for the "+
+			"wrong network - got %d, want %d", hash, serializedNet,
+			uint32(s.network))
+		return nil, makeDbErr(database.ErrDriverSpecific, str, nil)
```
jrick: wrong error code?

davecgh: It's correct.  The generic database interface errors have no notion of separate block files.  Thus, it's a driver-specific error should the wrong block files be put into the directory (not as hard as you might think when sizes grow and people start symlinking things).

database2/ffboltdb/blockio.go
davecgh: Good catch.  I was originally incrementing the cursor offset after the error, but changed it for rollback purposes.

database2/ffboltdb/blockio.go
davecgh: It might be a little tricky with the mutexes.  Perhaps something like this?

```Go
// .... MUST be called with the write cursor current file lock held...
func (wc *writeCursor) Write(data []byte, fieldName string) error {
	n, err := wc.curFile.WriteAt(data, int64(wc.curOffset))
	wc.curOffset += uint32(n)
	if err != nil {
		str := fmt.Sprintf("failed to write %s to file %d at "+
			"offset %d: %v", fieldName, wc.curFileNum,
			wc.curOffset-n, err)
		return makeDbErr(database.ErrDriverSpecific, str, err)
	}
	return nil
}
```

Then all of these sections would just become:

```Go
if err := wc.Write(scratch[:], "network"); err != nil {
	return blockLocation{}, err
}
if err := wc.Write(scratch[:], "block length"); err != nil {
	return blockLocation{}, err
}
...
```

jrick: I didn't consider the mutexes but that seems nice enough.

database2/ffboltdb/blockio.go
```diff
@@ -0,0 +1,746 @@
...
+	serializedData := make([]byte, numBytes)
+	_, err = blockFile.file.ReadAt(serializedData, int64(readOffset))
+	if err != nil {
+		blockFile.RUnlock()
+		str := fmt.Sprintf("failed to read region from block file %d, "+
+			"offset %d, len %d: %v", loc.blockFileNum, readOffset,
+			numBytes, err)
+		return nil, makeDbErr(database.ErrDriverSpecific, str, err)
+	}
+	blockFile.RUnlock()
```
jrick: This is unlocked as the first statement both inside and after the error check.  Move it to before the check?

database2/ffboltdb/blockio.go
```diff
@@ -0,0 +1,746 @@
...
+
+	return serializedData, nil
+}
+
+// handleRollback rolls the block files on disk back to the provided file number
+// and offset.  This involves potentially deleting and truncating the files that
+// were partially written.
+//
+// There are effective two scenarios to consider here:
```
jrick: effectively

database2/ffboltdb/blockio.go
```diff
@@ -0,0 +1,745 @@
...
+		wc.curFile.Lock()
+		if wc.curFile.file != nil {
+			_ = wc.curFile.file.Close()
+			wc.curFile.file = nil
+		}
+		wc.curFile.Unlock()
+	}
+	for ; wc.curFileNum > oldBlockFileNum; wc.curFileNum-- {
+		if err := s.deleteFileFunc(wc.curFileNum); err != nil {
+			_ = log.Warnf("ROLLBACK: Failed to delete block file "+
```
jrick: Maybe we should use `Critical`/`Criticalf` for logging rollback errors.

database2/ffboltdb/blockio.go
```diff
@@ -0,0 +1,745 @@
...
+	}
+
+	// Sync the file to disk.
+	if err := wc.curFile.file.Sync(); err != nil {
+		wc.curFile.Unlock()
+		_ = log.Warnf("ROLLBACK: Failed to sync file %d: %v",
+			wc.curFileNum, err)
+		return
+	}
+	wc.curFile.Unlock()
```
jrick: Unlocks on both branches, can move before the check (need to split up the if statement).

database2/ffboltdb/blockio.go
```diff
@@ -0,0 +1,745 @@
...
+// to detect unexpected shutdowns in the middle of writes so the block files
+// can be reconciled.
+func scanBlockFiles(dbPath string) (int, uint32) {
+	lastFile := -1
+	fileLen := uint32(0)
+	for i := 0; ; i++ {
+		filePath := blockFilePath(dbPath, uint32(i))
+		st, err := os.Stat(filePath)
+		if err != nil {
+			break
```
jrick: might want to check for `os.IsNotExist` here.  Other errors might need to be returned.

database2/ffboltdb/blockio.go
davecgh: I decided to just define a helper function on the store just above it, but it's refactored in essentially the same way as discussed.

database2/ffboltdb/blockio.go
```diff
@@ -0,0 +1,750 @@
...
+
+		writeCursor: &writeCursor{
+			curFile:    &lockableFile{},
+			curFileNum: uint32(fileNum),
+			curOffset:  uint32(fileOff),
+		},
+	}
+	store.openFileFunc = store.openFile
+	store.openWriteFileFunc = store.openWriteFile
+	store.deleteFileFunc = store.deleteFile
```
jrick: May want to make these free funcs so you don't accidentally call them (statically) when running tests.

database2/ffboltdb/db.go
```diff
@@ -0,0 +1,1510 @@
...
+	// metadata.
+	blockIdxBucketName = []byte("ffboltdb-blockidx")
+
+	// writeLocKeyName is the key used to store the current write file
+	// location.
+	writeLocKeyName = []byte("ffboltdb-writeloc")
+)
+
+// Common error strings.
+var (
```
jrick: These can be `const`s.

database2/ffboltdb/db.go
```diff
@@ -0,0 +1,1510 @@
...
+// Enforce cursor implements the database.Cursor interface.
+var _ database.Cursor = (*cursor)(nil)
+
+// Bucket returns the bucket the cursor was created for.
+//
+// This function is part of the database.Cursor interface implementation.
+func (c *cursor) Bucket() database.Bucket {
+	// Ensure transaction state is valid.
+	if err := c.bucket.tx.checkClosed(); err != nil {
+		return nil
```
jrick: Are callers expected to check this?

davecgh: Well callers should never be closing the db and/or transaction out from under themselves, but it has the same semantics as bolt - it'll panic.

jrick: ok, that sounds good to me

database2/ffboltdb/db.go
```diff
@@ -0,0 +1,1510 @@
...
+// This function is part of the database.Bucket interface implementation.
+func (b *bucket) ForEach(fn func(k, v []byte) error) error {
+	// Ensure transaction state is valid.
+	if err := b.tx.checkClosed(); err != nil {
+		return err
+	}
+
+	// Keep track of the caller returned the error so it can be
+	// differentiated from a bolt error which needs to be converted.
+	var callerErr error
```
jrick: bolt returns the exact error the caller returned; it doesn't mangle it.

database2/ffboltdb/db.go
```diff
@@ -0,0 +1,1510 @@
...
+
+	// Ensure the transaction is writable.
+	if !tx.writable {
+		str := "store block requires a writable database transaction"
+		return makeDbErr(database.ErrTxNotWritable, str, nil)
+	}
+
+	// Reject the block if it already exists.
+	blockHash := block.Sha()
+	if tx.pendingBlocks != nil {
```
jrick: This nil check isn't needed (map keys can be looked up from a nil map without panicking).

database2/ffboltdb/db.go
```diff
@@ -0,0 +1,1510 @@
...
+// This function is part of the database.Tx interface implementation.
+func (tx *transaction) Metadata() database.Bucket {
+	return tx.metaBucket
+}
+
+// hasBlock returns whether or not a block with the given hash exists.
+func (tx *transaction) hasBlock(hash *wire.ShaHash) bool {
+	// Return true if the block is pending to be written on commit since
+	// it exists from the viewpoint of this transaction.
+	if tx.pendingBlocks != nil {
```
jrick: unneeded nil check

database2/ffboltdb/db.go
jrick: Also, this whole check can be removed, since it's also done by `(*transaction).hasBlock`.

database2/ffboltdb/db.go
```diff
@@ -0,0 +1,1510 @@
...
+// This function is part of the database.Tx interface implementation.
+func (tx *transaction) FetchBlockHeader(hash *wire.ShaHash) ([]byte, error) {
+	// Ensure transaction state is valid.
+	if err := tx.checkClosed(); err != nil {
+		return nil, err
+	}
+
+	// When the block is pending to be written on commit return the bytes
+	// from there.
+	if tx.pendingBlocks != nil {
```
jrick: unneeded check

database2/ffboltdb/db.go
```diff
@@ -0,0 +1,1510 @@
...
+
+	return blockRow, nil
+}
+
+// FetchBlockHeader returns the raw serialized bytes for the block header
+// identified by the given hash.  The raw bytes are in the format returned by
+// Serialize on a wire.BlockHeader.
+//
+// Returns the following errors as required by the interface contract:
+//   - ErrBlockNotFound if the request block hash does not exist
```
jrick: requested

database2/ffboltdb/db.go
```diff
@@ -0,0 +1,1510 @@
...
+	endOffset := blockLocSize + blockHdrSize
+	return blockRow[blockLocSize:endOffset:endOffset], nil
+}
+
+// FetchBlockHeaders returns the raw serialized bytes for the block headers
+// identified by the given hashes.  The raw bytes are in the format returned by
+// Serialize on a wire.BlockHeader.
+//
+// Returns the following errors as required by the interface contract:
+//   - ErrBlockNotFound if the request block hash does not exist
```
jrick: "if any of the requested block hashes"

database2/ffboltdb/db.go
```diff
@@ -0,0 +1,1510 @@
...
+	// values, so optimize for the common case.
+
+	// Load the headers.
+	headers := make([][]byte, len(hashes))
+	for i := range hashes {
+		hash := &hashes[i]
+
+		// When the block is pending to be written on commit return the
+		// bytes from there.
+		if tx.pendingBlocks != nil {
```
jrick: unneeded check

database2/ffboltdb/db.go
```diff
@@ -0,0 +1,1510 @@
...
+
+	return headers, nil
+}
+
+// FetchBlock returns the raw serialized bytes for the block identified by the
+// given hash.  The raw bytes are in the format returned by Serialize on a
+// wire.MsgBlock.
+//
+// Returns the following errors as required by the interface contract:
+//   - ErrBlockNotFound if the request block hash does not exist
```
jrick: requested

jrick: You can fix the rest of these with a global search and replace :)

database2/ffboltdb/db.go
```diff
@@ -0,0 +1,1510 @@
...
+// This function is part of the database.Tx interface implementation.
+func (tx *transaction) FetchBlock(hash *wire.ShaHash) ([]byte, error) {
+	// Ensure transaction state is valid.
+	if err := tx.checkClosed(); err != nil {
+		return nil, err
+	}
+
+	// When the block is pending to be written on commit return the bytes
+	// from there.
+	if tx.pendingBlocks != nil {
```
jrick: Unneeded check, I'd search the code for all of these and remove all of the checks when the map is not assigned to.

database2/ffboltdb/db.go
davecgh: Right.  It was there before because I was keeping the index to know when to replace the block later, but I forget to remove this check when I removed that flag.

database2/ffboltdb/db.go
```diff
@@ -0,0 +1,1510 @@
...
+
+// fetchPendingRegion attempts to fetch the provided region from any block which
+// are pending to be written on commit.  It will return nil for the byte slice
+// when there region references a block which is not pending.  When the region
+// does reference a pending block, it is bounds checked and returns
+// ErrBlockRegionInvalid if invalid.
+func (tx *transaction) fetchPendingRegion(region *database.BlockRegion) ([]byte, error) {
+	// When the block is pending to be written on commit return the bytes
+	// from there.
+	if idx, exists := tx.pendingBlocks[*region.Hash]; exists {
```
jrick: Might read a little better if this check was inversed with an early return (save an indent).

davecgh: Agreed

database2/ffboltdb/blockio.go
davecgh: I'd rather leave them as they are otherwise it will need an additional closure over the store state and complicate all the calling code.

database2/ffboltdb/db.go
davecgh: Agreed

database2/ffboltdb/db.go
davecgh: I agree with you on the ones that are just checking the map, but I think this one is better to stay because of the call to the function.  I know the function will return nil, but I think it makes the code a little clearer in this instance.

database2/ffboltdb/db.go
```diff
@@ -0,0 +1,1495 @@
...
+	return nil
+}
+
+// rollback rollsback the underly bolt database and closes the transaction.  It
+// is separated mainly so the code panics on attempts to commit or rollback a
+// managed transaction can rollback first.
+func (tx *transaction) rollback() error {
+	// Regardless of whether the rollback succeeds, the transaction is
+	// closed on return.
+	defer tx.close()
```
jrick: This can be easily (and readibly) be moved after the rollback without using the defer.

database2/ffboltdb/db.go
davecgh: Like the other one I mentioned, this is doing more than just checking the map, so I think this one is better kept.

I've changed all the others.

database2/ffboltdb/db.go
davecgh: It would need to be in 2 places without the defer.  I can do that if you prefer.

jrick:
```Go
err := tx.boltTx.Rollback()
tx.close()
if err != nil {
    return convertErr("failed to rollback underlying bolt tx", err)
}
return nil
```

davecgh: Fair enough.  I'll do that.

database2/ffboltdb/reconcile.go
```diff
@@ -0,0 +1,117 @@
...
+			"at file %d, offset %d", curFileNum, curOffset,
+			wc.curFileNum, wc.curOffset)
+		pdb.store.handleRollback(curFileNum, curOffset)
+		log.Infof("Database sync complete")
+	}
+
+	// When the write cursor position found by scanning the block files on
+	// disk is BEFORE the position the metadata believes to be true, return
+	// a corruption error.  Since sync is called after each block is written
+	// and before the metadata is updated, this should only happen, in the
```
davecgh: , after happen not needed.

database2/ffboltdb/interface_test.go
```diff
@@ -0,0 +1,2266 @@
...
+			return false
+		}
+	}
+
+	// Ensure the first transaction of every block fetched in bulk block
+	// regions from the database matches the expected bytes.
+	allRegionBytes, err := tx.FetchBlockRegions(allBlockRegions)
+	if err != nil {
+		tc.t.Errorf("FetchBlockRegions: unexpected error: %v", err)
+
```
jrick: extra newline or missing return?

database2/ffboltdb/interface_test.go
```diff
@@ -0,0 +1,2266 @@
...
+			return false
+		}
+	}
+
+	// Ensure the bulk determination of whether a set of block hashes are in
+	// the database returns true for all loaded blocks.
+	hasBlocks, err := tx.HasBlocks(allBlockHashes)
+	if err != nil {
+		tc.t.Errorf("HasBlocks: unexpected error: %v", err)
+	}
```
jrick: missing return?

database2/ffboltdb/interface_test.go
davecgh: Yep missing a return there.  It catches the mismatch just below, but better to be consistent.

database2/ffboltdb/interface_test.go
davecgh: Yes.

davecgh:
Rebased to latest master and squashed all review up to this point.

davecgh:
Rebased to latest master and updated for relicensing.

davecgh:
Updated to add new `ffldb` backend and rebased to latest master.

davecgh:
Updated to remove `ffboltdb` backend, rebase to latest master, and squashed to a single commit.

database2/ffldb/blockio.go
```diff
@@ -0,0 +1,749 @@
...
+// blockFilePath return the file path for the provided block file number.
+func blockFilePath(dbPath string, fileNum uint32) string {
+	fileName := fmt.Sprintf(blockFilenameTemplate, fileNum)
+	return filepath.Join(dbPath, fileName)
+}
+
+// openWriteFile returns a file handle for the passed flat file number in
+// read/write mode.  The file will be created if needed.  It is typically used
+// for the current file that will have all new data appended.  Unlike openFile,
+// this function does not keep track the open file and it is not subject to the
```
Roasbeef: Typo: `does not keep track the open file`
Correction: `does not keep track *of* the open file`

davecgh: Updated.

davecgh:
Rebased to the latest master.

davecgh:
Added a database cache layer implementation.  It's ready for review.

davecgh:
For reference, this PR should be ready once the other required PRs discussed in the description are done.  The only thing this one could use is perhaps a few more whitebox tests to hit a few extremely highly unlikely to happen corner cases.  However, the test coverage is already over 90% and includes a significant amount of both positive and negative code path testing.

database2/README.md
```diff
@@ -0,0 +1,77 @@
...
+However, this package could be extremely useful for any applications requiring
+Bitcoin block storage capabilities.
+
+As of July 2015, there are over 365,000 blocks in the Bitcoin block chain and
+and over 76 million transactions (which turns out to be over 35GB of data).
+This package provides a database layer to store and retrieve this data in a
+simple and efficient manner.
+
+The default backend, ffldb, has a strong focus on speed, efficiency, and
+robustness.  It makes use leveldb for the metadata, flat files for block
```
wallclockbuilder: It makes use leveldb for the metadata 
--->
It makes use **of** leveldb for the metadata

davecgh: Fixed.

davecgh:
Squashed reviewed code and updates into the base commit and rebased to the latest master.  The unreviewed database cache layer code has been left separate.

davecgh:
Rebased to latest master.

davecgh:
Rebased to latest master.

davecgh:
Rebased to latest master.

davecgh:
Rebased to latest master.

davecgh:
Rebased to latest master.

davecgh:
OK, so I'm making a push to get this merged within the new couple of days.  It has been running for months with no issues.

Calling for final reviews.

davecgh:
Reviewed 29 of 42 files at r1, 4 of 7 files at r2, 1 of 2 files at r4, 5 of 8 files at r5, 5 of 5 files at r6.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/380)*

Dirbaio:
LGTM :+1:

I've been running with database2 in production for a few months and it works great.

dajohi:
OK from me.  Been running for months without issue.

jrick:
ok here too.

davecgh:
Reviewed 39 of 39 files at r7.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/380)*

PR #385, opened 2015-04-17
Fix typo and clean up comments for godoc.

jrick:
Lists and other such things should have a two space prefix in the comment so they are rendered as blocks (and with a monospace typeface) by godoc.  Otherwise a list will be interpreted as a single paragraph.


davecgh:
OK

PR #386, opened 2015-04-20
blockchain: Split block and header validation.

davecgh:
This pull request refactors the consensus rule checks for block headers and blocks in the blockchain package into separate functions.  These changes contain no modifications to consensus rules and the code still passes all block consensus tests.  It is only a refactoring.

This is being done to help pave the way toward supporting concurrent downloads.  While the package already supports headers-first mode up through the latest checkpoint through the use of the BFFastAdd flag and hard-coded checkpoints, it currently only works when downloading from a single peer.  In order to support concurrent downloads from multiple peers, the ability for the caller to do things such as independently checking a block header (both context-free and full-context checks) will be needed.

There are several more changes that will be necessary to support concurrent downloads as well, such as making the package concurrent safe, modifying it to make use of the new database API, etc.  Those changes are planned for future commits.

### DIFF
--- a/blockchain/accept.go
+++ b/blockchain/accept.go
@@ -4,11 +4,7 @@
 
 package blockchain
 
-import (
-	"fmt"
-
-	"github.com/btcsuite/btcutil"
-)
+import "github.com/btcsuite/btcutil"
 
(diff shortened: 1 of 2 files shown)


jrick:
ok

PR #387, opened 2015-04-20
txscript: Remove unneeded param from NewScript.

davecgh:
This pull request removes the unnecessary sigScript parameter from the txscript.NewScript function.  This has bothered me for a while because it can and really should be obtained from the provided transaction and input index.  The way it was, the passed script could technically be different than what is in the transaction.  Obviously that would be an improper use of the API, but it's safer and more convenient to simply pull it from the provided transaction and index.

Also, since the function signature is changing anyways, make the input index parameter come after the transaction which it references.


dajohi:
OK

PR #389, opened 2015-04-20
txscript: Rename Script to Engine.

davecgh:
This pull request renames the Script type to Engine to better reflect its purpose.  It also renames the NewScript function to NewEngine to match.

This is being done because name Script for the engine is confusing since it implies it is an actual script rather than the execution environment for the script.  It also paves the way for eventually supplying a
ParsedScript type which will be less likely to be confused with the execution environment.

While moving the code, some additional variable names and comments have been updated to better match the style used throughout the rest of the code base.  In addition, an attempt has been made to use consistent naming of the engine as 'vm' instead of using different variables names as it was previously.

Finally, the relevant engine code has been moved into a new file named engine.go and related tests moved to engine_test.go.


jrick:
ok

PR #390, opened 2015-04-20
txscript: Move error definitions to error.go.

davecgh:
This is more consistent with the rest of the code base and also will make it easier to improve the errors to provide more details at a later date.


dajohi:
OK

PR #391, opened 2015-04-20
txscript: Separate signing code.

davecgh:
This pull request separates the transaction signing code into sign.go and the related tests into sign_test.go.


dajohi:
OK

PR #392, opened 2015-04-21
txscript: Store flags in instance versus bools.

davecgh:
Rather than storing a separate bool for whether or not each flag is set in every script engine instance, store the flags and check if the relevant flag is set from each specific location.

This reduces the memory needed by each script engine instance and means future flags will not require new fields.


dajohi:
OK

PR #393, opened 2015-04-21
txscript: Define opcodes in hex and sync BC opcodes.

davecgh:
This pull request modifies the definition of the opcodes to their hex counterparts rather than decimal since it is far more common to see scripts in hex.  This makes it easier when manually looking at script dumps to correlate opcodes.  However, since there are also cases where it is useful to see the decimal value of the opcode, the decimal value has been left as a comment.  Obviously converting the numbers is trivial, but
it is handy when looking at the opcode definitions to already have it there.

In addition, it syncs the opcodes with the latest Bitcoin Core internal opcodes for completeness and modifies the tests accordingly.


dajohi:
OK

PR #395, opened 2015-04-21
txscript: Convert opcode map to an array.

davecgh:
Depends on #393.

This pull request converts the opcode map to an array to improve performance.

Benchmark of executing a standard p2pk transaction:
```
New: BenchmarkExecute            2000            784349 ns/op
Old: BenchmarkExecute            2000            792600 ns/op
```
The time is dominated by the signature checking as expected, however there is still an increase in speed.


dajohi:
OK

PR #397, opened 2015-04-22
Create and use a new SOCKS5 subpackage.

dajohi:
The new SOCKS5 subpackage includes all the existing funcationality
plus supports Tor stream isolation.  This can be acheived by using
the --proxyrandomize option.

Idea from Wladimir J. van der Laan via Bitcoin Core:
https://github.com/bitcoin/bitcoin/pull/5911

### DIFF
--- a/config.go
+++ b/config.go
@@ -19,10 +19,10 @@ import (
 	"github.com/btcsuite/btcd/database"
 	_ "github.com/btcsuite/btcd/database/ldb"
 	_ "github.com/btcsuite/btcd/database/memdb"
+	"github.com/btcsuite/btcd/socks5"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcutil"
 	flags "github.com/btcsuite/go-flags"
-	"github.com/btcsuite/go-socks/socks"
 )
 
 const (
@@ -91,6 +91,7 @@ type config struct {
 	Proxy              string        `long:"proxy" description:"Connect via SOCKS5 proxy (eg. 127.0.0.1:9050)"`
 	ProxyUser          string        `long:"proxyuser" description:"Username for proxy server"`
 	ProxyPass          string        `long:"proxypass" default-mask:"-" description:"Password for proxy server"`
+	ProxyRandomize     bool          `long:"proxyrandomize" description:"Randomize user credentials for each proxy connection. This enables Tor stream isolation"`
 	OnionProxy         string        `long:"onion" description:"Connect to tor hidden services via SOCKS5 proxy (eg. 127.0.0.1:9050)"`
 	OnionProxyUser     string        `long:"onionuser" description:"Username for onion proxy server"`
 	OnionProxyPass     string        `long:"onionpass" default-mask:"-" description:"Password for onion proxy server"`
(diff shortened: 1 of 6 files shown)


doc.go
```diff
@@ -58,6 +58,7 @@ Application Options:
       --proxy=             Connect via SOCKS5 proxy (eg. 127.0.0.1:9050)
       --proxyuser=         Username for proxy server
       --proxypass=         Password for proxy server
+      --proxyrandomize     Randomize user credentials for each proxy connection. This enables Tor stream isolation
```
davecgh: This doesn't match how it's actually shown.  Please use `./btcd -h` and copy it from there so this matches.

davecgh:
All of the functions need comments.

socks5/socks5.go
```diff
@@ -0,0 +1,273 @@
...
+		statusNetworkUnreachable:      errors.New("network unreachable"),
+		statusHostUnreachable:         errors.New("host unreachable"),
+		statusConnectionRefused:       errors.New("connection refused by destination host"),
+		statusTtlExpired:              errors.New("TTL expired"),
+		statusCommandNotSupport:       errors.New("command not supported / protocol error"),
+		statusAddressTypeNotSupported: errors.New("address type not supported"),
+	}
+)
+
+// Socks5
```
davecgh: This appears to be unfinished.

davecgh:
As Travis pointed out, there are a lot of lint issues that need to be resolved.

davecgh:
Package needs a `README.md`.

davecgh:
Please updated the [Tor documentation](https://github.com/btcsuite/btcd/blob/master/docs/configuring_tor.md) in light of the new functionality.

socks5/socks5.go
```diff
@@ -0,0 +1,273 @@
...
+
+	if _, err := io.ReadFull(conn, res2[:2]); err != nil {
+		conn.Close()
+		return nil, err
+	}
+
+	return conn, nil
+}
+
+func (s *Socks5) Randomize() bool {
```
davecgh: Should probably used `Randomized`.  This reads like a verb as if it's going to do something rather than returning the state of randomization.

socks5/socks5.go
```diff
@@ -0,0 +1,273 @@
...
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+
+	return s.randomize
+}
+
+func (s *Socks5) SetRandomize(randomize bool) {
```
davecgh: If `Randomize` is changed to `Randomized`, this should go to `SetRandomized` and the internal variable should be `randomized` to match.

config.go
```diff
@@ -736,14 +739,14 @@ func loadConfig() (*config, []string, error) {
...
 	// proxy than normal traffic.
 	if cfg.OnionProxy != "" {
-		cfg.oniondial = func(a, b string) (net.Conn, error) {
-			proxy := &socks.Proxy{
-				Addr:     cfg.OnionProxy,
-				Username: cfg.OnionProxyUser,
-				Password: cfg.OnionProxyPass,
-			}
-			return proxy.Dial(a, b)
+		onionProxy, err := socks5.New(cfg.OnionProxy)
```
davecgh: I'd prefer the options to be included in the call to `New`, like the previous package.  It's easy to forget to set additional options when they're aren't part of the `New` call struct params.

config.go
```diff
@@ -714,11 +715,13 @@ func loadConfig() (*config, []string, error) {
 	cfg.dial = net.Dial
 	cfg.lookup = net.LookupIP
 	if cfg.Proxy != "" {
-		proxy := &socks.Proxy{
-			Addr:     cfg.Proxy,
-			Username: cfg.ProxyUser,
-			Password: cfg.ProxyPass,
+		proxy, err := socks5.New(cfg.Proxy)
```
davecgh: Would obviously affect this too.

peer.go
```diff
@@ -93,17 +92,6 @@ func newNetAddress(addr net.Addr, services wire.ServiceFlag) (*wire.NetAddress,
 		return na, nil
 	}
 
-	// addr will be a socks.ProxiedAddr when using a proxy.
-	if proxiedAddr, ok := addr.(*socks.ProxiedAddr); ok {
```
davecgh: This should likely be uppated for the new package rather than just removed.

davecgh:
This needs to have the licenses updated for the relicense since it contains new files.

dajohi:
Closing in favor of new PR (for now)

PR #398, opened 2015-04-22
Fix 'add/delnode' type switch evaluation in server

Roasbeef:
* The cases for the 'addnode' command were previously
   stacked on top the new cases for the 'node' command.
   The intended behavior was to create a fall through and
   handle both commands. However, trying to use this
   syntax with a type switch caused the first case to be
   ignored.
* The 'addnode' cases are now handled separately from the
   'node' commands in the server.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -364,11 +364,11 @@ func handleAddNode(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (in
 	var err error
 	switch c.SubCmd {
 	case "add":
-		err = s.server.AddAddr(addr, true)
+		err = s.server.ConnectNode(addr, true)
 	case "remove":
-		err = s.server.RemoveAddr(addr)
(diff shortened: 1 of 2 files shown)


davecgh:
Rather than duplicating the code, can you refactor it out into functions that both call?

Roasbeef:
Agree about avoiding code duplication. I've taken a slightly different angle in the latest commit. Instead of adding new functions factoring out the repeated logic, the previous `addnode` specific functions and msg structs are now proxied to the new `node` functions. This should properly capture the behavior I was trying to create initially. Namely, re-using the shared code for 'add/remove' and 'connect/remove'.

davecgh:
Thanks.  I'm going to test this, but it reads well.

davecgh:
I've tested that both the node and the addnode command work properly with this.

OK after rebase.

Roasbeef:
Rebased.

PR #399, opened 2015-04-23
Remove account defaults for wallet requests.

jrick:
btcwallet will need to check whether these requests were unset, so defaults cannot be automatically filled in by btcjson.

~~This currently depends on several unmerged btcwallet PRs so it cannot be merged at this time.~~ ready


davecgh:
OK once the deps are in place.

davecgh:
OK

PR #400, opened 2015-04-23
txscript: Improve conditional stack.

davecgh:
This pull request improves the way the conditional execution stack is handled in a few ways.

First, the current execution state is now pushed onto the end of the slice rather than the front of it.  This has been done because it results in fewer allocations and is therefore more efficient.

Second, the need for allocating and setting an initial true in the conditional stack has been eliminated.  The vast majority of scripts don't contain any conditionals, so there is no reason to allocate a slice when
it isn't needed.

Third, a new function has been added to the engine to determine if the current conditional branch is executing named `isBranchExecuting` which handles the fact the conditional execution stack can now be empty and
improves the readability of the code.

Finally, it removes a couple of TODOs which I have verified do not apply.

### DIFF
--- a/txscript/engine.go
+++ b/txscript/engine.go
@@ -100,6 +100,17 @@ func (vm *Engine) hasFlag(flag ScriptFlags) bool {
 	return vm.flags&flag == flag
 }
 
+// isBranchExecuting returns whether or not the current conditional branch is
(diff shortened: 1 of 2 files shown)


dajohi:
OK

PR #402, opened 2015-04-23
btcjson: Remove NewOutPointFromWire and wire dep.

### LINKED ISSUE #401: package v2/btcjson only dependencies are core Go packages
package v2/btcjson
has no dependencies on btcd or any other packages except in ONE place
and that is here...

func NewOutPointFromWire(op *wire.OutPoint) *OutPoint {
and this function currently is not being used by anyone

so if you wanted to have v2/btcjson only depend on Go's core packages

you could remove that reference and then everyone could use it going forward without needing to use anything but the core go packages

davecgh:
This pull request removes the `NewOutPointFromWire` function from the `btcjson` version 2 package so it can be used without needing the wire package as a dependency.  It also updates the tests accordingly.

This results in the package only depending on core Go packages.

Closes #401.

### DIFF
--- a/btcjson/v2/btcjson/chainsvrwscmds.go
+++ b/btcjson/v2/btcjson/chainsvrwscmds.go
@@ -7,10 +7,6 @@
 
 package btcjson
 
-import (
-	"github.com/btcsuite/btcd/wire"
-)
-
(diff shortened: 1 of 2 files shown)


jrick:
ok

PR #403, opened 2015-04-25
txscript: Unexport Stack type.

davecgh:
This pull request unexports the `txscript.Stack` type since it is only intended to be used internally during script execution.  Further, the engine exposes the `{G,S}etStack` and `{G,S}etAltStack` functions which return the items as a slice of byte slices ([][]byte) for caller access while stepping.


jrick:
ok

PR #405, opened 2015-04-27
txscript: Move opcode execution logic to engine

davecgh:
This pull request moves the opcode execution logic from the opcode type to the engine type because execution of an opcode modifies the engine state (primarily the main and alternate data stacks) as opposed to the state of the opcode.  Making the engine the receiver more clearly indicates this fact.


dajohi:
OK

PR #407, opened 2015-04-28
Create `generate` RPC command to close #362

### LINKED ISSUE #362: [rpc] Add new 'generate' command
It is quite useful when doing regression and simulation testing to be able to generate a specific number of blocks.  Currently, the only available option is to call `setgenerate` to enable and disable block generation, but there is no way to generate a specific number of blocks.

Thus, a new command named `generate` should be introduced which takes a single numeric (int) parameter with the number of blocks to generate.  The RPC should block until the blocks are generated and it should return the hashes of the generated blocks as a JSON array of strings.

aakselrod:
Create GenerateCmd in btcjson v2. Update tests to check GenerateCmd.

Update chaincfg/params.go with a new bool in Params, GenerateSupported,
with true values in SimNetParams and RegressionNetParams and false in
the others.

Create new flag, discreteMining, in CPUMiner struct.

Add GenerateNBlocks function to cpuminer.go and handleGenerate
function to rpcserver.go.

Update documentation for the RPC calls.


btcjson/v2/btcjson/btcdextcmds.go
```diff
@@ -58,6 +58,19 @@ func NewDebugLevelCmd(levelSpec string) *DebugLevelCmd {
 	}
 }
 
+// GenerateCmd defines the generate JSON-RPC command.
+type GenerateCmd struct {
+	NumBlocks uint32 `json:"numblocks"`
```
davecgh: No need to specify the tag on the command.  JSON-RPC 1.0 uses positional params.

btcjson/v2/btcjson/chainsvrresults.go
```diff
@@ -6,6 +6,11 @@ package btcjson
 
 import "encoding/json"
 
+// GenerateResult models the data returned from the generate command.
+type GenerateResult struct {
+	Blockhashes []string `json:"blockhashes"`
```
davecgh: This type isn't needed (or correct).  The result type of the `generate` RPC should just be a primitive JSON array.  By putting it into a struct like this, it would result in a json object like so:
`{'blockhashes': ["hash1","hash2"]}`

Instead of the desired:

`["hash1","hash2"]`

rpcserver.go
```diff
@@ -784,6 +785,50 @@ func handleDecodeScript(s *rpcServer, cmd interface{}, closeChan <-chan struct{}
...
+	// Respond with an error if the client is requesting 0 blocks to be generated.
+	if c.NumBlocks == 0 {
+		return nil, &btcjson.RPCError{
+			Code:    btcjson.ErrRPCInternal.Code,
+			Message: "Please request a nonzero number of blocks to generate.",
+		}
+	}
+
+	// Create a reply
+	reply := btcjson.GenerateResult{
```
davecgh: As mentioned previously, this would create a JSON object instead of just a JSON array of strings.  Thus, this should just be:

`reply := make([]string, c.NumBlocks)`

rpcserver.go
```diff
@@ -784,6 +785,50 @@ func handleDecodeScript(s *rpcServer, cmd interface{}, closeChan <-chan struct{}
...
+		return nil, &btcjson.RPCError{
+			Code:    btcjson.ErrRPCInternal.Code,
+			Message: err.Error(),
+		}
+	}
+
+	// Mine the correct number of blocks, assigning the hex representation of the
+	// hash of each one to its place in the reply.
+	for i, hash := range blockHashes {
+		reply.Blockhashes[i] = hash.String()
```
davecgh: Then:

`reply[i] = hash.String()`

rpcserverhelp.go
```diff
@@ -506,6 +514,7 @@ var rpcResultTypes = map[string][]interface{}{
 	"debuglevel":            []interface{}{(*string)(nil), (*string)(nil)},
 	"decoderawtransaction":  []interface{}{(*btcjson.TxRawDecodeResult)(nil)},
 	"decodescript":          []interface{}{(*btcjson.DecodeScriptResult)(nil)},
+	"generate":              []interface{}{(*btcjson.GenerateResult)(nil)},
```
davecgh: Then this would be:

` []interface{}{(*[]string)(nil)},`

rpcserverhelp.go
```diff
@@ -99,6 +99,14 @@ var helpDescsEnUS = map[string]string{
 	"decodescript--synopsis": "Returns a JSON object with information about the provided hex-encoded script.",
 	"decodescript-hexscript": "Hex-encoded script",
 
+	// GenerateCmd help
+	"generate--synopsis": "Generates a set number of blocks (simnet or regtest only) and returns a JSON\n" +
+		" array of their hashes.",
+	"generate-numblocks": "Number of blocks to generate",
+
+	// GenerateResult help
+	"generateresult-blockhashes": "The hashes, in order, of blocks generated by the call",
```
davecgh: This would become:

`"generateresult--result0": "The hashes, in order, of blocks generated by the call"`,

cpuminer.go
```diff
@@ -508,3 +509,112 @@ func newCPUMiner(s *server) *CPUMiner {
...
+	// Start a ticker which is used to signal checks for stale work and
+	// updates to the speed monitor.
+	ticker := time.NewTicker(time.Second * hashUpdateSecs)
+	defer ticker.Stop()
+
+	for {
+		// Wait until there is a connection to at least one other peer
+		// since there is no way to relay a found block or receive
+		// transactions to work on when there are no connected peers.
+		if m.server.ConnectedCount() == 0 {
```
davecgh: This makes since for the normal mining path, but I don't think it should be checked for the discreet `generate` path.  It is pretty common to want to create a few blocks against a single node when doing regression testing.

cpuminer.go
```diff
@@ -508,3 +509,112 @@ func newCPUMiner(s *server) *CPUMiner {
 		updateHashes:      make(chan uint64),
 	}
 }
+
+// generateNBlocks generates the requested number of blocks. It is self
+// contained in that it creates block templates and attempts to solve them while
+// detecting when it is performing stale work and reacting accordingly by
+// generating a new block template.  When a block is solved, it is submitted.
+// The function returns a list of the hashes of generated blocks.
+func (m *CPUMiner) generateNBlocks(n uint32) ([]*wire.ShaHash, error) {
```
davecgh: Please define this before the `newCPUMiner` function so it is logically grouped with the other receiver functions.

cpuminer.go
aakselrod: ACK. That makes testing faster for me. Note that this will work for simnet as well, so you can have a single-node simnet now.

cpuminer.go
aakselrod: ACK.

btcjson/v2/btcjson/btcdextcmds.go
aakselrod: ACK.

btcjson/v2/btcjson/chainsvrresults.go
aakselrod: ACK.

rpcserver.go
aakselrod: ACK.

rpcserver.go
aakselrod: ACK.

rpcserverhelp.go
aakselrod: ACK.

rpcserverhelp.go
aakselrod: ACK.

rpcserverhelp.go
```diff
@@ -99,6 +99,14 @@ var helpDescsEnUS = map[string]string{
 	"decodescript--synopsis": "Returns a JSON object with information about the provided hex-encoded script.",
 	"decodescript-hexscript": "Hex-encoded script",
 
+	// GenerateCmd help
+	"generate--synopsis": "Generates a set number of blocks (simnet or regtest only) and returns a JSON\n" +
+		" array of their hashes.",
+	"generate-numblocks": "Number of blocks to generate",
+
+	// GenerateResult help
```
davecgh: Super Nit: This comment can go away since the type has been removed.   Once that goes away, so should the blank line above it.

cpuminer.go
davecgh: Also, please capitalize this function since it's intended to be called from outside of the miner (by the RPC server).  I'm thinking ahead to separation.  Don't forget to update the comment too when you change it!

cpuminer.go
```diff
@@ -496,6 +497,107 @@ func (m *CPUMiner) NumWorkers() int32 {
...
+		}
+
+		// No point in searching for a solution before the chain is
+		// synced.  Also, grab the same lock as used for block
+		// submission, since the current block will be changing and
+		// this would otherwise end up building a new block template on
+		// a block that is in the process of becoming stale.
+		m.submitBlockLock.Lock()
+		_, curHeight := m.server.blockManager.chainState.Best()
+		if curHeight != 0 && !m.server.blockManager.IsCurrent() {
```
davecgh: Like the `ConnectedCount` check you already removed, I believe this check can go away too.  It makes sense for the normal mining path, but I don't think it should be checked for the discreet generate path.  What will happen is if you have your simnet node with new blocks for ~24 hours, you'd get stuck here since there are no other blocks coming in and this condition can never bet met.

davecgh:
Also, I managed to hit a case where it was generating a bunch of blocks and I issued a `setgenerate false` from another thread.  That incorrectly allowed it to stop the mining process which leads to a (recovered) panic due to an attempted double close of the channel.

I believe this will need a separate flag to prevent any changes via `setgenerate` while discreet mining is active.  The reverse case seem to already be covered (can't issue `generate` while mining is already active).

rpcserverhelp.go
aakselrod: ACK, removed comment/blank line.

cpuminer.go
aakselrod: ACK, removed IsCurrent() check and updated comments.

aakselrod:
ACK, I created a new discreteMining flag and check it in Stop(), Start(), and GenerateNBlocks where it's set/cleared.

cpuminer.go
aakselrod: ACK, this is done. Also updated rpcserver.go to call the function using the new name.

davecgh:
Code updates look good.  I'm testing it out now.

PR #408, opened 2015-04-29
txscript: Code consistency and doco improvements.

davecgh:
This pull request contains a lot of cleanup on the txscript code to make it more consistent with the code throughout the rest of the project.  It doesn't change any operational logic.

- Add a significant number of comments throughout in order to better explain what the code is doing
- Fix several comment typos
- Move a couple of constants only used by the engine to engine.go
- Move a variable only used by the engine to engine.go
- Fix a couple of format specifiers in the test prints
- Reorder functions so they're defined before/closer to use
- Make the code lint clean with the exception of the opcode definitions

### DIFF
--- a/txscript/engine.go
+++ b/txscript/engine.go
@@ -8,11 +8,12 @@ import (
 	"fmt"
 	"math/big"
 
+	"github.com/btcsuite/btcd/btcec"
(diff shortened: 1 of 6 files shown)


dajohi:
OK

PR #409, opened 2015-04-30
txscript: Convert to new scriptnum type.

davecgh:
NOTE: Depends on #408.

This pull request implements a new type, named scriptNum, for handling all numeric values used in scripts and converts the code over to make use of it.  This is being done for a few of reasons.

First, the consensus rules for handling numeric values in the scripts require special handling with subtle semantics.  By encapsulating those details into a type specifically dedicated to that purpose, it simplifies the code and generally helps prevent improper usage.

Second, the new type is quite a bit more efficient than big.Ints which are designed to be arbitrarily large and thus involve a lot of heap allocations and additional multi-precision bookkeeping.  Because this new type is based on an int64, it allows the numbers to be stack allocated thereby eliminating a lot of GC and also eliminates the extra multi-precision arithmetic bookkeeping.

The use of an int64 is possible because the consensus rules dictate that when data is interpreted as a number, it is limited to an int32 even though results outside of this range are allowed so long as they are not
interpreted as integers again themselves.   Thus, the maximum possible result comes from multiplying a max int32 by itself which safely fits into an int64 and can then still appropriately provide the serialization
of the larger number as required by consensus.

Finally, it more closely resembles the implementation used by Bitcoin Core and thus makes is easier to compare the behavior between the two implementations.

This pull request also includes a full suite of tests with 100% coverage of the semantics of the new type.

### DIFF
--- a/mining.go
+++ b/mining.go
@@ -182,7 +182,8 @@ func mergeTxStore(txStoreA blockchain.TxStore, txStoreB blockchain.TxStore) {
 // the extra nonce as well as additional coinbase flags.
 func standardCoinbaseScript(nextBlockHeight int64, extraNonce uint64) ([]byte, error) {
 	return txscript.NewScriptBuilder().AddInt64(nextBlockHeight).
-		AddUint64(extraNonce).AddData([]byte(coinbaseFlags)).Script()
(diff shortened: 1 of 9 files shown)


davecgh:
Rebased to lasest master and updated for relicensing.

dajohi:
OK

PR #410, opened 2015-04-30
txscript: Separate code for standard scripts.

davecgh:
Depends on #409.

This pull request moves all code related to standard scripts into a separate file named standard.go as well as the associated tests into standard_test.go.  Since the code in address.go and address_test.go is only related to standard scripts, it has been combined into the new files and the old files deleted.

The intent here is to make it clear that the code in standard.go is not related to consensus.


davecgh:
Rebased and updated for relicensing.

dajohi:
OK

PR #412, opened 2015-05-02
txscript: Test consistency and cleanup.

davecgh:
- Move reference tests to test package since they are intended to exercise the engine as callers would
- Improve the short form script parsing to allow additional opcodes: DATA_#, OP_#, FALSE, TRUE
- Make use of a function to decode hex strings rather than manually defining byte slices
- Update the tests to make use of the short form script parsing logic rather than manually defining byte slices
- Consistently replace all []byte{} and [][]byte{} with nil
- Define tests only used in a specific function inside that func
- Move invalid flag combination test to engine_test since that is what it is testing
- Remove all redundant script tests in favor of the JSON-based tests in the data directory.
- Move several functions from internal_test.go to the test files associated with what the tests are checking

### DIFF
--- a/txscript/engine_test.go
+++ b/txscript/engine_test.go
@@ -54,7 +54,7 @@ func TestBadPC(t *testing.T) {
 		TxOut: []*wire.TxOut{
 			{
 				Value:    1000000000,
-				PkScript: []byte{},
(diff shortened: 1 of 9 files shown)


dajohi:
OK

PR #417, opened 2015-05-04
rpcserver: Omit empty getblock->nextblockhash.

### LINKED ISSUE #416: `getblock` on tip returns an empty string for `nextblockhash`
Not sure whether this is the intended behavior. On bitcoin core, `nextblockhash` is not in the dictionary.

davecgh:
This pull request omits the nextblockhash field from the getblock RPC result when there isn't one (i.e for current the tip).

Closes #416.


jrick:
ok

PR #419, opened 2015-05-05
Add new option --maxorphantx

dajohi:
The option `--maxorphantx` allows the user to specify the number of
orphan transactions to keep in memory.


config.go
```diff
@@ -600,6 +602,16 @@ func loadConfig() (*config, []string, error) {
 		return nil, nil, err
 	}
 
+	// Limit the max orphan count to a sane vlue.
+	if cfg.MaxOrphanTxs < 0 {
```
davecgh: I'd suggest making the config option a `uint32` so the parser will prevent negatives for you.

davecgh:
Also, this would need to set the default to the current value of 10k.

davecgh:
Also, please don't forget that new options need entries in doc.go and `sample-btcd.conf`.

config.go
dajohi: That produces ugly help output and also requires casting then.

mempool.go
```diff
@@ -500,11 +500,6 @@ func (mp *txMemPool) maybeAddOrphan(tx *btcutil.Tx) error {
 	// orphans.  In the case there is a valid transaction larger than this,
 	// it will ultimtely be rebroadcast after the parent transactions
 	// have been mined or otherwise received.
-	//
-	// Note that the number of orphan transactions in the orphan pool is
-	// also limited, so this equates to a maximum memory used of
-	// maxOrphanTxSize * maxOrphanTransactions (which is 500MB as of the
```
davecgh: Please leave the note here.  Just modify it to say * cfg.MaxOrphanTransactions and `which is 500MB with the default values as of the time this comment was written).

config.go
davecgh: Fair enough.  I prefer to use appropriate types in general (if something can't be negative, make it unsigned), but I suppose the nicer error message is worth it.

doc.go
```diff
@@ -81,7 +81,8 @@ Application Options:
       --limitfreerelay=    Limit relay of transactions with no transaction fee
                            to the given amount in thousands of bytes per minute
                            (15)
-
+      --maxorphantx=       Max number of orphan transactions to keep in memory
+                           (10000)
```
davecgh: As discussed, let's lower the default max orphans to 1000 along with this PR.

sample-btcd.conf
```diff
@@ -204,6 +204,12 @@
 ; notls=1
 
 ; ------------------------------------------------------------------------------
+; Mempool Settings - The following options
+; ------------------------------------------------------------------------------
+; Limit orphan transaction pool to 100 transactions.
+; maxorphantx=100
```
davecgh: This should also be the default value, which will be 1000.

davecgh:
OK

PR #422, opened 2015-05-06
rpcserver: Omit empty getrawtransaction->confirmations.

### LINKED ISSUE #420: `getrawtransactions` includes confirmations=0 for unconfirmed tx
Bitcoin core omits the field `confirmation` in case the transaction in unconfirmed, while btcd includes it with value 0.

Similar to #416

davecgh:
This pull request omits the confirmations field from the `getrawtransactions` RPC result when it's zero.

Closes #420.


jrick:
ok

PR #425, opened 2015-05-06
txscript: Consensus audit.

davecgh:
This pull request contains fixes from the results of a thorough audit of txscript to find any cases of script evaluation which doesn't match the required consensus behavior.  These conditions are fairly obscure and
highly unlikely to happen in any real scripts, but they could have nevertheless been used by a clever attacker with malicious intent to cause a fork.

Test cases which exercise these conditions have been added to the reference tests and will contributed upstream to improve the quality for the entire ecosystem.


dajohi:
OK

jonasnick:
Passes all of the testcases that spotted the aforementioned bugs.

PR #427, opened 2015-05-06
Fix listtransactions/gettransaction result structs.

jrick:
The following changes were made to ListTransactionsResult (which
models the long result format used by listtransactions,
listsinceblock, etc.):

  - Fee made optional (float64 -> *float64 + omitempty)
  - BlockIndex made optional (int64 + omitempty -> *int64 + omitempty)
  - InvolvesWatchOnly added (bool + omitempty)
  - Vout added (uint32)

The following changes were made to GetTransactionDetailsResult (which
models the short result format of listtransactions):

  - InvolvesWatchOnly added (bool + omitempty)
  - Fee added (*float64 + omitempty)
  - Vout added (uint32)

The combination of pointer types and the omitempty struct tag allow
excluding the field from the JSON object, or including it with the
zero value.  This is useful in particular for the fee fields, which
should be included whenever the category is "send" even if the fee is
zero.  Other optional fields which are only added to the result object
with non-zero values (such as includeswatchonly) can be reduced to
simply an omitempty tag without the pointer type.

### DIFF
--- a/btcjson/helpers.go
+++ b/btcjson/helpers.go
@@ -60,6 +60,14 @@ func Uint64(v uint64) *uint64 {
 	return p
 }
 
+// Float64 is a helper routine that allocates a new float64 value to store v and
+// returns a pointer to it.  This is useful when assigning optional parameters.
+func Float64(v float64) *float64 {
+	p := new(float64)
+	*p = v
+	return p
+}
+
 // String is a helper routine that allocates a new string value to store v and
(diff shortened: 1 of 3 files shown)


jrick:
This obviously isn't ready yet since tests don't pass nor does btcd build :)

jrick:
Ready for review now

davecgh:
OK.

PR #429, opened 2015-05-12
Update docs for IRC move to freenode.

dajohi:


luke-jr:
Maybe it should mention #bitcoin-dev too?

davecgh:
The main purpose of this channel is for btcd support as opposed to development.

OK on PR.

PR #433, opened 2015-05-14
Keep track of peers with maps instead of lists.

dajohi:


davecgh:
Reads well.  I'll spend some time testing it out.

davecgh:
I've been running this for a while now with no issues.

OK after rebase..

ishbir:
Just curious: why was this change necessary?

davecgh:
It isn't strictly necessary, but it's more efficient for a few reasons:

- O(1) lookups instead of O(N).  This isn't a huge savings since there generally aren't a lot of peers and the list isn't currently accessed all that frequently, but with the upcoming concurrent downloads, accessing multiple peers via the known set is going to become more frequent. 
- The items in a `list.List` are all `interface{}` so accessing the individual items required type asserting them (unboxing).  This approach avoids that overhead.
- The intent is to simply maintain a set of peers, so a list isn't the most optimal data structure for that purpose.  Also, the `struct{}` type doesn't use any memory.

So, summarizing, it uses less memory and provides faster access to the set elements.

PR #438, opened 2015-05-25
btcec: Mitigate timing attacks while using btcec.Decrypt

ishbir:
A side channel attack vector in btcec.Decrypt has been closed. bytes.Equal returns false as soon as the first byte fails to match. An attacker could guess the correct HMAC using careful timing analysis and then spoof it. Changing it to hmac.Equal, which has a constant run time, ensures that this doesn't happen.


davecgh:
OK

PR #441, opened 2015-05-27
Add new option --maxoutboundpeers.

### LINKED ISSUE #434: maxOutboundPeers should be configurable
[This](https://github.com/btcsuite/btcd/blob/master/server.go#L666) value is arbitrary and should be configurable in the same way [maxpeers](https://github.com/btcsuite/btcd/blob/master/doc.go#L35) is.
Maybe use a config flag such as --maxoutboundpeers.

dajohi:
The --maxoutboundpeers option allows the user to set the maximum
number of outbound peers to connect to.

Closes #434


davecgh:
I know you mentioned it in IRC, but adding it here for tracking purposes as well.  This need the doco updates (`doc.go`, `btcd-sample.conf`).

davecgh:
I'm closing this as there does not seem be sufficient interest nor consensus on providing an option by default for it.

As discussed here https://github.com/btcsuite/btcd/issues/434#issuecomment-103198376, often increasing this value leads to poorer performance and also has the issue of saturating full node connection slots with no increased benefits.

That said, now that the PR exists, users with specialized needs can apply it to their own forks if they desire.

PR #442, opened 2015-05-29
Update README.md install section.

davecgh:
This commit adds an additional step to the README.md install section to run the go version command and check the version so people that are installing it for the first time and ensure they are running a high enough version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,10 +48,21 @@ https://github.com/btcsuite/btcd/releases
 - Install Go according to the installation instructions here:
   http://golang.org/doc/install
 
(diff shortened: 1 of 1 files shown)


jrick:
perhaps: "the correct version" -> "a supported version"

davecgh:
Sounds good.

davecgh:
I also added a big more about checking the `GOPATH` and `GOROOT` aren't the same and `GOPATH` is set to a directory in the home directory.

jrick:
ok

PR #443, opened 2015-05-31
Fix incorrect ip connection attempt logic.

DanielKrawisz:
The comment says "only allow recent nodes (10mins) after we failed 30 times", but the server actually did the opposite and allowed only recent nodes before 30 failed connection attempts. This corrects the server's behavior.

https://golang.org/pkg/time/#Time.After 
"(t Time) After(u Time) After reports whether the time instant t is after u. "

So, previously, if now is AFTER the last connect attempt + ten minutes, in other words, if the last connect attempt was greater than ten minutes ago, don't bother trying to connect!

### DIFF
--- a/server.go
+++ b/server.go
@@ -769,8 +769,7 @@ out:
 
 			// only allow recent nodes (10mins) after we failed 30
 			// times
-			if time.Now().After(addr.LastAttempt().Add(10*time.Minute)) &&
-				tries < 30 {
+			if tries < 30 && time.Now().Sub(addr.LastAttempt()) < 10*time.Minute {
 				continue
 			}
 


dajohi:
I'd change it to:
```golang
if tries < 30 && time.Now().Sub(addr.LastAttempt()) < 10 * time.Minute {
```

DanielKrawisz:
Ok, I changed it.

dajohi:
@DanielKrawisz 
Code looks good.
Can you clean up the commit message?

```
addrmgr: Fix bug .... (50 chars)

summary here
```

See http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

davecgh:
Can we get the commit message cleaned up as suggested by @dajohi?  Also, while doing that, please rebase it to the latest master.

DanielKrawisz:
Oops! Sorry I forgot I was doing this. I'll take care of it. :)

davecgh:
Thanks for the update.  Can you rebase it to the latest master and I'll get it merged.  Please comment when you've rebased since there are no notifications otherwise.

DanielKrawisz:
Ok, I rebased with latest master and fixed it all up. Sorry to take so long!

PR #444, opened 2015-06-02
Ignore getheaders requests if not synced.

dajohi:
This mimics Bitcoin Core commit a1ba0778dd3c784046dea334e5d39f37eca264f7


davecgh:
OK

PR #446, opened 2015-06-10
Fix --onion.

dajohi:
The configured onion proxy was not being used due to checking if the
passed string had suffix '.onion', which never matched because the
port number is part of the string.


jrick:
Looks ok but I would add a precondition in the function comment that the address needs the ":port" at the end.

davecgh:
OK

PR #449, opened 2015-06-18
Notify mined transactions before connected block.

jrick:
This allows clients watching for both to know when all mined
transaction notifications for the block have been received.
Otherwise, clients would be aware that there is a new block, see the
exact same block hash/height in the next tx notifications, but never
know when all txs from the block have been received and processed.

This fixes a synchronization issue in btcwallet where the wallet would
mark itself synced to some block height before any of those blocks
transactions were processed.  If the RPC connection is lost between
sending the blockconnected notification and receiving the last
transaction notification, the wallet would not know of this and
continue with missing transactions.


jrick:
I would prefer these notifications to be grouped together, but without changing the existing API, the best we can do is reorder these and add an additional guarantee that they will always remain ordered this way.

davecgh:
OK

PR #452, opened 2015-06-20
RPC:  Add --norpcauth and change --notls

### LINKED ISSUE #451: What are the reasons for encouraging/requiring use of TLS/authentication for RPC?
When are the data transmitted over RPC actually private? Doesn't btcd really only handle publicly available data, given that it has no wallet functionality? I imagine that there are cases where it's somewhat advantageous to mask the content of transaction broadcasts, to avoid linking identities with Bitcoin addresses, but I don't see how that justifies forcing client applications to use TLS whenever they're not on the same machine (and indeed sometimes when they *are* in the same machine, but just in two different Docker containers or virtual machines, for example).

Moreover, why even handle TLS yourselves at all, instead of outsourcing that functionality to a dedicated tool like `stunnel`?

Similarly, why *require* any authentication? If all you're doing is reading from btcd's API, esp. on `localhost`, it's often just a configuration headache.

dajohi:
New option --norpcauth disables RPC authentication.  Existing option
--notls has been modified to make all listening sockets, not just
localhost, cleartext.

Closes #451

### DIFF
--- a/config.go
+++ b/config.go
@@ -84,6 +84,7 @@ type config struct {
 	RPCKey             string        `long:"rpckey" description:"File containing the certificate key"`
 	RPCMaxClients      int           `long:"rpcmaxclients" description:"Max number of RPC clients for standard connections"`
 	RPCMaxWebsockets   int           `long:"rpcmaxwebsockets" description:"Max number of RPC websocket connections"`
+	DisableRPCAuth     bool          `long:"norpcauth" description:"Disable RPC Authentication"`
 	DisableRPC         bool          `long:"norpc" description:"Disable built-in RPC server -- NOTE: The RPC server is disabled by default if no rpcuser/rpcpass or rpclimituser/rpclimitpass is specified"`
 	DisableTLS         bool          `long:"notls" description:"Disable TLS for the RPC server -- NOTE: This is only allowed if the RPC server is bound to localhost"`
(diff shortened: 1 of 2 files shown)


adamkrellenstein:
Just tested both options out, and it worked well for me.

davecgh:
As mentioned over in #451, I think this is a really, really bad idea.

Being an open source project, if the majority of the members on the team want to allow this, then I'll merge it (although doco updates would be needed before it's ready for merge).  However, I **strongly** advise against this.

History has shown that options like this always lead to guides and users who are not well versed in security matters making use of them without understanding the implications.  I would much rather enforce multi-layer security to prevent the possibility from the outset.

Rather than taking this approach, I'd rather see something like a setup guide or a setup tool which eases the process of configuring authentication and TLS.

adamkrellenstein:
If all you're worried about is users not understanding the implications, just put a big "DANGER" sign in the documentation and `--help` output. You can't force users to use secure passhrases, either.

It's absurd to say that e.g. *allowing* users to turn off TLS will result in the end of the world, when Bitcoin Core, which does the same thing as `btcd` (and has a built-in wallet!) but is much more widely used, always leaves that option off by default.

davecgh:
The team discussed this issue this morning and we've decided that we won't be merging this for reasons discussed on #451.

However, now that the patch exists, users can apply it to their own forks if they really want to ignore the advice of the btcd team.  That is one of the beautiful things about open source!  Companies and/or users are free to make modifications in their own forks.

PR #453, opened 2015-06-23
Support getmempoolinfo.

dajohi:


davecgh:
>	rpcserverhelp_test.go:17: RPC handler defined for method 'getmempoolinfo' without also specifying result types
>	rpcserverhelp_test.go:44: Failed to generate help for method 'getmempoolinfo': no result types specified for method getmempoolinfo

rpcserver.go
```diff
@@ -1938,6 +1939,24 @@ func handleGetInfo(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (in
 	return ret, nil
 }
 
+// handleGetMempoolInfo implements the getmempoolinfo command.
+func handleGetMempoolInfo(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (interface{}, error) {
+	mp := s.server.txMemPool
+	txD := mp.TxDescs()
```
davecgh: I'd combine these two lines.  The `mp` local isn't used anywhere else.

davecgh:
As Travis pointed out, the help text for the command is missing.

dajohi:
OK

davecgh:
I still need to test it out, but it reads well now.  Thanks!

davecgh:
Also, the [JSON-RPC API docs](https://github.com/btcsuite/btcd/blob/master/docs/json_rpc_api.md) need to be updated to include the new command.

dajohi:
Updated.

davecgh:
Updates look good.  Thanks.

davecgh:
Tested and confirmed it's working properly.  OK.

PR #454, opened 2015-06-26
txscript: Move lockTimeThreshold to txscript

dajohi:
Move lockTimeThreshold to txscript and export it.  This is a
consensus value which txscript will need in an upcoming diff.


davecgh:
No objection to the idea, but it shouldn't go under `standard.go`since it's a consensus value.

dajohi:
Move it to a new file, consensus.go

davecgh:
OK

dajohi:
@mably @petertodd's CHECKLOCKTIMEVERIFY

mably:
Good news, should have guessed it, thanx.

PR #455, opened 2015-06-26
txscript: Change makeScriptNum take a length argument

dajohi:
While current existing numeric opcodes are limited to 4 bytes, new
opcodes may need different limits.

This mimics Bitcoin Core commit 99088d60d8a7747c6d1a7fd5d8cd388be1b3e138

### DIFF
--- a/txscript/scriptnum.go
+++ b/txscript/scriptnum.go
@@ -8,9 +8,9 @@ const (
 	maxInt32 = 1<<31 - 1
 	minInt32 = -1 << 31
 
-	// maxScriptNumLen is the maximum number of bytes data being interpreted
-	// as an integer may be.
-	maxScriptNumLen = 4
+	// defaultScriptNumLen is the default number of bytes
+	// data being interpreted as an integer may be.
+	defaultScriptNumLen = 4
 )
 
 // scriptNum represents a numeric value used in the scripting engine with
@@ -36,9 +36,9 @@ const (
 //
 // Then, whenever data is interpreted as an integer, it is converted to this
 // type by using the makeScriptNum function which will return an error if the
-// number is out of range (or not minimally encoded depending on a flag).  Since
-// all numeric opcodes involve pulling data from the stack and interpreting it
-// as an integer, it provides the required behavior.
+// number is out of range or not minimally encoded depending on parameters.
+// Since all numeric opcodes involve pulling data from the stack and
+// interpreting it as an integer, it provides the required behavior.
 type scriptNum int64
 
 // checkMinimalDataEncoding returns whether or not the passed byte array adheres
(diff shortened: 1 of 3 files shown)


txscript/scriptnum.go
```diff
@@ -165,9 +160,9 @@ func (n scriptNum) Int32() int32 {
 // requireMinimal enabled.
 //
 // See the Bytes function documentation for example encodings.
-func makeScriptNum(v []byte, requireMinimal bool) (scriptNum, error) {
-	// Interpreting data as an integer requires that it is not larger than
-	// a 32-bit integer.
+func makeScriptNum(v []byte, requireMinimal bool, maxScriptNumLen int) (scriptNum, error) {
+	// Interpreting data requires that it is not larger than
```
davecgh: the "as an integer" part should still be retained.  Regardless of the number of bytes, it's still being interpreted as an integer.

davecgh:
I think there should be new tests added to deal with the new parameter as well.  I see that the existing tests were changed to use the default max size, however, that's only testing the single value.

In particular, I'd add tests for all variants of 5 bytes since that is how it will be used.

txscript/scriptnum.go
```diff
@@ -152,11 +156,6 @@ func (n scriptNum) Int32() int32 {
 // makeScriptNum interprets the passed serialized bytes as an encoded integer
 // and returns the result as a script number.
 //
-// Since the consensus rules dictate the serialized bytes interpreted as ints
```
davecgh: Rather than removing this important information, I'd prefer it simply be modified slightly to reflect the existence of a parameter.  Perhaps:

```Go
// Since the consensus rules dictate that serialized bytes interpreted as ints
// are only allowed to be in the range determined by the maximum number of
// bytes, an error will be returned when the provided bytes would result in a
// number outside of that range.  In particular, the range for the vast majority
// of opcodes dealing with numeric values are limited to 4 bytes and therefore
// will pass that value to this function resulting in an allowed range of
// [-2^31 + 1, 2^31 - 1].
```

txscript/scriptnum.go
```diff
@@ -165,10 +164,10 @@ func (n scriptNum) Int32() int32 {
...
 //
 // See the Bytes function documentation for example encodings.
-func makeScriptNum(v []byte, requireMinimal bool) (scriptNum, error) {
-	// Interpreting data as an integer requires that it is not larger than
-	// a 32-bit integer.
-	if len(v) > maxScriptNumLen {
+func makeScriptNum(v []byte, requireMinimal bool, scriptNumLen int) (scriptNum, error) {
+	// Interpreting data requires that it is not larger than
+	// the maximum scriptnum length or the passed scriptNumLen value.
+	if scriptNumLen > maxScriptNumLen || len(v) > scriptNumLen {
```
davecgh: The initial check against 9 can't be here.  While I agree with you that anything large than that number would cause an overflow and therefore should be invalidated, it can't be because that would disagree with the consensus rules.

txscript/scriptnum.go
```diff
@@ -8,9 +8,13 @@ const (
 	maxInt32 = 1<<31 - 1
 	minInt32 = -1 << 31
 
-	// maxScriptNumLen is the maximum number of bytes data being interpreted
-	// as an integer may be.
-	maxScriptNumLen = 4
+	// maxScriptNumLen is the maximum number of bytes
+	// data being interpreted as an integer may be.
+	maxScriptNumLen = 9
```
davecgh: There is no need for this.  As discussed below where it is used, it would break consensus.

davecgh:
As we discussed on IRC, the comments on `makeScriptNum` `scriptNum.Int32`, and `type scriptNum int64` need to updated to reflect these changes.

txscript/scriptnum.go
```diff
@@ -164,11 +168,16 @@ func (n scriptNum) Int32() int32 {
 // [0x7f 0x00 0x00 ...], etc.  All forms except [0x7f] will return an error with
 // requireMinimal enabled.
 //
+// The scriptNumLen is the size of the data to be encoded.
```
davecgh: This isn't quite right.  The number is already encoded when this function is called.  Perhaps the following?

```Go
// The scriptNumLen is the maximum number of bytes the encoded value can be
// before an ErrStackNumberTooBig is returned.  This effectivley limits the
// range of allowed values.
// WARNING:  Great care should be taken if passing a value larger than
// defaultScriptNumLen, since it could lead to addition and multiplication
// overflows.
//
```

davecgh:
OK

PR #456, opened 2015-06-26
Move IsFinalizedTransaction to txscript.

dajohi:
This change moves IsFinalizedTransaction to txscript and also changes
the first argument to take a wire.MsgTx instead of btcutil.Tx.  This
is needed for an upcoming diff in which txscript will require
IsFinalizedTransaction and we do not want to import the btcd/blockchain.

### DIFF
--- a/blockchain/validate.go
+++ b/blockchain/validate.go
@@ -127,41 +127,6 @@ func IsCoinBase(tx *btcutil.Tx) bool {
 	return IsCoinBaseTx(tx.MsgTx())
 }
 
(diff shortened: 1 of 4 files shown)


davecgh:
I vote no on this one.  Whether or not a transaction is finalized is not a function of script.

dajohi:
Yes.  txscript will need it but wanted to get around cyclic imports.  Perhaps a new subpkg is needed.

davecgh:
After discussing this in IRC a bit, we all agree that it doesn't really belong in txscript, but given the current architecture, it appears to be  the least bad option.

PR #457, opened 2015-06-27
Use NewMsgInvSizeHint when we know the size.

dajohi:
When trickling inv's to peers, allocate for the number of inv's
in the queue instead of the default 1000.  This should save on
memory.


davecgh:
OK

PR #458, opened 2015-06-29
Revert "Move IsFinalizedTransaction to txscript."

dajohi:
This reverts commit 17da2ba7fa691c666d70948217ffcd5b6ac148f8.

This was done prematurely.  This will be revisited when a code
restructure is more urgent.


davecgh:
OK

PR #461, opened 2015-07-06
Implement getblockheader RPC.

dajohi:
This mimics Bitcoin Core commit 076badb60f33f0c32b035de220ca14c52a423a2a

### DIFF
--- a/btcjson/chainsvrcmds_test.go
+++ b/btcjson/chainsvrcmds_test.go
@@ -200,6 +200,20 @@ func TestChainSvrCmds(t *testing.T) {
 			unmarshalled: &btcjson.GetBlockHashCmd{Index: 123},
 		},
 		{
+			name: "getblockheader",
+			newCmd: func() (interface{}, error) {
+				return btcjson.NewCmd("getblockheader", "123")
+			},
+			staticCmd: func() interface{} {
+				return btcjson.NewGetBlockHeaderCmd("123", nil)
+			},
+			marshalled: `{"jsonrpc":"1.0","method":"getblockheader","params":["123"],"id":1}`,
+			unmarshalled: &btcjson.GetBlockHeaderCmd{
+				Hash:    "123",
+				Verbose: btcjson.Bool(true),
+			},
+		},
+		{
 			name: "getblocktemplate",
 			newCmd: func() (interface{}, error) {
 				return btcjson.NewCmd("getblocktemplate")
@@ -926,6 +940,7 @@ func TestChainSvrCmds(t *testing.T) {
 			t.Errorf("Test #%d (%s) unexpected marshalled data - "+
 				"got %s, want %s", i, test.name, marshalled,
 				test.marshalled)
+			t.Errorf("\n%s\n%s", marshalled, test.marshalled)
 			continue
 		}
 
--- a/docs/json_rpc_api.md
+++ b/docs/json_rpc_api.md
@@ -157,27 +157,28 @@ the method name for further details such as parameter and return information.
 |7|[getblock](#getblock)|Y|Returns information about a block given its hash.|
 |8|[getblockcount](#getblockcount)|Y|Returns the number of blocks in the longest block chain.|
 |9|[getblockhash](#getblockhash)|Y|Returns hash of the block in best block chain at the given height.|
-|10|[getconnectioncount](#getconnectioncount)|N|Returns the number of active connections to other peers.|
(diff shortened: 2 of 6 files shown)


btcjson/chainsvrcmds_test.go
```diff
@@ -200,6 +200,18 @@ func TestChainSvrCmds(t *testing.T) {
 			unmarshalled: &btcjson.GetBlockHashCmd{Index: 123},
 		},
 		{
+			name: "getblockheader",
+			newCmd: func() (interface{}, error) {
+				return btcjson.NewCmd("getblockheader", "blockHash")
```
davecgh: Can you make the hash string "123" so it more closely matches `getblock` and is a valid hash?

docs/json_rpc_api.md
```diff
@@ -157,27 +157,28 @@ the method name for further details such as parameter and return information.
...
-|24|[ping](#ping)|N|Queues a ping to be sent to each connected peer.|
-|25|[sendrawtransaction](#sendrawtransaction)|Y|Submits the serialized, hex-encoded transaction to the local peer and relays it to the network.<br /><font color="orange">btcd does not yet implement the `allowhighfees` parameter, so it has no effect</font>|
-|26|[setgenerate](#setgenerate) |N|Set the server to generate coins (mine) or not.<br/>NOTE: Since btcd does not have the wallet integrated to provide payment addresses, btcd must be configured via the `--miningaddr` option to provide which payment addresses to pay created blocks to for this RPC to function.|
-|27|[stop](#stop)|N|Shutdown btcd.|
-|28|[submitblock](#submitblock)|Y|Attempts to submit a new serialized, hex-encoded block to the network.|
-|29|[validateaddress](#validateaddress)|Y|Verifies the given address is valid.  NOTE: Since btcd does not have a wallet integrated, btcd will only return whether the address is valid or not.|
-|30|[verifychain](#verifychain)|N|Verifies the block chain database.|
+|10|[getblockheader](#getblockheader)|Y|Returns the blockheader of the block.|
```
davecgh: `block header` should be two words imo.

rpcserver.go
```diff
@@ -1089,6 +1090,71 @@ func handleGetBlockHash(s *rpcServer, cmd interface{}, closeChan <-chan struct{}
...
+			MerkleRoot:    msgBlock.Header.MerkleRoot.String(),
+			NextHash:      shaNextStr,
+			PreviousHash:  msgBlock.Header.PrevBlock.String(),
+			Nonce:         uint64(msgBlock.Header.Nonce),
+			Time:          msgBlock.Header.Timestamp.Unix(),
+			Bits:          strconv.FormatInt(int64(msgBlock.Header.Bits), 16),
+			Difficulty:    getDifficultyRatio(msgBlock.Header.Bits),
+		}
+		return blockHeaderReply, nil
+	} else {
```
davecgh: As Travis correctly pointed out, there is no need for the else here since the if block above ends with a return.

rpcserver.go
```diff
@@ -1089,6 +1090,71 @@ func handleGetBlockHash(s *rpcServer, cmd interface{}, closeChan <-chan struct{}
...
+// handleGetBlockHeader implements the getblockheader command.
+func handleGetBlockHeader(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (interface{}, error) {
+	c := cmd.(*btcjson.GetBlockHeaderCmd)
+
+	sha, err := wire.NewShaHashFromStr(c.Hash)
+	if err != nil {
+		return nil, err
+	}
+
+	if *c.Verbose {
```
davecgh: The `Verbose` flag is optional, so this needs to be checked for nil before dereferencing it.

`if c.Verbose == nil || *c.Verbose {`

rpcserver.go
```diff
@@ -1089,6 +1090,71 @@ func handleGetBlockHash(s *rpcServer, cmd interface{}, closeChan <-chan struct{}
...
+		blkHeader, err := s.server.db.FetchBlockHeaderBySha(sha)
+		if err != nil {
+			return nil, &btcjson.RPCError{
+				Code:    btcjson.ErrRPCInvalidAddressOrKey,
+				Message: "Invalid address or key: " + err.Error(),
+			}
+		}
+
+		buf := bytes.NewBuffer(make([]byte, 0, wire.MaxBlockHeaderPayload))
+		err = blkHeader.Serialize(buf)
```
davecgh: This should be using `BtcEncode` instead of `Serialize`.  While the two encodings are the same at the moment, the RPC call should return the wire encoded block-header, not the one encoded for database storage.

davecgh:
Looks good with latest changes.

OK

PR #466, opened 2015-07-20
switch maxDataCarrierSize to public const

bclermont:


bclermont:
Ref https://github.com/btcsuite/btcd/commit/005b5408958d2704f0bcc2d24796dc6ebab5d934#commitcomment-12246583

davecgh:
OK

dajohi:
OK.

PR #467, opened 2015-07-22
Implmement BlockHeader BtcEncode/BtcDecode.

dajohi:
At the current time, there is no difference between the wire encoding
at protocol version 0 and the stable long-term storage format.  These
methods are simply for consistency with the other types.

### DIFF
--- a/wire/blockheader_test.go
+++ b/wire/blockheader_test.go
@@ -50,6 +50,7 @@ func TestBlockHeader(t *testing.T) {
 // protocol versions.
 func TestBlockHeaderWire(t *testing.T) {
 	nonce := uint32(123123) // 0x1e0f3
+	pver := uint32(70001)
 
 	// baseBlockHdr is used in the various tests as a baseline BlockHeader.
 	bits := uint32(0x1d00ffff)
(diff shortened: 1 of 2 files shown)


wire/blockheader_test.go
```diff
@@ -156,6 +156,81 @@ func TestBlockHeaderWire(t *testing.T) {
 	}
 }
 
+// TestBlockHeaderBtcEncode tests BlockHeader encode and decode.
+func TestBlockHeaderBtcEncode(t *testing.T) {
```
davecgh: I'd prefer not having a separate test func here.  Just add the relevant encode/decode bits to `TestBlockHeaderWire`.  Also, the name doesn't match what the function does anyways since it's testing both encode/decode.

For example, after the call to

```Go
err := wire.TstWriteBlockHeader(&buf, test.pver, test.in)
...
if !bytes.Equal(buf.Bytes(), test.buf) {
...
```

Add:

```Go
+		err = test.in.BtcEncode(&buf, pver)
+		if err != nil { ... }
+		if !bytes.Equal(buf.Bytes(), test.buf) { ... }
```

Then the same deal for the decode.

davecgh:
Looks good now.  OK

PR #468, opened 2015-07-22
wire: Update tests to force error in tx lock time.

davecgh:
This pull request updates the wire tests for transactions which force serialization and deserialization errors to force an error in the the transaction lock time path.

This brings the wire test coverage back up to 100%.


dajohi:
OK

PR #469, opened 2015-07-23
Removed unnecessary loop in calcSignatureHash.

jongillham:
Not completely sure why this loop was here in the first place as it appears to do nothing. It has been here since the first commit though (https://github.com/btcsuite/btcscript/blob/927d846f067df1ef6f4dbc615fe592a7c2d4fd33/script.go#L640).

### DIFF
--- a/wire/blockheader.go
+++ b/wire/blockheader.go
@@ -56,6 +56,22 @@ func (h *BlockHeader) BlockSha() ShaHash {
 	return DoubleSha256SH(buf.Bytes())
 }
 
+// BtcDecode decodes r using the bitcoin protocol encoding into the receiver.
+// This is part of the Message interface implementation.
+// See Deserialize for decoding block headers stored to disk, such as in a
+// database, as opposed to decoding block headers from the wire.
+func (h *BlockHeader) BtcDecode(r io.Reader, pver uint32) error {
+	return readBlockHeader(r, pver, h)
+}
+
+// BtcEncode encodes the receiver to w using the bitcoin protocol encoding.
+// This is part of the Message interface implementation.
+// See Serialize for encoding block headers to be stored to disk, such as in a
+// database, as opposed to encoding block headers for the wire.
+func (h *BlockHeader) BtcEncode(w io.Writer, pver uint32) error {
+	return writeBlockHeader(w, pver, h)
+}
+
 // Deserialize decodes a block header from r into the receiver using a format
 // that is suitable for long-term storage such as a database while respecting
 // the Version field.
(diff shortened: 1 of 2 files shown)


davecgh:
I'm guessing you figured out why since you closed it, but it's required.

jongillham:
I closed the pull request because Travis failed but I can't actually see why.

jongillham:
I'm just going to play around with this to figure out why it doesn't build with Travis.

jongillham:
It appears the build failed first time round because Travis CI stalled trying to download a package, not because the unit tests failed.

davecgh:
I'll take a look at this more closely over the weekend.  At a quick glance, I believe it should be fine, but I want to spend some time giving it a proper analysis and testing.  I think it's a hold over from before `MsgTx` had a deep copy method.

The important part is that calculating a signature hash for a transaction requires modifying that transaction.  Thus, a copy is needed to avoid modifying the original.

davecgh:
As an update to this, I spent some time looking at it and agree that there is no reason it should need to do the copy again since I verified the `tx.Copy` method does indeed copy it.  Nice catch!

In fact, the inputs just above it also don't need to be copied again either since they were copied by the `tx.Copy` call as well.  Since you're already submitting a change for this, would you mind removing those too and updating the commit/pr messages?

In particular, I'm referring to this:

```diff
diff --git a/txscript/script.go b/txscript/script.go
index 333774c..a811680 100644
--- a/txscript/script.go
+++ b/txscript/script.go
@@ -300,9 +300,6 @@ func calcSignatureHash(script []parsedOpcode, hashType SigHashType, tx *wire.Msg
        // inputs that are not currently being processed.
        txCopy := tx.Copy()
        for i := range txCopy.TxIn {
-               var txIn wire.TxIn
-               txIn = *txCopy.TxIn[i]
-               txCopy.TxIn[i] = &txIn
                if i == idx {
                        // UnparseScript cannot fail here because removeOpcode
                        // above only returns a valid script.
```

I also did some additional manual testing to prove it's working as intended with both changes.

davecgh:
Also, since you'll be modifying the commit anyways, please include `txscript:` at the start of it.

Something like: `txscript: Remove unneeded signature hash copies.` would be ideal so it conforms to the [model git commit message format](https://web.archive.org/web/20100308125500/http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).

jongillham:
Changes discussed here were consolidated in pull request https://github.com/btcsuite/btcd/pull/475.

PR #471, opened 2015-07-24
btcctl: Add getwalletinfo command

drazisil:
Adds wallet server command getwalletinfo


davecgh:
Thanks for the PR.  Looks good.

OK

dajohi:
OK..  but can you add the Result too?

See https://github.com/bitcoin/bitcoin/blob/master/src/wallet/rpcwallet.cpp#L2184-L2224

dajohi:
but OK on this diff :)

davecgh:
The result isn't needed for btcctl.  Once `btcwallet` implements the command, the result can be added then.

PR #473, opened 2015-07-27
docs: Add info describing model commit messages.

davecgh:


jcvernaleo:
Looks good to me.  OK

PR #474, opened 2015-07-28
peer: Optimize merkle block handling.

davecgh:
This pull request updates the merkle block handling to for the latest changes to the `btcutil` API and optimizes it along the way.

Previously, the code was inefficiently reloading the transactions for the matched hashes from the database instead of simply pulling them from the full block that was used to create the merkle block.

Depends on btcsuite/btcutil#55


dajohi:
OK

PR #475, opened 2015-07-28
txscript: Remove unneeded signature hash copies

jongillham:
These copies were likely left over from when MsgTx had no deep copy
functionality.

### DIFF
--- a/txscript/script.go
+++ b/txscript/script.go
@@ -300,9 +300,6 @@ func calcSignatureHash(script []parsedOpcode, hashType SigHashType, tx *wire.Msg
 	// inputs that are not currently being processed.
 	txCopy := tx.Copy()
 	for i := range txCopy.TxIn {
-		var txIn wire.TxIn
-		txIn = *txCopy.TxIn[i]
-		txCopy.TxIn[i] = &txIn
 		if i == idx {
 			// UnparseScript cannot fail here because removeOpcode
 			// above only returns a valid script.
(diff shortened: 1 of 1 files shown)


jongillham:
This follows on from pull request https://github.com/btcsuite/btcd/pull/469.

davecgh:
Thanks @jongillham .  For future reference, you can simply update an existing pr by force pushing versus closing and opening a new one which makes it easier to track history.  See [ Managing a btcd Project for Code Contibutions ](https://gist.github.com/davecgh/04cb601a147f8c831ca5) for more details pull request management.

As for the PR itself, as previously commented on #469, I spent some time looking at it and agree that there is no reason it needs to do the copies again since I verified the `tx.Copy` method does indeed copy both the inputs and outputs.  I also did some additional manual testing to prove it's working as intended with both changes.

OK

dajohi:
OK

jongillham:
Cheers @davecgh I managed to mangle my commits up by synching to the head of `btcsuite/btcd` and then trying to squash commits together. In the end I managed to accidentally 'auto close' the previous PR so created a new one instead. Definitely a nascence for tracking history.

jongillham:
@davecgh the merge seems to have caused `goclean.sh` to fail. Any ideas?

davecgh:
It's due to a change to `btcutil` which the next commit to btcd make use of.  I merged it too quickly before travis was done building this pr.  It's not a real error.

PR #476, opened 2015-07-28
rpcserver: Allow tx result creation without block.

davecgh:
This pull request modifies the `createTxRawResult` code path along with callers to work with block headers as opposed to `btcutil.Block`s.  This in turn allows the code in `handleGetRawTransaction` and `handleSearchRawTransactions` to perform a much cheaper block header load as opposed to a full block load.

While here, also very slightly optimize the `createVinList` function to avoid creating a `util.Tx` wrapper and to take advantage of the `btcutil.Amount` type added after the function was originally written


jrick:
ok

davecgh:
@dajohi: Moved

PR #477, opened 2015-07-29
txscript: Correct SignTxOutput documentation.

jongillham:


davecgh:
Hmmm.  I'm pretty sure the comment is accurate as it is.  You are signing the output of an origin transaction in order to generate a signature script that can be used as an input to a new redeeming transaction.

jongillham:
I am pretty sure you are correct on reflection. I mistook its different use to the `idx` param on `SignatureScript`.

jongillham:
Sorry @davecgh I am still having trouble with this. (I know it is fundamental to understanding transaction signing.) I still can't figure out why "SignTxOutput signs *output* idx of the given tx". I don't see why it needs to know the output of the origin transaction, only the pkScript. However, it does need to know the input `idx` of the redeeming transaction especially for `SigHashNone` and `SigHashSingle`.

davecgh:
I think the issue here is that the index is really used for both depending on the signature hash type.  I misspoke a bit in my comment above.  It's not the output of the origin transaction, rather both the input and the output of the redeeming transaction that is being passed.

The reason for this is that there are different signing types as specified by the signature hash type.  When you sign a transaction, you are really just signing a hash that is generated by a specific serialization of some combination of inputs and outputs in the redeeming transaction in conjunction with your specific input concatenated with the origin public key script.  The signature hash type controls how it's serialized therefore effectively giving some measure of control over which inputs and outputs you are signing in the redeeming transaction.

It would probably be instructive to look at `calcSignatureHash` to see how the transaction is transformed according to the signature hash type for more details.  In short though, the default type of `SigHashAll` specifies that all outputs should be signed, `SigHashNone` means none of the outputs should be signed (i.e the signer doesn't care where the Bitcoins go), and `SigHashSingle` specifies only a single output should be signed (i.e the signer only cares about where a single output goes).  There is also the `SigHashAnyOneCanPay` flag which specifies that  that signer does not care where the other inputs to the transaction come from, so it essentially allows other people to add inputs.  Heed special attention to the comments in `calcSignatureHash` regarding `SigHashSingle`.

In my opinion, the specific mechanism used for the signature hash types was not particularly well thought out when implemented in the original Satoshi client (not even including the `SigHashSingle` bug that is still problematic), but they are unfortunately part of the consensus now, so they can't be changed without a hard fork.

jongillham:
Thank you for the thorough explanation, it's really appreciated!

So shouldn't the documentation for SignTxOutput be

> SignTxOutput creates a signature script of type hashType for input idx of tx to resolve the script given in pkScript...

instead of

> SignTxOutput signs output idx of the given tx to resolve the script given in pkScript with a signature type of hashType...

The latter suggests that only output idx of tx is being signed when this really all depends on the hashType.

Really sorry if I am being dense!

davecgh:
I agree that the comment could be improved, though I'm not sure it makes it any better to say input instead of output because both are slightly wrong and could be misleading.  However, I don't have any particular objections to your rewording though.

jongillham:
Cheers @davecgh. I won't change anything yet because don't seem to understand something crucial if changing 'output' to 'input' still makes `SignTxOutput` description slightly wrong. (I am assuming when we talk about input and outputs we are talking about `TxIn`'s and `TxOut`s.)

Maybe if I explain how I understand `SignTxOutput` to work, you can point out where my assumptions are wrong:

I pass in a `tx` to `SignTxOutput` and get back a script that can *only* be used for `tx.TxIn[idx].SignatureScript` to enable me to spend some previous transaction's `TxOut` `pkScript`. The signage process and how much of `tx` is included in the signature depends on `hashType`. For example `SigHashAll` hashes all of `tx.TxIn` and `tx.TxOut`.

In fact it makes more sense to me if `SignTxOutput` was renamed `SignTx` because I don't know what the `Output` part of `SignTxOutput` has to do with anything. Therefore my comment would start:

> SignTx creates a signature script of type hashType for input idx of tx to resolve the script given in pkScript...

Thanks for guiding me through this!

davecgh:
Your understanding is correct.  The reason I said that 'input' is still slightly wrong is for the case of `SigHashSingle`.  In that case, it acts as both the input index and the output index.

I'm not really 100% sure why it was named that since I wasn't directly involved with the addition of the signing code (essentially all of the code that is now in `sign.go`).  However, I would guess it's called that because the action of spending coins is proving you are the owner of the coins "sent" by an _output_ of the origin transaction.

That said, I completely agree with you that the name is confusing given what really ends up happening.  In fact, one of the things on my ever growing list is to give `sign.go` a good polishing as I've noticed before some of the variable naming in the various functions could be made clearer (e.g. `originPkScript` instead of `subscript` and `inputIdx` instead of `idx`), there are several places efficiency can be improved through signature caching, and a few other things.  I doubt I'll have time to get to that anytime soon, though, so if you have time and desire to take a crack at it, we'd certainly appreciate the help.

jongillham:
> Your understanding is correct.

Phew, I am glad that's the case and will send a pull request with what I feel is a more appropriate name and comment for `SignTxOutput`.

I would be happy to preen the code – in fact I find it hard to resist if see something that bugs me. :smile:

PR #478, opened 2015-08-01
txscript:  New function IsUnspendable

dajohi:
IsUnspendable takes a public key script and returns whether it is
spendable.

Additionally, hook this into the mempool isDust function, since
unspendable outputs can't be spent.

This mimics Bitcoin Core commit 0aad1f13b2430165062bf9436036c1222a8724da

### DIFF
--- a/mempool.go
+++ b/mempool.go
@@ -107,6 +107,11 @@ type txMemPool struct {
 // relay fee.  In particular, if the cost to the network to spend coins is more
 // than 1/3 of the minimum transaction relay fee, it is considered dust.
 func isDust(txOut *wire.TxOut) bool {
+	// Unspendable outputs are considered dust.
+	if txscript.IsUnspendable(txOut.PkScript) {
+		return true
+	}
+
 	// The total serialized size consists of the output and the associated
 	// input script to redeem it.  Since there is no input script
 	// to redeem it yet, use the minimum size of a typical input script.
--- a/txscript/script.go
+++ b/txscript/script.go
@@ -458,3 +458,15 @@ func GetPreciseSigOpCount(scriptSig, scriptPubKey []byte, bip16 bool) int {
 	shPops, _ := parseScript(shScript)
 	return getSigOpCount(shPops, true)
 }
+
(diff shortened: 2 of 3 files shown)


mempool.go
```diff
@@ -107,6 +107,11 @@ type txMemPool struct {
 // relay fee.  In particular, if the cost to the network to spend coins is more
 // than 1/3 of the minimum transaction relay fee, it is considered dust.
 func isDust(txOut *wire.TxOut) bool {
+	// Unspendable outputs are considered dust.
+	if txscript.IsUnspendable(txOut.ScriptPubKey) {
```
davecgh: `txOut.PkScript`

txscript/script.go
```diff
@@ -458,3 +458,15 @@ func GetPreciseSigOpCount(scriptSig, scriptPubKey []byte, bip16 bool) int {
 	shPops, _ := parseScript(shScript)
 	return getSigOpCount(shPops, true)
 }
+
+// IsUnspendable returns whether the passed scriptPubKey is unspendable, or
+// guaranteed to fail at execution.  This allows inputs to be pruned instantly
+// when entering the UTXO set.
+func IsUnspendable(scriptPubKey []byte) bool {
```
davecgh: Please make the variable name match the rest of the code base.  In particular, it's referred to as `pkScript`.

txscript/script.go
```diff
@@ -458,3 +458,15 @@ func GetPreciseSigOpCount(scriptSig, scriptPubKey []byte, bip16 bool) int {
 	shPops, _ := parseScript(shScript)
 	return getSigOpCount(shPops, true)
 }
+
+// IsUnspendable returns whether the passed scriptPubKey is unspendable, or
```
davecgh: You'll need to fix the reference to the variable name too when you update it.  This is why I don't generally like putting variable names in the comments.  It's better to simply describe what the function does.

dajohi:
Updated.

davecgh:
OK

PR #479, opened 2015-08-02
Make PubKey variable names consistent.

jongillham:


jongillham:
I don't suffer with OCD or anything but this was bothering me! ;-)

In a similar vein I noticed that some defined errors use title case and some don't e.g. https://github.com/btcsuite/btcd/blob/master/txscript/error.go. It seems that idiomatic Go (i.e. the standard library) generally uses all lower case and sometimes prefixes the error with a package name e.g. https://golang.org/pkg/crypto/dsa/#pkg-variables.

davecgh:
OK.  I also agree that the error strings should be all lowercase so they can be concatenated into other errors more easily.

PR #483, opened 2015-08-09
txscript: Rename SignTxOutput to SignTx and improve comment.

jongillham:
Note that this is an API breaking change and will make the `btcwallet` build fail so it will need to be coordinated with an appropriate `btcwallet` change.

Along with this commit I would like to make `txscript.RawTxInSignature` and `txscript.SignatureScript` private functions because in my opinion they complicate the API and as far as I can see add nothing that `txscript.SignTxOutput` (now `txscript.SignTx`) don't already do.

There is probably a reason why `RawTxInSignature` and `SignatureScript` are public but I can't see one. So I though I would raise the issue here before I add a couple of commits to this pull request making them private.

In any case, you might not want to change the API at this point, in which case I will revert back to just the `txscript: Make error strings idiomatic.` commit 27f7f8235518af9c7f1bdc99db0cf81d2a9a1a3d.

### DIFF
--- a/txscript/error.go
+++ b/txscript/error.go
@@ -28,11 +28,11 @@ var (
 
 	// ErrStackOpDisabled is returned when a disabled opcode is encountered
 	// in the script.
-	ErrStackOpDisabled = errors.New("Disabled Opcode")
+	ErrStackOpDisabled = errors.New("disabled opcode")
 
 	// ErrStackVerifyFailed is returned when one of the OP_VERIFY or
 	// OP_*VERIFY instructions is executed and the conditions fails.
-	ErrStackVerifyFailed = errors.New("Verify failed")
+	ErrStackVerifyFailed = errors.New("verify failed")
 
 	// ErrStackNumberTooBig is returned when the argument for an opcode that
 	// should be an offset is obviously far too large.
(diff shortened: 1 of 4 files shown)


davecgh:
I haven't reviewed this yet, but I did want to comment in regards to `txscript.RawTxInSignature` and `txscript.SignatureScript` that they definitely need to be exported because not all users will be dealing with standard transaction types and those functions provide a mechanism to do non-standard things.  For example, the voting pool functionality in btcwallet uses `txscript.RawTxInSignature`.

jongillham:
@davecgh good point, I wonder if `txscript.RawTxInSignature` and `txscript.SignatureScript` could/should be consolidated into `txscript.SignTx`?

jongillham:
If you think `originPkScript` and `inputIdx` are appropriate variable names it would be good for me to change all the variable names in other functions to be consistent.

davecgh:
I'd rather keep the functions separate and exported.  They are specifically exported to allow callers working with non-standard script and transaction types of which the `SignTx` can't possibly understand all of.

As far as the `originPkScript` and `inputIdx`, those sound good to me. You'll want to be careful when updating them though as some of them have to work with subscript as opposed to the raw `originPkScript`, so a blanket copy and paste won't have the desired results.

PR #445, opened 2015-06-03
[peer] Refactor peer code into its own package

tuxcanfly:
This pull request introduces package peer which contains peer related features
refactored from `peer.go`.

The following is an overview of the features the package provides:

- Provides a basic concurrent safe bitcoin peer for handling bitcoin
    communications via the peer-to-peer protocol
- Full duplex reading and writing of bitcoin protocol messages
- Automatic handling of the initial handshake process including protocol
    version negotiation
- Automatic periodic keep-alive pinging and pong responses
- Asynchronous message queuing of outbound messages with optional
    channel for notification when the message is actually sent
- Inventory message batching and send trickling with known inventory
    detection and avoidance
- Ability to wait for shutdown/disconnect
- Flexible peer configuration
    - Caller is responsible for creating outgoing connections and listening
    for incoming connections so they have flexibility to establish
    connections as they see fit (proxies, etc.)
    - User agent name and version
    - Bitcoin network
    - Service support signalling (full nodes, bloom filters, etc.)
    - Maximum supported protocol version
    - Ability to register callbacks for handling bitcoin protocol messages
- Proper handling of bloom filter related commands when the caller does
    not specify the related flag to signal support
    - Disconnects the peer when the protocol version is high enough
    - Does not invoke the related callbacks for older protocol versions
- Snapshottable peer statistics such as the total number of bytes read
    and written, the remote address, user agent, and negotiated protocol
    version
- Helper functions for pushing `addresses`, `getblocks`, `getheaders`, and
    `reject` messages
    - These could all be sent manually via the standard message output
    function, but the helpers provide additional nice functionality such
    as duplicate filtering and address randomization
- Full documentation with example usage
- Test coverage

In addition to the addition of the new package, `btcd` has been refactored
to make use of the new package by extending the basic peer it provides to
work with the `blockmanager` and `server` to act as a full node.  The
following is a broad overview of the changes to integrate the package:

- The `server` is responsible for all connection management including
    persistent peers and banning
- Callbacks for all messages that are required to implement a full node
    are registered
- Logic necessary to serve data and behave as a full node is now in the
    callback registered with the peer

Finally, the following peer-related things have been improved as a part
of this refactor:

- Don't log or send reject message due to peer disconnects
- Remove trace logs that aren't particularly helpful
- Finish an old TODO to switch the queue `WaitGroup` over to a channel
- Improve various comments and fix some code consistency cases
- Improve a few logging bits
- Implement a most-recently-used nonce tracking for detecting self
[...]

PR #484, opened 2015-08-13
Keep track of recently rejected transactions.

dajohi:
This prevents the node from repeatedly requesting and rejecting the
same transaction as different peers inv the same transaction.

Idea from Bitcoin Core commit 0847d9cb5fcd2fdd5a21bde699944d966cf5add9.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/484)

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -6,6 +6,8 @@ package main
 
 import (
 	"container/list"
+	"crypto/rand"
+	"math/big"
 	"net"
 	"os"
 	"path/filepath"
@@ -32,6 +34,18 @@ const (
 	// database type is appended to this value to form the full block
 	// database name.
 	blockDbNamePrefix = "blocks"
+
+	// maxRejectedTxns is the maximum number of rejected transactions
+	// shas to store in memory.
+	maxRejectedTxns = 1000
+
+	// maxRequestedBlocks is the maximum number of requested block
+	// shas to store in memory.
+	maxRequestedBlocks = wire.MaxInvPerMsg
+
+	// maxRequestedTxns is the maximum number of requested transactions
+	// shas to store in memory.
+	maxRequestedTxns = wire.MaxInvPerMsg
 )
 
 // zeroHash is the zero value hash (all zeros).  It is defined as a convenience.
@@ -189,6 +203,7 @@ type blockManager struct {
 	started           int32
 	shutdown          int32
 	blockChain        *blockchain.BlockChain
+	rejectedTxns      map[wire.ShaHash]struct{}
 	requestedTxns     map[wire.ShaHash]struct{}
 	requestedBlocks   map[wire.ShaHash]struct{}
 	progressLogger    *blockProgressLogger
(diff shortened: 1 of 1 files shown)


blockmanager.go
```diff
@@ -1069,6 +1106,40 @@ func (b *blockManager) handleInvMsg(imsg *invMsg) {
 	}
 }
 
+// limitNumRejectedTxns limits the number of rejected transactions by evicting
+// a random rejected transaction if adding a new one would cause it to overflow
+// the max allowed.
+func (b *blockManager) limitNumRejectedTxns() error {
```
davecgh: I'd prefer this be defined above the only function it's called from so it's consistent with the vast majority of the code base.

davecgh:
Other than my comment this looks good.  I'll do some manual testing on it as well.

dajohi:
I will wait on this as I'd like to add limits to requestedTxns and requestedBlocks as well.

dajohi:
I wonder if this should wait til after the #445?

blockmanager.go
```diff
@@ -479,11 +496,20 @@ func (b *blockManager) handleTxMsg(tmsg *txMsg) {
 	// already knows about it and as such we shouldn't have any more
 	// instances of trying to fetch it, or we failed to insert and thus
 	// we'll retry next time we get an inv.
-	txHash := tmsg.tx.Sha()
 	delete(tmsg.peer.requestedTxns, *txHash)
 	delete(b.requestedTxns, *txHash)
 
 	if err != nil {
+		// Do not request this transactions again until a new block
```
wallclockbuilder: We're working with one tx here. Update comment to reflect. ~~transactions~~ transaction.

davecgh:
@dajohi: I know this is waiting on #445 to go in, but would you please address the comments and rebase it on top of 445.  I suspect that one should be going in within a week or so and I'd like to have this one ready.  Thanks!

davecgh:
#445 has been merged.  Please update this accordingly.

dajohi:
rebased.

blockmanager.go
```diff
@@ -1072,6 +1141,40 @@ func (b *blockManager) handleInvMsg(imsg *invMsg) {
...
 }
 
+// limitMap is a helper function for maps that require a maximum limit by
+// evicting a random rejected transaction if adding a new value would cause it
+// to overflow the maximum allowed.
+func (b *blockManager) limitMap(m map[wire.ShaHash]struct{}, limit int) error {
+	if len(m)+1 > limit {
+		// Generate a cryptographically random hash.
+		randHashBytes := make([]byte, wire.HashSize)
+		_, err := rand.Read(randHashBytes)
```
Roasbeef: So, I'm not sure we really need strong _cryptographic_ randomness here. I suspect reading from `/dev/urandom` may become an unnecessary bottleneck when processing transactions. In order to speed up the execution of this function, we may want to instead use a fast PRNG such as [xorshift](https://en.wikipedia.org/wiki/Xorshift). There exists a pure-go implementation, which can be found [here](https://github.com/lazybeaver/xorshift). 

NOTE: I haven't benchmarked this yet, but this is a recommendation based on past experience when optimizing the `SigCache` in #598.

davecgh: Yeah, you have a fair point and I think that's probably a reasonable place to look at for optimizing.

However, given this is really just a refactor of the existing code, I don't suspect it should hold up this PR.

Issue #448, opened 2015-06-15
Address indexing does not gracefully shutdown


carryforward:
I am running code tagged 0.11.1.  The blockchain is fully synced.  The address index had been dropped, due to upgrading from v.10.  

When I type ctrl+c btcd starts to shut down.  It doesn't look like the address indexing subsystem gets this shutdown message.  It continues to try to index blocks after shutdown starts.
```
15:56:58 2015-06-15 [INF] ADXR: Indexed addresses of 1 block in the last 30.9s (1369 transactions, height 312115, 2014-07-23 07:06:19 -0500 CDT)
15:57:10 2015-06-15 [INF] ADXR: Indexed addresses of 4 blocks in the last 12.35s (3045 transactions, height 312119, 2014-07-23 08:15:59 -0500 CDT)
15:57:21 2015-06-15 [INF] ADXR: Indexed addresses of 5 blocks in the last 10.11s (3255 transactions, height 312124, 2014-07-23 09:18:41 -0500 CDT)
15:57:42 2015-06-15 [INF] ADXR: Indexed addresses of 4 blocks in the last 21.61s (2133 transactions, height 312128, 2014-07-23 10:00:02 -0500 CDT)
15:57:55 2015-06-15 [INF] ADXR: Indexed addresses of 8 blocks in the last 12.95s (3497 transactions, height 312136, 2014-07-23 11:14:03 -0500 CDT)
15:58:45 2015-06-15 [INF] ADXR: Indexed addresses of 2 blocks in the last 50.16s (1675 transactions, height 312138, 2014-07-23 11:42:17 -0500 CDT)
15:58:56 2015-06-15 [INF] ADXR: Indexed addresses of 15 blocks in the last 10.55s (8085 transactions, height 312153, 2014-07-23 13:59:52 -0500 CDT)
15:59:07 2015-06-15 [INF] ADXR: Indexed addresses of 6 blocks in the last 11.55s (3105 transactions, height 312159, 2014-07-23 15:01:56 -0500 CDT)
^C15:59:18 2015-06-15 [INF] BTCD: Received SIGINT (Ctrl+C).  Shutting down...
15:59:18 2015-06-15 [INF] BTCD: Gracefully shutting down the server...
15:59:18 2015-06-15 [WRN] SRVR: Server shutting down
15:59:18 2015-06-15 [WRN] RPCS: RPC server shutting down
15:59:18 2015-06-15 [INF] RPCS: RPC server shutdown complete
[...]

PR #490, opened 2015-08-25
Logging of address indexer on shutdown

tuxcanfly:
Refs #448 

The index catchup worker goroutines might error out on shutdown because the db is closed. Updated to only log the error if we're not shutting down.

### DIFF
--- a/chainindexer.go
+++ b/chainindexer.go
@@ -400,9 +400,11 @@ out:
 		case indexJob := <-workChan:
 			addrIndex, err := a.indexBlockAddrs(indexJob.blk)
 			if err != nil {
-				adxrLog.Errorf("Unable to index transactions of"+
-					" block: %v", err)
-				a.server.Stop()
+				// Only log the error if we're not forcibly shutting down.
+				if atomic.LoadInt32(&a.shutdown) == 0 {
+					adxrLog.Errorf("Unable to index transactions of"+
+						" block: %v", err)
+				}
 				break out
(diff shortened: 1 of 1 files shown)


chainindexer.go
```diff
@@ -400,9 +400,11 @@ out:
 		case indexJob := <-workChan:
 			addrIndex, err := a.indexBlockAddrs(indexJob.blk)
 			if err != nil {
-				adxrLog.Errorf("Unable to index transactions of"+
-					" block: %v", err)
-				a.server.Stop()
```
Roasbeef: Why remove the server shutdown?

tuxcanfly: IIRC it was already being shutdown from other goroutines.

Roasbeef: The other critical shutdowns are triggered by distinct error conditions. I think the server shutdown should be left in place, but within this new "are we shutting down" conditional. 

Additionally, this "are we already shutting down" guard of error logging should be added to the other areas within the `chainindexer`that may trigger a critical shutdown. The extraneous error messages are harmless, but may cause users to conclude that an attempt at a clean shutdown failed.

davecgh: I agree with @Roasbeef here.  The shutdown should remain.

tuxcanfly:
OK, thanks, let me close this. The guard for shutdown error logging as described by @Roasbeef could be added in a new pr, if required.

PR #491, opened 2015-08-26
blockchain: Rework to use new db interface.

davecgh:
**This pull request requires #380**

# NOTE: This pull request requires a full redownload of the chain.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/491)

This pull request is the first stage of several that are planned to convert the `blockchain` package into a concurrent safe package that will ultimately allow support for multi-peer download and concurrent chain processing.  The goal is to update `btcd` proper after each step so it can take advantage of the enhancements as they are developed.

In addition to the aforementioned benefit, this staged approach has been chosen since it is absolutely critical to maintain consensus.  Separating the changes into several stages makes it easier for reviewers to logically follow what is happening and therefore helps prevent consensus bugs.  Naturally there are significant automated tests to help prevent consensus issues as well.

The main focus of this stage is to convert the `blockchain` package to use the new database interface and implement the chain-related functionality which it no longer handles while aiming to improve efficiency along the way in various areas by making use of the new database and chain capabilities.

The following is an overview of the chain changes:

- Update to use the new database interface
- Add chain-related functionality that the old database used to handle
  - Main chain structure and state
  - Transaction spend tracking
- Implement a new pruned unspent transaction output (utxo) set
  - Provides efficient direct access to the unspent transaction outputs
  - Uses a domain specific compression algorithm that understands the standard transaction scripts in order to significantly compress them
  - Removes reliance on the transaction index and paves the way toward eventually enabling block pruning
- Modify the New function to accept a `Config` struct instead of inidividual parameters
- Replace the old TxStore type with a new UtxoViewpoint type that makes use of the new pruned utxo set
- Convert code to treat the new UtxoViewpoint as a rolling view that is used between connects and disconnects to improve efficiency
- Make best chain state always set when the chain instance is created
  - Remove now unnecessary logic for dealing with unset best state
- Make all exported functions concurrent safe
  - Currently using a single chain state lock as it provides a straight forward and easy to review path forward however this can be improved with more fine grained locking
- Optimize various cases where full blocks were being loaded when only
  the header is needed to help reduce the I/O load
- Add the ability for callers to get a snapshot of the current best chain stats in a concurrent safe fashion
  - Does not block callers while new blocks are being processed
- Make error messages that reference transaction outputs consistently use `<transaction hash>:<output index>`
-  Introduce a new AssertError type an convert internal consistency checks to use it
- Update tests and examples to reflect the changes
- Add a full suite of tests to ensure correct functionality of the new code

The following is an overview of the btcd changes:

- Update to use the new database and chain interfaces
- Temporarily remove all code related to the transaction index
- Temporarily remove all code related to the address index
- Convert all code that uses transaction stores to use the new utxo view
- Rework several calls that required the block manager for safe concurrency to use the chain package directly now that it is concurrent safe
- Change all calls to obtain the best hash to use the new best state snapshot capability from the chain package
- Remove workaround for limits on fetching height ranges since the new database interface no longer imposes them
- Correct the `gettxout` RPC handler to return the best chain hash as opposed the hash the txout was found in
- Optimize various RPC handlers:
  - Change several of the RPC handlers to use the new chain snapshot capability to avoid needlessly loading data
  - Update several handlers to use new functionality to avoid accessing the block manager so they are able to return the data without blocking when the server is busy processing blocks
  - Update non-verbose `getblock` to avoid deserialization and serialization overhead
  - Update `getblockheader` to request the block height directly from chain and only load the header
  - Update `getdifficulty` to use the new cached data from chain
  - Update `getmininginfo` to use the new cached data from chain
  - Update non-verbose `getrawtransaction` to avoid deserialization and serialization overhead
  - Update `gettxout` to use the new utxo view versus loading full transactions using the transaction index

The following is an overview of the utility changes:
- Update `addblock` to use the new database and chain interfaces
- Update `findcheckpoint` to use the new database and chain interfaces
- Remove the `dropafter` utility which is no longer supported

**NOTE:** In order to get testing started on this code this commit is being pushed via a pull request before everything is done although it should be stable.  Before it can be merged to master, the following additional
things are needed:

- [x] Implement a proper pruned utxo set and reorg journal
- [x] Reimplement the transaction index using the new database interface including implementing the functionality that used to live in the old database, but will now be needed in btcd proper (#650)
- [x] Reimplement the address index using the new database interface including implementing the functionality that used to live in the old database, but will now be needed in btcd proper (#650)
- [x] Increase test coverage


blockchain/blocklocator.go
```diff
@@ -77,73 +85,94 @@ func (b *BlockChain) BlockLocatorFromHash(hash *wire.ShaHash) BlockLocator {
...
+		return nil
+	})
 
 	// Append the appropriate genesis block.
 	locator = append(locator, b.chainParams.GenesisHash)
 	return locator
 }
 
+// BlockLocatorFromHash returns a block locator for the passed block hash.
+// See BlockLocator for details on the algotirhm used to create a block locator.
```
Roasbeef: algotirhm -> algorithm

blockchain/chain.go
```diff
@@ -137,17 +122,49 @@ func removeChildNode(children []*blockNode, node *blockNode) []*blockNode {
...
 }
 
+// BestState houses information about the current best block and other info
+// related to the state of the main chain as it exists from the point of view of
+// the current best block.
+//
+// The BestSnapshot method can be used to obtain access to this information
+// in a concurrent safe manner and the data will not be changed out from under
+// the caller when chain state changes occur as the function name implies.
+// However, the returned snapshot must be treated as immutable since it is
```
Roasbeef: Makes me wish Go had support for compiler enforced immutable structs. Which might look something like this: 
```go
type BestState immutable struct {
 ...
}
```

However, it's possible to emulate this sort of struct immutability today. We can do this by moving `BestState` to its own module and making all the attributes of the struct private (lower case), and only providing access of the attributes via publicly exported methods.

blockchain/chain.go
```diff
@@ -742,16 +715,65 @@ func (b *BlockChain) getReorganizeNodes(node *blockNode) (*list.List, *list.List
...
+
+	// Atomically insert info into the database.
+	err := b.db.Update(func(dbTx database.Tx) error {
+		// Update best block state.
+		err := dbPutBestState(dbTx, state, node.workSum)
+		if err != nil {
+			return err
+		}
+
+		// Add the block hash and heigt to the block index which tracks
```
Roasbeef: heigt -> height

blockchain/chain.go
```diff
@@ -763,31 +785,101 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block) error {
...
+	// accessing node.parent directly as it will dynamically create previous
+	// block nodes as needed.  This helps allow only the pieces of the chain
+	// that are needed to remain in memory.
 	prevNode, err := b.getPrevNodeFromNode(node)
 	if err != nil {
 		return err
 	}
-	err = b.db.DropAfterBlockBySha(prevNode.hash)
+
+	// Load the previos block since some details for it are needed below.
```
Roasbeef: previos -> previous

blockchain/chain.go
```diff
@@ -763,31 +785,101 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block) error {
...
+	state := newBestState(prevNode, blockSize, numTxns, newTotalTxns)
+
+	err = b.db.Update(func(dbTx database.Tx) error {
+		// Update best block state.
+		err := dbPutBestState(dbTx, state, node.workSum)
+		if err != nil {
+			return err
+		}
+
+		// Remove the block hash and heigt from the block index which
```
Roasbeef: heigt -> height

blockchain/chainio.go
```diff
@@ -0,0 +1,944 @@
...
+// isNotInMainChainErr return whether or not the passed error is an
+// errNotInMainChain error.
+func isNotInMainChainErr(err error) bool {
+	_, ok := err.(errNotInMainChain)
+	return ok
+}
+
+// -----------------------------------------------------------------------------
+// The transaction index consists of an entry for every transaction in the main
+// chain.  Since it possible for multiple transcations to have the same hash as
```
Roasbeef: transcations -> transactions

blockchain/chainio.go
```diff
@@ -0,0 +1,944 @@
...
+	offset += 8
+	byteOrder.PutUint32(serializedData[offset:], workSumBytesLen)
+	offset += 4
+	copy(serializedData[offset:], workSumBytes)
+	return serializedData[:]
+}
+
+// deserializeBestChainState deserializes the passed serialized best chain
+// state.  This is data stored in the chain state bucket and is updated after
+// ever block is connected or disconnected form the main chain.
```
Roasbeef: ever -> every

blockchain/utxolookup.go
```diff
@@ -0,0 +1,548 @@
...
+import (
+	"fmt"
+
+	database "github.com/btcsuite/btcd/database2"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+)
+
+// utxoOutput houses details about an invidual unspent transaction output such
```
Roasbeef: invidual -> individual

blockchain/utxolookup.go
```diff
@@ -0,0 +1,548 @@
...
+
+	return entry.outputs[packedIndex].amount
+}
+
+// PkScript returns the public key script for the provided output index.
+//
+// Returns nil if the output index references an output that does not exist
+// either due to it being invalid or because the output is not part of the store
+// due to previously being spent/pruned.
+func (entry *UtxoEntry) PkScript(outputIndex uint32) []byte {
```
Roasbeef: Here's a suggestion for a more descriptive function name instead of `PkScript` (which sounds like an attribute): `PkScriptByIndex`.

blockchain/utxolookup.go
```diff
@@ -0,0 +1,548 @@
...
+//
+// The unspent outputs are need by other transactions for things such as script
+// validation and double spend prevention.
+type UtxoStore map[wire.ShaHash]*UtxoEntry
+
+// Entry returns information about a given transaction according to the current
+// state of the store.  It will return nil if the passed transaction hash does
+// not exist in the store or is otherwise not available such as when it has been
+// disconnected during a reorg.
+func (store UtxoStore) Entry(txHash *wire.ShaHash) *UtxoEntry {
```
Roasbeef: When used above it wasn't immediately apparent what this function did from simply its name. How about calling it either 'FetchEntry' or 'GetEntry'  in order to make the name less ambiguous.

blockchain/utxolookup.go
```diff
@@ -0,0 +1,548 @@
...
+
+	return entry
+}
+
+// addTxOuts adds all outputs in the passed transaction which are not provably
+// unspendable as available unspent transaction outputs.  This function only
+// differs from AddTxOuts in that it accepts an entry that has already been
+// looked up in the store.  This allow an extra extra map lookup to be avoided
+// when connecting transactions.
+func (store UtxoStore) addTxOuts(entry *UtxoEntry, tx *btcutil.Tx, blockHeight int32) {
```
Roasbeef: IMO the usage of `adds` here is a bit unclear. How about revising the comment to read: `adds all outputs in the passed transactions to the UtxoStore...`.

The same applies to the publicly exported version of the function below.

blockchain/validate.go
```diff
@@ -976,19 +946,20 @@ func (b *BlockChain) checkConnectBlock(node *blockNode, block *btcutil.Block) er
 	// it isn't currently necessary.
 
 	// The coinbase for the Genesis block is not spendable, so just return
-	// now.
-	if node.hash.IsEqual(b.chainParams.GenesisHash) && b.bestChain == nil {
-		return nil
+	// error now.
```
Roasbeef: error now -> an error now

Roasbeef:
Great work @davecgh! Very clean code, and highly descriptive comments as usual.

I've completed my first initial pass. But seeing as this PR is rather large and touches some critical sections. I'll be making *at least* one additional review pass. 

I'm also looking forward to running some basic benchmarks on the new `rpcserver` (new as in no longer blocking other clients when DB access is needed), we should see some significant throughput gains with concurrent clients.

blockchain/chain.go
davecgh: True, but I really didn't want to add the overhead of function calls for access to the fields.

davecgh:
@Roasbeef:  Thanks for the review.  I'll get everything addressed.

blockchain/utxolookup.go
davecgh: I agree a better name could be beneficial here.  I don't either `GetEntry` or `FetchEntry` would work well though because `GetEntry` is against [effective Go guidelines](http://golang.org/doc/effective_go.html) and in general `FetchX` is used throughout the code base to mean it's actually hitting the database such as in the case of `FetchUtxoStore`, `fetchInputUtxos`, etc.

In this case, it's simply attempting to return a transaction entry from the utxo store which may may not be there (a map lookup).  How about `LookupEntry`?

blockchain/utxolookup.go
davecgh: Sounds good to me.  I'll make that change.

davecgh:
Also, in regards to the non-blocking related enhancements to the `rpcserver`, I've done some testing myself and the difference is huge, especially when doing IBD when the database and block manager is super busy.  I'm interested to hear your results as well.

blockchain/utxolookup.go
wallclockbuilder: 145 | +// The unspent outputs are need by other transactions

need -> needed

davecgh: @wallclockbuilder Fixed in latest commit.  Thanks.

blockchain/chain.go
Roasbeef: Fair enough. I agree that the additional overhead should be avoided if possible. 

We'll just have to be vigilant during future reviews of PR's involving `BestState` to ensure prohibited state mutation doesn't pass through unnoticed.

blockchain/utxolookup.go
Roasbeef: Ah of course, one of those naming suggestions would indeed violate the guidelines in Effective Go.

Also agree that we should restricted `FetchX`, to database related functions in order to maintain constancy throughout the code w.r.t semantic naming. IMO naming consistency, makes the codebase easier to follow for new readers. 

I like `LookupEntry`, lets go with that :+1:

blockchain/chain.go
```diff
@@ -742,16 +715,65 @@ func (b *BlockChain) getReorganizeNodes(node *blockNode) (*list.List, *list.List
...
+		// Add the block hash and height to the block index which tracks
+		// the main chain.
+		err = dbPutBlockIndex(dbTx, block.Sha(), node.height)
+		if err != nil {
+			return err
+		}
+
+		// Add transaction index entries for all of the transactions in
+		// the block being connected.
+		err = dbAddTxIndexEntries(dbTx, block)
```
Roasbeef: Once the pruned utxo set has been implemented this can be made optional to a degree with a `--txidx` flag. And as we discussed on IRC it would be wise to consolidate all indexing operations (`tx\addr`) in a single batched atomic write. 

Leaving this comment here as a reminder.

davecgh:
Rebased.

davecgh:
Rebased to latest master (on top of database2 branch).

davecgh:
Rebased to latest master (on top of database2 branch).  This includes resolving all conflicts created by the recent peer refactor pull request (#445).


---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/491)

davecgh:
Rebased to latest master (on top of database2 branch). This includes resolving all conflicts created by the recent pull requests.


---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/491)

davecgh:
Rebased to latest master (on top of database2 branch). This includes resolving all conflicts created by the recent pull requests.

davecgh:
Rebased to latest master (on top of database2 branch). This includes resolving all conflicts created by the recent pull requests.

Also add optimized utxo set and view.

Dirbaio:
Review status: 0 of 84 files reviewed at latest revision, 2 unresolved discussions.

---

*[blockchain/utxoviewpoint.go, line 131 \[r4\]](https://reviewable.io:443/reviews/btcsuite/btcd/491#-K5S8eioO_CyWeI3KpQU:-K5S8eioO_CyWeI3KpQV:-135772379) ([raw file](https://github.com/btcsuite/btcd/blob/cde690532fe9092d59850e178f47e0edefe75317/blockchain/utxoviewpoint.go#L131)):*
I'd name this AmountByIndex, for consistency with PkScriptByIndex

---

*[blockchain/utxoviewpoint.go, line 551 \[r4\]](https://reviewable.io:443/reviews/btcsuite/btcd/491#-K5S9SvKMeIlOM68SOMv:-K5S9SvKMeIlOM68SOMw:445121488) ([raw file](https://github.com/btcsuite/btcd/blob/cde690532fe9092d59850e178f47e0edefe75317/blockchain/utxoviewpoint.go#L551)):*
UtxoViewpoint

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/491)*

davecgh:
Review status: 0 of 84 files reviewed at latest revision, 2 unresolved discussions.

---

*[blockchain/utxoviewpoint.go, line 131 \[r4\]](https://reviewable.io:443/reviews/btcsuite/btcd/491#-K5S8eioO_CyWeI3KpQU:-K5TZ9ikLUSxbTU2CJN8:1346045320) ([raw file](https://github.com/btcsuite/btcd/blob/cde690532fe9092d59850e178f47e0edefe75317/blockchain/utxoviewpoint.go#L131)):*
That's reasonable.  Originally, `PkScriptByIndex` was just `PkScript`, but it was suggested to add the `ByIndex` to make it more clear.

Done.

---

*[blockchain/utxoviewpoint.go, line 551 \[r4\]](https://reviewable.io:443/reviews/btcsuite/btcd/491#-K5S9SvKMeIlOM68SOMv:-K5TZ5sg-GdEh3hDB_aV:676635175) ([raw file](https://github.com/btcsuite/btcd/blob/cde690532fe9092d59850e178f47e0edefe75317/blockchain/utxoviewpoint.go#L551)):*
Done

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/491)*

davecgh:
Reworked the `UtxoEntry` struct a bit to better support the upcoming cache as well as a few other things described in the commit.

Dirbaio:
Review status: 0 of 84 files reviewed at latest revision, 4 unresolved discussions.

---

*[blockchain/utxoviewpoint.go, line 237 \[r7\]](https://reviewable.io:443/reviews/btcsuite/btcd/491#-K6-gp-WJl7VjNb0tapV:-K6-gp-WJl7VjNb0tapW:941988020) ([raw file](https://github.com/btcsuite/btcd/blob/315459419ad86a5d3891d1ec6dfb981862027752/blockchain/utxoviewpoint.go#L237)):*
Shouldn't this set output.compressed = false?

---

*[blockchain/validate.go, line 873 \[r7\]](https://reviewable.io:443/reviews/btcsuite/btcd/491#-K6-hUzKI4pcJG1WpK0e:-K6-hUzKI4pcJG1WpK0f:2000373971) ([raw file](https://github.com/btcsuite/btcd/blob/315459419ad86a5d3891d1ec6dfb981862027752/blockchain/validate.go#L873)):*
This error message could be confusing if the cause is actually an out-of-bounds index. How about something like "Transaction X tried to spend output Y, which is either already spent or out of bounds."

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/491)*

davecgh:
Review status: 0 of 84 files reviewed at latest revision, 4 unresolved discussions.

---

*[blockchain/utxoviewpoint.go, line 237 \[r7\]](https://reviewable.io:443/reviews/btcsuite/btcd/491#-K6-gp-WJl7VjNb0tapV:-K6-kQDZ9R9hF4phudMg:257553225) ([raw file](https://github.com/btcsuite/btcd/blob/315459419ad86a5d3891d1ec6dfb981862027752/blockchain/utxoviewpoint.go#L237)):*
Nope, the fields of a `MsgTx` are never compressed.

---

*[blockchain/validate.go, line 873 \[r7\]](https://reviewable.io:443/reviews/btcsuite/btcd/491#-K6-hUzKI4pcJG1WpK0e:-K6-kwef5VKPTdkI3ByH:238393811) ([raw file](https://github.com/btcsuite/btcd/blob/315459419ad86a5d3891d1ec6dfb981862027752/blockchain/validate.go#L873)):*
Well, that is intended to show the specific input transaction/input index combo which is needed when debugging since details from the pair are needed.  For example, both the signature script that comes from the transaction input and the public key script that comes from the output as well as the sequence number from the input.

Would " transaction **input** %s:%d ..." be more clear?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/491)*

davecgh:
Review status: 0 of 84 files reviewed at latest revision, 4 unresolved discussions.

---

*[blockchain/utxoviewpoint.go, line 237 \[r7\]](https://reviewable.io:443/reviews/btcsuite/btcd/491#-K6-gp-WJl7VjNb0tapV:-K6-nOIlv_3A3O7EDlUQ:123008706) ([raw file](https://github.com/btcsuite/btcd/blob/315459419ad86a5d3891d1ec6dfb981862027752/blockchain/utxoviewpoint.go#L237)):*
Oh, I misread your comment and thought it said `= true`.  Yes, you're right, it set set it to false in the case of being replace in case the previous one was not.  Nice catch.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/491)*

davecgh:
Review status: 0 of 84 files reviewed at latest revision, 4 unresolved discussions.

---

*[blockchain/utxoviewpoint.go, line 237 \[r7\]](https://reviewable.io:443/reviews/btcsuite/btcd/491#-K6-gp-WJl7VjNb0tapV:-K6-oTEHdjARk93SQiRV:676635175) ([raw file](https://github.com/btcsuite/btcd/blob/315459419ad86a5d3891d1ec6dfb981862027752/blockchain/utxoviewpoint.go#L237)):*
Done

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/491)*

davecgh:
Rebased to latest master and merged a few of the commits for good housekeeping.

Dirbaio:
Review status: 0 of 85 files reviewed at latest revision, 5 unresolved discussions.

---

*[blockchain/chainio.go, line 917 \[r9\]](https://reviewable.io:443/reviews/btcsuite/btcd/491#-K8Hj7j4cNoHdBt5PYVy:-K8Hj7j5xia8MMHuG7b7:-1748727034) ([raw file](https://github.com/btcsuite/btcd/blob/f2a1e201a35637cec65a85cb33417f3414f5dbfd/blockchain/chainio.go#L917)):*
This should be 

    copy(serializedData[1:], existing[1:])

Otherwise it overwrites serializedData[0], which is the entry count. I've run into crashes due to this.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/491)*

davecgh:
Reviewed 51 of 78 files at r1, 2 of 8 files at r2, 7 of 17 files at r3, 7 of 12 files at r4, 2 of 6 files at r5, 1 of 3 files at r6, 1 of 3 files at r7, 1 of 1 files at r8, 2 of 3 files at r9, 11 of 11 files at r10.

---

*[blockchain/chainio.go, line 917 \[r9\]](https://reviewable.io:443/reviews/btcsuite/btcd/491#-K8Hj7j4cNoHdBt5PYVy:-K9Tu4n5nuL2duWvuvpN:1141213963) ([raw file](https://github.com/btcsuite/btcd/blob/f2a1e201a35637cec65a85cb33417f3414f5dbfd/blockchain/chainio.go#L917)):*
Nice catch.  Fixed.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/491)*

davecgh:
Rebased over the latest master which has now had #380 merged.

davecgh:
Reviewed 1 of 17 files at r3, 36 of 36 files at r11, 4 of 41 files at r12.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/491)*

davecgh:
Reviewed 1 of 17 files at r3, 14 of 14 files at r13.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/491)*

davecgh:
As an update, this has been rebased to the latest master and I've pushed a new commit to address all outstanding review items.

This PR is now done with the exception of the required address index pieces that are being worked on via #573.

davecgh:
Reviewed 1 of 17 files at r3, 42 of 42 files at r15.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/491)*

davecgh:
Updated to accept `Config` struct instead of individual parameters.  This is useful for the upcoming indexing infrastructure changes since it means less code changes when new optional parameters are added.

davecgh:
This has been updated to remove the code related to the transaction index that is no longer required.

It will be implemented again in a separate indexing pull request which will be required to go in at the same time as this one.  In the mean time, the `getrawtransaction` RPC will not be able to retrieve transactions that aren't in the memory pool when using this PR alone.

---

Reviewed 5 of 7 files at r16, 5 of 5 files at r17.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/491#-:-KBA-wr6oJLpN6Lh16Pv:-746973525)*

davecgh:
Rebased to the latest master and updated to ensure code is simplified per `go fmt -s`.

---

Reviewed 12 of 12 files at r18.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/491#-:-KBP0nBnvFI52947s4ux:536069256)*

davecgh:
This has been rebased to the latest master.  Also, now that #650 has landed, it is ready for master once that PR has been reviewed and accepted.

davecgh:
Reviewed 2 of 2 files at r19, 1 of 1 files at r20.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/491)*

davecgh:
Reviewed 4 of 4 files at r21.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/btcsuite/btcd/491)*

davecgh:
This has been rebased to the latest master with all conflicts resolved.

davecgh:
This has been rebased to the latest master.

---

Reviewed 3 of 3 files at r22.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/btcsuite/btcd/491#-:-KF55CgPsyzwMbGj_AYs:-695469866)*

davecgh:
This has been rebased to the latest master with all conflicts resolved.

---

Reviewed 3 of 3 files at r23.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/btcsuite/btcd/491#-:-KF6MXK5rLEV88JOv8lj:-1146252506)*

dajohi:
Been running with this for months.  OK to put into master which will force more testing :)

PR #492, opened 2015-08-26
database: Replace with new version.

davecgh:
**Requires #380, #491, and #650.**

This pull request removes the old database package, moves the new package into its place, and updates all imports accordingly.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/492)


davecgh:
Rebased.

davecgh:
Rebased.

davecgh:
Rebased to latest master (on top of blockchain_database2 branch).


---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/492)

davecgh:
Rebased to latest master (on top of blockchain_database2 branch).


---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/492)

davecgh:
Rebased to latest master (on top of blockchain_database2 branch).

davecgh:
Rebased to latest master (on top of blockchain_database2 branch).

jongillham:
Reviewed 9 of 97 files at r1, 5 of 134 files at r2, 6 of 12 files at r3, 3 of 7 files at r4, 119 of 125 files at r5.

---

*[database/interface.go, line 22 \[r5\]](https://reviewable.io:443/reviews/btcsuite/btcd/492#-K7gUIt-wrLfr_kCGXde:-K7gUIt-wrLfr_kCGXdf:828737186) ([raw file](https://github.com/btcsuite/btcd/blob/5f7d5ee2ba2132087f12c0a5a51d005bccfcea37/database/interface.go#L22)):*
Most of the `Cursor` methods don't return errors. What would be the best way to expose errors in the underlying storage implementation without these methods returning errors?

---

*[database/interface.go, line 177 \[r5\]](https://reviewable.io:443/reviews/btcsuite/btcd/492#-K7gSXWyW_DRTeYjR7FT:-K7gSXWyW_DRTeYjR7FU:-668507850) ([raw file](https://github.com/btcsuite/btcd/blob/5f7d5ee2ba2132087f12c0a5a51d005bccfcea37/database/interface.go#L177)):*
Would it make sense for this method to also return an error such as `Get(key []byte) ([]byte, error)` to signify errors in the underlying storage?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/492)*

davecgh:

---

*[database/interface.go, line 22 \[r5\]](https://reviewable.io:443/reviews/btcsuite/btcd/492#-K7gUIt-wrLfr_kCGXde:-K8KKRM-tsYYlML3knwu:-1842149340) ([raw file](https://github.com/btcsuite/btcd/blob/5f7d5ee2ba2132087f12c0a5a51d005bccfcea37/database/interface.go#L22)):*
Hmm.  This is a fair point, but in my experience, the vast majority of cursor type interfaces only have errors on the initial cursor creation and deletion.  The rest is simply iterating through snapshots.  It's something worth considering.

---

*[database/interface.go, line 177 \[r5\]](https://reviewable.io:443/reviews/btcsuite/btcd/492#-K7gSXWyW_DRTeYjR7FT:-K8KL2OvAD8Wyxi49mP4:-1675017078) ([raw file](https://github.com/btcsuite/btcd/blob/5f7d5ee2ba2132087f12c0a5a51d005bccfcea37/database/interface.go#L177)):*
I honestly prefer the simplicity of not having an error since the caller is going to need check for malformed data when doing things with it anyways in which case it'll see the data is nil when it shouldn't be and know there is something kind of database error anyways.

That said, it probably should be changed since you can get better errors out of the underlying storage system that way.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/492)*

jongillham:

---

*[database/interface.go, line 22 \[r5\]](https://reviewable.io:443/reviews/btcsuite/btcd/492#-K7gUIt-wrLfr_kCGXde:-K8Ks_zPGe5yYjSVf2_L:1438468109) ([raw file](https://github.com/btcsuite/btcd/blob/5f7d5ee2ba2132087f12c0a5a51d005bccfcea37/database/interface.go#L22)):*
I just reviewed several distributed key/value stores that I use/have used including Mongo, Cassandra & Google Datastore which all produce errors when stepping through their cursors as they stream their data. Of course you can preloading their entire cursor datasets into memory to simulate a snapshot although in this case `Bucket.Cursor() Cursor` would need to be changed to `Bucket.Cursor() (Cursor, error)` to cater for errors generating the snapshot. If this were the case then we are now limited to cursors that are small enough to fit into memory (which might be perfectly acceptable.)

---

*[database/interface.go, line 177 \[r5\]](https://reviewable.io:443/reviews/btcsuite/btcd/492#-K7gSXWyW_DRTeYjR7FT:-K8KdQ1mnT45xoHv7zyb:1179399087) ([raw file](https://github.com/btcsuite/btcd/blob/5f7d5ee2ba2132087f12c0a5a51d005bccfcea37/database/interface.go#L177)):*
I also prefer the simplicity of just `Get(key []byte) []byte`. It does however preclude the use of `Get` to test for the existence of a key value pair as the caller would not be able to differentiate between a pair not existing and a pair existing but there being an underlying database error. That might not be a contract that is needed with the interface though.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/492)*

jongillham:
Review status: 5 of 110 files reviewed at latest revision, 3 unresolved discussions.

---

*[database/interface.go, line 104 \[r7\]](https://reviewable.io:443/reviews/btcsuite/btcd/492#-KATxfilOI6gEufB8_Ar:-KATxfilOI6gEufB8_As:1966010023) ([raw file](https://github.com/btcsuite/btcd/blob/9bc9ad27e6282e3dbd343a20f30f9c3114641acd/database/interface.go#L104)):*
It seems implicit that `DeleteBucket` should also recursively delete all nested buckets otherwise you would have nested buckets without any ability to access them. Maybe this should be explicitly stated?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/492)*

davecgh:
Rebased to the latest master.

---

Reviewed 1 of 134 files at r2, 38 of 125 files at r5, 8 of 143 files at r6, 106 of 125 files at r7, 29 of 29 files at r8.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

*[database/interface.go, line 22 \[r5\]](https://reviewable.io:443/reviews/btcsuite/btcd/492#-K7gUIt-wrLfr_kCGXde:-KBP8lSU_BDvcKuA8bSt:-521727589) ([raw file](https://github.com/btcsuite/btcd/blob/5f7d5ee2ba2132087f12c0a5a51d005bccfcea37/database/interface.go#L22)):*
Yeah, after doing a bit more research, I agree that you're correct.  I've mainly dealt with cursors on memory-mapped databases and hence the data is effectively in memory.  I was modelling it after that given that ultimately I would like to see a memory-mapped database backend.  I'm sure you've noticed several of the comments make reference to that.

However, since the interface is intended to be generic enough to support any database backend, the assumptions do not hold.

That said, I don't want to make the change in this pull request because it is only intended to delete the old database and move the new database2 package (which is already in master) in its place.

I think opening up a separate issue and creating a separate PR after the chain PR #491 and this one are in to expose errors is a fine idea.

---

*[database/interface.go, line 177 \[r5\]](https://reviewable.io:443/reviews/btcsuite/btcd/492#-K7gSXWyW_DRTeYjR7FT:-KBP7IOTfPOgYyl7y3pR:454940737) ([raw file](https://github.com/btcsuite/btcd/blob/5f7d5ee2ba2132087f12c0a5a51d005bccfcea37/database/interface.go#L177)):*
It is actually possible to test for existence because the interface guarantees that an empty byte slice will be returned for keys that exist as opposed to nil for those that don't.  As a result you can do the following:

```Go
value := bucket.Get(key)
if value == nil {
	fmt.Println("key does not exist")
} else if len(value) == 0 {
	fmt.Println("key exists but has no value")
}
```

I should also note that another nice property is taking the len of nil or an empty slice both result in 0, so in many cases where you don't care if it exists and simply that it's not empty, the following works:

```Go
value := bucket.Get(key)
if len(value) == 0 {
	fmt.Println("key either does not exist or exists but has no value")
}
```

---

*[database/interface.go, line 104 \[r7\]](https://reviewable.io:443/reviews/btcsuite/btcd/492#-KATxfilOI6gEufB8_Ar:-KBP7F6zm0WRjDfe96z6:-499146009) ([raw file](https://github.com/btcsuite/btcd/blob/9bc9ad27e6282e3dbd343a20f30f9c3114641acd/database/interface.go#L104)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/492#-:-KBPANU4a4CHlK36Ogm3:-1823062526)*

jongillham:
Review status: 108 of 110 files reviewed at latest revision, 1 unresolved discussion.

---

*[database/interface.go, line 177 \[r5\]](https://reviewable.io:443/reviews/btcsuite/btcd/492#-K7gSXWyW_DRTeYjR7FT:-KBPTGsrE99Khf6JEScz:1587060339) ([raw file](https://github.com/btcsuite/btcd/blob/5f7d5ee2ba2132087f12c0a5a51d005bccfcea37/database/interface.go#L177)):*
My specific issue was if there's an underlying database error when `bucket.Get` is called. `value` might return `nil` but you wouldn't be able to distinguish whether that was because the key really doesn't exist or just temporarily couldn't be fetched from the database due to an underlying error.

On a separate but slightly related note, I have attempted to implement `database.Interface` to a couple of NoSQL databases and had limited success with the `Bucket` and `Cursor` interface mainly due to the nesting functionality. However, I imagine that these interface issues could be tackled in subsequent pull requests once your awesome work on `database`, `blockchain` and the ffldb implementation have been merged into master. (I feel that there should be some orthogonality between the `Tx`, blockchain storage functions and `Bucket/Cursor` interfaces as well as the complete removal of non-managed transaction options. Although I haven't reviewed all your code to understand why the design decisions are they way they are and I know I am late in the game to comment on such drastic changes.)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/492)*

davecgh:
Reviewed 1 of 125 files at r5, 1 of 125 files at r7, 1 of 29 files at r8, 2 of 2 files at r9.

---

*[database/interface.go, line 177 \[r5\]](https://reviewable.io:443/reviews/btcsuite/btcd/492#-K7gSXWyW_DRTeYjR7FT:-KBRRFnjcIVCUu2jUIPt:1369913110) ([raw file](https://github.com/btcsuite/btcd/blob/5f7d5ee2ba2132087f12c0a5a51d005bccfcea37/database/interface.go#L177)):*
Agreed.  The interface will likely need some tweaks.  I had it up on github as a PR for something like 6 months and didn't ever really got a lot of feedback on the interface itself, so I'm glad to read you've been taking a serious look at it to identify these types of issues.  The speed improvements with all of the new PRs you mentioned versus the current code in master are *huge*, so I'm definitely focused on getting it finished up and merged to master before focusing on interface changes at this point.  However, after all of that code makes it in to master, we'll definitely want to address any interface issues so other backends can become a reality 

As far as the buckets, you'll probably have to implement them similar to how I did it in ffldb.  In particular it keeps a unique bucket counter that gives every new bucket (nested or not) a unique value.  Then, it maintains a bucket index that consists of keys which are a prefix followed by the bucket's parent id followed by its name and the value is the bucket's unique id.  With that, all keys are stored prefixed with the unique bucket id.  In that way, you can easily create a cursor over all items in a bucket by doing a prefix scan for the unique bucket id.  As far as checking for nested buckets, that too is easily done by a prefix scan for the bucket index prefix + parent bucket id.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/492)*

davecgh:
Rebased to latest master (on top of blockchain_database2_index branch).

---

Reviewed 3 of 136 files at r1, 1 of 134 files at r2, 2 of 12 files at r3, 38 of 125 files at r5, 38 of 125 files at r7, 7 of 137 files at r10, 130 of 130 files at r11.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/492#-:-KE-3k06Yjl6XpRwfGGL:-1742881976)*

davecgh:
Reviewed 1 of 134 files at r2, 38 of 125 files at r5, 38 of 125 files at r7, 38 of 130 files at r11, 130 of 130 files at r13.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/492)*

davecgh:
Rebased to the latest master (on top of blockchain_database2_index branch).

---

Reviewed 1 of 134 files at r2, 38 of 125 files at r5, 38 of 125 files at r7, 38 of 130 files at r11, 38 of 130 files at r13, 2 of 132 files at r14, 130 of 130 files at r15.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/btcsuite/btcd/492#-:-KEn60oFleCQxvsugZi6:-1157934953)*

davecgh:
Rebased to the latest master (on top of blockchain_database2_index branch).

---

Reviewed 1 of 134 files at r2, 38 of 125 files at r5, 38 of 125 files at r7, 38 of 130 files at r11, 38 of 130 files at r13, 38 of 130 files at r15, 1 of 131 files at r16, 130 of 130 files at r17.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/btcsuite/btcd/492#-:-KF57zpPtgI_uvGMFJ_4:-1157934953)*

davecgh:
Rebased to the latest master (on top of blockchain_database2_index branch).

---

Reviewed 1 of 134 files at r2, 38 of 125 files at r5, 38 of 125 files at r7, 38 of 130 files at r11, 38 of 130 files at r13, 38 of 130 files at r15, 38 of 130 files at r17, 1 of 131 files at r18, 130 of 130 files at r19.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/btcsuite/btcd/492#-:-KF6VvLu3rSlluD2cOw-:-1157934953)*

dajohi:
OK

jrick:
ok

PR #494, opened 2015-08-29
Improve mru inventory map and test cov to 100%.

davecgh:
This pull request improves the most-recently used inventory map human readable string to only show the inventory vectors. and adds tests for the entire structure to bring its coverage to 100%.

In addition, removes the type assertion check in the Add function since the internal inventory list is only managed by the type itself and the tests would now catch any mistakes in the type of entries in the list.


jcvernaleo:
@davecgh New tests look at all good to me.  So OK to go in by me.

PR #495, opened 2015-08-31
Fix skip not being applied to mempool txns in searchrawtransactions

Dirbaio:
Currently, skip is not being applied properly to mempool transactions in searchrawtransactions. This PR fixes that.

I had to change FetchTxsForAddr to additionally return the number of transactions skipped, to properly calculate how many transactions in mempool to skip. (For example, if FetchTxsForAddr returns an empty array, I need to know whether there was no transactions at all, or there were some but they were all skipped).

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -2939,6 +2939,43 @@ func handlePing(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (inter
 	return nil, nil
 }
 
+// getMempoolTxsForAddressRange looks up and returns all transactions from the
+// mempool related to the given address. The, `limit` parameter
+// should be the max number of transactions to be returned. Additionally, if the
+// caller wishes to seek forward in the results some amount, the 'seek'
+// represents how many results to skip.
+// It will return the array of fetched transactions, along with the amount
+// of transactions that were actually skipped.
+func getMempoolTxsForAddressRange(s *rpcServer, addr btcutil.Address, skip int,
+	limit int) ([]*database.TxListReply, int, error) {
+
+	memPoolTxs, err := s.server.txMemPool.FilterTransactionsByAddress(addr)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	// If we're asked to skip more transactions than we have,
+	// we skip them all and return an empty slice.
+	if skip >= len(memPoolTxs) {
+		return nil, len(memPoolTxs), nil
+	}
+
+	var result []*database.TxListReply
+
+	// Otherwise, calculate the range we have to return and return it.
+	rangeEnd := skip + limit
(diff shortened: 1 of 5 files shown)


dajohi:
Reads OK.

rpcserver.go
```diff
@@ -2939,6 +2939,43 @@ func handlePing(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (inter
...
+
+	var result []*database.TxListReply
+
+	// Otherwise, calculate the range we have to return and return it.
+	rangeEnd := skip + limit
+	if rangeEnd > len(memPoolTxs) {
+		rangeEnd = len(memPoolTxs)
+	}
+
+	for _, tx := range memPoolTxs[skip:rangeEnd] {
```
Roasbeef: So, for the bits of the `addrindex` stored in the DB, `skip` allows one to move forward in terms of chronological/dependency order since the `blkHeight` and `txoffset` portions of the index are stored in `BigEndian` order. 

However, the way `skip` is currently implemented within the `mempool` (was this way before your PR), returned entries are always randomized (Go's map iteration is random). If it enhances your use case, it may be worth it to also define some sort of ordering for txs returned from the mempool. A possible order (suggested in the [commit which added ordering](https://github.com/aakselrod/btcd/commit/ead39153aff20d191e4d28c39b86e82cce0924d6)), would arise if one did a topological sort of the mempool txns. If implemented, this would result in txns being returned in dependency order.

Dirbaio: I agree ordering the mempool txs would be nice, but II think it should be addressed in a future PR.

In my testing, mempool ordering is not that random, it doesn't change randomly with every call, probably due to the addrindex in mempool being a ```map[string]map[wire.ShaHash]struct{}```, so the order of the txs of a single address doesn't get "disturbed" unless there's a new transaction for that address.

This fix's goal is simply to be able to iterate through all the mempool transactions when paging. For example, if there are 100 mempool txs and I use 10-txs pages, before the fix the client will get the first 10 transactions regardless of the skip value. After the fix, though, the client can successfully retrieve all the 100 transactions as long as there are no new transactions for that address. In my opinion this change alone makes the behavior much better.

Of course, real ordering will be even better. I'll try to have a look at topological sorting after this.

davecgh:
This reads correctly and I've also tested it for proper functionality.

OK

Roasbeef:
Tested locally, works as advertised.

OK.

davecgh:
@Dirbaio: Can you please [rebase](https://gist.github.com/davecgh/04cb601a147f8c831ca5#rebasing-one-of-your-existing-pull-requests) this to the latest master.  I'll merge it once that's done.

Dirbaio:
Done

PR #497, opened 2015-08-31
Add reverse order option to searchrawtransactions rpc

Dirbaio:
***This PR requires #495***

This PR adds a new parameter to searchrawtransactions that lists transactions in reverse order. 

The reason I believe this is useful is that many apps (wallets and block explorers, for example) list newer transactions on the top, and display transactions in pages. Without the `reverse` flag, clients need to fetch all the transactions to show just the most recent ones, because the total tx count is not known. If it was known, clients could request the last transactions using high skip values, which is still not optimal since it involves skipping all the addrindex rows in the database. 

This solution iterates the addrindex database from the end, so requesting latest transactions just iterates over the last keys in the addrindex. 

Mempool txs are appended at the beginning or end of the result depending on the value of the `reverse flag`. Ordering of the mempool txs themselves is not actually reversed: as discussed in #495 the mempool order is random right now. This can be addressed in the future when an actual ordering is established for mempool txs.


davecgh:
I haven't reviewed this yet, but I did want to point out that the [btcrpcclient](https://github.com/btcsuite/btcrpcclient) will also need a PR ready to go in concurrently with this.  See #487 and btcsuite/btcrpcclient#61 as an example.

rpcserver.go
```diff
@@ -2939,6 +2939,43 @@ func handlePing(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (inter
 	return nil, nil
 }
 
+// getMempoolTxsForAddressRange looks up and returns all transactions from the
+// mempool related to the given address. The, `limit` parameter
+// should be the max number of transactions to be returned. Additionally, if the
+// caller wishes to seek forward in the results some amount, the 'seek'
```
davecgh: Shouldn't this be the 'skip'?

davecgh:
Sorry for the delay in reviewing this.  I got busy working on a bunch of other things and hadn't been able to get to it.  Other than the comment I made, this looks good.  Very well done with updating the tests and documentation!

If you will fix the comment and rebase this to the lastest master, it looks ready to merge.

OK

davecgh:
Created btcsuite/btcrpcclient#64 to coincide with this change.

Dirbaio:
Done :)

PR #499, opened 2015-09-08
Implement SFNodeBloom.

dajohi:
SFNodeBloom is a new service flag that a node is required to use to indicate that it supports bloom filtering.  This includes a protocol version bump to 70011 and a wire version bump to 0.3.0.

btcd:
The SFNodeBloom flag is set by default.  A new configuration option `--nopeerbloomfilters` has been added to to disable bloom filtering.

Any node advertising a version greater than or equal to 70011 that attempts to use bloom filtering will be disconnected if bloom filtering is disabled.

This mimics Bitcoin Core commit afb0ccaf9c9e4e8fac7db3564c4e19c9218c6b03

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -21,86 +21,91 @@ Usage:
   btcd [OPTIONS]
 
 Application Options:
-  -V, --version            Display version information and exit
-  -C, --configfile=        Path to configuration file
-  -b, --datadir=           Directory to store data
-  -a, --addpeer=           Add a peer to connect with at startup
-      --connect=           Connect only to the specified peers at startup
-      --nolisten           Disable listening for incoming connections -- NOTE:
-                           Listening is automatically disabled if the --connect
-                           or --proxy options are used without also specifying
-                           listen interfaces via --listen
-      --listen=            Add an interface/port to listen for connections
-                           (default all interfaces port: 8333, testnet: 18333)
-      --maxpeers=          Max number of inbound and outbound peers (125)
-      --banduration=       How long to ban misbehaving peers.  Valid time units
-                           are {s, m, h}.  Minimum 1 second (24h0m0s)
-  -u, --rpcuser=           Username for RPC connections
-  -P, --rpcpass=           Password for RPC connections
-      --rpclimituser=      Username for limited RPC connections
-      --rpclimitpass=      Password for limited RPC connections
-      --rpclisten=         Add an interface/port to listen for RPC connections
-                           (default port: 8334, testnet: 18334)
-      --rpccert=           File containing the certificate file
-      --rpckey=            File containing the certificate key
-      --rpcmaxclients=     Max number of RPC clients for standard connections
-                           (10)
-      --rpcmaxwebsockets=  Max number of RPC clients for standard connections
-                           (25)
-      --norpc              Disable built-in RPC server -- NOTE: The RPC server
-                           is disabled by default if no rpcuser/rpcpass is
-                           specified
-      --notls              Disable TLS for the RPC server -- NOTE: This is only
-                           allowed if the RPC server is bound to localhost
-      --nodnsseed          Disable DNS seeding for peers
-      --externalip:        Add an ip to the list of local addresses we claim to
-                           listen on to peers
-      --proxy=             Connect via SOCKS5 proxy (eg. 127.0.0.1:9050)
-      --proxyuser=         Username for proxy server
-      --proxypass=         Password for proxy server
-      --onion=             Connect to tor hidden services via SOCKS5 proxy (eg.
-                           127.0.0.1:9050)
-      --onionuser=         Username for onion proxy server
-      --onionpass=         Password for onion proxy server
-      --noonion=           Disable connecting to tor hidden services
-      --torisolation       Enable Tor stream isolation by randomizing user
-                           credentials for each connection.
-      --testnet=           Use the test network
-      --regtest=           Use the regression test network
-      --nocheckpoints=     Disable built-in checkpoints.  Don't do this unless
-                           you know what you're doing.
-      --dbtype=            Database backend to use for the Block Chain (leveldb)
-      --profile=           Enable HTTP profiling on given port -- NOTE port must
-                           be between 1024 and 65536 (6060)
-      --cpuprofile=        Write CPU profile to the specified file
-  -d, --debuglevel:        Logging level for all subsystems {trace, debug, info,
-                           warn, error, critical} -- You may also specify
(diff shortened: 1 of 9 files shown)


wire/msgversion.go
```diff
@@ -18,7 +18,7 @@ import (
 const MaxUserAgentLen = 2000
 
 // DefaultUserAgent for wire in the stack
-const DefaultUserAgent = "/btcwire:0.2.1/"
+const DefaultUserAgent = "/btcwire:0.2.2/"
```
davecgh: This should probably be  `0.3.0`.  Typically it's `<major>.<minor>.<patch>` and this is adding a new flag and bumping the protocol version, so it's more than a patch.

server.go
```diff
@@ -109,6 +109,8 @@ type server struct {
 	nat                  NAT
 	db                   database.Db
 	timeSource           blockchain.MedianTimeSource
+	services             wire.ServiceFlag
+	servicesMtx          sync.Mutex
```
davecgh: This mutex shouldn't be needed.  The `services` field is set during the initial server creation before and it it's not changed after that.

server.go
```diff
@@ -1241,6 +1243,16 @@ out:
 	s.wg.Done()
 }
 
+// Services returns the services that the server supports.  This is safe for
+// concurrent access.
+func (s *server) Services() wire.ServiceFlag {
```
davecgh: As mentioned elsewhere, the server services don't need to be protected by a mutex since they are not mutated after initial creation.  That means this entire function can go away.

server.go
```diff
@@ -1250,6 +1262,11 @@ func newServer(listenAddrs []string, db database.Db, chainParams *chaincfg.Param
 		return nil, err
 	}
 
+	services := defaultServices
+	if cfg.NoPeerBloomFilters {
+		services &= ^wire.SFNodeBloom
```
jrick: Can use &^= here

peer.go
```diff
@@ -1093,6 +1093,18 @@ func (p *peer) handleGetHeadersMsg(msg *wire.MsgGetHeaders) {
 // filter.  The peer will be disconnected if a filter is not loaded when this
 // message is received.
 func (p *peer) handleFilterAddMsg(msg *wire.MsgFilterAdd) {
+	if p.server.services&wire.SFNodeBloom != wire.SFNodeBloom {
```
davecgh: This code section is identical with the exception of the command name in the debug message for all three functions.  It should probably be refactored out into a function.

davecgh: Perhaps something like the following:

```Go
// Comments ....
func (p *peer) enforceFilterSupport(msg wire.Message) bool {
	if p.server.services&wire.SFNodeBloom != wire.SFNodeBloom {
		if p.ProtocolVersion() >= wire.BIP0111Version {
			peerLog.Debugf("%s sent an unsupported %s request "+
				"-- disconnecting", p, msg.Command())
			p.Disconnect()
			return true
		}
		peerLog.Debugf("Ignoring %s request from %s -- bloom "+
			"support is disabled", msg.Command(), p)
		return true
	}
	return false
}
```

Then in the callers:
```Go
if p.enforceFilterSupport(msg) {
	return
}

...
```

davecgh:
Commit message has `-nopeerbloomfilters` which should be `--nopeerbloomfilters`.

doc.go
```diff
@@ -21,86 +21,64 @@ Usage:
...
-                           a block (750000)
-      --blockprioritysize= Size in bytes for high-priority/low-fee transactions
-                           when creating a block (50000)
-      --getworkkey=        DEPRECATED -- Use the --miningaddr option instead
-      --addrindex=         Build and maintain a full address index. Currently
-                           only supported by leveldb.
-      --dropaddrindex=     Deletes the address-based transaction index from the
-                           database on start up, and the exits.
+  -V, --version             Display version information and exit
+  -C, --configfile=         Path to configuration file (/home/dhill/.btcd/btcd.conf)
```
davecgh: The defaults should be removed here.  Notice it wasn't there previously.

doc.go
```diff
@@ -21,86 +21,64 @@ Usage:
...
+  -b, --datadir=            Directory to store data (/home/dhill/.btcd/data)
```
davecgh: Same.

doc.go
```diff
@@ -21,86 +21,64 @@ Usage:
...
+      --logdir=             Directory to log output. (/home/dhill/.btcd/logs)
```
davecgh: Same.

doc.go
```diff
@@ -21,86 +21,64 @@ Usage:
...
+                            used without also specifying listen interfaces via --listen
+      --listen=             Add an interface/port to listen for connections (default all interfaces port: 8333, testnet: 18333)
+      --maxpeers=           Max number of inbound and outbound peers (125)
+      --banduration=        How long to ban misbehaving peers.  Valid time units are {s, m, h}.  Minimum 1 second (24h0m0s)
+  -u, --rpcuser=            Username for RPC connections
+  -P, --rpcpass=            Password for RPC connections
+      --rpclimituser=       Username for limited RPC connections
+      --rpclimitpass=       Password for limited RPC connections
+      --rpclisten=          Add an interface/port to listen for RPC connections (default port: 8334, testnet: 18334)
+      --rpccert=            File containing the certificate file (/home/dhill/.btcd/rpc.cert)
```
davecgh: Same

doc.go
```diff
@@ -21,86 +21,64 @@ Usage:
...
+      --rpckey=             File containing the certificate key (/home/dhill/.btcd/rpc.key)
```
davecgh: Same

doc.go
```diff
@@ -21,86 +21,64 @@ Usage:
...
-                           only supported by leveldb.
-      --dropaddrindex=     Deletes the address-based transaction index from the
-                           database on start up, and the exits.
+  -V, --version             Display version information and exit
+  -C, --configfile=         Path to configuration file
+  -b, --datadir=            Directory to store data
+      --logdir=             Directory to log output.
+  -a, --addpeer=            Add a peer to connect with at startup
+      --connect=            Connect only to the specified peers at startup
+      --nolisten            Disable listening for incoming connections -- NOTE: Listening is automatically disabled if the --connect or --proxy options are
```
davecgh: This and several others no longer match the actual wrapping done by the server when `-h` is specified and it will not show up right on godoc either.

davecgh:
Also `wire/doc.go` needs to have BIP0111 added to the list of supported BIPS.

davecgh:
Nice work.  Everything looks good to me now.

OK

jrick:
ok

Issue #229, opened 2015-01-16
BTCD miss a block and subsequent blocks are all orphan


savardiego:
New installation of btcd 0.9 on a linux debian. Today my node (depsi.zapto.org) has block count 339153 since last night. From the log it looks like it missed the 339154 block, and since the 339155 every block is added as orphan, even if the hash it logs correspond to the right current block (checked on blockchain.info).

Sometimes it seems to start requesting some old block but never the 339154. There's no other error.
The same thing happened yesterday, and after a restart everything went fine.
The node is at my home, and I have a dynamic ip... but other than this problem everything seems to work well.

Thanks.

04:54:26 2015-01-16 [INF] BMGR: Processed 1 block in the last 1m22.57s (375 transactions, height 339153, 2015-01-16 03:55:40 +0100 CET)
04:54:26 2015-01-16 [INF] BMGR: Lost peer 85.175.5.129:21291 (inbound)
04:55:18 2015-01-16 [INF] BMGR: New valid peer 148.251.238.178:60281 (inbound) (/getaddr.bitnodes.io:0.1/)
04:55:18 2015-01-16 [INF] BMGR: Lost peer 148.251.238.178:60281 (inbound)
04:55:39 2015-01-16 [INF] BMGR: New valid peer 112.124.22.80:53714 (inbound) (/bitcoin-seeder:0.01/)
04:55:41 2015-01-16 [INF] BMGR: Lost peer 112.124.22.80:53714 (inbound)
04:55:46 2015-01-16 [INF] BMGR: New valid peer 82.130.102.226:33830 (inbound) (/Snoopy:0.1/)
04:55:46 2015-01-16 [INF] BMGR: Lost peer 82.130.102.226:33830 (inbound)
04:55:48 2015-01-16 [INF] BMGR: New valid peer 27.204.8.63:49756 (inbound) (/Satoshi:0.9.3/)
04:56:30 2015-01-16 [INF] BMGR: New valid peer 88.198.131.156:39072 (inbound) (/bitcoin-seeder:0.01/)
04:56:31 2015-01-16 [INF] BMGR: Lost peer 88.198.131.156:39072 (inbound)
04:57:35 2015-01-16 [INF] BMGR: New valid peer 199.127.225.195:43501 (inbound) (/bitcoin-seeder:0.01/)
04:58:42 2015-01-16 [INF] BMGR: New valid peer 172.245.62.102:36818 (inbound) (/bitcoin-seeder:0.01/)
[...]

Issue #486, opened 2015-08-19
BlockManager stops syncing during initial sync


stevenroose:
I recently deployed btcd to a server with discoverability in order.

I noticed on getaddr.nodelist.io that the block height was no longer increasing (it was still catching up with the network), and this is what I found in the log:
https://gist.github.com/stevenroose/ce3003b8b61b86de5d11

Before that, it was processing blocks as normal, and suddenly it stops. After this excerpt are numerous lines of new connections and lost connections, mostly from `bitcoin-seeder` and `getaddr.bitnodes.io`.

davecgh:
Thanks for the report.  This is a known issue during IBD which is the result of having a single sync peer.  There have somewhat recently been an influx of peers which don't properly serve the chain and unfortunately when the selected sync peer doesn't serve the blocks, there is currently no code to notice this, so it essentially sits there idle waiting for blocks which never come.  Note that this only applies during the initial sync.  Once you're synced, it isn't a problem since the blocks are requested from multiple peers as needed.  Unfortunately, there isn't much you can do at the moment short of disconnecting the peer via btcctl or stopping/restarting btcd.

The current top priority item being working toward is multi-peer downloads which will fix this (and a few other similar type of things) as a part of it.

stevenroose:
Ok, good that you are aware of this. Will restart then.

aquarat:
I've also noticed this issue.

stevenroose:
This really bugs me. btcd stalls more than once a day and I have to restart it manually all the time.

PR #501, opened 2015-09-16
Introduce timer to disconnect peers stalling block download

Roasbeef:
This commit fixes a long standing bug (#486) which previously, would cause a
syncing node to get "stuck", halting all forward progress with respect
to keeping up with the main chain.

There are many nodes on the network which claim to be a fully
functioning full-node, but don't actually properly implement the entire
p2p protocol. For this particular case, after receiving a "getdata" requesting
blocks, these nodes fail to respond with a "block" message.
This would cause btcd to stall block download indefinitely, waiting for
a response which never arrives. Since btcd currently syncs new blocks
via a single syncPeer, it would never attempt to seek to new, active
peers to download the next block from. To fix this issue, we now
disconnect peers whom we deem are stalling forward progress after a
specified timeout.

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -32,6 +32,11 @@ const (
 	// database type is appended to this value to form the full block
 	// database name.
 	blockDbNamePrefix = "blocks"
+
+	// blockStallTimeout is the number of seconds we will wait for a
+	// "block" response after we send out a "getdata" for an announced
+	// block before we deem the peer inactive, and disconnect it.
+	blockStallTimeout = 10 * time.Second
 )
 
 // newPeerMsg signifies a newly connected peer to the block handler.
@@ -775,6 +780,12 @@ func (b *blockManager) fetchHeaderBlocks() {
 		}
 	}
 	if len(gdmsg.InvList) > 0 {
+		// In order to avoid unnecessarily stalling initial block
+		// download due to an unresponsive peer, we initialize a
+		// timer which will disconnect this peer after
+		// blockStallTimeout seconds, unless we receive a block before
+		// then.
+		b.syncPeer.SetBlockStallTimer(blockStallTimeout)
 		b.syncPeer.QueueMessage(gdmsg, nil)
 	}
 }
@@ -1032,6 +1043,7 @@ func (b *blockManager) handleInvMsg(imsg *invMsg) {
 	numRequested := 0
 	gdmsg := wire.NewMsgGetData()
 	requestQueue := imsg.peer.requestQueue
+	blocksRequested := false
 	for len(requestQueue) != 0 {
 		iv := requestQueue[0]
 		requestQueue[0] = nil
@@ -1047,6 +1059,8 @@ func (b *blockManager) handleInvMsg(imsg *invMsg) {
 				gdmsg.AddInvVect(iv)
 				numRequested++
 			}
+
+			blocksRequested = true
 
 		case wire.InvTypeTx:
 			// Request the transaction if there is not already a
@@ -1065,6 +1079,14 @@ func (b *blockManager) handleInvMsg(imsg *invMsg) {
 	}
 	imsg.peer.requestQueue = requestQueue
 	if len(gdmsg.InvList) > 0 {
+		if blocksRequested {
+			// In order to avoid unnecessarily stalling initial block
+			// download due to an unresponsive peer, we initialize a
+			// timer which will disconnect this peer after
+			// blockStallTimeout seconds unless we receive a block
+			// before the timeout.
+			imsg.peer.SetBlockStallTimer(blockStallTimeout)
+		}
 		imsg.peer.QueueMessage(gdmsg, nil)
 	}
 }
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -1076,9 +1076,8 @@ func (b *blockManager) handleInvMsg(imsg *invMsg) {
 				imsg.peer.requestedBlocks[iv.Hash] = struct{}{}
 				gdmsg.AddInvVect(iv)
 				numRequested++
+				blocksRequested = true
 			}
-
-			blocksRequested = true
 
 		case wire.InvTypeTx:
 			// Request the transaction if there is not already a
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -302,7 +302,7 @@ func (b *blockManager) startSync(peers *list.List) {
 
 		// Remove sync candidate peers that are no longer candidates due
 		// to passing their latest known block.  NOTE: The < is
-		// intentional as opposed to <=.  While techcnically the peer
+		// intentional as opposed to <=.  While technically the peer
 		// doesn't have a later block when it's equal, it will likely
 		// have one soon so it is a reasonable choice.  It also allows
 		// the case where both are at 0 such as during regression test.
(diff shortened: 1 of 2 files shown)


Roasbeef:
[Here's an alternative implementation](https://github.com/Roasbeef/btcd/commit/ae680ddc127fc80caf97f0724bdd9a945b1e89fa) that does everything automatically within the peer itself. 

The `blockmanager` approach (as implemented) may possibly, as @dajohi pointed out run into an edge-case where the `outqueue` is backed up and the timeout elapses before the message is even sent out to the peer. If this edge-case is executed, then the approach still "works" but it may pre-maturely boot a peer.

To get around this, and ensure a more precise timeout period, it uses some communication between the `inHandler` and the `outHandler` of the particular peer. It takes advantage of the fact that a select statement will never block on a nil channel. When an outgoing `MsgGetData` is received by the `outHandler`, if a timer isn't already active, it sets up a timer to boot the peer if the `inHandler` doesn't receive a `MsgBlock` in 5 seconds it disconnects the peer. If the `inHandler` gets the message in time, then it cancels the (non-blocking) timeout within the `outHandler` goroutine.

Roasbeef:
After some deliberation, I've come up with a *third* approach to introducing a block stall timer for peers within `btcd`. 

This [new approach](https://github.com/Roasbeef/btcd/commit/102297c239f1b453543b21f18481f3cfc8bd7a4c), in a way, merges the two previous proposals. The timer state is managed solely by the peer within its `inHandler` and `outHandler` goroutines. However, the responsibility of activation of the timer belongs to the blockmanager. 

This way, the timer is started/stopped at the precise points where a peer handles new incoming/outgoing messages. But, the blockmanager (and later executed callbacks with the new `peer` package) have control over the length of the timeout, and can manually activate block stall timeouts on several peers (behavior required to implemented concurrent block download).

It's worth mentioning that I haven't looked very closely at the soon to be merged work to create a new peer package. So I'm not sure which of these approaches would cause the least code/merge-conflict disruption.

davecgh:
@tuxcanfly Can you take a look at these proposals please?  I know you're working on the peer package and so you probably are best qualified to say which one will fit best with your work.

tuxcanfly:
@davecgh Sure, will do.

tuxcanfly:
Ported the patch in the third approach to the new refactored branch:

https://github.com/tuxcanfly/btcd/commit/18151a49a75abc778c53884ca981bc7a4512e636

Had to move the const `BlockStallTimeout` to peer and export it. Other than that, seems fairly straightforward.

davecgh:
From looking at it, I think the latest approach will work well too.  It also seems to fit pretty cleanly with the peer PR.

Mind updating this PR to it?  Then I can comment on it on the files tab and not lose history.

wire/msggetdata.go
```diff
@@ -35,6 +35,17 @@ func (msg *MsgGetData) AddInvVect(iv *InvVect) error {
 	return nil
 }
 
+// RequestsBlock returns true if this GetData message contains a request for a
+// block. Otherwise, it returns false.
+func (msg *MsgGetData) RequestsBlock() bool {
```
Roasbeef: Perhaps this should just be a function defined within `peer.go`?

Roasbeef:
@davecgh done, the PR has been updated to propose only the third proposal :+1:

blockmanager.go
```diff
@@ -1048,6 +1054,10 @@ func (b *blockManager) handleInvMsg(imsg *invMsg) {
 				numRequested++
 			}
 
+			if !blocksRequested {
```
davecgh: I'd just do `blocksRequested = true` here.  There is unlikely to be any efficiency difference, but it would be a little cleaner I think.

wire/msggetdata.go
davecgh: Yeah I think it should be in `peer.go`.

peer.go
```diff
@@ -1796,6 +1827,28 @@ out:
...
+		case timeout := <-p.blockStallActivate:
+			peerLog.Debugf("Activating block stall timer (%v "+
+				"seconds) for: %v", timeout, p)
+			blockStallActive = true
+			stallTimeout = timeout
+		case <-p.blockStallCancel:
+			// The inHandler received a MsgBlock before
+			// blockStallTimeout seconds had elapsed. So we set the
+			// blockStallTimer and blockStallCancel to nil so the
+			//select loop won't block on those cases in the future.
```
davecgh: //select -> // select

davecgh:
I'm fairly certain this is not working as intended.  I turned on peer logging and started a fresh chain sync and there are not any stall-related log messages.

I suspect it's because when it's before the final checkpoint and operating in headers-first mode, inventory announcements are ignored and thus it's never setting the stall timer.  When in headers-first mode the `fetchHeaderBlocks` function is the one that generates the `getdata` messages that request the blocks.

Roasbeef:
I had previously focused my testing on the case wherein after being fully synced, `btcd` would fail to keep up with the best chain, causing it to fall behind. I had overlooked the initial syncing state of headers-first mode during which,  all incoming `invs` are ignored. 

I've updated the PR to also start the stall timer when requesting blocks while in headers-first mode. Additionally, I think this might also fix #229. Previously, during initial sync, `btcd` would never re-request blocks while connected to the same unresponsive `syncNode`. This would cause all other received blocks to be marked as orphan blocks, since we never re-requested the original parent. 

I'm still in the process of testing to ensure the headers-first case is now addressed. I'll report back with my findings.

peer.go
```diff
@@ -1531,6 +1543,9 @@ out:
 			p.handleTxMsg(msg)
 
 		case *wire.MsgBlock:
+			if p.blockStallCancel != nil {
```
davecgh: This is reading the peer field that is being written in `outHandler` which is a separate goroutine.  So it's a race.  The race detect confirms:

==================
WARNING: DATA RACE
Read by goroutine 55:
  main.(*peer).inHandler()
      peer.go:1546 +0x147b

Previous write by goroutine 57:
  main.(*peer).outHandler()
      peer.go:1835 +0x12b1
==================

toddfries:
I'm testing this, and turns out if I start rescanning with 5 btcwallets concurrently ... btcd stalls out enough on the ibd that it thinks it is the peer that is not responding and disconnects the peer.  Somehow detecting we are the problem not them might be in order, but I suspect this will be sorted when async bits are run in parallel not in serial fashion (fix coming in another branch I'm told) .. but in the mean time I think this pull request would do good to set a longer timer as well as adjust the output to something abit more sensible than "... for 5s seconds .." ... see http://sprunge.us/OYYe and http://sprunge.us/EIaO.

peer.go
```diff
@@ -1796,6 +1827,28 @@ out:
 			p.sendDoneQueue <- struct{}{}
 			peerLog.Tracef("%s: acked queuehandler", p)
 
+		case timeout := <-p.blockStallActivate:
+			peerLog.Debugf("Activating block stall timer (%v "+
+				"seconds) for: %v", timeout, p)
```
davecgh: The timeout variable is a `time.Duration` and the print message uses `%v` which means it will print `5s`.  In words, the extra `seconds` is redundant.

peer.go
Roasbeef: Nice catch! I'd forgotten about the existence of Go's [race detector](https://golang.org/doc/articles/race_detector.html). 

I've added a mutex: `blockStallMtx` to protect access to `p.blockStallCancel` between the `inHandler` and the `outHandler`.

peer.go
Roasbeef: Fixed.

davecgh:
As a data point, I'm still seeing stalls during IBD with the latest changes on this PR.

> 12:18:17 2015-10-02 [INF] BMGR: Syncing to block height 375495 from peer 52.xx.xxx.xxx:46325
I noticed a few hours later that I hadn't received any blocks, so I used `btcctl` to find the peer and manually disconnect it `btcctl node disconnect <id>` which led to:
>15:21:27 2015-10-02 [INF] BMGR: Lost peer 52.xx.xxx.xxx:46325 (inbound)                                            
>15:21:27 2015-10-02 [INF] BMGR: Syncing to block height 377164 from peer 98.xxx.xx.xxx:57974
>15:21:49 2015-10-02 [INF] BMGR: Processed 1 block in the last 3h3m32.44s (1828 transactions, height 374848, 2015-09-16 17:42:11 -0500 HC)

As you can see by the `3h3m32.44s` there it took since it processed the last block, the download was clearly stalled on the peer until I manually disconnected it.

davecgh:
Were you able to make any headway on this?

Roasbeef:
Apologies for the lack of an update, I've been rather busy lately. But I haven't forgotten about this PR!

I've pushed two commits. The first adds stall timers when sending off `getheaders` and `getdata` messages. The second aims to fix some false positive peer disconnects which were triggered by the placement of the `blocksRequested` toggle in `blockmanager.handleInvMsg`. 

Additionally, I'm planning on writing a set of tests using the [new peer package](https://github.com/btcsuite/btcd/pull/445) (will post commit here so others can run) in order to ensure the pull works as advertised.

davecgh:
I have a few finishing touches to put on it, but as an update to this, I have a separate PR I will be submitting which approaches this in a different way due to some of the issues we've found with the approach used in this PR.

davecgh:
As previously discussed, I've submitted #518 which is an alternative approach that is a a more general solution and also fixes some concurrency-related issues with that PR as a side effect.

Roasbeef:
Closing in favor of #518.

PR #502, opened 2015-09-17
Update TravisCI to test against golang 1.5.1.

davecgh:
This pull request updates TravisCI to use the most recent versions of Go and also modifies the `goclean.sh` script to quote the test command arguments and update the vet test to exclude recent false positives.


jrick:
ok

PR #506, opened 2015-09-25
Implement and integrate an ECDSA signature verification cache into btcd

Roasbeef:
This PR proposes the addition of an ECDSA signature verification cache
 with a randomized entry eviction policy into btcd. 

 The benefits of SigCache are two fold. Firstly, usage of SigCache mitigates 
 a DoS attack wherein an attacker causes a victim's client to hang due to worst-case 
 behavior triggered while processing attacker crafted invalid transactions. A 
 detailed description of the mitigated DoS attack can be found [here](https://bitslog.wordpress.com/2013/01/23/fixed-bitcoin-vulnerability-explanation-why-the-signature-cache-is-a-dos-protection/). 
 Secondly, usage of the SigCache introduces a signature verification 
 optimization which speeds up the validation of transactions within a block, 
 if they've already once been seen and verified within the mempool.

 Implementation wise: 

 The server itself manages the sigCache instance. The blockManager and
 txMempool both respectively now receive pointers to the created sigCache
 instance. All read (sig triplet existence) operations on the sigCache
 will not block unless a separate goroutine is adding an entry (writing)
 to the sigCache. GetBlockTemplate generation now also utilizes the
 sigCache in order to avoid unnecessarily double checking signatures
 when generating a template after previously accepting a txn to the
 mempool. Consequently, the CPU miner now also employs the same
 optimization.

 The maximum number of entires for the sigCache has been introduced as a
 config parameter in order to allow users to configure the amount of
 memory consumed by this new additional caching.

### DIFF
--- a/config.go
+++ b/config.go
@@ -45,6 +45,7 @@ const (
 	defaultBlockPrioritySize = 50000
 	defaultGenerate          = false
 	defaultAddrIndex         = false
+	defaultSigCacheMaxSize   = 50000
 )
 
 var (
@@ -117,6 +118,7 @@ type config struct {
 	AddrIndex          bool          `long:"addrindex" description:"Build and maintain a full address index. Currently only supported by leveldb."`
 	DropAddrIndex      bool          `long:"dropaddrindex" description:"Deletes the address-based transaction index from the database on start up, and the exits."`
 	NoPeerBloomFilters bool          `long:"nopeerbloomfilters" description:"Disable bloom filtering support."`
+	SigCacheMaxSize    uint          `long:"sigcachemaxsize" description:"The maximum number of entries in the signature verification cache."`
 	onionlookup        func(string) ([]net.IP, error)
 	lookup             func(string) ([]net.IP, error)
 	oniondial          func(string, string) (net.Conn, error)
@@ -322,6 +324,7 @@ func loadConfig() (*config, []string, error) {
 		BlockMinSize:      defaultBlockMinSize,
 		BlockMaxSize:      defaultBlockMaxSize,
 		BlockPrioritySize: defaultBlockPrioritySize,
+		SigCacheMaxSize:   defaultSigCacheMaxSize,
 		MaxOrphanTxs:      maxOrphanTransactions,
 		Generate:          defaultGenerate,
 		AddrIndex:         defaultAddrIndex,
--- a/doc.go
+++ b/doc.go
@@ -105,6 +105,8 @@ Application Options:
       --dropaddrindex       Deletes the address-based transaction index from the
                             database on start up, and the exits.
       --nopeerbloomfilters  Disable bloom filtering support.
+      --sigcachemaxsize=    The maximum number of entries in the signature
+                            verification cache.
 
 Help Options:
   -h, --help           Show this help message
--- a/mining.go
+++ b/mining.go
@@ -366,9 +366,9 @@ func medianAdjustedTime(chainState *chainState, timeSource blockchain.MedianTime
 //  |  transactions (while block size   |   |
 //  |  <= cfg.BlockMinSize)             |   |
 //   -----------------------------------  --
-func NewBlockTemplate(mempool *txMemPool, payToAddress btcutil.Address) (*BlockTemplate, error) {
-	blockManager := mempool.server.blockManager
-	timeSource := mempool.server.timeSource
+func NewBlockTemplate(server *server, payToAddress btcutil.Address) (*BlockTemplate, error) {
+	blockManager := server.blockManager
+	timeSource := server.timeSource
 	chainState := &blockManager.chainState
 
 	// Extend the most recently known best block.
@@ -404,7 +404,7 @@ func NewBlockTemplate(mempool *txMemPool, payToAddress btcutil.Address) (*BlockT
 	// Also, choose the initial sort order for the priority queue based on
 	// whether or not there is an area allocated for high-priority
 	// transactions.
-	mempoolTxns := mempool.TxDescs()
+	mempoolTxns := server.txMemPool.TxDescs()
 	sortedByFee := cfg.BlockPrioritySize == 0
 	priorityQueue := newTxPriorityQueue(len(mempoolTxns), sortedByFee)
 
@@ -474,7 +474,7 @@ mempoolLoop:
 			originIndex := txIn.PreviousOutPoint.Index
 			txData, exists := txStore[*originHash]
 			if !exists || txData.Err != nil || txData.Tx == nil {
-				if !mempool.HaveTransaction(originHash) {
+				if !server.txMemPool.HaveTransaction(originHash) {
 					minrLog.Tracef("Skipping tx %s because "+
 						"it references tx %s which is "+
 						"not available", tx.Sha,
@@ -656,7 +656,7 @@ mempoolLoop:
 			continue
 		}
 		err = blockchain.ValidateTransactionScripts(tx, blockTxStore,
-			txscript.StandardVerifyFlags)
+			txscript.StandardVerifyFlags, server.sigCache)
 		if err != nil {
 			minrLog.Tracef("Skipping tx %s due to error in "+
 				"ValidateTransactionScripts: %v", tx.Sha(), err)
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -1449,7 +1449,7 @@ func (state *gbtWorkState) updateBlockTemplate(s *rpcServer, useCoinbaseValue bo
 		// block template doesn't include the coinbase, so the caller
 		// will ultimately create their own coinbase which pays to the
 		// appropriate address(es).
-		blkTemplate, err := NewBlockTemplate(s.server.txMemPool, payAddr)
+		blkTemplate, err := NewBlockTemplate(s.server, payAddr)
 		if err != nil {
 			return internalRPCError("Failed to create new block "+
 				"template: "+err.Error(), "")
@@ -2609,7 +2609,7 @@ func handleGetWorkRequest(s *rpcServer) (interface{}, error) {
 		// Choose a payment address at random.
 		payToAddr := cfg.miningAddrs[rand.Intn(len(cfg.miningAddrs))]
 
-		template, err := NewBlockTemplate(s.server.txMemPool, payToAddr)
+		template, err := NewBlockTemplate(s.server, payToAddr)
 		if err != nil {
 			context := "Failed to create new block template"
 			return nil, internalRPCError(err.Error(), context)
--- a/txscript/engine.go
+++ b/txscript/engine.go
@@ -86,6 +86,7 @@ type Engine struct {
 	condStack       []int
 	numOps          int
 	flags           ScriptFlags
+	sigCache        *SigCache
 	bip16           bool     // treat execution as pay-to-script-hash
 	savedFirstStack [][]byte // stack from first script for bip16 scripts
 }
@@ -573,7 +574,7 @@ func (vm *Engine) SetAltStack(data [][]byte) {
 // NewEngine returns a new script engine for the provided public key script,
 // transaction, and input index.  The flags modify the behavior of the script
 // engine according to the description provided by each flag.
-func NewEngine(scriptPubKey []byte, tx *wire.MsgTx, txIdx int, flags ScriptFlags) (*Engine, error) {
+func NewEngine(scriptPubKey []byte, tx *wire.MsgTx, txIdx int, flags ScriptFlags, sigCache *SigCache) (*Engine, error) {
 	// The provided transaction input index must refer to a valid input.
 	if txIdx < 0 || txIdx >= len(tx.TxIn) {
 		return nil, ErrInvalidIndex
@@ -588,7 +589,7 @@ func NewEngine(scriptPubKey []byte, tx *wire.MsgTx, txIdx int, flags ScriptFlags
 	// allowing the clean stack flag without the P2SH flag would make it
 	// possible to have a situation where P2SH would not be a soft fork when
 	// it should be.
-	vm := Engine{flags: flags}
+	vm := Engine{flags: flags, sigCache: sigCache}
 	if vm.hasFlag(ScriptVerifyCleanStack) && !vm.hasFlag(ScriptBip16) {
 		return nil, ErrInvalidFlags
 	}
--- a/txscript/opcode.go
+++ b/txscript/opcode.go
@@ -1850,8 +1850,21 @@ func opcodeCheckSig(op *parsedOpcode, vm *Engine) error {
 		return nil
 	}
 
-	ok := signature.Verify(hash, pubKey)
-	vm.dstack.PushBool(ok)
+	var valid bool
+	if vm.sigCache != nil {
+		var sigHash wire.ShaHash
+		copy(sigHash[:], hash)
+
+		valid = vm.sigCache.Exists(sigHash, signature, pubKey)
+		if !valid && signature.Verify(hash, pubKey) {
+			vm.sigCache.Add(sigHash, signature, pubKey)
+			valid = true
+		}
+	} else {
+		valid = signature.Verify(hash, pubKey)
+	}
+
+	vm.dstack.PushBool(valid)
 	return nil
 }
 
@@ -2051,7 +2064,21 @@ func opcodeCheckMultiSig(op *parsedOpcode, vm *Engine) error {
 		// Generate the signature hash based on the signature hash type.
 		hash := calcSignatureHash(script, hashType, &vm.tx, vm.txIdx)
 
-		if parsedSig.Verify(hash, parsedPubKey) {
+		var valid bool
+		if vm.sigCache != nil {
+			var sigHash wire.ShaHash
+			copy(sigHash[:], hash)
+
+			valid = vm.sigCache.Exists(sigHash, parsedSig, parsedPubKey)
+			if !valid && parsedSig.Verify(hash, parsedPubKey) {
+				vm.sigCache.Add(sigHash, parsedSig, parsedPubKey)
+				valid = true
+			}
+		} else {
+			valid = parsedSig.Verify(hash, parsedPubKey)
+		}
+
+		if valid {
 			// PubKey verified, move on to the next signature.
 			signatureIdx++
 			numSignatures--
--- a/txscript/reference_test.go
+++ b/txscript/reference_test.go
@@ -190,40 +190,51 @@ func TestScriptInvalidTests(t *testing.T) {
 			err)
 		return
 	}
-	for i, test := range tests {
-		// Skip comments
-		if len(test) == 1 {
-			continue
-		}
-		name, err := testName(test)
-		if err != nil {
-			t.Errorf("TestBitcoindInvalidTests: invalid test #%d",
-				i)
-			continue
-		}
-		scriptSig, err := parseShortForm(test[0])
-		if err != nil {
-			t.Errorf("%s: can't parse scriptSig; %v", name, err)
-			continue
-		}
-		scriptPubKey, err := parseShortForm(test[1])
-		if err != nil {
-			t.Errorf("%s: can't parse scriptPubkey; %v", name, err)
-			continue
-		}
-		flags, err := parseScriptFlags(test[2])
-		if err != nil {
-			t.Errorf("%s: %v", name, err)
-			continue
-		}
-		tx := createSpendingTx(scriptSig, scriptPubKey)
-		vm, err := NewEngine(scriptPubKey, tx, 0, flags)
-		if err == nil {
-			if err := vm.Execute(); err == nil {
-				t.Errorf("%s test succeeded when it "+
-					"should have failed\n", name)
+	sigCache := NewSigCache(10)
+	sigCacheToggle := []bool{true, false}
+	for _, useSigCache := range sigCacheToggle {
+		for i, test := range tests {
+			// Skip comments
+			if len(test) == 1 {
+				continue
 			}
-			continue
+			name, err := testName(test)
+			if err != nil {
+				t.Errorf("TestBitcoindInvalidTests: invalid test #%d",
+					i)
+				continue
+			}
+			scriptSig, err := parseShortForm(test[0])
+			if err != nil {
+				t.Errorf("%s: can't parse scriptSig; %v", name, err)
+				continue
+			}
+			scriptPubKey, err := parseShortForm(test[1])
+			if err != nil {
+				t.Errorf("%s: can't parse scriptPubkey; %v", name, err)
+				continue
+			}
+			flags, err := parseScriptFlags(test[2])
+			if err != nil {
+				t.Errorf("%s: %v", name, err)
+				continue
+			}
+			tx := createSpendingTx(scriptSig, scriptPubKey)
+
+			var vm *Engine
+			if useSigCache {
+				vm, err = NewEngine(scriptPubKey, tx, 0, flags, sigCache)
+			} else {
+				vm, err = NewEngine(scriptPubKey, tx, 0, flags, nil)
+			}
+
+			if err == nil {
+				if err := vm.Execute(); err == nil {
+					t.Errorf("%s test succeeded when it "+
+						"should have failed\n", name)
+				}
+				continue
+			}
 		}
 	}
 }
@@ -244,42 +255,53 @@ func TestScriptValidTests(t *testing.T) {
 			err)
 		return
 	}
-	for i, test := range tests {
-		// Skip comments
-		if len(test) == 1 {
-			continue
-		}
-		name, err := testName(test)
-		if err != nil {
-			t.Errorf("TestBitcoindValidTests: invalid test #%d",
-				i)
-			continue
-		}
-		scriptSig, err := parseShortForm(test[0])
-		if err != nil {
-			t.Errorf("%s: can't parse scriptSig; %v", name, err)
-			continue
-		}
-		scriptPubKey, err := parseShortForm(test[1])
-		if err != nil {
-			t.Errorf("%s: can't parse scriptPubkey; %v", name, err)
-			continue
-		}
-		flags, err := parseScriptFlags(test[2])
-		if err != nil {
-			t.Errorf("%s: %v", name, err)
-			continue
-		}
-		tx := createSpendingTx(scriptSig, scriptPubKey)
-		vm, err := NewEngine(scriptPubKey, tx, 0, flags)
-		if err != nil {
-			t.Errorf("%s failed to create script: %v", name, err)
-			continue
-		}
-		err = vm.Execute()
-		if err != nil {
-			t.Errorf("%s failed to execute: %v", name, err)
-			continue
+	sigCache := NewSigCache(10)
+	sigCacheToggle := []bool{true, false}
+	for _, useSigCache := range sigCacheToggle {
+		for i, test := range tests {
+			// Skip comments
+			if len(test) == 1 {
+				continue
+			}
+			name, err := testName(test)
+			if err != nil {
+				t.Errorf("TestBitcoindValidTests: invalid test #%d",
+					i)
+				continue
+			}
+			scriptSig, err := parseShortForm(test[0])
+			if err != nil {
+				t.Errorf("%s: can't parse scriptSig; %v", name, err)
+				continue
+			}
+			scriptPubKey, err := parseShortForm(test[1])
+			if err != nil {
+				t.Errorf("%s: can't parse scriptPubkey; %v", name, err)
+				continue
+			}
+			flags, err := parseScriptFlags(test[2])
+			if err != nil {
+				t.Errorf("%s: %v", name, err)
+				continue
+			}
+			tx := createSpendingTx(scriptSig, scriptPubKey)
+
+			var vm *Engine
+			if useSigCache {
+				vm, err = NewEngine(scriptPubKey, tx, 0, flags, sigCache)
+			} else {
+				vm, err = NewEngine(scriptPubKey, tx, 0, flags, nil)
+			}
+
+			if err != nil {
+				t.Errorf("%s failed to create script: %v", name, err)
+				continue
+			}
+			err = vm.Execute()
+			if err != nil {
+				t.Errorf("%s failed to execute: %v", name, err)
+				continue
+			}
 		}
 	}
 }
@@ -414,7 +436,7 @@ testloop:
 			// These are meant to fail, so as soon as the first
 			// input fails the transaction has failed. (some of the
 			// test txns have good inputs, too..
-			vm, err := NewEngine(pkScript, tx.MsgTx(), k, flags)
+			vm, err := NewEngine(pkScript, tx.MsgTx(), k, flags, nil)
 			if err != nil {
 				continue testloop
 			}
@@ -555,7 +577,7 @@ testloop:
 					k, i, test)
 				continue testloop
 			}
-			vm, err := NewEngine(pkScript, tx.MsgTx(), k, flags)
+			vm, err := NewEngine(pkScript, tx.MsgTx(), k, flags, nil)
 			if err != nil {
 				t.Errorf("test (%d:%v:%d) failed to create "+
 					"script: %v", i, test, k, err)
--- a/txscript/sigcache.go
+++ b/txscript/sigcache.go
@@ -1,0 +1,113 @@
+// Copyright (c) 2015 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package txscript
+
+import (
+	"bytes"
+	"crypto/rand"
+	"sync"
+
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/btcsuite/btcd/wire"
+)
+
+// sigInfo represents an entry in the SigCache. Entries in the sigcache are a
+// 3-tuple: (sigHash, sig, pubKey).
+type sigInfo struct {
+	sigHash wire.ShaHash
+	sig     string
+	pubKey  string
+}
+
+// SigCache implements an ECDSA signature verification cache with a randomized
+// entry eviction policy. Only valid signatures will be added to the cache. The
+// benefits of SigCache are two fold. Firstly, usage of SigCache mitigates a DoS
+// attack wherein an attack causes a victim's client to hang due to worst-case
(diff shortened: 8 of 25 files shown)


config.go
```diff
@@ -116,6 +117,7 @@ type config struct {
 	GetWorkKeys        []string      `long:"getworkkey" description:"DEPRECATED -- Use the --miningaddr option instead"`
 	AddrIndex          bool          `long:"addrindex" description:"Build and maintain a full address index. Currently only supported by leveldb."`
 	DropAddrIndex      bool          `long:"dropaddrindex" description:"Deletes the address-based transaction index from the database on start up, and the exits."`
+	SigCacheMaxSize    int           `long:"sigcachemaxsize" description:"hello"`
```
wallclockbuilder: SigCacheMaxSize    int           `long:"sigcachemaxsize" description:"hello"`
SigCacheMaxSize    int           `long:"sigcachemaxsize" description:"~~hello~~Allows users to configure the amount of memory consumed by this new additional caching."`

txscript/sigcache.go
```diff
@@ -0,0 +1,119 @@
...
+
+	var candidateHash wire.ShaHash
+	copy(candidateHash[:], sigHash)
+
+	info := sigInfo{candidateHash, *sig, *pubKey}
+	_, ok := s.validSigs[info]
+	return ok
+}
+
+// Add adds an entry for a signature over 'sigHash' under public kye 'pubKey'
```
wallclockbuilder: +// Add adds an entry for a signature over 'sigHash' under public ~~kye~~ key 'pubKey'

wallclockbuilder:
:100:

rpcserver.go
```diff
@@ -1406,7 +1406,7 @@ func (state *gbtWorkState) templateUpdateChan(prevHash *wire.ShaHash, lastGenera
 // addresses.
 //
 // This function MUST be called with the state locked.
-func (state *gbtWorkState) updateBlockTemplate(s *rpcServer, useCoinbaseValue bool) error {
+func (state *gbtWorkState) updateBlockTemplate(s *rpcServer, useCoinbaseValue bool, sigCache *txscript.SigCache) error {
```
davecgh: Rather than passing the sigCache as a separate parameter, why not just access it via `s.server.sigCache`?

rpcserver.go
```diff
@@ -1685,7 +1685,7 @@ func handleGetBlockTemplateLongPoll(s *rpcServer, longPollID string, useCoinbase
 	// be manually unlocked before waiting for a notification about block
 	// template changes.
 
-	if err := state.updateBlockTemplate(s, useCoinbaseValue); err != nil {
+	if err := state.updateBlockTemplate(s, useCoinbaseValue, s.server.sigCache); err != nil {
```
davecgh: No change would be needed here if `updateBlockTemplate` accessed `s.server.sigCache` directly.

txscript/sigcache.go
```diff
@@ -0,0 +1,119 @@
...
+	return &SigCache{validSigs: make(map[sigInfo]struct{}), maxEntries: maxEntries}
+}
+
+// Exists returns true if an existing entry of 'sig' over 'sigHash' for public
+// key 'pubKey' is found within the SigCache. Otherwise, false is returned.
+//
+// NOTE: This function is safe for concurrent access. Readers won't be blocked
+// unless there exists a writer, adding an entry to the SigCache.
+func (s *SigCache) Exists(sigHash []byte, sig *btcec.Signature, pubKey *btcec.PublicKey) bool {
+	s.RLock()
```
davecgh: Probably not a big deal, but this lock is being held longer than it needs to be since it is only protecting access to the map and defer has non-trivial overhead.  I'd suggest:

```Go
s.RLock()
_, ok := s.validSigs[info]
s.RUnlock()
```

txscript/sigcache_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	sig, err := privKey.Sign(msgHash)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	return msgHash, sig, privKey.PubKey(), nil
+}
+
+// TestSigCacheAdd tests the ability to add, and later check the existence of a
```
davecgh: `TestSigCacheAddExists`

txscript/engine.go
```diff
@@ -75,6 +75,7 @@ var halfOrder = new(big.Int).Rsh(btcec.S256().N, 1)
 
 // Engine is the virtual machine that executes scripts.
 type Engine struct {
+	sigCache        *SigCache
```
davecgh: While it doesn't really matter from a functionality standpoint, this should probably go after `flags` since it mirrors the order it is specified in `New`.

davecgh:
The new parameter will need to be added to `doc.go` and `sample-btcd.conf`.

davecgh:
I think this should also be applied to `opcodeCheckMultiSig` where it has:

````Go
if parsedSig.Verify(hash, parsedPubKey) {
```

txscript/reference_test.go
```diff
@@ -244,42 +257,55 @@ func TestScriptValidTests(t *testing.T) {
...
-		}
-		err = vm.Execute()
-		if err != nil {
-			t.Errorf("%s failed to execute: %v", name, err)
-			continue
+	sigCache := NewSigCache(10)
+	sigCacheToggle := []bool{true, false}
+	for _, useSigCache := range sigCacheToggle {
+		for i, test := range tests {
+			var vm *Engine
```
davecgh: This should be declared just before the `if useSigCache` branch below.  Also `err` doesn't need to be declared since it will already exist by virtual of the previous statements doing `foo, err := ...`.

txscript/sigcache.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/btcsuite/btcd/wire"
+)
+
+// sigInfo represents an entry in the SigCache. Entries in the sigcache are a
+// 3-tuple: (sigCache, sig, pubKey).
+type sigInfo struct {
+	// TODO(roasbeef): Make this a string so it's hash func blocksize
+	// generic?
+	sigHash wire.ShaHash
+	sig     btcec.Signature
```
davecgh: This will not work.  Both `btcec.Signature` and `btcec.PublicKey` contain `big.Int` pointers.  This means two signatures and/or pubkeys that are otherwise identical will not match because the pointers are pointing to different memory.  To prove it, I did the following:

```diff
diff --git a/txscript/sigcache_test.go b/txscript/sigcache_test.go
index 6a1fda6..34375b0 100644
--- a/txscript/sigcache_test.go
+++ b/txscript/sigcache_test.go
@@ -49,7 +49,8 @@ func TestSigCacheAddExists(t *testing.T) {
        sigCache.Add(msg1, sig1, key1)

        // The previously added triplet should now be found within the sigcache.
-       if !sigCache.Exists(msg1, sig1, key1) {
+       sig1Copy, _ := btcec.ParseSignature(sig1.Serialize(), btcec.S256())
+       if !sigCache.Exists(msg1, sig1Copy, key1) {
                t.Errorf("previously added item not found in signature cache")
        }
 }
```
> $ go test
> --- FAIL: TestSigCacheAddExists (0.00s)
> 	sigcache_test.go:57: previously added item not found in signature cache

txscript/sigcache_test.go
```diff
@@ -0,0 +1,130 @@
...
+	msg1, sig1, key1, err := genRandomSig()
+	if err != nil {
+		t.Errorf("unable to generate random signature test data")
+	}
+
+	// Add the triplet to the signature cache.
+	sigCache.Add(msg1, sig1, key1)
+
+	// The previously added triplet should now be found within the sigcache.
+	if !sigCache.Exists(msg1, sig1, key1) {
```
davecgh: As I partially showed in another comment about why the `sigInfo` struct won't work properly, this test really should be using deep copies of the message, signature, and key (can be obtained by serializing and deserializing them) to make sure that you're actually checking for existence of the actual message/sig/key versus the memory they point to.

davecgh:
Also, looks like a few updates in the tests were missed:

> blockchain/common_test.go:119: not enough arguments in call to blockchain.New
> blockchain/example_test.go:47: not enough arguments in call to blockchain.New
> blockchain/scriptval_test.go:40: not enough arguments in call to blockchain.TstCheckBlockScripts

txscript/sigcache_test.go
Roasbeef: Fixed.

txscript/sigcache.go
Roasbeef: Indeed! I forgot that the sig and pubkey structs themselves actually contain pointers. I've changed the `sigInfo` struct to include the serialized versions of `btcec.Signature` and `btcec.PublicKey`. 

Additionally, I've updated all new tests to serialize, then deserialize both the sig and pubkey before checking the `sigCache` for existence.

txscript/reference_test.go
Roasbeef: Fixed.

txscript/engine.go
Roasbeef: Changed.

txscript/sigcache_test.go
Roasbeef: Fixed.

txscript/sigcache.go
Roasbeef: Great suggestion, I've opted to make the locking more fine grained mirroring your suggested code fragment

rpcserver.go
Roasbeef: Indeed, I've made the change you suggested above.

rpcserver.go
Roasbeef: Hmm I can't recall why I previously passed the sigCache in as a separate parameter. 

I've updated the fragment and surrounding code to just use: `s.server.sigCache`.

txscript/sigcache.go
Roasbeef: Fixed.

config.go
Roasbeef: Fixed. 

The `hello` was left over from when I just needed it all to compile :)

Roasbeef:
Updated `doc.go`, `sample-btcd.conf`, added usage of the `sigCache` to `opcodeCheckMultiSig`, and  addressed the above code review comments.

doc.go
```diff
@@ -21,91 +21,90 @@ Usage:
...
+  -P, --rpcpass=           Password for RPC connections
+      --rpclimituser=      Username for limited RPC connections
+      --rpclimitpass=      Password for limited RPC connections
+      --rpclisten=         Add an interface/port to listen for RPC connections
+                           (default port: 8334, testnet: 18334)
+      --rpccert=           File containing the certificate file
+      --rpckey=            File containing the certificate key
+      --rpcmaxclients=     Max number of RPC clients for standard connections
+                           (10)
+      --rpcmaxwebsockets=  Max number of RPC clients for standard connections
```
davecgh: Look like another one got changed incorrectly.  It should be `Max number of RPC websocket connections (25)`.

doc.go
```diff
@@ -21,91 +21,90 @@ Usage:
...
+                           (25)
+      --norpc              Disable built-in RPC server -- NOTE: The RPC server
+                           is disabled by default if no rpcuser/rpcpass is
```
davecgh: And this lost the additional `rpclimituser/rpclimitpass` bits.

doc.go
Roasbeef: Fixed.

doc.go
Roasbeef: Fixed.

doc.go
```diff
@@ -21,91 +21,89 @@ Usage:
...
+      --rpclisten=         Add an interface/port to listen for RPC connections
+                           (default port: 8334, testnet: 18334)
+      --rpccert=           File containing the certificate file
+      --rpckey=            File containing the certificate key
+      --rpcmaxclients=     Max number of RPC clients for standard connections
+                           (10)
+      --rpcmaxwebsockets=  Max number of RPC websocket connections (25)
+      --norpc              Disable built-in RPC server -- NOTE: The RPC server
+                           is disabled by default if no rpcuser/rpcpass or
+			   rpclimituser/rpclimitpass is specified
```
davecgh: `r` is not aligned.

doc.go
```diff
@@ -21,91 +21,89 @@ Usage:
...
+      --onion=             Connect to tor hidden services via SOCKS5 proxy (eg.
+                           127.0.0.1:9050)
+      --onionuser=         Username for onion proxy server
+      --onionpass=         Password for onion proxy server
+      --noonion=           Disable connecting to tor hidden services
+      --torisolation       Enable Tor stream isolation by randomizing user
+                           credentials for each connection.
+      --testnet=           Use the test network
+      --regtest=           Use the regression test network
+      --nocheckpoints=     Disable built-in checkpoints.  Don't do this unless
```
davecgh: Lost `--simnet`.

davecgh:
Speaking of the `doc.go` changes, I don't see why those lines would even need to be changed.  The new option is not wider than the existing ones, so really the only thing should be the no option.  I'd suggest checkingout the existing `doc.go` from master and just copying the single new `--sigcachemaxsize` line into it.  That should mean only a single one-line diff.

mining.go
```diff
@@ -366,7 +366,7 @@ func medianAdjustedTime(chainState *chainState, timeSource blockchain.MedianTime
 //  |  transactions (while block size   |   |
 //  |  <= cfg.BlockMinSize)             |   |
 //   -----------------------------------  --
-func NewBlockTemplate(mempool *txMemPool, payToAddress btcutil.Address) (*BlockTemplate, error) {
+func NewBlockTemplate(mempool *txMemPool, payToAddress btcutil.Address, sigCache *txscript.SigCache) (*BlockTemplate, error) {
```
davecgh: Maybe change this to accept the server instead of the separate `mempool` and `sigCache`?  I suggest that because I noticed all the callers have `s.server.txMemPool` and `s.server.sigCache` for the params.

txscript/reference_test.go
davecgh: I see you noted fixed, but it's still here.  Lost in a rebase?

txscript/reference_test.go
```diff
@@ -190,40 +190,53 @@ func TestScriptInvalidTests(t *testing.T) {
...
-		if err == nil {
-			if err := vm.Execute(); err == nil {
-				t.Errorf("%s test succeeded when it "+
-					"should have failed\n", name)
+	sigCache := NewSigCache(10)
+	sigCacheToggle := []bool{true, false}
+	for _, useSigCache := range sigCacheToggle {
+		for i, test := range tests {
+			var vm *Engine
+			var err error
```
davecgh: And error is still here.

txscript/sigcache.go
```diff
@@ -0,0 +1,119 @@
...
+
+// SigCache implements an ECDSA signature verification cache with a randomized
+// entry eviction policy. The benefits of SigCache are two fold. Firstly, usage
+// of SigCache mitigates a DoS attack wherein an attack causes a victim's client
+// to hang due to worst-case behavior triggered while processing attacker
+// crafted invalid transactions. A detailed description of the mitigated DoS
+// attack can be found here: https://bitslog.wordpress.com/2013/01/23/fixed-bitcoin-vulnerability-explanation-why-the-signature-cache-is-a-dos-protection/.
+// Secondly, usage of the SigCache introduces a signature verification
+// optimization which speeds up the validation of transactions within a block,
+// if they've already once been seen and verified within the mempool.
```
davecgh: I think "once" is a typo here.

txscript/sigcache.go
```diff
@@ -0,0 +1,119 @@
...
+type SigCache struct {
+	sync.RWMutex
+	validSigs  map[sigInfo]struct{}
+	maxEntries int
+}
+
+// NewSigCache creates and initializes a new instance of SigCache. Its sole
+// parameter 'maxEntries' represents the maximum number of entries allowed to
+// exist in the SigCache and any particular moment. If the addition of a new
+// 'sigInfo' entry puts the number of entries in the SigCache over
```
davecgh: Since `sigInfo` is an internal type, it probably doesn't make sense to make reference to it on the exported godoc.  Perhaps:

`Random entries are evicted to make room for new entries that would cause the number of entries in the cache to exceed the max.`

config.go
```diff
@@ -117,6 +118,7 @@ type config struct {
 	AddrIndex          bool          `long:"addrindex" description:"Build and maintain a full address index. Currently only supported by leveldb."`
 	DropAddrIndex      bool          `long:"dropaddrindex" description:"Deletes the address-based transaction index from the database on start up, and the exits."`
 	NoPeerBloomFilters bool          `long:"nopeerbloomfilters" description:"Disable bloom filtering support."`
+	SigCacheMaxSize    int           `long:"sigcachemaxsize" description:"The maximum number of entries in the signature verification cache."`
```
davecgh: I'd make this a `uint` so it will disallow negative numbers for you.

txscript/sigcache.go
```diff
@@ -0,0 +1,119 @@
...
+// to hang due to worst-case behavior triggered while processing attacker
+// crafted invalid transactions. A detailed description of the mitigated DoS
+// attack can be found here: https://bitslog.wordpress.com/2013/01/23/fixed-bitcoin-vulnerability-explanation-why-the-signature-cache-is-a-dos-protection/.
+// Secondly, usage of the SigCache introduces a signature verification
+// optimization which speeds up the validation of transactions within a block,
+// if they've already once been seen and verified within the mempool.
+type SigCache struct {
+	sync.RWMutex
+	validSigs  map[sigInfo]struct{}
+	maxEntries int
```
davecgh: Probably should be `uint` since I don't think a negative number of max entries makes sense.

txscript/sigcache.go
```diff
@@ -0,0 +1,119 @@
...
+// existing entry it randomly chosen to be evicted in order to make space for
+// the new entry.
+//
+// NOTE: This function is safe for concurrent access. Writers will block
+// simultaneous readers until function execution has concluded.
+func (s *SigCache) Add(sigHash []byte, sig *btcec.Signature, pubKey *btcec.PublicKey) {
+	s.Lock()
+	defer s.Unlock()
+
+	if s.maxEntries <= 0 {
```
davecgh: If `maxEntries` is changed to uint, this could safely be changed to `== 0`, although it probably wouldn't hurt leaving the `<`.

txscript/sigcache.go
```diff
@@ -0,0 +1,119 @@
...
+
+// Exists returns true if an existing entry of 'sig' over 'sigHash' for public
+// key 'pubKey' is found within the SigCache. Otherwise, false is returned.
+//
+// NOTE: This function is safe for concurrent access. Readers won't be blocked
+// unless there exists a writer, adding an entry to the SigCache.
+func (s *SigCache) Exists(sigHash []byte, sig *btcec.Signature, pubKey *btcec.PublicKey) bool {
+	var candidateHash wire.ShaHash
+	copy(candidateHash[:], sigHash)
+
```
davecgh: Nitpick: This has a blank line while the equivalent code below does not.

txscript/sigcache_test.go
```diff
@@ -0,0 +1,141 @@
...
+		msg, sig, key, err := genRandomSig()
+		if err != nil {
+			t.Fatalf("unable to generate random signature test data")
+		}
+
+		sigCache.Add(msg, sig, key)
+
+		sigCopy, _ := btcec.ParseSignature(sig.Serialize(), btcec.S256())
+		keyCopy, _ := btcec.ParsePubKey(key.SerializeCompressed(), btcec.S256())
+
```
davecgh: Inconsistent blank line.

txscript/sigcache_test.go
```diff
@@ -0,0 +1,141 @@
...
+	// The sigcache should still have sigCache entries.
+	if len(sigCache.validSigs) != sigCacheSize {
+		t.Fatalf("sigcache should now have %v entries, instead it has %v",
+			sigCacheSize, len(sigCache.validSigs))
+	}
+
+	sigNewCopy, _ := btcec.ParseSignature(sigNew.Serialize(), btcec.S256())
+	keyNewCopy, _ := btcec.ParsePubKey(keyNew.SerializeCompressed(), btcec.S256())
+
```
davecgh: Inconsistent blank line.  The copies are created after the comment in the code below.

davecgh:
I'm done with this round of review.  It looks like functionally, everything is good to go.  I'm going to do some manual testing on it as well.

txscript/reference_test.go
Roasbeef: Hmm seems some line numbers got mixed up. I corrected the issue in `TestScriptValidTests`, but not `TestScriptInvalidTests`. Should be all remedied now in the latest snapshot.

txscript/sigcache.go
Roasbeef: Yeah it reads better without "once". I removed updated the comment to remove the extraneous word.

txscript/sigcache.go
Roasbeef: Agreed, updated the latter half of the comment with your suggestion.

config.go
Roasbeef: Agreed, changed to a `uint`.

txscript/sigcache.go
Roasbeef: Fixed.

txscript/sigcache.go
Roasbeef: I've changed `maxEntries` to be a `uint`, opting to leave the `<`.

txscript/sigcache.go
Roasbeef: Removed extra blank line to remain consistent with the code fragment below.

txscript/sigcache_test.go
Roasbeef: Removed.

txscript/sigcache_test.go
Roasbeef: Fixed.

mining.go
Roasbeef: I've changed `NewBlockTemplate` to directly take the `server` as a parameter as suggested.

txscript/sigcache_test.go
```diff
@@ -0,0 +1,139 @@
...
+	if !sigCache.Exists(msg1, sig1Copy, key1Copy) {
+		t.Errorf("previously added item not found in signature cache")
+	}
+}
+
+// TestSigCacheAddEvictEntry tests the eviction case where a new signature
+// triplet is added to a full signature cache which should trigger randomized
+// eviction, followed by adding the new element to the cache.
+func TestSigCacheAddEvictEntry(t *testing.T) {
+	// Create a sigcache that can hold up to 10 entries.
```
davecgh: Comment is incorrect here.

davecgh:
Other than the comment I mentioned, this looks good to go.  I've done a full chain download with checkpoints disabled (to force every script in the chain to be executed).  I've also confirmed that it's using cached signatures (by adding a temporary logging statement) for transactions in the memory pool that are included in new block as expected.

Please fix the comment and rebase/squash to latest master.

txscript/opcode.go
```diff
@@ -1850,7 +1850,17 @@ func opcodeCheckSig(op *parsedOpcode, vm *Engine) error {
 		return nil
 	}
 
-	ok := signature.Verify(hash, pubKey)
+	var ok bool
+	if vm.sigCache != nil {
+		ok = vm.sigCache.Exists(hash, signature, pubKey)
```
davecgh: This assumes existence in the signature cache means it is a valid signature and the below only adds the result to the cache if it's a valid signature.  I'm fairly certain this means that invalid signatures will have to be re-checked which partially defeats the DoS protection aspects.

txscript/opcode.go
```diff
@@ -2051,7 +2061,18 @@ func opcodeCheckMultiSig(op *parsedOpcode, vm *Engine) error {
 		// Generate the signature hash based on the signature hash type.
 		hash := calcSignatureHash(script, hashType, &vm.tx, vm.txIdx)
 
-		if parsedSig.Verify(hash, parsedPubKey) {
+		var valid bool
+		if vm.sigCache != nil {
+			valid = vm.sigCache.Exists(hash, parsedSig, parsedPubKey)
```
davecgh: The same is true here as above regarding existence signalling the signature is valid.

davecgh:
For reference, I've done a full chain download with checkpoints disabled (to force every script in the chain to be executed), and I've also confirmed that it's using cached signatures (by adding a temporary logging statement) for transactions in the memory pool that are included in new block as expected.

Aside from the incorrect comment, this is ready.  Please fix the comment and rebase/squash to the latest master.

Roasbeef:
I've fixed the comment (also clarifying that the cache is only for valid signatures), and rebase/squashed to the latest master :+1:

davecgh:
OK

txscript/sigcache.go
```diff
@@ -0,0 +1,118 @@
...
+// benefits of SigCache are two fold. Firstly, usage of SigCache mitigates a DoS
+// attack wherein an attack causes a victim's client to hang due to worst-case
+// behavior triggered while processing attacker crafted invalid transactions. A
+// detailed description of the mitigated DoS attack can be found here: https://bitslog.wordpress.com/2013/01/23/fixed-bitcoin-vulnerability-explanation-why-the-signature-cache-is-a-dos-protection/.
+// Secondly, usage of the SigCache introduces a signature verification
+// optimization which speeds up the validation of transactions within a block,
+// if they've already been seen and verified within the mempool.
+type SigCache struct {
+	sync.RWMutex
+	validSigs  map[sigInfo]struct{}
```
jrick: This might not work as you expect since sigInfo contains string fields (the shallow equality operator might not work if the two strings are different e.g. one is a static string, and the other was created at runtime).

txscript/sigcache.go
```diff
@@ -0,0 +1,118 @@
...
+
+		// Try to find the first entry that is greater than the random
+		// hash. Use the first entry (which is already pseudo random due
+		// to Go's range statement over maps) as a fall back if none of
+		// the hashes in the rejected transactions pool are larger than
+		// the random hash.
+		var foundEntry *sigInfo
+		for sigEntry := range s.validSigs {
+			if foundEntry == nil {
+				foundEntry = &sigEntry
```
jrick: I would make a copy of the entry here and use that for the foundEntry (and below this too, after the byte comparison), otherwise the foundEntry will keep changing every iteration (range will overwrite it, because you're pointing to the iterator's current element).

txscript/sigcache.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// Add adds an entry for a signature over 'sigHash' under public key 'pubKey'
+// to the signature cache. In the event that the SigCache is 'full', an
+// existing entry it randomly chosen to be evicted in order to make space for
+// the new entry.
+//
+// NOTE: This function is safe for concurrent access. Writers will block
+// simultaneous readers until function execution has concluded.
+func (s *SigCache) Add(sigHash []byte, sig *btcec.Signature, pubKey *btcec.PublicKey) {
```
jrick: A slice for the sigHash is reasonable if you need then extra performance of not creating a new wire.ShaHash array, but at the very least this function needs to check that the length is correct.  As it is, I can pass something not 32 bytes long, and this function will either extend or truncate it to a ShaHash when the sigInfo is saved.

txscript/sigcache.go
jrick: nevermind: http://play.golang.org/p/Phwr-adqy7

txscript/sigcache.go
davecgh: It will break out of the loop, thereby leaving it at the correct entry when a match is found, so this works properly.

txscript/opcode.go
jrick: For this same dance below, in opcodeCheckMultiSig, you use valid instead of ok.  Better to keep them consistent, better yet to push this out to a new function.

txscript/sigcache_test.go
```diff
@@ -0,0 +1,139 @@
...
+	key1Copy, _ := btcec.ParsePubKey(key1.SerializeCompressed(), btcec.S256())
+	if sigCache.Exists(msg1, sig1Copy, key1Copy) {
+		t.Errorf("previously added not found in sigcache, but shouldn't" +
+			" have been")
+	}
+
+	// There shouldn't be any entries in the sigCache.
+	if len(sigCache.validSigs) != 0 {
+		t.Errorf("items found in sigcache, no items should have been added")
+	}
```
dajohi: please report how many items were found.

txscript/sigcache_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// Add the triplet to the signature cache.
+	sigCache.Add(msg1, sig1, key1)
+
+	// The generated triplet should not be found.
+	sig1Copy, _ := btcec.ParseSignature(sig1.Serialize(), btcec.S256())
+	key1Copy, _ := btcec.ParsePubKey(key1.SerializeCompressed(), btcec.S256())
+	if sigCache.Exists(msg1, sig1Copy, key1Copy) {
+		t.Errorf("previously added not found in sigcache, but shouldn't" +
```
dajohi: s/not/signature

txscript/sigcache.go
jrick: Sure but even though this code reads as if it will default to the first item it ranged over, it won't (it would be the last).  It's probably less performant too because it has to re-write to the heap allocation each and every loop iteration.

txscript/opcode.go
```diff
@@ -1850,7 +1850,17 @@ func opcodeCheckSig(op *parsedOpcode, vm *Engine) error {
...
+	var ok bool
+	if vm.sigCache != nil {
+		ok = vm.sigCache.Exists(hash, signature, pubKey)
+		if !ok && signature.Verify(hash, pubKey) {
+			vm.sigCache.Add(hash, signature, pubKey)
+			ok = true
+		}
+	} else {
+		ok = signature.Verify(hash, pubKey)
+	}
```
jrick: also I think this could would be a little more readable as:

```Go
valid := false
if vm.sigCache != nil {
    valid = vm.sigCache.Exists(...)
}
if !valid {
    valid = signature.Verify(...)
    if valid {
        vm.sigCache.Add(...)
    }
}
```

txscript/sigcache.go
```diff
@@ -0,0 +1,118 @@
...
+	"bytes"
+	"crypto/rand"
+	"sync"
+
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/btcsuite/btcd/wire"
+)
+
+// sigInfo represents an entry in the SigCache. Entries in the sigcache are a
+// 3-tuple: (sigCache, sig, pubKey).
```
kargakis: s/sigCache/sigHash

txscript/sigcache.go
```diff
@@ -0,0 +1,118 @@
...
+	sig     string
+	pubKey  string
+}
+
+// SigCache implements an ECDSA signature verification cache with a randomized
+// entry eviction policy. Only valid signatures will be added to the cache. The
+// benefits of SigCache are two fold. Firstly, usage of SigCache mitigates a DoS
+// attack wherein an attack causes a victim's client to hang due to worst-case
+// behavior triggered while processing attacker crafted invalid transactions. A
+// detailed description of the mitigated DoS attack can be found here: https://bitslog.wordpress.com/2013/01/23/fixed-bitcoin-vulnerability-explanation-why-the-signature-cache-is-a-dos-protection/.
```
kargakis: Mind putting the link in its own line?

txscript/sigcache.go
wallclockbuilder: an existing entry ~~it~~ is randomly chosen to be evicted in order to make space for the new entry.

txscript/sigcache.go
Roasbeef: Done.

txscript/sigcache.go
Roasbeef: Nice catch! Fixed the comment.

txscript/opcode.go
Roasbeef: I disagree. As written now, both cases (`sigCache` vs no `sigCache`) are logically separated and handled distinctly.

txscript/sigcache_test.go
Roasbeef: Fixed.

txscript/sigcache_test.go
Roasbeef: Updated the error string to include how many items were found.

txscript/opcode.go
Roasbeef: I've opted to uniformly use `valid` both here and in `opcodeCheckMultiSig`. But, I don't see any gains to be hand by pushing out the fragment to a new function.

txscript/sigcache.go
Roasbeef: I've changed the function to return an error if passed a `sigHash` that isn't exactly 32 bytes. As a result, `sigCache.Add` now once again returns an error (which is always ignored in `opcode.go` when executing).

txscript/sigcache.go
Roasbeef: Excellent observation. I've changed the loop to instead use a copy, rather than pointing to the iterator's current element.

txscript/sigcache.go
```diff
@@ -0,0 +1,124 @@
...
+func (s *SigCache) Add(sigHash []byte, sig *btcec.Signature, pubKey *btcec.PublicKey) error {
+	s.Lock()
+	defer s.Unlock()
+
+	if s.maxEntries <= 0 {
+		return nil
+	}
+
+	if len(sigHash) != wire.HashSize {
+		return errors.New("sigHash must be a byte slice of length 32")
```
davecgh: This assumes the size of the constant in the error message and could get out of sync.

`return fmt.Errorf("sigHash must be a byte slice of length %d", wire.HashSize)`

I'd personally prefer the parameter take a wire.ShaHash directly so there isn't even a reason to have an error path at all and it would make it more explicit what type of hash is expected here given there is more than one type used in scripts (sha256(x), sha256(sha256(x)), ripemd160(sha256(x)), etc).

That said this way is fine too as long as the error message adapts to the constant.

Roasbeef: Agreed, it's better to make the interface to `SigCache` more explicit. In the latest commit, I've made `SigCache.Add` and `SigCache.Exists` simply take a `wire.Shahash` directly (eliminating the error path).

davecgh:
Latest changes look good.

OK after rebase/squash.

PR #507, opened 2015-09-26
wire: Reject non-canonically encoded varints.

davecgh:
The Bitcoin wire protocol includes several fields with their lengths encoded according to a variable length integer encoding scheme that does not enforce a unique encoding for all numbers.

This can lead to a situation where deserializing and re-serializing the same data can result in different bytes.  There are no currently known issues due to this, but it is safer to reject such subtle differences as
they could potentially lead to exploits.

Consequently, this pull request modifies the varint decoding function to error when the value is not canonically encoded which effectively means that all messages with varints that are not canonically encoded will now be rejected.  This will not cause issues with old client versions in regards to blocks and transactions since the data is deserialized into memory and then reserialized before being relayed thereby effectively erasing any non-canonical encodings.

Also, new tests have been added to ensure non-canonical encodings are properly rejected and exercise the new code, and the default user agent version for wire has been bumped to version 0.2.1 to differentiate the new behavior.

The equivalent logic was implemented in Bitcoin Core by PR 2884.

### DIFF
--- a/wire/common.go
+++ b/wire/common.go
@@ -17,6 +17,11 @@ import (
 // Maximum payload size for a variable length integer.
 const MaxVarIntPayload = 9
 
+// errNonCanonicalVarInt is the common format string used for non-canonically
+// encoded variable length integer errors.
+var errNonCanonicalVarInt = "non-canonical varint %x - discriminant %x must " +
+	"encode a value greater than %x"
+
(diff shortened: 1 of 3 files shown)


jrick:
ok

dajohi:
Looks good.  OK.

PR #508, opened 2015-09-28
txscript: Optimize DisasmString function.

davecgh:
This pull request modifies the `DisasmString` function to use a bytes buffer for constructing the disassembled string instead of naive string concatenation.  This makes a huge difference when disassembling scripts with large numbers of opcodes.


dajohi:
OK

PR #509, opened 2015-09-28
rpcserver: Optimize JSON raw tx input list create.

davecgh:
This pull request optimizes the `createVinList` function which is used to generate the JSON list of transaction inputs.  It also makes it more consistent with the `createVinListPrevOut` function.

In particular, it entails the following changes:
- Only do a single coinbase check and return right away instead of checking multiple times inside the loop over the inputs
- Use a pointer for populating the details of each entry to avoid multiple unnecessary array lookups and bounds checks
- Group all fields that populate the entry for better readability


dajohi:
OK

PR #511, opened 2015-10-09
btcjson: Add errors to InfoChainResult

dajohi:
The getinfo RPC will now include the errors attribute in the
result.


davecgh:
OK

PR #512, opened 2015-10-09
txscript:  Add ScriptVerifyLowS to the standard flags

dajohi:
We've already been generating lowS sigs for quite a while.  This removes
the malleability vector.

This mimics Bitcoin Core commit 49dd5c629df0a08cf3b1ea8085c03312d1a81696


davecgh:
OK

PR #514, opened 2015-10-16
wire: Export var length string serialization funcs.

davecgh:
This pull request exports the `ReadVarString` and `WriteVarString` functions so they are available for callers to use.

A variable length string is encoded as a variable length integer containing the length of the string followed by the bytes that representthe string itself.


jrick:
ok

dajohi:
OK

PR #515, opened 2015-10-16
rpcserver: Copy btcwallet fix for verifymessage RPC.

jrick:


davecgh:
I've tested this works properly using signed messages created by `signmessage` using both btcwallet and Bitcoin Core.

OK.

PR #516, opened 2015-10-17
adds filteraddrs param to searchrawtransactions API

dan-da:
# abstract
The new filteraddrs parameter enables the caller to pass a list of addresses to the searchrawtransactions API. Only inputs and outputs that match addresses in the list will be returned.

Additionally, if at least 1 filter address is passed then the (loooong) Hex attribute of the transaction will be empty.  ( though key is still present. )

# motivation
To increase efficiency of the API particularly in terms of bandwidth usage, processing overhead for the caller, and overall roundtrip time.    The API is presently returning datasets that are 20+ MB for some addresses when the application may need only a tiny fraction of that.

# example
My application displays a transaction history for one or more addresses.  To do this, it calls searchrawtransaction once per address.  It only needs to see the vin/vout entries that pay to or from that address.  Below we call this address passed as the first param to searchrawtransactions the "target address".

For bitcoin addresses with many inputs and outputs such as those involved in mining pool payouts, the bulk of the inputs/outputs are NOT related to the target address.  All this extra data is unnecessary and wasteful.

Additionally, my application does not use the Hex attribute of the transaction, but a large amount of unused data is returned with it.

Using the new filteraddrs param, my application simply passes the target address as the only element in the filteraddrs array, and the majority of unused data disappears.

This implementation makes it possible to filter vin/vout by an address other than the target address.  Or by multiple addresses.  But likely the most common use-case will be to filter by target address only.

# how much savings?
Obviously the savings will vary depending on the address.  But the savings can be very significant.  For one address that I know was receiving mining pool payouts for a couple months, the total size of returned JSON is:
  without filteraddrs: 22938656 bytes,  6.911 seconds
  with filteraddrs: 123229 bytes, 2.087 seconds

So 22.9 MB vs 123KB.  That's 186 times smaller.  Also a 3x roundtrip speedup on localhost.  The speedup would of course be larger if making the call over the internet given the size of data being transferred.

Also, many addresses exist with many times more transactions and vin/vout.  I expect there are addresses that would return 100+ MB result sets.

# testing performed

1) Updated test cases for searchrawtransactions and added 1 new test case for the filteraddrs param.
2) passes goclean.sh tests.
3) Verified manually that the API works as expected when called (a) without filteraddrs param, (b) with filteraddrs set to nil, (c) with filteraddrs set to array containing target address only, (d) with filteraddrs set to array with multiple addresses from the result set.
4) Used in my application and verified that the transaction history appears the same with or without the filterAddrs param in use.

# further thoughts / notes

## bool flag vs array/list
If we assume that the target address is going to be the desired filter 99.99% of the time then it is simpler just to use a boolean "filterAddr=true" than a list.  In fact, my initial implementation did this.  However, for public use I imagine there will likely be other use cases, and the performance is about the same, so I decided to expose a filterlist to the caller.

## empty filter list behavior
In this implementation the behavior is the same if filteraddrs arg is omitted or if an empty array is passed.  In either case, all vin/vout and also the tx.Hex attribute are returned as per legacy behavior.

It is possible that a caller might wish to omit ALL vin/vout and might try passing an empty array to achieve this.  But that won't work.   A workaround for this person would be to pass a single impossible address.  That will never match anything, so all will be filtered out.

Alternatively, it would be possible to change the behavior so that an empty array filters out all, but nil/null filters nothing.

## Tx.Hex attribute

This patch presently includes the hex attribute but with an empty value when filterAddrs contains 1 or more values.  We could hide/remove it entirely via the json omitempty flag.   Alternatively, this change could be de-bundled from the patch and be part of a more comprehensive "user chooses fields" feature.  see below.

## further size savings
Additional savings are possible.  I tried to make this patch conservative by removing only the worst offenders and thereby getting the most bang for the buck.  A possible future approach could be to allow the caller to specify which fields to receive/omit for the tx, vins, vouts via a separate parameter.  This would be similar to a SQL select list.  I'm not proposing to work on this, as the present patch is good enough for my needs.  Just throwing it out there as food for thought.

## parameter creep
searchrawtransactions is such a useful function it would not surprise me if future parameters get added, and each one changes the signature and adds new default.  So an idea for the dev team to consider would be creating a new param called "options" which is a struct.  So adding new params is as simple as adding a new key to this struct, and callers do not have to specify all the defaults for previous params.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/516)

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -664,22 +664,41 @@ func createVinList(mtx *wire.MsgTx) []btcjson.Vin {
 	return vinList
 }
 
+// stringInSlice returns true if string a is found in array list.
+func stringInSlice(a string, list []string) bool {
+	for _, b := range list {
+		if b == a {
+			return true
+		}
+	}
+	return false
+}
+
 // createVinList returns a slice of JSON objects for the inputs of the passed
 // transaction.
-func createVinListPrevOut(s *rpcServer, mtx *wire.MsgTx, chainParams *chaincfg.Params, vinExtra int) []btcjson.VinPrevOut {
+func createVinListPrevOut(s *rpcServer, mtx *wire.MsgTx, chainParams *chaincfg.Params, vinExtra int, filterAddrMap map[string]struct{}) []btcjson.VinPrevOut {
+	// We use a dynamically sized list to accomodate address filter.
+	vinList := make([]btcjson.VinPrevOut, 0, len(mtx.TxIn))
+
 	// Coinbase transactions only have a single txin by definition.
-	vinList := make([]btcjson.VinPrevOut, len(mtx.TxIn))
 	if blockchain.IsCoinBaseTx(mtx) {
+		// include tx only if filterAddrMap is empty because coinbase has no address
+		// and so would never match a non-empty filter.
+		if len(filterAddrMap) != 0 {
+			return vinList
+		}
+		var vinEntry btcjson.VinPrevOut
 		txIn := mtx.TxIn[0]
-		vinList[0].Coinbase = hex.EncodeToString(txIn.SignatureScript)
-		vinList[0].Sequence = txIn.Sequence
+		vinEntry.Coinbase = hex.EncodeToString(txIn.SignatureScript)
+		vinEntry.Sequence = txIn.Sequence
+		vinList = append(vinList, vinEntry)
 		return vinList
 	}
 
 	// Lookup all of the referenced transactions needed to populate the
 	// previous output information if requested.
 	var txStore blockchain.TxStore
-	if vinExtra != 0 {
+	if vinExtra != 0 || len(filterAddrMap) > 0 {
 		tx := btcutil.NewTx(mtx)
 		txStoreNew, err := s.server.txMemPool.fetchInputTransactions(tx, true)
 		if err == nil {
@@ -687,26 +706,15 @@ func createVinListPrevOut(s *rpcServer, mtx *wire.MsgTx, chainParams *chaincfg.P
 		}
 	}
 
-	for i, txIn := range mtx.TxIn {
+	for _, txIn := range mtx.TxIn {
+		// reset filter flag for each.
+		passesFilter := len(filterAddrMap) == 0
+
 		// The disassembled string will contain [error] inline
 		// if the script doesn't fully parse, so ignore the
 		// error here.
 		disbuf, _ := txscript.DisasmString(txIn.SignatureScript)
 
-		vinEntry := &vinList[i]
-		vinEntry.Txid = txIn.PreviousOutPoint.Hash.String()
-		vinEntry.Vout = txIn.PreviousOutPoint.Index
-		vinEntry.Sequence = txIn.Sequence
-		vinEntry.ScriptSig = &btcjson.ScriptSig{
-			Asm: disbuf,
-			Hex: hex.EncodeToString(txIn.SignatureScript),
-		}
-
-		// Only populate previous output information if requested and
-		// available.
-		if vinExtra == 0 || len(txStore) == 0 {
-			continue
-		}
 		txData := txStore[txIn.PreviousOutPoint.Hash]
 		if txData == nil {
 			continue
@@ -717,20 +725,42 @@ func createVinListPrevOut(s *rpcServer, mtx *wire.MsgTx, chainParams *chaincfg.P
 		// Ignore the error here since an error means the script
 		// couldn't parse and there is no additional information about
 		// it anyways.
-		var strAddrs []string
 		_, addrs, _, _ := txscript.ExtractPkScriptAddrs(
 			originTxOut.PkScript, chainParams)
-		if addrs != nil {
-			strAddrs = make([]string, len(addrs))
-			for j, addr := range addrs {
-				strAddrs[j] = addr.EncodeAddress()
+
+		encodedAddrs := make([]string, len(addrs))
+		for j, addr := range addrs {
+			encodedAddrs[j] = addr.EncodeAddress()
+
+			if len(filterAddrMap) > 0 {
+				if _, exists := filterAddrMap[encodedAddrs[j]]; exists {
+					passesFilter = true
+				}
 			}
 		}
 
-		vinEntry.PrevOut = &btcjson.PrevOut{
-			Addresses: strAddrs,
-			Value:     btcutil.Amount(originTxOut.Value).ToBTC(),
+		if !passesFilter {
+			continue
 		}
+
+		var vinEntry btcjson.VinPrevOut
+		vinEntry.Txid = txIn.PreviousOutPoint.Hash.String()
+		vinEntry.Vout = txIn.PreviousOutPoint.Index
+		vinEntry.Sequence = txIn.Sequence
+		vinEntry.ScriptSig = &btcjson.ScriptSig{
+			Asm: disbuf,
+			Hex: hex.EncodeToString(txIn.SignatureScript),
+		}
+
+		// Only populate previous output information if requested
+		if vinExtra != 0 {
+			vinEntry.PrevOut = &btcjson.PrevOut{
+				Addresses: encodedAddrs,
+				Value:     btcutil.Amount(originTxOut.Value).ToBTC(),
+			}
+		}
+
+		vinList = append(vinList, vinEntry)
 	}
 
 	return vinList
@@ -738,34 +768,47 @@ func createVinListPrevOut(s *rpcServer, mtx *wire.MsgTx, chainParams *chaincfg.P
 
 // createVoutList returns a slice of JSON objects for the outputs of the passed
 // transaction.
-func createVoutList(mtx *wire.MsgTx, chainParams *chaincfg.Params) []btcjson.Vout {
-	voutList := make([]btcjson.Vout, len(mtx.TxOut))
+func createVoutList(mtx *wire.MsgTx, chainParams *chaincfg.Params, filterAddrMap map[string]struct{}) []btcjson.Vout {
+	voutList := make([]btcjson.Vout, 0, len(mtx.TxOut))
 	for i, v := range mtx.TxOut {
-		voutList[i].N = uint32(i)
-		voutList[i].Value = btcutil.Amount(v.Value).ToBTC()
+		// reset filter flag for each.
+		passesFilter := len(filterAddrMap) == 0
 
 		// The disassembled string will contain [error] inline if the
 		// script doesn't fully parse, so ignore the error here.
 		disbuf, _ := txscript.DisasmString(v.PkScript)
-		voutList[i].ScriptPubKey.Asm = disbuf
-		voutList[i].ScriptPubKey.Hex = hex.EncodeToString(v.PkScript)
 
 		// Ignore the error here since an error means the script
 		// couldn't parse and there is no additional information about
 		// it anyways.
 		scriptClass, addrs, reqSigs, _ := txscript.ExtractPkScriptAddrs(
 			v.PkScript, chainParams)
-		voutList[i].ScriptPubKey.Type = scriptClass.String()
-		voutList[i].ScriptPubKey.ReqSigs = int32(reqSigs)
 
-		if addrs == nil {
-			voutList[i].ScriptPubKey.Addresses = nil
-		} else {
-			voutList[i].ScriptPubKey.Addresses = make([]string, len(addrs))
-			for j, addr := range addrs {
-				voutList[i].ScriptPubKey.Addresses[j] = addr.EncodeAddress()
+		encodedAddrs := make([]string, len(addrs))
+		for j, addr := range addrs {
+			encodedAddrs[j] = addr.EncodeAddress()
+
+			if len(filterAddrMap) > 0 {
+				if _, exists := filterAddrMap[encodedAddrs[j]]; exists {
+					passesFilter = true
+				}
 			}
 		}
+
+		if !passesFilter {
+			continue
+		}
+
+		var vout btcjson.Vout
+		vout.N = uint32(i)
+		vout.Value = btcutil.Amount(v.Value).ToBTC()
+		vout.ScriptPubKey.Addresses = encodedAddrs
+		vout.ScriptPubKey.Asm = disbuf
+		vout.ScriptPubKey.Hex = hex.EncodeToString(v.PkScript)
+		vout.ScriptPubKey.Type = scriptClass.String()
+		vout.ScriptPubKey.ReqSigs = int32(reqSigs)
+
+		voutList = append(voutList, vout)
 	}
 
 	return voutList
@@ -775,18 +818,24 @@ func createVoutList(mtx *wire.MsgTx, chainParams *chaincfg.Params) []btcjson.Vou
 // to a raw transaction JSON object, possibly with vin.PrevOut section.
 func createSearchRawTransactionsResult(s *rpcServer, chainParams *chaincfg.Params, mtx *wire.MsgTx,
 	txHash string, blkHeader *wire.BlockHeader, blkHash string,
-	blkHeight int32, chainHeight int32, vinExtra int) (*btcjson.SearchRawTransactionsResult, error) {
+	blkHeight int32, chainHeight int32, vinExtra int, filterAddrMap map[string]struct{}) (*btcjson.SearchRawTransactionsResult, error) {
 
-	mtxHex, err := messageToHex(mtx)
-	if err != nil {
-		return nil, err
+	// omit hex if filterAddrMap are present.  When filtering, typically the
+	// goal is to reduce unnecessary bloat in the result.
+	var mtxHex string
+	if len(filterAddrMap) == 0 {
+		mtxHexTmp, err := messageToHex(mtx)
+		if err != nil {
+			return nil, err
+		}
+		mtxHex = mtxHexTmp
 	}
 
 	txReply := &btcjson.SearchRawTransactionsResult{
 		Hex:      mtxHex,
 		Txid:     txHash,
-		Vout:     createVoutList(mtx, chainParams),
-		Vin:      createVinListPrevOut(s, mtx, chainParams, vinExtra),
+		Vout:     createVoutList(mtx, chainParams, filterAddrMap),
+		Vin:      createVinListPrevOut(s, mtx, chainParams, vinExtra, filterAddrMap),
 		Version:  mtx.Version,
 		LockTime: mtx.LockTime,
 	}
@@ -816,7 +865,7 @@ func createTxRawResult(chainParams *chaincfg.Params, mtx *wire.MsgTx,
 	txReply := &btcjson.TxRawResult{
 		Hex:      mtxHex,
 		Txid:     txHash,
-		Vout:     createVoutList(mtx, chainParams),
+		Vout:     createVoutList(mtx, chainParams, nil),
 		Vin:      createVinList(mtx),
 		Version:  mtx.Version,
 		LockTime: mtx.LockTime,
@@ -861,7 +910,7 @@ func handleDecodeRawTransaction(s *rpcServer, cmd interface{}, closeChan <-chan 
 		Version:  mtx.Version,
 		Locktime: mtx.LockTime,
 		Vin:      createVinList(&mtx),
-		Vout:     createVoutList(&mtx, s.server.chainParams),
+		Vout:     createVoutList(&mtx, s.server.chainParams, nil),
 	}
 	return txReply, nil
 }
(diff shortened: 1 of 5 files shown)


davecgh:
Thanks for the pull request.  I haven't reviewed the code yet, but I think the idea is a good one.

I am a little concerned about the parameter creep, as you pointed out.  I believe it has had 3 (and this would be the 4th) new parameter added in the last few months.  The way optional parameters are handled is one of the things I really dislike about JSON-RPC 1.0, but unfortunately it's required for compatibility purposes.  I suspect adding an options struct could help, although that would also have the downside of making it slightly harder to specify via the command-line such as with `btcctl`.  The said, it's probably easier than having to specify defaults for all of the other optional parameter you probably don't really care about due to the positional requirements.

dan-da:
Yeah, so I'm thinking that a simple and backward-compatible way forward would be to accept this pull request basically as-is, then create a new API "searchrawtransactionsv2" or whatever you want to call it that accepts only two params:  addr and options.  Any new options would go into the new API, and the old one could be frozen/deprecated.

Ideally addr would be changed to addr_list, in order to more efficiently retrieve transactions for multiple addresses at once.  ( Bitpay's Insight API already supports this. )  With this change in mind, perhaps the new API could be called searchtransactionsforaddresses or something like that and offer more selectivity in which fields are returned.

up to you... I'm just brainstorming.  :-)

davecgh:
For an update to this, I haven't forgot about it, just been rather busy with the PR barrage as you've probably noted.

dan-da:
thx for the update.  Yeah, I'm not in a real big rush about this.   Though my tool/app does use the filter and it would be nice if I could announce it on reddit one of these days and have it work with official btcd instead of a forked copy.  So that was my primary incentive for finishing up the pull request.

rpcserver.go
```diff
@@ -667,13 +685,20 @@ func createVinListPrevOut(s *rpcServer, mtx *wire.MsgTx, chainParams *chaincfg.P
...
-	for i, txIn := range mtx.TxIn {
+	for _, txIn := range mtx.TxIn {
+		var vinEntry btcjson.VinPrevOut
+
+		// reset filter flag for each vin.
+		passesFilter := true
+		if len(filterAddrs) > 0 {
+			passesFilter = false
+		}
+
```
dajohi: perhaps

```go
passesFilter := len(filterAddrs) == 0
```

rpcserver.go
```diff
@@ -718,34 +751,50 @@ func createVinListPrevOut(s *rpcServer, mtx *wire.MsgTx, chainParams *chaincfg.P
...
+	voutList := []btcjson.Vout{}
 	for i, v := range mtx.TxOut {
-		voutList[i].N = uint32(i)
-		voutList[i].Value = btcutil.Amount(v.Value).ToBTC()
+		var vout btcjson.Vout
+		passesFilter := true
+		if len(filterAddrs) > 0 {
+			passesFilter = false
+		}
+
```
dajohi: same as comment above

dajohi:
I am OK with this diff.  My comments above are merely style.

dan-da:
@dajohi thanks for the review.  I made the change you suggested.  The multiline if() bothered me too.  I'm still getting used to Go and wanted to use the ? : operator which doesn't seem to exist.  Your suggestion is a nice shortcut.

davecgh:
Reviewed 3 of 4 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 13 unresolved discussions.

---

<sup>**[btcjson/chainsvrcmds.go, line 554 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K2nRMPZh-GkfCoK_9fQ)** ([raw file](https://github.com/btcsuite/btcd/blob/23eb93bdadb538560f14110b2d675ccae781941a/btcjson/chainsvrcmds.go#L554)):</sup>
What do you think about making this a `*map[string]struct{}` instead of a `*[]string`?  My thinking is it would automatically prevent duplicate addresses, something the current approach doesn't deal with, and it would also generally make checking if an address was specified faster versus the current need for the `stringInSlice` function which searches through the slice each time.


---

<sup>**[rpcserver.go, line 661 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K2nEG4AIOLdoFPt6zNZ)** ([raw file](https://github.com/btcsuite/btcd/blob/23eb93bdadb538560f14110b2d675ccae781941a/rpcserver.go#L661)):</sup>
This should preallocate the entries in the backing array even if only some of them end up being added.  The problem with the current approach is you force several reallocates and memory copies since the original backing array is empty.  It is hard on the garbage collector and less efficient.

```Go
vinList := make([]btcjson.VinPrevOut, 0, len(mtx.TxIn))
```

Notice the addition of the extra 0 for the number of entries while the preallocated space in the backing array is the number of inputs.  The rest of the code then can stay the same.


---

<sup>**[rpcserver.go, line 667 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K2nKpy84WtwZTMrLPon)** ([raw file](https://github.com/btcsuite/btcd/blob/23eb93bdadb538560f14110b2d675ccae781941a/rpcserver.go#L667)):</sup>
Please reverse this so the code below it is unindented and the return is instant.  That is:

```Go
// Don't include tx if filterAddrs is not empty because coinbase
// has no address and so would never match a non-empty filter.
if len(filterAddrs) != 0 {
	return nil
}

...
return vinList
```


---

<sup>**[rpcserver.go, line 689 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K2nQPkKAmiTojfcykaG)** ([raw file](https://github.com/btcsuite/btcd/blob/23eb93bdadb538560f14110b2d675ccae781941a/rpcserver.go#L689)):</sup>
Please move this down to just before it's used.  This is exactly where the old `vinEntry := ...` was.


---

<sup>**[rpcserver.go, line 730 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K2nJ8ja3RSYPY3IK5w6)** ([raw file](https://github.com/btcsuite/btcd/blob/23eb93bdadb538560f14110b2d675ccae781941a/rpcserver.go#L730)):</sup>
Please add the condition after the && on the next line.  We aim to keep the code at 80 cols.  A few chars is acceptable here and there, but this one is pushing 106.


---

<sup>**[rpcserver.go, line 752 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K2nS4PrUSOKuiH4lkiD)** ([raw file](https://github.com/btcsuite/btcd/blob/23eb93bdadb538560f14110b2d675ccae781941a/rpcserver.go#L752)):</sup>
Same here regarding the preallocation.


---

<sup>**[rpcserver.go, line 754 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K2nS8-qlSjgeSzd84zJ)** ([raw file](https://github.com/btcsuite/btcd/blob/23eb93bdadb538560f14110b2d675ccae781941a/rpcserver.go#L754)):</sup>
Please move it down to specify it just before it's used for the first time to be consistent.


---

<sup>**[rpcserver.go, line 771 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K2nSt6bMlJUaDXaRks_)** ([raw file](https://github.com/btcsuite/btcd/blob/23eb93bdadb538560f14110b2d675ccae781941a/rpcserver.go#L771)):</sup>
I'd suggest moving most of the vout field setting code below this point and simply check if it passes the filter here.  There is no point in doing all of the work to build a vout only to ignore it later.  Since `EncodeAddress` is relatively expensive, I'd definitely build the list of encoded address strings here too since they're needed to check the filter.  For example:

```Go
encodedAddrs := make([]string, len(addrs))
for j, addr := range addrs {
	encodedAddrs[j] = addr.EncodeAddress()
}
if len(filterAddrMap) > 0 {
	for _, addr := range encodedAddrs {
		if _, exists := filterAddrMap[addr]; !exists {
			continue nextTxOut
		}
	}
}

var vout btcjson.Vout
vout.N = uint32(i)
...
vout.ScriptPubKey.Addresses = encodedAddrs
...
```


---

<sup>**[rpcserver.go, line 791 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K2nVKJfdHlbpEMmuB2q)** ([raw file](https://github.com/btcsuite/btcd/blob/23eb93bdadb538560f14110b2d675ccae781941a/rpcserver.go#L791)):</sup>
This won't be needed if following my advice above.


---

<sup>**[rpcserver.go, line 805 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K2nVm4KmSSqsm2oWGKq)** ([raw file](https://github.com/btcsuite/btcd/blob/23eb93bdadb538560f14110b2d675ccae781941a/rpcserver.go#L805)):</sup>
Please move this just before the if for consistency.


---

<sup>**[rpcserver.go, line 810 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K2nVs7wNgaYFeN3DK0W)** ([raw file](https://github.com/btcsuite/btcd/blob/23eb93bdadb538560f14110b2d675ccae781941a/rpcserver.go#L810)):</sup>
Extra blank line that shouldn't be here.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/516)

dan-da:

---

<sup>**[btcjson/chainsvrcmds.go, line 554 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K2nRMPZh-GkfCoK_9fQ)** ([raw file](https://github.com/btcsuite/btcd/blob/23eb93bdadb538560f14110b2d675ccae781941a/btcjson/chainsvrcmds.go#L554)):</sup>
I suppose that it is worth using a map for the (probably unusual) case that a large filter list is received.    For small lists (common case) performance is probably similar or may actually be better for the array loop than a hash lookup.

I will attempt the change.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/516)

dan-da:

---

<sup>**[btcjson/chainsvrcmds.go, line 554 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K2nRMPZh-GkfCoK_9fQ)** ([raw file](https://github.com/btcsuite/btcd/blob/23eb93bdadb538560f14110b2d675ccae781941a/btcjson/chainsvrcmds.go#L554)):</sup>
Hmm, if we change it here, then doesn't that force the caller to send the data as a map, eg:  { addr1 :  true,  addr2: true,  addr3:  true }, etc?

From the API caller's perspective, that seems messier and uglier than before:, which was simply:  [addr1, addr2, addr3].

It could be that we keep the external interface an array but then copy it into a map internally.  of course that adds some overhead.

thoughts?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/516)

rpcserver.go
```diff
@@ -644,22 +644,40 @@ func createVinList(mtx *wire.MsgTx) []btcjson.Vin {
...
 		return vinList
 	}
 
 	// Lookup all of the referenced transactions needed to populate the
 	// previous output information if requested.
 	var txStore blockchain.TxStore
-	if vinExtra != 0 {
+	if vinExtra != 0 || len(filterAddrs) > 0 {
 		tx := btcutil.NewTx(mtx)
 		txStoreNew, err := s.server.txMemPool.fetchInputTransactions(tx, true)
```
dajohi: Should this be FetchInputTransactions, or does the mempool need locking?

davecgh:

---

<sup>**[btcjson/chainsvrcmds.go, line 554 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K2nRMPZh-GkfCoK_9fQ)** ([raw file](https://github.com/btcsuite/btcd/blob/23eb93bdadb538560f14110b2d675ccae781941a/btcjson/chainsvrcmds.go#L554)):</sup>
Yes it would force the caller to specify it that way.  I'm ok with building a map out of it internally.  That would provided the desired duplicate filtering functionality and scale well with large numbers of addresses.


---

<sup>**[rpcserver.go, line 682 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K1HRbgQcypoY5ov1VTa-r2-682)** ([raw file](https://github.com/btcsuite/btcd/blob/23eb93bdadb538560f14110b2d675ccae781941a/rpcserver.go#L682)):</sup>
Yes.  It should be.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/516)

dan-da:
Ok, I've made all the requested changes and I think it is ready for another (hopefully final) review.

I changed the internal filterAddrs var from []string to map[string]bool.  This is copied from the JSON input, which is still []string.

I also refactored createVinListPrevOut a bit so it is does not create the btcjson.VinPrevOut entry until after the filter passes.  This matches a small refactor in createVoutList requested by @davecgh.

I tested the latest code once again with:
 vinExtra=0, filterParams omitted
 vinExtra=1, filterParams omitted
 filterParams empty
 filterParams non-empty

each case appears to work correctly and my client app that uses the data displays transaction list with correct values for inputs and outputs.

dan-da:
Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

<sup>**[rpcserver.go, line 682 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K1HRbgQcypoY5ov1VTa-r2-682)** ([raw file](https://github.com/btcsuite/btcd/blob/23eb93bdadb538560f14110b2d675ccae781941a/rpcserver.go#L682)):</sup>
I'm not clear on this.  Do I need to change something, or not?  Anyway, this call was pre-existing to this pull request.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/516)

davecgh:
Thanks for the updates and your work on this.  I've noted a few minor efficiency-related things.


---

Reviewed 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

<sup>**[rpcserver.go, line 682 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K1HRbgQcypoY5ov1VTa-r2-682)** ([raw file](https://github.com/btcsuite/btcd/blob/23eb93bdadb538560f14110b2d675ccae781941a/rpcserver.go#L682)):</sup>
We can do it in a separate PR as I realize it's existing code.  Don't worry about it in this PR.


---

<sup>**[rpcserver.go, line 691 \[r4\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K36Up05X05F0cGZzJHO)** ([raw file](https://github.com/btcsuite/btcd/blob/b0290add236e6f6cb2ea994f0159c19acbb8a5e3/rpcserver.go#L691)):</sup>
With the other changes that cause a continue as soon as possible, this variable should no longer be needed.


---

<sup>**[rpcserver.go, line 718 \[r4\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K36RYsga80Z-Z1XKG7I)** ([raw file](https://github.com/btcsuite/btcd/blob/b0290add236e6f6cb2ea994f0159c19acbb8a5e3/rpcserver.go#L718)):</sup>
Once it passes the filter, the continue can happen immediately if you add a label to the outer for loop:

```Go
nextTxIn:
	for _, txIn := range mtx.TxIn {
...
		if _, exists := filterAddrMap[encodedAddrs[j]]; !exists {
			continue nextTxIn
		}
```

I suggest this since there isn't any reason to continue encoding all the other addresses when it's just going to continue and skip them anyways.  


---

<sup>**[rpcserver.go, line 756 \[r4\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K36VjpaVx_iYKgh3JC2)** ([raw file](https://github.com/btcsuite/btcd/blob/b0290add236e6f6cb2ea994f0159c19acbb8a5e3/rpcserver.go#L756)):</sup>
Nor this one.


---

<sup>**[rpcserver.go, line 775 \[r4\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K36Svs-HQORgtKV_Bnl)** ([raw file](https://github.com/btcsuite/btcd/blob/b0290add236e6f6cb2ea994f0159c19acbb8a5e3/rpcserver.go#L775)):</sup>
The same here as above in the input section:

```Go
nextTxOut:
	for _, txOut := range mtx.TxOut {
...
		if _, exists := filterAddrMap[encodedAddrs[j]]; !exists {
			continue nextTxOut
		}
<!-- Reviewable comment -K36Svs01-Hnm1pBsFf7:-1172789231 -->

---

<sup>**[rpcserver.go, line 3180 \[r4\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K36U9SHKJ1P7mJOyXbv)** ([raw file](https://github.com/btcsuite/btcd/blob/b0290add236e6f6cb2ea994f0159c19acbb8a5e3/rpcserver.go#L3180)):</sup>
Since this is only acting a set, please use `make(map[string]struct[})` and `filterAddrs[addr] = struct{}{}`.  This approach causes no memory to be allocated for the keys which aren't used.  The rest of the syntax that checks for key existence (`if _, exists := filterAddrs[addr]; exists { ... }`) stays the same.
<!-- Reviewable comment -K36U9SHKJ1P7mJOyXbw:1182175949 -->

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/516)
<!-- Sent from Reviewable.io -->

**Comment by dan-da** [comment]
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

<sup>**[rpcserver.go, line 718 \[r4\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K36RYsga80Z-Z1XKG7I)** ([raw file](https://github.com/btcsuite/btcd/blob/b0290add236e6f6cb2ea994f0159c19acbb8a5e3/rpcserver.go#L718)):</sup>
If I'm not mistaken, this changes the logic.   My code was saying: "if any address in encodedAddrs matches the filter, then pass filter".   This code (if I understand it correctly) is saying: "if any address in encodedAddrs does not match filter, then fail filter and go to next vin".  ( because the if/continue take place inside the for loop that iterates over encodedAddrs. )

Now the common/usual case for Vin is that len(encodedAddrs) == 1.  So they get the same result. 

But if len(encodedAddrs) > 1 then I believe my logic is more correct, as the goal is to retrieve all inputs and outputs that contain the filter addr(s).

Am I wrong?

also, I'm not sure exactly how/when len(encodedAddrs) > 1.    multisig tx perhaps?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/516)

**Comment by dan-da** [comment]
I changed the type of filterAddrs to map[string]struct{}.

I also renamed filterAddrs to filterAddrMap.

I did not yet make the change to remove the passesFilter var and use a continue/label because I believe that changes the logic for cases when len(filterAddrMap) > 1.  See my previous comment.

**Comment by davecgh** [comment]
Reviewed 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

<sup>**[rpcserver.go, line 718 \[r4\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K36RYsga80Z-Z1XKG7I)** ([raw file](https://github.com/btcsuite/btcd/blob/b0290add236e6f6cb2ea994f0159c19acbb8a5e3/rpcserver.go#L718)):</sup>
Alright, I see.  You're right, but then there is an issue with the current code which is what confused me.  You're resetting `passesFilter` to `true` whenever there are items in the filter and the loop iterates with `passesFilter := len(filterAddrMap) == 0`.  Then this code does `if len(filterAddrMap) > 0 { /* maybe set it to true, but it's already true from above... */ }`.  Given that, I believe it should be resetting `passesFilter` to `false` on each loop iteration as I'm pretty sure the current code will _always_ pass the filter whenever it has any entries which isn't the desired behavior.  So, I think it would be better to simply declare the variable right before the block that uses it like so:

```Go
var passesFilter bool
encodedAddrs := make([]string, len(addrs))
for j, addr := range addrs {
	...
}
if !passesFilter {
	continue
}
```


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/516)

davecgh:
<sup>**[rpcserver.go, line 718 \[r4\]](https://reviewable.io:443/reviews/btcsuite/btcd/516#-K36RYsga80Z-Z1XKG7I)** ([raw file](https://github.com/btcsuite/btcd/blob/b0290add236e6f6cb2ea994f0159c19acbb8a5e3/rpcserver.go#L718)):</sup>
Nevermind, you also want it to pass when there are no addresses specified.  It's correct the way it is.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/516)

davecgh:
Thanks.  You're correct about the code the way it is.  Just need the reviewable comments to be acknowledged by you and @dajohi and this is good to go.

It does need to be rebased and squashed to the latest master however.

See [Managing a btcd Pull Request for Code Contibutions ](https://gist.github.com/davecgh/04cb601a147f8c831ca5#rebasing-one-of-your-existing-pull-requests) if you need instructions.


---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/516)

dan-da:
ackknowledged.   rebased/squashed.   pushed.

paging @dajohi

davecgh:
Thanks.  Though, it doesn't look like it was rebased properly.  There are 15 commits instead of one that is on top of b691a222d564e222801d2136489f8758b9154a16.

dan-da:
git is crazy, but I think I beat it into submission.   :-)   It's a single commit now.


---

Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/516)

davecgh:
Yep.  Looks good.  Once @dajhoi acks the final discussion, I'll get it merged.


---

Reviewed 5 of 5 files at r6.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/516)

dajohi:
so does @dan-da get to do a btcsuite/btcrpcclient PR too? :)

PR #517, opened 2015-10-20
peer: Ping on interval instead of delayed timer.

davecgh:
**NOTE: Requires pull request #445**

This pull request modifies the ping logic in the peer to ping on an interval regardless of what other messages are being sent versus the previous method of delaying the ping each time a message that is expected to receive data is sent.

This helps improve the ping statistics and simplifies its logic.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/517)

### DIFF
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -43,9 +43,9 @@ const (
 	// inventory cache.
 	maxKnownInventory = 1000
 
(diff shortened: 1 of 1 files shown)


dajohi:
I had the same diff in a branch.  OK.

PR #518, opened 2015-10-20
peer: Implement stall detection.

### LINKED ISSUE #486: BlockManager stops syncing during initial sync

davecgh:
**Requires #445**

*This is an alternative approach to #501 that is a more general solution and also fixes some concurrency-related issues with that PR as a side effect.*

This pull request implements stall detection logic at the peer level to detect and disconnect peers that are either not following the protocol in regards to expected response messages or have otherwise stalled.  This
is accomplished by setting deadlines for each message type which expects a response and periodically checking them while properly taking into account processing time.

There are an increasing number of nodes on the network which claim to be full nodes, but don't actually properly implement the entire p2p protocol even though they implement it enough to cause properly implemented nodes to make data requests to which they never respond.

Since btcd currently only syncs new blocks via single sync peer and, prior to this commit only had very basic stall detection, this could lead to a situation where the block download became stalled indefinitely due to one of these misbehaving peers.  This commit fixes that issue since the stalled peer will now be detected and disconnected which leads to a new sync peer being selected.

This logic will also fit nicely with the future multi-peer sync model which is on the roadmap and for which infrastructure work is underway.

Fixes #486 and #229 .

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/518)

### DIFF
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -27,11 +27,6 @@ const (
 	// MaxProtocolVersion is the max protocol version the peer supports.
 	MaxProtocolVersion = 70011
 
-	// BlockStallTimeout is the number of seconds we will wait for a
-	// "block" response after we send out a "getdata" for an announced
-	// block before we deem the peer inactive, and disconnect it.
-	BlockStallTimeout = 5 * time.Second
-
 	// outputBufferSize is the number of elements the output channels use.
 	outputBufferSize = 50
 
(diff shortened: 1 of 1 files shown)


davecgh:
Updated to improve log message on stalled peer and made it debug level.

toddfries:
I used to have 'orphan block only' symptoms multiple times per day, but definately every 12-24hrs at the minimum... with this pull request, my btcd has not skipped a beat since.  Why is this not merged yet? ;-) ;-) ;-)

dajohi:
No issues in hours with dozens of hosts.  OK

jrick:
Reads like a hack but solves the immediate problem.  Although it's not a fool proof solution (deadlines can be overwritten for example) it will be updated when multipeer syncing is added.  ok

PR #519, opened 2015-10-20
Correct a few style related issues found by golint.

davecgh:
Also, update TravisCI goclean script to remove the special casing which ignored `Id` from the lint output since that exception is no longer needed.  It was previously required due to the old version of `btcjson`,
but that is no longer in the repo.


dajohi:
OK

PR #520, opened 2015-10-20
config: New option --minrelaytxfee

dajohi:
--minrelaytxfee allows the user to specify the minimum transaction
fee in BTC/kB in which the fee is considered a non-zero fee.

### DIFF
--- a/config.go
+++ b/config.go
@@ -106,6 +106,7 @@ type config struct {
 	CPUProfile         string        `long:"cpuprofile" description:"Write CPU profile to the specified file"`
 	DebugLevel         string        `short:"d" long:"debuglevel" description:"Logging level for all subsystems {trace, debug, info, warn, error, critical} -- You may also specify <subsystem>=<level>,<subsystem2>=<level>,... to set the log level for individual subsystems -- Use show to list available subsystems"`
 	Upnp               bool          `long:"upnp" description:"Use UPnP to map our listening port outside of NAT"`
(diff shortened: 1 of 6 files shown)


config.go
```diff
@@ -119,11 +119,13 @@ type config struct {
 	DropAddrIndex      bool          `long:"dropaddrindex" description:"Deletes the address-based transaction index from the database on start up, and the exits."`
 	NoPeerBloomFilters bool          `long:"nopeerbloomfilters" description:"Disable bloom filtering support."`
 	SigCacheMaxSize    uint          `long:"sigcachemaxsize" description:"The maximum number of entries in the signature verification cache."`
+	MinRelayTxFee      float64       `long:"minrelaytxfee" description:"The minimum transaction fee in BTC/kB be considered a non-zero fee."`
```
davecgh: to be considered

doc.go
```diff
@@ -107,6 +107,8 @@ Application Options:
       --nopeerbloomfilters  Disable bloom filtering support.
       --sigcachemaxsize=    The maximum number of entries in the signature
                             verification cache.
+      --minrelaytxfee=      The minimum transaction fee in BTC/kB be considered
```
davecgh: Same

davecgh:
I've tested this and reviewed the code.  The fee is set properly and the `getinfo` RPC call returns the configured value.

OK.

Issue #521, opened 2015-10-20
Mempool fee calculation should not be calculated per full kB


jrick:
When Core added floating fees to their wallet the mempool behavior changed as well to calculate the fee as a percentage of the rate times the size (fee = rate * size / 1000), rather than using integer division first to calculate the full number of kilobytes of the transaction size (fee = (1 + size/1000) * rate).  This can cause transactions accepted by Core's mempool to be rejected by btcd's due to the fee being too low.

Example:

Given a transaction that is 1.5 kB large, and a fee rate of 0.00005 BTC/kB, Core would calculate the required fee to be 0.00075 (this equals the rate * 1.5).  btcd with the same fee rate would instead require the transaction to have a fee of at least 0.00010 (the rate * 2) because the size is greater than 1 kB but does not exceed 2 kB.

davecgh:
Implemented by PR #537.

PR #522, opened 2015-10-20
Increase default minimum relay tx fee to 5000 satoshi.

dajohi:
This matches Bitcoin Core commit 28e3249e53b8ef7516636df0f1406466a513095d

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -82,7 +82,7 @@ Application Options:
                             to list available subsystems (info)
       --upnp                Use UPnP to map our listening port outside of NAT
       --minrelaytxfee=      The minimum transaction fee in BTC/kB to be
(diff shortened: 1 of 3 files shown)


davecgh:
Grumble.  OK.

jrick:
ok, but note that while we're making mempool match with latest Core we should fix #521 as well.  btcwallet will need to be bumped too, but the more granular fees there can wait (they just need to be above the new minimum).

davecgh:
After mulling it over a bit (and as evidenced by my initial Grumble remark), I'm really not happy with this change.  There is now a knob to change it if users need to and this is really not the right way to fix the motivating case anyways.

So, I'm going to say no on this.

PR #523, opened 2015-10-21
Transaction input and output sorting

T909:
This adds a Sort() method to wire.MsgTx.  It sorts the inputs and outputs based on BIP LI01 
https://github.com/kristovatlas/rfc/blob/master/bips/bip-li01.mediawiki

It's implemented using golang's standard library sort.Sort(), which does a quicksort, and needs three methods on the *[]MsgTx slice type: Len(), Less(), and Swap().  Len() and Swap() are easy; the Less() comparison operation is where the actual BIP is implemented.

Note that this does change the txid when it actually makes changes.  The idea is to minimize information leakage by having a standard ordering for inputs and outputs.  A standard ordering can also help minimize communications requirements when constructing transactions in a multi-party use case.

I think the code is pretty straight-forward but this is my first PR to btcd so it could be in the wrong place.  Operating on btcutil.Tx would be another possibility, but wire.MsgTx seems better as it changes txids.

Also I have not added tests to msgtx_test.go.  If tests are desired I can write some but I'm not sure the best way to make tests as none of the added methods can return errors.

Thanks for looking at this and hope it's helpful.  Also could rename to SortLI01 or something if there are other sorting methods wanted.


davecgh:
Thanks for the PR.  I haven't reviewed the code very closely yet, but there are a few things of note:

1. This doesn't seem like it belongs in the wire package which is intended to deal with the wire protocol.  Perhaps `btcutil`?  I'd like to hear opinions from others on that.
2. There are no tests to ensure the sorting is working as expected.  See the [Testing](https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md#Testing) section of the [Code Contribution Guidelines](https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md) for more info.
3. This would need to be made into a single commit that is rebased on top of the current btcd master.  See [btcd Pull Request Management](https://gist.github.com/davecgh/04cb601a147f8c831ca5) for details.

Roasbeef:
I have mixed feelings on whether the new functionality should reside in `btcutil` or `wire`. On one hand,  sorting the inputs/outputs of a transactions may change its `txid`. If the order of the inputs/outputs on the underlying `wire.MsgTx` are changed, then the [cached txid](https://github.com/btcsuite/btcutil/blob/master/tx.go#L39) will need to be re-generated. So it follows that that an operation which invalidates the cache should live on the same abstraction level as the cache itself. 

However, in my experience, when crafting transactions, one is typically manipulating an instance of `wire.MsgTx`. Its helper functions like `AddTxIn/AddTxOut` lend well to the incremental creation of transactions. But if sorting is implemented at this level, then this complicates `txid` cache invalidation at higher layers/wrappers (eg. `btcutil.Tx`). Though I guess one could just add a `dirty` bit to `wire.MsgTx` which wrappers would be expected to check, updating their caching accordingly. 

This also raises another question: should we make sorting the *default* behavior? Meaning `wire.MsgTx.TxSha()` would automatically sort before hashing the transaction. 

With regard to tests, I think the [examples](https://github.com/kristovatlas/rfc/blob/master/bips/bip-li01.mediawiki#Examples) section of the BIP should be sufficient for initial coverage and verification that the implementation *at least* meets spec.

jrick:
It doesn't need to be a method of btcutil.Tx.  It could just be the function `SortTx(tx *wire.MsgTx)` in the util package (or elsewhere).

jrick:
`This also raises another question: should we make sorting the default behavior? Meaning  wire.MsgTx.TxSha()  would automatically sort before hashing the transaction.`

This won't work.  It's a domain specific function (for applications that create their own transactions and need to ensure they aren't leaking implementation details).  But unsorted transactions can still be valid and must be processed as is by btcd.

davecgh:
> With regard to tests, I think the examples section of the BIP should be sufficient for initial coverage and verification that the implementation at least meets spec. 

Agreed.  I had edited my original message to call out the examples as well.  I'm pretty sure that will exercise all code paths anyways since it's simple sorting.

> It doesn't need to be a method of btcutil.Tx. It could just be the function SortTx(tx *wire.MsgTx) in the util package (or elsewhere).

That is what I was thinking as well.  In fact, we might want the func to return a new sorted transaction instead of modifying the existing one.  This would probably depend on the intended use cases, but what I'd be worried about is something like:

```Go
btcutil.SortTx(msgBlock.Transactions[0])   // or with the current PR even msgBlock.Transactions[0].Sort()
```

That would (potentially if the original tx in the block is not already lexographically ordered) invalidate the block.

jrick:
There's not a whole lot we can do about that in a language without immutable data.  By convention we just have to treat all published transactions as immutable.

davecgh:
It's true that we can't stop it, but you can certainly make it harder to accidentally do.  I think the probability of making that mistake is quite high here since I could easily see code that wants to take published transactions and sort them.  In that case, even if the language did have an immutable keyword, the result would be that you need the sorted version to be a copy.  Given that, I think it's a good indicator to how it should be handled from the start.

jrick:
But we already have deep copying.

```
txCopy := msgBlock.Transactions[0].Copy()
btcutil.SortTx(txCopy)
```

davecgh:
Right, and that was actually what I had in mind for inside the function:

```Go
func SortTx(tx *wire.MsgTx) *wire.msgTx {
	txCopy := tx.Copy()
	// Sort it
	return txCopy
}
```

 However, I suppose having the flexibility to do both could be useful.

jrick:
Basically I see two uses of this function, the first would be by the author of a transaction to sort before publishing to mempool, in which case an in place sort is fine due to them being the only one with the transaction.  The second use is to check whether a published transaction is sorted or not.  The second use doesn't actually require the caller to get back the sorted transaction, they just need a bool.  This can be implemented by sorting again and comparing hashes, but is more efficient to just check for sorted inputs and outputs directly.

davecgh:
That sounds reasonable.  So perhaps a `SortTx(tx *wire.MsgTx)` and `IsTxSorted(tx *wire.MsgTx) bool`.

Roasbeef:
> Basically I see two uses of this function...

A third use case arises within the context of multi-party smart contract protocols built on top of Bitcoin. If all parties ascribe to an agreed upon canonical ordering of inputs/outputs, then when gathering signatures for "template contract" transactions, only signatures need to be exchanged instead of the txn itself (which can become relatively large). Depending on the protocol, a majority of these transactions will never make it into the mempool of *any* node due to the incentives related to prior state revocation.

The proposed "copy-on-sort" behavior addresses my prior concerns related to cache invalidation within `btcutil.Tx`.

`btcutil.Sort(tx *wire.MsgTx)` and `btcutil.IsTxSorted(tx *wire.MsgTx)` (as described above) sounds good to me :+1:

T909:
OK cool I will also make an IsSorted(tx *wire.MsgTx) bool function, and move them to btcutil.  Also merging the commits together and stuff, forgot about that!

Also, just to make sure, these are both functions, not methods, and the btcutil.Sort(tx *wire.MsgTx) returns a new wire.MsgTx, or a new *wire.MsgTX?  So don't touch the tx in the argument.

I will try to update stuff today but may not get to it until later.  Thanks for the discussion and pointers!

davecgh:
Correct.  You can't define a method on a type defined in another package, so it has to be a function.

davecgh:
Also, I suggest separate files in btcutil, `txsort.go`, and `txsort_test.go`.

T909:
OK, should I close this and open a new PR in btcutil?  It's a different repo so maybe I'll link to this discussion from btcutil's page.

davecgh:
Yes, that makes sense.  Thanks!

T909:
closed and moved to here: btcsuite/btcutil#58

PR #524, opened 2015-10-21
peer: Don't log write errs due to disconnect.

davecgh:
**Requires #445**

This pull request modifies the peer logging so that it will not log an error when it's due to the remote peer disconnecting or the peer being forcibly disconnected or shutdown.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/524)


jcvernaleo:
I think we discussed this one on irc.  As stated there, I'm in favor of relatively clean logs at the 'normal' level so I like this change.
Code looks good.  Assuming travis ever finishes, OK on this one too.

PR #525, opened 2015-10-22
Move non-mempool specific functions to new file.

dajohi:
No functional change. Add tests.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/525)

### DIFF
--- a/policy_test.go
+++ b/policy_test.go
@@ -1,0 +1,242 @@
+// Copyright (c) 2013-2015 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package main
+
+import (
+	"testing"
+
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+)
+
+// TestCalcMinRequiredTxRelayFee tests the calcMinRequiredTxRelayFee API.
+func TestCalcMinRequiredTxRelayFee(t *testing.T) {
+	tests := []struct {
+		name     string         // test description.
+		size     int64          // Transaction size in bytes.
+		relayFee btcutil.Amount // minimum relay transaction fee.
+		want     int64          // Expected fee.
+	}{
+		{
+			"zero value with default minimum relay fee",
+			0,
+			defaultMinRelayTxFee,
+			int64(defaultMinRelayTxFee),
+		},
+		{
+			"100 bytes with default minimum relay fee",
+			100,
+			defaultMinRelayTxFee,
+			int64(defaultMinRelayTxFee),
+		},
+		{
+			"max standard tx size with default minimum relay fee",
+			maxStandardTxSize,
+			defaultMinRelayTxFee,
+			101000,
+		},
+		{
+			"max standard tx size with max satoshi relay fee",
+			maxStandardTxSize,
+			btcutil.MaxSatoshi,
+			btcutil.MaxSatoshi,
+		},
+	}
+
+	for _, test := range tests {
+		got := calcMinRequiredTxRelayFee(test.size, test.relayFee)
+		if got != test.want {
+			t.Errorf("TestCalcMinRequiredTxRelayFee test '%s' "+
+				"failed: got %v want %v", test.name, got,
+				test.want)
+			continue
+		}
+	}
+}
+
+// TestCheckPkScriptStandard tests the checkPkScriptStandard API.
+func TestCheckPkScriptStandard(t *testing.T) {
+	var pubKeys [][]byte
+	for i := 0; i < 4; i++ {
+		pk, err := btcec.NewPrivateKey(btcec.S256())
+		if err != nil {
+			t.Fatalf("TestCheckPkScriptStandard NewPrivateKey failed: %v",
+				err)
+			return
+		}
+		pubKeys = append(pubKeys, pk.PubKey().SerializeCompressed())
+	}
+
+	tests := []struct {
+		name       string // test description.
+		script     *txscript.ScriptBuilder
+		isStandard bool
+	}{
+		{
+			"key1 and key2",
+			txscript.NewScriptBuilder().AddOp(txscript.OP_2).
+				AddData(pubKeys[0]).AddData(pubKeys[1]).
+				AddOp(txscript.OP_2).AddOp(txscript.OP_CHECKMULTISIG),
+			true,
+		},
+		{
+			"key1 or key2",
+			txscript.NewScriptBuilder().AddOp(txscript.OP_1).
+				AddData(pubKeys[0]).AddData(pubKeys[1]).
+				AddOp(txscript.OP_2).AddOp(txscript.OP_CHECKMULTISIG),
+			true,
+		},
+		{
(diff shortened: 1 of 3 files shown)


davecgh:
As discussed on IRC, I like the concept.

policy_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"testing"
+
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+)
+
+// TestCalcMinRequiredTxRelayFee tests the calcMinRequiredTxRelayFee API.
+func TestCalcMinRequiredTxRelayFee(t *testing.T) {
+	tests := []struct {
+		size     int64          // Transaction size in bytes.
```
davecgh: Please give the tests a `name  string` field and include it in the error output.  I know there are several tests that don't do that currently, but we've been trying to move that direction since it's much harder to figure out which test failed when only given a number.  Granted it's not too difficult when there are only a few tests, but I'd like to be consistent.

policy_test.go
```diff
@@ -0,0 +1,104 @@
...
+// TestDust tests the isDust API.
+func TestDust(t *testing.T) {
+	pkScript := []byte{0x76, 0xa9, 0x21, 0x03, 0x2f, 0x7e, 0x43,
+		0x0a, 0xa4, 0xc9, 0xd1, 0x59, 0x43, 0x7e, 0x84, 0xb9,
+		0x75, 0xdc, 0x76, 0xd9, 0x00, 0x3b, 0xf0, 0x92, 0x2c,
+		0xf3, 0xaa, 0x45, 0x28, 0x46, 0x4b, 0xab, 0x78, 0x0d,
+		0xba, 0x5e, 0x88, 0xac}
+
+	tests := []struct {
+		txOut  wire.TxOut
```
davecgh: Same here.

policy_test.go
```diff
@@ -0,0 +1,104 @@
...
+		},
+		{
+			// Unspendable pkScript due to an invalid public key
+			// script.
+			wire.TxOut{5000, []byte{0x01}},
+			true,
+		},
+	}
+	for x, test := range tests {
+		res := isDust(&test.txOut, defaultMinRelayTxFee)
```
davecgh: I think the test struct should include the relay fee and the tests should specify different values.  First, this will properly test the function works with different relay fees, and second, it will prevent the tests from failing if the default relay fee is changed.

davecgh:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[policy.go, line 27 \[r1\]](https://reviewable.io:443/reviews/btcsuite/btcd/525#-K1HSVWJi8De90rWYFaL)** ([raw file](https://github.com/btcsuite/btcd/blob/d9a673460e3724e8ab675ea7b143ba940d150a2d/policy.go#L27)):</sup>
Please don't wrap function definitions.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/525)

davecgh:
Reviewed 1 of 2 files at r2.
Review status: 2 of 3 files reviewed at latest revision, 9 unresolved discussions, all commit checks successful.

---

<sup>**[policy_test.go, line 72 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/525#-K1_pnfZR87rRi_kNYVj)** ([raw file](https://github.com/btcsuite/btcd/blob/2499e410e6f57bee5d5dbdcda4a542560337a03a/policy_test.go#L72)):</sup>
This looks leftover.  Please use `t.Fatalf`.


---

<sup>**[policy_test.go, line 79 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/525#-K1_q9B7UdTxNksbgQAy)** ([raw file](https://github.com/btcsuite/btcd/blob/2499e410e6f57bee5d5dbdcda4a542560337a03a/policy_test.go#L79)):</sup>
Why not just make this `*txscript.ScriptBuilder` and avoid the extra indirects below?


---

<sup>**[policy_test.go, line 158 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/525#-K1_qIHNU6aM1GRdrcpn)** ([raw file](https://github.com/btcsuite/btcd/blob/2499e410e6f57bee5d5dbdcda4a542560337a03a/policy_test.go#L158)):</sup>
Better error message please.  This won't be very useful if it gets hit due to an invalid test case.


---

<sup>**[policy_test.go, line 198 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/525#-K1_qc0QJZ_AMzJfeCO0)** ([raw file](https://github.com/btcsuite/btcd/blob/2499e410e6f57bee5d5dbdcda4a542560337a03a/policy_test.go#L198)):</sup>
Please put the actual relay fee that is being tested, not the default.  If you want to prove the default hasn't changed, then add another test like `func TestDefaultMinRelayFeePolicy(t *testing.T) {...}`.


---

<sup>**[policy_test.go, line 204 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/525#-K1_qpp43NExZBVJNU-n)** ([raw file](https://github.com/btcsuite/btcd/blob/2499e410e6f57bee5d5dbdcda4a542560337a03a/policy_test.go#L204)):</sup>
Same.


---

<sup>**[policy_test.go, line 210 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/525#-K1_qqZiCZdcWmBMY4fB)** ([raw file](https://github.com/btcsuite/btcd/blob/2499e410e6f57bee5d5dbdcda4a542560337a03a/policy_test.go#L210)):</sup>
Same.


---

<sup>**[policy_test.go, line 217 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/525#-K1_qrNKRFvBxfmbq7Y-)** ([raw file](https://github.com/btcsuite/btcd/blob/2499e410e6f57bee5d5dbdcda4a542560337a03a/policy_test.go#L217)):</sup>
Same.


---

<sup>**[policy_test.go, line 224 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/525#-K1_qs4qF-Q3dEwaSQ6r)** ([raw file](https://github.com/btcsuite/btcd/blob/2499e410e6f57bee5d5dbdcda4a542560337a03a/policy_test.go#L224)):</sup>
Same.


---

<sup>**[policy_test.go, line 232 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/525#-K1_qsngCyEVztKnkg0d)** ([raw file](https://github.com/btcsuite/btcd/blob/2499e410e6f57bee5d5dbdcda4a542560337a03a/policy_test.go#L232)):</sup>
Same. 

Also, please use a few different values.  Only testing 0 and 0.00001 doesn't really exercise the function very well.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/525)

davecgh:
Reviewed 1 of 2 files at r2, 1 of 1 files at r3.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/525)

davecgh:
OK

PR #526, opened 2015-10-22
txscript: fix isMultiSig bug.

dajohi:
isMultiSig was not verifying the number of pubkeys specified matched
the number of pubkeys provided.  This caused certain non-standard
scripts to be considered multisig scripts.

However, the script still would have failed during execution.

NOTE: This only affects whether or not the script is considered
standard and does NOT affect consensus.


davecgh:
OK

davecgh:
OK

PR #530, opened 2015-10-23
docs: Make various README.md files consistent.

davecgh:
First, it removes the documentation section from all the `README.md` files and instead puts a web-based godoc badge and link at the top with the other badges.  This is being done since the local godoc tool no longer ships with Go by default, so the instructions no longer work without first installing godoc. Due to this, pretty much everyone uses the web-based godoc these days anyways.  Anyone who has manually installed godoc won't need instructions.

Second, it makes sure the ISC license badge is at the top with the other badges and removes the textual reference in the overview section.

Finally, it's modifies the `Installation` section to `Installation and Updating` and adds a `-u` to the `go get` command since it works for both and thus is simpler.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/530)

### DIFF
--- a/blockchain/README.md
+++ b/blockchain/README.md
@@ -4,6 +4,8 @@ blockchain
 [![Build Status](http://img.shields.io/travis/btcsuite/btcd.svg)]
 (https://travis-ci.org/btcsuite/btcd) [![ISC License]
 (http://img.shields.io/badge/license-ISC-blue.svg)](http://copyfree.org)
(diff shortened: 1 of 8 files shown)


jcvernaleo:
While I'm not happy about what has happened to godoc (I prefer it to the web stuff) that's the way it is going so we need to adapt.
This looks good to me.
OK

PR #534, opened 2015-10-26
peer: Don't stall on pongs.

davecgh:
[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/534)


jrick:
This makes a good case for reverting the always ping change that went in last week.  If we're on IBD and already have deadlines for getdata why do we need pings as well?  No point in sending them if we aren't going to bother checking for the pong.

davecgh:
The main purpose of sending them is for updated ping statistics.

jrick:
I'll throw in my ok on this one since the ping only exists to update stats.  The other deadlines should be sufficient to catch most stalls until multipeer is in.

dajohi:
Seconded.

PR #537, opened 2015-10-29
Altered min fee calculation function to match current Core behavior.

lologarithm:
Specifically the number of KB is no longer rounded down to next integer.

Since this changes slightly the meaning of the 'minRelayTxFee' variable it has been renamed to 'relayTxFeeRate' and have deprecated the old config value associated with it and have added a new one.

See issue: https://github.com/btcsuite/btcd/issues/521

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/537)

### DIFF
--- a/policy.go
+++ b/policy.go
@@ -26,11 +26,14 @@ const (
 func calcMinRequiredTxRelayFee(serializedSize int64, minRelayTxFee btcutil.Amount) int64 {
 	// Calculate the minimum fee for a transaction to be allowed into the
 	// mempool and relayed by scaling the base fee (which is the minimum
-	// free transaction relay fee).  minRelayTxFee is in Satoshi/KB, so
-	// divide the transaction size by 1000 to convert to kilobytes.  Also,
-	// integer division is used so fees only increase on full kilobyte
-	// boundaries.
-	minFee := (1 + serializedSize/1000) * int64(minRelayTxFee)
+	// free transaction relay fee). minTxRelayFee is in Satoshi/kB so
+	// multiply by serializedSize (which is in bytes) and divide by 1000 to get
+	// minimum Satoshis.
+	minFee := (serializedSize * int64(minRelayTxFee)) / 1000
+
+	if minFee == 0 && minRelayTxFee > 0 {
+		minFee = int64(minRelayTxFee)
+	}
 
 	// Set the minimum fee to the maximum possible value if the calculated
 	// fee is not in the valid range for monetary amounts.
--- a/policy_test.go
+++ b/policy_test.go
@@ -22,28 +22,60 @@ func TestCalcMinRequiredTxRelayFee(t *testing.T) {
 		want     int64          // Expected fee.
 	}{
 		{
-			"zero value with default minimum relay fee",
-			0,
-			defaultMinRelayTxFee,
-			int64(defaultMinRelayTxFee),
+			// Ensure combination of size and fee that are less than 1000
+			// produce a non-zero fee.
+			"250 bytes with relay fee of 3",
+			250,
+			3,
+			3,
 		},
 		{
 			"100 bytes with default minimum relay fee",
 			100,
 			defaultMinRelayTxFee,
-			int64(defaultMinRelayTxFee),
+			100,
 		},
 		{
 			"max standard tx size with default minimum relay fee",
 			maxStandardTxSize,
 			defaultMinRelayTxFee,
-			101000,
+			100000,
 		},
 		{
 			"max standard tx size with max satoshi relay fee",
 			maxStandardTxSize,
 			btcutil.MaxSatoshi,
 			btcutil.MaxSatoshi,
+		},
+		{
+			"1500 bytes with 5000 relay fee",
+			1500,
+			5000,
+			7500,
+		},
+		{
+			"1500 bytes with 3000 relay fee",
+			1500,
+			3000,
+			4500,
+		},
+		{
+			"782 bytes with 5000 relay fee",
+			782,
+			5000,
+			3910,
+		},
+		{
+			"782 bytes with 3000 relay fee",
+			782,
+			3000,
+			2346,
+		},
+		{
+			"782 bytes with 2550 relay fee",
+			782,
+			2550,
+			1994,
 		},
 	}
 


jrick:
Couple issues here:

KB should be kB since it's kilo not kibi (1000 bytes, not 1024).

It also uses the rate as the minimum allowed fee, but it should be calculated to be less than that when the transaction size is less than 1000 bytes.  The serializedSize > 1000 check should be removed.

policy.go
```diff
@@ -26,11 +26,14 @@ const (
...
-	// divide the transaction size by 1000 to convert to kilobytes.  Also,
-	// integer division is used so fees only increase on full kilobyte
-	// boundaries.
-	minFee := (1 + serializedSize/1000) * int64(minRelayTxFee)
+	// free transaction relay fee). minTxRelayFee is in Satoshi/KB so
+	// multiply by serializedSize (which is in KB) and divide by 1000 to get
+	// minimum Satoshis.
+	minFee := int64(minRelayTxFee)
+	// If we have more than 1KB calculate the new minFee
+	if serializedSize > 1000 {
```
davecgh: This does not match the behavior.  It should be:

```Go
minFee := (serializedSize * int64(minRelayTxFee)) / 1000
if minFee == 0 && minRelayTxFee > 0 {
	minFee = minRelayTxFee
}
```

For example, let's take some representative values.  A serialized transaction size of 284 bytes and a min fee of 5000.

The code in this PR as it is now would result in `minFee = 5000`.  However it should be `minFee = 1420`.

In fact, I think it would be  a good idea to add this specific test to `TestCalcMinRequiredTxRelayFee`.

policy.go
```diff
@@ -26,11 +26,14 @@ const (
 func calcMinRequiredTxRelayFee(serializedSize int64, minRelayTxFee btcutil.Amount) int64 {
 	// Calculate the minimum fee for a transaction to be allowed into the
 	// mempool and relayed by scaling the base fee (which is the minimum
-	// free transaction relay fee).  minRelayTxFee is in Satoshi/KB, so
-	// divide the transaction size by 1000 to convert to kilobytes.  Also,
-	// integer division is used so fees only increase on full kilobyte
-	// boundaries.
-	minFee := (1 + serializedSize/1000) * int64(minRelayTxFee)
+	// free transaction relay fee). minTxRelayFee is in Satoshi/KB so
+	// multiply by serializedSize (which is in KB) and divide by 1000 to get
```
davecgh: The serialized transaction size is in bytes, not KB.

policy.go
davecgh: More examples of tx sizes and fees (I'd suggest adding these to the tests):

tx size: 1500 bytes
minRelayTxFee: 5000
minFee: 7500

tx size: 1500 bytes
minRelayTxFee: 3000
minFee: 4500

tx size: 782 bytes
minRelayTxFee: 5000
minFee: 3910

tx size: 782 bytes
minRelayTxFee: 3000
minFee: 2346

tx size: 782 bytes
minRelayTxFee: 2550
minFee: 1994

jrick: May want to rename minRelayTxfee to relayTxFeeRate or something else that doesn't imply it is a minimum for all transactions regardless of their size.

davecgh:
@jrick: Yes, I'm fine with using `kB` to mean 1000 bytes since that matches the standard.  Of course, I also prefer KiB to mean 1024 bytes since it is unambiguous.

policy.go
lologarithm: I will add all those unit tests. I like the idea of changing the variable name as it clearly confused me ;)

lologarithm:
I will make sure to use kB to signify 1000 instead of 1024.

lologarithm:
I have reverted the change I made to make 'minRelayTxFee' the minimum. I have one outstanding question - what is the expected fee for a 0 byte transfer? With the new logic it becomes 0 but that seems bad. Should it default to 1? Or should it be 0?

davecgh:
A 0 byte transfer would not require any fee.  Since it's impossible to have a zero byte serialized transaction, the only way the result can be 0 is if the relay fee is set to 0 in which case, it should be 0.

lologarithm:
Cool - and lastly jrick suggested that the variable name be changed to rate - should I do that now?

lologarithm:
Everything should be fixed now. I thinks If you want to change the minRelayTxfee variable name we could make a separate issue.

davecgh:
I think it probably a good idea to rename it, as it is a relic of the past when it actually was minimum (due to the full kB boundary logic).

Speaking of which, since the logic is changing, I think the command line option `minrelaytxfee` probably needs to be updated in this same PR.

I'm not a huge fan of changing CLI options since it can break existing setups, so I suspect we should add a new option to all the rate to be configured and make the old CLI a _deprecated_ alias for it (see getworkkey for an example).

@jrick Thoughts on the CLI option bit?

jrick:
Deprecating and adding an alias sounds ok to me.

policy.go
```diff
@@ -26,11 +26,10 @@ const (
...
 	// mempool and relayed by scaling the base fee (which is the minimum
-	// free transaction relay fee).  minRelayTxFee is in Satoshi/KB, so
-	// divide the transaction size by 1000 to convert to kilobytes.  Also,
-	// integer division is used so fees only increase on full kilobyte
-	// boundaries.
-	minFee := (1 + serializedSize/1000) * int64(minRelayTxFee)
+	// free transaction relay fee). minTxRelayFee is in Satoshi/kB so
+	// multiply by serializedSize (which is in bytes) and divide by 1000 to get
+	// minimum Satoshis.
+	minFee := (serializedSize * int64(minRelayTxFee)) / 1000
```
davecgh: There is a case missing here (I had it in my original comment):

```Go
if minFee == 0 && minRelayTxFee > 0 {
    minFee = minRelayTxFee
}
```

This is because there are 2 cases in which the `minFee` can be zero:
1) A zero serializedSize (which is actually impossible in practice)
2) When serializedSize * minRelayTxFee < 1000 (due to integer division)

The additional if statement covers the second case to ensure a combination of the serialized tx size and relay rates that would result in values too small will not result in a 0 fee.

policy_test.go
```diff
@@ -25,19 +25,19 @@ func TestCalcMinRequiredTxRelayFee(t *testing.T) {
 			"zero value with default minimum relay fee",
 			0,
 			defaultMinRelayTxFee,
-			int64(defaultMinRelayTxFee),
+			0,
```
davecgh: This should technically be `int64(defaultMinRelayTxFee)` as it was, but really I see no point in having a test for it since a 0-byte transaction is not possible.

I better test, in my opinion, would be:

```Go
		{
			// Ensure combination of size and fee that are less than 1000
			// bytes produce a non-zero fee.
			"250 bytes with relay fee of 3",
			250,
			3,
			3,
		},
```

lologarithm: I will update that test.

lologarithm:
New name for min fee will be relayTxFeeRate instead of minRelayTxFee.

I am also learning the config system now and am going to deprecate the old value and make a new one following the new name.

config.go
```diff
@@ -115,7 +115,8 @@ type config struct {
 	CPUProfile         string        `long:"cpuprofile" description:"Write CPU profile to the specified file"`
 	DebugLevel         string        `short:"d" long:"debuglevel" description:"Logging level for all subsystems {trace, debug, info, warn, error, critical} -- You may also specify <subsystem>=<level>,<subsystem2>=<level>,... to set the log level for individual subsystems -- Use show to list available subsystems"`
 	Upnp               bool          `long:"upnp" description:"Use UPnP to map our listening port outside of NAT"`
-	MinRelayTxFee      float64       `long:"minrelaytxfee" description:"The minimum transaction fee in BTC/kB to be considered a non-zero fee."`
+	MinRelayTxFee      float64       `long:"minrelaytxfee" description:"DEPRECATED -- Use the --relaytxfeerate"`
```
davecgh: `Use the --relaytxfeerate option instead` or simply `Use --relaytxfeerate instead`.

config.go
```diff
@@ -616,6 +618,20 @@ func loadConfig() (*config, []string, error) {
 		return nil, nil, err
 	}
 
+	// If RelayTxFeeRate is non-default we apply it over cfg.relayTxFeeRate
+	// However if it is default we shouldn't because its possible
+	// that the deprecated 'minRelayTxFee' is being used.
+	if cfg.RelayTxFeeRate != defaultRelayTxFeeRate.ToBTC() {
```
davecgh: Rather than repeating all of this, I think it would be cleaner to prevent both from being set at the same time (before the previous validation section), then set `cfg.RelayTxFeeRate = cfg.MinRelayTxFee` if `cfg.MinRelayTxFee` is not the default (i.e it has been set).  That way only a single parse/validate is needed.

In any case, both should not be allowed to be set simultaneously since they conflict.

davecgh:
After thinking about it some more, I'm not sure the new name really reflects the behavior.  It's still a minimum tx fee rate and the new name doesn't convey that.  Since it's in the unit of Satoshi/kB, tx sizes < 1000 scale it downwards appropriately just as sizes > 1000 will scale it upwards.

I understand the confusion of thinking it's an absolute floor, however that can be resolved but comments.

Also, given the fact renaming it would cause churn to the CLI params (which as I previously stated, I'm not a huge fan of), I think I'd prefer not to rename it.  @dajohi, @jrick thoughts?

davecgh:
Rather than churning this PR, how about we just get the logic changes in without the rename, then if we really want to rename, do it in another PR?

lologarithm:
I am ok either way. Let me know soon you decide and I will push this up.

lologarithm:
Hadn't heard anything so I reverted the renaming changes. Another ticket for naming seems like a fairly easy thing to do once the new names and descriptions are chosen.

davecgh:
Thanks.  I was waiting for feedback from @jrick and @dajohi.  However, I think we can safely worry about any potential renames in a different PR.  This one should be ready.  Please go ahead and rebase/squash it to a single commit (Instruction in [btcd Pull Request Management](https://gist.github.com/davecgh/04cb601a147f8c831ca5) if you need it).

dajohi:
OK on this diff.

jrick:
ok

davecgh:
Would you please make the commit message following the [Model Git Commit Message Format](https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md#ModelGitCommitMessages) described in the [Code Contribution Guidelines](https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md)?  Thanks!

lologarithm:
Sure - sorry about that.

davecgh:
OK

PR #538, opened 2015-11-04
server: Allow IPv6 addresses with zone id.

davecgh:
This modifies the IP parsing code to work with IPv6 zone ids.  This is needed since the `net.ParseIP` function does not allow zone ids even though `net.Listen` does.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/538)

### DIFF
--- a/server.go
+++ b/server.go
@@ -14,6 +14,7 @@ import (
 	"net"
 	"runtime"
 	"strconv"
+	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


davecgh:
Bo on IRC (the person who originally reported zone ids weren't parsing correctly) confirmed that this code allows IPV6 addresses with zone ids to be specified.

dajohi:
OK

PR #539, opened 2015-11-05
Fix mempool transactions disappearing if a parent tx is confirmed.

Dirbaio:
There's a bug in the logic that processes new blocks that causes txs to be unnecessarily dropped from mempool.

Basically:

There are two txs A and B in mempool. Tx B uses an output from A.
A new block is received that contains A but not B. 

RemoveTransaction(A) will be called on the mempool, which will recursively remove B. However, B should not be removed because it's not orphan, its parent is now on the blockchain.

This fixes the bug by not recursively removing depending transactions when processing blocks.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/539)

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -1215,11 +1215,11 @@ func (b *blockManager) handleNotifyMsg(notification *blockchain.Notification) {
 		// connected block from the transaction pool.  Secondly, remove any
 		// transactions which are now double spends as a result of these
 		// new transactions.  Finally, remove any transaction that is
-		// no longer an orphan.  Note that removing a transaction from
-		// pool also removes any transactions which depend on it,
-		// recursively.
+		// no longer an orphan. Transactions which depend on a confirmed
+		// transaction are NOT removed recursively because they are still
+		// valid.
 		for _, tx := range block.Transactions()[1:] {
-			b.server.txMemPool.RemoveTransaction(tx)
+			b.server.txMemPool.RemoveTransaction(tx, false)
 			b.server.txMemPool.RemoveDoubleSpends(tx)
 			b.server.txMemPool.RemoveOrphan(tx.Sha())
 			b.server.txMemPool.ProcessOrphans(tx.Sha())
@@ -1261,7 +1261,7 @@ func (b *blockManager) handleNotifyMsg(notification *blockchain.Notification) {
 				// Remove the transaction and all transactions
 				// that depend on it if it wasn't accepted into
 				// the transaction pool.
-				b.server.txMemPool.RemoveTransaction(tx)
+				b.server.txMemPool.RemoveTransaction(tx, true)
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


dajohi:
I much rather prefer RemoveTransaction get the API change like removeTransaction, and get rid of RemoveConfirmedTransaction.   The update all callers of RemoveTransaction() to specify whether they want to remove depending or not.

davecgh:
The concept is correct, but I agree with @dajohi regarding the implementation.

davecgh:
Reviewed 2 of 2 files at r1, 2 of 2 files at r2.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/539)

dajohi:
OK

davecgh:
OK

jrick:
Not sure depending is the best name for those transactions, because depending could also refer to parent txs.  Perhaps removeChildren?

davecgh:
I suppose the name could be confusing.  Given the comment accurately describes it I didn't worry about it, but I suppose `removeRedeemers` or `removeChildren` might be less confusing.

jrick:
+1 for removeRedeemers

Dirbaio:
Changed.

davecgh:
Reviewed 1 of 1 files at r3.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/539)

jrick:
ok

blockmanager.go
```diff
@@ -1215,11 +1215,10 @@ func (b *blockManager) handleNotifyMsg(notification *blockchain.Notification) {
 		// connected block from the transaction pool.  Secondly, remove any
 		// transactions which are now double spends as a result of these
 		// new transactions.  Finally, remove any transaction that is
-		// no longer an orphan.  Note that removing a transaction from
-		// pool also removes any transactions which depend on it,
-		// recursively.
+		// no longer an orphan. Transactions which depend on a confirmed
+		// transaction are NOT removed recursively because they are still valid.
```
dajohi: While here, can we make the above comment limited to 80 chars?

Dirbaio:
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

<sup>**[blockmanager.go, line 1219 \[r3\]](https://reviewable.io:443/reviews/btcsuite/btcd/539#-K2NNo3fQaFNc1u3mdZ1-r3-1219)** ([raw file](https://github.com/btcsuite/btcd/blob/3dd6c396f453e0d9410f7fb1e9e850216f6c456a/blockmanager.go#L1219)):</sup>
Done.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/539)

davecgh:
Reviewed 1 of 1 files at r4.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/539)

dajohi:
OK

PR #540, opened 2015-11-06
blockchain: Finish TODO by improving error message.

jrick:
Removes a TODO by improving the error message.

In the presence of overflow, the actual output sum is still not negative and should not be reported as so.

I also realize that for bitcoin the overflow would have never occurred to begin with, but in other cryptocurrencies 2*MaxBaseUnits (whatever they call them) may overflow an int64 and the total in the error message would be negative.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/540)

### DIFF
--- a/blockchain/validate.go
+++ b/blockchain/validate.go
@@ -240,13 +240,14 @@ func CheckTransactionSanity(tx *btcutil.Tx) error {
 			return ruleError(ErrBadTxOutValue, str)
 		}
 
-		// TODO(davec): No need to check < 0 here as satoshi is
-		// guaranteed to be positive per the above check.  Also need
-		// to add overflow checks.
+		// Two's complement int64 overflow guarantees that any overflow
+		// is detected and reported.  This is impossible for Bitcoin, but
+		// perhaps possible if an alt increases the total money supply.
 		totalSatoshi += satoshi
 		if totalSatoshi < 0 {
 			str := fmt.Sprintf("total value of all transaction "+
-				"outputs has negative value of %v", totalSatoshi)
+				"outputs exceeds max allowed value of %v",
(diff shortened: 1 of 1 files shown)


davecgh:
After reviewing this and discussing it in IRC, I believe we can simply remove the TODO. 

The reason is that while it's true it's only adding positive values and therefore can't be negative under normal circumstances, because it's an int64, if it overflows, it will become negative and thus the negative check will properly detect it and give an error message.  However, we might want to change the error message to indicate overflow since that is what it means.

For example, consider the code with some smaller values:

```Go
package main

import "fmt"

const MaxSatoshi = 125

func main() {
	var totalSatoshi int8
	// The total here would sum to 140, but since it overflows an int8, it'll trigger the error.
	for i := 0; i < 7; i++ {
		satoshi := int8(20)
		// Range checks that are already there
		totalSatoshi += satoshi
		if totalSatoshi < 0 {
			fmt.Printf("total value of all transaction outputs has negative value of %v\n", totalSatoshi)
			return
		}
		if totalSatoshi > MaxSatoshi {
			fmt.Printf("total value of all transaction outputs is %v which is higher than max allowed value of %v", totalSatoshi, MaxSatoshi)
			return
		}
	}
}
```

Output:

```total value of all transaction outputs has negative value of -116```


---

Reviewed 1 of 1 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/540)

davecgh:
Reviewed 1 of 1 files at r2.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/540)

davecgh:
OK

PR #543, opened 2015-11-09
Fix failing test case because of wrong TimeOffset

tuxcanfly:
In rare cases, the build fails with the error:

    --- FAIL: TestPeerConnection (0.01s)
        peer_test.go:296: Running 2 tests
        peer_test.go:155: testPeer: wrong TimeOffset - got -1, want 0

https://travis-ci.org/btcsuite/btcd/jobs/87692240#L542

One reason this might happen is when the current timestamp is `1s` ahead of the message timestamp and the tick happens when the message is in transit.

Since the protocol is limited to one second precision, we do not have enough data to compare the timestamps any further. Therefore, I guess there will be some peer connections which inadvertently have an offset of `1s`. Updated the test case to handle this.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/543)

### DIFF
--- a/peer/peer_test.go
+++ b/peer/peer_test.go
@@ -151,8 +151,11 @@ func testPeer(t *testing.T, p *peer.Peer, s peerStats) {
 		return
 	}
 
-	if p.TimeOffset() != s.wantTimeOffset {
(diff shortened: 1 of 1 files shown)


davecgh:
OK

davecgh:
Needs rebase though.

tuxcanfly:
OK, rebased.

PR #544, opened 2015-11-09
chaincfg: Move DNS Seeds to chaincfg.

dajohi:
This allows API users access to the DNS Seeds for use with SPV
clients, seeders, etc.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/544)


davecgh:
OK

davecgh:
For reference, this is being done since the seeds are needed in wallet for SPV support among other applications.

PR #547, opened 2015-11-11
mempool: Have ProcessTransaction return orphan status.

dajohi:
It is not the responsibility of mempool to relay transactions, so
return whether the newly accepted transaction is an orphan to the
caller.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/547)

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -454,7 +454,7 @@ func (b *blockManager) handleTxMsg(tmsg *txMsg) {
 	// Process the transaction to include validation, insertion in the
 	// memory pool, orphan handling, etc.
 	allowOrphans := cfg.MaxOrphanTxs > 0
-	err := b.server.txMemPool.ProcessTransaction(tmsg.tx,
+	acceptedTxs, err := b.server.txMemPool.ProcessTransaction(tmsg.tx,
 		allowOrphans, true)
 
 	// Remove transaction from request maps. Either the mempool/chain
@@ -489,6 +489,8 @@ func (b *blockManager) handleTxMsg(tmsg *txMsg) {
 			false)
 		return
 	}
+
+	b.server.AnnounceNewTransactions(acceptedTxs)
 }
 
 // current returns true if we believe we are synced with our peers, false if we
@@ -1209,7 +1211,8 @@ func (b *blockManager) handleNotifyMsg(notification *blockchain.Notification) {
 			b.server.txMemPool.RemoveTransaction(tx, false)
 			b.server.txMemPool.RemoveDoubleSpends(tx)
 			b.server.txMemPool.RemoveOrphan(tx.Sha())
-			b.server.txMemPool.ProcessOrphans(tx.Sha())
+			acceptedTxs := b.server.txMemPool.ProcessOrphans(tx.Sha())
+			b.server.AnnounceNewTransactions(acceptedTxs)
 		}
 
 		if r := b.server.rpcServer; r != nil {
--- a/mempool.go
+++ b/mempool.go
@@ -52,10 +52,6 @@ type mempoolConfig struct {
 	// the current best chain.
 	Chain *blockchain.BlockChain
 
-	// RelayNtfnChan defines the channel to send newly accepted transactions
-	// to.  If unset or set to nil, notifications will not be sent.
-	RelayNtfnChan chan *btcutil.Tx
-
 	// SigCache defines a signature cache to use.
 	SigCache *txscript.SigCache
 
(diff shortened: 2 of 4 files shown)


davecgh:
This isn't quite right as is.  A transaction being processed can cause transactions which were previously orphans to no longer be orphans.  Those also need to be relayed when they are accepted to the memory pool.

I suggest returning a list of all transaction hashes that were added to the memory pool as a result of the transaction being added (and the transaction itself for caller simplicity).  That way they can all be relayed by the caller.

dajohi:
Updated to address @davecgh comments.

davecgh:
I'll need to review this more in-depth, but I like the concept.  It separates the server relay logic out of the mempool.

davecgh:
Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

_[mempool.go, line 1092 \[r1\]](https://reviewable.io:443/reviews/btcsuite/btcd/547#-K3bmQ6qjjcvGXPHpVQC:-K3bmQ6qjjcvGXPHpVQD:-1341796361) ([raw file](https://github.com/btcsuite/btcd/blob/b10043e42c3c50842c40296c58df22679afd4f72/mempool.go#L1092)):_
Please update the comment to describe the fact it returns a list of transactions (if any) that were added (like was done with `ProcessTransaction` below).

---

_[mempool.go, line 1104 \[r1\]](https://reviewable.io:443/reviews/btcsuite/btcd/547#-K3bnHbHu5Zos6Qy9_cB:-K3bnHbHu5Zos6Qy9_cC:-1340881350) ([raw file](https://github.com/btcsuite/btcd/blob/b10043e42c3c50842c40296c58df22679afd4f72/mempool.go#L1104)):_
Probably should call out that the list includes the one being added as well.

Perhaps:

```Go
//                        (notice the extra blank line here)
// It returns a slice of transactions added to the mempool.  When the
// error is nil, the list will include the passed transaction itself along
// with any additional orphan transaactions that were added as a result of
// the passed one being accepted.
```

---

_[mempool.go, line 1129 \[r1\]](https://reviewable.io:443/reviews/btcsuite/btcd/547#-K3brfyQWVyARMPwWWJB:-K3brfyRMmGl8rhoNlUF:1274308274) ([raw file](https://github.com/btcsuite/btcd/blob/b10043e42c3c50842c40296c58df22679afd4f72/mempool.go#L1129)):_
I think this should prepend the tx instead of appending it so that the returned slice has the transactions in parent order.  That way when the inventory gets relayed, the remote nodes won't get orphans.

Something like this should do the trick:

```Go
acceptedTxs := make([]*btcutil.Tx, len(newTxs)+1)
acceptedTxs[0] = tx
copy(acceptedTxs[1:], newTxs...)
```

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/547)_

davecgh:
Reviewed 3 of 4 files at r2, 1 of 1 files at r3.

---

*[blockmanager.go, line 550 \[r3\]](https://reviewable.io:443/reviews/btcsuite/btcd/547#-KFAvRbR-TIw7tPkosrQ:-KFAvRbSGkHktWmkSuS0:-2128164890) ([raw file](https://github.com/btcsuite/btcd/blob/1359039d448eb4b4428ca1c5741d7ece3304062f/blockmanager.go#L550)):*
This is identical code as that in the `handleNotifyMsg` code below.  I'd prefer this be refactored into a function that is invoked in both places.

---

*[mempool.go, line 1012 \[r3\]](https://reviewable.io:443/reviews/btcsuite/btcd/547#-KFArXTupC1i8vcNZt0d:-KFArXTupC1i8vcNZt0e:-684995152) ([raw file](https://github.com/btcsuite/btcd/blob/1359039d448eb4b4428ca1c5741d7ece3304062f/mempool.go#L1012)):*
Would you move this after the comment please?  It's cleaner, in my opinion, for the comment to refer to the block of code including the declaration.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/btcsuite/btcd/547)*

davecgh:
Thanks for the updates.  The new function refactored out looks a lot cleaner.  There was one nit left to address and then this is good to go.

---

Reviewed 4 of 4 files at r4.

---

*[mempool.go, line 1012 \[r3\]](https://reviewable.io:443/reviews/btcsuite/btcd/547#-KFArXTupC1i8vcNZt0d:-KFDOAZyTfXDHxjuxcr6:373569341) ([raw file](https://github.com/btcsuite/btcd/blob/1359039d448eb4b4428ca1c5741d7ece3304062f/mempool.go#L1012)):*
Reminder to move this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/btcsuite/btcd/547#-:-KFDOE0IXxgSFFLeNnlF:-1989824280)*

dajohi:
code revised.

davecgh:
Reviewed 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/btcsuite/btcd/547)*

davecgh:
All code review comments have been been addressed and I tested this overnight and reviewed the logs to ensure it's working properly.

OK

PR #549, opened 2015-11-12
Fix build due to hang in peer tests

tuxcanfly:
In rare cases, the build fails due to a hang in the peer tests:

https://travis-ci.org/btcsuite/btcd/jobs/90591996#L550

From the logs, I see that two peers in `TestOutboundPeer` are trying to write to the writer belonging to the same mock connection. Fixed this by using a new connection for each peer.

Also, cleaned up a chan that was not being used.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/549)

### DIFF
--- a/peer/peer_test.go
+++ b/peer/peer_test.go
@@ -569,7 +569,7 @@ func TestOutboundPeer(t *testing.T) {
 	// Test Queue Message
 	fakeMsg := wire.NewMsgVerAck()
 	p.QueueMessage(fakeMsg, nil)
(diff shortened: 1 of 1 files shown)


davecgh:
Thanks.  I hadn't looked into this yet, but I did notice it was happening.

davecgh:
OK.  Needs a rebase.

tuxcanfly:
Thanks, rebased.

PR #550, opened 2015-11-12
Add signature hash calculation tests from Bitcoin Core.

aarbt:
Add 500 tests with various transactions and scripts, verifying that
calcSignatureHash generates the expected hash in each case.

This requires changing SigHashType to uint32; that won't affect the
standard use-cases, but will make calcSignatureHash behave more like the
Core counterpart for non-standard SigHashType settings, like those in
some of these tests.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/550)


davecgh:
Nice!  Thanks for the pull request.  This is definitely useful.  I do have a couple of requests however.

- Rather than converting the test data into Go source, I'd prefer for it to be left as JSON and added to the `txscript/data` directory in the same manner as the `script_valid.json`, `tx_valid.json`, etc.  I almost hate to ask this because I'm sure it was a lot of work to convert it, but it makes it much easier to diff, is more consistent with the existing imported Core tests, and makes it clear the data is intended to be cross implementation test data.
- Move the new test function into the `reference_test.go` file which is intended for the tests which came from Core and change it to read the JSON data and use it akin to how `TestTxValidTests` does it.


---

Reviewed 3 of 3 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/550)

dajohi:
Thank you for doing this!

aarbt:
Thanks for the feedback. I moved the test to reference_test.go and kept the vectors in sighash.json. The format of the json file isn't ideal, but I think I got it parsed reasonably well. Comments are obviously welcome.

davecgh:
Thanks for the updates.  Looks good.

OK


---

Reviewed 2 of 3 files at r2.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/550)

davecgh:
If you will rebase and squash this to a single commit on the latest master, I'll get it merged.

See [Managing a btcd Pull Request for Code Contibutions ](https://gist.github.com/davecgh/04cb601a147f8c831ca5#rebasing-one-of-your-existing-pull-requests) if you need instructions.

aarbt:
Done.

davecgh:
Thanks for squashing.  However, it's still not on the latest master.  This is on top of commit 58e2762 instead of the latest master which is 365b1bd.

aarbt:
Sorry about that; should be fixed now.

PR #551, opened 2015-11-13
mempool:  convert orphansByPrev from a list to a map.

dajohi:
This is more efficient and prevents duplicate entries which can
lead to no-longer-orphans being attempted be added to the mempool
multiple times.

Bug found by me, debugged with @davecgh, and patch from @davecgh

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/551)


davecgh:
OK

PR #553, opened 2015-11-18
Add new configuration option --blocksonly

dajohi:
--blocksonly essentially disables the mempool.  Transactions will
not be accepted nor relayed.

Idea from Bitcoin Core.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/553)

### DIFF
--- a/server.go
+++ b/server.go
@@ -440,6 +440,12 @@ func (sp *serverPeer) OnMemPool(p *peer.Peer, msg *wire.MsgMemPool) {
 // handler this does not serialize all transactions through a single thread
 // transactions don't rely on the previous one in a linear fashion like blocks.
 func (sp *serverPeer) OnTx(p *peer.Peer, msg *wire.MsgTx) {
+	if cfg.BlocksOnly {
+		peerLog.Tracef("Ignoring tx %v from %v - blocksonly enabled",
+			msg.TxSha(), p)
+		return
+	}
+
 	// Add the transaction to the known inventory for the peer.
 	// Convert the raw MsgTx to a btcutil.Tx which provides some convenience
 	// methods and things such as hash caching.
@@ -487,7 +493,36 @@ func (sp *serverPeer) OnBlock(p *peer.Peer, msg *wire.MsgBlock, buf []byte) {
 // accordingly.  We pass the message down to blockmanager which will call
 // QueueMessage with any appropriate responses.
 func (sp *serverPeer) OnInv(p *peer.Peer, msg *wire.MsgInv) {
-	sp.server.blockManager.QueueInv(msg, sp)
+	if !cfg.BlocksOnly {
+		if len(msg.InvList) > 0 {
+			sp.server.blockManager.QueueInv(msg, sp)
+		}
+		return
+	}
+
+	newInv := wire.NewMsgInvSizeHint(uint(len(msg.InvList)))
+	for _, invVect := range msg.InvList {
+		if invVect.Type == wire.InvTypeTx {
+			peerLog.Tracef("Ignoring tx %v in inv from %v -- "+
+				"blocksonly enabled", invVect.Hash, p)
+			if p.ProtocolVersion() >= wire.BIP0037Version {
+				peerLog.Infof("Peer %v is announcing "+
+					"transactions -- disconnecting", p)
+				p.Disconnect()
+				return
+			}
+			continue
+		}
+		err := newInv.AddInvVect(invVect)
+		if err != nil {
+			peerLog.Errorf("Failed to add inventory vector: %v", err)
+			break
+		}
+	}
+
+	if len(newInv.InvList) > 0 {
+		sp.server.blockManager.QueueInv(newInv, sp)
+	}
 }
 
 // OnHeaders is invoked when a peer receives a headers bitcoin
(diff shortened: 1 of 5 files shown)


davecgh:
I haven't reviewed this, but it's failing Travis.

Roasbeef:
Due to the [significant bandwidth savings to be had as a result of the PR](https://bitcointalk.org/index.php?topic=1377345.0), I propose that we expedite the review->testing->merge workflow. 

I've completed my initial review, this PR as is gets an "ok" from me. However, I've yet to test this PR locally. I'll report back with an rpctest using the [proposed `rpctest` package](https://github.com/btcsuite/btcd/pull/371). I feel that an rpctest exercising this PR would be well suited for the newly proposed package. I'll link to the commit adding rpctests for `blocksonly`, @dajohi fee free to cherry-pick the commit to this PR once I post it. 

In my opinion, once `rpctest` is merged, all future PR's adding new daemon behavior should be accompanied by a related rpctest excising the new features.

Roasbeef:
I think we should also include a new config parameter which specifies that transactions should still be fetched/received from peers added via `--connect` and peers connected to as "permanent" via the `btcctl node add` command. There may exist uses cases wherein one would only wish to accept tx's and tx-inv's for a set of white listed peers.

server.go
```diff
@@ -398,6 +398,12 @@ func (sp *serverPeer) OnMemPool(p *peer.Peer, msg *wire.MsgMemPool) {
 // handler this does not serialize all transactions through a single thread
 // transactions don't rely on the previous one in a linear fashion like blocks.
 func (sp *serverPeer) OnTx(p *peer.Peer, msg *wire.MsgTx) {
+	if cfg.BlocksOnly {
+		peerLog.Debugf("Ignoring tx %v from %v - blocksonly enabled",
```
davecgh: This seems like it would be pretty noisy for the Debug level.  Perhaps Trace?

server.go
```diff
@@ -445,7 +451,30 @@ func (sp *serverPeer) OnBlock(p *peer.Peer, msg *wire.MsgBlock, buf []byte) {
...
-	sp.server.blockManager.QueueInv(msg, sp)
+	if !cfg.BlocksOnly {
+		sp.server.blockManager.QueueInv(msg, sp)
+		return
+	}
+
+	newInv := wire.NewMsgInv()
+	for _, invVect := range msg.InvList {
+		if invVect.Type == wire.InvTypeTx {
+			peerLog.Debugf("Ignoring tx %v in inv from %v -- "+
```
davecgh: Same here.

davecgh:
I need to test this still, but it reads well other than the debug noise level I called out.

@Roasbeef:  I agree that it would be ideal to have some type of white listing.  However, since there is no whitelisting infrastructure at all currently, I think that might be better done in a separate series of PRs.  I'd really prefer to avoid holding this one up for that behavior even though it's clearly useful.

Also, did you have get an RPC test using the [proposed `rpctest` package](https://github.com/btcsuite/btcd/pull/371)?

dajohi:
addressed concerns and rebased.

server.go
```diff
@@ -487,7 +493,36 @@ func (sp *serverPeer) OnBlock(p *peer.Peer, msg *wire.MsgBlock, buf []byte) {
...
 func (sp *serverPeer) OnInv(p *peer.Peer, msg *wire.MsgInv) {
-	sp.server.blockManager.QueueInv(msg, sp)
+	if !cfg.BlocksOnly {
+		if len(msg.InvList) > 0 {
+			sp.server.blockManager.QueueInv(msg, sp)
+		}
+		return
+	}
+
+	newInv := wire.NewMsgInv()
```
davecgh: I'd suggest specifying a size hint here in order to cut down on the allocations from growing the backing array.  

```Go
	newInv := wire.NewMsgInvSizeHint(uint(len(msg.InvList)))
```

It does mean the backing array would be larger than it has to be in the case the remote is an older version and txns end up getting filtered out, but that case should be relatively rare.

davecgh:
This appears to handle the client-side concerns correctly of telling the remote we're not interested in receiving inventory about new transactions, disconnecting peers with a higher enough protocol version that aren't respecting that setting, and ignoring the transaction inventory messages from older peers, however, it looks to me like there are a couple of issues where the transaction inventory will be sent even when in this new `--blocksonly` mode.

We can see in `handleRelayInvMsg`:
```Go
			// Don't relay the transaction to the peer when it has
			// transaction relaying disabled.
			if sp.relayTxDisabled() {
				return
			}
```

Which is accurate.  However, if we examine the cases where that flag is set:

* It's set in `OnVersion` to whatever the remote peer requested:
```Go
	// Choose whether or not to relay transactions before a filter command
	// is received.
	sp.setDisableRelayTx(msg.DisableRelayTx)
```

I believe that should also take into account the `--blocksonly` flag.  Something like:
```Go
	// Choose whether or not to relay transactions before a filter command
	// is received and whether or not the server is running in blocks only
	// mode.
	sp.setDisableRelayTx(cfg.BlocksOnly || msg.DisableRelayTx)
```

* The `OnFilterLoad` handler unconditionally enabled transaction inventory relay:
```Go
	sp.setDisableRelayTx(false)
```

This should also take into account the `--blocksonly` flag:
```Go
	// Enable transaction relay when not in blocks only mode.
	if !cfg.BlocksOnly {
		sp.setDisableRelayTx(false)
	}
```

dajohi:
In --blocksonly mode, the only transactions that will make it to mempool and relayed are those from RPC sendrawtransaction.  Your change wouldn't announce any transactions, which would not be good for a wallet connected to it.

server.go
dajohi: Yes, in --blocksonly mode, inv's from most peers should only contain blocks.

davecgh:
I've been testing this for a while now and everything is working as expected.  I've tested both with and without `--blocksonly` and used logging as well as a secondary peer to ensure the behavior is accurate.

OK

PR #554, opened 2015-11-20
Add 0.12.0 deps to deps.txt.

davecgh:
[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/554)


jrick:
http://sprunge.us/AicQ 

Missing golangcrypto, snappy-go, and go-spew if you want to count that one too (although it isn't a btcsuite dep).

davecgh:
Ah yes, we did vendor those.  Good catch.

davecgh:
Updated to include `golangcrypto`and `snappy-go`.

jrick:
ok

PR #556, opened 2015-11-21
peer: Combine stats struct into peer struct.

davecgh:
This defines the peer stat fields directly in the Peer struct instead of defining them in a separate struct and embedding them.  It is slightly more efficient this way and there is really no reason to have them
defined separately anyways since they are not passed around or otherwise used independently.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/556)

### DIFF
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -322,23 +322,6 @@ type stallControlMsg struct {
 	message wire.Message
 }
 
-// stats is the collection of stats related to a peer.
-type stats struct {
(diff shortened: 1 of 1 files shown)


dajohi:
OK

jrick:
This should have identical performance characteristics as before but the change looks ok regardless.

davecgh:
Reviewed 1 of 1 files at r1.

---



_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/556)_

PR #557, opened 2015-11-21
peer: Unexport the mru inventory map.

davecgh:
This unexports the `mruInventoryMap` type since it is only needed within the peer package.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/557)


jrick:
ok

PR #558, opened 2015-11-22
mining: Remove comment and change fee calc to int.

davecgh:
Now that the memory pool minimum fee calculation code is also calculating a more precise value instead of rounding up to the nearest kilobyte boundary, the comment in `NewBlockTemplate` regarding this behavior is no longer accurate.  Thus, this removes the comment.

Also, while here, change the calculation to use an int64 instead of float since it matches the precision of the new calculation code used by the memory pool and can also avoid the need for the slower floating point math.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/558)

### DIFF
--- a/mining.go
+++ b/mining.go
@@ -47,7 +47,7 @@ type txPrioItem struct {
 	tx       *btcutil.Tx
 	fee      int64
 	priority float64
-	feePerKB float64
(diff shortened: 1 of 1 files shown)


dajohi:
OK

PR #559, opened 2015-11-22
mining: Refactor policy into its own struct.

davecgh:
**This pull request requires #558.**

This introduces the concept of a mining policy struct which is used to control block template generation instead of directly accessing the config struct.  This is a step toward decoupling the mining code from the internals of btcd.  Ultimately the intent is to create a separate mining package.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/559)


dajohi:
OK

PR #561, opened 2015-11-22
server: Improve handling of disconnected peers.

davecgh:
When the peer code was refactored, the lists of peers were converted to maps however the code which runs when a peer disconnects still iterates them like a slice.  This is no longer necessary since they are maps
which means the peer can simply be looked up by its ID.

Also, the old code was comparing the map entry and the peer being removed by their pointers which could lead to potentially not properly finding the peer.  This issue is also resolved by this change since it looks up the peer by its ID.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/561)

### DIFF
--- a/server.go
+++ b/server.go
@@ -1117,27 +1117,26 @@ func (s *server) handleDonePeerMsg(state *peerState, sp *serverPeer) {
 	} else {
 		list = state.outboundPeers
 	}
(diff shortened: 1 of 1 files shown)


dajohi:
OK, nice catch.

PR #562, opened 2015-11-22
server: Improve the persistent peer retry logic.

davecgh:
**This pull request requires #561.**

This fixes an issue introduced during the peer refactor where persistent peers that failed the initial connection are not retried as intended.

It also improves the retry logic as follows:

- Make the retry handler goroutine simply use a for loop instead of launching a new goroutine for each backoff attempt.  Even though goroutines are fairly cheap to create, it is much more efficient to simply loop
- Rather than dividing the const interval by 2 everywhere, just half the constant

Finally, include the address of the peer in the error message when a new outbound peer can't be created.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/562)

### DIFF
--- a/server.go
+++ b/server.go
@@ -48,7 +48,7 @@ const (
 	// connectionRetryInterval is the base amount of time to wait in between
 	// retries when connecting to persistent peers.  It is adjusted by the
 	// number of retries such that there is a retry backoff.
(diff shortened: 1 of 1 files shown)


jrick:
ok

PR #563, opened 2015-11-22
mempool/mining: Introduce TxSource interface.

davecgh:
**This pull request requires #559 and #566.**

This introduces the concept of a new interface named `TxSource` which aims to generically provide a concurrent safe source of transactions to be considered for inclusion in a new block.  This is a step towards decoupling the mining code from the internals of btcd.  Ultimately the intent is to create a separate `mining` package.

The new `TxSource` interface relies on a new struct named `miningTxDesc`, which describes each entry in the transaction source.  Once this code is refactored into a separate mining package, the mining prefix can simply be dropped leaving the type exported as `mining.TxDesc`.

To go along with this, the existing `TxDesc` type in the mempool has been renamed to `mempoolTxDesc` and changed to embed the new `miningTxDesc type`.  This allows the mempool to efficiently implement the `MiningTxDescs` method needed to satisfy the `TxSource` interface.

This approach effectively separates the direct reliance of the mining code on the `mempool` and its data structures.  Even though the memory pool will still be the default concrete implementation of the interface,
making it an interface offers much more flexibility in terms of testing and even provides the potential to allow more than one source (perhaps multiple independent relay networks, for example).

Finally, the memory pool and all of the mining code has been updated to implement and use the new interface.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/563)

### DIFF
--- a/cpuminer.go
+++ b/cpuminer.go
@@ -53,6 +53,7 @@ var (
 type CPUMiner struct {
 	sync.Mutex
 	policy            *miningPolicy
+	txSource          TxSource
 	server            *server
 	numWorkers        uint32
 	started           bool
@@ -184,7 +185,7 @@ func (m *CPUMiner) solveBlock(msgBlock *wire.MsgBlock, blockHeight int32,
 
 	// Initial state.
 	lastGenerated := time.Now()
-	lastTxUpdate := m.server.txMemPool.LastUpdated()
+	lastTxUpdate := m.txSource.LastUpdated()
 	hashesCompleted := uint64(0)
 
 	// Note that the entire extra nonce range is iterated and the offset is
@@ -219,7 +220,7 @@ func (m *CPUMiner) solveBlock(msgBlock *wire.MsgBlock, blockHeight int32,
 				// has been updated since the block template was
 				// generated and it has been at least one
 				// minute.
-				if lastTxUpdate != m.server.txMemPool.LastUpdated() &&
+				if lastTxUpdate != m.txSource.LastUpdated() &&
 					time.Now().After(lastGenerated.Add(time.Minute)) {
 
 					return false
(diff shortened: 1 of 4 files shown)


davecgh:
This has been updated based on feedback to specify the interface in `mining.go` (which will eventually become part of as separate mining package) and use a concrete struct for the descriptors for efficiency reasons.

jrick:
I like this version much better than the previous that used interfaces for each entry.

Why not create the mining package right now to avoid switching types around everywhere later?

davecgh:
Well, I suppose we could create the package and have nothing the but the type definitions in it, but the code still needs quite a bit of refactoring before the package can be fully separated.  I was planning on untangling everything by refactoring in steps and then doing a final refactor into the package along with proper documentation.

dajohi:
This diff includes too many changes to review.  I recommend a separate PR first for just the priority function changes.

davecgh:
Alright, I'll see if I can break it down some more.  I think it's going to be hard to break it up much more as this is already pretty well stripped down compared to everything that needs to happen.

davecgh:
I've split the priority bits off into #566 and rebased this on top.

dajohi:
I like where this thread is heading.  OK

davecgh:
Now that the priority code was factored out, we can revisit @jrick's suggestion to create a skeleton WIP mining package now which simply has the type definitions in it.

I think this might make sense because otherwise the mempool package would be blocked until the mining package is done.

@dajohi Thoughts?

davecgh:
I went ahead and created a separate PR to do the skeleton package (#567).

davecgh:
@dajohi / @jrick:  Anything else or is this good to go?

jrick:
ok by me

dajohi:
ok

PR #566, opened 2015-11-25
mempool/mining: Decouple and optimize priority calcs.

davecgh:
This does three things:

- Splits the priority calculation logic from the `TxDesc` type
- Modifies the `calcPriority` function to perform the value age calculation instead of accepting it as a parameter
- Changes the starting priority to be calculated when the transaction is added to the pool

The first is useful as it is a step towards decoupling the mining code from the internals of the memory pool.  Also, the concept of priority is related to mining policy, so it makes more sense to have the calculations separate than being defined on the memory pool tx descriptor.

The second change has been made because everywhere that uses the `calcPriority` function first has to calculate the value age anyways and by making it part of the function it can be avoided altogether in
certain circumstances thereby provided a bit of optimization.

The third change ensures the starting priority is safe for reentrancy which will be important once the mempool is split into a separate package.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/566)

### DIFF
--- a/mempool.go
+++ b/mempool.go
@@ -47,7 +47,7 @@ type TxDesc struct {
 	Added            time.Time   // Time when added to pool.
 	Height           int32       // Blockheight when added to pool.
 	Fee              int64       // Transaction fees.
-	startingPriority float64     // Priority when added to the pool.
+	StartingPriority float64     // Priority when added to the pool.
 }
 
 // txMemPool is used as a source of transactions that need to be mined into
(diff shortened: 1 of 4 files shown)


dajohi:
OK

davecgh:
Reviewed 4 of 4 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/566)*

jrick:
ok

PR #567, opened 2015-11-25
mining: Create skeleton package.

davecgh:
**This pull request requires #563.**

This creates a skeleton mining package that simply contains a few of the definitions used by the mining and mempool code.

This is a step towards decoupling the mining code from the internals of btcd and ultimately will house all of the code related to creating block templates and CPU mining.

The main reason a skeleton package is being created before the full blown package is ready is to avoid blocking mempool separation which relies on these type definitions.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/567)

### DIFF
--- a/cpuminer.go
+++ b/cpuminer.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcd/blockchain"
+	"github.com/btcsuite/btcd/mining"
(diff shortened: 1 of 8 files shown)


dajohi:
Yes please.  OK.

davecgh:
@jrick: Updated.

jrick:
ok

PR #568, opened 2015-11-27
mempool: Introduce mempoolConfig.

dajohi:
This is in preparation of moving mempool to its own subpackage.  No
functional change.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/568)

### DIFF
--- a/mempool.go
+++ b/mempool.go
@@ -50,12 +50,51 @@ type mempoolTxDesc struct {
 	StartingPriority float64
 }
 
+// mempoolConfig is a descriptor containing the memory pool configuration.
+type mempoolConfig struct {
+	// DisableRelayPriority defines whether to relay free or low-fee
+	// transactions that do not have enough priority to be relayed.
+	DisableRelayPriority bool
+
+	// EnableAddrIndex defines whether the address index should be enabled.
+	EnableAddrIndex bool
+
+	// FetchTransactionStore defines the function to use to fetch
+	// transacation information.
+	FetchTransactionStore func(*btcutil.Tx, bool) (blockchain.TxStore, error)
+
+	// FreeTxRelayLimit defines the given amount in thousands of bytes
+	// per minute that transactions with no fee are rate limited to.
+	FreeTxRelayLimit float64
+
+	// MaxOrphanTxs defines the maximum number of orphan transactions to
+	// keep in memory.
+	MaxOrphanTxs int
+
+	// MinRelayTxFee defines the minimum transaction fee in BTC/kB to be
+	// considered a non-zero fee.
+	MinRelayTxFee btcutil.Amount
+
+	// NewestSha defines the function to retrieve the newest sha
+	NewestSha func() (*wire.ShaHash, int32, error)
+
+	// RelayNtfnChan defines the channel to send newly accepted transactions
+	// to.  If unset or set to nil, notifications will not be sent.
+	RelayNtfnChan chan *btcutil.Tx
+
+	// SigCache defines a signature cache to use.
+	SigCache *txscript.SigCache
+
+	// TimeSource defines the timesource to use.
+	TimeSource blockchain.MedianTimeSource
+}
+
 // txMemPool is used as a source of transactions that need to be mined into
 // blocks and relayed to other peers.  It is safe for concurrent access from
 // multiple peers.
 type txMemPool struct {
 	sync.RWMutex
-	server        *server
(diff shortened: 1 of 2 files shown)


davecgh:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[mempool.go, line 46 \[r1\]](https://reviewable.io:443/reviews/btcsuite/btcd/568#-K49DmuqmCpwByuFBpqz:-K49DmuqmCpwByuFBpr-:1246643559) ([raw file](https://github.com/btcsuite/btcd/blob/777e8e7fc0307df946d5ebeea9db77d77b72dda5/mempool.go#L46)):*
I'd name this `EnableAddrIndex` to be more consistent with the rest of the code base which uses the convention `DisableX`/`EnableX`.

---

*[mempool.go, line 69 \[r1\]](https://reviewable.io:443/reviews/btcsuite/btcd/568#-K49E3Wu5DumzkLzLq5r:-K49E3Wu5DumzkLzLq5s:992187142) ([raw file](https://github.com/btcsuite/btcd/blob/777e8e7fc0307df946d5ebeea9db77d77b72dda5/mempool.go#L69)):*
Same here:  `DisableRelayPriority`.

---

*[mempool.go, line 73 \[r1\]](https://reviewable.io:443/reviews/btcsuite/btcd/568#-K49ELMnMtDUVeF6Ch6_:-K49ELMnMtDUVeF6Ch6a:171604828) ([raw file](https://github.com/btcsuite/btcd/blob/777e8e7fc0307df946d5ebeea9db77d77b72dda5/mempool.go#L73)):*
I think this should be optional.  In other words, specify in the comment that it can be `nil` if the caller is not interested in receiving them.

---

*[mempool.go, line 980 \[r1\]](https://reviewable.io:443/reviews/btcsuite/btcd/568#-K49EYgwDo8iU6H7g3V7:-K49EYgx-ZfMRkMUbBue:-1444543345) ([raw file](https://github.com/btcsuite/btcd/blob/777e8e7fc0307df946d5ebeea9db77d77b72dda5/mempool.go#L980)):*
As mentioned previously, I think this should be made optional.  That would require this to be:

```Go
if mp.cfg.RelayNtfnChan != nil {
	mp.cfg.RelayNtfnChan <- tx
}
```

---

*[mempool.go, line 1077 \[r1\]](https://reviewable.io:443/reviews/btcsuite/btcd/568#-K49FwM9NX6FC3k_ak_k:-K49FwM9NX6FC3k_ak_l:-693307963) ([raw file](https://github.com/btcsuite/btcd/blob/777e8e7fc0307df946d5ebeea9db77d77b72dda5/mempool.go#L1077)):*
I'd prefer this to take a pointer and then copy it when setting the `cfg` field below.  This, of course, means the `txMemPool` should by of type `txConfig` instead of `*txConfig`.  The reason for this is it will avoid a ton of pointer derefs all throughout the mempool code.

---

*[server.go, line 1906 \[r1\]](https://reviewable.io:443/reviews/btcsuite/btcd/568#-K49GR_vJQkncxZ2fsKg:-K49GR_vJQkncxZ2fsKh:-884996826) ([raw file](https://github.com/btcsuite/btcd/blob/777e8e7fc0307df946d5ebeea9db77d77b72dda5/server.go#L1906)):*
This needs to pass the `tx` as the second parameter as the filter handling in `handleRelayInvMsg` requires it.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/568)*

dajohi:
updated.

davecgh:
Reviewed 2 of 2 files at r2, 1 of 1 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/568)*

davecgh:
Also, I'd call this `mempoolConfig` so when it is all split into the mempool package it will simply become `mempool.Config`.

dajohi:
pushed.

davecgh:
Reviewed 2 of 2 files at r4.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/568)*

davecgh:
OK

davecgh:
Reviewed 1 of 1 files at r5.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/568)*

PR #569, opened 2015-11-29
rpcserver: Optimize filteraddr code.

davecgh:
This optimizes the filter address handling code in the RPC server `handleSearchRawTransasctions` path in a few ways:

- Only normalize the filter addresses provided in the command once rather than for every transaction being returned
- Reset the passes filter flag just before it's used
- Use a local for the encoded address to avoid the bounds checking overhead of accessing it from the slice
- Avoiding subsequent map lookups once the filter has already passed

Also, while here, modify a few of the related comments to match code style and consistency.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/569)


jrick:
ok

PR #570, opened 2015-12-01
policy:  Introduce policy package.

dajohi:
[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/570)

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -16,6 +16,7 @@ import (
 	"github.com/btcsuite/btcd/blockchain"
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/database"
+	"github.com/btcsuite/btcd/policy"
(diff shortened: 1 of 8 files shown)


davecgh:
I don't think policy should be a separate package.  There is already a `miningPolicy` and this is dealing with the mempool policy.  In my opinion, it should be a separate type `mempoolPolicy` which eventually would become `mempool.Policy` when the package splits.

jrick:
Agree with @davecgh.  The policy here is a policy of the tx source, in this case being mempool.  Other transaction sources (tests for the mining package come to mind) may have different policy requirements and a one size fits all policy type wouldn't work here.

If you wanted to further filter a transaction source with additional policy rules, wrap the tx source in a new type implementing that same interface.

dajohi:
Yes, new diff baking

PR #571, opened 2015-12-01
mempool:  Add a skeleton mempool pkg

dajohi:
[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/571)

### DIFF
--- a/config.go
+++ b/config.go
@@ -26,27 +26,29 @@ import (
 )
 
 const (
-	defaultConfigFilename    = "btcd.conf"
-	defaultDataDirname       = "data"
-	defaultLogLevel          = "info"
-	defaultLogDirname        = "logs"
-	defaultLogFilename       = "btcd.log"
-	defaultMaxPeers          = 125
-	defaultBanDuration       = time.Hour * 24
-	defaultBanThreshold      = 100
-	defaultMaxRPCClients     = 10
-	defaultMaxRPCWebsockets  = 25
-	defaultVerifyEnabled     = false
-	defaultDbType            = "leveldb"
-	defaultFreeTxRelayLimit  = 15.0
-	defaultBlockMinSize      = 0
-	defaultBlockMaxSize      = 750000
-	blockMaxSizeMin          = 1000
-	blockMaxSizeMax          = wire.MaxBlockPayload - 1000
-	defaultBlockPrioritySize = 50000
-	defaultGenerate          = false
-	defaultAddrIndex         = false
-	defaultSigCacheMaxSize   = 50000
+	defaultConfigFilename        = "btcd.conf"
+	defaultDataDirname           = "data"
+	defaultLogLevel              = "info"
+	defaultLogDirname            = "logs"
+	defaultLogFilename           = "btcd.log"
+	defaultMaxPeers              = 125
+	defaultBanDuration           = time.Hour * 24
+	defaultBanThreshold          = 100
+	defaultMaxRPCClients         = 10
+	defaultMaxRPCWebsockets      = 25
+	defaultVerifyEnabled         = false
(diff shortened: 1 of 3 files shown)


davecgh:
I'd prefer to not create the skeleton package unless it's absolutely needed.  The mining case was required because mempool separation would be blocked.  In this case, there is nothing else that needs `mempoolPolicy`.  In other words, I'd just introduce `type mempoolPolicy struct` and switch the code to make use of it (very close to what you've done here, but without the package).

dajohi:
How does this look?

config.go
```diff
@@ -337,7 +337,7 @@ func loadConfig() (*config, []string, error) {
 		BlockMaxSize:      defaultBlockMaxSize,
 		BlockPrioritySize: defaultBlockPrioritySize,
 		SigCacheMaxSize:   defaultSigCacheMaxSize,
-		MaxOrphanTxs:      maxOrphanTransactions,
+		MaxOrphanTxs:      1000,
```
davecgh: This should be a `defaultMaxOrphanTransactions` constant to be consistent with the surrounding code.

mempool.go
```diff
@@ -64,21 +46,12 @@ type mempoolConfig struct {
...
-	MaxOrphanTxs int
-
-	// MinRelayTxFee defines the minimum transaction fee in BTC/kB to be
-	// considered a non-zero fee.
-	MinRelayTxFee btcutil.Amount
-
 	// NewestSha defines the function to retrieve the newest sha
 	NewestSha func() (*wire.ShaHash, int32, error)
 
+	// Policy
```
davecgh: The comment is incomplete.

mempool.go
```diff
@@ -1103,9 +1110,10 @@ func (mp *txMemPool) LastUpdated() time.Time {
 
 // newTxMemPool returns a new memory pool for validating and storing standalone
 // transactions until they are mined into a block.
-func newTxMemPool(cfg *mempoolConfig) *txMemPool {
+func newTxMemPool(policy *mempoolPolicy, cfg *mempoolConfig) *txMemPool {
```
davecgh: The `mempoolConfig` struct has the `Policy` field in it.  There shouldn't be a need to pass them both.

mempool.go
```diff
@@ -1103,9 +1110,10 @@ func (mp *txMemPool) LastUpdated() time.Time {
...
 	memPool := &txMemPool{
 		cfg:           *cfg,
+		policy:        *policy,
```
davecgh: `cfg.Policy`

server.go
```diff
@@ -2362,19 +2362,24 @@ func newServer(listenAddrs []string, db database.Db, chainParams *chaincfg.Param
...
 		EnableAddrIndex:       cfg.AddrIndex,
 		FetchTransactionStore: s.blockManager.blockChain.FetchTransactionStore,
-		FreeTxRelayLimit:      cfg.FreeTxRelayLimit,
-		MaxOrphanTxs:          cfg.MaxOrphanTxs,
-		MinRelayTxFee:         cfg.minRelayTxFee,
 		NewestSha:             s.db.NewestSha,
 		RelayNtfnChan:         s.relayNtfnChan,
 		SigCache:              s.sigCache,
 		TimeSource:            s.timeSource,
 	}
```
davecgh:
```Go
	Policy: mempoolPolicy{
		...
	},
```

davecgh:
This looks good overall.  In addition to addressing the line comments, this needs a rebase since it can't be merged cleanly.

dajohi:
Rebased.

davecgh:
Thanks for addressing the review items.  I've reviewed all of the updates and tested them.

OK

PR #572, opened 2015-12-02
wire: Implement sendheaders command (BIP0130)

dajohi:
This implements the wire protocol encoding portion of a new sendheaders message as described by BIP0130.  It purpose is to request that a peer sends header commands instead of inv commands when announcing new blocks.  This includes a protocol version bump to 70012 and a wire version bump to 0.4.0.

Note that this does not implement logic to handle the command in btcd, rather it only makes the command available at the wire protocol level.  A future commit which honors the command and therefore provides full BIP0130 support is still required.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/572)

### DIFF
--- a/wire/msgsendheaders.go
+++ b/wire/msgsendheaders.go
@@ -1,0 +1,60 @@
+// Copyright (c) 2016 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package wire
+
+import (
+	"fmt"
+	"io"
+)
+
+// MsgSendHeaders implements the Message interface and represents a bitcoin
+// sendheaders message.  It is used to request the peer send block headers
+// rather than inventory vectors.
+//
+// This message has no payload and was not added until protocol versions
+// starting with SendHeadersVersion.
+type MsgSendHeaders struct{}
+
+// BtcDecode decodes r using the bitcoin protocol encoding into the receiver.
+// This is part of the Message interface implementation.
+func (msg *MsgSendHeaders) BtcDecode(r io.Reader, pver uint32) error {
+	if pver < SendHeadersVersion {
+		str := fmt.Sprintf("sendheaders message invalid for protocol "+
+			"version %d", pver)
+		return messageError("MsgSendHeaders.BtcDecode", str)
+	}
+
+	return nil
+}
+
+// BtcEncode encodes the receiver to w using the bitcoin protocol encoding.
+// This is part of the Message interface implementation.
+func (msg *MsgSendHeaders) BtcEncode(w io.Writer, pver uint32) error {
+	if pver < SendHeadersVersion {
+		str := fmt.Sprintf("sendheaders message invalid for protocol "+
+			"version %d", pver)
+		return messageError("MsgSendHeaders.BtcEncode", str)
+	}
+
+	return nil
+}
+
+// Command returns the protocol command string for the message.  This is part
+// of the Message interface implementation.
+func (msg *MsgSendHeaders) Command() string {
+	return CmdSendHeaders
+}
+
+// MaxPayloadLength returns the maximum length the payload can be for the
+// receiver.  This is part of the Message interface implementation.
+func (msg *MsgSendHeaders) MaxPayloadLength(pver uint32) uint32 {
+	return 0
+}
+
+// NewMsgSendHeaders returns a new bitcoin sendheaders message that conforms to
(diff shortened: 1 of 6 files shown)


davecgh:
I think it's worthwhile to call out in the commit description that this only implements support for the command at a protocol level and does not include support in btcd itself.

wire/msgsendheaders.go
```diff
@@ -0,0 +1,60 @@
...
+
+package wire
+
+import (
+	"fmt"
+	"io"
+)
+
+// MsgSendHeaders implements the Message interface and represents a bitcoin
+// sendheaders message.  It is used to request the peer send header's rather
```
davecgh: `header's` -> `block headers`
`inv's` -> `inventory vectors`

wire/msgsendheaders_test.go
```diff
@@ -0,0 +1,66 @@
...
+package wire_test
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/btcsuite/btcd/wire"
+)
+
+func TestSendHeaders(t *testing.T) {
```
davecgh: Missing comment.

davecgh: Missing comment.

davecgh:
Please add the following tests using the tests in `msgping_test.go` as a guide:

```Go
// TestSendHeaders tests the MsgSendHeaders API against the protocol version
// SendHeadersVersion.
func TestSendHeadersBIP130(t *testing.T) {
...
}

// TestSendHeadersCrossProtocol tests the MsgSendHeaders API when encoding with
// the latest protocol version and decoding with SendHeadersVersion.
func TestSendHeadersCrossProtocol(t *testing.T) {
...
}

// TestSendHeadersWire tests the MsgSendHeaders wire encode and decode for
// various protocol versions.
func TestSendHeadersWire (t *testing.T) {
...
}

// TestSendHeadersWireErrors performs negative tests against wire encode and
// decode of MsgSendHeaders to confirm error paths work correctly.
func TestSendHeadersWireErrors (t *testing.T) {
...
}
```

davecgh:
The commit and PR description should call out this is part of the BIP0130 implementation.

davecgh:
Should add support for the BIP to `doc.go` as well:


`	BIP0130	(https://github.com/bitcoin/bips/blob/master/bip-0130.mediawiki)`

wire/msgsendheaders.go
```diff
@@ -0,0 +1,60 @@
...
+	return CmdSendHeaders
+}
+
+// MaxPayloadLength returns the maximum length the payload can be for the
+// receiver.  This is part of the Message interface implementation.
+func (msg *MsgSendHeaders) MaxPayloadLength(pver uint32) uint32 {
+	return 0
+}
+
+// NewMsgSendHeaders returns a new bitcoin pong message that conforms to the
```
davecgh: `pong` -> `sendheaders`
`MsgPong` -> `MsgSendHeaders`

wire/msgsendheaders_test.go
```diff
@@ -0,0 +1,188 @@
...
+	"testing"
+
+	"github.com/btcsuite/btcd/wire"
+	"github.com/davecgh/go-spew/spew"
+)
+
+// TestSendHeaders tests the MsgSendHeaders API against the latest protocol
+// version.
+func TestSendHeaders(t *testing.T) {
+	pver := wire.SendHeadersVersion
```
davecgh: This should be `wire.ProtocolVersion` since the point of this test is to ensure it works with the latest protocol version.

wire/msgsendheaders_test.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		t.Errorf("decode of MsgSendHeaders failed [%v] err <%v>", buf,
+			err)
+	}
+}
+
+// TestSendHeadersWire tests the MsgSendHeaders wire encode and decode for
+// various protocol versions.
+func TestSendHeadersWire(t *testing.T) {
+	tests := []struct {
```
davecgh: For consistency, I'd suggest this be done like `msgfilterclear_test.go`.  In particular:

```Go
msgSendHeaders := wire.NewMsgSendHeaders()
msgSendHeadersEncoded := []byte{}
```

Then the tests below would use those variables.

davecgh:
The wire version should also be bumped to 0.3.1.

davecgh:
Looks good now.

OK

PR #574, opened 2015-12-03
peer: Add DisableRelayTx to config.

dajohi:
DisableRelayTx sets the DisableRelayTx value in the version
message which informs the remote peer on whether to send
inv messages for transactions.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/574)


davecgh:
OK

PR #578, opened 2015-12-08
server: Correct mempool/cpu miner initialize order.

davecgh:
The CPU miner relies on the mempool, so the mempool has to be created before calling the function to create the CPU miner.  When PR #568 introduced the mempool config struct, it moved the mempool creation
after the miner creation, which leads to the CPU miner crashing due to trying to access a nil mempool.

This move the CPU miner creation after the mempool creation appropriately.

Once the mempool and miner separation is done this won't matter anymore since the miner config will require a `mining.TxSource` which effectively enforces its creation first.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/578)

### DIFF
--- a/server.go
+++ b/server.go
@@ -2354,15 +2354,6 @@ func newServer(listenAddrs []string, db database.Db, chainParams *chaincfg.Param
 	}
 	s.blockManager = bm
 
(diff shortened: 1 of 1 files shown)


dajohi:
OK

PR #586, opened 2015-12-27
peer: remove getaddr msg from stall detection.

Dirbaio:
The getaddr msg is usually replied to with an addr msg, but if
the other peer does not have any addresses to share it will not
reply at all (instead of replying with an addr msg with 0 addresses).
Therefore, the getaddr msg is not guaranteed a reply, so this commit
removes it from stall detection to avoid incorrectly kicking
such peers.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/586)


davecgh:
OK

PR #587, opened 2015-12-27
server: remove peer from peerState even when doing a retry. Fixes #577

### LINKED ISSUE #577: Max peers possibly mis-counted
I have a network setup with several btcd nodes but only one connects out to the internet, and the others sync with it over the LAN.  
A few times I've gotten loops where it keeps disconnecting and reconnecting to internal nodes.
It seems it may be counting connections that are in the process of closing and getting mixed up due to that.

10.0.30.10's btcd.conf:
```
testnet=1
addrindex=1
rpcuser= -
rpcpass= -

nodnsseed=1

maxpeers=5
connect=na
connect=104.131.6.40
connect=122.18.83.210

nolisten=0

minrelaytxfee=0.00001
rpclisten=
notls=0
```

Log example of the internet connected node (10.0.30.10):
```
17:02:00 2015-12-07 [INF] SRVR: Max peers reached [5] - disconnecting peer na:18333 (outbound)
17:02:00 2015-12-07 [INF] BMGR: New valid peer na:18333 (outbound) (/btcwire:0.3.0/btcd:0.12.0/)
17:02:00 2015-12-07 [INF] BMGR: Lost peer na:18333 (outbound)
17:02:05 2015-12-07 [INF] BMGR: New valid peer na:18333 (outbound) (/btcwire:0.3.0/btcd:0.12.0/)
17:02:05 2015-12-07 [INF] SRVR: Max peers reached [5] - disconnecting peer na:18333 (outbound)
17:02:05 2015-12-07 [INF] BMGR: Lost peer na:18333 (outbound)
```
... repeats every 5 seconds till I kill btcd.

and an example log from "na" (10.0.30.11) the node being disconnected:
```
04:01:22 2015-12-08 [INF] BMGR: Syncing to block height 617134 from peer 10.0.30.10:35848
04:01:22 2015-12-08 [INF] BMGR: Lost peer 10.0.30.10:35848 (inbound)
04:01:27 2015-12-08 [INF] BMGR: New valid peer 10.0.30.10:35850 (inbound) (/btcwire:0.3.0/btcd:0.12.0/)
04:01:27 2015-12-08 [INF] BMGR: Syncing to block height 617134 from peer 10.0.30.10:35850
04:01:27 2015-12-08 [INF] BMGR: Lost peer 10.0.30.10:35850 (inbound)
04:01:32 2015-12-08 [INF] BMGR: New valid peer 10.0.30.10:35851 (inbound) (/btcwire:0.3.0/btcd:0.12.0/)
04:01:32 2015-12-08 [ERR] PEER: Can't read message from 10.0.30.10:35851 (inbound): read tcp4 10.0.30.11:18333->10.0.30.10:35851: read: connection reset by peer
04:01:32 2015-12-08 [INF] BMGR: Syncing to block height 617134 from peer 10.0.30.10:35851
04:01:32 2015-12-08 [INF] BMGR: Lost peer 10.0.30.10:35851 (inbound)
04:01:37 2015-12-08 [INF] BMGR: New valid peer 10.0.30.10:35853 (inbound) (/btcwire:0.3.0/btcd:0.12.0/)
```

restarting the internet facing node fixes it, but the problem comes back and the walled off node will fall behind on blocks.

Hope this helps and can provide more logs / test things.  Thanks!

Dirbaio:
The issue is that when the new connection is established, handleAddPeerMsg will be called later when the connection is established, the same way as regular (non-retry) connections. So the old peer needs to be deleted anyway, and the new one doesn't need to be added now. Otherwise entries from dead peers in the peerState stack up and hit the maxPeers limit.

I'm not sure if this is the best way to fix the issue since I'm not very familiar with the peer/server code. I'm open to alternate suggestions.

I'm able to reliably reproduce #577 in my setup, and I have tested this PR reliably fixes the issue: setting maxPeers to 1, using --connect, and causing the connection to fail and retry does not trigger the bug anymore, btcd will keep happily retrying forever.

Fixes #577.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/587)

### DIFF
--- a/server.go
+++ b/server.go
@@ -1124,12 +1124,10 @@ func (s *server) handleDonePeerMsg(state *peerState, sp *serverPeer) {
 		// persistent outbound connection.
 		if !sp.Inbound() && sp.persistent && atomic.LoadInt32(&s.shutdown) == 0 {
 			// Retry peer
-			sp = s.newOutboundPeer(sp.Addr(), sp.persistent)
-			if sp != nil {
-				go s.retryConn(sp, false)
+			sp2 := s.newOutboundPeer(sp.Addr(), sp.persistent)
+			if sp2 != nil {
+				go s.retryConn(sp2, false)
 			}
-			list[sp.ID()] = sp
-			return
 		}
 		if !sp.Inbound() && sp.VersionKnown() {
 			state.outboundGroups[addrmgr.GroupKey(sp.NA())]--


davecgh:
From looking at this, it appears the intention was to simply replace the existing peer with the new one in the map.  The issue here then is the fact that it's not removing the old peer ID before adding the one which of course results in the old one never being removed.

However, the fix here is not quite right because it will result in the outbound groups being miscounted as well as not having any entry in the map of outbound peers.  Also, the debug message shouldn't be saying the peer was removed from the server because it hasn't been given it's being retried.  Given all of that, I would suggest the following:

```Go
		if !sp.Inbound() && sp.persistent && atomic.LoadInt32(&s.shutdown) == 0 {
			// Retry peer
			newSP := s.newOutboundPeer(sp.Addr(), sp.persistent)
			if newSP != nil {
				delete(list, sp.ID())
				list[newSP.ID()] = newSP
				go s.retryConn(newSP, false)
				return
			}
		}
		if !sp.Inbound() && sp.VersionKnown() {
			...
```

Notice a few things here:
* The logic to remove the old ID, add the new ID, start the goroutine, and return is now inside the != nil check so either all of it, or none of it, happens.  If the new outbound peer fails to be created for some reason, the entire section will not run which will result in the code falling through to the normal cleanup code.
* The old ID is deleted from the map prior to adding the new one
* The new ID is added to the map _before_ the retry goroutine is started.  While this currently doesn't matter as nothing in that goroutine accesses the map, it's more future proof if that should ever change.

Dirbaio:
It seems IDs are assigned when the connection is complete (on receiving the version msg). Right after calling newOutboundPeer the id is 0.
https://github.com/btcsuite/btcd/blob/89af747603fad09130129fa4688eaab53c0abdba/peer/peer.go#L1024

Not sure how to deal with this...

davecgh:
Ah, yes you're right about the ID.  So, tracing through the code I can see that the retry handler calls `establishConn` which in turn call `Connect` on the peer, which in turn leads to the version message that is handled by  `OnVersion` which ultimately ends up calling `handleAddPeerMsg`.  That function handles updating the outbound groups counts and adding it back to the appropriate map (`persistentPeers` in this case).

So, in short, my comment above is incorrect in regards to your PR causing the counts to be incorrect and your PR is correct to let it fall through to handle decrementing the counts and removing it from the map (since it is ultimately added by later as described).

The only thing arguably incorrect about it then is the debug message saying the peer is being removed.  However, `retryConn` logs that it's being retried and `establishConn` logs the attempted and successful connections.  Given that, I don't see a big issue with leaving the removal debug message.

I also added some additional prints and tested this PR out to prove the correct behavior in regards to the group counting and also confirmed it fixes #577.  While doing so, I also noticed the persistent peer is not being added to the pending peers list in various places, so that could end up causing an additional connection to be made only to later be removed due to exceeding the limit of outbound peers.  However, that is a separate to this one and should not hold up this PR.

OK

davecgh:
Would you make the commit message a [Model Git Commit Message](https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md#44-code-documentation-and-commenting).

Dirbaio:
Done. I hope the commit message is clearer now.

davecgh:
Thanks.  Commit message looks good.

PR #590, opened 2015-12-29
txscript: Fix docs to match function.

wallclockbuilder:
Changed the order of return values described in the docs to be
consistent with the function’s actual return value signature.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/590)


davecgh:
OK

PR #593, opened 2016-01-03
Replace *KoblitzCurve by elliptic.Curve

mvdan:
Found via github.com/mvdan/interfacer.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/593)


davecgh:
Thanks.  I don't have any issues with this given the current code.  While it doesn't affect this particular PR, I'd actually like to move away from `crypto/elliptic` altogether in the future.

OK

PR #594, opened 2016-01-05
peer: declare QueueMessage()'s doneChan as send only.

dajohi:
This ensures the channel passed to QueueMessage is writable and that
QueueMessage will not read from the channel (write-only).

This change is merely a safety change.  If a user of the API passes
a read-only channel to QueueMessage, it will now be caught at compile
time instead of panicking during runtime.

Also update internal functions.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/594)


davecgh:
Looks good.  It reds and and I've also tested the code syncs as expected with it.

OK

PR #595, opened 2016-01-05
chaincfg: Remove testnet-seed.alexykot.me, it is defunct.

dajohi:
[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/595)


davecgh:
OK

PR #598, opened 2016-01-11
txscript: optimize sigcache lookup

### LINKED ISSUE #575: [txscript] Sigcache Implementation Needs Optimization
I was doing a bit of profiling for other reasons tonight and I noticed the recently added signature cache is taking way more CPU time than it should.  The following profile was generated when doing an initial block download, so there would be very few hits of the signature cache, nevertheless, as I've highlighted in the image, the signature cache additions and lookups are actually taking more time than the signature verifications!

sigcachecpuprof

I'm fairly certain this is the result of using a struct as the map key, which, in my experience, typically results in poor performance and this case appears to be no exception.  I suspect making the map key a smaller fixed size array and using a fast hashing function (does not need to be a cryptographic hash function) to hash just the signature, then using that as the key would make a huge difference.  Note that doing this would obviously also require making the struct that currently acts as the key the value for the entry and then ensuring that any hits are actually matches (since collisions would now be possible).  In practice, the number of collisions that would require additional checking will be minimal and the more expensive full check would barely ever have to run in the case of cache misses.

Roasbeef:
Profiles discovered that lookups into the signature cache included an
expensive comparison to the stored `sigInfo` struct. This lookup had the
potential to be more expensive than directly verifying the signature
itself!

In addition, evictions were rather expensive because they involved
reading from /dev/urandom, or equivalent, for each eviction once the
signature cache was full as well as potentially iterating over every
item in the cache in the worst-case.

To remedy this poor performance several changes have been made:
* Change the lookup key to the fixed sized 32-byte signature hash
* Perform a full equality check only if there is a cache hit which
    results in a significant  speed up for both insertions and existence
checks
* Override entries in the case of a colliding hash on insert Add an
* .IsEqual() method to the Signature and PublicKey types in the
  btcec package to facilitate easy equivalence testing
* Allocate the signature cache map with the max number of entries in
  order to avoid unnecessary map re-sizes/allocations
* Optimize evictions from the signature cache Delete the first entry
* seen which is safe from manipulation due to
    the pre image resistance of the hash function
* Double the default maximum number of entries within the signature
  cache due to the reduction in the size of a cache entry
  * With this eviction scheme, removals are effectively O(1)

Fixes #575.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/598)


Roasbeef:
As pointed out in #575 by @wallclockbuilder, when certain types are used as keys within a map, they benefit from a special fast-path within the hashing operations for `map`. They types that are able to take advantage of the fast-path include: `int32`, `int64` and `string`. With this new PR, the `sigCache` keys are now just `[256]byte`, however, we may be able to gain an additional speed up by switching the key to simply be a `string`.

txscript/sigcache.go
```diff
@@ -7,18 +7,31 @@ package txscript
...
-// sigInfo represents an entry in the SigCache. Entries in the sigcache are a
-// 3-tuple: (sigHash, sig, pubKey).
-type sigInfo struct {
-	sigHash wire.ShaHash
-	sig     string
-	pubKey  string
+// sigKey represents the lookup key of the signature cache.
+type sigKey [wire.HashSize]byte
+
+// newSigKey creates a new sigcache lookup key using the passed paramters. This
```
wallclockbuilder: +// newSigKey creates a new sigcache lookup key using the passed paramters~~parameters~~. This

txscript/sigcache.go
```diff
@@ -7,18 +7,31 @@ package txscript
...
+// lookup key is is the result of: SHA-256(nonce || sigHash || signature || pubkey).
```
wallclockbuilder: +// lookup key is ~~is~~ the result of: SHA-256(nonce || sigHash || signature || pubkey).

txscript/sigcache.go
```diff
@@ -7,18 +7,31 @@ package txscript
 import (
 	"bytes"
 	"crypto/rand"
+	"crypto/sha256"
```
dajohi: @davecgh should we use github.com/btcsuite/fastsha256 here?

dajohi:
@Roasbeef Could you write some benchmarks for before and after?

txscript/sigcache.go
```diff
@@ -42,8 +56,19 @@ type SigCache struct {
...
 // cache to exceed the max.
-func NewSigCache(maxEntries uint) *SigCache {
-	return &SigCache{validSigs: make(map[sigInfo]struct{}), maxEntries: maxEntries}
+func NewSigCache(maxEntries uint) (*SigCache, error) {
+	cache := &SigCache{
+		validSigs:  make(map[sigKey]struct{}),
+		maxEntries: maxEntries,
+	}
+
+	// Read a 32 byte nonce to use as a salt the SHA-256 invocations for
```
wallclockbuilder: +	// Read a 32 byte nonce to use as a salt **for** the SHA-256 invocations for

Roasbeef:
So, I ran some preliminary benchmarks. The results indicate that this PR, in terms of CPU efficiency, is actually *inferior*. 

For the benchmarks, I created a [new branch](https://github.com/Roasbeef/btcd/tree/sigcache-optimize-bench), with **4** versions of the `SigCache`. 
  * `SigCacheOld` 
      * The version of the `SigCache` currently checked into master. 
  * `SigCacheOldString` 
      * `SigCacheOld`, with one slight modification. The `sigInfo` struct has been changed such that, `sigHash` is now a converted into a `string` instead of a `wire.ShaHash`. The justification for this change is the following: if the key to a map is a struct, and all attributes within that struct are of type `string`, then the entire struct is able to take advantage of the special [optimized fast path](https://github.com/golang/go/commit/423b0cc25ad8f75825db70a370ea9b696a17074d) available for `string`, `int`, and `int64. 
  * `SigCacheNew`
      * The new version of the `SigCache` proposed within this PR. 
  * `SigCacheNewString`
      * The same as `SigCacheNew`, however instead of storing a `wire.ShaHash` as the `sigKey`, a `string` is instead stored. The justification here is identical to `SigCacheOldString`: there a special optimizations in the compiler to avoid the extra copy when converting a type into a string for a map key, and the special fast path. 

Finally, the result of the preliminary benchmark. Both the `Add` and `Exists` operations of the `SigCache` we're benchmarked across all four implementations. 

```
BenchmarkSigCacheOldExists-4             2000000               960 ns/op
BenchmarkSigCacheOldStringExists-4       2000000               932 ns/op
BenchmarkSigCacheNewExists-4              500000              2638 ns/op
BenchmarkSigCacheNewStringExists-4        500000              2708 ns/op
BenchmarkSigCacheOldAdd-4                1000000              1078 ns/op
BenchmarkSigCacheOldStringAdd-4          1000000              1081 ns/op
BenchmarkSigCacheNewAdd-4                 500000              2745 ns/op
BenchmarkSigCacheNewStringAdd-4           500000              2867 ns/op
ok      github.com/btcsuite/btcd/txscript       13.609s
```

It's evident from the following benchmark, that the *old* `SigCache` with the `sigInfo` struct consisting of all `string` types is actually the most efficient. This version doesn't require the expensive `sha256` operation, and takes advantage of certain compiler optimizations. However, the clear downside to this approach is the large size of each entry within the `SigCache`. With `SigCacheNew`, each entry (ignoring map overhead) is 32-bytes. With `SigCacheOldString` each entry (again, ignoring overhead) is 135 bytes (32 + 71/72 + 32). 

A benchmark of checking for membership within the `SigCache` vs. performing full signature verification seems to contradict the results of profiling which prompted this PR. 
```
BenchmarkSigCacheOldStringExists-4       2000000               932 ns/op
BenchmarkSigVerify-4                        2000           1096216 ns/op
ok      github.com/btcsuite/btcd/txscript       5.166s
```

jrick:
I've been playing around with this for a small while now and I suggest using the following map type for the cache as a starting point:

```
type sigCacheEntry struct {
    signature *btcec.Signature
    publicKey *btcec.PublicKey
}

type SigCache struct {
    // ...
    validSigs map[wire.ShaHash]sigCacheEntry // key is the sighash
}
```

The reason I'd suggest this is that in my benchmarks it is much more efficient to insert these keys and values.  I'm seeing about a 5000% increase here compared to @Roasbeef's version (~1947ns down to 38), and a 2000% improvement compared to master (~791ns to 38).

This is only measuring inserts and I have yet to do any testing on cache existence yet, but I'd expect good gains with this data structure as well since it's creating the exact same map lookup keys.  Exists does become a little more expensive than inserting however as simply checking for the existence in the map is not enough.  The signature and pubkey must also be compared for equality.  However, I expect this to be rather quick (or at least quicker than generating the map keys with any of the slow methods) since big.Int equality checks are straightforward.

This brings up the question of what to do when two sighashes collide.  I believe that simply overwriting any previous entries will be sufficient.  If/when the original is looked up, exists will now return false and the signature will have to be verified the long way.  Even though this will lower the total number of unique entries that the cache can store, I believe it to be highly likely that total execution time will drop significantly due to the more efficient map operations.

I'd also like to suggest passing an initial size to the sigcache map when it is created to avoid map reallocations.  The max will eventually be reached given long enough runtime so if there is not enough memory to allocate a sigcache that large, the minimum needs to be dropped instead.

jrick:
Cache eviction is also very inefficient and I expect this to be the primary bottleneck once the configured max entries is reached.  If we are unable to increase the efficiency of the eviction (crypto rand and manually iterating the map...) then it should probably evict multiple at a time (say, 1000) so it runs less often.

Roasbeef:
Excellent suggestion @jrick. I've updated the PR with a few commits implementing your suggested optimization. In the process I've added some helper methods for computing equality to `btcec.PublicKey`, and `btcec.Signature`. 

With the change to `sigCacheEntry`, and abandoning the calls to `sha256`, my benchmarks now show a _considerable_ speed up from the `SigCache` checked into master: 
```
BenchmarkSigCacheOldExists-4     2000000               972 ns/op
BenchmarkSigCacheNewExists-4     5000000               264 ns/op
BenchmarkSigCacheOldAdd-4        1000000              1052 ns/op
BenchmarkSigCacheNewAdd-4       10000000               194 ns/op
```

However, these new optimizations don't address the efficiency of the randomized cache eviction in both versions. 
```
BenchmarkSigCacheOldAddEviction-4         300000              6010 ns/op
``` 

The culprit here lies in requiring a read from `/dev/urandom/` or equivalent with each addition once the `SigCache` is full. As suggested by `ceejep` on IRC, we don't require such strong randomness for our particular use case. Therefore, I investigated ditching `crypto/rand` in favor of using `math/rand`. 

Using `math/rand` instead of reading the full 32-byte random value, we instead  generate 4 random `int64`'s, then serialize+concatenate them for usage as a random hash. This also resulted in a significant speed up. However, concurrent calls to `math/rand` are synchronized with a global package level lock. IIRC we use `math/rand` in packages within `btcd` such as `addrmgr`. I suspect concurrent usage would result in unnecessary contention, which is undesirable. 

`ceejep` also suggested using a [xorshit](https://en.wikipedia.org/wiki/Xorshift) PRNG in place of `math/rand`. The xorshift+ PRNG is extremely fast on modern CPU's, has a simple implementation, and passes statistical tests. I ran (single threaded) benchmarks with both approaches, and found `xorshift+` outperforms `math/rand`: 

```
BenchmarkSigCacheOldAddEvict-4    300000              6017 ns/op
BenchmarkSigCacheNewAddEvict-4   2000000               617 ns/op
BenchmarkSigCacheXorAddEvict-4   3000000               466 ns/op
```

The latest benchmark branch can be found [here](https://github.com/Roasbeef/btcd/tree/sigcache-new-bench).

jrick:
This looks great, nice job.  I only have two remaining (minor) complains.  The first is that the sigcache map is still not being allocated with an initial size suitable for the eventual max.  The second is that during eviction the map is still iterated with a range until a suitable key is found, which in the worst case will iterate the entire map.  I know that the range map order isn't really random (it can be quite predictable actually) but for the purposes of this cache, is that really a problem?

Essentially I'm asking whether something like this could be done:

```
for k := range s.validSigs {
    delete(s.validSigs, k)
    break
}
```

Roasbeef:
Ah, forgot about that. I've pushed a commit that fully allocates the map with the maximum number of entries to be stored within the `SigCache`. :+1: 

Great point. In the worst-case, we'll be forced to iterate over *every* entry within the `SigCache` when evicting entries. It's worth noting that my benchmarks above don't capture this worst-case performance, as the eviction cases were executed with only 1 element present in the `SigCache` (although full). You're correct that iteration order for maps is already [randomized](https://github.com/golang/go/blob/master/src/runtime/hashmap.go#L647), so it should be sufficient to just stop, and break after the first element from the `range` statement. 

I've pushed another commit which abandons xorshift, in favor of Go's randomized iteration order for maps. Benchmarking once again, shows an improvement (384 vs 466 ns). 
```
BenchmarkSigCacheOldAddEvict-4    300000              5834 ns/op
BenchmarkSigCacheNewAddEvict-4   2000000               384 ns/op
```

So then, I guess the final question is: are we comfortable relying on the runtime for randomly evicting an element from the `SigCache`? 

I'm comfortable with it myself. I fail to see how an adversary could manipulate the RNG output which determines the starting entry, if the RNG seed has sufficiently high entropy. 

I'm curious as to the others' opinion on this matter.

jrick:
You're missing a break after the delete.  Right now every entry will always be evicted.

There are certainly places where the randomness of random map iteration can't be relied on, but I don't think this is one of them.

davecgh:
I personally don't think relying on the randomness of map iteration order is a good idea.  The spec says:

> The iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next.

The fact that `gc` implements the randomization behavior (which was only added in 1.2 if memory serves) does not mean that `gccgo` or any other compilers such as clang will necessarily implement it either.

jrick:
The randomness only has to do with that you will see different values for the same map each time you iterate it, not that you will iterate keys in order.  That is impossible to do regardless of the implementation (unless you went out of your way to do it slowly) since they've been hashed to a bucket index.

jrick:
I'm also not sure how you could exploit the randomness or not of the map hashing unless you've also broken sha256.. as which point you're not using bitcoin anymore.

davecgh:
What's the status of this one?

jrick:
The xorshift package should probably be vendored but otherwise the current code has an ok from me.

Roasbeef:
I've reverted the PR to once again use `xorshift` rather than depending on the behavior of Go's map iteration.

Roasbeef:
Actually, ignore my previous comment. I've reverted my reversion. I agree with jrick. I'm unable to see how an adversary can influence which elements are evicted from the cache without breaking sha256 in the process. 

As a result, we no longer need to vendor the `xorshift` package, since it's no longer used.

davecgh:
Would you rebase this to the latest master please so it picks up the latest Travis updates and thus will pass the tests?

btcec/pubkey_test.go
```diff
@@ -247,3 +247,39 @@ func TestPubKeys(t *testing.T) {
 		}
 	}
 }
+
+func TestPublicKeyIsEqual(t *testing.T) {
+	pubKey1, err := btcec.ParsePubKey(
+		[]byte{0x03, 0x26, 0x89, 0xc7, 0xc2, 0xda, 0xb1, 0x33,
```
davecgh: These could use `decodeHex` since they are hard coded and it would make the code cleaner I think.

davecgh: Actually, if you'd prefer, I can just do this in a separate PR because there are other cases in the tests that should be using `decodeHex` as well.

txscript/sigcache.go
```diff
@@ -42,8 +43,11 @@ type SigCache struct {
 // exist in the SigCache at any particular moment. Random entries are evicted
 // to make room for new entries that would cause the number of entries in the
 // cache to exceed the max.
-func NewSigCache(maxEntries uint) *SigCache {
-	return &SigCache{validSigs: make(map[sigInfo]struct{}), maxEntries: maxEntries}
+func NewSigCache(maxEntries uint) (*SigCache, error) {
```
davecgh: Why was an error added when the function can't fail?

txscript/sigcache.go
```diff
@@ -79,35 +84,11 @@ func (s *SigCache) Add(sigHash wire.ShaHash, sig *btcec.Signature, pubKey *btcec
...
-			return
-		}
-
-		// Try to find the first entry that is greater than the random
-		// hash. Use the first entry (which is already pseudo random due
-		// to Go's range statement over maps) as a fall back if none of
-		// the hashes in the rejected transactions pool are larger than
-		// the random hash.
-		var foundEntry sigInfo
 		for sigEntry := range s.validSigs {
```
davecgh: Please add a comment that calls out this is relying on the fact that Go iterators maps in a pseudo-random fashion.  People who aren't familiar with that particular nuance will otherwise think this code is just deleting the first entry.

jrick: It would be better to use a comment similar to limitMap's in #658.

Roasbeef: I've updated this fragment with an explanatory comment in the latest commit.

txscript/sigcache.go
Roasbeef: No idea, the error has been removed.

btcec/pubkey_test.go
Roasbeef: I'd like to defer this change to a later PR.

davecgh: Sounds good.

davecgh:
Thanks for the updates.  I'm going to give this one more round of testing, but I believe it's good to go.

davecgh:
OK

davecgh:
Reviewed 3 of 5 files at r2, 1 of 4 files at r4, 2 of 4 files at r8, 1 of 2 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/btcsuite/btcd/598)*

davecgh:
I've updated the PR description to match reality of all of the changes and what the rebased and squash commit message will be.

PR #599, opened 2016-01-12
Use atomic operations instead of mutexes.

dajohi:
[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/599)

### DIFF
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -387,10 +387,14 @@ type HostToNetAddrFunc func(host string, port uint16,
 // of specific types that typically require common special handling are
 // provided as a convenience.
 type Peer struct {
-	started    int32
-	connected  int32
-	disconnect int32 // only to be used atomically
-	conn       net.Conn
+	// The following variables must only be used atomically
+	started       int32
+	connected     int32
+	disconnect    int32
+	bytesReceived uint64
+	bytesSent     uint64
+
+	conn net.Conn
 
 	// These fields are set at creation time and never modified, so they are
 	// safe to read from concurrently without a mutex.
@@ -423,8 +427,6 @@ type Peer struct {
 	timeConnected      time.Time
 	lastSend           time.Time
 	lastRecv           time.Time
-	bytesReceived      uint64
-	bytesSent          uint64
 	startingHeight     int32
 	lastBlock          int32
 	lastAnnouncedBlock *wire.ShaHash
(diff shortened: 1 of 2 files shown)


jrick:
Put the 64-bit atomic fields at the top of the structs so they are guaranteed to be 64-bit aligned.

jrick:
I take that ok back...

For the Peer struct, the stats mutex is still held when copying out every atomic field into the stats copy.  Memory safety isn't violated here but the whole thing is not atomic either, since changes to the atomic fields are not synchronized with the copy.

davecgh:
I'm fine with this change in general, however we have to be really careful to order the struct fields properly and need to make sure to test on 32-bit platforms as well as arm.  This type of change has led to panics on such platfoms in the past due to the alignment requirement.

Also, please comment accordingly that the fields MUST stay at the top of the struct in order to ensure 64-bit alignment and that must only used atomically.

davecgh:
@jrick has a point about accessing all of the stats as an atomic group versus individually, but I don't think it really matters here since the stats are only ever dealt with in the context of a snapshot and are not codependent.

jrick:
Example of how reading of the stats is not being done atomically anymore:

1. GR 1 calls SnapsSnapshot
2. GR 1 is preempted by GR 2 which reads a message and atomically updates the bytesRead field.
3. GR 2 writes a message but before it can atomically write the stats to bytesSent, it is preempted by GR 1
3. GR 1 sees the new bytesRead value and the old bytesSent value

And there are many other orderings that could result in unexpected values.

jrick:
It's still memory safe, but could result in wacky statistics.  Your call @davecgh.

davecgh:
Personally, I prefer to avoid using atomic most of the time due to the alignment requirement and how easy it is for it to be accidentally violated when adding or removing struct fields.  Obviously the tests should ideally catch it, but that assumes the tests run on every single platform which Travis does not support.

Given the fact that bytes are always going to be a snapshot or "best effort" and they aren't codependent with any other stats, I don't believe they could ever really result in wackiness.

That said, I'm not really sure there is a clear benefit here either.  It might be ever so slightly faster in exchange for more fragile code.

dajohi:
I was just getting opinions.

I do realize the snapshot would no longer be exactly in sync, but since it was just byte counters, I didn't think it as mattering negatively.  Also, send and receive wait on each other because of the mutexing and thought the flow would be nicer if they didn't.  It is a lot of Lock/Unlock with each read and write on the wire.

jrick:
I'm leaning towards supporting this change just because it removes the undesired synchronization between reads and writes.  If we want to update the read and write stats with the read and write operations, that would require replacing these atomics with 2 new mutexes.

davecgh:
All requested changes have been made, and as @jrick pointed out, this makes is clear that the read and write stats are independent.

OK

jrick:
ok

PR #608, opened 2016-01-23
peer: Simplify PushAddrMsg method loop.

jongillham:
[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/608)

### DIFF
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2013-2015 The btcsuite developers
+// Copyright (c) 2013-2016 The btcsuite developers
 // Use of this source code is governed by an ISC
 // license that can be found in the LICENSE file.
 
@@ -10,7 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	prand "math/rand"
+	"math/rand"
 	"net"
 	"strconv"
 	"sync"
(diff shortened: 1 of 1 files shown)


jongillham:
Note that I removed `rand.New(prand.NewSource(time.Now().UnixNano()))`. Would it be easier and more efficient to seed the pseudorandom number generator somewhere like btcd's main function. It would potentially allow better deterministic testing of packages such as `peer`.

davecgh:
Thanks for the PR.  I like where you're going here.

I would like for this commit to also seed the prng in an init function for the package.  It can be overridden by an init function in the test package for deterministic testing as you suggest.  This is the same model I used for the treaps in my forthcoming database branch.  See [main package](https://github.com/davecgh/btcd/blob/blockchain_database2/database2/internal/treap/common.go#L127-L129) and [the override for the tests](https://github.com/davecgh/btcd/blob/blockchain_database2/database2/internal/treap/common_test.go#L118-L121).

peer/peer.go
```diff
@@ -811,39 +811,38 @@ func (p *Peer) pushVersionMsg() error {
 }
 
 // PushAddrMsg sends an addr message to the connected peer using the provided
-// addresses.  This function is useful over manually sending the message via
+// addresses. This function is useful over manually sending the message via
```
davecgh: This doesn't really matter, but if you look throughout the project there are pretty much always two spaces after a period.  I'd prefer to keep it that way for consistency.

peer/peer.go
```diff
@@ -811,39 +811,38 @@ func (p *Peer) pushVersionMsg() error {
...
-// are too many.  No message will be sent if there are no entries in the
-// provided addresses slice.
-// It is safe for concurrent access.
-func (p *Peer) PushAddrMsg(addresses []*wire.NetAddress) ([]*wire.NetAddress, error) {
+// are too many. It returns the addresses that were actually sent and no message
+// will be sent if there are no entries in the provided addresses slice.
+//
+// This function is safe for concurrent access.
+func (p *Peer) PushAddrMsg(addresses []*wire.NetAddress) (
+	[]*wire.NetAddress, error) {
```
davecgh: Please leave this on the same line.  Almost everywhere else in the project function definitions don't break across lines unless they are excessively long.  Aside from just general consistency which is probably the most important reason, we've also received requests to do it because splitting it doesn't work well with the intellisense features of some editors or godoc.

jongillham:
@davecgh thanks for the feedback. I've made the changes and amended the commit.

davecgh:
Thanks for the updates!  The changes look good.

I also just noticed the year in the license header needs to be updated to 2016 since the file is being modified.  After that change if you would also please rebase to the latest master, I'll get this merged.

jongillham:
Nice spot. All done.

davecgh:
OK

Issue #120, opened 2014-04-17
Automatic RPC Configuration


davecgh:
When `btcd` starts without a config file present, it should ideally copy the sample configuration into place along with automatically generating a reasonably secure RPC username and password and populating it into the config file.

Further, a configuration file should be created and copied into place for `btcctl` as well, or perhaps `btcctl` should do this when it starts up as well.  In either case, the credentials should be automatically configured.

I suspect this also means `btcctl` will need a way to specify a different username and password for `btcd` and `btcwallet` which are selected based on the `--wallet` flag since they might not be the same.

The end result of this should be the RPC server is running on localhost only with `btcctl` already setup to work without any need for further user configuration.

stevenroose:
Is this fixed by #609?

davecgh:
Yes, it is.  Thanks!

PR #609, opened 2016-01-24
Add automatic RPC configuration.

hectorj:
Implements #120 

When running btcd, if no configuration file is found, the sample-btcd.conf file is copied to the default path, with the rpcuser and rpcpass lines populated with random base64 encoded values.

When running btcctl, if no configuration file is found, it attempts to read the btcd.conf file at its default path, and create a basic btcctl.conf file by copying the rpcuser and rpcpass.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/609)

### DIFF
--- a/cmd/btcctl/config.go
+++ b/cmd/btcctl/config.go
@@ -6,9 +6,11 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/btcsuite/btcd/btcjson"
@@ -215,6 +217,13 @@ func loadConfig() (*config, []string, error) {
 		os.Exit(0)
 	}
 
+	if _, err := os.Stat(preCfg.ConfigFile); os.IsNotExist(err) {
+		err := createDefaultConfigFile(preCfg.ConfigFile)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error creating a default config file: %v\n", err)
+		}
+	}
+
 	// Load additional config from file.
 	parser := flags.NewParser(&cfg, flags.Default)
 	err = flags.NewIniParser(parser).ParseFile(preCfg.ConfigFile)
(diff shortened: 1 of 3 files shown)


davecgh:
Thanks for the PR.  I haven't review this yet, but I noticed it failed the CI checks.  Looking at the Travis logs I see the following reasons:

> config.go:884:2: var generatedRpcUser should be generatedRPCUser
> config.go:890:2: var generatedRpcPass should be generatedRPCPass

hectorj:
I forgot to run goclean.sh, thanks.
Now fixed.

cmd/btcctl/config.go
```diff
@@ -268,3 +277,59 @@ func loadConfig() (*config, []string, error) {
...
+	}
+
+	// Create the destination file and write the rpcuser and rpcpass to it
+	dest, err := os.OpenFile(destinationPath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0700)
+	if err != nil {
+		return err
+	}
+	defer dest.Close()
+
+	dest.WriteString(fmt.Sprintf(`[Application Options]
```
davecgh: The `[Application Options]` header isn't required, so I would prefer it's left off.  Also, please do this using a regular string with \n so it doesn't mess the source code formatting up.

```Go
	dest.WriteString(fmt.Sprintf("rpcuser=%s\nrpcpass=%s",
		string(userSubmatches[1]), string(passSubmatches[1])))
```

hectorj: Done (sorry for the delay)

hectorj:
Rebased on latest master and squashed

davecgh:
I just tested this out and the generated `conf` file contains extra new lines after every line:

```

; Data settings

```

davecgh:
`btcctl` is able to properly extract the `rpcuser` and `rpcpass` and generates the conf file automatically when it doesn't exist as expected.  That is super nice!

hectorj:
New lines issue fixed

config.go
```diff
@@ -880,6 +891,66 @@ func loadConfig() (*config, []string, error) {
...
+	reader := bufio.NewReader(src)
+	for err != io.EOF {
+		var line string
+		line, err = reader.ReadString('\n')
+		if err != nil && err != io.EOF {
+			return err
+		}
+
+		if strings.Contains(line, "rpcuser=") {
+			line = "  rpcuser=" + string(generatedRPCUser) + "\n"
```
davecgh: I don't believe this should have the extra 2 spaces at the front (nor the `rpcpass` below).

It ends up like the following:

```
; full set of credentials - limited or admin - or the RPC server will
; be disabled.
  rpcuser=<user>
  rpcpass=<pass>
; rpclimituser=whatever_limited_username_you_want
; rpclimitpass=
```

Instead of:

```
; full set of credentials - limited or admin - or the RPC server will
; be disabled.
rpcuser=<user>
rpcpass=<pass>
; rpclimituser=whatever_limited_username_you_want
; rpclimitpass=
```

davecgh:
Confirmed.  The only thing left is the extra two spaces before `rpcuser` and `rpcpass` that I commented on.  After that this it good to go.

hectorj:
- removed extra spaces
- removed debug code leftover in test (I was printing the generated file to check the newlines and forgot to remove it)
- refactored test by using ioutil.ReadFile instead of os.Open + ioutil.ReadAll
- squashed it all

davecgh:
Thanks.  I reviewed all the changes and tested it as follows:

- Ran btcd with existing conf file to ensure it didn't overwrite
- Ran btcctl with existing conf file to ensure it didn't overwrite
- Removed both btcd and btcctl conf files
- Ran btcd and confirmed it created the config file and formatted it appropriately
- Ran btcctl and confirmed it created the config file with the rpcuser and rpcpass extracted from the config file

Everything looks good.

OK

waxmihs2902:
Reviewed 1 of 1 files at r3, 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/btcsuite/btcd/609)*

PR #611, opened 2016-01-27
Implement banning based on dynamic ban scores

tb00:
Dynamic ban scores consist of a persistent and a decaying component. The
persistent score can be used to create simple additive banning policies
simlar to those found in other bitcoin node implementations. The
decaying score enables the creation of evasive logic which handles
misbehaving peers (especially application layer DoS attacks) gracefully
by disconnecting and banning peers attempting various kinds of flooding.
Dynamic ban scores allow these two approaches to be used in tandem.

This pull request includes the following:

 - Dynamic ban score type & functions, with tests for core functionality
 - Ban score of connected peers can be queried via rpc (getpeerinfo)
 - Example policy with decaying score increments on mempool and getdata
 - Logging of misbehavior once half of the ban threshold is reached
 - Banning logic can be disabled via configuration (enabled by default)
 - User defined ban threshold can be set via configuration

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/611)

### DIFF
--- a/dynamicbanscore.go
+++ b/dynamicbanscore.go
@@ -1,0 +1,146 @@
+// Copyright (c) 2016 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package main
+
+import (
+	"fmt"
+	"math"
+	"sync"
+	"time"
+)
+
+const (
+	// Halflife defines the time (in seconds) by which the transient part
+	// of the ban score decays to one half of it's original value.
+	Halflife = 60
+
+	// lambda is the decaying constant.
+	lambda = math.Ln2 / Halflife
+
+	// Lifetime defines the maximum age of the transient part of the ban
+	// score to be considered a non-zero score (in seconds).
+	Lifetime = 1800
+
+	// precomputedLen defines the amount of decay factors (one per second) that
+	// should be precomputed at initialization.
+	precomputedLen = 64
+)
+
+// precomputedFactor stores precomputed exponential decay factors for the first
+// 'precomputedLen' seconds starting from t == 0.
+var precomputedFactor [precomputedLen]float64
+
+// init precomputes decay factors.
+func init() {
+	for i := range precomputedFactor {
+		precomputedFactor[i] = math.Exp(-1.0 * float64(i) * lambda)
+	}
+}
+
+// decayFactor returns the decay factor at t seconds, using precalculated values
+// if available, or calculating the factor if needed.
+func decayFactor(t int64) float64 {
+	if t < precomputedLen {
+		return precomputedFactor[t]
+	}
+	return math.Exp(-1.0 * float64(t) * lambda)
+}
+
+// dynamicBanScore provides dynamic ban scores consisting of a persistent and a
+// decaying component. The persistent score could be utilized to create simple
+// additive banning policies similar to those found in other bitcoin node
+// implementations.
+//
+// The decaying score enables the creation of evasive logic which handles
+// misbehaving peers (especially application layer DoS attacks) gracefully
+// by disconnecting and banning peers attempting various kinds of flooding.
+// dynamicBanScore allows these two approaches to be used in tandem.
+//
+// Zero value: Values of type dynamicBanScore are immediately ready for use upon
+// declaration.
+type dynamicBanScore struct {
+	lastUnix   int64
+	transient  float64
+	persistent uint32
+	sync.Mutex
+}
+
+// String returns the ban score as a human-readable string.
+func (s *dynamicBanScore) String() string {
+	s.Lock()
+	r := fmt.Sprintf("persistent %v + transient %v at %v = %v as of now",
+		s.persistent, s.transient, s.lastUnix, s.Int())
+	s.Unlock()
+	return r
+}
+
+// Int returns the current ban score, the sum of the persistent and decaying
+// scores.
+//
+// This function is safe for concurrent access.
+func (s *dynamicBanScore) Int() uint32 {
+	s.Lock()
+	r := s.int(time.Now())
+	s.Unlock()
+	return r
+}
+
+// Increase increases both the persistent and decaying scores by the values
+// passed as parameters. The resulting score is returned.
+//
+// This function is safe for concurrent access.
+func (s *dynamicBanScore) Increase(persistent, transient uint32) uint32 {
+	s.Lock()
+	r := s.increase(persistent, transient, time.Now())
+	s.Unlock()
+	return r
+}
+
+// Reset set both persistent and decaying scores to zero.
+//
+// This function is safe for concurrent access.
+func (s *dynamicBanScore) Reset() {
+	s.Lock()
+	s.persistent = 0
+	s.transient = 0
+	s.lastUnix = 0
+	s.Unlock()
+}
+
+// int returns the ban score, the sum of the persistent and decaying scores at a
+// given point in time.
+//
+// This function is not safe for concurrent access. It is intended to be used
+// internally and during testing.
+func (s *dynamicBanScore) int(t time.Time) uint32 {
+	dt := t.Unix() - s.lastUnix
+	if s.transient < 1 || dt < 0 || Lifetime < dt {
+		return s.persistent
+	}
+	return s.persistent + uint32(s.transient*decayFactor(dt))
+}
+
+// increase increases the persistent, the decaying or both scores by the values
+// passed as parameters. The resulting score is calculated as if the action was
+// carried out at the point time represented by the third paramter. The
+// resulting score is returned.
+//
+// This function is not safe for concurrent access.
+func (s *dynamicBanScore) increase(persistent, transient uint32, t time.Time) uint32 {
+	s.persistent += persistent
+	tu := t.Unix()
+	dt := tu - s.lastUnix
+
+	if transient > 0 {
+		if Lifetime < dt {
+			s.transient = 0
+		} else if s.transient > 1 && dt > 0 {
+			s.transient *= decayFactor(dt)
+		}
+		s.transient += float64(transient)
+		s.lastUnix = tu
+	}
+	return s.persistent + uint32(s.transient)
+}
--- a/dynamicbanscore_test.go
+++ b/dynamicbanscore_test.go
@@ -1,0 +1,68 @@
+// Copyright (c) 2016 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package main
+
+import (
+	"math"
+	"testing"
+	"time"
+)
+
+// TestDynamicBanScoreDecay tests the exponential decay implemented in
+// dynamicBanScore.
+func TestDynamicBanScoreDecay(t *testing.T) {
+	var bs dynamicBanScore
+	base := time.Now()
+
+	r := bs.increase(100, 50, base)
+	if r != 150 {
+		t.Errorf("Unexpected result %d after ban score increase.", r)
+	}
+
+	r = bs.int(base.Add(time.Minute))
+	if r != 125 {
+		t.Errorf("Halflife check failed - %d instead of 125", r)
+	}
+
+	r = bs.int(base.Add(7 * time.Minute))
+	if r != 100 {
+		t.Errorf("Decay after 7m - %d instead of 100", r)
+	}
+}
+
+// TestDynamicBanScoreLifetime tests that dynamicBanScore properly yields zero
+// once the maximum age is reached.
+func TestDynamicBanScoreLifetime(t *testing.T) {
+	var bs dynamicBanScore
+	base := time.Now()
+
+	r := bs.increase(0, math.MaxUint32, base)
+	r = bs.int(base.Add(Lifetime * time.Second))
+	if r != 3 { // 3, not 4 due to precision loss and truncating 3.999...
+		t.Errorf("Pre max age check with MaxUint32 failed - %d", r)
+	}
+	r = bs.int(base.Add((Lifetime + 1) * time.Second))
+	if r != 0 {
+		t.Errorf("Zero after max age check failed - %d instead of 0", r)
+	}
+}
+
+// TestDynamicBanScore tests exported functions of dynamicBanScore. Exponential
+// decay or other time based behavior is tested by other functions.
+func TestDynamicBanScoreReset(t *testing.T) {
+	var bs dynamicBanScore
+	if bs.Int() != 0 {
+		t.Errorf("Initial state is not zero.")
+	}
+	bs.Increase(100, 0)
+	r := bs.Int()
+	if r != 100 {
+		t.Errorf("Unexpected result %d after ban score increase.", r)
+	}
+	bs.Reset()
+	if bs.Int() != 0 {
+		t.Errorf("Failed to reset ban score.")
+	}
+}
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -2382,7 +2382,7 @@ func handleGetPeerInfo(s *rpcServer, cmd interface{}, closeChan <-chan struct{})
 			Inbound:        statsSnap.Inbound,
 			StartingHeight: statsSnap.StartingHeight,
 			CurrentHeight:  statsSnap.LastBlock,
-			BanScore:       0,
+			BanScore:       int32(p.banScore.Int()),
 			SyncNode:       p == syncPeer,
 		}
 		if p.LastPingNonce() != 0 {
--- a/config.go
+++ b/config.go
@@ -33,6 +33,7 @@ const (
 	defaultLogFilename       = "btcd.log"
 	defaultMaxPeers          = 125
 	defaultBanDuration       = time.Hour * 24
+	defaultBanThreshold      = 100
 	defaultMaxRPCClients     = 10
 	defaultMaxRPCWebsockets  = 25
 	defaultVerifyEnabled     = false
@@ -84,7 +85,9 @@ type config struct {
 	DisableListen      bool          `long:"nolisten" description:"Disable listening for incoming connections -- NOTE: Listening is automatically disabled if the --connect or --proxy options are used without also specifying listen interfaces via --listen"`
 	Listeners          []string      `long:"listen" description:"Add an interface/port to listen for connections (default all interfaces port: 8333, testnet: 18333)"`
 	MaxPeers           int           `long:"maxpeers" description:"Max number of inbound and outbound peers"`
+	DisableBanning     bool          `long:"nobanning" description:"Disable banning of misbehaving peers"`
 	BanDuration        time.Duration `long:"banduration" description:"How long to ban misbehaving peers.  Valid time units are {s, m, h}.  Minimum 1 second"`
+	BanThreshold       uint32        `long:"banthreshold" description:"Maximum allowed ban score before disconnecting and banning misbehaving peers."`
 	RPCUser            string        `short:"u" long:"rpcuser" description:"Username for RPC connections"`
 	RPCPass            string        `short:"P" long:"rpcpass" default-mask:"-" description:"Password for RPC connections"`
 	RPCLimitUser       string        `long:"rpclimituser" description:"Username for limited RPC connections"`
@@ -324,6 +327,7 @@ func loadConfig() (*config, []string, error) {
 		DebugLevel:        defaultLogLevel,
 		MaxPeers:          defaultMaxPeers,
 		BanDuration:       defaultBanDuration,
+		BanThreshold:      defaultBanThreshold,
 		RPCMaxClients:     defaultMaxRPCClients,
 		RPCMaxWebsockets:  defaultMaxRPCWebsockets,
 		DataDir:           defaultDataDir,
--- a/doc.go
+++ b/doc.go
@@ -34,6 +34,9 @@ Application Options:
       --listen=             Add an interface/port to listen for connections
                             (default all interfaces port: 8333, testnet: 18333)
       --maxpeers=           Max number of inbound and outbound peers (125)
+      --nobanning           Disable banning of misbehaving peers
+      --banthreshold=       Maximum allowed ban score before disconnecting and
+                            banning misbehaving peers.
       --banduration=        How long to ban misbehaving peers.  Valid time units
                             are {s, m, h}.  Minimum 1 second (24h0m0s)
   -u, --rpcuser=            Username for RPC connections
--- a/sample-btcd.conf
+++ b/sample-btcd.conf
@@ -103,6 +103,12 @@
 ; Maximum number of inbound and outbound peers.
 ; maxpeers=125
 
+; Disable banning of misbehaving peers.
+; nobanning=1
+
+; Maximum allowed ban score before disconnecting and banning misbehaving peers.`
+; banthreshold=100
+
 ; How long to ban misbehaving peers. Valid time units are {s, m, h}.
 ; Minimum 1s.
 ; banduration=24h
--- a/server.go
+++ b/server.go
@@ -217,8 +217,8 @@ type serverPeer struct {
 	requestedBlocks map[wire.ShaHash]struct{}
 	filter          *bloom.Filter
 	knownAddresses  map[string]struct{}
+	banScore        dynamicBanScore
 	quit            chan struct{}
-
 	// The following chans are used to sync blockmanager and server.
 	txProcessed    chan struct{}
 	blockProcessed chan struct{}
@@ -291,6 +291,40 @@ func (sp *serverPeer) pushAddrMsg(addresses []*wire.NetAddress) {
 	sp.addKnownAddresses(known)
 }
 
+// addBanScore increases the persistent and decaying ban score fields by the
+// values passed as parameters. If the resulting score exceeds half of the ban
+// threshold, a warning is logged including the reason provided. Further, if
+// the score is above the ban threshold, the peer will be banned and
+// disconnected.
+func (sp *serverPeer) addBanScore(persistent, transient uint32, reason string) {
+	// No warning is logged and no score is calculated if banning is disabled.
+	if cfg.DisableBanning {
+		return
+	}
+	warnThreshold := cfg.BanThreshold >> 1
+	if transient == 0 && persistent == 0 {
+		// The score is not being increased, but a warning message is still
+		// logged if the score is above the warn threshold.
+		score := sp.banScore.Int()
+		if score > warnThreshold {
+			peerLog.Warnf("Misbehaving peer %s: %s -- ban score is %d, "+
+				"it was not increased this time", sp, reason, score)
+		}
+		return
+	}
+	score := sp.banScore.Increase(persistent, transient)
+	if score > warnThreshold {
+		peerLog.Warnf("Misbehaving peer %s: %s -- ban score increased to %d",
+			sp, reason, score)
+		if score > cfg.BanThreshold {
+			peerLog.Warnf("Misbehaving peer %s -- banning and disconnecting",
+				sp)
+			sp.server.BanPeer(sp)
+			sp.Disconnect()
+		}
+	}
+}
+
 // OnVersion is invoked when a peer receives a version bitcoin message
 // and is used to negotiate the protocol version details as well as kick start
 // the communications.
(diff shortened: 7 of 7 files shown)


davecgh:
Nice.  Thanks for the PR.  I haven't given it a thorough review yet, but on a quick pass it looks well done overall with comments and some tests.

One thing I did notice that the banning is opt-in.  I personally think it should be opt-out (if anything) instead.  As you probably noticed while digging through to implement this, there was already a bit of infrastructure in place to support it, but the dynamic scoring bits was never finished, so I'm really glad to see this.

I'd like to hear from others what they think about having banning by default or not. @dajohi, @jrick, @Roasbeef

dynamicbanscore.go
```diff
@@ -0,0 +1,140 @@
...
+type dynamicBanScore struct {
+	lastUnix   int64
+	transient  float64
+	persistent uint32
+	sync.Mutex
+}
+
+// String returns the ban score as a human-readable string.
+func (s *dynamicBanScore) String() string {
+	return fmt.Sprintf("persistent %v + transient %v at %v = %v as of now",
```
dajohi: missing mutex

dynamicbanscore.go
```diff
@@ -0,0 +1,140 @@
...
+		s.persistent, s.transient, s.lastUnix, s.Int())
+}
+
+// Int returns the current ban score, the sum of the persistent and decaying
+// scores.
+//
+// This function is safe for concurrent access.
+func (s *dynamicBanScore) Int() uint32 {
+	return s.int(time.Now())
```
dajohi: the mutex in int() should be moved here to match other packages.  exported functions handle the mutexes.

dynamicbanscore_test.go
```diff
@@ -0,0 +1,48 @@
...
+package main
+
+import (
+	"math"
+	"testing"
+	"time"
+)
+
+// TestDynamicBanScore tests dynamicBanScore.
+func TestDynamicBanScore(t *testing.T) {
```
dajohi: function uses both exported and unexported functions.

dynamicbanscore.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// Increase increases both the persistent and decaying scores by the values
+// passed as parameters. The resulting score is returned.
+//
+// This function is safe for concurrent access.
+func (s *dynamicBanScore) Increase(persistent, transient uint32) uint32 {
+	s.Lock()
+	defer s.Unlock()
+	return s.increase(persistent, transient, time.Now())
```
dajohi: Try not to use defer when it is easy.

s.Lock()
result := s.increase(persistent, transient, time.Now())
s.Unlock()
return result

dynamicbanscore.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// int returns the ban score, the sum of the persistent and decaying scores at a
+// given point in time.
+//
+// This function is safe for concurrent access. It is intended to be used
+// internally and during testing.
+func (s *dynamicBanScore) int(t time.Time) uint32 {
+	// reduce the amount of time the lock is held
+	s.Lock()
```
dajohi: Move mutex.  move to Int()

rpcserver.go
```diff
@@ -2357,6 +2357,7 @@ func handleGetPeerInfo(s *rpcServer, cmd interface{}, closeChan <-chan struct{})
 	infos := make([]*btcjson.GetPeerInfoResult, 0, len(peers))
 	for _, p := range peers {
 		statsSnap := p.StatsSnapshot()
+		score := p.banScore.Int()
```
dajohi: remove

rpcserver.go
```diff
@@ -2373,7 +2374,7 @@ func handleGetPeerInfo(s *rpcServer, cmd interface{}, closeChan <-chan struct{})
 			Inbound:        statsSnap.Inbound,
 			StartingHeight: statsSnap.StartingHeight,
 			CurrentHeight:  statsSnap.LastBlock,
-			BanScore:       0,
+			BanScore:       score,
```
dajohi: p.banScore.Int(),

dynamicbanscore.go
tb00: @dajohi: My initial version had the mutex in Int(). The reason I moved it to int() it was to release the lock as early as possible (line 109). Having said that, I can of course revert it as per your comment if you prefer.

tb00:
@dajohi: Thanks for the feedback! Please let me know if you have further recommendations.

tb00:
Review status: 0 of 8 files reviewed at latest revision, 7 unresolved discussions.

---

*[dynamicbanscore_test.go, line 14 \[r1\]](https://reviewable.io:443/reviews/btcsuite/btcd/611#-K9-vio1HvOkju33fepd-r1-14:-K98ZKSAh6p4Mm0VeapT:-1248164669) ([raw file](https://github.com/btcsuite/btcd/blob/0dbc19985ab5c30d40ad73f50d65435dc5612785/dynamicbanscore_test.go#L14)):*
Added separate test functions for exported and unexported functions.

---

*[rpcserver.go, line 2377 \[r1\]](https://reviewable.io:443/reviews/btcsuite/btcd/611#-K9-vio1HvOkju33fep_-r1-2377:-K98Z_uFWlGTDWZawot5:-499146009) ([raw file](https://github.com/btcsuite/btcd/blob/0dbc19985ab5c30d40ad73f50d65435dc5612785/rpcserver.go#L2377)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/611)*

Roasbeef:
Excellent PR! This is a great start towards a fine-grained dynamic resource utilization scheduler. Banning after exceeding a threshold is the a first step, which could be taken even further by adding "de-prioritization". As an example: throttling an SPV peer, or a peer doing a from-scratch historical block download. Additionally, once's dave's database PR gets in, one could easily persist the current state of banned peers to the DB. Similar to the decaying score, this state could be pruned periodically (ban for 24 hrs, etc.) 

Concerning defaults: in my opinion, the base banscores should be on by default. Mitigating the increased resource utilization on our users caused by malicious parties on the network should be a priority. 

However, I think we should take care with the thresholds we set for the default banscores. The scoring policy should be tuned in order to ensure it avoids unwanted false positives. For example, with the current ban policy for `getdata`, is it possible a syncing node connected to btcd via a low latency link is erroneously banned? Note that I haven't yet tested this PR.

dynamicbanscore.go
```diff
@@ -0,0 +1,142 @@
...
+	"math"
+	"sync"
+	"time"
+)
+
+const (
+	// Halflife defines the time (in seconds) by which the transient part
+	// of the ban score decays to one half of it's original value.
+	Halflife = 60
+	lambda   = math.Ln2 / Halflife
```
Roasbeef: In order to be consistent with the definitions of variables inside the `const` clause throughout the rest of the codebase, there should be a newline separating each variable initialization.

tb00: Alternatively, I could simply inline it, it is used at two locations. Let me know which one is preferred and I will make the change.

tb00:
@Roasbeef: I could not think of a universal policy which fits nodes with different workload and resource constraints equally well. I experimented with the policy on my particular environment, which is statistically not significant. This is why I refer to them as example policies intended as a first step.

Regarding potential false positives raised by the policy for `getdata` when latency is low: I see the point and believe increasing the ban threshold (which can be configured by the end user) would work.

This of course brings up the question whether or not the current default ban threshold is reasonable for the majority of deployments. My intention with having banning off by default is to decrease the impact of picking a default threshold that does not fit the majority. Having said that, I am open to enabling banning by default, and would kindly ask others to evaluate on their environment what ban threshold fits them. Then, we can iteratively refine both the default threshold and policies based on feedback.

tb00:
All requested changes have been applied. Please let me know any further comment.

davecgh:
Thanks for the updates.  I should be able to give this a through review within the next couple of days.

dynamicbanscore.go
```diff
@@ -0,0 +1,143 @@
...
+	if t < precomputedLen {
+		return precomputedFactor[t]
+	}
+	return math.Exp(-1.0 * float64(t) * lambda)
+}
+
+// dynamicBanScore provides dynamic ban scores consisting of a persistent and a
+// decaying component. The persistent score can be utilized to create simple
+// additive banning policies typically found in other bitcoin node
+// implementations, such as increasing ban score on invalid messages.
```
jrick: I would revise this part of the comment since it's not actually being used for invalid messages, just flooding protection.  If a peer is sending an invalid (but recognized) message, they should be disconnected immediately.

davecgh:
I've completed my review.  Very nicely done overall.  Thank you for sticking with the same style used in the code base, that helps a lot.

---

Reviewed 1 of 8 files at r1, 2 of 3 files at r2, 5 of 5 files at r3.
Review status: all files reviewed at latest revision, 13 unresolved discussions.

---

*[btcjson/chainsvrresults.go, line 180 \[r3\]](https://reviewable.io:443/reviews/btcsuite/btcd/611#-KABM_BeFP4cms5prrL6:-KABM_BeFP4cms5prrL7:-647454566) ([raw file](https://github.com/btcsuite/btcd/blob/f5ded656366989112afa359733a4a90d38bf8a8a/btcjson/chainsvrresults.go#L180)):*
While I agree with you that uint32 is better (I *MUCH* prefer to use unsigned when possible), I believe this has to be an int32 in order to remain compatible with Core which uses a signed int.  Changing this would cause btctl/btcrpcclient to fail to unparse responses from Core which have a negative `banscore`.

---

*[dynamicbanscore.go, line 18 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/611#-K9-vio1HvOkju33fepc-r2-18:-KABP1s4fmdnLdgdgeuz:77532325) ([raw file](https://github.com/btcsuite/btcd/blob/393c52a2f6bd8d678aa093249545426136c7175b/dynamicbanscore.go#L18)):*
He was referring to the following (I also agree):

```Go
const (
	// Halflife defines the time (in seconds) by which the transient part
	// of the ban score decays to one half of it's original value.
	Halflife = 60
			<---------- newline here
	// lambda is the dacying constant.
	lambda = math.Ln2 / Halflife
			<---------- newline here
	// Liftime ....
```

---

*[dynamicbanscore.go, line 124 \[r3\]](https://reviewable.io:443/reviews/btcsuite/btcd/611#-KAC-gsWXPmSYkYjF2bj:-KAC-gsXqUCQUopPLUsr:-1901079618) ([raw file](https://github.com/btcsuite/btcd/blob/3dd3b56fb779b61415cb7e883f19b65b48cd5072/dynamicbanscore.go#L124)):*
carreid -> carried

---

*[dynamicbanscore_test.go, line 52 \[r3\]](https://reviewable.io:443/reviews/btcsuite/btcd/611#-KACYQTWml-XiQ9qKkD8:-KACYQTX8Qaolico5D66:-1367567537) ([raw file](https://github.com/btcsuite/btcd/blob/3dd3b56fb779b61415cb7e883f19b65b48cd5072/dynamicbanscore_test.go#L52)):*
function -> functions

---

*[dynamicbanscore_test.go, line 53 \[r3\]](https://reviewable.io:443/reviews/btcsuite/btcd/611#-KACYV7rSrThumRxnoNq:-KACYV7rSrThumRxnoNr:1061418796) ([raw file](https://github.com/btcsuite/btcd/blob/3dd3b56fb779b61415cb7e883f19b65b48cd5072/dynamicbanscore_test.go#L53)):*
tested by other functions

---

*[dynamicbanscore_test.go, line 61 \[r3\]](https://reviewable.io:443/reviews/btcsuite/btcd/611#-KACYi-rhs-2vtHkJmg8:-KACYi-rhs-2vtHkJmg9:-81386728) ([raw file](https://github.com/btcsuite/btcd/blob/3dd3b56fb779b61415cb7e883f19b65b48cd5072/dynamicbanscore_test.go#L61)):*
Shouldn't this just be `if r != 100 {` ?

---

*[server.go, line 299 \[r3\]](https://reviewable.io:443/reviews/btcsuite/btcd/611#-KAC_-4dedOyCvNRdtla:-KAC_-4e1eAOh6Bm6Tv1:916276196) ([raw file](https://github.com/btcsuite/btcd/blob/3dd3b56fb779b61415cb7e883f19b65b48cd5072/server.go#L299)):*
Please reverse this condition check so the main code is not indented in a conditional.  i.e.

```Go
// No warning is logged and no score is calculated if banning is
// disabled.
if cfg.DisableBanning {
	return
}

warnThreshold := cfg.BanThreshold >> 1
...
```

---

*[server.go, line 309 \[r3\]](https://reviewable.io:443/reviews/btcsuite/btcd/611#-KAC_tmZTfcZwpdOT5a2:-KAC_tmZTfcZwpdOT5a3:1605019768) ([raw file](https://github.com/btcsuite/btcd/blob/3dd3b56fb779b61415cb7e883f19b65b48cd5072/server.go#L309)):*
Please add a return and remove the else then outdent the block.  We typically try to keep the code from getting nested as much as possible.

i.e

```Go
// The score is not being increased, but a warning message is still
// logged if the score is above the warn threshold.
if transient == 0 && persistent == 0 {
	score := sp.banScore.Int()
	if score > warnThreshold {
		...
	}
	return
}

// Increase the ban score and warn if the new score is above the warn
// threshold.
score := sp.banScore.Increase(persistent, transient)

---

*[server.go, line 323 \[r3\]](https://reviewable.io:443/reviews/btcsuite/btcd/611#-KACc6y0-4C9K4TRFGIo:-KACc6y0-4C9K4TRFGIp:-724132454) ([raw file](https://github.com/btcsuite/btcd/blob/3dd3b56fb779b61415cb7e883f19b65b48cd5072/server.go#L323)):*
This would be move up to the original if statement as shown in the other comment.

---

*[server.go, line 397 \[r3\]](https://reviewable.io:443/reviews/btcsuite/btcd/611#-KAC_bUwiTDxsPPajIwZ:-KAC_bUxB4V-QErDldsn:577646539) ([raw file](https://github.com/btcsuite/btcd/blob/3dd3b56fb779b61415cb7e883f19b65b48cd5072/server.go#L397)):*
half of its value  (no apostrophe)

---

*[server.go, line 507 \[r3\]](https://reviewable.io:443/reviews/btcsuite/btcd/611#-KACcZxWhK18NJtgyZZv:-KACcZxX1ZdxALmhsDgg:577646539) ([raw file](https://github.com/btcsuite/btcd/blob/3dd3b56fb779b61415cb7e883f19b65b48cd5072/server.go#L507)):*
half of its value  (no apostrophe)

---

*[server.go, line 508 \[r3\]](https://reviewable.io:443/reviews/btcsuite/btcd/611#-KACtzUy-BK7V62cCrcB:-KACtzUz-lALeFNvfucR:-1553247685) ([raw file](https://github.com/btcsuite/btcd/blob/3dd3b56fb779b61415cb7e883f19b65b48cd5072/server.go#L508)):*
Perhaps I'm reading this wrong, but won't this cause a request with max inventory vectors to be banned immediately?

1 + uint32(50000)*100 / 50000 = 101

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/611#-:-KABO6IiRGaFK0y0wUJf:1705965062)*
<!-- Sent from Reviewable.io -->

File: dynamicbanscore.go (hunk 9 above)
Comment (tb00) [inline dynamicbanscore.go]: Yes, a peer sending an invalid message will be disconnected - there was no intention to imply the contrary. What I meant is that the non-decaying score could be used to implement more traditional banning logic, similar to those found in other implementations.

I will rephrase the comment to avoid misunderstanding.

**Comment by tb00** [comment]
Thanks for the precise review. I addressed all remarks. Please take a look and let me know if further updates are needed.

**Comment by davecgh** [comment]
Thanks for the updates.  I've reviewed them and updated my comments.  There are also 4 comments from the original review that are unaddressed.  I'll add them via the github interface so they stand out.

---

Reviewed 5 of 5 files at r4.

---

*[server.go, line 509 \[r4\]](https://reviewable.io:443/reviews/btcsuite/btcd/611#-KAL9AbTrkWUyS3DbV13:-KAL9AbUVsj7Zj3mYhKc:-977729441) ([raw file](https://github.com/btcsuite/btcd/blob/913d6980dec8a254755b1a996d38dc1027dfef46/server.go#L509)):*
I believe this will still ban on a max inventory request.  1 + uint32(50000)*99 / 50000 = 100

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/611#-:-KAL9eRRR51D-wpgjMOk:-1799169810)*

File: dynamicbanscore.go [hunk 10]
```diff
@@ -0,0 +1,146 @@
...
+	dt := t.Unix() - s.lastUnix
+	if s.transient < 1 || dt < 0 || Lifetime < dt {
+		return s.persistent
+	}
+	return s.persistent + uint32(s.transient*decayFactor(dt))
+}
+
+// increase increases the persistent, the decaying or both scores by the values
+// passed as parameters. The resulting score is calculated as if the action was
+// carried out at the point time represented by the third paramter. The
```
davecgh: carried -> carried

dynamicbanscore_test.go
```diff
@@ -0,0 +1,68 @@
...
+	if r != 3 { // 3, not 4 due to precision loss and truncating 3.999...
+		t.Errorf("Pre max age check with MaxUint32 failed - %d", r)
+	}
+	r = bs.int(base.Add((Lifetime + 1) * time.Second))
+	if r != 0 {
+		t.Errorf("Zero after max age check failed - %d instead of 0", r)
+	}
+}
+
+// TestDynamicBanScore tests exported functions of dynamicBanScore. Exponential
```
davecgh: function -> functions

davecgh:
Nevermind about the unaddressed comments, I guess I missed your updates somehow.

tb00:

---

*[server.go, line 509 \[r4\]](https://reviewable.io:443/reviews/btcsuite/btcd/611#-KAL9AbTrkWUyS3DbV13:-KALBiG-xvYrPbVnX2f9:-1164670106) ([raw file](https://github.com/btcsuite/btcd/blob/913d6980dec8a254755b1a996d38dc1027dfef46/server.go#L509)):*
It will ban if the score exceeds the threshold.
`if score > cfg.BanThreshold {`

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/611)*

davecgh:

---

*[server.go, line 509 \[r4\]](https://reviewable.io:443/reviews/btcsuite/btcd/611#-KAL9AbTrkWUyS3DbV13:-KALEPM3172dmRcBNAdS:631767152) ([raw file](https://github.com/btcsuite/btcd/blob/913d6980dec8a254755b1a996d38dc1027dfef46/server.go#L509)):*
Ah right, I remember seeing that.  So this looks like it has a good progression now:

```
1 + uint32(1)*99 / 50000 = 1
1 + uint32(100)*99 / 50000 = 1
1 + uint32(500)*99 / 50000 = 1
1 + uint32(2000)*99 / 50000 = 4
1 + uint32(10000)*99 / 50000 = 20
1 + uint32(25000)*99 / 50000 = 50
1 + uint32(50000)*99 / 50000 = 100
```

I am a little concerned this might be too aggressive for header requests though since headers first mode requests the entire chain of headers up front.  Given the chain is almost 400,000 and 50,000 headers is only 4MB, I could easily see requesting more than 50k in 30 seconds.

I'd like to do a bit of testing here to get a better feel for the numbers.  I'll do that later today as I'd like to get this PR in as soon as possible.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/611)*

davecgh:
OK, so I confirmed my suspicion about the banscore being too aggressive.  I connected both a separate btcd node and a Core both starting with no blockchain to a local btcd node running this patch and, this is what happens:
```
20:47:38 2016-02-12 [WRN] PEER: Misbehaving peer 127.0.0.1:51215 (inbound): getdata -- ban score increased to 51
20:47:38 2016-02-12 [WRN] PEER: Misbehaving peer 127.0.0.1:51215 (inbound): getdata -- ban score increased to 52
20:47:38 2016-02-12 [WRN] PEER: Misbehaving peer 127.0.0.1:51215 (inbound): getdata -- ban score increased to 53
20:47:39 2016-02-12 [WRN] PEER: Misbehaving peer 127.0.0.1:51215 (inbound): getdata -- ban score increased to 101
20:47:39 2016-02-12 [WRN] PEER: Misbehaving peer 127.0.0.1:51215 (inbound) -- banning and disconnecting
...

**Comment by tb00** [comment]
Thank you for testing it, this is exactly the feedback I am looking for. As I noted earlier, I have tuned the formula while experimenting on my specific environment and the current policies need to be adjusted in order to avoid false positives as much as possible while still providing reasonable protection for the majority of users and typical use cases. I am looking for suggestions.

I believe there is no universal policy which fits nodes with different usage patterns, workload characteristics and resource constraints equally well, my objectives are the following:
- implement a flexible framework which offers both decaying and non decaying scoring
- provide policies which are reasonable for typical use with default configuration
- allow the end user to adjust the sensitivity by overriding one single configuration parameter -- as opposed to having to deal with many parameters governing how different kinds of messages are weighted and scored
- allow the end user to completely disable banning via configuration

Based on your feedback it is clear that banning of `getdata` needs to be adjusted. Let me recap to make sure I am not misinterpreting anything (please correct as needed):
- btcd or Core performing initial blockchain download
- peers connected to localhost – close to zero network latency and infinite bandwidth
- as I see from your comment, the score was incremented by 1 upon each `getdata` message
- a large amount of such messages arrived within a second, the peer got banned after a few seconds

The current formula takes the form `C+M*len(InvList)/MaxInvLen` with `C = 1` and `M = 99`. This implies each `getdata` message in your test had an InvList smaller than 506, and many small  as opposed to fewer large messages were sent within a given time window. It means, that the factor M in the formula does not impact this use case at all, but the constant C does. It also means that the exponential decay had no chance to kick in as things happened within two seconds. Setting C to zero would eliminate the issue but at the same time would not allow to detect flooding with small `getdata` messages at all.

Could you please replicate the use case with a ban threshold of 1000 instead of the default 100? Or alternatively, with the diff below (so you would still receive warnings whenever the score is above 50)? I would like to see where the score peaks with the current formula in your case during IBD. Based on this, we could either adjust the formula or consider a reasonably higher default ban threshold.

```
index 838b464..2d17b07 100644
--- a/server.go
+++ b/server.go
@@ -315,7 +315,7 @@ func (sp *serverPeer) addBanScore(persistent, transient uint32, reason string) {
        if score > warnThreshold {
                peerLog.Warnf("Misbehaving peer %s: %s -- ban score increased to %d",
                        sp, reason, score)
-               if score > cfg.BanThreshold {
+               if score > cfg.BanThreshold + 1000000 {
                        peerLog.Warnf("Misbehaving peer %s -- banning and disconnecting",
                                sp)
                        sp.server.BanPeer(sp)
```

I have found that remote peers connected over the internet to my environment typically peak at around 80, but this is a different story as latency and bandwidth limitations apply and I also have bitcoin specific traffic shaping and other network security measures in place which might have impacted my the results.

**Comment by davecgh** [comment]
Thanks for a thorough reply.  I definitely appreciate your attention to detail!  I completely agree with you that there isn't a universal policy the fits all use cases and your stated objectives seem very reasonable to me.  I have tested this PR pretty extensively now and I think it does a good job of meeting your objectives with the exception of the getdata issue that we're working on now.

In regards to the conditions, all of the ones you recapped are correct however there is one major one that I failed to mention and I think it's very important here -- I tested this again PR #491.

The reason this matters so much is because the current code on master is _abysmally_ slow at serving headers because it stores the blocks in leveldb and, as a result, there is no way to only grab the header.  Instead is has to load the entire block from the database and slice off the header.  The end result is it takes ages (we're talking minutes here) to serve the full chain of headers on cold cache.

On the other hand, PR #491 makes use of the new database backend (along with other massive optimizations) that I wrote and merged via #380 which stores the blocks on the file system and therefore headers can be served extremely fast (full chain in around ~1.3 seconds).

```
$ ./dbtool loadheaders
...
20:02:17 2016-02-13 [INF] Loading headers for 398322 blocks...
20:02:18 2016-02-13 [INF] Loaded 398322 headers in 1.277132s
```

I'll test with your modifications here shortly and follow up with the results.  Also, I'll turn on debug logging so we can get a good picture of the number of requested items per `getdata` request coming from Core (which is currently the vast majority of the network by far anyways).

**Comment by davecgh** [comment]
Alright, here are the results.  I modified your suggested diff a bit in order to report a high watermark:

```diff
diff --git a/server.go b/server.go
index 6b02d94..d07bbf9 100644
--- a/server.go
+++ b/server.go
@@ -218,7 +218,9 @@ type serverPeer struct {
        filter          *bloom.Filter
        knownAddresses  map[string]struct{}
        banScore        dynamicBanScore
+       highWatermark   uint32
        quit            chan struct{}
@@ -321,9 +323,14 @@ func (sp *serverPeer) addBanScore(persistent, transient uint32, reason string) {
        }
        score := sp.banScore.Increase(persistent, transient)
        if score > warnThreshold {
-               peerLog.Warnf("Misbehaving peer %s: %s -- ban score increased to %d",
-                       sp, reason, score)
-               if score > cfg.BanThreshold {
+               if score > sp.highWatermark {
+                       sp.highWatermark = score
+                       peerLog.Warnf("Misbehaving peer %s: %s -- new high "+
+                               "watermark %d", sp, reason, score)
+               }
+               if score > cfg.BanThreshold+1000000 {
                        peerLog.Warnf("Misbehaving peer %s -- banning and disconnecting",
                                sp)
                        sp.server.BanPeer(sp)
```


The result looks like it hit a high watermark of 20652:

```
21:29:54 2016-02-13 [WRN] PEER: Misbehaving peer 127.0.0.1:63113 (inbound): getdata -- new high watermark 20652
```

Also, I confirmed via debug logging that Core is only requesting a single block at a time as opposed to multiple per `getdata` request.

tb00:
I am a bit puzzled about this. The high watermark you captured seems extremely high. In order to reach it with `getdata` messages smaller than 500 and the current formula, a peer would have to send 240 messages each second, and sustain this rate for about 7 minutes. Maintaining the same rate, after about 15 minutes the score would stabilize at 20895 and not change any more unless the message rate changes.

To me it seems that a sustained rate of 240 `getdata` messages per second from a remote peer qualifies as misbehavior in general, but of course I agree that low latency peers performing IDB should not be banned right away after a few seconds.

At this point, I see two options. I can set constant C to zero (remove it from the formula) which will ensure banning does not interfere with the use case (zero latency and IDB) but at the same time it will disable flood detection for `getdata` (but still penalize huge `getdata` messages), or we could agree on a maximum message rate that is to be considered healthy for typical use cases and tune the formula and default threshold accordingly. The default should be established such that it does not harm the p2p network (not ban remote peers performing IDB). I am inclined to consider a zero latency / localhost IDB a special case where both peers are run by the same entity. I would not necessarily tune the defaults to cover this case – keeping in mind defaults can be adjusted via configuration. 

A third option would be to narrow down somehow what kind of `getdata` we should penalize and what we should not. In addition to `InvList` size, we could involve `InvType` as you pointed out, but I am unsure if applying different weighting for Tx, Block and FilteredBlock would allow us to distinguish between good and bad.

How shall we proceed?

davecgh:
I think what might make the most sense is to remove that particular part of this PR, since it is the only problematic part, so we can get it merged.

Then we can open another PR specifically to address adding the `getdata` protections.

At any rate, in regards to your calculations, they look to match up pretty closely to what my log is showing.  I enabled debug logging for the PEER subsystem so each `getdata` request is logged and grepped them out per minute for the first three minutes:

```
$ cat btcd.log | grep "Received getdata" | grep "21:22" | wc -l
13951
$ cat btcd.log | grep "Received getdata" | grep "21:23" | wc -l
14520
$ cat btcd.log | grep "Received getdata" | grep "21:24" | wc -l
14280
```

I should also note, the only actual problematic part of IBD (at the current time anyways, future enhancements could definitely change this as network speeds and disk speeds increase of course) is near the beginning of the chain.  This makes sense to me because almost all of the initial blocks consist of very few transactions.  The number of transactions really didn't start going up to appreciable numbers until around block 185000 (June 2012).  I would guess that the average block size of those first 185,000 blocks is easily under 50k.  At a rate of 240 blocks per second (14,400 blocks per minute) up to that point, that amounts to almost 13 minutes.

I should also note that these figures coincides quite well with data I have gathered on IBD across the network.  I have done several test runs and I invariably reach block 182,000 within roughly 14-15 minutes.  That would indicate it's slightly slower across the network (as expected), but if we take the value of 15 minutes, that still implies a sustained rate of 182,000 / (15*60) ~= 202 blocks/sec.

tb00:
Yes, this makes perfect sense. I have removed the problematic part (constant C) so `getdata` messages with an `InvList` smaller than 506 are not penalized at all. Huge inventory queries are still penalized as discussed earlier. Flooding with small messages is not addressed but that is challenge for another day.

davecgh:
Thanks for the update.  Changes look good.  Please squash everything to a single commit, update the commit message and PR description to reflect reality, and rebase on top of the latest mater (commit 907152cef9714be35f6e32352268a23dc76bdb9c).  In case you aren't familiar with rebasing, see the following [gist](https://gist.github.com/davecgh/04cb601a147f8c831ca5#rebasing-one-of-your-existing-pull-requests).

I'll get it merged after that.

tb00:
Done.

davecgh:
Reviewed 1 of 1 files at r5, 2 of 2 files at r6.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/611)*

davecgh:
OK

PR #612, opened 2016-01-29
WIP: peer: Refactor to decrease complexity.

jongillham:
I have been refactoring the `peer` package for my own use outside of `btcd`. I wanted more flexibility on when I disconnect a peer, returned errors from `QueueMessage` and most importantly I wanted to simplify the code base to improve maintainability for myself.

At this point in my refactoring I have two options; to continue to make changes that suit me but diverge further away from `btcd` or to integrate it back into `btcd` if you think the new API and refactored 'engine' are heading in a better direction than the package as it stands now. I know that the Bitcoin peer protocol has a few peculiarities that might make this type of change too risky for your appetites which is why I am showing the code now.

The main changes are as follows:

- `NewInboundPeer` and `NewOutboundPeer` now start the peer and only return once the protocol has been negotiated. There is no longer `Start` or `Connect`. This significantly simplifies the codebase with a lot less synchronisation state needing to be saved on the `Peer` struct. It also allows consolidation of all the code specific to inbound and outbound peers into one function each instead of being dotted around the place.
- `QueueMessage`, `QueueInventory` and related push messages now all return a `chan error` which can be used to either act on an error and/or block until the message is sent to the network. This is in place of the previous `wait` parameter to indicate that a method should block.
- The connection stall handler pushes instead of polls. (I removed the `offset` code that accounts for slow local `MessageListeners` when detecting a peer stall as I don't need it. However it can be added back if necessary.)

I'm happy to do the integration work if you think these changes will benefit `btcd`. However you might feel them too sweeping to be worth the risk.

The package itself compiles and the tests run but `btcd` won't compile with it.

Significant changes are:
```go
Peer.writeHandler()
Peer.writeMsgQueueHandler()
Peer.writeInvVectQueueHandler()
Peer.readHandler()
Peer.pingTicker()
Peer.negotiateOutboundVersion()
Peer.negotiateInboundVersion()
```

There are a couple of things left to do:
- Decide whether the onus should be on the `Peer` to disconnect/shutdown when errors are detected or on the client using `Peer`. (I think it should probably be `Peer`, as it is now, and the client should be notified through a listener or just `Peer.WaitForShutdown()`.) 
- Update documentation.
- Better unit tests. (The code passes the current unit tests but I think they can be improved as well as coverage increased.)

 What are your thoughts on the code and whether this is a worthwhile change to integrate back into `btcd`?

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/612)

### DIFF
--- a/peer/example_test.go
+++ b/peer/example_test.go
@@ -8,11 +8,9 @@ import (
 	"fmt"
 	"log"
 	"net"
-	"time"
 
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/peer"
-	"github.com/btcsuite/btcd/wire"
 )
 
 // mockRemotePeer creates a basic inbound peer listening on the simnet port for
@@ -62,23 +60,12 @@ func Example_newOutboundPeer() {
 	}
 
 	// Create an outbound peer that is configured to act as a simnet node
-	// that offers no services and has listeners for the version and verack
-	// messages.  The verack listener is used here to signal the code below
-	// when the handshake has been finished by signalling a channel.
-	verack := make(chan struct{}, 1)
+	// that offers no services.
 	peerCfg := &peer.Config{
 		UserAgentName:    "peer",  // User agent name to advertise.
 		UserAgentVersion: "1.0.0", // User agent version to advertise.
 		ChainParams:      &chaincfg.SimNetParams,
 		Services:         0,
-		Listeners: peer.MessageListeners{
-			OnVersion: func(p *peer.Peer, msg *wire.MsgVersion) {
-				fmt.Println("outbound: received version")
-			},
-			OnVerAck: func(p *peer.Peer, msg *wire.MsgVerAck) {
-				verack <- struct{}{}
-			},
-		},
 	}
 
 	// Establish the connection to the peer address and mark it connected.
@@ -93,16 +80,6 @@ func Example_newOutboundPeer() {
 		return
 	}
 
-	// Wait for the verack message or timeout in case of failure.
-	select {
-	case <-verack:
-	case <-time.After(time.Second * 1):
-		fmt.Printf("Example_peerConnection: verack timeout")
-	}
-
 	// Disconnect the peer.
 	p.Disconnect()
-
-	// Output:
-	// outbound: received version
 }
--- a/peer/example_test.go
+++ b/peer/example_test.go
@@ -6,6 +6,7 @@ package peer_test
 
 import (
 	"fmt"
+	"log"
 	"net"
 	"time"
 
@@ -38,10 +39,8 @@ func mockRemotePeer() error {
 		}
 
 		// Create and start the inbound peer.
-		p := peer.NewInboundPeer(peerCfg, conn)
-		if err := p.Start(); err != nil {
-			fmt.Printf("Start: error %v\n", err)
-			return
+		if _, err := peer.NewInboundPeer(peerCfg, conn); err != nil {
+			log.Fatal(err)
 		}
 	}()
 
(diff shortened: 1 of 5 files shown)


davecgh:
Thanks John!  Obviously this is pretty large so will take a while to get through it.  My initial impressions of your description of the major changes are as follows:

* I think getting rid of `Start` and `Connect` is a good idea in general.  When I was first reviewing the package I noted that it would be ideal to not have it.  `Start` is a pretty clear cut case.  There really isn't a reason not to combine that one.  `Connect` , on the other hand, I think might be a little trickier to deal with due to the automatic retry logic needed for persistent peers.  I haven't looked at your code yet, so it's possible you've already dealt with it.
* Again, I haven't looked at the code, so this is purely gut feeling and perhaps after review I'll have a change of opinon, but I don't think I care for making the error a channel that you have to block on versus a wait flag.
* Using pushes instead of polling for the connection stall handler is a good idea.  It is important to have the `offset` code though because often a callback is doing a time consuming task like validating a block.  Without accounting for the time spent in callbacks, this peers would easily be disconnected when they aren't doing anything wrong simply because it took a while to validate a block (or several).

jongillham:
Thanks @davecgh, I'll integrate into `btcd` and either ask for a further review or if I find it doesn't suit our mutual needs and/or code becomes too complex then I'll cancel the pull request.

jongillham:
@davecgh I thought I would change focus for this pull request and attempt to make the `peer` package simpler instead adapting it for my needs.

I found that by making `peer.NewOutboundPeer` and `peer.NewInboundPeer` only return after they have successfully negotiated the protocol simplifies the code greatly. It allowed me to remove most of the locking and synchronisation from the package. It also allowed me to better separate the concerns of each goroutine and make the deadline push based instead of poll. (I also ended up not liking the error channel idea so reverted the API back to how it originally was.)

The effect of having `peer.NewOutboundPeer` and `peer.NewInboundPeer` negotiate the protocol is that they are now blocking functions and can block as long as `negotiateTimeout`. This may cause an issue with the code in`btcd/server.go` but I need to do further testing.

I don't think this code will do much to change performance/throughput but you might feel the new architecture better for simplicity and maintainability. (You might not as of course – in which case I'll scrub the request.)

I have a couple of questions:

1. The deadline code in [`peer/peer.go`](https://github.com/btcsuite/btcd/blob/1944637333bce1d7adbd74c8c693cf0f8d286454/peer/peer.go#L1255) seems to reset the deadline every time we make a new request for data. I feel that the deadline should only be cancelled when we actually get data. Is this supposed to happen or is it a bug?
2. The original implementation of [`NewOutboundPeer`](https://github.com/btcsuite/btcd/blob/1944637333bce1d7adbd74c8c693cf0f8d286454/peer/peer.go#L2075) takes an `addr` however you then have to pass a `conn net.Conn` to [`Connect()`](https://github.com/btcsuite/btcd/blob/1944637333bce1d7adbd74c8c693cf0f8d286454/peer/peer.go#L1950) where `addr == conn.RemoteAddr().String()` doesn't have to hold true. So the peer thinks it's connecting to `addr` but it's actually connecting to `conn.RemoteAddr()`. I suspect this was done for a reason but can't find what that reason is. Do you know? (If it was an oversight then the code can be simplified even more.)

The pull request will need further testing and documentation before I am happy with it but overall the architecture is complete. Do you think it is worth carrying on or do you feel the risk of regression issues not warrant the code changes?

jongillham:
I'm abandoning this pull request. I think the changes, although beneficial, are too far reaching to be done in one go safely. I'll submit some more manageable requests instead over time.

davecgh:
I understand.  I do hope you'll continue to work on it though as I like where you're going with it.

PR #617, opened 2016-02-03
peer: Consolidate Connect, Disconnect, Start, Shutdown public methods.

jongillham:
This commit does not change functionality. It makes the creation of inbound and outbound peers more homogeneous. As a result the Start method of peer was removed as it was found not to be necessary. This is the first of several pull requests/commits designed to make the peer public API and internals less complex.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/617)

### DIFF
--- a/server.go
+++ b/server.go
@@ -1038,9 +1038,8 @@ func (s *server) handleAddPeerMsg(state *peerState, sp *serverPeer) bool {
 
 	// Ignore new peers if we're shutting down.
 	if atomic.LoadInt32(&s.shutdown) != 0 {
-		srvrLog.Infof("New peer %s ignored - server is shutting "+
-			"down", sp)
-		sp.Shutdown()
+		srvrLog.Infof("New peer %s ignored - server is shutting down", sp)
+		sp.Disconnect()
 		return false
 	}
 
@@ -1048,14 +1047,14 @@ func (s *server) handleAddPeerMsg(state *peerState, sp *serverPeer) bool {
 	host, _, err := net.SplitHostPort(sp.Addr())
 	if err != nil {
 		srvrLog.Debugf("can't split hostport %v", err)
-		sp.Shutdown()
+		sp.Disconnect()
 		return false
 	}
 	if banEnd, ok := state.banned[host]; ok {
 		if time.Now().Before(banEnd) {
-			srvrLog.Debugf("Peer %s is banned for another %v - "+
-				"disconnecting", host, banEnd.Sub(time.Now()))
-			sp.Shutdown()
+			srvrLog.Debugf("Peer %s is banned for another %v - disconnecting",
+				host, banEnd.Sub(time.Now()))
+			sp.Disconnect()
 			return false
 		}
 
@@ -1070,16 +1069,16 @@ func (s *server) handleAddPeerMsg(state *peerState, sp *serverPeer) bool {
 		if state.OutboundCount() >= state.maxOutboundPeers {
 			srvrLog.Infof("Max outbound peers reached [%d] - disconnecting "+
 				"peer %s", state.maxOutboundPeers, sp)
-			sp.Shutdown()
+			sp.Disconnect()
 			return false
 		}
 	}
 
 	// Limit max number of total peers.
 	if state.Count() >= cfg.MaxPeers {
-		srvrLog.Infof("Max peers reached [%d] - disconnecting "+
-			"peer %s", cfg.MaxPeers, sp)
-		sp.Shutdown()
+		srvrLog.Infof("Max peers reached [%d] - disconnecting peer %s",
+			cfg.MaxPeers, sp)
+		sp.Disconnect()
 		// TODO(oga) how to handle permanent peers here?
 		// they should be rescheduled.
 		return false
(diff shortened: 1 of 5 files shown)


server.go
```diff
@@ -1638,11 +1642,11 @@ out:
 		case qmsg := <-s.query:
 			s.handleQuery(state, qmsg)
 
-		// Shutdown the peer handler.
+		// Disconnect the peer handler.
 		case <-s.quit:
-			// Shutdown peers.
+			// Disconnect peers.
 			state.forAllPeers(func(sp *serverPeer) {
-				sp.Shutdown()
```
davecgh: Since the `Shutdown` function was removed an it has the trace logging for the peer being forcefully shutdown, I'd suggest adding the removed logging statement back to here before calling `Disconnect`.

server.go
```diff
@@ -1659,7 +1663,7 @@ out:
 		if !state.NeedMoreOutbound() || len(cfg.ConnectPeers) > 0 ||
 			atomic.LoadInt32(&s.shutdown) != 0 {
 			state.forPendingPeers(func(sp *serverPeer) {
-				sp.Shutdown()
```
davecgh: Same here regarding the trace log.

davecgh:
Reviewed 5 of 5 files at r1.

---

*[peer/peer.go, line 1954 \[r1\]](https://reviewable.io:443/reviews/btcsuite/btcd/617#-K9cR4vRShajlvQljIUF:-K9cR4vSNiHqoSsGdbgI:-1260650573) ([raw file](https://github.com/btcsuite/btcd/blob/ed2ca9d56a2f203cc488a6f202a4ef9f42f54029/peer/peer.go#L1954)):*
We'll need to make sure this works correctly with proxies.  I seem to recall the reason an address string was used is because the proxy package ends up giving you the address of the proxy here as opposed to the peer.

---

*[server.go, line 1645 \[r1\]](https://reviewable.io:443/reviews/btcsuite/btcd/617#-K9cVlWR8ijCw1Ak-ioE:-K9cVlWSm9PTOsIIDaSL:1041054191) ([raw file](https://github.com/btcsuite/btcd/blob/ed2ca9d56a2f203cc488a6f202a4ef9f42f54029/server.go#L1645)):*
This comment doesn't seem right (nor did it before).

How about the following?

```Go
// Disconnect all peers on server shutdown.
```

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/617)*

davecgh:
As discussed on your previous PR, I like the direction you're heading with this.

jongillham:
I didn't add the shutdown trace logging because the `Disconnect` function already logs that it has been called and I felt the current functionality was just doubling up logging without any benefit. Happy to add it back though.

---

---

*[peer/peer.go, line 1954 \[r1\]](https://reviewable.io:443/reviews/btcsuite/btcd/617#-K9cR4vRShajlvQljIUF:-K9gx3jqk3ALTZHqsypV:2001018710) ([raw file](https://github.com/btcsuite/btcd/blob/ed2ca9d56a2f203cc488a6f202a4ef9f42f54029/peer/peer.go#L1954)):*
That's the same functionality that exists now. I just moved it out of the `NewInboundPeer` constructor.

Ideally I don't want users of the `peer` package to be able to set address strings of the remote peer at all for outbound peers. I would like to change  the`NewOutboundPeer(cfg Config, addr string)` signature to `NewOutboundPeer(cfg Config)` just like `NewInboundPeer(cfg Config)` and get the `p.addr` from `conn.RemoteAddr().String()`. As it stands now `p.addr == conn.RemoteAddr().String()` doesn't necessarily hold true for outbound peers which is a bit confusing.

However, as you say, that might affect the proxy code and leak information it's not supposed to so I kept the functionality the same for the time being. I hope to look into simplifying this in a later commit.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/617#-:-K9h4cnpDQPWG5xU9Kq_:-684888602)*

davecgh:
Reviewed 1 of 5 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/617)*

davecgh:
Alright, I'm done with my review of the code.  I'm going to do a bit of manual testing and will report back.

davecgh:
I've finished testing this and did not find any new issues.

@jongillham Please rebase and squash this (there is a minor merge conflict with a recent peer change involving the started flag). Once that is done, I'll get it merged in.

jongillham:
Thanks @davecgh. The next step will be to make the protocol version negotiation half-duplex which simplifies the code quite a bit and makes it easier to simplify the ping and stall handler logic.

davecgh:
Sounds good. Thanks again for your work on this.  It'll help simply the SPV wallet work as well.

---

Reviewed 2 of 2 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/617#-:-K9ro7ubToufaVpjlroo:1388996200)*

davecgh:
OK

PR #618, opened 2016-02-06
peer: Implement sendheaders support (BIP0130).

dajohi:
This modifies the `peer` package to add support for the `sendheaders` protocol message introduced by `BIP0030`.

NOTE: This does not add support to btcd itself. That requires the server and sync code to make use of the new functionality exposed by these changes.  As a result, btcd will still be using protocol version 70011.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/618)

### DIFF
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -26,7 +26,7 @@ import (
 
 const (
 	// MaxProtocolVersion is the max protocol version the peer supports.
-	MaxProtocolVersion = 70011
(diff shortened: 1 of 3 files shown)


davecgh:
I've reviewed this and tested that these additions do not yet change the behavior of btcd as expected since that functionality will need to be part of a separate PR.

My only nitpick is that the commit message doesn't make the above clear.  However, I'll modify the commit message and PR description accordingly when merging.

OK

PR #619, opened 2016-02-08
Do not use unkeyed fields.

jcvernaleo:
Fix something goclean caught.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/619)


davecgh:
Thanks for the PR.  I need to see why Travis didn't catch this.  It should have in the goclean script.

OK

PR #621, opened 2016-02-10
Faster operations on the elliptical curve exponent by using nfield instead of big.Int

jimmysong:


davecgh:
Very nice!  Thanks a ton for working on this.  It's obviously going to take a lot of careful review and there are some Travis issues to take care of, but it will be nice to get this in not only for the speed up, but also because we can stop relying on the math/big package which has been rather buggy.

jongillham:
Review status: 0 of 8 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[btcec/nfield.go, line 156 \[r1\]](https://reviewable.io:443/reviews/btcsuite/btcd/621#-KAAUTh5N8fnEMmQonf5:-KAAUTh5N8fnEMmQonf6:725480431) ([raw file](https://github.com/btcsuite/btcd/blob/ea75bce77c4543b7c5eb1d5c0f7b2eb84f46b56c/btcec/nfield.go#L156)):*
Can the `nfieldVal` type be declared as `type nfieldValue [10]uint32` instead seeing as it only has one member? Not sure if it's possible but it might make the struct even more memory compact. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/621)*

jimmysong:
@davecgh any update on this?

dajohi:
rebase please

raedah:
I am wonder if it would useful for the big.Ints mode to be available to run in tandem with the field elements mode for testing and verification of results.

btcec/nfield.go
```diff
@@ -0,0 +1,1769 @@
...
+// license that can be found in the LICENSE file.
+
+package btcec
+
+// References:
+//   [HAC]: Handbook of Applied Cryptography Menezes, van Oorschot, Vanstone.
+//     http://cacr.uwaterloo.ca/hac/
+
+// All elliptic curve operations for secp256k1 are done in a finite field
+// characterized by a 256-bit prime.  Given this precision is larger than the
```
davecgh: These comments need to be updated to reflect reality.  This is implementing modular arithmetic on the order of the curve as opposed to the prime.

btcec/nfield.go
```diff
@@ -0,0 +1,1769 @@
...
+
+	// nfieldMSBBits is the number of bits in the most significant word used
+	// to represent the value.
+	nfieldMSBBits = 256 - (nfieldBase * (nfieldWords - 1))
+
+	// nfieldMSBMask is the mask for the bits in the most significant word
+	// needed to represent the value.
+	nfieldMSBMask = (1 << nfieldMSBBits) - 1
+
+	// nfieldPrimeWordZero is word zero of the secp256k1 prime in the
```
davecgh: Same here.  secp256k1 curve order, not prime.

davecgh: There are several more cases of this, but I'll not comment on every one to avoid sending a ton of e-mails.

btcec/nfield.go
```diff
@@ -0,0 +1,1769 @@
...
+	// internal nfield representation.  It is used during modular reduction
+	// and negation.
+	nfieldPrimeWordThree = 0x39abd22
+
+	// nfieldPrimeWordFour is word four of the secp256k1 prime in the
+	// internal nfield representation.  It is used during modular reduction
+	// and negation.
+	nfieldPrimeWordFour = 0x2baaedc
+
+	nfieldComplementZero  = 0x3c9bebf
```
davecgh: This all needs comments.

btcec/nfield.go
```diff
@@ -0,0 +1,1769 @@
...
+	nfieldReduceZero      = nfieldComplementZero * 16
+	nfieldReduceOne       = nfieldComplementOne * 16
+	nfieldReduceTwo       = nfieldComplementTwo * 16
+	nfieldReduceThree     = nfieldComplementThree * 16
+	nfieldReduceFour      = nfieldComplementFour * 16
+)
+
+// nfieldVal implements optimized fixed-precision arithmetic over the
+// secp256k1 finite nfield.  This means all arithmetic is performed modulo
+// 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2f.  It
```
davecgh: The is the prime, not N.

btcec/nfield.go
```diff
@@ -0,0 +1,1769 @@
...
+	nfieldComplementFour  = 0x1455123
+	nfieldReduceZero      = nfieldComplementZero * 16
+	nfieldReduceOne       = nfieldComplementOne * 16
+	nfieldReduceTwo       = nfieldComplementTwo * 16
+	nfieldReduceThree     = nfieldComplementThree * 16
+	nfieldReduceFour      = nfieldComplementFour * 16
+)
+
+// nfieldVal implements optimized fixed-precision arithmetic over the
+// secp256k1 finite nfield.  This means all arithmetic is performed modulo
```
davecgh: finite nfield -> finite curve order.

btcec/nfield.go
```diff
@@ -0,0 +1,1769 @@
...
+func (f *nfieldVal) SetHex(hexString string) *nfieldVal {
+	if len(hexString)%2 != 0 {
+		hexString = "0" + hexString
+	}
+	bytes, _ := hex.DecodeString(hexString)
+	return f.SetByteSlice(bytes)
+}
+
+// Normalize normalizes the internal nfield words into the desired range and
+// performs fast modular reduction over the secp256k1 prime by making use of the
```
davecgh: secp256k1 prime -> secp256k1 curve order

btcec/nfield.go
```diff
@@ -0,0 +1,1769 @@
...
+// special form of the prime.
```
davecgh: prime -> curve order.

btcec/nfield.go
```diff
@@ -0,0 +1,1769 @@
...
+	r = (r >> nfieldBase) + t9
+	t9 = r & nfieldMSBMask
+
+	// At this point, one more subtraction of the prime
+	// might be needed if the current result is greater than or equal to the
+	// n.
+
+	subtractN := false
+
+	if t9 == nfieldMSBMask && t8 == nfieldBaseMask && t7 == nfieldBaseMask && t6 == nfieldBaseMask && t5 == nfieldBaseMask {
```
davecgh: This is not constant time.

btcec/nfield.go
```diff
@@ -0,0 +1,1769 @@
...
+// NegateVal negates the passed value and stores the result in f.
+// The result is guaranteed to be normalized
+//
+// The nfield value is returned to support chaining.  This enables syntax like:
+// f.NegateVal(f2).AddInt(1) so that f = -f2 + 1.
+func (f *nfieldVal) NegateVal(val *nfieldVal) *nfieldVal {
+	// Negation in the nfield is just the prime minus the value.
+	// Normalize first
+	f.Normalize()
+	if f.n[0]+f.n[1]+f.n[2]+f.n[3]+f.n[4]+f.n[5]+f.n[6]+f.n[7]+f.n[8]+f.n[9] == 0 {
```
davecgh: Not constant time.

btcec/nfield.go
```diff
@@ -0,0 +1,1769 @@
...
+	//
+	// In order to efficiently compute a^(p-2), p-2 needs to be split into
+	// a sequence of squares and multipications that minimizes the number of
+	// multiplications needed (since they are more costly than squarings).
+	// Intermediate results are saved and reused as well.
+	//
+	// The secp256k1 n - 2 is
+	// 2^256 - 432420386565659656852420866394968145601.
+	//
+	// This has a cost of 258 nfield squarings and 33 nfield multiplications.
```
davecgh: This needs to be updated to reflect reality.

btcec/nfield.go
```diff
@@ -0,0 +1,1769 @@
...
+	f.Square().Square().Square().Square().Square() // f = a^(2^256 - 432420386565659656852420866394968145632)
+	f.Mul(&a21)                                    // f = a^(2^256 - 432420386565659656852420866394968145611)
+	return f.Mul(&a10)                             // f = a^(2^256 - 432420386565659656852420866394968145601) = a^(p-2)
+}
+
+// Cmp returns -1 if f < val, 0 if f == val, 1 if f > val
+func (f *nfieldVal) Cmp(val *nfieldVal) int {
+	f.Normalize()
+	val.Normalize()
+	if f.n[9] < val.n[9] {
```
davecgh: Not constant time.

btcec/nfield_test.go
```diff
@@ -0,0 +1,925 @@
...
+	}{
+		{"0", false},
+		{"1", true},
+		{"2", false},
+		// 2^32 - 1
+		{"ffffffff", true},
+		// 2^64 - 2
+		{"fffffffffffffffe", false},
+		// secp256k1 n
+		{"fffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2f", true},
```
davecgh: This should be the order of the curve instead of the prime.

btcec/nfield_test.go
```diff
@@ -0,0 +1,925 @@
...
+		b := new(big.Int).SetBytes(data)
+		c := big.NewInt(0)
+		c.Mul(a, b)
+		c.Mod(c, N)
+		bN := btcec.NewNfieldVal().SetByteSlice(b.Bytes())
+		cN := btcec.NewNfieldVal()
+		cN.Mul2(aN, bN)
+		cstr := fmt.Sprintf("%064x", c)
+		if cN.String() != cstr {
+			t.Fatalf("expected strings to be the same, was not: %v %v", cN.String(), cstr)
```
davecgh: I would highly suggest this test pick a random seed to use for the random number generation and log it in the event of failure.  Alternatively, this test should log the operands used in the calculation.

With neither of those things, if there is a failure, there would be no way to reproduce it in order to debug it.

davecgh: Something like the following:
```Go
	randSeed := rand.Int63()
	defer func() {
		if t.Failed() {
			t.Logf("Random numbers using seed: %v", randSeed)
		}
	}()
	prng := rand.New(rand.NewSource(randSeed))
	// Use prng to generate the random data so it is deterministic from the seed...
```

btcec/nfield_test.go
```diff
@@ -0,0 +1,925 @@
...
+	aN := btcec.NewNfieldVal().SetByteSlice(a.Bytes())
+	for i := 0; i < 100; i++ {
+		c := big.NewInt(0)
+		c.Mul(a, a)
+		c.Mod(c, N)
+		cN := btcec.NewNfieldVal()
+		cN.SquareVal(aN)
+		cstr := fmt.Sprintf("%064x", c)
+		if cN.String() != cstr {
+			t.Fatalf("expected strings to be the same, was not: %v %v", cN.String(), cstr)
```
davecgh: Same thing here regarding the random numbers and debugability.

btcec/nfield_test.go
```diff
@@ -0,0 +1,925 @@
...
+		}
+		a = c
+		aN = cN
+	}
+}
+
+// TestInverse ensures that finding the multiplicative inverse via Inverse works
```
davecgh: Comment doesn't match function name.

btcec/nfield_test.go
```diff
@@ -0,0 +1,925 @@
...
+		result := f.Inverse().Normalize()
+		if !result.Equals(expected) {
+			t.Errorf("nfieldVal.Inverse #%d wrong result\n"+
+				"got: %v\nwant: %v", i, result, expected)
+			continue
+		}
+	}
+}
+
+// TestNfieldInvRand tests the inverse correctness by doing the
```
davecgh: Comment doesn't match function name.

btcec/nfield_test.go
```diff
@@ -0,0 +1,925 @@
...
+			t.Fatalf("failed to read random data")
+		}
+		a := new(big.Int).SetBytes(data)
+		aN := btcec.NewNfieldVal().SetByteSlice(a.Bytes())
+		c := big.NewInt(0)
+		c.ModInverse(a, N)
+		cN := aN.Inverse()
+		cstr := fmt.Sprintf("%064x", c)
+		if cN.String() != cstr {
+			t.Fatalf("expected strings to be the same, was not: %v %v", cN.String(), cstr)
```
davecgh: Same thing here regarding the random numbers and debugability.

davecgh: I'll avoid repeating this comment for the rest, but there are several more cases that this applies to.

btcec/btcec_test.go
```diff
@@ -596,12 +596,83 @@ func TestScalarMult(t *testing.T) {
...
 	}
 }
 
+type splitKTest struct {
+	k      string
+	k1, k2 string
+	s1, s2 int
+}
+
+var s256SplitKTests = []splitKTest{
```
davecgh: Please move this inside the `TestSplitK` function so it is consistent with the field tests.

e.g.

```Go
	tests := []struct {
		k      string
		k1, k2 string
		s1, s2 int
	}{
		{
			"...",
		},
	}

	for i, test := range tests {
		// ...
	}
```

davecgh:
There are several issues reported by Travis that need to be resolved before this can be merged as well:

```
btcec/btcec.go:643:2: should replace c2.n[0] += 1 with c2.n[0]++
btcec/internal_test.go:91:21: exported func NewNfieldVal returns unexported type *btcec.nfieldVal, which can be annoying to use
btcec/internal_test.go:96:53: exported func TstNonceRFC6979 returns unexported type *btcec.nfieldVal, which can be annoying to use
btcec/signature.go:73:1: exported method Signature.Verify should have comment or be unexported
btcec/btcec.go:643:2: should replace c2.n[0] += 1 with c2.n[0]++
btcec/internal_test.go:91:21: exported func NewNfieldVal returns unexported type *btcec.nfieldVal, which can be annoying to use
btcec/internal_test.go:96:53: exported func TstNonceRFC6979 returns unexported type *btcec.nfieldVal, which can be annoying to use
btcec/signature.go:73:1: exported method Signature.Verify should have comment or be unexported
```

btcec/signature.go
```diff
@@ -21,22 +20,17 @@ import (
...
+	one                       = big.NewInt(1)
 )
 
 // Signature is a type representing an ecdsa signature.
 type Signature struct {
-	R *big.Int
-	S *big.Int
+	R       *nfieldVal
+	S       *nfieldVal
+	SNegPad bool
```
davecgh: This is a test-specific thing and shouldn't be part of the struct that non-test code has to use.

waxmihs2902:
Reviewed 1 of 8 files at r1.
Review status: 1 of 8 files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/btcsuite/btcd/621)*

waxmihs2902:
Reviewed 7 of 8 files at r1.
Review status: all files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/btcsuite/btcd/621)*

btcec/signature.go
```diff
@@ -76,10 +70,28 @@ func (sig *Signature) Serialize() []byte {
...
-// key.  It returns true if the signature is valid, false otherwise.
 func (sig *Signature) Verify(hash []byte, pubKey *PublicKey) bool {
-	return ecdsa.Verify(pubKey.ToECDSA(), hash, sig.R, sig.S)
+	zero := new(nfieldVal)
+	if sig.S.Cmp(zero) == 0 || sig.R.Cmp(zero) == 0 {
+		return false
+	}
+	c := pubKey.Curve
+	tmp := hashToInt(hash, c)
+	e := new(nfieldVal).SetByteSlice(tmp.Bytes())
```
davecgh: It would be ideal to avoid the need for `big.Int` here.  Something like a `hashToNfield(hash)`.

btcec/signature.go
```diff
@@ -307,10 +340,10 @@ func recoverKeyFromSignature(curve *KoblitzCurve, sig *Signature, msg []byte,
 	// We calculate the two terms sR and eG separately multiplied by the
 	// inverse of r (from the signature). We then add them to calculate
 	// Q = r^-1(sR-eG)
-	invr := new(big.Int).ModInverse(sig.R, curve.Params().N)
+	invr := new(big.Int).ModInverse(sigR, curve.Params().N)
```
davecgh: Any reason inverse isn't being done in the field where it's a lot faster and then converted to a `big.Int`?

btcec/signature.go
```diff
@@ -307,10 +340,10 @@ func recoverKeyFromSignature(curve *KoblitzCurve, sig *Signature, msg []byte,
...
 
 	// first term.
-	invrS := new(big.Int).Mul(invr, sig.S)
+	invrS := new(big.Int).Mul(invr, sigS)
```
davecgh: Same thing here.  This math should be done in the field where it's faster and convert the results to `big.Int`.

btcec/btcec.go
```diff
@@ -625,38 +626,53 @@ func (curve *KoblitzCurve) Double(x1, y1 *big.Int) (*big.Int, *big.Int) {
...
+	c2.Magnitude(curve.b1, nfieldK)
+	c2.n[0] += 1
+	// k1 = k - c1 * a1 - c2 * a2 from step 5
+	tmp1.Mul2(c1, curve.a1).Normalize()
+	tmp2.Mul2(c2, curve.a2).Normalize()
+	tmp1.Negate()
+	k1.Add2(nfieldK, tmp1)
+	tmp2.Negate().Normalize()
+	k1.Add(tmp2)
+	k1.Normalize().Normalize()
```
davecgh: Double call to `Normalize` here.

waxmihs2902:
:lgtm:

<details><summary><i>Previously, davecgh (Dave Collins) wrote…</i></summary>
> There are several issues reported by Travis that need to be resolved before this can be merged as well:
> 
> 
> 
> ```
> 
> btcec/btcec.go:643:2: should replace c2.n[0] += 1 with c2.n[0]++
> 
> btcec/internal_test.go:91:21: exported func NewNfieldVal returns unexported type *btcec.nfieldVal, which can be annoying to use
> 
> btcec/internal_test.go:96:53: exported func TstNonceRFC6979 returns unexported type *btcec.nfieldVal, which can be annoying to use
> 
> btcec/signature.go:73:1: exported method Signature.Verify should have comment or be unexported
> 
> btcec/btcec.go:643:2: should replace c2.n[0] += 1 with c2.n[0]++
> 
> btcec/internal_test.go:91:21: exported func NewNfieldVal returns unexported type *btcec.nfieldVal, which can be annoying to use
> 
> btcec/internal_test.go:96:53: exported func TstNonceRFC6979 returns unexported type *btcec.nfieldVal, which can be annoying to use
> 
> btcec/signature.go:73:1: exported method Signature.Verify should have comment or be unexported
> 
> ```
</details>


---

Review status: all files reviewed at latest revision, 23 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/btcsuite/btcd/621#-:-KJZ2Ewm5i5WxZIcW8y_:1417088721)*

btcec/nfield_test.go
jimmysong: @davecgh I'm using crypto/rand here. Is your example using something else?

btcec/signature.go
jimmysong: @davecgh SNegPad is used in several places below (non-test code)

waxmihs2902:
:lgtm:

<details><summary><i>Previously, waxmihs2902 (waxmigs2902) wrote…</i></summary>
> :lgtm:
</details>


---

Review status: 2 of 11 files reviewed at latest revision, 24 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/btcsuite/btcd/621#-:-KJZKfmoKf6NVBD2ReU6:1417088721)*

btcec/nfield_test.go
davecgh: Yes, this example is using a prng, which is preferred for testing like this since it is reproducible.

dajohi:
@jimmysong Can you rebase and address the comments?

jimmysong:
rebased, have addressed some of the comments. will put on my todo list.

This is a time-consuming PR to put into mind again. I can't promise this will be there that soon.

davecgh:
@jimmysong: Thanks.  If you would prefer, I can take this over and drive it to completion.  I will of course leave your name on the commit and give you the opportunity to review any modifications I might make.

jimmysong:
@davecgh I would like that very much. It'll be easier for me to review than to code right now.

stevenroose:
I tried to sync testnet with this PR and got this:
```
00:16:15 2017-01-22 [INF] BMGR: Rejected block 00000000036334b7b4272c4d3fc1a72e07159c7938859d1580aca99807232243 from 127.0.0.1:18333 (outbound): failed to validate input 95ec9f464983a63f393f3e7b20a974668213ad76b816206343718a13d9ce897a:0 which references output 19845d028aebcbfd19538fe84887c5a4f29bbf0629c68327b77dca95dc61acde:0 - false stack entry at end of script execution (input script bytes 493046022100d390d20333811dfdac7aa582649532ce02cc95b878d5fbd8f7906cf890d56515022100f442a1aeb43d665e31ae6021f9f596cd06c91ede3c3344f1ace0e61cc45ff0610121029c1d751235217819e6942d8f929ce56fe5e80bb9a12f34fddc9ed889df5a8e30, prev output script bytes 76a91400938bf234076a5ef3cf8936bbc4fa53c3e2661f88ac)
```

davecgh:
@stevenroose: Thanks for pasting results.  I haven't been able to go over this in great detail yet, so that will be handy to track down the issue when I eventually get to it.

stevenroose:
@davecgh I wanted to fix some of the remarks you made on the code in order to get this PR included. It seems like it could be a good performance bonus.

I figured you'd walk into the same issue yourself anyway, though.

I wondered how one would go about making a tree of if-blocks [like this](https://github.com/btcsuite/btcd/pull/621#discussion-diff-65740672R363) one into a constant-time implementation..

davecgh:
@stevenroose: You'd need to convert them to a series of bit manipulations with addition and/or subtraction as necessary to ensure that there are no differences in the branches (at least as much as possible, there is no way to full guarantee what the CPU will do even if the assembler is using constant time bitwise ops, but you still want to give it the best possible chance and in practice bitwise ops are the best bet for that) .

For example, consider the constant time implementation of `IsZero` in `field.go`:

```Go
	bits := f.n[0] | f.n[1] | f.n[2] | f.n[3] | f.n[4] | f.n[5] | f.n[6] | f.n[7] | f.n[8] | f.n[9]
	return bits == 0
```

Now compare that against the typical implementation that is not constant time:

```Go
	return f.n[0] == 0 && f.n[1] == 0 && ... f.n[9] == 0
```

That will short circuit and return on the first non-zero word that is encountered, so different values will result in different running time.  This is easy to see when looking at the generated assembler since it will employ a bunch of "jump if equal" comparison branches.

stevenroose:
Just out of curiosity, where is the limit? `crypto/subtle` provides a method `ConstantTimeEq(x, y int32) int`, which could be used to do all the comparisons like `t9 == nfieldMSBMask`.

bmperrea:
I took a turn at resolving the rest of the comments and rebased - [changes here](https://github.com/bmperrea/btcd/commit/a21b3c8bcf9e0b4b801a9425c8bcac2ed290c6b2). To aid in some of the constant time calculations I introduced a little library following crypto/subtle for uint32. I would appreciate any feedback - I talked to @jimmysong elsewhere and hope to get his approval on my changes soon to move this PR along.

bmperrea:
Also - in response to @stevenroose , I think that an effort to use crypto/subtle in place of any branching to avoid timing attacks based on branch prediction would be good for this repo. To check in on the relative performance of this approach I made [a little example repo](https://github.com/bmperrea/constant-time-go) benchmarking the crypto/subtle functions against their branching counterparts. For all but equality checks there is no measurable performance cost to using functions like in crypto/subtle in place of built-in comparisons with equal-time branches. I will recommend we convert some other constant time function to use these tools in another issue.

l0k1verloren:
I know constant time is critical for interactive EC/AES using TLS/SSL, and I apologise if this is a stupid question, but what benefit does constant time have for validating data that is not using any kind of symmetric encryption nor streaming data containing this across a network in an interactive session?

Wouldn't it be better if the overall processing time was lower? What exactly is the benefit, I saw above in some of the comments regarding a piece of code that compares a number of bitfields, and the non-constant time one falls out as soon as it finds zeroes whereas the constant time function evaluates every field regardless of the fact that finding a zero produces the result?

If security is at stake, sure, I will totally agree that is a needed change, but does anyone really think btcd needs code that slows it down any further? If security is not an issue then this is a total waste of time and effort.

jakesyl:
Any update here?

kcalvinalvin:
@jakesyl 

Btcd was abandoned by the original authors as they went off to make an altcoin. Safe to say this feature is abandoned.

jakesyl:
@kcalvinalvin as far as I can tell @Roasbeef is still active as of November. Decred split off a while ago

jakesyl:
@jcvernaleo (as per #1530)

- low priority
- enhancement
- outdated

PR #622, opened 2016-02-11
server: Make consistent use of server peer stringer.

davecgh:
This updates a couple of logging statements to use the `serverPeer` instance instead of the embedded `peer.Peer` so they are consistent with all of the other log statements.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/622)


jrick:
Should be using `%v` for the format verb (vet complains about this) but otherwise ok.

davecgh:
That's really odd because every other case is using `%s` too.  I don't mind switching them, but it's strange that it's complaining here and not for the others.

PR #623, opened 2016-02-12
database: Remove tx log in favor of new ldb txns.

davecgh:
This removes the intermediate transaction log that was introduced as a part of the database cache as a workaround for `leveldb` batches causing massive memory usage spikes in favor of the recently introduced `leveldb` transaction interface which no longer has the memory usage issues.

This approach is preferred because it can avoid the extra memory needed for the transaction log and therefore all of the intermediate states as well.  As a result, the default cache size has been doubled since it equals roughly the same amount of overall memory usage and the flush interval has been raised as well.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/623)

### DIFF
--- a/database2/ffldb/dbcache.go
+++ b/database2/ffldb/dbcache.go
@@ -6,6 +6,7 @@ package ffldb
 
 import (
 	"bytes"
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


dajohi:
Cannot comment on the code, but the commit messages makes sense.  I've also been running with this diff without regressions.

PR #626, opened 2016-02-12
peer: Extract protocol negotiation from main read and write loops.

jongillham:
This commit does not change functionality other than adjusting some logging messages. It moves version negotiation from the main read/write loops to initial Connect. This simplifies the message flow logic somewhat and will enable a simplified stall handler implementation in a future commit.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/626)

### DIFF
--- a/peer/example_test.go
+++ b/peer/example_test.go
@@ -39,10 +39,7 @@ func mockRemotePeer() error {
 
 		// Create and start the inbound peer.
 		p := peer.NewInboundPeer(peerCfg)
-		if err := p.Connect(conn); err != nil {
-			fmt.Printf("Connect: error %v\n", err)
-			return
-		}
+		p.Connect(conn)
 	}()
 
 	return nil
@@ -93,10 +90,7 @@ func Example_newOutboundPeer() {
 		fmt.Printf("net.Dial: error %v\n", err)
 		return
 	}
-	if err := p.Connect(conn); err != nil {
-		fmt.Printf("Connect: error %v\n", err)
-		return
-	}
+	p.Connect(conn)
 
 	// Wait for the verack message or timeout in case of failure.
 	select {
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -1932,6 +1932,7 @@ func (p *Peer) start() error {
 	case <-time.After(negotiateTimeout):
 		return errors.New("protocol negotiation timeout")
 	}
+	log.Debugf("Connected to %s", p.Addr())
 
 	// The protocol has been negotiated successfully so start processing input
 	// and output messages.
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -1953,7 +1953,7 @@ func (p *Peer) QueueInventory(invVect *wire.InvVect) {
 // the peer is already connected  will have no effect.
 func (p *Peer) Connect(conn net.Conn) error {
 	// Already connected?
-	if atomic.LoadInt32(&p.connected) != 0 {
+	if !atomic.CompareAndSwapInt32(&p.connected, 0, 1) {
 		return nil
 	}
 
@@ -1963,7 +1963,6 @@ func (p *Peer) Connect(conn net.Conn) error {
 	p.conn = conn
 	p.timeConnected = time.Now()
 
-	atomic.AddInt32(&p.connected, 1)
 	return p.start()
 }
 
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -775,7 +775,6 @@ func (p *Peer) pushVersionMsg() error {
 	// recently seen nonces.
 	nonce, err := wire.RandomUint64()
 	if err != nil {
-		fmt.Println(err)
 		return err
 	}
 	sentNonces.Add(nonce)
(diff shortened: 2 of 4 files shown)


jongillham:
The Travis build currently consistently fails with Go 1.5.1 and I can't seem to recreate it on my machine so am having to figure out the issue by trying to debug remotely. I have discovered numerous benign races within the peer package and believe it is likely to be due to one of them using a slightly different goroutine scheduler in 1.5.1.

jongillham:
A fresh install of Go 1.5.1 allows me to recreate the issue.

jongillham:
The CI error was due a race in the tests that was easy to address. The race was made more likely due to commit 383ed0 however it is of no consequence to the running of a peer.

davecgh:
Thanks for another PR regarding this.  I'm glad to see it getting some attention to make it more usable by 3rd parties as that was the intention of making it a package.

I probably won't be able to get to the reviewing this for a few days, but I wanted to drop a note letting you know it's not being ignored.

jongillham:
Cheers @davecgh. There are a few other things I want to do such as rework the stall handler, improve the unit tests and simplify the disconnect/shutdown code paths. Would you recommend I continue adding commits on top of this pull request and let you just choose the granularity of commits to master or create separate pull requests for each addition?

davecgh:
I prefer the smaller PRs like this.  It's easier to review and get them in with smaller chunks.  It also lends itself well if there are any controversial parts since it doesn't necessarily block the other ones.

Typically what I do is make the new PRs such that they are rebased on top of the commit(s) of the dependent PR(s) and add a comment that the PR requires the others.  It's a little more rebase work, but it tends to make the review process a lot smoother, means each PR can be pulled and tested easily, and help prevent blocking future work while the dependent PRs are being reviewed.

davecgh:
@jongillham: Sorry for the delay on these PRs.  As I'm sure you've noted there was a big push to finish up the major database/chain infrastructure changes and get them to master.  I'm starting to go over these again.

I noticed that none of them merge cleanly at the moment, so they'll need a rebase.

jongillham:
@davecgh not at all. It's great to see the progress made in the last couple of days! I'll rebase and retest over the next few days.

davecgh:
Would you rebase this to the latest master please?  It can't be merged cleanly at the moment.

davecgh:
The first commit which uses `CompareAndSwap` looks good and I agree with the change.

Overall I like where this is going, but I'm not following the reasoning for removing the error from `Connect` and making it launch a goroutine.  The change means any errors from the initial version push and negotiation can't be detected programmatically by the caller and also removes the caller's ability to choose whether or not it wants to block until that happens.

Other than that though, this looks good.

jongillham:
Hi @davecgh, thanks for reviewing the code.

Even though I removed the error from `Connect`, I don't particularly like how it is right now. Mainly because any errors generated by connection should really be returned to the `peer` client as you say.

Having said that, the reasons I removed the error return from `Connect` are as follows:

- It means that clients of the `peer` API only have one blocking function `WaitForDisconnect()` to worry about. Now everything else is non-blocking. This means that if a client wants to create ten peer connections, they just need to call `Connect` and block on `WaitForDisconnect()` ten times, _potentially_ allow us to remove some or all of the `pendingPeers` functionality in `server.go`.

- Aesthetically, it's symmetrical with `Disconnect`. :bowtie:

What I would actually like to see is `WaitForDisconnect()` to return an error from any issues such as connection, protocol negotiation, closing of the connection etc. In a similar vein to how we check `Err` in [`bufio.Scanner`](https://golang.org/pkg/bufio/#Scanner) after using it. It could be nice to use something like [`errors.Wrap`](https://godoc.org/github.com/pkg/errors#Wrap) for the returned `WaitForDisconnect()` error to ensure we catch cascading errors from the peer.

The reason I didn't try any of the above in this pull request is to keep it simple. From what I can see, all the returned `Connect` errors do in current `server.go` code is to call `Disconnect` and log the error which I believe I have replicated.

What do you think about the potential direction?

davecgh:
Hmm.  That makes sense, however I'm also trying to keep in mind #653 which is nearing completion.  The new code in the connection manager it provides makes use of the error on connect like so:

```Go
		Dial:          btcdDial,
		OnConnection: func(c *connmgr.ConnReq, conn net.Conn) {
			sp := s.newOutboundPeer(c.Addr, c.Permanent)
			if sp != nil {
				if err := sp.Connect(conn); err != nil {
					srvrLog.Errorf("Can't use connection: %v", err)
					return
				}
				sp.connReq = c
				srvrLog.Debugf("Connected to %s", sp.Addr())
				s.addrManager.Attempt(sp.NA())
			}
		},
```

However, I suspect that can be dropped without much hassle given a failed negotiation will result in a Disconnect which leads to the correct behavior.

davecgh:
Given the above, I'm going to test this out for a little bit and assuming nothing crops up, I'll get it merged.

davecgh:
I ran this over the weekend with no issues.

OK

PR #627, opened 2016-02-12
mempool:  reduce lock contention

dajohi:
[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/627)

### DIFF
--- a/mempool.go
+++ b/mempool.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2013-2014 The btcsuite developers
+// Copyright (c) 2013-2016 The btcsuite developers
 // Use of this source code is governed by an ISC
 // license that can be found in the LICENSE file.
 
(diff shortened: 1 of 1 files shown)


davecgh:
The idea of reducing contention is sound, but the particular approach used here is not future safe:

```Go
	t := time.Now().AddDate(300, 0, 0) // 300 years from now
	fmt.Println(t)
	fmt.Println(time.Unix(0, t.UnixNano()))
```

Output:
```
2316-02-12 13:27:49.9477848 -0600 CST
1731-07-24 13:53:16.238233184 -0600 CST
```

That said, the last update time really doesn't need nanosecond precision though, so I would suggest just using `t.Unix()` and `time.Unix(secs, 0)`.  That will work properly:

```
	t := time.Now().AddDate(10000, 0, 0) // 10,000 years from now
	fmt.Println(t)
	fmt.Println(time.Unix(t.Unix(), 0))
```

Output:
```
12016-02-12 13:33:01.2293125 -0600 CST
12016-02-12 13:33:01 -0600 CST
```

dajohi:
Updated.

davecgh:
OK

PR #629, opened 2016-02-19
peer: Separate ping ticker, inventory and message queues.

jongillham:
This pull request relies on #626. It separates the concerns of the ping ticker as well as the inventory and message queues and arguably allows easier rationalisation of execution paths.


davecgh:
Thanks for the PR!

I'll look at this a bit more closely after I get to #626 (probably this weekend), but I'm curious what the benefit of adding more goroutine overhead here is?  I generally prefer to reduce the number of goroutines rather than increase them when possible unless there is a clear reason to launch one.

It seems to me like trickling inventory fits well within the output handler goroutine as it is dealing with sending messages.  It also then can make use of the same queuing mechanism and ensure it intermingles well with the normal output path and doesn't allow one or the other to dominate since it is able to use a single select statement and the select statement has a fairness algorithm built in

Before the peer package refactor there were only 3 goroutines per peer.  Now there 6 I believe due to the changes that were made in server (I'm really not happy with the way the server handles peer connection right now at all to be honest, but that has nothing to do with this PR).  This PR would bring the total number of goroutines to 8.  Each goroutine also has roughly 4kb overhead, so we're talking about 32kb of overhead per peer here.  That might not seem like much for a normal home user, but when you're using it to back a service where there are thousands of nodes connecting, that can quickly balloon to hundreds of MBs.

jongillham:
Hi @davecgh thanks for the feedback. This pull request is really a midway point to removing the poll based stallHandler goroutine and replacing it with push based mechanism.

The benefit of 5 permanent goroutines and one temporary one per peer would be to gain a push based stall detector and much cleaner separation of concerns. I suppose in terms of performance we are reducing CPU cycles and increasing maintainability at the expense of memory. However if power users are mainly finding the limiting factor is memory then I'll need to figure out a way to reduce that whilst maintaining separation of concerns.

I completely forgot that the select statement had a fairness algorithm! I keep assuming select worked it's way down each case statement in the order they are written if everything was unblocked - at the expense of channel starvation. I might to have to make use of that for exactly the reasons you described.

I too am not particularly comfortable with the way the server makes use of peer connections which was the original motivation for looking at the peer code. I think the server codes responsiveness can be tightened somewhat. However I wanted to 'optimise' the peer code before I looked at the server side.

Maybe it would be best for me to submit the final pull request associated with this refactoring before you review anything. You'll then be able to see the finished product. It will be quite a big one but at least will only touch the internals of peer, not the server like this request and the previous one. If you like the direction/see any benefit then we can continue.

jongillham:
@davecgh after doing some testing I agree that not increasing the number of goroutines in the peer package is a good rule to follow, even though the resulting code may be less idiomatic Go. I've reworked this request a bit to keep the number of goroutines the same. I've also added a preceding pull request #639 to make review easier.

In the next one or two pull requests I'll be able to remove the stall handler goroutine as well. However, I would appreciate your review and merging if you're happy with the work before I carry on.

davecgh:
In regards to the initial discussion regarding the way the server handles peer connection, it has recently been greatly improved by the `connmgr` package, so I'll be reviewing this again with that in mind.

davecgh:
Once #639 goes in and this is rebased over it (and the necessary small tweaks that will entail), this should be able to go in without any issues.

While I don't find the current code regarding the inventory trickling to be particularly egregious since, as previously mentioned, it makes use of the fairness of the select statement, I also don't have any strong objections to splitting it out as this PR is doing.  Splitting it out like this does improve the readability and that is a clear win.

dajohi:
@jongillham can you please rebase?

jongillham:
Will do as soon as I can.

jakesyl:
@jcvernaleo (as per #1530)

- Low priority
- Outdated 
- Enhancement

jakesyl:
@jongillham can you rebase (or if you're not interested, mind if I rebase?)

jongillham:
@jakesyl very happy for you to! 🤠

jakesyl:
Cool - will go ahead and do so

jcvernaleo:
Closing this since the work has moved to #1544

PR #631, opened 2016-02-19
findcheckpoint: Update to allow first checkpoint.

davecgh:
This updates the `findcheckpoint` utility to work when there are not already any checkpoints.  This doesn't really matter for Bitcoin at the current time, but if a new `testnet` is created it will not have any
checkpoints to start with and this change also means the utility can work for alts.

While here, switch a couple of error prints to ensure they contain a final newline.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/631)


jcvernaleo:
Assuming travis every finishes, this looks good to me.

OK

PR #635, opened 2016-02-23
Look in datadir for RPC cert+key.

jcvernaleo:
btcwallet sets the path for the rpc cert and key
to the directory specified by the datadir option.

This adds the same logic to btcd.

Pointed out in decred/dcrd#52

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/635)

### DIFF
--- a/config.go
+++ b/config.go
@@ -435,6 +435,18 @@ func loadConfig() (*config, []string, error) {
 		return nil, nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


jcvernaleo:
As discussed in decred/dcrd#52 this isn't really the right way to solve this so I'm closing the issue.  I'll try to do a PR with a solution more like we discussed in the issue soon.

PR #636, opened 2016-02-25
multi: Update with result of gofmt -s.

davecgh:
This pull request updates the code to make use of the most recent simplified output from gofmt.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/636)


jrick:
README needs to be updated saying it required Go 1.5.

davecgh:
These changes shouldn't require Go 1.5.  It passed Travis which still runs Go 1.4.2 as well.

jrick:
I suppose, but only because there are no map key literals that were simplified.

https://golang.org/doc/go1.5#language

davecgh:
Right, I recall seeing that addition to 1.5, but there were no cases of that here.

jrick:
Since my only complaint was fixed separately on master: ok

davecgh:
Even though this doesn't require it anyways, the README has been updated for Go 1.6 via #638.

PR #639, opened 2016-02-26
peer: Separate ping ticker functionality from outHandler.

jongillham:

### DIFF
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -1746,10 +1746,6 @@ func (p *Peer) shouldLogWriteError(err error) bool {
 // goroutine.  It uses a buffered channel to serialize output messages while
 // allowing the sender to continue running asynchronously.
 func (p *Peer) outHandler() {
-	// pingTicker is used to periodically send pings to the remote peer.
-	pingTicker := time.NewTicker(pingInterval)
-	defer pingTicker.Stop()
-
 out:
 	for {
 		select {
@@ -1790,14 +1786,6 @@ out:
 			}
 			p.sendDoneQueue <- struct{}{}
 
-		case <-pingTicker.C:
-			nonce, err := wire.RandomUint64()
-			if err != nil {
-				log.Errorf("Not sending ping to %s: %v", p, err)
-				continue
-			}
-			p.QueueMessage(wire.NewMsgPing(nonce), nil)
-
 		case <-p.quit:
 			break out
 		}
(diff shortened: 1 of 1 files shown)


jongillham:
Requires pull request #626.

server.go
```diff
@@ -1573,9 +1568,7 @@ func (s *server) establishConn(sp *serverPeer) error {
 	if err != nil {
 		return err
 	}
-	if err := sp.Connect(conn); err != nil {
-		return err
-	}
+	sp.Connect(conn)
```
dajohi: If the connect fails, you will get both:
log.Errorf("Cannot start peer %v: %v", p, err)
srvrLog.Debugf("Connected to %s", sp.Addr())

jongillham: @dajohi thanks for the review. I'll update with a solution that gives non-contradictory logging.

jongillham:
@davecgh I thought I would start back with the peer refactoring. Better separation of concerns is the only benefit of this request - at the expense of more state variables. However it helps work towards removing the stalling handler polling and potentially remove the need for code to keep track of 'pending peers'.

davecgh:
@jongillham Sounds good.  Both this PR and #629 are almost to the top of me PR review list, so hopefully I should be able to get to them before too much longer.

jrick:
I agree with making it a new goroutine, as @davecgh suggested.  reads better and is easier to understand.

jongillham:
@davecgh it certainly reads easier to have a separate goroutine for ping ticker. That is how the pull request originally was however I remember having a discussion with you on another thread about the desire to keep the number of goroutines for each peer down to a minimum.

After that I had a quick look at the Go internals to see how `time.AfterFunc` worked and got the impression that it amortised its cost by mainly creating one goroutine to handle all firing of timer events. Here's a [link](https://github.com/golang/go/blob/2481481ff798636907376bfdf7e8c7558b8b930e/src/runtime/time.go#L64) to the code I looked at.

Of course, if I misread the `time` code and one goroutine is created for each `time.AfterFunc` and/or you don't have the requirement for keeping the number of goroutines down to a minimum then I would certainly vote for increased readability as you suggest.

davecgh:
@jongillham: That's a fair point about amortization and one I didn't actually know, although it does make sense the stdlib would do that.  I do recall having the discussion about goroutines (although I'd have to review the specific details), and I definitely think it's something we want to keep an eye on.

At any rate, I think given the readability improvement, and the fact that several people prefer the goroutine for this, it justifies the addition.  Sorry for the mixed signals on that front.

If you want to update it to the goroutine approach, I'll go ahead and merge it after that.  I'll be looking at your other peer PR (#629) next.

jongillham:
@davecgh updated.

PR #640, opened 2016-02-27
server: Appropriately name inbound peers map in peerState.

jongillham:
[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/640)


davecgh:
OK

PR #641, opened 2016-03-02
server: Allow peer requestQueue to be garbage collected.

jongillham:
I stumbled over `serverPeer.requestQueue` and it appears that with the [current implementation](https://github.com/btcsuite/btcd/blob/master/blockmanager.go#L1041) for reading from the queue will not allow the underlying slice to get garbage collected as each message is read. Instead the backing array will just get bigger until the peer is garbage collected. There's an easy fix but I took the liberty of creating fix that limits unnecessary allocations.

I haven't found the current (pre pull request) functionality to be an issue but I thought I would post this one anyway for you to decide if the added complexity is worth it. (This might be a case of optimising before necessary.)

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/641)

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -1030,45 +1030,55 @@ func (b *blockManager) handleInvMsg(imsg *invMsg) {
 		}
 	}
 
-	// Request as much as possible at once.  Anything that won't fit into
-	// the request will be requested on the next inv message.
-	numRequested := 0
-	gdmsg := wire.NewMsgGetData()
-	requestQueue := imsg.peer.requestQueue
-	for len(requestQueue) != 0 {
-		iv := requestQueue[0]
-		requestQueue[0] = nil
-		requestQueue = requestQueue[1:]
+	// Request the required inventory.
+
+	// Avoid new allocations by reusing the requestQueue's slice to filter only
+	// the invvects we need to send.
+	requiredInvVects := imsg.peer.requestQueue[:0]
+
+	for _, iv := range imsg.peer.requestQueue {
 
 		switch iv.Type {
 		case wire.InvTypeBlock:
-			// Request the block if there is not already a pending
-			// request.
+			// Request block if there is not already a pending request.
 			if _, exists := b.requestedBlocks[iv.Hash]; !exists {
 				b.requestedBlocks[iv.Hash] = struct{}{}
 				imsg.peer.requestedBlocks[iv.Hash] = struct{}{}
-				gdmsg.AddInvVect(iv)
-				numRequested++
+
+				requiredInvVects = append(requiredInvVects, iv)
 			}
 
 		case wire.InvTypeTx:
-			// Request the transaction if there is not already a
-			// pending request.
+			// Request transaction if there is not already a pending request.
 			if _, exists := b.requestedTxns[iv.Hash]; !exists {
 				b.requestedTxns[iv.Hash] = struct{}{}
 				imsg.peer.requestedTxns[iv.Hash] = struct{}{}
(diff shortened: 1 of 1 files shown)


davecgh:
I can see why this would look like it's not being garbage collected, but it actually is due to the functionality of append.  I agree that it's not very obvious, but what happens is effectively the following.  I've used `{}` to annotate the backing array and `[]` to annotate the slice within the backing array.

```
requestQueue := {[iv1, iv2, iv3, iv4]}
requestQueue = requestQueue[1:]: {nil, [iv2, iv3, iv4]} // offset 1, len 3, cap 3, backing array 4
requestQueue = requestQueue[1:]: {nil, nil, [iv3, iv4]} // offset 2, len 2, cap 2, backing array 4
requestQueue = requestQueue[1:]: {nil, nil, nil, [iv4]} // offset 3, len 1, cap 1, backing array 4
// appending exceeds the offset+cap, so so a new backing array with a cap of 2 is created
// with iv4 copied to offset 0 and iv5 copied to offset 1 thus freeing up the old backing array
requestQueue = append(requestQueue, iv5)
requestQueue: {[iv4, iv5]} // offset 0, len 2, cap 2
```

To see this in action, observe the following program that digs out the pointer to the slice data using reflect.

```Go
package main

import (
	"fmt"
	"reflect"
	"unsafe"
)

func main() {
	requestQueue := make([]int, 0, 2)
	fmt.Println("Notice how the data address doesn't change for the first two appends")
	fmt.Printf("%+v\n", (*reflect.SliceHeader)(unsafe.Pointer(&requestQueue)))
	requestQueue = append(requestQueue, 1)
	fmt.Printf("%+v\n", (*reflect.SliceHeader)(unsafe.Pointer(&requestQueue)))
	requestQueue = append(requestQueue, 2)
	fmt.Printf("%+v\n", (*reflect.SliceHeader)(unsafe.Pointer(&requestQueue)))
	fmt.Println("Notice how exceeding the cap causing a new backing array")
	requestQueue = append(requestQueue, 3)
	fmt.Printf("%+v\n", (*reflect.SliceHeader)(unsafe.Pointer(&requestQueue)))
}
```
Output:
```
Notice how the data address doesn't change for the first two appends
&{Data:826814800496 Len:0 Cap:2}
&{Data:826814800496 Len:1 Cap:2}
&{Data:826814800496 Len:2 Cap:2}
Notice how exceeding the cap causing a new backing array
&{Data:826814777216 Len:3 Cap:4}
```

Also, I believe this pull request would change the functionality.  Currently it only sends a single `getdata` message per batch of processed inventory vectors advertisements.  This pull request would potentially make it send multiple `getdata` messages per inventory vector advertisement which doesn't conform to the protocol.

I have no objections to making it easier to follow and/or more explicit, but we need to careful not to change the functionality.

jongillham:
Thanks for taking the time to explain! I had not considered append was being used like that.

Yes it did slightly change functionality. I checked the the [`bitcoind`](https://github.com/bitcoin/bitcoin/blob/8b70a64d62c6e64288762d062414cc979f880c54/src/net.cpp#L2436) implementation and `btcd` functionality seemed to already differ slightly from it so I thought I would add a potentially more effective way of requesting inventory and see what happened.

`bitcoind` appears to throw away pending messages if the following limits are reached:

```
if (mapAskFor.size() > MAPASKFOR_MAX_SZ || setAskFor.size() > SETASKFOR_MAX_SZ)
        return;
```

Where as `btcd` saves them for the next round of `invvect`s. (I'm quite possibly mistaken though as `bitcoind` code can be like a can of spaghetti.)

Issue #642, opened 2016-03-03
Blockchain stops processing after a while when the RPC server isn't used.


jongillham:
I can recreate this problem if I set `btcd --norpc --maxpeers=4`. (Setting `--maxpeers` to something small like 4 makes the issue occur quicker.)

The cause seems to be that [`rebroadcastHandler`](https://github.com/btcsuite/btcd/blob/master/server.go#L2060) is not started so the buffered channel [`relayNtfnChan`](https://github.com/btcsuite/btcd/blob/master/server.go#L1968) never gets drained. The `relayNtfnChan` is populated by the mempool's [`ProcessTransaction`](https://github.com/btcsuite/btcd/blob/master/mempool.go#L976) method. `relayNtfnChan` is buffered by `MaxPeers` amount so when the buffer is full it blocks the main `btcd` processing loop.

Looking into this I noticed a couple of potential problems in the current code that would also need to be addressed:

1. `RelayNtfnChan` is never nil and some code paths bogusly rely on detecting if `RelayNtfnChan` is nil or not.
2. When the RPC server is disabled I believe transactions aren't relayed to other peers because [`rebroadcastHandler`](https://github.com/btcsuite/btcd/blob/master/server.go#L1960) isn't started. I feel it would be cleaner for the rebroadcasting of inventory and relaying of transactions functionality should be separated out our `rebroadcastHandler`.

This issue might be related to #605 which is what I initially looked into and how I stumbled on this. My dev computer and test server are ancient and always under high load just running the operating system and so far I've *never* managed to run `btcd` more than a couple of hours (with the RPC server running) without it stalling. Fixing this issue will hopefully help track down the root cause of #605.

I'm happy to look further into this and provide a fix but my analysis to this point has just been visual inspection of the code and using `pprof`. Can anyone else recreate this issue? Do you have any thoughts on the best fix?

dajohi:
I believe you are on the right track..

1. mempool will become its own subpkg, so RelayNtfnChan could be nil as things other than btcd may use the package.

2. Yes, this is a bug and should be fixed.

jongillham:
@dajohi I created pull request #643 that fixes this issue before your reply about RelayNtfnChan nil. The request removes the check for nil. I feel this is probably the appropriate thing to do with the codebase as it stands but am happy to change it back if you want.

dajohi:
Please put the nil checks back.  The checks will be needed once mempool subpkg PR shows up making the diff bigger.

[...]

PR #643, opened 2016-03-03
Fix server stalling when RPC server is not running.

jongillham:
This resolves issue #642.

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/643)

### DIFF
--- a/server.go
+++ b/server.go
@@ -1954,14 +1954,10 @@ func (s *server) UpdatePeerHeights(latestBlkSha *wire.ShaHash, latestHeight int3
 	}
 }
 
-// rebroadcastHandler keeps track of user submitted inventories that we have
-// sent out but have not yet made it into a block. We periodically rebroadcast
-// them in case our peers restarted or otherwise lost track of them.
-func (s *server) rebroadcastHandler() {
-	// Wait 5 min before first tx rebroadcast.
-	timer := time.NewTimer(5 * time.Minute)
-	pendingInvs := make(map[wire.InvVect]interface{})
-
+// relayTransactionsHandler relays transactions sent on relayNtfnChan to other
+// peers and to the RPC infrastructure if necessary.  It must be run as a
+// goroutine.
+func (s *server) relayTransactionsHandler() {
 out:
 	for {
 		select {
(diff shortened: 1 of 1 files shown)


dajohi:
Please keep the name RelayNtfnChan.

jongillham:
Done. The commit is boring now! ;-)

server.go
```diff
@@ -1976,9 +1972,28 @@ out:
...
-				s.rpcServer.gbtWorkState.NotifyMempoolTx(s.txMemPool.LastUpdated())
+				s.rpcServer.gbtWorkState.NotifyMempoolTx(
+					s.txMemPool.LastUpdated())
 			}
 
+		case <-s.quit:
+			break out
+		}
+	}
+	s.wg.Done()
```
davecgh: I believe there will need to be  cleanup section here that drains the `s.relayNtfnChan` channel.  Otherwise shutdown could hang because queued items in the sender would block forever.

The following should do the trick:

```Go
	// Drain channels before exiting so nothing is left waiting around
	// to send.
cleanup:
	for {
		select {
		case <-s.relayNtfnChan:
		default:
			break cleanup
		}
	}
	s.wg.Done()
```

davecgh:
Thanks for tracking down the issue and submitting a PR.  I agree with you that it would be an issue when the `rebroadcastHandler` isn't started with the way it currently is.

jongillham:
@davecgh your cleanup comment just got deleted with my forced update, but nice catch. Cleanup code added.

davecgh:
@jongillham: Ah, I almost always use the files tab so that doesn't happen. I guess I was on the commit tab.  Oops!

Anyways, the code reads well to me now.  I'll try to reproduce the issue and verify this resolves it later tonight.

jongillham:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[server.go, line 1983 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/643#-KBwzilePIx0sF8eBwOG-r2-1983:-KCNBVMN4qzjM5p2oZZy:-499146009) ([raw file](https://github.com/btcsuite/btcd/blob/12dd38f6f5f9ece6ce0d14c2735ecaf52e2bad53/server.go#L1983)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/643)*

jongillham:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[server.go, line 1983 \[r2\]](https://reviewable.io:443/reviews/btcsuite/btcd/643#-KBwzilePIx0sF8eBwOG-r2-1983:-KCNBd7BntFh2TZRDtYY:-499146009) ([raw file](https://github.com/btcsuite/btcd/blob/12dd38f6f5f9ece6ce0d14c2735ecaf52e2bad53/server.go#L1983)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/btcsuite/btcd/643)*

davecgh:
I was able to reproduce the issue and confirm this PR fixes it.  Sorry for the delay.  For some reason I thought I had already done this.

OK

PR #648, opened 2016-03-11
docs: Update READMEs with current details.

### LINKED ISSUE #647: Remove btcgui reference from README
it's dead, jim.

davecgh:
This commit updates the main `README.md` and `docs/README.md` files to replace the references to the now dead btcgui project with the Windows-only Paymetheus project.

While here, it also updates some information to make it more current and accurately describe the current status.

Closes #647.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/648)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,7 @@ btcd
 btcd is an alternative full node bitcoin implementation written in Go (golang).
 
 This project is currently under active development and is in a Beta state.  It
-is extremely stable and has been in production use for over 6 months as of May
-2014, however there are still a couple of major features we want to add before
-we come out of beta.
+is extremely stable and has been in production use since October 2013.
 
 It properly downloads, validates, and serves the block chain using the exact
 rules (including bugs) for block acceptance as Bitcoin Core.  We have taken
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -7,9 +7,8 @@ btcd
 btcd is an alternative full node bitcoin implementation written in Go (golang).
 
 This project is currently under active development and is in a Beta state.  It
-is extremely stable and has been in production use for over 6 months as of May
-2014, however there are still a couple of major features we want to add before
-we come out of beta.
+is extremely stable and has been in production use for just under 2 and a half
+years as of March 2016.
```
jrick: Maybe say "in production use since ..." instead of updating this line each time?

README.md
```diff
@@ -29,9 +28,9 @@ wallet functionality and this was a very intentional design decision.  See the
 blog entry [here](https://blog.conformal.com/btcd-not-your-moms-bitcoin-daemon)
 for more details.  This means you can't actually make or receive payments
 directly with btcd.  That functionality is provided by the
-[btcwallet](https://github.com/btcsuite/btcwallet) and
-[btcgui](https://github.com/btcsuite/btcgui) projects which are both under
-active development.
+[btcwallet - CLI-only](https://github.com/btcsuite/btcwallet) and
```
jrick: Not sure I'd say CLI-only.  It powers paymetheus, after all.

README.md
```diff
@@ -29,9 +28,9 @@ wallet functionality and this was a very intentional design decision.  See the
...
+[Paymetheus - Windows-only](https://github.com/btcsuite/Paymetheus) projects
```
jrick: If the CLI-only line is removed from btcwallet, I think it would read better to put (Windows only) in parentheses after the link.

README.md
davecgh: Good point.  Will make that change.

README.md
davecgh: True.  I was trying to make it clear for GUI users that's probably not the one they want to use, but as long as the documentation for btcwallet calls that out, it make sense.

README.md
davecgh: Sounds good.  Will change.

jrick:
ok

PR #573, opened 2015-12-02
WIP: blockchain: implement new blockchain indexing infrastructure

Dirbaio:
**This PR requires #491**

As discussed with @davecgh over IRC, here's my proposed implementation for indexes on the new database.

Quick overview of the changes:

- Define a new interface, Index, with callbacks for connecting and disconnecting blocks.
- Add support in the blockchain package for enabling indexes when instancing a BlockChain.
- Make blockchain manage the lifecycle of index buckets, tracking of the index tips, catch up, call ConnectBlock/DisconnectBlock on the indexes. This avoids duplicating code per index type.
- Perform Block connection and index update atomically in the same database transaction.
- Add index registry, similar to the database2 driver registry, which enables registering indexes by name.
- Add new flags --index and --dropindex, for enabling and dropping indexes.
- Implement the old transaction-by-address as an index.

Indexes can be enabled and disabled at will. Dropping the index data is not necessary for disabling it, this allows a user to later re-enable an index without having to catch up from scratch again. Catch up works properly even if the index tip has been orphaned by a reorganize: catch up pops blocks from the index until the tip is back on the main chain, and then catches up to the main chain tip.

This introduces two new flags, --index and --dropindex. Both can be used multiple times to enable or drop several indexes, for example:
```
btcd --index=txbyaddr --index=txbyhash
```

For backwards compatibility, the --addrindex flag will be defined to be equivalent to --index=txbyaddr

The currently mandatory tx index will be rendered optional when @davecgh finishes work on the UTXO set: it can then be implemented as an optional index too. The flag --txindex could be added as an equivalent with --index=txbyhash for compatibility with Bitcoin Core.

TODO list:
- [ ] Make index catchup log progress like IBD and the old addrindex, and make it interruptible with ctrl+c.
- [x] Implement searchrawtransactions
- [x] Implement mempool addrindex
- [ ] Implement dropping indexes (--dropindex flag)
- [ ] Implement backwards-compatibility for -addrindex, same as --index=txbyaddr

[Review on Reviewable](https://reviewable.io/reviews/btcsuite/btcd/573)

PR #650, opened 2016-03-29
indexers: Implement optional tx/address indexes.

davecgh:
**This PR requires #491**

**NOTE: This pull request makes the transaction index optional.  If you want to retain the old functionality, you'll need to add `txindex=1` to the config file or enable it on the command line with `--txindex`.**

This pull request expands upon and finishes work done by @Dirbaio in #573 (which in turn was based on some of my own work) and therefore takes it place.  In particular, the credit for the majority of the address index level-based design and the idea to make the indexes a separate package goes to @Dirbaio.

----

This introduces a new indexing infrastructure for supporting optional indexes using the new database and `blockchain` infrastructure along with two concrete indexer implementations which provide both a transaction-by-hash and a transaction-by-address index.

The new infrastructure is mostly separated into a package named `indexers` which is housed under the `blockchain` package. In order to support this, a new interface named `IndexManager` has been introduced in the `blockchain` package which provides methods to be notified when the chain has been initialized and when blocks are connected and disconnected from the main chain.  A concrete implementation of an index manager is provided by the new `indexers` package.

The new `indexers` package also provides a new interface named `Indexer` which allows the index manager to manage concrete index implementations which conform to the interface.

The following is high level overview of the main index infrastructure changes:
- Define a new `IndexManager` interface in the `blockchain` package and modify the package to make use of the interface when specified
- Create a new `indexers` package
 - Provides an `Index` interface which allows concrete indexes to plugin to an index manager
  - Provides a concrete `IndexManager` implementation:
    - Handle the lifecycle of all indexes it manages
    - Tracks the index tips
    - Handles catching up disabled indexes that have been reenabled
    - Handles reorgs while the index was disabled
    - Invokes the appropriate methods for all managed indexes to allow them to index and deindex the blocks and transactions
  - Implement a transaction-by-hash index
    - Makes use of internal block IDs to save a significant amount of space and indexing costs over the old transaction index format
  - Implement a transaction-by-address index
    - Makes use of a leveling scheme in order to provide a good tradeoff between space required and indexing costs
- Supports enabling and disabling indexes at will
- Supports the ability to drop indexes if they are no longer desired

The following is an overview of the btcd changes:

- Add a new index logging subsystem
- Add new options `--txindex` and `--addrindex` in order to enable the optional indexes
 - NOTE: The transaction index will automatically be enabled when the address index is enabled because it depends on it
- Add new options `--droptxindex` and `--dropaddrindex` to allow the indexes to be removed
  - NOTE: The address index will also be removed when the transaction index is dropped because it depends on it
- Update `getrawtransactions` RPC to make use of the transaction index
- Reimplement the `searchrawtransaction` RPC that makes use of the address
- Update `sample-btcd.conf` to include sample usage for the new optional index flags

---
This change is [Reviewable](https://reviewable.io/reviews/btcsuite/btcd/650)


davecgh:
Travis failures are a false positive and only apply to Go 1.4.3 due to the lint package no longer being compatible.

blockchain/indexers/addrindex.go
```diff
@@ -0,0 +1,932 @@
...
+
+	// level0MaxEntries is the maximum number of transactions that are
+	// stored in level 0 of an address index entry.  Subsequent levels store
+	// 2^n * level0MaxEntries entries, or in words, double the maximum of
+	// the previous level.
+	level0MaxEntries = 8
+
+	// addrKeySize is the number of bytes an address key consumes in the
+	// index.  It consists of 1 byte address type + 20 bytes hash160.
+	addrKeySize = 1 + 20
```
Roasbeef: As part of the new segwit softfork, two new address types have been introduced. One of the new address types, pay-to-witness-script-hash or P2WSH, has the following "standard" pkScript template: 
  * `0 <32-byte-hash>`

Functionally, this is just regular p2sh, but now with a 32-byte hash (sha256) rather than a 20-byte hash (hash160).

As a result, all "traditional" addresses can no longer be uniquely pattern-matched with a 20-byte data push. I think it'd be prudent to modify the way the format of the `addrKey`, in addition to the way it's calculated once a "standard" address type is encountered. 

I propose the `addrKeySize` be promoted to be 33 bytes (1 byte address type + 32 byte sha256). While indexing a transaction's inputs, and outputs, whenever a known address is encountered if it contains a salient 20-byte data push (`btcutil.Address.ScriptAddress()`), then we hash that to create a 32-byte key. If the data push is instead 32-bytes, we just use that directly, skipping the intermediate hashing phase. 

Thoughts?

Roasbeef:
Nice work @davecgh! This is an excellent successor to the chain indexing work I implemented previously. The `IndexManager`, and `Index` interfaces make the infrastructure easily managed, and extended in the future. I also like how you decoupled the unconfirmed transaction address index from the mempool, letting the `AddrIndexer` manage that state itself. 

I'll be testing this PR over the next day or two.

blockchain/indexers/addrindex.go
Roasbeef: Alternatively, we could do the inverse: reducing the 32-byte hash down to a 20-byte hash via a hash160 call.

blockchain/indexers/manager.go
```diff
@@ -0,0 +1,648 @@
...
+			hash, height, err = dbFetchIndexerTip(dbTx, idxKey)
+			if err != nil {
+				return err
+			}
+			return nil
+		})
+		if err != nil {
+			return err
+		}
+
```
Roasbeef: Just ran into an issue while testing.

Steps to reproduce: 
  1. Start with a fresh `data` directory (no indexes, no blocks, etc)
  2. Sync up full chain
  3. Start with `btcd --txindex` or `btcd --addrindex`
 
Result: 
```
21:19:18 2016-03-29 [WRN] BTCD: open /home/laolu32/.btcd/btcd.conf: no such file or directory
21:19:18 2016-03-29 [INF] BTCD: Version 0.12.0-beta
21:19:18 2016-03-29 [INF] BTCD: Loading block database from '/home/laolu32/.btcd/data/testnet/blocks_ffldb'
21:19:18 2016-03-29 [INF] BTCD: Block database loaded
21:19:18 2016-03-29 [INF] INDX: Transaction index is enabled
21:19:18 2016-03-29 [ERR] BTCD: Unable to start server on [:18333]: block 0000000000000000000000000000000000000000000000000000000000000000 does not exist
```
`btcd` exits after printing the above log message.

Since none of the indexes has been created yet, `dbFetchIndexerTip` returns a `zeroHash` for the block hash. This hash isn't found in the main chain by `chain.MainChainHasBlock`, so the loop then proceeds to fetch a block with a block hash of all zeroes, which obviously does not exist. To remedy, there should be a check here for equality against the `zeroHash`, if `hash` is equivalent, then `continue` since the current index has never been created. Any other way to detect an uninitialized index should suffice also.

davecgh: Yep, just noticed that myself and pushed a fix for it.  Thanks!

jrick:
Should the --droptxindex and --dropaddrindex flags cause the indexes to actually be dropped if another one of the --txindex or --addrindex flags are enabled?  Currently they do, and then the program terminates.

jrick:
This seems unrelated to the indexes changes from this pr (just easier to hit), but when the program is quit early via SIGINT during a --txindex catchup, it attempts to start the RPC server but fails because the database is closed, and prints the error:

```
$ btcd --txindex
10:05:28 2016-04-01 [INF] BTCD: Version 0.12.0-beta
10:05:28 2016-04-01 [WRN] BTCD: WARNING: There are multiple block chain databases using different database types.
You probably don't want to waste disk space by having more than one.
Your current database is located at [/home/jrick/.btcd/data/mainnet/blocks_ffldb].
The additional database is located at [/home/jrick/.btcd/data/mainnet/blocks_leveldb]
10:05:28 2016-04-01 [INF] BTCD: Loading block database from '/home/jrick/.btcd/data/mainnet/blocks_ffldb'
10:05:28 2016-04-01 [INF] BTCD: Block database loaded
10:05:28 2016-04-01 [INF] INDX: Transaction index is enabled
10:05:28 2016-04-01 [INF] INDX: Catching up indexes from height -1 to 205538
10:05:38 2016-04-01 [INF] INDX: Indexed 93063 blocks in the last 10s (180159 transactions, height 93062, 2010-11-20 22:14:52 -0500 EST)
^C10:05:40 2016-04-01 [INF] BTCD: Received SIGINT (Ctrl+C).  Shutting down...
10:05:40 2016-04-01 [INF] BTCD: Gracefully shutting down the database...
10:05:42 2016-04-01 [ERR] BTCD: Unable to start server on [:8333]: database is not open
$
```

jrick:
`10:20:22 2016-04-01 [INF] INDX: Indexed 23199 blocks in the last 10s (102642 transactions, height 95943, 2010-12-06 06:12:46 -0500 EST)`

When addrindex is enabled, should this also print the number of addresses processed?

jrick:
Would it be possible to keep the RPC server running during the index catchup?  It's not disabled during IBD, so disabling it during the catchup state seems a bit weird.

davecgh:
@jrick: The code currently disallows the mix of `--dropX` and `-X` where X is the same index, but you're correct that it doesn't stop you from doing `-dropY` and `-X`.  In my opinion, the behavior is accurate in that if you specify one of more `--drop`, it will drop those indexes and exit regardless of any other flags.  I suppose an additional check to ensure there are no `--dropY` + `-X` combinations wouldn't hurt, but then it raises the question of every other flag too.  For example, I can do `./btcd --debuglevel=show --dropX` and the one that happens will be whichever the first one is that is processed.  I have no real objection to adding additional checks to error when a mix is specified even though the resulting behavior will be the same.

Regarding the failure to start when you stop it early, that is intentional.  When you issue a `SIGINT`, it closes the database (which of course has a mutex so it finishes any outstanding database transactions before closing).  The closing of the database is what triggers everything else to shutdown.

Not a bad idea, however, it would extremely expensive to log that right now.  The logging happens outside of the indexers (i.e there is a loop which passes the block to each indexer via an interface so each indexer does its own thing), so in order to count the number of addresses, it would require them all to be extracted outside of the index again (which is really expensive).  One option would be to have each index handle its own logging, but then you'd up with one info statement per 10 seconds per index.  If that's acceptable, it could be changed.

Regarding getting the RPC server running, ultimately I *very much* want btcd to be able to start the RPC server before anything else, but unfortunately, it can't do that currently because the RPC server relies on everything else (the server, the block manager, the blockchain instance, the mempool, etc) and several of those pieces require the indexes to be caught up before they are started.

davecgh:
@jrick: After thinking some more about the `--dropaddrindex` + `--txindex` case, I believe it's better the way it current is.  My reasoning is that users will likely have `txindex=1` and `addrindex=1` in their config files.

If they want to drop the address index, the current code will throw an error that `--addrindex` and `--dropaddrindex` can't be mixed, as expected.  So, they will remove the `addrindex=1` entry from their config file.  However, I think it's not very intuitive to have to remove the `txindex=1` entry as well in order to drop the address index.

jrick:
Yeah.. I think this just proves my point that these really should be subcommands rather than config options.  It doesn't make sense to have both --addrindex and --droptxindex flags at the same time, so make that impossible by removing the drop flags and making them commands.

davecgh:
That seems like a reasonable assessment.  I wouldn't have any issue with converting these type of things to subcommands as I agree it makes sense.  However, I'd prefer that be in another PR since it's an infrastructure we don't really have in place.

jrick:
Yep agree with that.  not a showstopper on this PR.  We already have flags that change the program behavior (disable the daemon mode) similar to the new drop flags introduced here (like --version, --debuglevel=show, etc.).

davecgh:
Rebased to latest master (on to of blockchain_database2 branch).

davecgh:
Rebased to latest master (on to of blockchain_database2 branch).

jrick:
Been running this for a couple weeks now without any issues.

ok

PR #652, opened 2016-04-02
server: Drain relay notification channel.

davecgh:
A new channel to handle inventory relaying asynchronous to the mempool was added when the `mempoolConfig` struct was introduced in commit 83bcfea271da8b0c890ede7ef374c5a74ec5aad7.

However, the new channel was not also added to the cleanup code which drains the channels and, as a result, it led to the possibility for the server to hang during shutdown.

This pull request resolves that issue.

---
This change is [Reviewable](https://reviewable.io/reviews/btcsuite/btcd/652)

### DIFF
--- a/server.go
+++ b/server.go
@@ -2018,6 +2018,7 @@ out:
 cleanup:
 	for {
 		select {
+		case <-s.relayNtfnChan:
 		case <-s.modifyRebroadcastInv:
 		default:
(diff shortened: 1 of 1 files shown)


davecgh:
I realized this is already fixed by another PR that does a better job of it anyways.  Closing.

Issue #564, opened 2015-11-23
[connmgr] Proposal to refactor connection logic into its own package


davecgh:
The current code which handles establishing and maintaining a specific number of connections to other peers is tightly integrated into the server code.  While this is not problematic in and of itself, I would like to propose it be refactored into a separate sub-package named `connmgr` just like `wire`, `peer`, `txscript`, `addrmgr`, etc.

To be a bit more concrete, the package would handle all of the general connection concerns such as maintaining a set number of outbound connections, sourcing peers from seeds and the `addrmgr` package, banning, limiting max connections, tor, etc.

Ideally the refactor should be generic enough to allow the caller the flexibility to create their own types which can close over the peer state rather than having the package simply return a `peer.Peer`.  For more details on this point, see the `serverPeer` type in the current code.  This flexibility would allow the code to be used both by btcd proper and the SPV wallet work which is currently under development could ultimately make use of it, and it could also be useful in other scenarios.

In order to accomplish this, I think it would make sense to simply have the connection manger establish the connections as needed according to the configuration options and notify the caller (probably via a channel) with the established `net.Conn`.  The caller would then be responsible for creating a `peer.Peer` (likely with an enclosing state), setting its connection to the provided one, and notifying the connection manager when it goes away.

This would have a few benefits such as the ability to create mock connections in the testing framework that are independent of the Bitcoin-specific peer logic to ensure it works properly and help prevent regressions, allow the core server logic to be simplified down to only dealing with bitcoin-specific details, and, as previously mentioned, allow this fairly commonly needed functionality to be used externally.

Obviously I'm just spitballing here and the details will undoubtedly have to be tweaked once the actual work begins, but I think it's worth exploring.

tuxcanfly:
Thanks, yeah this would help decouple the connection logic and remove a lot of duplicate code while implementing SPV wallet. Let me me take a dig at this and come up with a draft.

jongillham:
@tuxcanfly are you still working on this?

tuxcanfly:
@jongillham Yeah, I was playing around with some ideas on a local branch. I'll get that cleaned up and push it for review.

davecgh:
This has been implemented and merged via PR https://github.com/btcsuite/btcd/pull/653.

PR #653, opened 2016-04-04
connmgr: Add pkg and refactor

tuxcanfly:
Added pkg `connmgr` and moved common connection related functionalities into it. Working on refactoring all the functionalities mentioned in #564 and fixing related bugs.

### DIFF
--- a/config.go
+++ b/config.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/btcsuite/btcd/connmgr"
 	"github.com/btcsuite/btcd/database"
 	_ "github.com/btcsuite/btcd/database/ffldb"
 	"github.com/btcsuite/btcd/mempool"
@@ -859,7 +860,7 @@ func loadConfig() (*config, []string, error) {
 		cfg.dial = proxy.Dial
 		if !cfg.NoOnion {
 			cfg.lookup = func(host string) ([]net.IP, error) {
-				return torLookupIP(host, cfg.Proxy)
+				return connmgr.TorLookupIP(host, cfg.Proxy)
 			}
 		}
 	}
@@ -899,7 +900,7 @@ func loadConfig() (*config, []string, error) {
 			return proxy.Dial(a, b)
 		}
 		cfg.onionlookup = func(host string) ([]net.IP, error) {
-			return torLookupIP(host, cfg.OnionProxy)
+			return connmgr.TorLookupIP(host, cfg.OnionProxy)
 		}
 	} else {
 		cfg.oniondial = cfg.dial
--- a/connmgr/connmanager.go
+++ b/connmgr/connmanager.go
@@ -1,0 +1,371 @@
+// Copyright (c) 2016 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package connmgr
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// maxFailedAttempts is the maximum number of successive failed connection
+// attempts after which network failure is assumed and new connections will
+// be delayed by the configured retry duration.
+const maxFailedAttempts = 25
+
+var (
+	//ErrDialNil is used to indicate that Dial cannot be nil in the configuration.
+	ErrDialNil = errors.New("Config: Dial cannot be nil")
+
+	// maxRetryDuration is the max duration of time retrying of a persistent
+	// connection is allowed to grow to.  This is necessary since the retry
+	// logic uses a backoff mechanism which increases the interval base times
+	// the number of retries that have been done.
+	maxRetryDuration = time.Minute * 5
+
+	// defaultRetryDuration is the default duration of time for retrying
+	// persistent connections.
+	defaultRetryDuration = time.Second * 5
+
+	// defaultMaxOutbound is the default number of maximum outbound connections
+	// to maintain.
+	defaultMaxOutbound = uint32(8)
+)
+
+// DialFunc defines a function that dials a connection.
+type DialFunc func(string, string) (net.Conn, error)
+
+// AddressFunc defines a function that returns a network address to connect to.
+type AddressFunc func() (string, error)
+
+// ConnState represents the state of the requested connection.
+type ConnState uint8
+
+// ConnState can be either pending, established, disconnected or failed.  When
+// a new connection is requested, it is attempted and categorized as
+// established or failed depending on the connection result.  An established
+// connection which was disconnected is categorized as disconnected.
+const (
+	ConnPending ConnState = iota
+	ConnEstablished
+	ConnDisconnected
+	ConnFailed
+)
+
+// OnConnectionFunc is the signature of the callback function which is used to
+// subscribe to new connections.
+type OnConnectionFunc func(*ConnReq, net.Conn)
+
+// OnDisconnectionFunc is the signature of the callback function which is used to
+// notify disconnections.
+type OnDisconnectionFunc func(*ConnReq)
+
+// ConnReq is the connection request to a network address. If permanent, the
+// connection will be retried on disconnection.
+type ConnReq struct {
+	// The following variables must only be used atomically.
+	id uint64
+
+	Addr      string
+	Permanent bool
+
+	conn       net.Conn
+	state      ConnState
+	stateMtx   sync.RWMutex
+	retryCount uint32
+}
+
+// updateState updates the state of the connection request.
+func (c *ConnReq) updateState(state ConnState) {
+	c.stateMtx.Lock()
+	c.state = state
+	c.stateMtx.Unlock()
+}
+
+// ID returns a unique identifier for the connection request.
+func (c *ConnReq) ID() uint64 {
+	return atomic.LoadUint64(&c.id)
+}
+
+// State is the connection state of the requested connection.
+func (c *ConnReq) State() ConnState {
+	c.stateMtx.RLock()
+	state := c.state
+	c.stateMtx.RUnlock()
+	return state
+}
+
+// String returns a human-readable string for the connection request.
+func (c *ConnReq) String() string {
+	if c.Addr == "" {
+		return fmt.Sprintf("reqid %d", atomic.LoadUint64(&c.id))
+	}
+	return fmt.Sprintf("%s (reqid %d)", c.Addr, atomic.LoadUint64(&c.id))
+}
+
+// Config holds the configuration options related to the connection manager.
+type Config struct {
+	// MaxOutbound is the maximum number of outbound network connections to
+	// maintain. Defaults to 8.
+	MaxOutbound uint32
+
+	// RetryDuration is the duration to wait before retrying connection
+	// requests. Defaults to 5s.
+	RetryDuration time.Duration
+
+	// OnConnection is a callback that is fired when a new connection is
+	// established.
+	OnConnection OnConnectionFunc
+
+	// OnDisconnection is a callback that is fired when a connection is
+	// disconnected.
+	OnDisconnection OnDisconnectionFunc
+
+	// GetNewAddress is a way to get an address to make a network connection
+	// to.  If nil, no new connections will be made automatically.
+	GetNewAddress AddressFunc
+
+	// Dial connects to the address on the named network. It cannot be nil.
+	Dial DialFunc
+}
+
+// handleConnected is used to queue a successful connection.
+type handleConnected struct {
+	c    *ConnReq
+	conn net.Conn
+}
+
+// handleDisconnected is used to remove a connection.
+type handleDisconnected struct {
+	id    uint64
+	retry bool
+}
+
+// handleFailed is used to remove a pending connection.
+type handleFailed struct {
+	c   *ConnReq
+	err error
+}
+
+// ConnManager provides a manager to handle network connections.
+type ConnManager struct {
+	// The following variables must only be used atomically.
+	connReqCount uint64
+	start        int32
+	stop         int32
+
+	cfg            Config
+	wg             sync.WaitGroup
+	failedAttempts uint64
+	requests       chan interface{}
+	quit           chan struct{}
+}
+
+// handleFailedConn handles a connection failed due to a disconnect or any
+// other failure. If permanent, it retries the connection after the configured
+// retry duration. Otherwise, if required, it makes a new connection request.
+// After maxFailedConnectionAttempts new connections will be retried after the
+// configured retry duration.
+func (cm *ConnManager) handleFailedConn(c *ConnReq, retry bool) {
+	if atomic.LoadInt32(&cm.stop) != 0 {
+		return
+	}
+	if retry && c.Permanent {
+		c.retryCount++
+		d := time.Duration(c.retryCount) * cm.cfg.RetryDuration
+		if d > maxRetryDuration {
+			d = maxRetryDuration
+		}
+		log.Debugf("Retrying connection to %v in %v", c, d)
+		time.AfterFunc(d, func() {
+			cm.Connect(c)
+		})
+	} else if cm.cfg.GetNewAddress != nil {
+		cm.failedAttempts++
+		if cm.failedAttempts >= maxFailedAttempts {
+			log.Debugf("Max failed connection attempts reached: [%d] "+
+				"-- retrying connection in: %v", maxFailedAttempts,
+				cm.cfg.RetryDuration)
+			time.AfterFunc(cm.cfg.RetryDuration, func() {
+				cm.NewConnReq()
+			})
+		} else {
+			go cm.NewConnReq()
+		}
+	}
+}
+
+// connHandler handles all connection related requests.  It must be run as a
+// goroutine.
+//
+// The connection handler makes sure that we maintain a pool of active outbound
+// connections so that we remain connected to the network.  Connection requests
+// are processed and mapped by their assigned ids.
+func (cm *ConnManager) connHandler() {
+	conns := make(map[uint64]*ConnReq, cm.cfg.MaxOutbound)
+out:
+	for {
+		select {
+		case req := <-cm.requests:
+			switch msg := req.(type) {
+
+			case handleConnected:
+				connReq := msg.c
+				connReq.updateState(ConnEstablished)
+				connReq.conn = msg.conn
+				conns[connReq.id] = connReq
+				log.Debugf("Connected to %v", connReq)
+				connReq.retryCount = 0
+				cm.failedAttempts = 0
+
+				if cm.cfg.OnConnection != nil {
+					go cm.cfg.OnConnection(connReq, msg.conn)
+				}
+
+			case handleDisconnected:
+				if connReq, ok := conns[msg.id]; ok {
+					connReq.updateState(ConnDisconnected)
+					if connReq.conn != nil {
+						connReq.conn.Close()
+					}
+					log.Debugf("Disconnected from %v", connReq)
+					delete(conns, msg.id)
+
+					if cm.cfg.OnDisconnection != nil {
+						go cm.cfg.OnDisconnection(connReq)
+					}
+
+					cm.handleFailedConn(connReq, msg.retry)
+				} else {
+					log.Errorf("Unknown connection: %d", msg.id)
+				}
+
+			case handleFailed:
+				connReq := msg.c
+				connReq.updateState(ConnFailed)
+				log.Debugf("Failed to connect to %v: %v", connReq, msg.err)
+				cm.handleFailedConn(connReq, true)
+			}
+
+		case <-cm.quit:
+			break out
+		}
+	}
+
+	cm.wg.Done()
+	log.Trace("Connection handler done")
+}
+
+// NewConnReq creates a new connection request and connects to the
+// corresponding address.
+func (cm *ConnManager) NewConnReq() {
+	if atomic.LoadInt32(&cm.stop) != 0 {
+		return
+	}
+	if cm.cfg.GetNewAddress == nil {
+		return
+	}
+	c := &ConnReq{}
+	atomic.StoreUint64(&c.id, atomic.AddUint64(&cm.connReqCount, 1))
+	addr, err := cm.cfg.GetNewAddress()
+	if err != nil {
+		cm.requests <- handleFailed{c, err}
+		return
+	}
+	c.Addr = addr
+	cm.Connect(c)
+}
+
+// Connect assigns an id and dials a connection to the address of the
+// connection request.
+func (cm *ConnManager) Connect(c *ConnReq) {
+	if atomic.LoadInt32(&cm.stop) != 0 {
+		return
+	}
+	if atomic.LoadUint64(&c.id) == 0 {
+		atomic.StoreUint64(&c.id, atomic.AddUint64(&cm.connReqCount, 1))
+	}
+	log.Debugf("Attempting to connect to %v", c)
+	conn, err := cm.cfg.Dial("tcp", c.Addr)
+	if err != nil {
+		cm.requests <- handleFailed{c, err}
+	} else {
+		cm.requests <- handleConnected{c, conn}
+	}
+}
+
+// Disconnect disconnects the connection corresponding to the given connection
+// id. If permanent, the connection will be retried with an increasing backoff
+// duration.
+func (cm *ConnManager) Disconnect(id uint64) {
+	if atomic.LoadInt32(&cm.stop) != 0 {
+		return
+	}
+	cm.requests <- handleDisconnected{id, true}
+}
+
+// Remove removes the connection corresponding to the given connection
+// id from known connections.
+func (cm *ConnManager) Remove(id uint64) {
+	if atomic.LoadInt32(&cm.stop) != 0 {
+		return
+	}
+	cm.requests <- handleDisconnected{id, false}
+}
+
+// Start launches the connection manager and begins connecting to the network.
+func (cm *ConnManager) Start() {
+	// Already started?
+	if atomic.AddInt32(&cm.start, 1) != 1 {
+		return
+	}
+
+	log.Trace("Connection manager started")
+	cm.wg.Add(1)
+	go cm.connHandler()
+
+	for i := atomic.LoadUint64(&cm.connReqCount); i < uint64(cm.cfg.MaxOutbound); i++ {
+		go cm.NewConnReq()
+	}
+}
+
+// Wait blocks until the connection manager halts gracefully.
+func (cm *ConnManager) Wait() {
+	cm.wg.Wait()
+}
+
+// Stop gracefully shuts down the connection manager.
+func (cm *ConnManager) Stop() {
+	if atomic.AddInt32(&cm.stop, 1) != 1 {
+		log.Warnf("Connection manager already stopped")
+		return
+	}
+	close(cm.quit)
+	log.Trace("Connection manager stopped")
+}
+
+// New returns a new connection manager.
+// Use Start to start connecting to the network.
+func New(cfg *Config) (*ConnManager, error) {
+	if cfg.Dial == nil {
+		return nil, ErrDialNil
+	}
+	// Default to sane values
+	if cfg.RetryDuration <= 0 {
+		cfg.RetryDuration = defaultRetryDuration
+	}
+	if cfg.MaxOutbound == 0 {
+		cfg.MaxOutbound = defaultMaxOutbound
+	}
+	cm := ConnManager{
+		cfg:      *cfg, // Copy so caller can't mutate
+		requests: make(chan interface{}),
+		quit:     make(chan struct{}),
+	}
+	return &cm, nil
+}
--- a/connmgr/connmanager_test.go
+++ b/connmgr/connmanager_test.go
@@ -1,0 +1,389 @@
+// Copyright (c) 2016 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package connmgr
+
+import (
+	"bytes"
+	"errors"
+	"io"
+	"net"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/btcsuite/btclog"
+)
+
+func init() {
+	// Override the max retry duration when running tests.
+	maxRetryDuration = 2 * time.Millisecond
+}
+
+// mockAddr mocks a network address
+type mockAddr struct {
+	net, address string
+}
+
+func (m mockAddr) Network() string { return m.net }
+func (m mockAddr) String() string  { return m.address }
+
+// mockConn mocks a network connection by implementing the net.Conn interface.
+type mockConn struct {
+	io.Reader
+	io.Writer
+	io.Closer
+
+	// local network, address for the connection.
+	lnet, laddr string
+
+	// remote network, address for the connection.
+	rnet, raddr string
+}
+
+// LocalAddr returns the local address for the connection.
+func (c mockConn) LocalAddr() net.Addr {
+	return &mockAddr{c.lnet, c.laddr}
+}
+
+// RemoteAddr returns the remote address for the connection.
+func (c mockConn) RemoteAddr() net.Addr {
+	return &mockAddr{c.rnet, c.raddr}
+}
+
+// Close handles closing the connection.
+func (c mockConn) Close() error {
+	return nil
+}
+
+func (c mockConn) SetDeadline(t time.Time) error      { return nil }
+func (c mockConn) SetReadDeadline(t time.Time) error  { return nil }
+func (c mockConn) SetWriteDeadline(t time.Time) error { return nil }
+
+// mockDialer mocks the net.Dial interface by returning a mock connection to
+// the given address.
+func mockDialer(network, address string) (net.Conn, error) {
+	r, w := io.Pipe()
+	c := &mockConn{raddr: address}
+	c.Reader = r
+	c.Writer = w
+	return c, nil
+}
+
+// TestNewConfig tests that new ConnManager config is validated as expected.
+func TestNewConfig(t *testing.T) {
+	_, err := New(&Config{})
+	if err == nil {
+		t.Fatalf("New expected error: 'Dial can't be nil', got nil")
+	}
+	_, err = New(&Config{
+		Dial: mockDialer,
+	})
+	if err != nil {
+		t.Fatalf("New unexpected error: %v", err)
+	}
+}
+
+// TestUseLogger tests that a logger can be passed to UseLogger
+func TestUseLogger(t *testing.T) {
+	l, err := btclog.NewLoggerFromWriter(bytes.NewBuffer(nil), btclog.InfoLvl)
+	if err != nil {
+		t.Fatal(err)
+	}
+	UseLogger(l)
+}
+
+// TestStartStop tests that the connection manager starts and stops as
+// expected.
+func TestStartStop(t *testing.T) {
+	connected := make(chan *ConnReq)
+	disconnected := make(chan *ConnReq)
+	cmgr, err := New(&Config{
+		MaxOutbound:   1,
+		GetNewAddress: func() (string, error) { return "127.0.0.1:18555", nil },
+		Dial:          mockDialer,
+		OnConnection: func(c *ConnReq, conn net.Conn) {
+			connected <- c
+		},
+		OnDisconnection: func(c *ConnReq) {
+			disconnected <- c
+		},
+	})
+	if err != nil {
+		t.Fatalf("New error: %v", err)
+	}
+	cmgr.Start()
+	gotConnReq := <-connected
+	cmgr.Stop()
+	// already stopped
+	cmgr.Stop()
+	// ignored
+	cr := &ConnReq{Addr: "127.0.0.1:18555", Permanent: true}
+	cmgr.Connect(cr)
+	if cr.ID() != 0 {
+		t.Fatalf("start/stop: got id: %v, want: 0", cr.ID())
+	}
+	cmgr.Disconnect(gotConnReq.ID())
+	cmgr.Remove(gotConnReq.ID())
+	select {
+	case <-disconnected:
+		t.Fatalf("start/stop: unexpected disconnection")
+	case <-time.Tick(10 * time.Millisecond):
+		break
+	}
+}
+
+// TestConnectMode tests that the connection manager works in the connect mode.
+//
+// In connect mode, automatic connections are disabled, so we test that
+// requests using Connect are handled and that no other connections are made.
+func TestConnectMode(t *testing.T) {
+	connected := make(chan *ConnReq)
+	cmgr, err := New(&Config{
+		MaxOutbound: 2,
+		Dial:        mockDialer,
+		OnConnection: func(c *ConnReq, conn net.Conn) {
+			connected <- c
+		},
+	})
+	if err != nil {
+		t.Fatalf("New error: %v", err)
+	}
+	cr := &ConnReq{Addr: "127.0.0.1:18555", Permanent: true}
+	cmgr.Start()
+	cmgr.Connect(cr)
+	gotConnReq := <-connected
+	wantID := cr.ID()
+	gotID := gotConnReq.ID()
+	if gotID != wantID {
+		t.Fatalf("connect mode: %v - want ID %v, got ID %v", cr.Addr, wantID, gotID)
+	}
+	gotState := cr.State()
+	wantState := ConnEstablished
+	if gotState != wantState {
+		t.Fatalf("connect mode: %v - want state %v, got state %v", cr.Addr, wantState, gotState)
+	}
+	select {
+	case c := <-connected:
+		t.Fatalf("connect mode: got unexpected connection - %v", c.Addr)
+	case <-time.After(time.Millisecond):
+		break
+	}
+	cmgr.Stop()
+}
+
+// TestMaxOutbound tests the maximum number of outbound connections.
+//
+// We wait until all connections are established, then test they there are the
+// only connections made.
+func TestMaxOutbound(t *testing.T) {
+	maxOutbound := uint32(10)
+	connected := make(chan *ConnReq)
+	cmgr, err := New(&Config{
+		MaxOutbound:   maxOutbound,
+		Dial:          mockDialer,
+		GetNewAddress: func() (string, error) { return "127.0.0.1:18555", nil },
+		OnConnection: func(c *ConnReq, conn net.Conn) {
+			connected <- c
+		},
+	})
+	if err != nil {
+		t.Fatalf("New error: %v", err)
+	}
+	cmgr.Start()
+	for i := uint32(0); i < maxOutbound; i++ {
+		<-connected
+	}
+
+	select {
+	case c := <-connected:
+		t.Fatalf("max outbound: got unexpected connection - %v", c.Addr)
+	case <-time.After(time.Millisecond):
+		break
+	}
+	cmgr.Stop()
+}
+
+// TestRetryPermanent tests that permanent connection requests are retried.
+//
+// We make a permanent connection request using Connect, disconnect it using
+// Disconnect and we wait for it to be connected back.
+func TestRetryPermanent(t *testing.T) {
+	connected := make(chan *ConnReq)
+	disconnected := make(chan *ConnReq)
+	cmgr, err := New(&Config{
+		RetryDuration: time.Millisecond,
+		MaxOutbound:   1,
+		Dial:          mockDialer,
+		OnConnection: func(c *ConnReq, conn net.Conn) {
+			connected <- c
+		},
+		OnDisconnection: func(c *ConnReq) {
+			disconnected <- c
+		},
+	})
+	if err != nil {
+		t.Fatalf("New error: %v", err)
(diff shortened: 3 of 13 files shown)


rpcserver.go
```diff
@@ -2382,7 +2383,6 @@ func handleGetPeerInfo(s *rpcServer, cmd interface{}, closeChan <-chan struct{})
 			Inbound:        statsSnap.Inbound,
 			StartingHeight: statsSnap.StartingHeight,
 			CurrentHeight:  statsSnap.LastBlock,
-			BanScore:       int32(p.banScore.Int()),
```
davecgh: This shouldn't be removed.

connmgr/connmanager.go
```diff
@@ -0,0 +1,229 @@
...
+	"sync"
+	"time"
+
+	"github.com/btcsuite/btcd/addrmgr"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/wire"
+)
+
+var (
+	//MaxOutboundPeers is the number of max outbound connections.
```
davecgh: Space after comment for consistency

davecgh:
Looks like the banning code has all been removed and shouldn't have been.

config.go
```diff
@@ -855,31 +856,20 @@ func loadConfig() (*config, []string, error) {
...
-// btcdDial connects to the address on the named network using the appropriate
-// dial function depending on the address and configuration options.  For
-// example, .onion addresses will be dialed using the onion specific proxy if
-// one was specified, but will otherwise use the normal dial function (which
-// could itself use a proxy or not).
-func btcdDial(network, address string) (net.Conn, error) {
-	if strings.Contains(address, ".onion:") {
-		return cfg.oniondial(network, address)
+	// Provide connection manager config
+	connmgr.Dial = func(network, address string) (net.Conn, error) {
```
davecgh: Setting globals like this is super racy.

davecgh: Well on second look, I guess they're only be set up startup and never changed later.  Nevertheless, the package really shouldn't be using globals like this.  The code should be able to get an instance of a connection manager which houses all state.

Using globals in packages is almost always bad.

connmgr/connmanager.go
```diff
@@ -0,0 +1,229 @@
...
+	MaxOutboundPeers = 8
```
davecgh: All of these need to be part of a `Config` struct that is provided when creating a new connmgr instance instead of globals.

connmgr/connmanager.go
```diff
@@ -0,0 +1,229 @@
...
+
+// connectionHandler is a service that monitors requests for new
+// connections or closed connections and maps addresses to their
+// respective connections.
+func (cm *ConnManager) connectionHandler() {
+out:
+	for {
+		select {
+		case cr := <-cm.newConnections:
+			cm.connections[cr.Addr] = cr.Conn
```
davecgh: There can be multiple connections from the same address.  This doesn't seem very robust to me.

connmgr/connmanager.go
```diff
@@ -0,0 +1,229 @@
...
+
+		case addr := <-cm.closeConnections:
+			c := cm.connections[addr]
```
davecgh: Accessing fields on the connection manager form multiple goroutines without a mutex will definitely race.

connmgr/connmanager.go
```diff
@@ -0,0 +1,229 @@
...
+// WaitForShutdown blocks until the connection handlers are stopped.
+func (cm *ConnManager) WaitForShutdown() {
+	cm.wg.Wait()
+}
+
+// Stop gracefully shuts down the connection manager.
+func (cm *ConnManager) Stop() {
+	log.Warnf("Connection Manager shutting down")
+	close(cm.quit)
+	return
```
davecgh: There is no return value, so no need for this statement.

connmgr/connmanager.go
```diff
@@ -0,0 +1,229 @@
...
+		case <-cm.quit:
+			break out
+		}
+	}
+	cm.wg.Done()
+}
+
+// Start launches the connection manager.
+func (cm *ConnManager) Start() {
+	cm.seedFromDNS(ChainParams.DNSSeeds)
```
davecgh: As mentioned elsewhere, the `ChainParams` should come from a `Config` on the `ConnManager` the same as other packages handle it.

Also, I don't see any reason to pass this variable to the `cm.seedFromDNS` function since it has access to the `ConnManager` state and it's not called from anywhere else with different values.

connmgr/connmanager.go
```diff
@@ -0,0 +1,229 @@
...
+	amgr           *addrmgr.AddrManager
+	connections    map[string]net.Conn
+	outboundGroups map[string]int
+
+	NewConnectionHandler func(<-chan *ConnResult)
+}
+
+// seedFromDNS uses DNS seeding to populate the address manager with peers.
+func (cm *ConnManager) seedFromDNS(DNSSeeds []string) {
+	for _, seeder := range DNSSeeds {
```
davecgh: Once the `ConnManager` has a `Config` instance set in the state as mentioned elsewhere, this would just become:

```Go
for _, seeder := range cm.cfg.DNSSeeds {
...
}
```

davecgh:
We need to examine the goroutines carefully.  It seems like this is launching way more goroutines than it needs to be for every connection.

davecgh:
I know you said this isn't ready for review yet, so I'm not getting into the details, but I did want to point out those overall architectural issues so you can keep them in mind as you update the PR.

tuxcanfly:
Thanks, will update the PR and keep the points mentioned in mind.

tuxcanfly:
Refactored to address the review items, notably using a `Config` struct, connection ID instead of `addr` as the connections map key. Working on the following TODO items in issue:

- [ ] Move ban, retry, group logic to `connmgr`
- [ ] Simplify `server.go` peer management (goroutines etc)
- [ ] Add `connmgr` tests

tuxcanfly:
Refactored to simplify the connection manager so that it now only manages connections to the required fixed number of them. The package will be integrated once the core components are in place, so I have restored the earlier changes to `server.go` et al.

tuxcanfly:
The core connection handler is now in place. I've started integrating the pkg and from initial tests it works well. I have removed some of the server code which is now unnecessary, to simplify some of the connection related bits.

connmgr/connmanager.go
```diff
@@ -0,0 +1,235 @@
...
+
+// ConnState represents the state of the requested connection.
+type ConnState uint8
+
+// ConnState can be either pending, established, disconnected or failed.  When
+// a new connection is requested, it is attempted and categorized as
+// established or failed depending on the connection result.  An established
+// connection which was disconnected is categorized as disconnected.
+const (
+	ConnPending      ConnState = 0x0
```
davecgh: I'd suggest using iota here:
```Go
ConnPending      ConnState = iota
ConnEstablished
ConnDisconnected
ConnFailed
```

connmgr/connmanager.go
```diff
@@ -0,0 +1,235 @@
...
+	Addr      string
+	Permanent bool
+
+	state      ConnState
+	retryCount uint32
+}
+
+// State is the connection state of the requested connection.
+func (c *ConnReq) State() ConnState {
+	return c.state
```
davecgh: This will race.  It can access the state variable from any goroutine and it's set in the `connHandler` goroutine below.

connmgr/connmanager.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+	close(cm.quit)
+	log.Trace("Connection manager stopped")
+}
+
+// New returns a new connection manager.
+// Use Start to start connecting to the network.
+func New(cfg *Config) (*ConnManager, error) {
+	cm := ConnManager{
+		cfg:      cfg,
```
davecgh: I'd suggest making the `ConnManager` field a concrete type and making this a copy like `peer` does [here](https://github.com/btcsuite/btcd/blob/master/peer/peer.go#L2062).  This will prevent the caller from being able to mutate it after the fact.

So:
```Go
		cfg:      *cfg,
```

connmgr/connmanager.go
```diff
@@ -0,0 +1,235 @@
...
+
+// handleFailed is used to remove a pending connection.
+type handleFailed struct {
+	c   *ConnReq
+	err error
+}
+
+// ConnManager provides a manager to handle network connections.
+type ConnManager struct {
+	cfg      *Config
```
davecgh: As mentioned below, this would become:
```Go
	cfg      Config
```

connmgr/connmanager.go
```diff
@@ -0,0 +1,235 @@
...
+	wg       sync.WaitGroup
+	conns    map[*ConnReq]net.Conn
```
davecgh: Using points in a map is dangerous because it's too easy to end up with a pointer to a copy of the object which is otherwise functionally the same.  I would suggest giving a `ConnReq` a unique uint64 id (probably atomically incremented on the `ConnManager` instance itself) and keying the map by uint64.

connmgr/connmanager.go
```diff
@@ -0,0 +1,235 @@
...
+	cm.requests <- handleDisconnected{c}
+}
+
+// Start launches the connection manager and begins connecting to the network.
+func (cm *ConnManager) Start() {
+	log.Trace("Connection manager started")
+	cm.wg.Add(1)
+	go cm.connHandler()
+
+	for i := uint32(0); i < cm.cfg.MaxOutboundConnections; i++ {
```
davecgh: This appears to be ignoring configured persistent peers which would reduce this number.

connmgr/connmanager.go
davecgh: You'd either need to put a mutex around it or create a couple of new channels to have the `connHandler` goroutine return the state.

connmgr/connmanager.go
```diff
@@ -0,0 +1,235 @@
...
+// are processed and mapped to their corresponding network connections.
+func (cm *ConnManager) connHandler() {
+out:
+	for {
+		select {
+		case req := <-cm.requests:
+			switch msg := req.(type) {
+
+			case handleConnected:
+				msg.c.state = ConnEstablished
```
davecgh: And then this of course would need to grab the write lock:

```Go
cr := msg.c
cr.stateMtx.Lock()
cr.state = ConnEstablished
cr.Unlock()
...
```

connmgr/connmanager.go
```diff
@@ -0,0 +1,235 @@
...
+				cm.conns[msg.c] = msg.conn
+				log.Infof("Connection established: %v", msg.c.Addr)
+				msg.c.retryCount = 0
+
+				if cm.cfg.OnConnection != nil {
+					go cm.cfg.OnConnection(msg.c, msg.conn)
+				}
+
+			case handleDisconnected:
+				msg.c.state = ConnDisconnected
```
davecgh: Same here.

connmgr/connmanager.go
```diff
@@ -0,0 +1,235 @@
...
+				if conn, ok := cm.conns[msg.c]; ok {
+					conn.Close()
+				}
+				delete(cm.conns, msg.c)
+				log.Debugf("Connection closed: %v", msg.c.Addr)
+
+				cm.handleFailedConn(msg.c)
+
+			case handleFailed:
+				msg.c.state = ConnFailed
```
davecgh: Same here.

connmgr/connmanager.go
```diff
@@ -0,0 +1,235 @@
...
+		d := time.Duration(c.retryCount) * cm.cfg.RetryDuration
+		if d > maxRetryDuration {
+			d = maxRetryDuration
+		}
+		log.Debugf("Retrying connection to %v in: %v", c.Addr, d)
+		time.AfterFunc(d, func() {
+			cm.Connect(c)
+		})
+	} else if uint32(len(cm.conns)) < cm.cfg.MaxOutboundConnections {
+		go cm.Connect(&ConnReq{Addr: cm.cfg.GetNewAddress()})
```
davecgh: This looks to me like it will have a degenerate condition when the network adapter goes down (such as a wireless card turned off).  It will launch a new goroutine which will fail immediately due to the adapter being down and turn send a message back to the `connHandler` that it failed which will call this function and launch another one, etc.  

This would happen incredibly fast and almost positively outpace the goroutine schedule such that it would balloon the number of goroutines and goes memory usage to go nuts.

Have you tried this scenario?

connmgr/connmanager.go
davecgh: Something to keep in mind however is that there is also the case when there are > `cm.cfg.MaxOutboundConnections` persistent peers which all are not able to connect and so considering them could lead to a situation where there would be no connections at all until (or if) one or more of those persistent peer remotes came back online.

connmgr/connmanager_test.go
```diff
@@ -0,0 +1,83 @@
...
+// given address.
+func mockDial(network, address string) (net.Conn, error) {
+	r, w := io.Pipe()
+	c := &mockConn{raddr: address}
+	c.Reader = r
+	c.Writer = w
+	return c, nil
+}
+
+func TestStartStop(t *testing.T) {
```
davecgh: Missing comment.

davecgh:
Needs a `README.md`.

server.go
```diff
@@ -1198,13 +1180,11 @@ func (s *server) handleAddPeerMsg(state *peerState, sp *serverPeer) bool {
...
 
 	// Limit max outbound peers.
-	if _, ok := state.pendingPeers[sp.Addr()]; ok {
-		if state.OutboundCount() >= state.maxOutboundPeers {
-			srvrLog.Infof("Max outbound peers reached [%d] - disconnecting "+
-				"peer %s", state.maxOutboundPeers, sp)
-			sp.Disconnect()
-			return false
-		}
+	if state.OutboundCount() >= state.maxOutboundPeers {
```
davecgh: Shouldn't this be handled by the connection manager?  The point of the connection manager is the server doesn't have to worry about tracking this type of stuff.

davecgh:
Perhaps I'm missing it, but this new code doesn't seem to properly handle the `--connect` option which says to only connect to the specified peers and no others.

connmgr/connmanager.go
tuxcanfly: I did try this and found that number of goroutines and memory remained almost the same but let me double check that.

server.go
tuxcanfly: Yes this would be superflous now. Should have been removed  during the integration, I'll do that shortly.

tuxcanfly:
@davecgh Yeah missed the `--connect` case, I'll add a flag on the manager to handle it. Thanks for catching that.

tuxcanfly:
In case the caller doesn't pass `GetNewAddress`, the `connmgr` will run in `--connect` mode i.e it does not connect to new addresses automatically.

Checking for existing connection requests at startup by using a counter for them which also serves as the source of ID.

Added a counter for successive failed connections and using it to handle the degenerate case i.e network failure.

tuxcanfly:
Working on the `README`. Also still need to consider the case where persistent connections > `cm.cfg.MaxOutboundConnections`

tuxcanfly:
Added README with a general overview and some of the features of the pkg.

Also for the persistent case, we are launching temporary connections in addition to the persistent connections and which ever connect first end up in the connection manager. This has the benefit of choosing faster connections but the drawback is that it sometimes disregards the persistent connections especially if they are slower. One idea is to have maybe 50% of the max slots allocated to persistent connections and the rest to temporary ones but I'm not sure if that would be worth implementing.

tuxcanfly:
Trying out an idea where if a permanent connection is successful and we are already full, we replace a temporary connection, this way we make use of all slots and don't lose permanent connections as well.

tuxcanfly:
To keep it simple I've updated to only launch the number of connection requests required to maintain the `MaxOutbound` number of peers. So the past two comments can be ignored.

connmgr/connmanager.go
```diff
@@ -0,0 +1,321 @@
...
+	ConnFailed
+)
+
+// ConnReq is the connection request to a network address. If permanent, the
+// connection will be retried on disconnection.
+type ConnReq struct {
+	Addr      string
+	Permanent bool
+
+	id         uint64
```
davecgh: Since this is being used atomically, it needs to be at the top of the struct to ensure alignment.  Otherwise it will end up panicing on platforms such as arm.

tuxcanfly: Thanks, will move that.

connmgr/connmanager.go
```diff
@@ -0,0 +1,321 @@
...
+	c   *ConnReq
+	err error
+}
+
+// ConnManager provides a manager to handle network connections.
+type ConnManager struct {
+	cfg            Config
+	wg             sync.WaitGroup
+	conns          map[uint64]*ConnReq
+	connReqCount   uint64
```
davecgh: The same deal regarding all of the fields used with `sync/atomic`.  They all need to be at the top.  I'd follow the pattern used in the `peer` package.  Namely:

```Go
	// The following variables must only be used atomically.
	connReqCount   uint64
	stop           int32
```

connmgr/connmanager.go
```diff
@@ -0,0 +1,321 @@
...
+func (cm *ConnManager) connHandler() {
+out:
+	for {
+		select {
+		case req := <-cm.requests:
+			switch msg := req.(type) {
+
+			case handleConnected:
+				connReq := msg.c
+				connReq.stateMtx.Lock()
```
davecgh: Since changing the state requires the mutex and it's duplicated in several places, I would suggest an `UpdateState` function on the `connReq` type like so:

```Go
// UpdateState changes the connection state to the provided state.
func (c *ConnReq) UpdateState(state ConnState) {
	c.stateMtx.Lock()
	c.state = state
	c.stateMtx.Unlock()
}
```

Then, this would just be:

```Go
connReq.UpdateState(ConnEstablished)
```

The same thing goes for the other places the connection request state is updated.

davecgh: Actually, since `ConnReq` is exported, I'd suggest making that func unexported instead so callers can't mess with the state.

```Go
func (c *ConnReq) updateState(state ConnState) {
...
}
```

tuxcanfly: Yup, will do that.

connmgr/connmanager.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+	close(cm.quit)
+	log.Trace("Connection manager stopped")
+}
+
+// New returns a new connection manager.
+// Use Start to start connecting to the network.
+func New(cfg *Config) (*ConnManager, error) {
+	if cfg.Dial == nil {
+		return nil, errors.New("Config: Dial cannot be nil")
```
davecgh: Please make this an exported error.  I prefer all errors returned from a package to be programmatically distinguishable in general.

tuxcanfly: OK, will do.

connmgr/connmanager.go
```diff
@@ -0,0 +1,321 @@
...
+	// maxRetryDuration is the max amount of time retrying of a persistent
+	// connection is allowed to grow to.  This is necessary since the retry
+	// logic uses a backoff mechanism which increases the interval base done
+	// the number of retries that have been done.
+	maxRetryDuration = time.Minute * 5
+
+	// maxFailedAttempts is the maximum number of successive failed connection
+	// attempts after which network failure is assumed and new connections will
+	// be delayed by the configured retry duration.
+	maxFailedAttempts = uint64(10)
```
davecgh: This should be an untyped constant imo.

connmgr/connmanager.go
```diff
@@ -0,0 +1,321 @@
...
+
+// New returns a new connection manager.
+// Use Start to start connecting to the network.
+func New(cfg *Config) (*ConnManager, error) {
+	if cfg.Dial == nil {
+		return nil, errors.New("Config: Dial cannot be nil")
+	}
+	// Default to sane values
+	if cfg.RetryDuration == 0 {
+		cfg.RetryDuration = time.Second
```
davecgh: I'd suggest `time.Second * 5` here to match the current default.

davecgh: Also, I'd prefer definitions for the defaults at the top of the file.  Same with the `.MaxOutbound` default below.

tuxcanfly: Thanks, updated.

connmgr/connmanager_test.go
```diff
@@ -0,0 +1,373 @@
...
+		Dial:          mockDialer,
+		OnConnection: func(c *connmgr.ConnReq, conn net.Conn) {
+			connected <- c
+		},
+	})
+	if err != nil {
+		t.Fatalf("connmgr.New error: %v", err)
+	}
+	cmgr.Start()
+	go cmgr.Wait()
```
davecgh: What's the point of this?

connmgr/connmanager_test.go
```diff
@@ -0,0 +1,373 @@
...
+	gotConnReq := <-connected
+	cmgr.Stop()
+	// already stopped
+	cmgr.Stop()
+	// ignored
+	cr := &connmgr.ConnReq{Addr: "127.0.0.1:18555", Permanent: true}
+	cmgr.Connect(cr)
```
davecgh: This section doesn't actually appear to be testing the result?  If it wasn't working properly and ignoring the request due to it being stopped, how would the test know?

connmgr/connmanager_test.go
```diff
@@ -0,0 +1,373 @@
...
+	}
+	gotState := cr.State()
+	wantState := connmgr.ConnEstablished
+	if gotState != wantState {
+		t.Fatalf("connect mode: %v - want state %v, got state %v", cr.Addr, wantState, gotState)
+	}
+	select {
+	case c := <-connected:
+		t.Fatalf("connect mode: got unexpected connection - %v", c.Addr)
+	case <-time.After(time.Millisecond):
```
davecgh: This is probably a bit too short to detect if it actually weren't working properly.

connmgr/connmanager_test.go
```diff
@@ -0,0 +1,373 @@
...
+		t.Fatalf("connmgr.New error: %v", err)
+	}
+
+	cr := &connmgr.ConnReq{Addr: "127.0.0.1:18555", Permanent: true}
+	go cmgr.Connect(cr)
+	cmgr.Start()
+	// retry in 1ms
+	// retry in 2ms - max retry duration reached
+	// retry in 2ms - timedDialer returns mockDial
+	<-connected
```
davecgh: What if it doesn't work as expected?  The text will never fail here.  Need a select statement with a timeout longer than `timedDailer` with a `t.Fatal` branch.

connmgr/connmanager_test.go
```diff
@@ -0,0 +1,373 @@
...
+}
+
+// TestNetworkFailure tests that the connection manager handles a network
+// failure graciously.
```
davecgh: graciously -> gracefully

connmgr/connmanager_test.go
```diff
@@ -0,0 +1,373 @@
...
+			t.Fatalf("network failure: got unexpected connection - %v", c.Addr)
+		},
+	})
+	if err != nil {
+		t.Fatalf("connmgr.New error: %v", err)
+	}
+	cmgr.Start()
+	time.AfterFunc(10*time.Millisecond, cmgr.Stop)
+	cmgr.Wait()
+	if atomic.LoadUint32(&dials) > 25 {
```
davecgh: 25 seems a bit high here.  There should only be 5 simultaneous attempts with a 5 millisecond retry.  Thus there will be 5 attempts, then 5 milliseconds later another 5 attempts, and then 5 more attempts after 10 milliseconds.  The last one is unlikely to be hit since it's stopped after 10 milliseconds, but to be safe I think 15 would cover it.

connmgr/seed.go
```diff
@@ -0,0 +1,62 @@
...
+
+// onSeed is the signature of the callback function which is invoked when DNS
+// seeding is succesfull.
+type onSeed func(addrs []*wire.NetAddress)
+
+// lookupFunc is the signature of the DNS lookup function.
+type lookupFunc func(string) ([]net.IP, error)
+
+// SeedFromDNS uses DNS seeding to populate the address manager with peers.
+func SeedFromDNS(chainParams *chaincfg.Params, lookupFn lookupFunc, seedFn onSeed) {
```
davecgh: Using unexported types in functions is generally not a good thing to do.  I'd suggest exporting the `OnSeed` and `LookupFunc` types.

server.go
```diff
@@ -1741,20 +1576,30 @@ func (s *server) peerHandler() {
...
 		permanentPeers = cfg.AddPeers
 	}
 	for _, addr := range permanentPeers {
-		sp := s.newOutboundPeer(addr, true)
-		if sp != nil {
-			go s.retryConn(sp, true)
-		}
+		go s.connManager.Connect(&connmgr.ConnReq{Addr: addr, Permanent: true})
 	}
+	// TODO: Sync Connects
```
davecgh: Reminder to finish this TODO.

server.go
```diff
@@ -1811,77 +1656,11 @@ out:
...
-			}
-		}
-
-		// We need more peers, wake up in ten seconds and try again.
-		if state.NeedMoreOutbound() {
-			time.AfterFunc(10*time.Second, func() {
-				s.wakeup <- struct{}{}
-			})
-		}
+		// TODO: Connect to more peers - this is now managed by connmgr.
```
davecgh: Reminder.

server.go
```diff
@@ -2557,6 +2335,39 @@ func newServer(listenAddrs []string, db database.DB, chainParams *chaincfg.Param
...
+					srvrLog.Errorf("Can't use connection: %v", err)
+					return
+				}
+				sp.connReq = c
+				srvrLog.Debugf("Connected to %s", sp.Addr())
+				s.addrManager.Attempt(sp.NA())
+			}
+		},
+	}
+	if len(cfg.ConnectPeers) == 0 {
```
davecgh: I'd reverse this a bit so the config is all done in one shot and without the need for the local like so:

```Go
	// Only setup a function to return new addresses to connect to when
	// not running in connect-only mode.
	var newAddressFunc connmgr.AddressFunc
	if len(cfg.ConnectPeers) == 0 {
		newAddressFunc = func() string {
			// ...
		}
	}

	// Create a connection manager.
	cmgr, err := connmgr.New(&connmgr.Config{
		RetryDuration: connectionRetryInterval
		// ...
		GetNewAddress: newAddressFunc,
		Dial:          btcdDial,
	})
	if err != nil {
		return nil, err
	}
```

tuxcanfly: Sounds good. Let me address that and the preceding review items.

connmgr/connmanager_test.go
tuxcanfly: Removed, it was left over from a refactor.

connmgr/connmanager_test.go
tuxcanfly: Updated. We want to make sure that requests are ignored after `Stop` is called. For `Connect`, we check that no `ID` is assigned. For `Disconnect` and `Remove` we check that no callback is fired.

connmgr/connmanager_test.go
tuxcanfly: Tested by forcing a mock connection in `Start` and checked that the test failed. Since we're using `mockDial` I think it is fast enough.

connmgr/connmanager_test.go
tuxcanfly: Thanks, updated.

connmgr/connmanager_test.go
tuxcanfly: Done.

connmgr/seed.go
tuxcanfly: Thanks, done.

server.go
tuxcanfly: Refactored as per the above suggestion.

connmgr/connmanager.go
```diff
@@ -0,0 +1,336 @@
...
+// be delayed by the configured retry duration.
+const maxFailedAttempts = 10
+
+var (
+	//ErrDialNil is used to indicate that Dial cannot be nil in the configuration.
+	ErrDialNil = errors.New("Config: Dial cannot be nil")
+
+	// maxRetryDuration is the max duration of time retrying of a persistent
+	// connection is allowed to grow to.  This is necessary since the retry
+	// logic uses a backoff mechanism which increases the interval base done
```
Roasbeef: I think the end of this comment should instead read: 
> which increases the interval base **times** the number of retries that have been done.

connmgr/connmanager.go
```diff
@@ -0,0 +1,336 @@
...
+					}
+					log.Debugf("Connection closed: %v", connReq.Addr)
+					delete(cm.conns, msg.id)
+
+					if cm.cfg.OnDisconnection != nil {
+						go cm.cfg.OnDisconnection(connReq)
+					}
+
+					cm.handleFailedConn(connReq, msg.retry)
+				}
```
Roasbeef: I think an error log message should be added here in the scenario that we get a `handleDisconnected` with an id that we don't know of. At the moment, I can't think of a case where this would arise, but it may help us track down bugs in the future.

connmgr/connmanager.go
```diff
@@ -0,0 +1,336 @@
...
+// id from known connections.
+func (cm *ConnManager) Remove(id uint64) {
+	if atomic.LoadInt32(&cm.stop) != 0 {
+		return
+	}
+	cm.requests <- handleDisconnected{id, false}
+}
+
+// Start launches the connection manager and begins connecting to the network.
+func (cm *ConnManager) Start() {
```
Roasbeef: This currently allows the `Start()` method to be called multiple times in a non-idempotent manner, that doesn't seem desirable. 

I think the code block below should be guarded with something like: 
```go
if !atomic.CompareAndSwapUint64(&cm.stop, 0, 1) {
    return
}
```

With this guard, calling the `Start` method multiple times (even concurrently) is idempotent. Alternatively, we could use a `sync.Once` here, though I lean towards the CAS call.

connmgr/connmanager.go
tuxcanfly: Thanks, added that.

server.go
tuxcanfly: Restored the checks we had earlier. For reference:

https://github.com/btcsuite/btcd/blob/5a9bac96/server.go#L753-L783

connmgr/connmanager.go
tuxcanfly: Added a atomic int guard similar to `addrmgr` to make sure `Start` can be called only once.

server.go
tuxcanfly: The checks have been restored but I'm not sure if keeping the old `tries` loop is a good way to handle this. For reference:

https://github.com/btcsuite/btcd/pull/653/commits/54cad6a05c16e047235be05b895fa16791094a21#diff-34c6b408d72845d076d47126c29948d1L2327

connmgr/connmanager.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+// New returns a new connection manager.
+// Use Start to start connecting to the network.
+func New(cfg *Config) (*ConnManager, error) {
+	if cfg.Dial == nil {
+		return nil, ErrDialNil
+	}
+	// Default to sane values
+	if cfg.RetryDuration == 0 {
```
dajohi: <= 0

connmgr/connmanager.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+// Config holds the configuration options related to the connection manager.
+type Config struct {
+	// MaxOutbound is the maximum number of outbound network connections to
+	// maintain. Defaults to 8.
+	MaxOutbound uint32
+
+	// RetryDuration is the duration to wait before retrying connection
+	// requests. Defaults to 1s.
```
dajohi: Code says 5 * time.Second.

connmgr/connmanager.go
```diff
@@ -0,0 +1,345 @@
...
+	c   *ConnReq
+	err error
+}
+
+// ConnManager provides a manager to handle network connections.
+type ConnManager struct {
+	// The following variables must only be used atomically.
+	start        int32
+	stop         int32
+	connReqCount uint64
```
dajohi: Please move connReqCount before start and stop.

dajohi:
The timeouts in _test.go are pretty short.  5ms and 10ms.  go test fails on my machine, but they pass when set them to 1 second.

tuxcanfly:
Rebased onto master. Investing two tests which seem to fail sometimes - `TestMaxRetryDuration` and `TestNetworkFailure`.

tuxcanfly:
Fixed the issues with unexpected test failures.

server.go
```diff
@@ -1853,78 +1702,9 @@ out:
 		if cfg.SimNet {
```
davecgh: This section can be removed since it no longer applies here, however the behavior needs to be retained in connection manager and currently it has not.

In fact, you can see this issue when running with the `--simnet` flag:

```
$ btcd --simnet -ddebug

21:34:24 2016-10-20 [DBG] CMGR: Attempting connection:
21:34:24 2016-10-20 [DBG] CMGR: Connection failed : dial tcp: missing address
21:36:26 2016-10-20 [DBG] CMGR: Attempting connection:
21:36:26 2016-10-20 [DBG] CMGR: Attempting connection:
21:36:26 2016-10-20 [DBG] CMGR: Attempting connection:
21:36:26 2016-10-20 [DBG] CMGR: Attempting connection:
21:36:26 2016-10-20 [DBG] CMGR: Attempting connection:
21:36:26 2016-10-20 [DBG] CMGR: Attempting connection:
21:36:26 2016-10-20 [DBG] CMGR: Attempting connection:
21:36:26 2016-10-20 [DBG] CMGR: Connection failed : dial tcp: missing address
21:36:26 2016-10-20 [DBG] CMGR: Connection failed : dial tcp: missing address
```

The key point of the simuation test network is that it is only intended to connect to specified peers and actively avoids advertising and connecting to discovered peers to prevent it from becoming a public test network.  You can see this throughout the code where it avoids adding any address to the address manager, does not provide any peers in response to a `getaddr` message, and ignores any addresses received via an `addr` message.  That means the connection manager should only ever connect to peers specified via `--addpeer` or `--connect` when on simnet, and, due to that fact, `--addpeer` is effectively identical to `--connect` on simnet.

Luckily, the fix should be as simple as:

```diff
	// Only setup a function to return new addresses to connect to when
-	// not running in connect-only mode.
+	// not running in connect-only mode.  The simulation network is always
+	// in connect-only mode since it is only intended to connect to
+	// specified peers and actively avoid advertising and connecting to
+	// discovered peers in order to prevent it from becoming a public test
+	// network.
	var newAddressFunc connmgr.AddressFunc
-	if len(cfg.ConnectPeers) == 0 {
+	if !cfg.SimNet && len(cfg.ConnectPeers) == 0 {
```

server.go
```diff
@@ -1713,67 +1609,21 @@ func (s *server) peerDoneHandler(sp *serverPeer) {
...
 // peerHandler is used to handle peer operations such as adding and removing
 // peers to and from the server, banning peers, and broadcasting messages to
 // peers.  It must be run in a goroutine.
 func (s *server) peerHandler() {
-	// Start the address manager and block manager, both of which are needed
-	// by peers.  This is done here since their lifecycle is closely tied
-	// to this handler and rather than adding more channels to sychronize
-	// things, it's easier and slightly faster to simply start and stop them
-	// in this handler.
+	// Start the address manager, block manager and connection manager, all of
```
davecgh: This comment update doesn't appear accurate any longer given the connection manager is not started until lower down.

server.go
```diff
@@ -2602,6 +2389,75 @@ func newServer(listenAddrs []string, db database.DB, chainParams *chaincfg.Param
...
+
+	// Create a connection manager.
+	cmgr, err := connmgr.New(&connmgr.Config{
+		RetryDuration: connectionRetryInterval,
+		MaxOutbound:   defaultMaxOutbound,
+		Dial:          btcdDial,
+		OnConnection: func(c *connmgr.ConnReq, conn net.Conn) {
+			sp := s.newOutboundPeer(c.Addr, c.Permanent)
+			if sp != nil {
+				sp.Connect(conn)
```
davecgh: Can we rename `peer.Connect` to something like `peer.AssociateConnection` in a separate PR?  Given there is a `connmgr.Connect` which actually establishes a connection, it seems rather confusing to have the same name on a peer when it doesn't actually connect, rather it just associates an existing connection with it.

connmgr/seed.go
```diff
@@ -0,0 +1,62 @@
+package connmgr
```
davecgh: Missing copyright and license header.

connmgr/connmanager.go
```diff
@@ -0,0 +1,345 @@
...
+
+// OnDisconnectionFunc is the signature of the callback function which is used to
+// notify disconnections.
+type OnDisconnectionFunc func(*ConnReq)
+
+// ConnState can be either pending, established, disconnected or failed.  When
+// a new connection is requested, it is attempted and categorized as
+// established or failed depending on the connection result.  An established
+// connection which was disconnected is categorized as disconnected.
+const (
```
davecgh: Please move this up to just after the type definition for `ConnState` so they're grouped.

server.go
```diff
@@ -1248,16 +1210,6 @@ func (s *server) handleAddPeerMsg(state *peerState, sp *serverPeer) bool {
...
-		if state.OutboundCount() >= state.maxOutboundPeers {
-			srvrLog.Infof("Max outbound peers reached [%d] - disconnecting "+
-				"peer %s", state.maxOutboundPeers, sp)
-			sp.Disconnect()
-			return false
-		}
-	}
-
 	// Limit max number of total peers.
 	if state.Count() >= cfg.MaxPeers {
```
davecgh: As @tuxcanfly pointed out, the connection manager only handles outbound connections, so this code does in fact have to be in the server to limit overall connections.  My previous comment can be ignored.

connmgr/connmanager.go
```diff
@@ -342,3 +342,15 @@ func New(cfg *Config) (*ConnManager, error) {
 	}
 	return &cm, nil
 }
+
+// TstSetMaxRetryDuration makes the ability to set the max retry duration
+// available to the test package.
+func TstSetMaxRetryDuration(d time.Duration) {
```
davecgh: This function isn't necessary anymore since the tests are in the same package.  They can just do `maxRetryDuration = whatever`.

connmgr/connmanager.go
```diff
@@ -342,3 +342,15 @@ func New(cfg *Config) (*ConnManager, error) {
...
+	maxRetryDuration = d
+}
+
+// TstSetConnManagerStop makes the ability to set cm.stop available to the test
+// package.
+func (cm *ConnManager) TstSetConnManagerStop(i int32) {
```
davecgh: Same here.

connmgr/connmanager_test.go
```diff
@@ -13,14 +13,11 @@ import (
...
 
-	"github.com/btcsuite/btcd/connmgr"
 	"github.com/btcsuite/btclog"
 )
 
-const maxRetryDuration = 2 * time.Millisecond
-
 func init() {
-	connmgr.TstSetMaxRetryDuration(maxRetryDuration)
+	TstSetMaxRetryDuration(2 * time.Millisecond)
```
davecgh: Per the previous comment:

```Go
	// Override the max retry duration when running tests.
	maxRetryDuration = 2 * time.Millisecond

**Comment by davecgh** [comment]
There is still an issue with bootstrapping addresses.  When you start with no known addresses, it takes time to query addresses from DNS and seed the address manager.  Meanwhile, the connection manager is requesting addresses and can easily end up getting an error due to the address manager not having any addresses in it yet.  When this happens, the connection manager never retries which means you would end up forever losing 1 or more outbound connections.

The can be seen easily with the following diff which intentionally forces the first 8 returned address to error out simulating the case of an initial empty address manager that takes a while to seed:

```diff
--- a/server.go
+++ b/server.go
@@ -2389,7 +2389,7 @@ func newServer(listenAddrs []string, db database.DB, chainParams *chaincfg.Param
        // network.
        var newAddressFunc connmgr.AddressFunc
        if !cfg.SimNet && len(cfg.ConnectPeers) == 0 {
-               newAddressFunc = func() (string, error) {
+               realNewAddressFunc := func() (string, error) {
                        for tries := 0; tries < 100; tries++ {
                                addr := s.addrManager.GetAddress("any")
                                if addr == nil {
@@ -2422,6 +2422,14 @@ func newServer(listenAddrs []string, db database.DB, chainParams *chaincfg.Param
                        }
                        return "", errors.New("no valid connect address")
                }
+               var numConnTries uint64
+               newAddressFunc = func() (string, error) {
+                       tries := atomic.AddUint64(&numConnTries, 1)
+                       if tries <= 8 {
+                               return "", errors.New("no valid connect address")
+                       }
+                       return realNewAddressFunc()
+               }
        }
```

With that diff applied observe the following output.  The errors are indeed expected, but also notice how it never retries and you are forever left with no outbound connections:

```
$ btcd --testnet --nolisten -dCMGR=debug,AMGR=debug
...
16:45:47 2016-10-21 [ERR] CMGR: Error fetching address: no valid connect address
16:45:47 2016-10-21 [ERR] CMGR: Error fetching address: no valid connect address
16:45:47 2016-10-21 [ERR] CMGR: Error fetching address: no valid connect address
16:45:47 2016-10-21 [ERR] CMGR: Error fetching address: no valid connect address
16:45:47 2016-10-21 [ERR] CMGR: Error fetching address: no valid connect address
16:45:47 2016-10-21 [ERR] CMGR: Error fetching address: no valid connect address
16:45:47 2016-10-21 [ERR] CMGR: Error fetching address: no valid connect address
16:45:47 2016-10-21 [ERR] CMGR: Error fetching address: no valid connect address
```

tuxcanfly:
Thanks @davecgh Yeah, I can see the issue. We're skipping over making the connection request when there's an err getting a new address. Looking into handling this.

davecgh:
Here is a diff that creates a stringer for the `ConnReq` type and makes the log statements more consistent with the previous prints.  The reason for consistency is some users have some logging filters setup that I'd like to avoid breaking.

```diff
diff --git a/connmgr/connmanager.go b/connmgr/connmanager.go
index 6c38b0c..1c88fc4 100644
--- a/connmgr/connmanager.go
+++ b/connmgr/connmanager.go
@@ -6,6 +6,7 @@ package connmgr

 import (
        "errors"
+       "fmt"
        "net"
        "sync"
        "sync/atomic"
@@ -99,6 +100,14 @@ func (c *ConnReq) State() ConnState {
        return state
 }

+// String returns a human-readable string for the connection request.
+func (c *ConnReq) String() string {
+       if c.Addr == "" {
+               return fmt.Sprintf("reqid %d", atomic.LoadUint64(&c.id))
+       }
+       return fmt.Sprintf("%s (reqid %d)", c.Addr, atomic.LoadUint64(&c.id))
+}
+
 // Config holds the configuration options related to the connection manager.
 type Config struct {
        // MaxOutbound is the maximum number of outbound network connections to
@@ -172,7 +181,7 @@ func (cm *ConnManager) handleFailedConn(c *ConnReq, retry bool) {
                if d > maxRetryDuration {
                        d = maxRetryDuration
                }
-               log.Debugf("Retrying connection to %v in: %v", c.Addr, d)
+               log.Debugf("Retrying connection to %v in %v", c, d)
                time.AfterFunc(d, func() {
                        cm.Connect(c)
                })
@@ -210,7 +219,7 @@ out:
                                connReq.updateState(ConnEstablished)
                                connReq.conn = msg.conn
                                conns[connReq.id] = connReq
-                               log.Infof("Connection established: %v", connReq.Addr)
+                               log.Debugf("Connected to %v", connReq)
                                connReq.retryCount = 0
                                cm.failedAttempts = 0

@@ -224,7 +233,7 @@ out:
                                        if connReq.conn != nil {
                                                connReq.conn.Close()
                                        }
-                                       log.Debugf("Connection closed: %v", connReq.Addr)
+                                       log.Debugf("Connection closed %v", connReq)
                                        delete(conns, msg.id)

                                        if cm.cfg.OnDisconnection != nil {
@@ -239,7 +248,8 @@ out:
                        case handleFailed:
                                connReq := msg.c
                                connReq.updateState(ConnFailed)
-                               log.Debugf("Connection failed %v: %v", connReq.Addr, msg.err)
+                               log.Debugf("Failed to connect to %v: %v", connReq,
+                                       msg.err)

                                cm.handleFailedConn(connReq, true)
                        }
@@ -282,7 +292,7 @@ func (cm *ConnManager) Connect(c *ConnReq) {
        if atomic.LoadUint64(&c.id) == 0 {
                atomic.StoreUint64(&c.id, atomic.AddUint64(&cm.connReqCount, 1))
        }
-       log.Debugf("Attempting connection: %v", c.Addr)
+       log.Debugf("Attempting to connect to %v", c)
        conn, err := cm.cfg.Dial("tcp", c.Addr)
        if err != nil {
                cm.requests <- handleFailed{c, err}
diff --git a/server.go b/server.go
index c7a17d4..aa71aca 100644
--- a/server.go
+++ b/server.go
@@ -2434,7 +2443,6 @@ func newServer(listenAddrs []string, db database.DB, chainParams *chaincfg.Param
                        if sp != nil {
                                sp.AssociateConnection(conn)
                                sp.connReq = c
-                               srvrLog.Debugf("Connected to %s", sp.Addr())
                                s.addrManager.Attempt(sp.NA())
                        }
                },
```

PR #655, opened 2016-04-07
TravisCI: Remove external go vet reference.

davecgh:
The vet tool moved into the Go source tree as of Go 1.5.  Its previous location in the `x/tools` repo was deprecated at that time and has now been removed.

This updates the `.travis.yml` configuration to avoid fetching vet from the old location and to simply use the version now available as part of the standard Go install.

Also, while here, remove the check for changing to tool path since it is no longer needed.

---
This change is [Reviewable](https://reviewable.io/reviews/btcsuite/btcd/655)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,9 @@ go:
 sudo: false
 before_install:
   - gotools=golang.org/x/tools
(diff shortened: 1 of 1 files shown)


jcvernaleo:
Big OK from me on this (since I want it for dcrd).

PR #658, opened 2016-04-10
server: Optimize map limiting in block manager.

davecgh:
This optimizes the way in which the maps are limited by the block manager.

Previously the code would read a cryptographically random value large enough to construct a hash, find the first entry larger than that value, and evict it.

That approach is quite inefficient and could easily become a bottleneck when processing transactions due to the need to read from a source such as `/dev/urandom` and all of the subsequent hash comparisons.

Luckily, strong cryptographic randomness is not needed here. The primary intent of limiting the maps is to control memory usage with a secondary concern of making it difficult for adversaries to force eviction of specific entries.

Consequently, this changes the code to make use of the pseudorandom iteration order of Go's maps along with the preimage resistance of the hashing function to provide the desired functionality.  It has previously been discussed that the specific pseudorandom iteration order is not guaranteed by the Go spec even though in practice that is how it is implemented.  This is not a concern however because even if the specific compiler doesn't implement that, the preimage resistance of the hashing function alone is enough.

Thanks to @Roasbeef for pointing out the efficiency concerns and the fact that strong cryptographic randomness is not necessary.

---
This change is [Reviewable](https://reviewable.io/reviews/btcsuite/btcd/658)

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -6,8 +6,6 @@ package main
 
 import (
 	"container/list"
-	"crypto/rand"
-	"math/big"
 	"net"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


blockmanager.go
```diff
@@ -1144,35 +1120,23 @@ func (b *blockManager) handleInvMsg(imsg *invMsg) {
...
-			return err
-		}
-		randHashNum := new(big.Int).SetBytes(randHashBytes)
-
-		// Try to find the first entry that is greater than the random
-		// hash.  Use the first entry (which is already pseudorandom due
-		// to Go's range statement over maps) as a fallback if none of
-		// the hashes in the map are larger than the random hash.
-		var foundHash *wire.ShaHash
+		// Remove a random entry from the map.  This relies on the fact
```
jrick: I wouldn't say relies here since that contradicts what follows, which said that the iteration order does not matter at all due to the cryptographic hash.

davecgh: Good point.  Updated.

jrick:
ok

PR #659, opened 2016-04-10
mining: Export block template fields.

davecgh:
This simply exports and adds some comments to the fields of the `BlockTemplate` struct.

This is primarily being done as a step toward being able to separate the mining code into its own package, but also it makes sense on its own because code that requests new block template necessarily examines the returned fields which implies they should be exported.

---
This change is [Reviewable](https://reviewable.io/reviews/btcsuite/btcd/659)


jrick:
ok

PR #660, opened 2016-04-10
chaincfg: Register networks instead of hard coding.

davecgh:
This modifies the `chaincfg` package to register the default network params via the `init` function instead of manually hard coding their data into the maps.  This is less error prone when adding new default networks.

A new function named `MustRegister` has been introduced that panics if there are any errors when registering the network that the new code makes use of and appropriate tests have been added.

---
This change is [Reviewable](https://reviewable.io/reviews/btcsuite/btcd/660)

### DIFF
--- a/chaincfg/params.go
+++ b/chaincfg/params.go
@@ -337,31 +337,10 @@ var (
 )
 
 var (
-	registeredNets = map[wire.BitcoinNet]struct{}{
-		MainNetParams.Net:       {},
-		TestNet3Params.Net:      {},
-		RegressionNetParams.Net: {},
-		SimNetParams.Net:        {},
-	}
-
-	pubKeyHashAddrIDs = map[byte]struct{}{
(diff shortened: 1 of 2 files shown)


jcvernaleo:
I'm mostly looking at it from an alt perspective, but I'm totally in favor of this.
OK

chaincfg/params.go
```diff
@@ -384,6 +363,14 @@ func Register(params *Params) error {
 	return nil
 }
 
+// MustRegister performs the same function as Register except it panics if there
+// is an error.  This should only be called from package init functions.
+func MustRegister(params *Params) {
```
jrick: can we unexport this?

davecgh: I was thinking it would be useful for others to be able to register their custom networks at init time too, but I suppose they could just write something similar.

davecgh: Done.

jrick:
ok

PR #661, opened 2016-04-10
chaincfg: Use hex strings for PoW limits.

davecgh:
Rather than having separate variables for each network's proof-of-work limit parameter, introduce a new function to convert a hex string to a `big.Int` and make use of it.  This has a couple of benefits such as being easier to visually verify the limit at a glance and removing the need to skip between two different areas of the file when looking at the parameters.

---
This change is [Reviewable](https://reviewable.io/reviews/btcsuite/btcd/661)

### DIFF
--- a/chaincfg/internal_test.go
+++ b/chaincfg/internal_test.go
@@ -12,3 +12,14 @@ func TestInvalidShaStr(t *testing.T) {
 	}()
 	newShaHashFromStr("banana")
 }
+
+// TestInvalidBigIntStrPanic ensures the newBigFromHex function panics when
+// called with invalid hex.
+func TestInvalidBigIntStrPanic(t *testing.T) {
+	defer func() {
+		if r := recover(); r == nil {
+			t.Errorf("Expected panic for invalid big int string")
+		}
+	}()
+	newBigFromHex("banana")
+}
--- a/chaincfg/params.go
+++ b/chaincfg/params.go
@@ -11,31 +11,6 @@ import (
 	"github.com/btcsuite/btcd/wire"
 )
 
-// These variables are the chain proof-of-work limit parameters for each default
-// network.
-var (
-	// bigOne is 1 represented as a big.Int.  It is defined here to avoid
-	// the overhead of creating it multiple times.
-	bigOne = big.NewInt(1)
-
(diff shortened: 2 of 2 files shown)


jcvernaleo:
I like the idea of moving the values to the new place but I'm not sure how I feel about setting it directly from hex rather than showing the math in big Ints.  Makes it harder to setup new networks.
At the very least, I think you should keep the old comments that state what the pow limits are:
// have for the main network.  It is the value 2^224 - 1.
rather than just showing the hex.  That helps keep btcd as a good reference for bitcoin values (not everyone remembers those numbers off hand) and makes it easier for an alt derived from this to change it.

jrick:
I do prefer the actual numeric calculation to be done in code rather than just converting a hexadecimal number.  It's unfortunate that code is so noisy, but that comes with the territory when using go (no operator overloading to clean it up).

davecgh:
Interesting.  Perhaps it's just because I'm so familiar with it, but I find the hex much cleaner and easier to understand than the `big.Int` calcs.  The reason being is that the point of the PoW limit is to have a specific number of leading zeroes.  The can be directly read from the hex whereas the calculation does not make it so obvious.

For example, consider the new `segnet4` network and compare `000001ffffffffffffffffffffffffff` with `new(big.Int).Sub(new(big.Int).Lsh(bigOne, 233), bigOne)`.  The latter you have to intrinsically just know that it's a 256-bit value in question and thus there are 23 leading zero bits.  The hex, on the hand makes it patently obvious that there are 23 leading zero bits.

However, if you guys are still against this, I'll go ahead and close it.

jrick:
> The hex, on the hand makes it patently obvious that there are 23 leading zero bits.

Disagree, because you could add extra leading zeros or take some away and the result would be the same, but if you eyeball the number of leading zeros, you will come up with something unexpected.

davecgh:
True that you _could_ write it that way, but that isn't how it's actually written, so when it comes to people reviewing the code, they'll see the properly written form and would be easier to spot any mistakes.

Anyways, I'm closing this since nobody else seems to agree!

jrick:
After some reconsideration, I would be ok with this change so long as the function performs a length check, as well as keeping the number it represents in a comment.

PR #662, opened 2016-04-11
chaincfg: Consolidate tests into the chaincfg pkg.

davecgh:
Putting the test code in the same package makes it easier for forks since they don't have to change the import paths as much and it also gets rid of the need for ` internal_test.go` to bridge.

This same thing should probably be done for the majority of the code base.

---
This change is [Reviewable](https://reviewable.io/reviews/btcsuite/btcd/662)


jcvernaleo:
I'm in favor of this since I did not enjoy changing those paths for dcrd.
OK

PR #663, opened 2016-04-13
Optimize by removing defers

dajohi:
defer's are nice for readability but they do add overhead.  This
gets rid of defer's where it is just as easy as not to use one.

### DIFF
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -284,9 +284,10 @@ func (mp *TxPool) isTransactionInPool(hash *chainhash.Hash) bool {
 func (mp *TxPool) IsTransactionInPool(hash *chainhash.Hash) bool {
 	// Protect concurrent access.
 	mp.mtx.RLock()
-	defer mp.mtx.RUnlock()
+	inPool := mp.isTransactionInPool(hash)
+	mp.mtx.RUnlock()
 
-	return mp.isTransactionInPool(hash)
+	return inPool
 }
 
 // isOrphanInPool returns whether or not the passed transaction already exists
@@ -308,9 +309,10 @@ func (mp *TxPool) isOrphanInPool(hash *chainhash.Hash) bool {
 func (mp *TxPool) IsOrphanInPool(hash *chainhash.Hash) bool {
 	// Protect concurrent access.
 	mp.mtx.RLock()
-	defer mp.mtx.RUnlock()
+	inPool := mp.isOrphanInPool(hash)
+	mp.mtx.RUnlock()
 
-	return mp.isOrphanInPool(hash)
+	return inPool
 }
 
 // haveTransaction returns whether or not the passed transaction already exists
@@ -328,9 +330,10 @@ func (mp *TxPool) haveTransaction(hash *chainhash.Hash) bool {
 func (mp *TxPool) HaveTransaction(hash *chainhash.Hash) bool {
 	// Protect concurrent access.
 	mp.mtx.RLock()
-	defer mp.mtx.RUnlock()
+	haveTx := mp.haveTransaction(hash)
+	mp.mtx.RUnlock()
 
-	return mp.haveTransaction(hash)
+	return haveTx
 }
 
 // removeTransaction is the internal function which implements the public
@@ -375,9 +378,8 @@ func (mp *TxPool) removeTransaction(tx *btcutil.Tx, removeRedeemers bool) {
 func (mp *TxPool) RemoveTransaction(tx *btcutil.Tx, removeRedeemers bool) {
 	// Protect concurrent access.
 	mp.mtx.Lock()
-	defer mp.mtx.Unlock()
-
 	mp.removeTransaction(tx, removeRedeemers)
+	mp.mtx.Unlock()
 }
 
 // RemoveDoubleSpends removes all transactions which spend outputs spent by the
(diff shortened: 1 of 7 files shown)


jrick:
While I'm in favor of improving performance, and this should do the trick, I would like to see numbers supporting these changes before they're made.  Defer does have one potentially nice property when unlocking mutexes, and that's the they would be locked again even if the critical section panics.  Since panics are automatically recovered from by the rpc server's http request handler, this could potentially lead to deadlocks where a mutex is never unlocked.

mempool/mempool.go
```diff
@@ -875,15 +880,15 @@ func (mp *TxPool) ProcessOrphans(hash *chainhash.Hash) []*btcutil.Tx {
 //
 // This function is safe for concurrent access.
 func (mp *TxPool) ProcessTransaction(tx *btcutil.Tx, allowOrphan, rateLimit bool) ([]*btcutil.Tx, error) {
+	log.Tracef("Processing transaction %v", tx.Hash())
+
 	// Protect concurrent access.
 	mp.mtx.Lock()
-	defer mp.mtx.Unlock()
```
davecgh: I'd prefer to leave this one.  When it's a simple function with only a single control path, I think it makes sense to avoid to the defer.  However, in a more complex function like this with multiple control paths, it's too easy to add a branch and forget to release the mutex.

davecgh:
I don't see any particular issue with doing this (except the one that I noted), but I do question how much of a speed benefit it will amount to.  There is definitely a lot of overhead associated with defer, so I could see that it might compound enough to make a difference.  However, I suspect compared to the other things going on, it probably won't make a lot huge difference.

That said, I have no objections.

stevenroose:
Isn't it good practice to have a single way to handle mutex unlocks across the whole project?

davecgh:
@stevenroose: Originally we exclusively used defer, but profiling has shown that it adds significant overhead (2-3x on average) in hot paths such that it made since to switch over to avoiding them for simple functions with a single exit point.

stevenroose:
Looks like the Go compiler could optimize when it recognizes a defer in a
single-exit-point function... But that's out of the scope of this project
ofc, removing such an overhead makes sense.

jrick:
@stevenroose the compiler would never be able to do that, unless it could statically tell that the code in the critical section never panics (same goes for inserting multiple unlock statements in each of the exit points).  Otherwise the two versions are not semantically the same (defer will still unlock the mutex during stack unwind).

mempool/mempool.go
```diff
@@ -393,6 +394,8 @@ func (mp *TxPool) RemoveDoubleSpends(tx *btcutil.Tx) {
 			}
 		}
 	}
+
```
davecgh: For consistency with the rest of the code and relevant changes, please wrap the lock unlock around the critical section without blank lines.

```Go
func (mp *TxPool) RemoveDoubleSpends(tx *btcutil.Tx) {
	mp.mtx.Lock()
	for _, txIn := range tx.MsgTx().TxIn {
		...
	}
	mp.mtx.Unlock()
}

File: mempool/mempool.go (hunk 1 above)
Comment (davecgh) [inline mempool/mempool.go]: Reminder.

File: mempool/mempool.go [hunk 3]
```diff
@@ -947,7 +956,6 @@ func (mp *TxPool) Count() int {
 // This function is safe for concurrent access.
 func (mp *TxPool) TxHashes() []*chainhash.Hash {
 	mp.mtx.RLock()
-	defer mp.mtx.RUnlock()
```
davecgh: Same thing here regarding the lock around the critical section and spacing consistency.

mempool/mempool.go
```diff
@@ -985,7 +994,6 @@ func (mp *TxPool) TxDescs() []*TxDesc {
 // concurrent access as required by the interface contract.
 func (mp *TxPool) MiningDescs() []*mining.TxDesc {
 	mp.mtx.RLock()
-	defer mp.mtx.RUnlock()
```
davecgh: Same thing here regarding the lock around the critical section and spacing consistency.

peer/mruinvmap.go
```diff
@@ -109,12 +107,13 @@ func (m *mruInventoryMap) Add(iv *wire.InvVect) {
 // This function is safe for concurrent access.
 func (m *mruInventoryMap) Delete(iv *wire.InvVect) {
 	m.invMtx.Lock()
-	defer m.invMtx.Unlock()
```
davecgh: Same thing here regarding the lock around the critical section and spacing consistency.

peer/mrunoncemap.go
```diff
@@ -107,12 +105,13 @@ func (m *mruNonceMap) Add(nonce uint64) {
 // This function is safe for concurrent access.
 func (m *mruNonceMap) Delete(nonce uint64) {
 	m.mtx.Lock()
-	defer m.mtx.Unlock()
```
davecgh: Same thing here regarding the lock around the critical section and spacing consistency.

peer/peer.go
```diff
@@ -1057,7 +1075,6 @@ func (p *Peer) handlePingMsg(msg *wire.MsgPing) {
 // ping was not previously sent.
 func (p *Peer) handlePongMsg(msg *wire.MsgPong) {
 	p.statsMtx.Lock()
```
davecgh: Same thing here regarding the lock around the critical section and spacing consistency.  After the comment just before the if statement, please.

dajohi:
Updated to address comments.

stevenroose:
https://tip.golang.org/doc/go1.8#defer

Yaay! Revert all the changes!!

PR #665, opened 2016-04-13
blockchain: Remove unneeded unspentness size check.

### LINKED ISSUE #664: chainio.go assumes 64-bit ints
Compiling for linux/arm7 I get:
```
# github.com/btcsuite/btcd/blockchain
blockchain/chainio.go:627: constant 4294967295 overflows int
```
The comparison numBitmapBytes > maxUint32 uses the output from utxoEntryHeaderCode:
```
func utxoEntryHeaderCode(entry *UtxoEntry, highestOutputIndex uint32) (uint64, int, error) {
```
which is an 'int' and depends on CPU architecture.  I tried replacing it but there's a bunch of functions and I don't know this code, so figured I'd just make the issue.

(Personally I don't even like that golang HAS arch-dependent ints but that's another story :)

davecgh:
The current code is needlessly checking the number of bytes needed to serialize the unspentness bitmap in the utxo against a maximum value that could never be returned because the function takes a `uint32` output index which is treated as a bit offset, and converts it bytes, which will necessarily be less than a max `uint32`.

This check also causes a compile error on arm where native integers are 32 bits.

This simply removes the unneeded check.

Fixes #664.

---
This change is [Reviewable](https://reviewable.io/reviews/btcsuite/btcd/665)

### DIFF
--- a/blockchain/chainio.go
+++ b/blockchain/chainio.go
@@ -16,10 +16,6 @@ import (
 	"github.com/btcsuite/btcutil"
 )
 
-const (
-	maxUint32 = 1<<32 - 1
-)
-
 var (
 	// hashIndexBucketName is the name of the db bucket used to house to the
 	// block hash -> block height index.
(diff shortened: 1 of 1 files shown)


dajohi:
OK

davecgh:
@T909 Can you confirm this resolves your issue as well?

T909:
Yup, compiles now, thanks!

(also... getting a crash now when running it on arm, but that's a different issue; ``atomic.AddUint64(&s.bytesSent, bytesSent)`` on line 1982 of server.go.  Weird.  It's actually ``/usr/local/go/src/sync/atomic/64bit_arm.go:31`` that crashes it, which may indicate a compiler problem on my end; will look into it.)

davecgh:
Thanks for testing.

PR #666, opened 2016-04-13
mining: Add tests for prio queue.

davecgh:
This adds a test to ensure the priority queue works properly both for sorting by fee per KB and priorities.

Thanks to @ceejep for the original test code and idea which was subsequently modified and cleaned up a bit to the code here.

---
This change is [Reviewable](https://reviewable.io/reviews/btcsuite/btcd/666)


cjepson:
tACK, good to have more mining tests in place.

PR #667, opened 2016-04-13
mining: Improve tests for prio queue.

davecgh:
This improves the tests of the priority queue to include the secondary sort ordering as well as adds some manual entries to ensure the edge conditions are properly tested.

This also brings the priority queue test coverage up to 100%.

---
This change is [Reviewable](https://reviewable.io/reviews/btcsuite/btcd/667)

### DIFF
--- a/mining_test.go
+++ b/mining_test.go
@@ -15,51 +15,96 @@ import (
 // TestTxFeePrioHeap ensures the priority queue for transaction fees and
 // priorities works as expected.
 func TestTxFeePrioHeap(t *testing.T) {
-	// Create priority items with random fees and priorites.
-	const numItems = 1000
-	prioItems := make([]*txPrioItem, numItems)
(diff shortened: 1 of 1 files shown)


jrick:
This looks ok.  My only suggestion is to use set a random seed and if any of the tests fail, then the seed can be printed so the test can be reproduced, but this is still an improvement by itself so that's no reason to hold up the PR.

davecgh:
Good suggestion.  I've updated the PR to create a new random source from a random seed and log the seed.

jrick:
I would only print it if the test failed:

```Go
defer func() {
    if t.Failed() {
        t.Log("Random numbers using seed: %v", randSeed)
    }
}()
```

davecgh:
Updated to only print seed on defer.

jrick:
ok

PR #668, opened 2016-04-14
fix memory alignment for 32-bit architectures

T909:
having 3 int32s above the uint64s in the struct
will cause mis-alignment for some 32-bit architectures.
see https://golang.org/pkg/sync/atomic/#pkg-note-BUG
This alligns bytesReceived and bytesSent.

---
This change is [Reviewable](https://reviewable.io/reviews/btcsuite/btcd/668)

### DIFF
--- a/server.go
+++ b/server.go
@@ -170,11 +170,12 @@ func (ps *peerState) forAllPeers(closure func(sp *serverPeer)) {
 // bitcoin peers.
 type server struct {
 	// The following variables must only be used atomically.
+	// Putting the uint64s first makes them 64-bit aligned for 32-bit systems.
+	bytesReceived uint64 // Total bytes received from all peers since start.
+	bytesSent     uint64 // Total bytes sent by all peers since start.
 	started       int32
 	shutdown      int32
 	shutdownSched int32
-	bytesReceived uint64 // Total bytes received from all peers since start.
-	bytesSent     uint64 // Total bytes sent by all peers since start.
 
 	listeners            []net.Listener
 	chainParams          *chaincfg.Params


server.go
```diff
@@ -170,11 +170,12 @@ func (ps *peerState) forAllPeers(closure func(sp *serverPeer)) {
 // bitcoin peers.
 type server struct {
 	// The following variables must only be used atomically.
+	// Putting the uint64s first makes them 64-bit alligned for 32-bit systems.
```
davecgh: alligned -> aligned

T909:
btcd would crash on arm7 (raspi 2):
```
02:41:03 2016-04-14 [INF] CHAN: Chain state (height 0, hash 18fb5ff510c09532033d2137a6914010509ee6258275a4b7e1b7b24b1d2191b2, totaltx 1, work 8388672)
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x4 pc=0x1688d0]

goroutine 58 [running]:
sync/atomic.addUint64(0x11056014, 0x89, 0x0, 0x89, 0x0)
	/usr/local/go/src/sync/atomic/64bit_arm.go:31 +0x68
main.(*server).AddBytesSent(0x11056000, 0x89, 0x0)
	/home/tx/pr/go/src/github.com/btcsuite/btcd/server.go:2008 +0x48
main.(*serverPeer).OnWrite(0x10fe2070, 0x10c761c0, 0x89, 0x76db0d30, 0x11012090, 0x0, 0x0)
	/home/tx/pr/go/src/github.com/btcsuite/btcd/server.go:939 +0x38
main.(*serverPeer).OnWrite-fm(0x10c761c0, 0x89, 0x76db0d30, 0x11012090, 0x0, 0x0)
	/home/tx/pr/go/src/github.com/btcsuite/btcd/server.go:1520 +0x4c
github.com/btcsuite/btcd/peer.(*Peer).writeMessage(0x10c761c0, 0x76db0d30, 0x11012090, 0x1, 0x0, 0x0)
	/home/tx/pr/go/src/github.com/btcsuite/btcd/peer/peer.go:1200 +0x518
github.com/btcsuite/btcd/peer.(*Peer).outHandler(0x10c761c0)
	/home/tx/pr/go/src/github.com/btcsuite/btcd/peer/peer.go:1874 +0x2e0
created by github.com/btcsuite/btcd/peer.(*Peer).start
	/home/tx/pr/go/src/github.com/btcsuite/btcd/peer/peer.go:2044 +0x258
```
(line numbers are a little off, may have been a few commits behind)
I'm not 100% sure how this works, but at the bottom of the "atomic" docs here:
https://golang.org/pkg/sync/atomic/#pkg-note-BUG
it says 

> On both ARM and x86-32, it is the caller's responsibility to arrange for 64-bit alignment of 64-bit words accessed atomically. The first word in a global variable or in an allocated struct or slice can be relied upon to be 64-bit aligned.

I moved bytesReceived and bytesSent to the top of the server struct.  That fixed it.  This is a bit outside my area of expertise so there may be better ways to achieve this fix, but it seems to work OK.

davecgh:
Right, I've run into it before where it has to be aligned.  I wasn't aware the 64-bits had to go before the 32-bit ones, however it does make sense given 32 + 32 + 32 is not 64-bit aligned.  This is why I'm generally not very thrilled with PRs that add them even though they are quite a lot faster than a mutex for counters like this.

T909:
Got it; it does seem a bit fragile.  I don't know much about the code in server.go so I don't want to make too many changes; swapping these lines was a quick fix I figure I'd share.  Maybe this is more of an 'informational' pull request rather than an actual request to get this merged in :)

davecgh:
It's a correct fix.  I appreciate the PR.  I'll merge it after you fix the comment typo.

T909:
Hmm I spelled that wrong in a bunch of places, oops :)  Fixed those

davecgh:
Looks good.

OK

PR #670, opened 2016-04-14
peer: use atomics instead of mutexes

dajohi:
This change is [Reviewable](https://reviewable.io/reviews/btcsuite/btcd/670)

### DIFF
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -392,10 +392,12 @@ type HostToNetAddrFunc func(host string, port uint16,
 // provided as a convenience.
 type Peer struct {
 	// The following variables must only be used atomically.
-	connected     int32
-	disconnect    int32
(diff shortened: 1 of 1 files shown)


davecgh:
This is a reasonable change, however it will need to be tested on ARM and other 32-bit platforms since atomic relies on 64-bit alignment of the fields.  I believe the alignment is correct based on the field order, but given it will panic if not on 32-bit architectures, it's important to test it.

dajohi:
I wonder if @T909 has the ability to test?

davecgh:
For reference, I did test this on 32-bit Windows and it works without issue, so I'm pretty confident it will work properly on ARM as well.

dajohi:
rebased

davecgh:
OK

PR #671, opened 2016-04-15
btcd: sendheaders server support

dajohi:
This adds support for serving headers instead of inventory messages in accordance with BIP0130.  btcd itself does not yet make use of the feature when receiving data.

This change is [Reviewable](https://reviewable.io/reviews/btcsuite/btcd/671)

### DIFF
--- a/server.go
+++ b/server.go
@@ -1302,6 +1302,26 @@ func (s *server) handleRelayInvMsg(state *peerState, msg relayMsg) {
 			return
 		}
 
+		// If the inventory is a block and the peer prefers headers,
+		// generate and send a headers message instead of an inventory
+		// message.
+		if msg.invVect.Type == wire.InvTypeBlock && sp.WantsHeaders() {
+			blockHeader, ok := msg.data.(wire.BlockHeader)
+			if !ok {
+				peerLog.Warnf("Underlying data for headers" +
+					" is not a block header")
+				return
+			}
+			msgHeaders := wire.NewMsgHeaders()
+			if err := msgHeaders.AddBlockHeader(&blockHeader); err != nil {
+				peerLog.Errorf("Failed to add block"+
+					" header: %v", err)
+				return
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -341,6 +341,17 @@ func (sp *serverPeer) addBanScore(persistent, transient uint32, reason string) {
 	}
 }
 
+// OnVerAck is invoked when a peer receives a verack bitcoin message.
+func (sp *serverPeer) OnVerAck(p *peer.Peer, msg *wire.MsgVerAck) {
```
davecgh: I'd rather leave this out completely until it is ready to be added.  I prefer to avoid big `TODO` blocks laying around the code base in general as issues are a better medium for that since they allow discussion.

server.go
```diff
@@ -1302,6 +1313,23 @@ func (s *server) handleRelayInvMsg(state *peerState, msg relayMsg) {
 			return
 		}
 
+		if msg.invVect.Type == wire.InvTypeBlock && sp.WantsHeaders() {
```
davecgh: Comment please.

server.go
```diff
@@ -1501,6 +1529,7 @@ func disconnectPeer(peerList map[int32]*serverPeer, compareFunc func(*serverPeer
 func newPeerConfig(sp *serverPeer) *peer.Config {
 	return &peer.Config{
 		Listeners: peer.MessageListeners{
+			OnVerAck:      sp.OnVerAck,
```
davecgh: This would go away when the function stub is removed as well.

davecgh:
Thanks for the PR.  It will be nice to properly support serving this feature.  Would you please update the commit message to include some verbiage that specifically calls out this is only for serving headers instead of inventory messages in accordance with BIP0130 and does not yet make use of the feature itself when receiving data.

davecgh:
OK

PR #675, opened 2016-04-20
btcd: Remove peer-specific logging funcs.

davecgh:
This removes the logging functions that are now implemented in the peer package as they are no longer used by `btcd` itself and should have been removed when they were copied into the peer package.

---
This change is [Reviewable](https://reviewable.io/reviews/btcsuite/btcd/675)


jrick:
ok

PR #677, opened 2016-04-24
txscript: Reduce script parse opcode allocs.

davecgh:
This changes the script template parsing function to use a pointer into the constant global opcode array for parsed opcodes as opposed to making a copy of the opcode entries which causes unnecessary allocations.

Profiling showed that after roughly 48 hours of operation, this copy was the culprit of 207 million unnecessary allocations.

---
This change is [Reviewable](https://reviewable.io/reviews/btcsuite/btcd/677)

### DIFF
--- a/txscript/script.go
+++ b/txscript/script.go
@@ -105,8 +105,8 @@ func parseScriptTemplate(script []byte, opcodes *[256]opcode) ([]parsedOpcode, e
 	retScript := make([]parsedOpcode, 0, len(script))
 	for i := 0; i < len(script); {
 		instr := script[i]
-		op := opcodes[instr]
-		pop := parsedOpcode{opcode: &op}
+		op := &opcodes[instr]
+		pop := parsedOpcode{opcode: op}
 
 		// Parse data out of instruction.
 		switch {


jrick:
It's impossible to enforce this since and very hard to tell at a glance since there's no immutability, but are you certain that none of the opcode pointers are used to modify the global array (whereas previously, mutating the data would only change that one opcode)?

davecgh:
Yes, I did go through and manually check every case to ensure they are only being used immutably.

davecgh:
In order to be careful, here is a grep of all references:

```
$ grep "\.opcode" *.go
engine.go:      if pop.opcode.value > OP_16 {
engine.go:              pop.opcode.value >= 0 && pop.opcode.value <= OP_PUSHDATA4 {
engine.go:      return pop.opcode.opfunc(pop, vm)
opcode.go:      switch pop.opcode.value {
opcode.go:      switch pop.opcode.value {
opcode.go:      switch pop.opcode.value {
opcode.go:      opcode := pop.opcode.value
opcode.go:      opcodeName := pop.opcode.name
opcode.go:              if pop.opcode.length == 1 {
opcode.go:      if pop.opcode.length == 1 {
opcode.go:      switch pop.opcode.length {
opcode.go:      if pop.opcode.length > 0 {
opcode.go:              retbytes = make([]byte, 1, pop.opcode.length)
opcode.go:                      pop.opcode.length)
opcode.go:      retbytes[0] = pop.opcode.value
opcode.go:      if pop.opcode.length == 1 {
opcode.go:      nbytes := pop.opcode.length
opcode.go:      if pop.opcode.length < 0 {
opcode.go:              switch pop.opcode.length {
opcode.go:      vm.dstack.PushInt(scriptNum((op.opcode.value - (OP_1 - 1))))
opcode.go:      switch op.opcode.value {
opcode.go:                              "upgrades", op.opcode.value-(OP_NOP1-1))
script.go:              pops[0].opcode.value == OP_HASH160 &&
script.go:              pops[1].opcode.value == OP_DATA_20 &&
script.go:              pops[2].opcode.value == OP_EQUAL
script.go:              if pop.opcode.value > OP_16 {
script.go:              if pop.opcode.value != opcode {
script.go:      opcode := pop.opcode.value
script.go:              switch pop.opcode.value {
script.go:                              pops[i-1].opcode.value >= OP_1 &&
script.go:                              pops[i-1].opcode.value <= OP_16 {
script.go:                              nSigs += asSmallInt(pops[i-1].opcode)
script.go:      return len(pops) > 0 && pops[0].opcode.value == OP_RETURN
scriptbuilder_test.go:          for _, opcode := range test.opcodes {
standard.go:            pops[1].opcode.value == OP_CHECKSIG
standard.go:            pops[0].opcode.value == OP_DUP &&
standard.go:            pops[1].opcode.value == OP_HASH160 &&
standard.go:            pops[2].opcode.value == OP_DATA_20 &&
standard.go:            pops[3].opcode.value == OP_EQUALVERIFY &&
standard.go:            pops[4].opcode.value == OP_CHECKSIG
standard.go:    if !isSmallInt(pops[0].opcode) {
standard.go:    if !isSmallInt(pops[l-2].opcode) {
standard.go:    if pops[l-1].opcode.value != OP_CHECKMULTISIG {
standard.go:    if l-2-1 != asSmallInt(pops[l-2].opcode) {
standard.go:    if l == 1 && pops[0].opcode.value == OP_RETURN {
standard.go:            pops[0].opcode.value == OP_RETURN &&
standard.go:            pops[1].opcode.value <= OP_PUSHDATA4 &&
standard.go:            return asSmallInt(pops[0].opcode) + 1
standard.go:    numSigs := asSmallInt(pops[0].opcode)
standard.go:    numPubKeys := asSmallInt(pops[len(pops)-2].opcode)
standard.go:            } else if pop.opcode.value == OP_0 {
standard.go:            requiredSigs = asSmallInt(pops[0].opcode)
standard.go:            numPubKeys := asSmallInt(pops[len(pops)-2].opcode)
```

As can be seen, the vast majority are just performing comparisons on the `value` and `length` fields.  The only cases that are using the opcode structure directly, and thus also need to be verified do not do any mutations, are those that call `isSmallInt` and `asSmallInt` functions with it.  Looking at those function, we can also see they are simply immutably accessing the `value` field.

jrick:
looks good. ok

PR #678, opened 2016-04-25
wire: Use large tx for deserialize benchmark.

davecgh:
This adds a benchmark for deserializing a large transaction that is often referred to as the megatransaction since it is the largest Bitcoin transaction mined to date.

It consists of 5569 inputs and 1 output and its hash is:
bb41a757f405890fb0f5856228e23b715702d714d59bf2b1feb70d8b2b4e3e08.

This is being done so there is a benchmark that tests more of a worst-case scenario which is a better candidate for identifying and testing improvements.

The following benchmark results shows the how much more intensive this transaction is over the existing mock transaction:

```
DeserializeTxSmall  1000000    1751 ns/op      376 B/op     16 allocs/op
DeserializeTxLarge  300     5093980 ns/op  1672829 B/op  33430 allocs/op
```


wire/bench_test.go
```diff
@@ -306,7 +308,7 @@ func BenchmarkWriteTxIn(b *testing.B) {
 }
 
 // BenchmarkDeserializeTx performs a benchmark on how long it takes to
-// deserialize a transaction.
+// deserialize a small transaction.
 func BenchmarkDeserializeTx(b *testing.B) {
```
jrick: Can we add Small here as well to match?

davecgh: Done

jrick:
ok

PR #679, opened 2016-04-25
wire: Don't allocate new readers in benchmarks.

davecgh:
NOTE: This requires #678 in order for the reported allocation statistics to be correct.

This modifies the benchmarks in the wire package to avoid creating a new reader for each iteration.  This is useful since it means that showing the memory allocations will only show the function under test instead of the allocation for the benchmark setup as well.

The following is a before and after comparison of the allocations with the benchmarks that did not change removed:

```
benchmark              old allocs     new allocs     delta
------------------------------------------------------------
ReadVarInt1            2              1              -50.00%
ReadVarInt3            2              1              -50.00%
ReadVarInt5            2              1              -50.00%
ReadVarInt9            2              1              -50.00%
ReadVarStr4            4              3              -25.00%
ReadVarStr10           4              3              -25.00%
ReadOutPoint           2              1              -50.00%
ReadTxOut              4              3              -25.00%
ReadTxIn               6              5              -16.67%
DeserializeTxSmall     16             15             -6.25%
DeserializeTxLarge     33430          33428          -0.01%
ReadBlockHeader        8              7              -12.50%
```

### DIFF
--- a/wire/bench_test.go
+++ b/wire/bench_test.go
@@ -152,8 +152,10 @@ func BenchmarkWriteVarInt9(b *testing.B) {
 // a single byte variable length integer.
 func BenchmarkReadVarInt1(b *testing.B) {
 	buf := []byte{0x01}
+	r := bytes.NewReader(buf)
 	for i := 0; i < b.N; i++ {
-		ReadVarInt(bytes.NewReader(buf), 0)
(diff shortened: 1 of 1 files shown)


jrick:
ok

PR #680, opened 2016-04-25
blockchain: Remove unused root field.

davecgh:
This removes the root field and all references to it from the `BlockChain` struct since it is no longer required.

It was previously required because the chain state was not initialized when the instance was created.  However, that is no longer the case, so there is no reason to keep it around any longer.

---
This change is [Reviewable](https://reviewable.io/reviews/btcsuite/btcd/680)


jrick:
ok

PR #682, opened 2016-04-25
wire: Add several decode benchmarks.

davecgh:
This adds decode benchmarks for several of the messages that profiling has identified to cause a lot of allocations in addition to those that already exist.  By adding these benchmarks, it makes it easier to get allocation and speed statistics which can in turn be used to compare future improvements.

The following bencharmarks have been added:

`DecodeGetHeaders`, `DecodeHeaders`, `DecodeGetBlocks`, `DecodeAddr`, `DecodeInv`,
`DecodeNotFound`, and `DecodeMerkleBlock`

For reference, here is the benchmark data as of this commit.

```
DecodeGetHeaders     93261 ns/op     24120 B/op     1004 allocs/op
DecodeHeaders      2071263 ns/op    368399 B/op    18002 allocs/op
DecodeGetBlocks      92486 ns/op     24120 B/op     1004 allocs/op
DecodeAddr          850608 ns/op    136202 B/op     9002 allocs/op
DecodeInv         17107172 ns/op   3601447 B/op   150004 allocs/op
DecodeNotFound    17522225 ns/op   3601444 B/op   150004 allocs/op
DecodeMerkleBlock    21062 ns/op      5192 B/op      222 allocs/op
```

### DIFF
--- a/wire/bench_test.go
+++ b/wire/bench_test.go
@@ -7,7 +7,9 @@ package wire
 import (
 	"bytes"
 	"compress/bzip2"
+	"fmt"
 	"io/ioutil"
+	"net"
 	"os"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


jrick:
All of the `StopTimer` and `StartTimer` calls can be replaced with just a `ResetTimer` before the actual work to be benchmarked.

davecgh:
Updated to use `ResetTimer` as suggested.

wire/bench_test.go
```diff
@@ -432,6 +434,221 @@ func BenchmarkWriteBlockHeader(b *testing.B) {
...
+	}
+
+	// Serialize it so the bytes are available to test the decode below.
+	var bb bytes.Buffer
+	if err := m.BtcEncode(&bb, pver); err != nil {
+		b.Fatalf("MsgGetHeaders.BtcEncode: unexpected error: %v", err)
+	}
+	buf := bb.Bytes()
+
+	b.ResetTimer()
```
jrick: Shouldn't the timer resets go just before the loop? Otherwise this will include an additional allocs and CPU time for the reader creation and message declaration.

davecgh: I can change it, but I'm pretty sure `ResetTimer` doesn't reset the allocation count because that is an average.

davecgh: Done.

jrick:
ok

PR #683, opened 2016-04-25
Concurrently handle websocket client JSON-RPC requests

### LINKED ISSUE #504: Websocket RPCs should not be processed synchronously.
The current websocket RPC server will block both the read and servicing of all later requests until the current request is finished being processed and the response is sent.  While not inherently wrong, this is less useful because two individual requests are not guaranteed to be processed in order, and doing so will cause short requests (e.g. session) that do not depend on database state to block on previous requests that do make database calls.

Before fixing this, we will probably want to consider throttling db calls per websocket client, to avoid issues where one client connects and starts hundreds of db calls which will significantly slow down block processing and other clients.

jrick:
There are two possible designs that can be used for throttling: fanout with a set number of goroutines that are always running, or a semaphore with a new goroutine started each time.  I am in favor of using the semaphore as it allows easily bumping up the max number of concurrent requests without paying a higher constant overhead for the added goroutines.

Fixes #504.

### DIFF
--- a/config.go
+++ b/config.go
@@ -40,6 +40,7 @@ const (
 	defaultBanThreshold          = 100
 	defaultMaxRPCClients         = 10
 	defaultMaxRPCWebsockets      = 25
+	defaultMaxRPCConcurrentReqs  = 20
 	defaultVerifyEnabled         = false
 	defaultDbType                = "ffldb"
 	defaultFreeTxRelayLimit      = 15.0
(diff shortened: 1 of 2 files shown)


jrick:
A few more questions remain:

1. Do we want to use a semaphore or fanout design for limiting requests?  Or something else?
2. What should the default max number of concurrent requests be?  I would like it to be a higher number (e.g. 200) so that it can better support resource intensive applications like block explorers.  This is closely related to question 1 since a higher limit means higher memory overhead for each worker's stack.
3. Should there be a timeout for requests that take too long to be serviced?

rpcwebsocket.go
```diff
@@ -36,8 +36,19 @@ const (
...
+	maxConcurrentWebsocketClientRequests = 20
 )
 
+type semaphore chan struct{}
+
+func makeSemaphore(n int) semaphore {
+	return make(chan struct{}, n)
+}
+
+func (s semaphore) p() { s <- struct{}{} }
```
davecgh: I think these should be named `Acquire` and `Release`.  I'm of course familiar with the `P(S)` and `V(S)` syntax when dealing with semaphores, but I prefer the code to be readable by everyone and that nomenclature is not particularly illuminating for those who aren't familiar with primitive semaphores.

davecgh:
1) I think the semaphore is a good choice.  It does mean more goroutines are launched along with their associated overhead, but Go is particularly well suited and dealing with launching a lot of short lived goroutines, so I don't think that will be an issue.
2) I'm not opposed to a relative high amount of concurrent requests such as 200 being allowed, though I do think we should load it up and see how it performs under such a scenario.
3) I'd prefer there to be a timeout, but this could be tricky to get right since some things, like rescan and chain verification inherently take a long time.  The timeout would probably have to be selectively applied.
4) If it's made a flag, we can make the default something lower like 20 and more intensive applications like block explorers could bump it up, so that's probably a win-win.

jrick:
Regarding 3, the timeout is specifically for beginning to service the request (while waiting for other requests to finish), not the total time to respond to the request.

jrick:
I believe that there is a bug with how WaitForShutdown attempts to acquire all of the semaphore counts to ensure that no more requests are currently being handled.  First of all, this prevents WaitForShutdown to be used by multiple callers, since they would deadlock each other due to semaphore contention.  Finally, I'm not even sure that there needs to be any waiting for these handlers to finish running since the websocket client connection is closed from Disconnect and the results of these handlers are immediately thrown away.  We could get away with just leaving those goroutines behind.

It would be better behavior to return some kind of "shutting down" or "cancelled" error to the client instead of immediately disconnecting them but this appears to be a bug with the code on master as well so I'm not sure it's within the scope of this PR.

davecgh:
Yeah, putting the semaphore in `WaitForShutdown` would lead to issues.  That is why it currently just uses the `sync.WaitGroup` to wait for the goroutines to shutdown.

One issue I can see about this new design is there is no cancellation on disconnect.  So, let's say I connect, launch a bunch of long-running requests, and then disconnect.  Those tasks will still be running.  Previously this wasn't an issue because there was only allowed to be one "standard" handler and one "async" one running.

rpcwebsocket.go
```diff
@@ -38,6 +38,15 @@ const (
 	websocketSendBufferSize = 50
 )
 
+type semaphore chan struct{}
```
davecgh: Missing comments on all of these.

jrick:
I had to avoid the waitgroup because wg.Wait is already called in another goroutine.  It is incorrect goroutine usage to positively increment waitgrout concurrently with a call to Wait.

I'm confused by your last comment.  It is *already* an issue on master.  Granted, there could be only a max of two requests currently running (rescan for the async request, and any other non-async request) but the requests would be run to completion (or early cancellation, in the case of rescan) and **no response would be sent to the client** since the websocket connection was already disconnected.

davecgh:
Also, I much prefer the pevious `handleMessage`.  It split out the handling code from the input routine and the code was clearer as it handled the entire path when the user was not authenticated in one area.  Now it's all split up and I'd have to review it really carefully, but I suspect the semantics have changed as a result.

rpcwebsocket.go
```diff
@@ -1102,6 +1067,29 @@ out:
 	rpcsLog.Tracef("Websocket client input handler done for %s", c.addr)
 }
 
+func (c *wsClient) serviceRequest(r *parsedRPCCmd) {
```
davecgh: Missing comment.

rpcwebsocket.go
jrick: Would a `// http://lmgtfy.com/?q=semaphore` work? :p

These kind of useless comments make me want to just remove the type instead and deal with a normal `chan struct{}`.  The type name and method names already document its intended usage.

rpcwebsocket.go
jrick: Yep, this one deserves a comment.  will add

davecgh:
In practice, it's not an issue on master because all long-running RPCs are marked and executed via the async handler, which means it's limited to a single request.  Also, as you noted, in the case of rescan, it will cancel early when it tries to send the response and fails.

So, the end result is that on master there could only be a max of *one* outstanding long-running request that would finish and throw away the response.

However, with this new code, there can be up to whatever the maximum number of allowed async requests all running to completion after the client has long disconnected.

jrick:
I understand that this behavior is desired.  However, **none** of the existing request handlers, except for rescan, check for cancellation.  You can't terminate another goroutine.  It has to exit by itself, so every handler will need to be rewritten.

jrick:
I think that is the wrong way to go about it.  We want early cancellation, not restrictions on what combinations of RPCs are allowed together.  The latter just makes the RPC server as a whole less useful.

When I was talking about returning an early cancellation error, it would be due to the server disconnecting from the client.  There is nothing to prevent the client from disconnecting, leaving those handlers running (until completion, or early cancellation).

Even if we tried to limit the number of concurrent rescans, an (authenticated, mind you) client can still make multiple connections to the server to still execute each rescan in parallel.

This is a non-issue IMO.  Cancellation errors when disconnected by the server would be nice, but it doesn't solve what you want to solve.

davecgh:
Given it's an authenticated interface, I think it's probably reasonable to avoid the restrictions.  With an unauthenticated REST interface though, I definitely would want to ensure certain *long-running* requests are limited to one at a time.

Anyways, I'll do another review of this with the recent updates shortly.

jrick:
i'll get around to rebasing this one of these days since the new db is in and it should not be an issue to make many concurrent db accesses as before.

jrick:
Rebased and squashed over master.

config.go
```diff
@@ -82,73 +83,74 @@ func minUint32(a, b uint32) uint32 {
...
+	RPCLimitUser                      string        `long:"rpclimituser" description:"Username for limited RPC connections"`
+	RPCLimitPass                      string        `long:"rpclimitpass" default-mask:"-" description:"Password for limited RPC connections"`
+	RPCListeners                      []string      `long:"rpclisten" description:"Add an interface/port to listen for RPC connections (default port: 8334, testnet: 18334)"`
+	RPCCert                           string        `long:"rpccert" description:"File containing the certificate file"`
+	RPCKey                            string        `long:"rpckey" description:"File containing the certificate key"`
+	RPCMaxClients                     int           `long:"rpcmaxclients" description:"Max number of RPC clients for standard connections"`
+	RPCMaxWebsockets                  int           `long:"rpcmaxwebsockets" description:"Max number of RPC websocket connections"`
+	RPCMaxWebsocketConcurrentRequests int           `long:"rpcmaxwebsocketconcurrentrequests" description:"Max number of concurrent RPC requests that may be processed concurrently"`
```
jrick: I prefer concurrent instead of parallel, but sure I will make it shorter.

davecgh:
I have pretty extensively tested this using both [btcrpcclient](https://github.com/btcsuite/btcrpcclient) and via JavaScript through Firefox and Chrome.  It is working as expected as far as concurrently processing the incoming requests and generating the responses.

Unfortunately the underlying websockets package does not support concurrent responses, so all of the *responses* end up being delivered single threaded.  It would be better if that weren't the case, but this is nevertheless still a huge improvement as the majority of time serving requests is spent in the handlers which this PR allows to happen in parallel.

PR #684, opened 2016-04-26
wire: Significantly reduce decoding allocs.

davecgh:
Live profiling data pulled from an instance of btcd running for a week shows that the vast majority of in-use (aka live) objects were the result of the deserialization process in wire.  In particular, the following were the top two entries of live objects when the memory pool contained 46,396 transactions:

```
4,637,941 39.44% 39.44%  wire.(*MsgTx).BtcDecode
4,511,685 38.37% 77.81%  wire.ReadVarBytes
```

This clearly shows that `77.81%` of all live allocations were allocated by transaction decoding and amounted to a total of `9,149,626` allocations.  Combining this with the fact there were `46,396` transactions in the memory pool, the average number of live allocations per deserialized transaction can be calculated as `196`.

Also, as shown by the following graph, the runtime overhead of tracking the massive amount of allocations was over `500MB` and climbing.

pr_memstats_before

Consequently, this pull request consists of a series of commits which significantly reduce the number of allocations the wire package performs with a focus on transactions since the profiling data showed they were the biggest offender.  It has been split into a series of logical commits that each reduce the number of allocations the wire package performs across the board.  Each commit contains the benchmark comparisons to show the effects of each change.

The end result of all changes in this PR is the number of allocations required to deserialize a transaction has been reduced to 6 regardless of the number of inputs and outputs.  In effect this equates to a `1-(6/196) ~= 97%` reduction in the number of live allocations across the total deserialized transactions.

The following graph shows the same running time and a very similar number of transaction in the mempool with this pull request applied.  As can be seen, the runtime overhead stabalized around `300MB` and, as a result, the overall memory usage for the entire process dropped by approximately `200MB`.  NOTE: This graph also includes PR #686.

pr_memstats_after

Finally, the following is a before and after comparison of the allocations:
```
benchmark              old allocs     new allocs     delta
-------------------------------------------------------------
WriteVarInt1           1              0              -100.00%
WriteVarInt3           1              0              -100.00%
WriteVarInt5           1              0              -100.00%
WriteVarInt9           1              0              -100.00%
ReadVarInt1            1              0              -100.00%
ReadVarInt3            1              0              -100.00%
ReadVarInt5            1              0              -100.00%
ReadVarInt9            1              0              -100.00%
ReadVarStr4            3              2              -33.33%
ReadVarStr10           3              2              -33.33%
WriteVarStr4           2              1              -50.00%
WriteVarStr10          2              1              -50.00%
ReadOutPoint           1              0              -100.00%
WriteOutPoint          1              0              -100.00%
ReadTxOut              3              0              -100.00%
WriteTxOut             2              0              -100.00%
ReadTxIn               5              0              -100.00%
WriteTxIn              3              0              -100.00%
DeserializeTxSmall     15             5              -66.67%
DeserializeTxLarge     33428          6              -99.98%
SerializeTx            8              0              -100.00%
ReadBlockHeader        7              0              -100.00%
WriteBlockHeader       10             4              -60.00%
DecodeGetHeaders       1004           2              -99.80%
DecodeHeaders          18002          2              -99.99%
DecodeGetBlocks        1004           2              -99.80%
DecodeAddr             9002           2002           -77.76%
DecodeInv              150005         2              -100.00%
DecodeNotFound         150004         3              -100.00%
DecodeMerkleBlock      222            3              -98.65%
TxSha                  10             2              -80.00%
```

### DIFF
--- a/wire/common.go
+++ b/wire/common.go
@@ -14,8 +14,161 @@ import (
 	"github.com/btcsuite/fastsha256"
 )
 
-// MaxVarIntPayload is the maximum payload size for a variable length integer.
-const MaxVarIntPayload = 9
+const (
+	// MaxVarIntPayload is the maximum payload size for a variable length integer.
+	MaxVarIntPayload = 9
+
+	// binaryFreeListMaxItems is the number of buffers to keep in the free
+	// list to use for binary serialization and deserialization.
+	binaryFreeListMaxItems = 1024
+)
+
+var (
+	// littleEndian is a convenience variable since binary.LittleEndian is
+	// quite long.
+	littleEndian = binary.LittleEndian
+
+	// bigEndian is a convenience variable since binary.BigEndian is quite
+	// long.
+	bigEndian = binary.BigEndian
+)
+
+// binaryFreeList defines a concurrent safe free list of byte slices (up to the
+// maximum number defined by the binaryFreeListMaxItems constant) that have a
+// cap of 8 (thus it supports up to a uint64).  It is used to provide temporary
+// buffers for serializing and deserializing primitive numbers to and from their
+// binary encoding in order to greatly reduce the number of allocations
+// required.
+//
+// For convenience, functions are provided for each of the primitive unsigned
+// integers that automatically obtain a buffer from the free list, perform the
+// necessary binary conversion, read from or write to the given io.Reader or
+// io.Writer, and return the buffer to the free list.
+type binaryFreeList chan []byte
+
+// Borrow returns a byte slice from the free list with a length of 8.  A new
+// buffer is allocated if there are not any available on the free list.
+func (l binaryFreeList) Borrow() []byte {
+	var buf []byte
+	select {
+	case buf = <-l:
+	default:
+		buf = make([]byte, 8)
+	}
+	return buf[:8]
+}
+
+// Return puts the provided byte slice back on the free list.  The buffer MUST
+// have been obtained via the Borrow function and therefore have a cap of 8.
+func (l binaryFreeList) Return(buf []byte) {
+	select {
+	case l <- buf:
+	default:
+		// Let it go to the garbage collector.
+	}
+}
+
+// Uint8 reads a single byte from the provided reader using a buffer from the
+// free list and returns it as a uint8.
+func (l binaryFreeList) Uint8(r io.Reader) (uint8, error) {
+	buf := l.Borrow()[:1]
+	if _, err := io.ReadFull(r, buf); err != nil {
+		l.Return(buf)
+		return 0, err
+	}
+	rv := buf[0]
+	l.Return(buf)
+	return rv, nil
+}
+
+// Uint16 reads two bytes from the provided reader using a buffer from the
+// free list, converts it to a number using the provided byte order, and returns
+// the resulting uint16.
+func (l binaryFreeList) Uint16(r io.Reader, byteOrder binary.ByteOrder) (uint16, error) {
+	buf := l.Borrow()[:2]
+	if _, err := io.ReadFull(r, buf); err != nil {
+		l.Return(buf)
+		return 0, err
+	}
+	rv := byteOrder.Uint16(buf)
+	l.Return(buf)
+	return rv, nil
+}
+
+// Uint32 reads four bytes from the provided reader using a buffer from the
+// free list, converts it to a number using the provided byte order, and returns
+// the resulting uint32.
+func (l binaryFreeList) Uint32(r io.Reader, byteOrder binary.ByteOrder) (uint32, error) {
+	buf := l.Borrow()[:4]
+	if _, err := io.ReadFull(r, buf); err != nil {
+		l.Return(buf)
+		return 0, err
+	}
+	rv := byteOrder.Uint32(buf)
+	l.Return(buf)
+	return rv, nil
+}
+
+// Uint64 reads eight bytes from the provided reader using a buffer from the
+// free list, converts it to a number using the provided byte order, and returns
+// the resulting uint64.
+func (l binaryFreeList) Uint64(r io.Reader, byteOrder binary.ByteOrder) (uint64, error) {
+	buf := l.Borrow()[:8]
+	if _, err := io.ReadFull(r, buf); err != nil {
+		l.Return(buf)
+		return 0, err
+	}
+	rv := byteOrder.Uint64(buf)
+	l.Return(buf)
+	return rv, nil
+}
+
+// PutUint8 copies the provided uint8 into a buffer from the free list and
+// writes the resulting byte to the given writer.
+func (l binaryFreeList) PutUint8(w io.Writer, val uint8) error {
+	buf := l.Borrow()[:1]
+	buf[0] = val
+	_, err := w.Write(buf)
+	l.Return(buf)
+	return err
+}
+
+// PutUint16 serializes the provided uint16 using the given byte order into a
+// buffer from the free list and writes the resulting two bytes to the given
+// writer.
+func (l binaryFreeList) PutUint16(w io.Writer, byteOrder binary.ByteOrder, val uint16) error {
+	buf := l.Borrow()[:2]
+	byteOrder.PutUint16(buf, val)
+	_, err := w.Write(buf)
+	l.Return(buf)
+	return err
+}
+
+// PutUint32 serializes the provided uint32 using the given byte order into a
+// buffer from the free list and writes the resulting four bytes to the given
+// writer.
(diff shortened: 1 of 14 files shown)


wire/common.go
```diff
@@ -10,75 +10,251 @@ import (
...
+	var buf []byte
+	select {
+	case buf = <-l:
+	default:
+		buf = make([]byte, 8)
+	}
+	return buf[:8]
+}
+
+// Return puts the provided byte slice back on the free list.  The buffer is
```
jongillham: I think the 'is' is not needed on this line.

davecgh: Nice catch.  Updated.

wire/msgtx.go
```diff
@@ -57,8 +57,80 @@ const (
...
+	// concatenated into a single contiguous buffers.  This value was chosen
+	// because it slightly more than twice the size of the vast majority of
+	// all "standard" scripts.  Larger scripts are still deserialize just
+	// fine as the free list will simply be bypassed for them.
+	freeListMaxScriptSize = 512
+
+	// freeListMaxItems is the number of buffers to keep in the free list
+	// to use for script deserialization.  This value allows up to 100
+	// scripts per transaction being simultaneously deserialized by 125
+	// peers.  Thus, the peak usage of the free list is 125,000 * 512 =
```
davecgh: This should read 12,500 * 512.

wire/msgtx.go
```diff
@@ -58,8 +57,80 @@ const (
...
```
davecgh: This should read 12,500.

wire/msgtx.go
```diff
@@ -58,8 +57,80 @@ const (
 	// number of transaction outputs 1 byte + LockTime 4 bytes + min input
 	// payload + min output payload.
 	minTxPayload = 10
+
+	// freeListMaxScriptSize is the size of each buffer in the free list
+	// that	is used for deserializing scripts from the wire before they are
+	// concatenated into a single contiguous buffers.  This value was chosen
+	// because it slightly more than twice the size of the vast majority of
```
wallclockbuilder: // because it **is** slightly more than twice the size of the vast majority of

wire/msgtx.go
```diff
@@ -58,8 +57,80 @@ const (
...
+	// all "standard" scripts.  Larger scripts are still deserialize just
```
wallclockbuilder: // all "standard" scripts. Larger scripts are still deserialize**d** just

wire/msgtx.go
davecgh: Thanks.  Fixed.

wire/msgtx.go
davecgh: Fixed.

davecgh:
Rebased to the latest master and squashed all review comment fixes into their respective commits.

wire/common.go
```diff
@@ -390,31 +508,30 @@ func ReadVarInt(r io.Reader, pver uint32) (uint64, error) {
...
 	}
 
-	var buf [9]byte
-	buf[0] = 0xff
-	binary.LittleEndian.PutUint64(buf[1:], val)
-	_, err := w.Write(buf[:])
-	return err
+	err := binarySerializer.PutUint8(w, 0xff)
+	if err != nil {
+		return err
```
Roasbeef: This may just be Github's formatting (I'm viewing this using the split diff), but it looks like this line hasn't gone through `go fmt`.

wire/msgtx.go
```diff
@@ -57,8 +57,80 @@ const (
...
+func (c scriptFreeList) Borrow(size uint64) []byte {
+	if size > freeListMaxScriptSize {
+		return make([]byte, size, size)
+	}
+
+	var buf []byte
+	select {
+	case buf = <-c:
+	default:
+		buf = make([]byte, freeListMaxScriptSize, freeListMaxScriptSize)
```
Roasbeef: This line be be reduced to simply: 
```go
buf = make([]byte, freeListMaxScriptSize)
```
Since if the capacity parameter is omitted, then the capacity of the slice will be set to the specified lenght.

Roasbeef:
Excellent PR! 

I've suggested in the past that we may eventually want to look into using such free lists for `[]byte` across the codebase, but never got around to investigating such an implementation. It's great that you've gone ahead and implemented these gc/malloc optimizations. In testing my segwit PR on segnet, I've seen some rather large bursts caused by serialization/deserializing some of the mega-blocks I've created which have up to 6k+ transactions. This PR, and the recent related work should serve to dampen such bursts. 

The resulting reduction in allocations across the board due to this PR are very impressive! 

I especially liked the usage of contiguous slices (with the original slice of pointers referencing it), to batch allocations and reduce memory fragmentation. In *retrospect* it's a no-brainer for a memory optimization stand point. I agree that the rest of the `wire` structs should eventually be altered in order to take advantage of the layout optimization. In the same spirit, using a contiguous byte slice for *all* `sigScripts` in a transaction is another excellent optimization. The technique can also be extended to handle allocations for the new `wire.TxWitness` field introduced as part of my segwit PR.

Roasbeef:
There's a typo in the body of the commit message for `wire: Significantly reduce decoding allocs.`: 

> This could be improved upon even further by converting all of the **ata**
structures...

Should be: 
> This could be improved upon even further by converting all of the **data**
structures...

wire/msgtx.go
davecgh: Fair point.  I was being overly explicit and you're right to note it's not needed.  Changed.

davecgh:
@Roasbeef: Thanks for the thorough review!  I've fixed the commit message typo and addressed your comments.

stevenroose:
This is interesting. Is this a technique typical for Go? Or might this improve things in other languages too?

davecgh:
@stevenroose A couple of techniques used here are pretty typical.  Namely, the idea of using a free list and preferring larger contiguous allocations.

However, the techniques I used in the script allocation and rerouting the public API pointers to an internal contiguous slice I have not seen anywhere else in the Go ecosystem.  That isn't to say the idea don't exist elsewhere, but I haven't read about them in any of the numerous articles I've read.

stevenroose:
I think I'm not familiar enough with Go channels to make sense of it all. It seems strange that when you allocate a series of byte buffers of undefined length reduces the amount of allocations. How much space is pre-allocated for these 12500 buffers if the size of these buffers is not known? 

Anyways, it's not you task to teach me Go, I'll read myself into it at some point :) And btw, I spotted a minor typo on the comments here: https://github.com/btcsuite/btcd/pull/684/files#diff-86047f0a6960e8232783735fbda7b70eR130
"maintined" should be "maintained"

davecgh:
Thanks for pointing out the typo.  It has been fixed.

Each of the buffers are allocated at 512 bytes (`freeListMaxScriptSize` in the code).  In the case the script is larger than that, it simply allocated a new buffer as it normally would at whatever size is needed.  The vast majority of scripts are less than that amount, so they fit nicely within the temporary buffer until the final size is determined.  Once the final size is determined, a single allocation is made of that size and all of the script are packed/copied into it.  At that point, the script locations withing the transaction data structs are updated to subslice into the single contiguous buffer and the original script buffers (those that have a cap of 512 bytes) are returned back to the pool for the next transaction.

Graphically, it's doing this:
```
(Script Pool)
     |  Borrow buffers
     v
[sig1][sig2][sig3][...][pk1][pk2][...]  -->(single alloc, copy)  [sig1sig2sig3...pk1pk2...]
     |  Return Buffers
     v
(Script Pool)
```

So, you can see the end result is a single allocation while the

stevenroose:
Hmm, I think that I got that :) What I didn't get is how the allocation of the script pool knows that it should allocate `12500 * 512` bytes. It's done like [below](https://github.com/btcsuite/btcd/pull/684/files#diff-86047f0a6960e8232783735fbda7b70eR132), but how can it know the size of the byte arrays that will be put into it?
```go
var scriptPool scriptFreeList = make(chan []byte, freeListMaxItems)
```
The size of that buffers is set [here](https://github.com/btcsuite/btcd/pull/684/files#diff-86047f0a6960e8232783735fbda7b70eR103) as I understand it:
```go
buf = make([]byte, freeListMaxScriptSize)
```

Anyways, the results are there and I will stop wasting your time.

davecgh:
Ah, I see what you're asking.  It allocates the script buffers on the fly until the pool hits the max by using a buffered channel.  Borrowing a script buffer either pulls it out of the buffered channel queue (when there are any available) or creates a new one.  Returning a buffer either puts it back into the channel queue or lets it go to the garbage collector if the queue is already full.

davecgh:
Rebased to the latest master and squashed review updates into the relevant commits.

PR #685, opened 2016-04-27
mining: Correctly format log messages.

jongillham:
Type feePerKB is an int64 and therefore is not correctly interpreted by format verb f.


jrick:
ok

davecgh:
Nice catch!

OK

Issue #649, opened 2016-03-17
btcd using 60% of memory, bitcoind 4%.


dan-da:
This is from top.  Same machine.   48 gigs ram.   Both have been running for weeks.

29797 btcd      20   0 33.157g 0.028t   5176 S  29.2 60.5  78941:49 btcd                                         
 2667 bitcoin   20   0 5047952 1.833g   6128 S   9.9  3.9   5034:52 bitcoind                                     

btcd version 0.11.1-beta

Is there a config option I should be using to limit mem usage somehow?

davecgh:
How many entries do you have in the memory pool?  I suspect that's the issue, but would like to confirm.

dan-da:
$ ./btcctl getmempoolinfo
{
  "size": 73902,
  "bytes": 879551016
}

tb00:
My general experience is that btcd uses less memory than bitcoind. This is my personal experience with my nodes similar in configuration, your mileage may vary of course.

Is there a significant difference between mempoolinfo from your btcd versus bitcoind instances? 

If no, then this response will not be relevant for your case. If yes, maybe they differ in which transactions they accept - maybe the mempool of your btcd holds many low fee transactions which might make their way into blocks very slowly. Many nodes are operated with customized settings controlling which transactions are accepted into the mempool and under which condition they are included in blocks.

Version 0.12.0 of btcd offers the following configuration options which I believe are not present in 0.11.1:
- `limitfreerelay`: rate-limiting free transactions could cut the rate at which your mempool grows
- `minrelaytxfee`: allows setting the minimum transaction fee that will be considered non-zero

These should be compared with similar parameters of your bitcoind and tuned carefully if needed.

dan-da:
Ok, I just upgraded to 0.12.0 and restarted.  I didn't change any config options.    At present it is using 0.3% of memory.   I'll watch what it does.

One thing about this node that could differ from some is that its primary purpose is to provide the searchrawtransactions API, and it receives some heavy calls that return thousands of transactions.

Most likely the issue is/was related to the mempool and not API caching or leak.  I just mention it as a data point.

dan-da:
I just remembered to check up on btcd memory usage.  It's been 18 days.

bitcoind is still using about 5.2% of available ram, btcd has grown to 15.3%.

22381 btcd      20   0 10.518g 7.214g   5668 S  24.5 15.3  10186:45 btcd                                         
[...]

PR #686, opened 2016-04-27
main: Limit garbage collection percentage.

davecgh:
This reduces the target ratio of freshly allocated data to live data to 10% in order to limit excessive overallocations by the garbage collector during data bursts such as processing complex blocks or rapidly receiving a lot of large transactions.

This is part of the work towards addressing #649.

### DIFF
--- a/btcd.go
+++ b/btcd.go
@@ -11,6 +11,7 @@ import (
 	_ "net/http/pprof"
 	"os"
 	"runtime"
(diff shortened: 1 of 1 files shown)


dajohi:
OK

dajohi:
My btcd usually sits around 490M-500M memory usage.  With this PR after 6 days, btcd is using 251M.

PR #688, opened 2016-05-04
btcd: handle signal SIGTERM

dajohi:
When an OS reboots or shuts down, it sends all processes SIGTERM before
sending SIGKILL.  This allows btcd to do a proper shutdown which most
importantly closes the database.

### DIFF
--- a/signal.go
+++ b/signal.go
@@ -16,6 +16,10 @@ var interruptChannel chan os.Signal
 // to be invoked on SIGINT (Ctrl+C) signals.
 var addHandlerChannel = make(chan func())
 
+// signals defines the default signals to catch in order to do a proper
+// shutdown.
+var signals = []os.Signal{os.Interrupt}
+
 // mainInterruptHandler listens for SIGINT (Ctrl+C) signals on the
 // interruptChannel and invokes the registered interruptCallbacks accordingly.
 // It also listens for callback registration.  It must be run as a goroutine.
@@ -32,16 +36,17 @@ func mainInterruptHandler() {
 
 	for {
 		select {
-		case <-interruptChannel:
+		case sig := <-interruptChannel:
 			// Ignore more than one shutdown signal.
 			if isShutdown {
-				btcdLog.Infof("Received SIGINT (Ctrl+C).  " +
-					"Already shutting down...")
+				btcdLog.Infof("Received signal (%s).  "+
+					"Already shutting down...", sig)
 				continue
 			}
 
 			isShutdown = true
-			btcdLog.Infof("Received SIGINT (Ctrl+C).  Shutting down...")
+			btcdLog.Infof("Received signal (%s).  Shutting down...",
+				sig)
 
 			// Run handlers in LIFO order.
 			for i := range interruptCallbacks {
@@ -74,7 +79,7 @@ func addInterruptHandler(handler func()) {
 	// all other callbacks and exits if not already done.
 	if interruptChannel == nil {
 		interruptChannel = make(chan os.Signal, 1)
-		signal.Notify(interruptChannel, os.Interrupt)
+		signal.Notify(interruptChannel, signals...)
 		go mainInterruptHandler()
 	}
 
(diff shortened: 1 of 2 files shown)


signal.go
```diff
@@ -32,16 +33,17 @@ func mainInterruptHandler() {
...
 	for {
 		select {
-		case <-interruptChannel:
+		case sig := <-interruptChannel:
 			// Ignore more than one shutdown signal.
 			if isShutdown {
-				btcdLog.Infof("Received SIGINT (Ctrl+C).  " +
-					"Already shutting down...")
+				btcdLog.Infof("Received signal (%s).  "+
+					"Already shutting down...", sig.String())
```
davecgh: Since sig is already a `fmt.Stringer`, there is no need to explicitly call `.String()`.

The same is true for the next one.

davecgh:
The `syscall` package is not portable and as such not all Operating Systems will provide the constant.

This would need to be indirected out somehow such that the OS-specific handling can be provided.

jrick:
```
// signals_unix.go
//+build linux darwin freebsd openbsd netbsd
package main
import "signal"
var signals []os.Signal = []os.Signal{os.Interrupt, syscall.SIGTERM}
```

```
// signals_windows.go
package main
import "signal"
var signals []os.Signal = []os.Signal{os.Interrupt, syscall.SIGTERM}
```

```
// signals_portable.go
//+build !windows,!linux,!darwin,!freebsd,!openbsd,!netbsd
package main
import "signal"
var signals []os.Signal = []os.Signal{os.Interrupt}
```

Then use `signal.Notify(interruptChannel, signals...)`

davecgh:
The latest code which defines the portable signals right in `signal.go` and overrides them in `signalsigterm.go` is nice and clean.

We need to test it on Windows and OSX since Travis is testing neither, but I like the approach and it reads well.

signalsigterm.go
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2016 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+// +build darwin linux windows unix
```
jrick: is unix a recognized build tag? i thought this would need to explicitly add each unix OS.

davecgh:
I believe the list of supported `GOOS` is currently: 

darwin, dragonfly, freebsd, linux, netbsd, openbsd, plan9, solaris, windows

jrick:
Tested on windows.

It does not run the interrupt handlers when the console window receives the WM_CLOSE signal, either when manually closed or during shutdown.

davecgh:
After a bit of research, I believe the only way to catch shutdown/reboot and closing of the window is via the [SetConsoleCtrlHandler](https://msdn.microsoft.com/en-us/library/windows/desktop/ms686016%28v=vs.85%29.aspx) function in the Win API.

That means it would be necessary to create a windows-specific file (something like `signalsigterm_windows.go`) that makes use of `syscall.LoadLibrary` and `syscall.GetProcAddress` in order to get a reference to the [SetConsoleCtrlHandler](https://msdn.microsoft.com/en-us/library/windows/desktop/ms686016%28v=vs.85%29.aspx) function and invoke it accordingly.

jrick:
Per that MSDN documentation, since Windows 7, only CTRL_C_EVENT and CTRL_BREAK_EVENT are handled by SetConsoleCtrlHandler.  These signals are already caught by the runtime and used to spoof the Unix SIGINT signal.

In order to catch the equivalent of the Unix SIGTERM, the application would need to create a hidden window in order to receive the WM_QUERYENDSESSION and WM_ENDSESSION signals.

davecgh:
It will definitely take a bit of work to make it handle those events on Windows.

At any rate, I think for now we should just go ahead and remove windows from the build tag and get this PR merged.  I say that because it will improve the functionality on other operating systems while keeping the same behavior on Windows.

dajohi:
Removed windows.  Ready for merge.

jrick:
ok

davecgh:
OK

PR #689, opened 2016-05-06
build: Convert project to use glide.

### LINKED ISSUE #676: Add vendoring support
Hi

Since Go 1.6, it is recommended to vendor the dependencies:

https://golang.org/doc/go1.6#go_command

Could you do this in this project? This way, the exact version of the dependencies is fixed, and in my project, I only have to directly vendor btcd, and not manually transitively vendor all its dependencies.

davecgh:
This converts the project to allow `btcd` to be used with the `glide` package manager in order to provide stable and reproducible builds without the user having to jump through all of the hoops as they do today.

It consists of adding a `glide.yaml` file which identifies the project dependencies and locations along with a `glide.lock` file which contains the complete dependency tree pinned to specific versions.  Glide uses these files to download the packages (or updates) to a local `vendor` directory and checkout the correct pinned versions.  The go tool, in turn, is used to build/install `btcd` and will use the pinned versions in the `vendor` directory.

This also updates TravisCI to build using glide, removes some of the exceptions in the lint checks which are no longer required, and updates the `README.md` with the new instructions needed to build the project with `glide`.

Closes #676.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ btcd.db
 # Folders
 _obj
 _test
+vendor
 
 # Architecture specific extensions/prefixes
 *.[568vq]
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,11 @@ sudo: false
 before_install:
   - gotools=golang.org/x/tools
 install:
-  - go get -d -t -v ./...
+  - go get -v github.com/Masterminds/glide
+  - glide install
   - go get -v $gotools/cmd/cover
-  - go get -v github.com/bradfitz/goimports
   - go get -v github.com/golang/lint/golint
 script:
   - export PATH=$PATH:$HOME/gopath/bin
+  - export GO15VENDOREXPERIMENT=1
   - ./goclean.sh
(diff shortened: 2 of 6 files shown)


jcvernaleo:
When I run glide install I get:
```
jcv@triforce btcd $ glide install
[WARN] Lock file may be out of date. Hash check of YAML failed. You may need to 
run 'update'
```

jcvernaleo:
Other than that all seems to work on Bitrig, Linux, and OSX.

jrick:
I see that as well.  After running glide update:

```diff
diff --git a/glide.lock b/glide.lock
index dca29b8..2e6f905 100644
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 7fe92afbacb875acf2e7c13c431269d70f992e43c791028a580974b2997e80cc
-updated: 2016-05-05T21:56:18.3094742-05:00
+hash: 9b8ff781a12daad991983e9a421a320e905eecc4e9ff0b0643e790f472bc78c8
+updated: 2016-05-06T09:36:18.6338973-04:00
 imports:
 - name: github.com/btcsuite/btclog
   version: f96df2375f37300305f329b8e5258764b4f19a7f
```

So none of the deps are actually wrong, just the hash wasn't updated properly. (why?)

davecgh:
I updated the `glide.lock` file.  The warning should be gone now.

jrick:
ok

PR #692, opened 2016-05-06
Do not add the same transaction back to the priority heap

### LINKED ISSUE #691: Transactions with the same dependent tx can be readded to prio queue multiple times
Issue is in this chunk of code of mining.go.

```
			for e := deps.Front(); e != nil; e = e.Next() {
				// Add the transaction to the priority queue if
				// there are no more dependencies after this
				// one.
				item := e.Value.(*txPrioItem)
				delete(item.dependsOn, *tx.Sha())
				if len(item.dependsOn) == 0 {
					heap.Push(priorityQueue, item)
				}
```

If the same transaction is referenced multiple times in the input, it will be pushed on to the prio queue multiple times in this loop. Then there are many needless checks after this, because after the first time the transaction is added it will be rechecked but then fail due to double spend since tx is already present in tx store.

cjepson:
A bug that caused the same transaction to be added to the
mining priority heap when a transaction spends multiple
outputs from the same input transaction was fixed by first
checking to see if the transaction already exists in the
priority heap.

Fixes #691.

### DIFF
--- a/mining.go
+++ b/mining.go
@@ -67,6 +67,20 @@ type txPriorityQueue struct {
 	items    []*txPrioItem
 }
 
+// Contains checks to see whether the passed txPrioItem currently exists within
+// the priority queue.
+func (pq *txPriorityQueue) Contains(prioItem *txPrioItem) bool {
+	itemHash := prioItem.tx.Sha()
+	for i := range pq.items {
+		thisItemHash := pq.items[i].tx.Sha()
+		if *itemHash == *thisItemHash {
+			return true
+		}
+	}
+
+	return false
+}
+
 // Len returns the number of items in the priority queue.  It is part of the
 // heap.Interface implementation.
 func (pq *txPriorityQueue) Len() int {
@@ -695,7 +709,7 @@ mempoolLoop:
 				// one.
 				item := e.Value.(*txPrioItem)
 				delete(item.dependsOn, *tx.Sha())
-				if len(item.dependsOn) == 0 {
+				if len(item.dependsOn) == 0 && !priorityQueue.Contains(item) {
 					heap.Push(priorityQueue, item)
 				}
 			}


mining.go
```diff
@@ -695,7 +709,7 @@ mempoolLoop:
 				// one.
 				item := e.Value.(*txPrioItem)
 				delete(item.dependsOn, *tx.Sha())
-				if len(item.dependsOn) == 0 {
+				if len(item.dependsOn) == 0 && !priorityQueue.Contains(item) {
```
davecgh: I think looping the entire priority queue here will cause performance issues.  Keep in mind that there often by 30,000+ transaction in the pq after the node has been running for a while.  Due to that, I believe it would be a lot more efficient to simply keep track of the hashes of the added transactions (priority queue items) in a map and avoid adding them that way.

The following diff should do the trick and remove the need for the `Contains` function as well:

```diff
diff --git a/mining.go b/mining.go
index ccf3d34..5955931 100644
--- a/mining.go
+++ b/mining.go
@@ -689,14 +689,22 @@ mempoolLoop:
 		// have any other unsatisified dependencies) to the priority
 		// queue.
 		if deps != nil {
+			pushedTxns := make(map[wire.ShaHash]struct{})
 			for e := deps.Front(); e != nil; e = e.Next() {
 				// Add the transaction to the priority queue if
 				// there are no more dependencies after this
-				// one.
+				// one and it wasn't already added via another
+				// dependency.
 				item := e.Value.(*txPrioItem)
 				delete(item.dependsOn, *tx.Sha())
 				if len(item.dependsOn) == 0 {
+					itemHash := item.tx.Sha()
+					if _, ok := pushedTxns[*itemHash]; ok {
+						continue
+					}
+
 					heap.Push(priorityQueue, item)
+					pushedTxns[*itemHash] = struct{}{}
 				}
 			}
 		}
```

davecgh: After thinking about this a bit more, I think it would be even better to simply stop duplicate dependencies from ever being added to the dependency list anyways.  That would avoid the need for allocating a bunch of new maps here in this code.

Since the list of dependencies for a given transaction are extremely unlikely to get massively large, unlike the entire priority queue, I believe an iterative approach like you have in this PR will work well and avoid the need for extra map allocations.

The following diff would approach it that way:

```diff
$ git diff
diff --git a/mining.go b/mining.go
index ccf3d34..55d11bb 100644
--- a/mining.go
+++ b/mining.go
@@ -181,6 +181,18 @@ type BlockTemplate struct {
        ValidPayAddress bool
 }

+// containsPrioItem returns whether or not the passed list of dependcies
+// contains the transaction associated with the passed priority queue item.
+func containsPrioItem(deps *list.List, item *txPrioItem) bool {
+       for e := deps.Front(); e != nil; e = e.Next() {
+               listItem := e.Value.(*txPrioItem)
+               if *listItem.tx.Sha() == *item.tx.Sha() {
+                       return true
+               }
+       }
+       return false
+}
+
 // mergeUtxoView adds all of the entries in view to viewA.  The result is that
 // viewA will contain all of its original entries plus all of the entries
 // in viewB.  It will replace any entries in viewB which also exist in viewA
@@ -505,7 +517,9 @@ mempoolLoop:
                                        depList = list.New()
                                        dependers[*originHash] = depList
                                }
-                               depList.PushBack(prioItem)
+                               if !containsPrioItem(depList, prioItem) {
+                                       depList.PushBack(prioItem)
+                               }
                                if prioItem.dependsOn == nil {
                                        prioItem.dependsOn = make(
                                                map[wire.ShaHash]struct{})
```

cjepson:
The map is really the way to go, as we've seen in memory pool with dependencies. I can try to update my PR to reflect this later.

davecgh:
Closing in favor of #759 which implements the discussed map approach.

PR #693, opened 2016-05-07
addblock: Add support for generating indexes.

davecgh:
This adds two new flags, `--txindex` and `--addrindex`, to the `addblock` utility which mirror the flags on `btcd`.  They serve to to specify that the transaction index and/or address index, respectively, should be
built while importing from the bootstrap file.

This is technically not 100% required since `btcd` will build the indexes on the first load (when enabled) if they aren't already built, however it is much faster to build the indexes as the blocks are being validated (particularly for the address index), so this makes the capability available.

### DIFF
--- a/cmd/addblock/import.go
+++ b/cmd/addblock/import.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcd/blockchain"
+	"github.com/btcsuite/btcd/blockchain/indexers"
 	"github.com/btcsuite/btcd/database"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcutil"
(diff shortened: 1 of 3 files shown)


cmd/addblock/import.go
```diff
@@ -295,9 +296,40 @@ func (bi *blockImporter) Import() chan *importResults {
...
+			log.Infof("Transaction index enabled because it is " +
+				"required by the address index")
+			cfg.TxIndex = true
+		} else {
+			log.Info("Transaction index is enabled")
+		}
+		indexes = append(indexes, indexers.NewTxIndex(db))
+	}
+	if cfg.AddrIndex {
+		log.Info("Address index is enabled")
```
jcvernaleo: Wouldn't it make more sense to check this option and show it in the logs before the tx index check since the txindex gets turned on by this one if you haven't already set it?

jcvernaleo:
Other than that one comment above this would be OK by me.

cmd/addblock/import.go
davecgh: That's doable, although I think you would need the extra check one way or the other in order for the logging message to say when the `txindex` has been enabled due to the `address` index versus due to the config setting.  The way it is now, the transaction index state is always listed before the address index state. I sort of like showing the index that is a dependency first, but I don't guess it would matter much to reverse them.

For example, the reversed approach might look like:

```Go
var forcedTxIndex bool
if cfg.AddrIndex {
	forcedTxIndex = true
	cfg.TxIndex = true
	log.Info("Address index is enabled")
	indexes = append(indexes, indexers.NewAddrIndex(db, activeNetParams))
}
if cfg.TxIndex {
	if forcedTxIndex {
		log.Infof("Transaction index enabled because ...")
	} else {
		log.Info("Transaction index is enabled")
	}
	indexes = append(indexes, indexers.NewTxIndex(db))
}
```

There isn't much difference to my eyes.

jcvernaleo: It wasn't so much the extra check that bothered me (since that seemed reasonable).  I was just thinking of the logical flow of the output messages (as in saying txindex enabled because of addrindex before saying addrindex enabled felt a little odd to me).  It isn't a huge deal either way so I'd still OK this PR either way.

PR #696, opened 2016-05-15
btcec: Minor correction to GenerateSharedSecret documentation.

IngCr3at1on:
This really bothered me when I read it for some reason...


davecgh:
Nice catch.  Thanks.

OK

PR #698, opened 2016-05-21
Update installation instructions using Glide

stevenroose:
The main README.md file had newer installation that uses Glide, while these did not.


davecgh:
Nice catch.  Thanks for the PR!

OK

Issue #632, opened 2016-02-20
[rpcserver] Improve JSON-RPC compatibility


tb00:
I was having trouble getting CKPool to work with btcd. I isolated multiple issues and narrowed down the root causes. The pool software builds on the behavior of bitcoind/Core which is not mirrored in btcd:

 - Responding to RPC request without an "id" member
 - Terminating the RPC response with a newline

I made changes to CKPool around how JSON-RPC is handled but then figured making btcd more interoperable in this area could be beneficial for others as it would avoid similar compatibility issues with other software that relies on the behavior of bitcoind's JSON-RPC implementation.

I thoroughly read the JSON-RPC specification (1.0 & 2.0) and performed tests with both Core and btcd before raising this issue. If you agree that mirroring Core behavior is beneficial, I will send a PR.

davecgh:
We do generally strive for compatibility where it makes sense.  To that end, and I don't remember what they are now off the top of my head, but there are some cases with Core uses a sort of in-between JSON-RPC 1.0 and JSON-RPC 2.0 and doesn't conform to either spec perfectly.  We had to avoid using the stdlib jsonrpc package specifically because of that in the name of compatibility.

In the case of the id member, do keep in mind the btcd support JSON-RPC 2.0 notifications (via websockets) which the spec specifically states "A Notification is a Request object without an "id" member."  I'm not sure of the specific cases you're hitting in regards to the id, but I thought that was worth mentioning, because there are cases where the id is intentionally omitted.

I'd certainly appreciate a PR that makes it more compatible, so long as it doesn't break JSON-RPC 2.0 notifications via Websockets of course.

davecgh:
Oh I should also note, if you're in the JSON-RPC code anyways and have any desire to work on it, it would be ideal to address #329 as well.  I believe there are several different mining/pool software implementations which make use of JSON-RPC 2.0 batching.

jrick:
> A Notification is a Request object without an "id" member. A Request object that is a Notification signifies the Client's lack of interest in the corresponding Response object, and as such no Response object needs to be returned to the client. The Server MUST NOT reply to a Notification, including those that are within a batch request.

http://www.jsonrpc.org/specification#notification

tb00:
[...]

PR #701, opened 2016-05-24
rpcserver: Improve JSON-RPC compatibility

tb00:
This PR aims to improve JSON-RPC compatibility with software that relies on the behavior of the reference implementation. The PR itself consists of two commits which resolve two separate compatibility issues reported under #632. 

### The first commit

In order to avoid interoperability issues with software that relies on Core's behavior, terminate HTTP POST JSON-RPC responses with a newline.

Core always terminates the response with a newline. While this is not mandated by the specification, the lack of a closing newline may cause undesired side effects with software sharing code for handling JSON-RPC over HTTP and raw socket transports (e.g. stratum). It certainly causes issues with ckpool.

This commit intentionally limits the scope of this compatibility improvement to only affect HTTP POST, not WebSocket.

### The second commit

Avoid compatibility issues with software that relies on the behavior of bitcoind's JSON-RPC implementation, which does not strictly conform to any version of the JSON-RPC specification.

The JSON-RPC 1.0 spec defines that notifications must have their "id" set to null and states that notifications do not have a response.

A JSON-RPC 2.0 notification is a request with "json-rpc":"2.0", and without an "id" member. The specification states that notifications must not be responded to. JSON-RPC 2.0 permits the null value as a valid request id, therefore such requests are not notifications.

Bitcoin Core serves requests with "id":null or even an absent "id", and responds to such requests with "id":null in the response.

Btcd does not respond to any request without and "id" or with "id":null, regardless the indicated JSON-RPC protocol version.

In order to avoid compatibility issues with software that relies on Core's behavior, this commit implements "quirks mode" as follows:
 - quirks mode can be enabled via configuration (disabled by default)
 - If no JSON-RPC version is indicated in the request, accept and respond to request with "id":null
 - If no JSON-RPC version is indicated in the request, accept and respond to requests without an "id" member
 - In both cases above, use "id":null in the response
 - Do not respond to request without an "id" or with "id":null when JSON-RPC version is indicated in the request (process as notification)

The proposed change allows "quirks mode" to be enabled consistently for both or none of HTTP POST and WebSocket transports, however, it can be changed to only affect HTTP POST.

As a closing remark, I have modified ckpool to always include an "id" member and submitted a PR. Still I decided to share this as it might be useful for others when facing similar interoperability issues.

### DIFF
--- a/config.go
+++ b/config.go
@@ -107,6 +107,7 @@ type config struct {
 	RPCMaxClients        int           `long:"rpcmaxclients" description:"Max number of RPC clients for standard connections"`
 	RPCMaxWebsockets     int           `long:"rpcmaxwebsockets" description:"Max number of RPC websocket connections"`
 	RPCMaxConcurrentReqs int           `long:"rpcmaxconcurrentreqs" description:"Max number of concurrent RPC requests that may be processed concurrently"`
+	RPCQuirks            bool          `long:"rpcquirks" description:"Mirror some JSON-RPC quirks of Bitcoin Core -- NOTE: Discouraged unless interoperability issues need to be worked around"`
 	DisableRPC           bool          `long:"norpc" description:"Disable built-in RPC server -- NOTE: The RPC server is disabled by default if no rpcuser/rpcpass or rpclimituser/rpclimitpass is specified"`
 	DisableTLS           bool          `long:"notls" description:"Disable TLS for the RPC server -- NOTE: This is only allowed if the RPC server is bound to localhost"`
 	DisableDNSSeed       bool          `long:"nodnsseed" description:"Disable DNS seeding for peers"`
--- a/doc.go
+++ b/doc.go
@@ -50,6 +50,9 @@ Application Options:
       --rpcmaxclients=      Max number of RPC clients for standard connections
                             (10)
       --rpcmaxwebsockets=   Max number of RPC websocket connections (25)
+      --rpcquirks           Mirror some JSON-RPC quirks of Bitcoin Core -- NOTE:
+                            Discouraged unless interoperability issues need to
+                            be worked around
       --norpc               Disable built-in RPC server -- NOTE: The RPC server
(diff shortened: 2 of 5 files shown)


config.go
```diff
@@ -99,6 +99,7 @@ type config struct {
 	RPCKey             string        `long:"rpckey" description:"File containing the certificate key"`
 	RPCMaxClients      int           `long:"rpcmaxclients" description:"Max number of RPC clients for standard connections"`
 	RPCMaxWebsockets   int           `long:"rpcmaxwebsockets" description:"Max number of RPC websocket connections"`
+	RPCQuirks          bool          `long:"rpcquirks" description:"Mirror some JSON-RPC quirks of the reference implementation -- NOTE: Discouraged unless interoperability issues need to be worked around"`
```
davecgh: Would you mind making this say `of Bitcoin Core`?  It's more descriptive for people that are unaware.

doc.go
```diff
@@ -50,6 +50,9 @@ Application Options:
       --rpcmaxclients=      Max number of RPC clients for standard connections
                             (10)
       --rpcmaxwebsockets=   Max number of RPC websocket connections (25)
+      --rpcquirks           Mirror some JSON-RPC quirks of the reference
```
davecgh: Same here.

davecgh:
Thanks for the PR.  Other than the one comment I made this looks good.

tb00:
Done. I changed the wording in the two locations you mentioned plus an additional one in `sample-btcd.conf`. I will squash and rebase on top of latest master if you confirm this is good to go..

dajohi:
@tb00 Can you rebase?

tb00:
Rebased on latest master. Should I squash the two commits?

davecgh:
@tb00: The two commits are fine.  I just merged it.  Thanks and sorry for the delay on merging it.  It sort of fell between the cracks with everything else going on.

PR #705, opened 2016-06-06
mining: Stop transactions losing their dependants.

jongillham:
This fixes a bug where a transaction would loose reference to other transactions dependant on it when being considered for inclusion in a new block template. The issue only occurs when the transaction being considered triggers a change of priority queue ordering to ordering by fee. It results in none of the dependant transactions being considered for inclusion in the new block template.

### DIFF
--- a/mining.go
+++ b/mining.go
@@ -554,12 +554,8 @@ mempoolLoop:
 		prioItem := heap.Pop(priorityQueue).(*txPrioItem)
 		tx := prioItem.tx
 
-		// Grab the list of transactions which depend on this one (if
-		// any) and remove the entry for this transaction as it will
-		// either be included or skipped, but in either case the deps
-		// are no longer needed.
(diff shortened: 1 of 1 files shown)


jongillham:
The mining code could probably do with a unit test to demonstrate the problem. However this might prove somewhat tricky with the arguments required to call `NewBlockTemplate`. I see someone might be putting some effort into moving the mining code to its own package which would hopefully make testing easier.

davecgh:
Would you rebase this please?  It won't merge cleanly, although the code change is fine.

jongillham:
Rebased.

davecgh:
I wanted to note the Travis CI build failure here is a false positive.  Something went wrong on the Travis side with the build where the build never actually ran and restarting it never even tried to execute it.  Now the original branch has been removed, so it's not available anymore in order to trigger a rebuild.

Issue #706, opened 2016-06-08
Normalizing bug with field.go when the number > P but the first word is less than P's first word.


jimmysong:
Currently, for the normalization function in field.go, there is a final step for subtracting P from the number. This step should occur if and only if the number is greater than P. However, on closer inspection, this only happens if the first and second words of the number being normalized are BOTH greater than or equal to the first and second numbers of P. The situation where the first word is actually less but the second word is larger results in a normalized number that's greater than P, yet not properly modulo'ed.

This issue is a lot easier to demonstrate than describe. I am adding a pull request which adds two test cases to TestNormalize one of which fails with the current implementation.

davecgh:
Thanks for the report.  I've confirmed this issue.  I'll push a PR shortly with a fix that is constant time.  I'll also squash and pull in your commits from #707 as a part of that PR so master stays green once merged.

PR #707, opened 2016-06-08
Added a test for field.go that demonstrates a bug

jimmysong:
This adds two tests which should demonstrate the bug in field.go described in #706.


davecgh:
Thanks for the tests.  I've squashed your commits to a single commit and included it as a part of PR #711 to ensure you get credit for them.  Closing this in favor of that PR.

PR #708, opened 2016-06-08
Bug corrected

jimmysong:
This also adds the correction that would solve #706 introduced in the other PR.

Note that this is probably not an acceptable solution as it's not constant-time. However, I wanted to demonstrate what the correct code would look like.


jimmysong:
See #707 and #706

stevenroose:
@jimmysong Is it impossible to solve in constant time? (I didn't completely look into the problem, but noticed your note on the second commit.)

jimmysong:
@stevenroose There probably is a way, but I'm not familiar enough with constant-time programming techniques to be able to do it at the moment.

raedah:
This pull request needs a better title.

davecgh:
Thanks for demonstrating the issue and fix.  As the initial PR description states, this implementation is not constant time.  I'll push a PR with a constant time fix shortly and close this one once that has been done.

davecgh:
Closing in favor of the constant time fix in #711.

PR #711, opened 2016-06-20
btcec: Ensure reduction when > P in all cases.

### LINKED ISSUE #706: Normalizing bug with field.go when the number > P but the first word is less than P's first word.

davecgh:
As noted in issue #706, the existing code had an issue where the normalized result was > P when both the first and second words of the field representation being normalized were BOTH greater than or equal to the first and second words of P.  Although this condition is rare in practice, it needs to be handled properly.

This resolves the issue by comparing the low words in the final reduction step against the normalized low order prime bits to ensure the final subtraction occurs correctly any time they're > P.  This approach
retains the constant time property as well.

Also, it merges and squashes in the additional commits with tests provided by @jimmysong in #707 to help prevent regressions.  It is being done this way to ensure master stays green.

Fixes #706.

### DIFF
--- a/btcec/field.go
+++ b/btcec/field.go
@@ -1,5 +1,5 @@
-// Copyright (c) 2013-2014 The btcsuite developers
-// Copyright (c) 2013-2014 Dave Collins
+// Copyright (c) 2013-2016 The btcsuite developers
+// Copyright (c) 2013-2016 Dave Collins
 // Use of this source code is governed by an ISC
 // license that can be found in the LICENSE file.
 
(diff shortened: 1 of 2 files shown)


davecgh:
For additional testing, I've done a full download of the chain with checkpoints disabled to ensure every signature in the chain verifies as expected with these changes.

cjepson:
OK

PR #712, opened 2016-06-21
Mempool-only BIP 113: Median Time Past for lock-time calculations

### LINKED ISSUE #645: Implement BIP0068 and BIP0113
BIP0068 and BIP113 are nearing merge in Core and should be implemented for btcd.

https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki
https://github.com/bitcoin/bips/blob/master/bip-0113.mediawiki

Roasbeef:
This PR implements [BIP 113](https://github.com/bitcoin/bips/blob/master/bip-0113.mediawiki) which modifies the time-lock threshold Bitcoin nodes use to determine if a transaction's lock-time has been satisfied or not. Before BIP 113, simply the time stamp within the blockheader of the block including a candidate transaction was consulted. However, this allows miners to fiddle with the timestamp on their blocks in order to include transactions which haven't yet matured by wall-clock time, thereby collecting more fees. 

After this PR, the MTP of the prior 11 blocks will be used in order to calculate the currently lock-time threshold for all transactions accepted into the mempool. 

A brief summary of this PR follows: 
 * A new closure within the `mempool`'s Policy config has been added which returns the current MTP as calculated from the tip of the best chain. Introducing such a closure which typically has the value cached should reduce the performance impact of the new MTP checks within the mempool. Otherwise, the MTP would need to be calculated from scratch each time negatively impacting validation efficiency. 
 * Transaction finality checks within the `mempool` now use the MTP of the past 11 blocks. As a result, the `TimeSource` field within the mempool's config has been removed, as MTP is now always used. 

This PR partially fixes #645.

### DIFF
--- a/mempool/policy_test.go
+++ b/mempool/policy_test.go
@@ -7,8 +7,8 @@ package mempool
 import (
 	"bytes"
 	"testing"
+	"time"
 
-	"github.com/btcsuite/btcd/blockchain"
 	"github.com/btcsuite/btcd/btcec"
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
@@ -466,11 +466,11 @@ func TestCheckTransactionStandard(t *testing.T) {
 		},
 	}
 
-	timeSource := blockchain.NewMedianTime()
+	pastMedianTime := time.Now()
 	for _, test := range tests {
 		// Ensure standardness is as expected.
 		err := checkTransactionStandard(btcutil.NewTx(&test.tx),
-			test.height, timeSource, DefaultMinRelayTxFee)
+			test.height, pastMedianTime, DefaultMinRelayTxFee)
 		if err == nil && test.isStandard {
 			// Test passes since function returned standard for a
 			// transaction which is intended to be standard.
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -59,6 +59,11 @@ type Config struct {
 	// the current best chain.
 	BestHeight func() int32
 
+	// MedianTimePast defines the function to use in order to access the
+	// median time past calculated from the point-of-view of the current
+	// chain tip within the best chain.
+	MedianTimePast func() time.Time
+
 	// SigCache defines a signature cache to use.
 	SigCache *txscript.SigCache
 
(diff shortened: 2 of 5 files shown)


Roasbeef:
I've rebased this PR on top the the latest changes to master.

Roasbeef:
I just pushed an integration test using the new `rpctest` package which exercises the intended behavior of BIP 113 both within the mempool and during block validation. 

The test could be made to be a bit terser with some minor additions to the `memWallet` within the `rpctest` package. The addition of manual coin selection (via a `listunspent`-like method) along with sparse signing would allow the `memWallet` to handle the functionality I created by hand within the integration tests.

Roasbeef:
With the latest set of commits, I'm removing the WIP tag from the PR description. 

I consider the commits displayed final, and ready to review. The only lingering item not finished within the talk-list for this PR is the BIP 9 integration. Integration of BIP 9 will likely eliminate the first commit, and slightly modify a few others to guard BIP 113 observance on the proper version bits state.

davecgh:
The BIP9 PR (#755) is in, so this can be updated accordingly.  I personally would rather see this PR (and the BIP68 and BIP112 ones) only deal with the mempool, and then have a separate PR which enables them all via BIP9.

Roasbeef:
I've updated the PR to remove most of the functionality that was in the prior version of the PR. The original tree can be found [here](https://github.com/Roasbeef/btcd/tree/median-time-past-full). What's remaining of the former PR is simply the addition to mempool policy rules which unconditionally use the MTP for all transaction lock-time evaluations.

mempool/policy_test.go
```diff
@@ -466,11 +466,11 @@ func TestCheckTransactionStandard(t *testing.T) {
 		},
 	}
 
-	timeSource := blockchain.NewMedianTime()
+	pastMedianTime := blockchain.NewMedianTime().AdjustedTime()
```
davecgh: This is the the equivalent of `time.Now()`. I'd suggest using that instead because this is the only thing left in the file that depends on `blockchain` now and thus the dependency could be dropped altogether.

Roasbeef:
Thanks! Happy to see this series of PR's finally merged. I've addressed your commit and squashed the fix into the current set of PR's.

PR #713, opened 2016-06-22
Mempool-only BIP 68: relative lock-time via consensus-enforced sequence numbers

### LINKED ISSUE #645: Implement BIP0068 and BIP0113

Roasbeef:
This commit implements [BIP 68](https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki) which repurposes the sequence number as defined within the `wire.TxIn`'s of a transaction. This repurposing allows for consensus enforced _relative_ lock-times based on the set sequence numbers of all inputs within a transaction. Transactions which violate the required input maturity governed by sequence numbers are not allowed into the mempool, rejected when found within a block, and skipped over when assembling blocks. This PR currently only enforces the behavior when accepting candidate transaction to the mempool. 

This PR is dependent on #712 as it rebases on top of the latest changes to that branch. By *default* all checks for input maturity based on sequence numbers must use the MTP of the block prior to the one which includes the referenced output in the case of a relative lock-time expressed in _seconds_. 

A brief summary of the changes in this PR follows: 
  * A new data-structure `SequenceLocks`, along with a new method on `BlockChain` to calculate sequence locks based on a transaction and `utxoView` has have been added to the `blockchain` package. 
    * Sequence locks are essentially the evaluated absolute lock-time based on an attempt to include a transaction within a block, accept to the mempool, or accept as valid within a new block. 
    * Currently in the implementation there's a `true` value in a boolean statement which determines if sequence locks should be enforced or not. Once BIP 9 has been implemented, this should instead be a check to the `versionbits` state to determine if the CSV package has been activated+locked-in. 
    * This function _takes_ a `UtxoViewPoint` rather than generating one internally in order to allow callers to properly populate the utxo view depending on the context. For example, when accepting a transaction into the mempool, any inputs referenced which are in the mempool may potentially need to be considered. Similarly when validating a block, the UTXO's of any transactions within the block, but not yet in the DB also need to be considered.  

* A new function `SequenceLocksActive` similar to `IsFinalizedTransaction` has been added to the `blockchain` package. This function is used to determine if the relative lock-times (sequence locks) for a transaction have reached maturity from the point-of-view of a block height or MTP. 

* When accepting a transaction into the mempool, validating a block, or assembling a new block template relative-time locks are currently _unconditionally_ evaluated. This OK for the mempool case, but when dealing with blocks, BIP 9 `versionbits` state should be examined to ensure we're enforcing the new consensus behavior at the correct time. 

This PR partially fixes #645.

### DIFF
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -828,6 +828,24 @@ func (b *BlockChain) calcSequenceLock(tx *btcutil.Tx, utxoView *UtxoViewpoint,
 	return sequenceLock, nil
 }
 
+// LockTimeToSequence converts the passed relative locktime to a sequence
+// number in accordance to BIP-68.
+// See: https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki
+//  * (Compatibility)
+func LockTimeToSequence(isSeconds bool, locktime uint32) uint32 {
+	// If we're expressing the relative lock time in blocks, then the
+	// corresponding sequence number is simply the desired input age.
+	if !isSeconds {
+		return locktime
+	}
+
+	// Set the 22nd bit which indicates the lock time is in seconds, then
+	// shift the locktime over by 9 since the time granularity is in
+	// 512-second intervals (2^9). This results in a max lock-time of
+	// 33,553,920 seconds, or 1.1 years.
+	return wire.SequenceLockTimeIsSeconds | uint32(locktime>>wire.SequenceLockTimeGranularity)
+}
+
 // getReorganizeNodes finds the fork point between the main chain and the passed
 // node and returns a list of block nodes that would need to be detached from
 // the main chain and a list of block nodes that would need to be attached to
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -511,6 +511,51 @@ func (b *BlockChain) getPrevNodeFromNode(node *blockNode) (*blockNode, error) {
 	return prevBlockNode, err
 }
 
+// relativeNode returns the ancestor block a relative 'distance' blocks before
+// the passed anchor block. While iterating backwards through the chain, any
+// block nodes which aren't in the memory chain are loaded in dynamically.
+//
+// This function MUST be called with the chain state lock held (for writes).
+func (b *BlockChain) relativeNode(anchor *blockNode, distance uint32) (*blockNode, error) {
+	var err error
+	iterNode := anchor
+
+	err = b.db.View(func(dbTx database.Tx) error {
+		// Walk backwards in the chian until we've gone 'distance'
+		// steps back.
+		for i := distance; i > 0; i-- {
+			switch {
+			// If the parent of this node has already been loaded
+			// into memory, then we can follow the link without
+			// hitting the database.
+			case iterNode.parent != nil:
+				iterNode = iterNode.parent
+
+			// If this node is the genesis block, then we can't go
+			// back any further, so we exit immediately.
+			case iterNode.hash.IsEqual(b.chainParams.GenesisHash):
+				return nil
+
+			// Otherwise, load the block node from the database,
+			// pulling it into the memory cache in the processes.
+			default:
+				iterNode, err = b.loadBlockNode(dbTx,
+					iterNode.parentHash)
+				if err != nil {
+					return err
+				}
+			}
+		}
+
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return iterNode, nil
+}
+
 // removeBlockNode removes the passed block node from the memory chain by
 // unlinking all of its children and removing it from the the node and
 // dependency indices.
@@ -543,52 +588,6 @@ func (b *BlockChain) removeBlockNode(node *blockNode) error {
 		// longer any nodes which depend on the parent hash.
 		if len(b.depNodes[*prevHash]) == 0 {
 			delete(b.depNodes, *prevHash)
-		}
-	}
-
-	return nil
-}
-
-// pruneBlockNodes removes references to old block nodes which are no longer
-// needed so they may be garbage collected.  In order to validate block rules
-// and choose the best chain, only a portion of the nodes which form the block
-// chain are needed in memory.  This function walks the chain backwards from the
-// current best chain to find any nodes before the first needed block node.
-//
-// This function MUST be called with the chain state lock held (for writes).
-func (b *BlockChain) pruneBlockNodes() error {
-	// Walk the chain backwards to find what should be the new root node.
-	// Intentionally use node.parent instead of getPrevNodeFromNode since
-	// the latter loads the node and the goal is to find nodes still in
-	// memory that can be pruned.
-	newRootNode := b.bestNode
-	for i := int32(0); i < b.minMemoryNodes-1 && newRootNode != nil; i++ {
-		newRootNode = newRootNode.parent
-	}
-
-	// Nothing to do if there are not enough nodes.
-	if newRootNode == nil || newRootNode.parent == nil {
-		return nil
-	}
-
-	// Push the nodes to delete on a list in reverse order since it's easier
-	// to prune them going forwards than it is backwards.  This will
-	// typically end up being a single node since pruning is currently done
-	// just before each new node is created.  However, that might be tuned
-	// later to only prune at intervals, so the code needs to account for
-	// the possibility of multiple nodes.
-	deleteNodes := list.New()
-	for node := newRootNode.parent; node != nil; node = node.parent {
-		deleteNodes.PushFront(node)
-	}
-
-	// Loop through each node to prune, unlink its children, remove it from
-	// the dependency index, and remove it from the node index.
-	for e := deleteNodes.Front(); e != nil; e = e.Next() {
-		node := e.Value.(*blockNode)
-		err := b.removeBlockNode(node)
-		if err != nil {
-			return err
 		}
 	}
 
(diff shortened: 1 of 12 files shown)


blockchain/chain.go
```diff
@@ -680,6 +680,128 @@ func (b *BlockChain) CalcPastMedianTime() (time.Time, error) {
...
+	b.stateLock.RLock()
+	nextHeight := b.stateSnapshot.Height + 1
+	b.stateLock.RUnlock()
+
+	for _, txIn := range mTx.TxIn {
+		utxo := utxoView.LookupEntry(&txIn.PreviousOutPoint.Hash)
+		if utxo == nil {
+			// TODO(rosbeef): better error message
+			return sequenceLock, fmt.Errorf("utxo not found")
+		}
```
dajohi: Please add which utxo...  "utxo %v not found"

blockchain/process.go
```diff
@@ -33,6 +33,14 @@ const (
 	// without modifying the current state.
 	BFDryRun
 
+	// BFMedianTimePast may be set to indciate that when validating
```
dajohi: indicate

afk11:
I think Core took the approach of getting the CSV-fork implementations merged first, and then merged something activating it much later. It let people hook into script/consensus.h to test the feature (it accepts arbitrary script flags, whereas a node determines them by consensus)

Roasbeef:
@afk11 Indeed. I'm open to taking that route as well once I patch up the holes in the test coverage w.r.t to this PR. 

I'll be shooting to get this PR finalized within the next week. The aforementioned `rpctest` package should be merged soon, allowing us to move ahead with utilizing the package to test all the outstanding soft-fork PR's.

Roasbeef:
I've pushed a "final" set of commits implementing the BIP, various refactors, and a suite of unit+integration tests.

davecgh:
The BIP9 PR (#755) is in, so this can be updated accordingly.  I personally would rather see this PR (and the BIP113 and BIP113 ones) only deal with the mempool, and then have a separate PR which enables them all via BIP9.

davecgh:
As an FYI, in #755, I'm adding a new folder/package for integration tests where I would prefer the tests fr this and the other chain-related tests end up going.  Currently, it doesn't have a primary harness because the tests really needs to start with a fresh chain in order to quickly and easily test the threshold state transitions.  IMO the `rpcserver_test.go` file should only be for testing the RPC interface as opposed to consensus rules.

Roasbeef:
I've updated the PR to only enforce BIP 68 behavior within the mempool _unconditionally_. The original "full" version of this PR can be found [here](https://github.com/Roasbeef/btcd/commits/sequence-locks-full).

I agree that the soft-fork related integration tests should be located within a distinct (possibly internal) package, as the soft-fork integration tests require a more precise set of initial conditions to properly test the behavior (pre-fork, activation, post-fork, etc).

blockchain/chain.go
```diff
@@ -694,6 +694,164 @@ func (b *BlockChain) CalcPastMedianTime() (time.Time, error) {
...
+	// If we're expressing the relative lock time in blocks, then the
+	// corresponding sequence number if simply the desired input age.
+	if !isSeconds {
+		return locktime
+	}
+
+	// Set the 22nd bit which indicates the lock time is in seconds, then
+	// shift the locktime over by 9 since the time granularity is in
+	// 512-second intervals (2^9). This results in a max lock-time of
+	// 33,554,431 seconds, or 1.06 years.
```
davecgh: This value looks incorrect to me (I know it's that way in the BIP too).

The mask is `0x0000ffff` which means the highest possible value it can hold is 65,535.   Since it's multiplied by 512 (left shift of 9), that is a max of 33,553,920.  Looking at it from the other direction, if you were to start at the stated 33,554,431 and divide it by 512 (right shift of 9 as is being done here), that is indeed the max value of 65,535, but since it's not evenly divisible by 9, you lose those extra 511 seconds due to the precision loss.  The result is you can't actually encode a max lock-time greater than 33,554,431 - 511 = 33,553,920 == 512 * 65,535.

blockchain/chain.go
```diff
@@ -694,6 +694,146 @@ func (b *BlockChain) CalcPastMedianTime() (time.Time, error) {
...
+// median time), or once the 'blockHeight' has been reached.
+type SequenceLock struct {
+	Seconds     int64
+	BlockHeight int32
+}
+
+// CalcSequenceLock computes a relative lock-time SequenceLock for the passed
+// transaction using the passed UtxoViewpoint to obtain the past median time
+// for blocks in which the referenced inputs of the transactions were included
+// within. The generated SequenceLock lock can be used in conjunction to a
```
Roasbeef: Fixed.

mempool/mempool.go
```diff
@@ -81,6 +81,10 @@ type Config struct {
 // Policy houses the policy (configuration parameters) which is used to
 // control the mempool.
 type Policy struct {
+	// TxVersion is the transaction version that the mempool should accept.
+	// All transactions above this version are rejected as non-standard.
+	TxVersion int32
```
Roasbeef: Great point, I've renamed the variable within the policy struct from `TxVersion` to `MaxTxVersion`.

blockchain/chain_test.go
```diff
@@ -110,3 +111,371 @@ func TestHaveBlock(t *testing.T) {
 		}
 	}
 }
+
+// TestCalcSequenceLock tests the LockTimeToSequence function, and the
+// CalcSequenceLock method of a Chain instance. The tests exercise several
+// combinations of inputs to the CalcSequenceLock function in order to ensure
+// the returned SequenceLocks are correct for each test instance.
+func TestCalcSequenceLock(t *testing.T) {
+	// Lock up the first 4 blocks of the main chain.
```
Roasbeef: A test chain is needed for the tests since `SequenceLocks` hits the DB to fetch the headers if the relative lock is seconds-based.

blockchain/validate.go
```diff
@@ -118,6 +118,38 @@ func IsCoinBase(tx *btcutil.Tx) bool {
...
+	// RelativeLockSecondsGranularity ...
+	RelativeLockSecondsGranularity = 9
+
+	// RelativeLockMask ...
+	RelativeLockMask = 0x0000ffff
+)
+
+// SequenceLockActive determines if a transaction's sequence locks have been
+// met, meaning that all the inputs of a given transaction have reached a
+// height sufficient for their relative lock-time maturity.
```
Roasbeef: Fixed. 11

blockchain/validate.go
```diff
@@ -118,6 +118,38 @@ func IsCoinBase(tx *btcutil.Tx) bool {
 	return IsCoinBaseTx(tx.MsgTx())
 }
 
+// TODO(rosabeef): remove these once PR created
```
Roasbeef: Rebased over 112 and fixed.

blockchain/chain.go
```diff
@@ -834,6 +834,24 @@ func (b *BlockChain) calcSequenceLock(tx *btcutil.Tx, utxoView *UtxoViewpoint,
 	return sequenceLock, nil
 }
 
+// LockTimeToSequence converts the passed relative locktime to a sequence
+// number in accordance to BIP-68.
+// See: https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki
+//  * (Compatibility)
+func LockTimeToSequence(isSeconds bool, locktime uint32) uint32 {
+	// If we're expressing the relative lock time in blocks, then the
+	// corresponding sequence number if simply the desired input age.
```
Roasbeef: Fixed.

blockchain/chain.go
Roasbeef: I believe you're correct! I've updated the comment to match the proper limits on time-based relative lock times. 

Will you be submitting a PR to update the BIP?

blockchain/chain.go
```diff
@@ -511,6 +511,52 @@ func (b *BlockChain) getPrevNodeFromNode(node *blockNode) (*blockNode, error) {
...
 }
 
+// ancestorNode returns the ancestor block 'distance' blocks before the passed
+// anchor block. While iterating backwards through the chain, any block nodes
+// which aren't in the memory chain are loaded in dynamically.
+//
+// This function MUST be called with the chain state lock held (for writes).
+func (b *BlockChain) ancestorNode(anchor *blockNode, distance uint32) (*blockNode, error) {
+
+	var err error
```
davecgh: Any reason for not just doing `err := ...` below?

blockchain/chain.go
```diff
@@ -511,6 +511,52 @@ func (b *BlockChain) getPrevNodeFromNode(node *blockNode) (*blockNode, error) {
 	return prevBlockNode, err
 }
 
+// ancestorNode returns the ancestor block 'distance' blocks before the passed
+// anchor block. While iterating backwards through the chain, any block nodes
+// which aren't in the memory chain are loaded in dynamically.
+//
+// This function MUST be called with the chain state lock held (for writes).
+func (b *BlockChain) ancestorNode(anchor *blockNode, distance uint32) (*blockNode, error) {
```
davecgh: I believe this will conflict with #755 which has the same function name, but different semantics.  This one goes back a certain distance while the one in #755 goes to a specific height.

We'll need to rename one or the other.  I'm open to suggestions for naming.

blockchain/chain.go
```diff
@@ -781,24 +828,32 @@ func (b *BlockChain) calcSequenceLock(tx *btcutil.Tx, utxoView *UtxoViewpoint,
...
+			if err != nil {
+				return sequenceLock, err
+			}
+
+			// The call above loaded all the headers between the
+			// inclusionary block and the chain tip. In order to
+			// properly calculate the proper relative lock time, we
+			// also need to load into memory enough blocks from the
+			// MTP calculation.
+			_, err = b.ancestorNode(b.bestNode, medianTimeBlocks)
```
davecgh: I don't believe this is needed, but if it were, I think it would need to be from the returned `blockNode`.  For example, say we're at block 450k and the input is from block 420k.  The call to `ancestorNode` above will get back to 420k, but you still need to end up going back to 420k-11 = 419989 in order to calculate the median time.

However, as I first mentioned, this shouldn't be needed.  The call to `b.calcPastMedianTime` will load the extra needed block nodes correctly so long as memory has at least the block after the most recent block the calculation starts from -- in keeping with the example, that would be 420k.

blockchain/chain.go
```diff
@@ -899,7 +899,7 @@ func LockTimeToSequence(isSeconds bool, locktime uint32) uint32 {
 	// Set the 22nd bit which indicates the lock time is in seconds, then
 	// shift the locktime over by 9 since the time granularity is in
 	// 512-second intervals (2^9). This results in a max lock-time of
-	// 33,554,431 seconds, or 1.06 years.
+	// 33,553,920 seconds, or 1.1 years.
```
davecgh: It's still actually 1.06 years, but it's rounded anyways, so probably doesn't matter.

mempool/policy.go
```diff
@@ -328,14 +328,15 @@ func isDust(txOut *wire.TxOut, minRelayTxFee btcutil.Amount) bool {
 // of recognized forms, and not containing "dust" outputs (those that are
 // so small it costs more to process them than they are worth).
 func checkTransactionStandard(tx *btcutil.Tx, height int32,
-	medianTimePast time.Time, minRelayTxFee btcutil.Amount) error {
+	medianTimePast time.Time, minRelayTxFee btcutil.Amount,
+	acceptedTxVersion int32) error {
```
dajohi: maxTxVersion might be a better name.

blockchain/chain.go
Roasbeef: I've renamed the function to `relativeNode`. Funnily, there's a bit of a double-entendre w.r.t the function name as it retrieves a node which is a relative of the ancestor node by walking back a relative distance into the chain 😎.

blockchain/chain.go
Roasbeef: Yep you're right, upon a closer look this additional call to the `ancestorNode` function isn't needed. Also calling the function with `b.bestNode` rather than `blockNode` is a typo/bug. 

I've removed the unnecessary (and incorrect) call to `ancestorNode`.

blockchain/chain.go
```diff
@@ -679,6 +678,174 @@ func (b *BlockChain) calcPastMedianTime(startNode *blockNode) (time.Time, error)
...
+		if utxo == nil {
+			return sequenceLock, fmt.Errorf("utxo %v not found",
+				txIn.PreviousOutPoint)
+		}
+
+		// If the input height is set to the mempool height, then we
+		// assume the transaction makes it into the next block when
+		// evaluating its sequence blocks.
+		inputHeight := utxo.BlockHeight()
+		if inputHeight == 0x7fffffff {
```
davecgh: Now that #813 is in, can you rebase and make use of `mining.UnminedHeight`?

Roasbeef: When attempting to implement this modification I ran into an import cycle between the `mining` package and the `blockchain` package. 

Dave and I discussed the issue on IRC, ultimately deciding to shelve the change for now to be revisited at a later time.

Roasbeef:
Squashed down to the original set of commits 👍

mempool/mempool.go
```diff
@@ -692,6 +694,19 @@ func (mp *TxPool) maybeAcceptTransaction(tx *btcutil.Tx, isNew, rateLimit, rejec
 		return missingParents, nil
 	}
 
+	// Don't allow the transaction into the mempool unless its sequence
+	// lock is active, meaning that it'll be allowed into the next block
+	// with respect to its defined relative lock times.
+	sequenceLock, err := mp.cfg.CalcSequenceLock(tx, utxoView)
+	if err != nil {
+		return nil, err
```
davecgh: Like other paths that involve return errors from `blockchain` this needs to be converted via `chainRuleError` if it's a `blockchain.RuleError`.  This is important so an appropriate reject message is generated and delivered back to the sending peer.

Here is the diff:

```diff
diff --git a/mempool/mempool.go b/mempool/mempool.go
index a9630a4..0a1ea9e 100644
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -705,6 +705,9 @@ func (mp *TxPool) maybeAcceptTransaction(tx *btcutil.Tx, isNew, rateLimit, rejec
        // with respect to its defined relative lock times.
        sequenceLock, err := mp.cfg.CalcSequenceLock(tx, utxoView)
        if err != nil {
+               if cerr, ok := err.(blockchain.RuleError); ok {
+                       return nil, chainRuleError(cerr)
+               }
                return nil, err
        }
```

Roasbeef: Fixed.

blockchain/chain.go
```diff
@@ -679,6 +678,154 @@ func (b *BlockChain) calcPastMedianTime(startNode *blockNode) (time.Time, error)
...
+		return sequenceLock, nil
+	}
+
+	// Grab the next height to use for inputs present in the mempool.
+	nextHeight := b.BestSnapshot().Height + 1
+
+	for _, txIn := range mTx.TxIn {
+		utxo := utxoView.LookupEntry(&txIn.PreviousOutPoint.Hash)
+		if utxo == nil {
+			return sequenceLock, fmt.Errorf("utxo %v not found",
```
Roasbeef: Fixed.

mempool/mempool.go
```diff
@@ -614,12 +624,15 @@ func (mp *TxPool) maybeAcceptTransaction(tx *btcutil.Tx, isNew, rateLimit, rejec
 	bestHeight := mp.cfg.BestHeight()
 	nextBlockHeight := bestHeight + 1
 
+	medianTimePast := mp.cfg.MedianTimePast()
```
Roasbeef: Fixed.

blockchain/chain.go
Roasbeef: Hmm it's possibly the most recent rebase got a bit hairy, just pushed a commit fixing it (again).

davecgh:
Latest commits look good.  If you would squash them into the commits, this is ready for merge.

PR #719, opened 2016-07-14
blockchain: Associate time src with chain instance.

davecgh:
Rather than making the caller pass in the median time source on `ProcessBlock` and `IsCurrent`, modify the `Config` struct to include the median time source and associate it with the chain instance when it is
created.

This is being done because both the `ProcessBlock` and `IsCurrent` functions require access to the `blockchain` state already, it is a little bit safer to ensure the time source matches the chain instance state, it simplifies the caller logic, and it also allows its use within the logic of the `blockchain` package itself which will be required by upcoming rule change warning logic that is part of BIP9.

### DIFF
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -164,6 +164,7 @@ type BlockChain struct {
 	checkpointsByHeight map[int32]*chaincfg.Checkpoint
 	db                  database.DB
 	chainParams         *chaincfg.Params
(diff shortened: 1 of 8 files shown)


dajohi:
OK

Roasbeef:
LGTM. 

Looking forward to the forthcoming BIP9 related PRs!

PR #720, opened 2016-07-20
blockchain: optimize HaveBlock

dajohi:
If a block is known to exist in the memory chain or database then
there is no need to check the orphan pool.


davecgh:
I'd prefer the travis bump be in a separate PR.  OK on the reversal of the checks.

dajohi:
updated

davecgh:
OK

PR #724, opened 2016-08-02
don't spin forever looking for peers

jadeblaquiere:
If you have a small number of nodes in a test network (e.g. two nodes where one is pending) then the number of tries may not get incremented resulting in a tight loop that spins essentially endlessly.

### DIFF
--- a/server.go
+++ b/server.go
@@ -1831,17 +1831,17 @@ out:
 				break
 			}
 
-			// Check that we don't have a pending connection to this addr.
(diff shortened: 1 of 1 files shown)


dajohi:
This looks correct.  OK.

davecgh:
Reads OK.  Need to test.

davecgh:
I've tested this and everything seems to work well.

OK

PR #726, opened 2016-08-04
peer: Remove mempool, getheaders, getblocks from stall detection.

### LINKED ISSUE #725: connect/disconnect cycling
I have a test network (based off a fork of btcd) where I'm seeing frequent connect/disconnect cycling. I am able to replicate that condition using the current btcd and testnet. Simply starting two isolated nodes (--nodnsseed) on different networks I see the pattern: 
- Connect new inbound peer
- Lose connection to outbound peer
- Request getblocks from inbound peer
- Receive getblocks from inbound peer
- Connect to new outbound peer
- Disconnect inbound peer (timeout from request)
- Request getblocks from outbound peer
- Receive getblocks from outbound peer

Digging in a bit (see detailed logs attached from both nodes) it seems the peer disconnects because it didn't receive a response to the getblocks request. The nodes are in sync and no miners are connected so there aren't any new blocks to send. And that appears to create a problem as the peer doesn't reply because it checks for n > 0 and silently swallows the response ( see : https://github.com/btcsuite/btcd/blob/master/server.go#L675 )

So... this seems broken to me, but there are two logical fixes... either the inv response should be optional (i.e. don't disconnect the peer if we got nothing back because it could be an empty list) or the peer should always send a response, even if it is an empty list. 

I like the second approach better as ultimately it does make sense to disconnect unresponsive peers and it is more consistent. I've tested by taking the conditional out and that seems to resolve the issue. 

Is that the right solution? does this affect bitcoind compatibility?

Logs attached here for reference:

[loga.txt](https://github.com/btcsuite/btcd/files/402231/loga.txt)
[logb.txt](https://github.com/btcsuite/btcd/files/402237/logb.txt)

dajohi:
They may not produce a response.

Fixes #725


davecgh:
NACK


This defeats the purpose of stall detection and just replaces one issue for another.  Specifically, if this PR were applied, then when a peer is selected as the sync peer and it never responds to a getheaders request, it would cause the chain sync to stall indefinitely.

PR #728, opened 2016-08-06
wire: Consolidate tests into the wire pkg.

davecgh:
Putting the test code in the same package makes it easier for forks since they don't have to change the import paths as much and it also gets rid of the need for internal_test.go to bridge.

This same thing should probably be done for the majority of the code base.


stevenroose:
Is there a reason not to do this other than wanting to do black-box testing? Does this add an overhead to usage of the package?

davecgh:
The main reason to put the tests in a separate package is, as you noted, black box testing.  It can help ensure the API is sane and is sufficiently usable.

However, it also has some downsides.  The main one is that it is cumbersome for forks since it means they have to change the import paths in all the test files.  Another one is that in addition to black box testing, you typically want to be able to do a bit of white box testing as well, so you end up having to write extra bridge code to be able to expose those internal functions to the test package in order to be able to do it.

Go only compiles files that are named `*_test.go` when running `go test`, so there is no additional overhead at all when not running the tests because the test code is simply never compiled.

jrick:
ok

cjepson:
ok; I realise the black box testing downside but I actually much prefer all the internal API to be tested as well. It's a shame there is no easy way to adjust global view of the functions in Golang when testing.

PR #729, opened 2016-08-07
chainhash: Abstract hash logic to new package.

davecgh:
**This PR requires #730 and btcsuite/btcutil#78.**

This is mostly a backport of some of the same modifications made in Decred along with a few additional things cleaned up.  In particular, this updates the code to make use of the new `chainhash` package.

Also, since this required API changes anyways and the hash algorithm is no longer tied specifically to SHA, all other functions throughout the code base which had `Sha` in their name have been changed to `Hash` so they are not incorrectly implying the hash algorithm.

- Remove the `wire.ShaHash` type
- Update all references to `wire.ShaHash` to the new `chainhash.Hash` type
- Rename the following functions and update all references:
  - `wire.BlockHeader.BlockSha` -> `BlockHash`
  - `wire.MsgBlock.BlockSha` -> `BlockHash`
  - `wire.MsgBlock.TxShas` -> `TxHashes`
  - `wire.MsgTx.TxSha` -> `TxHash`
  - `blockchain.ShaHashToBig` -> `HashToBig`
  - `peer.ShaFunc` -> `peer.HashFunc`
- Rename all variables that included `sha` in their name to include `hash` instead
- Update for function name changes in other dependent packages such as `btcutil`
- Update copyright dates on all modified files
- Update `glide.lock` file to use the required version of `btcutil`


stevenroose:
Is this just for for good practice? Or is the btcd codebase to be used for a different coin with a different hashfunction? Or are there plans to change the Bitcoin hash function? :-D

davecgh:
The btcd codebase is used for different coins with different hash functions however, so yes that is why it's being done.  There are no plans to change it for Bitcoin at the current time as far as I am aware, but I also would not be overly surprised if it were to change at some point in the far future.

jcvernaleo:
OK

marcopeereboom:
sweet mother of awesome! ok

cjepson:
OK

Need to backport tests to DCR eventually too

dajohi:
OK

PR #731, opened 2016-08-08
TravisCI: Set vendor experiment variable earlier.

davecgh:
This sets the `GO15VENDOREXPERIMENT` environment variable before glide is installed so vendoring is used when installing it as well.


jrick:
ok

dajohi:
ok

PR #732, opened 2016-08-09
chaincfg/blockchain: Parameterize more chain consts.

davecgh:
This moves several of the chain constants to the `Params` struct in the `chaincfg` package which is intended for that purpose.  This is mostly a backport of the same modifications made in `Decred` along with a few additional things cleaned up.

- Comment all fields in the `Params` struct definition
- Add locals to `BlockChain` instance for the calculated values based on the provided chain params
- Rename the following param fields:
  - `SubsidyHalvingInterval` -> `SubsidyReductionInterval`
  - `ResetMinDifficulty` -> `ReduceMinDifficulty`
- Add new Param fields:
  - `CoinbaseMaturity`
  - `TargetTimePerBlock`
  - `TargetTimespan`
  - `BlocksPerRetarget`
  - `RetargetAdjustmentFactor`
  - `MinDiffReductionTime`

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -2104,6 +2104,11 @@ func handleGetNetworkHashPS(s *rpcServer, cmd interface{}, closeChan <-chan stru
 		endHeight = best.Height
 	}
 
+	// Calculate the number of blocks per retarget interval based on the
+	// chain parameters.
+	blocksPerRetarget := int32(s.server.chainParams.TargetTimespan /
+		s.server.chainParams.TargetTimePerBlock)
+
 	// Calculate the starting block height based on the passed number of
 	// blocks.  When the passed value is negative, use the last block the
 	// difficulty changed as the starting height.  Also make sure the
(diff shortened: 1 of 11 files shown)


dajohi:
OK

rpcserver.go
```diff
@@ -2104,6 +2104,11 @@ func handleGetNetworkHashPS(s *rpcServer, cmd interface{}, closeChan <-chan stru
 		endHeight = best.Height
 	}
 
+	// Calculate the number of blocks per retarget interval based on the
+	// chain parameters.
+	blocksPerRetarget := int32(s.server.chainParams.TargetTimespan /
+		s.server.chainParams.TargetTimePerBlock)
+
```
dajohi: Perhaps blockchain could have an BlocksPerRetarget() int32 { }  that returns what it was configured with.. instead of recalcuating here?

davecgh: It's based on constants, so it's probably not a big deal that it's calculating it here.  However, I don't imagine it would hurt to expose it on the chain instance either.

cjepson:
OK. Goes a long way towards making these parameters more accessible by people who wish to alter them.

rpcserver.go
dajohi: That can be a separate PR though.  OK on this PR.

PR #733, opened 2016-08-10
btcd: Simplify shutdown signal handling logic.

davecgh:
This rewrites the shutdown logic to simplify the shutdown signalling.  All cleanup is now run from deferred functions in the main function and channels are used to signal shutdown either from OS signals or from
other subsystems such as the RPC server and windows service controller.

The RPC server has been modified to use a new channel for signalling shutdown that is exposed via the `RequestedProcessShutdown` function instead of directly calling `Stop` on the server as it previously did.

Finally, it adds a few checks for early termination during the main start sequence so the process can be stopped without starting all the subsystems if desired.

This is a backport of the equivalent logic from Decred with a few slight modifications.  Credits go to @jrick.

### DIFF
--- a/btcd.go
+++ b/btcd.go
@@ -19,8 +19,7 @@ import (
 )
 
 var (
-	cfg             *config
-	shutdownChannel = make(chan struct{})
+	cfg *config
 )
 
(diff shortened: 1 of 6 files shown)


dajohi:
OK

PR #734, opened 2016-08-12
txscript: Expose AddOps on ScriptBuilder.

### LINKED ISSUE #715: ScriptBuilder AddOperations (operations []bytes)
I have an algorithm that computes different parts of the scripts concurrently. Right now I have to collect all the prebuilt parts and stick them together in a binary presentation on my own.

Having AddOperations public interface in ScriptBuilder will allow me to do this in a more clean way. In fact, I noticed that there is such interface implemented already, but it is in the tests. Any special reason why this is not a public interface? Ofter all using or missing it is up to the consumers of the library.

davecgh:
This exposes a new function on the `ScriptBuilder` type named `AddOps` that allows multiple opcodes to be added via a single call and adds tests to exercise the new function.

Finally, it updates a couple of places in the signing code that were abusing the interface by setting its private script directly to use the new public function instead.

Implements and closes #715

### DIFF
--- a/txscript/scriptbuilder.go
+++ b/txscript/scriptbuilder.go
@@ -73,6 +73,27 @@ func (b *ScriptBuilder) AddOp(opcode byte) *ScriptBuilder {
 	return b
 }
 
+// AddOps pushes the passed opcodes to the end of the script.  The script will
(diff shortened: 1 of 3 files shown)


dajohi:
OK

Issue #139, opened 2014-06-11
Add estimatefee support


dajohi:
Add estimatefee RPC call support.

https://github.com/bitcoin/bitcoin/commit/171ca7745e77c9f78f26556457fe64e5b2004a75

jakesyl:
Addressed by #1180

PR #735, opened 2016-08-13
rpc method estimatefee

DanielKrawisz:
This pr implements estimatefee similar to the way it works in Bitcoin Core, as described [here](https://github.com/bitcoin/bitcoin/pull/3959). estimatefee allows for one parameter, NumBlocks. It attempts to provide an estimated fee / kb rate (in Bitcoins) for a transaction to be confirmed before at least NumBlocks (inclusive) have been mined. 

Every time a transaction is added to the mempool, it is registered with the fee estimator. Every time a new block is received, it is also registered with the fee estimator. The fee estimator keeps 25 bins containing information on transactions for which 1 to 25 blocks have been mined before being put in a block. Each bin contains information on up to 100 transactions. 

Let m be a bijective map from all sots in all bins to the set if indices from 1 to N, where N is the number of transactions in all bins. m maps the first slot in the first bin to 1, the second to 2, and so on, all the way through all the bins. The value returned by estimatefee is the fee / kb value of the transactions whose index in a sorted list of all transactions (with the highest value first) is the same as m[s], where s is the middle slot in bin NumBlocks. estimatefee therefore is a monotonically decreasing function of NumBlocks. 

It is also possible to rollback the effect of a limited number of blocks in case of orphans. Default number of rollbacks allowed is 2. 

Since it is very easy to store 25 values, all results for all values of NumBlocks are cached when estimatefee is first called. Each time a new block is registered or rolled back, the cache is deleted. 

~~In core, the set of observed transactions is saved between sessions in a file that also contains mempool transactions. btcd doesn't seem to have a similar file, so fee estimation starts over if the program restarts.~~

The server saves the state of the FeeEstimator on shutdown into the database and tries to restore it upon initialization. If it cannot, then it creates a new FeeEstimator. 

~~A corresponding pr to btcrpcclient has been made [here](https://github.com/btcsuite/btcrpcclient/pull/87).~~

### DIFF
--- a/mempool/estimatefee.go
+++ b/mempool/estimatefee.go
@@ -33,20 +33,27 @@ const (
 	// estimateFeeMaxReplacements is the max number of replacements that
 	// can be made by the txs found in a given block.
 	estimateFeeMaxReplacements = 10
+
+	bytePerKb = 1024
+
+	btcPerSatoshi = 1E-8
 )
 
 // SatoshiPerByte is number with units of satoshis per byte.
 type SatoshiPerByte float64
 
-// ToSatoshiPerKb returns a float value that represents the given
+// BtcPerKilobyte is number with units of bitcoins per kilobyte.
+type BtcPerKilobyte float64
+
+// ToBtcPerKb returns a float value that represents the given
 // SatoshiPerByte converted to satoshis per kb.
-func (rate SatoshiPerByte) ToSatoshiPerKb() float64 {
+func (rate SatoshiPerByte) ToBtcPerKb() BtcPerKilobyte {
 	// If our rate is the error value, return that.
 	if rate == SatoshiPerByte(-1.0) {
 		return -1.0
 	}
 
-	return float64(rate) * 1024
+	return BtcPerKilobyte(float64(rate) * bytePerKb * btcPerSatoshi)
 }
 
 // Fee returns the fee for a transaction of a given size for
@@ -88,7 +95,7 @@ type observedTransaction struct {
 // is used if Rollback is called to reverse the effect of registering
 // a block.
 type registeredBlock struct {
-	hash         *chainhash.Hash
+	hash         chainhash.Hash
 	transactions []*observedTransaction
 }
 
@@ -96,11 +103,11 @@ type registeredBlock struct {
 // fee estimations. It is safe for concurrent access.
 type FeeEstimator struct {
 	maxRollback uint32
-	binSize     int
+	binSize     int32
 
 	// The maximum number of replacements that can be made in a single
 	// bin per block. Default is estimateFeeMaxReplacements
-	maxReplacements int
+	maxReplacements int32
 
 	// The minimum number of blocks that can be registered with the fee
 	// estimator before it will provide answers.
@@ -109,17 +116,19 @@ type FeeEstimator struct {
 	// The last known height.
 	lastKnownHeight int32
 
-	sync.RWMutex
-	observed            map[chainhash.Hash]observedTransaction
-	bin                 [estimateFeeDepth][]*observedTransaction
-	numBlocksRegistered uint32 // The number of blocks that have been registered.
+	// The number of blocks that have been registered.
+	numBlocksRegistered uint32
+
+	mtx      sync.RWMutex
+	observed map[chainhash.Hash]*observedTransaction
+	bin      [estimateFeeDepth][]*observedTransaction
 
 	// The cached estimates.
 	cached []SatoshiPerByte
 
 	// Transactions that have been removed from the bins. This allows us to
 	// revert in case of an orphaned block.
-	dropped []registeredBlock
+	dropped []*registeredBlock
 }
 
 // NewFeeEstimator creates a FeeEstimator for which at most maxRollback blocks
@@ -132,21 +141,27 @@ func NewFeeEstimator(maxRollback, minRegisteredBlocks uint32) *FeeEstimator {
 		lastKnownHeight:     mining.UnminedHeight,
 		binSize:             estimateFeeBinSize,
 		maxReplacements:     estimateFeeMaxReplacements,
-		observed:            make(map[chainhash.Hash]observedTransaction),
-		dropped:             make([]registeredBlock, 0, maxRollback),
+		observed:            make(map[chainhash.Hash]*observedTransaction),
+		dropped:             make([]*registeredBlock, 0, maxRollback),
 	}
 }
 
 // ObserveTransaction is called when a new transaction is observed in the mempool.
 func (ef *FeeEstimator) ObserveTransaction(t *TxDesc) {
-	ef.Lock()
-	defer ef.Unlock()
+	ef.mtx.Lock()
+	defer ef.mtx.Unlock()
+
+	// If we haven't seen a block yet we don't know when this one arrived,
+	// so we ignore it.
+	if ef.lastKnownHeight == mining.UnminedHeight {
+		return
+	}
 
 	hash := *t.Tx.Hash()
 	if _, ok := ef.observed[hash]; !ok {
 		size := uint32(t.Tx.MsgTx().SerializeSize())
 
-		ef.observed[hash] = observedTransaction{
+		ef.observed[hash] = &observedTransaction{
 			hash:     hash,
 			feeRate:  NewSatoshiPerByte(btcutil.Amount(t.Fee), size),
 			observed: t.Height,
@@ -157,8 +172,8 @@ func (ef *FeeEstimator) ObserveTransaction(t *TxDesc) {
 
 // RegisterBlock informs the fee estimator of a new block to take into account.
 func (ef *FeeEstimator) RegisterBlock(block *btcutil.Block) error {
-	ef.Lock()
-	defer ef.Unlock()
+	ef.mtx.Lock()
+	defer ef.mtx.Unlock()
 
 	// The previous sorted list is invalid, so delete it.
 	ef.cached = nil
@@ -184,8 +199,8 @@ func (ef *FeeEstimator) RegisterBlock(block *btcutil.Block) error {
 	var replacementCounts [estimateFeeDepth]int
 
 	// Keep track of which txs were dropped in case of an orphan block.
-	dropped := registeredBlock{
-		hash:         block.Hash(),
+	dropped := &registeredBlock{
+		hash:         *block.Hash(),
 		transactions: make([]*observedTransaction, 0, 100),
 	}
 
@@ -200,41 +215,50 @@ func (ef *FeeEstimator) RegisterBlock(block *btcutil.Block) error {
 		}
 
 		// Put the observed tx in the oppropriate bin.
-		o.mined = height
-
 		blocksToConfirm := height - o.observed - 1
 
+		// This shouldn't happen if the fee estimator works correctly,
+		// but return an error if it does.
+		if o.mined != mining.UnminedHeight {
+			log.Error("Estimate fee: transaction ", hash.String(), " has already been mined")
+			return errors.New("Transaction has already been mined")
+		}
+
 		// This shouldn't happen but check just in case to avoid
-		// a panic later.
+		// an out-of-bounds array index later.
 		if blocksToConfirm >= estimateFeeDepth {
 			continue
 		}
 
 		// Make sure we do not replace too many transactions per min.
-		if replacementCounts[blocksToConfirm] == ef.maxReplacements {
+		if replacementCounts[blocksToConfirm] == int(ef.maxReplacements) {
 			continue
 		}
+
+		o.mined = height
 
 		replacementCounts[blocksToConfirm]++
 
 		bin := ef.bin[blocksToConfirm]
 
 		// Remove a random element and replace it with this new tx.
-		if len(bin) == ef.binSize {
-			l := ef.binSize - replacementCounts[blocksToConfirm]
+		if len(bin) == int(ef.binSize) {
+			// Don't drop transactions we have just added from this same block.
+			l := int(ef.binSize) - replacementCounts[blocksToConfirm]
 			drop := rand.Intn(l)
 			dropped.transactions = append(dropped.transactions, bin[drop])
 
 			bin[drop] = bin[l-1]
-			bin[l-1] = &o
+			bin[l-1] = o
 		} else {
-			ef.bin[blocksToConfirm] = append(bin, &o)
+			bin = append(bin, o)
 		}
+		ef.bin[blocksToConfirm] = bin
 	}
 
 	// Go through the mempool for txs that have been in too long.
 	for hash, o := range ef.observed {
-		if height-o.observed >= estimateFeeDepth {
+		if o.mined == mining.UnminedHeight && height-o.observed >= estimateFeeDepth {
 			delete(ef.observed, hash)
 		}
 	}
@@ -253,6 +277,14 @@ func (ef *FeeEstimator) RegisterBlock(block *btcutil.Block) error {
 	return nil
 }
 
+// LastKnownHeight returns the height of the last block which was registered.
+func (ef *FeeEstimator) LastKnownHeight() int32 {
+	ef.mtx.Lock()
+	defer ef.mtx.Unlock()
+
+	return ef.lastKnownHeight
+}
+
 // Rollback unregisters a recently registered block from the FeeEstimator.
 // This can be used to reverse the effect of an orphaned block on the fee
 // estimator. The maximum number of rollbacks allowed is given by
@@ -262,29 +294,24 @@ func (ef *FeeEstimator) RegisterBlock(block *btcutil.Block) error {
 // deleted if they have been observed too long ago. That means the result
 // of Rollback won't always be exactly the same as if the last block had not
 // happened, but it should be close enough.
-func (ef *FeeEstimator) Rollback(block *btcutil.Block) error {
-	ef.Lock()
-	defer ef.Unlock()
-
-	hash := block.Hash()
+func (ef *FeeEstimator) Rollback(hash *chainhash.Hash) error {
+	ef.mtx.Lock()
+	defer ef.mtx.Unlock()
 
 	// Find this block in the stack of recent registered blocks.
 	var n int
-	for n = 1; n < len(ef.dropped); n++ {
+	for n = 1; n <= len(ef.dropped); n++ {
 		if ef.dropped[len(ef.dropped)-n].hash.IsEqual(hash) {
 			break
 		}
 	}
 
-	if n == len(ef.dropped) {
+	if n > len(ef.dropped) {
 		return errors.New("no such block was recently registered")
 	}
 
 	for i := 0; i < n; i++ {
-		err := ef.rollback()
-		if err != nil {
-			return err
-		}
+		ef.rollback()
 	}
 
 	return nil
@@ -292,27 +319,23 @@ func (ef *FeeEstimator) Rollback(block *btcutil.Block) error {
 
 // rollback rolls back the effect of the last block in the stack
 // of registered blocks.
-func (ef *FeeEstimator) rollback() error {
-
+func (ef *FeeEstimator) rollback() {
 	// The previous sorted list is invalid, so delete it.
 	ef.cached = nil
 
 	// pop the last list of dropped txs from the stack.
 	last := len(ef.dropped) - 1
 	if last == -1 {
-		// Return if we cannot rollback.
-		return errors.New("max rollbacks reached")
+		// Cannot really happen because the exported calling function
+		// only rolls back a block already known to be in the list
+		// of dropped transactions.
+		return
 	}
 
-	ef.numBlocksRegistered--
-
 	dropped := ef.dropped[last]
-	ef.dropped = ef.dropped[0:last]
 
 	// where we are in each bin as we replace txs?
 	var replacementCounters [estimateFeeDepth]int
-
-	var err error
 
 	// Go through the txs in the dropped block.
 	for _, o := range dropped.transactions {
@@ -326,9 +349,8 @@ func (ef *FeeEstimator) rollback() error {
 		// Continue to go through that bin where we left off.
 		for {
 			if counter >= len(bin) {
-				// Create an error but keep going in case we can roll back
-				// more transactions successfully.
-				err = errors.New("illegal state: cannot rollback dropped transaction")
+				// Panic, as we have entered an unrecoverable invalid state.
+				panic(errors.New("illegal state: cannot rollback dropped transaction"))
 			}
 
 			prev := bin[counter]
@@ -344,6 +366,8 @@ func (ef *FeeEstimator) rollback() error {
 
 			counter++
 		}
+
+		replacementCounters[blocksToConfirm] = counter
 	}
 
 	// Continue going through bins to find other txs to remove
@@ -373,9 +397,11 @@ func (ef *FeeEstimator) rollback() error {
 		}
 	}
 
-	ef.lastKnownHeight--
+	ef.dropped = ef.dropped[0:last]
 
-	return err
+	// The number of blocks the fee estimator has seen is decrimented.
+	ef.numBlocksRegistered--
+	ef.lastKnownHeight--
 }
 
 // estimateFeeSet is a set of txs that can that is sorted
@@ -407,19 +433,26 @@ func (b *estimateFeeSet) estimateFee(confirmations int) SatoshiPerByte {
 		return 0
 	}
 
-	var min, max uint32 = 0, 0
-	for i := 0; i < confirmations-1; i++ {
-		min += b.bin[i]
-	}
-
-	max = min + b.bin[confirmations-1]
-
 	// We don't have any transactions!
-	if min == 0 && max == 0 {
+	if len(b.feeRate) == 0 {
 		return 0
 	}
 
-	return b.feeRate[(min+max-1)/2] * 1E-8
+	var min, max int = 0, 0
+	for i := 0; i < confirmations-1; i++ {
+		min += int(b.bin[i])
+	}
+
+	max = min + int(b.bin[confirmations-1]) - 1
+	if max < min {
+		max = min
+	}
+	feeIndex := (min + max) / 2
+	if feeIndex >= len(b.feeRate) {
+		feeIndex = len(b.feeRate) - 1
+	}
+
+	return b.feeRate[feeIndex]
 }
 
 // newEstimateFeeSet creates a temporary data structure that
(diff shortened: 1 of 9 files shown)


estimatefee.go
```diff
@@ -0,0 +1,434 @@
...
+	"math/rand"
+	"sort"
+	"sync"
+
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcutil"
+)
+
+const (
+	// estimateFeeDepth is the maximum number of blocks before a transaction
```
Roasbeef: The variable name as declared and what's referenced in the comment are mismatched. The comment should be updated to reference `estimateFeeBins` instead of `estimateFeeDepth`.

estimatefee.go
```diff
@@ -0,0 +1,434 @@
...
+
+	// estimateFeeMaxReplacements is the max number of replacements that
+	// can be made by the txs found in a given block.
+	estimateFeeMaxReplacements = 10
+)
+
+// observedTransaction represents an observed transaction and some
+// additional data required for the fee estimation algorithm.
+type observedTransaction struct {
+	hash     chainhash.Hash // a transaction hash.
```
Roasbeef: In order to maintain consistency with the rest of the codebase (for the most part), the comments for the struct's attributes should be placed a line `above` the attribute being described rather than `in-line`.

estimatefee.go
```diff
@@ -0,0 +1,434 @@
...
+	size     uint32         // size in bytes.
+	fee      uint64         // miner's fee.
```
Roasbeef: A minor recommendation: you might want to use `btcutil.Amount` in place of the `uint64` here. The former has some convenience methods for string formatting, multiplication with rounding, etc. 

Additionally, using a `btcutil.Amount` everywhere will save you from performing floating point operations for the most part throughout this PR.

estimatefee.go
```diff
@@ -0,0 +1,434 @@
...
+	feePerKb float64
+	observed int32 // The block height when it was observed.
+	mined    int32 // The block in which this tx was mined.
+}
+
+// registeredBlock
+type registeredBlock struct {
```
Roasbeef: Looks like a `gococ` style comment was started here but never completed.

estimatefee.go
```diff
@@ -0,0 +1,434 @@
...
+	// The last known height.
+	lastKnownHeight int32
+
+	sync.RWMutex
+	observed            map[chainhash.Hash]observedTransaction
+	bin                 [estimateFeeBins][]*observedTransaction
+	numBlocksRegistered uint32 // The number of blocks that have been registered.
+
+	// The cached estimates.
+	cached []float64
```
Roasbeef: Perhaps `cached` should be declared as a fixed-size array of length `estimateFeeBins`?

estimatefee.go
```diff
@@ -0,0 +1,434 @@
...
+func (ef *feeEstimator) RecordBlock(block *btcutil.Block) {
+	ef.Lock()
+	defer ef.Unlock()
+
+	// The previous sorted list is invalid, so delete it.
+	ef.cached = nil
+
+	height := block.Height()
+	if height != ef.lastKnownHeight+1 && ef.lastKnownHeight != mempoolHeight {
+		panic(fmt.Sprint("intermediate block not recorded; current height is ", ef.lastKnownHeight,
```
Roasbeef: This *definitely* should not be a panic. Instead the error created should be returned to the caller, deferring the responsibility of handling the error to them.

estimatefee.go
```diff
@@ -0,0 +1,434 @@
...
+			"; new height is ", height))
+	}
+
+	// Update the last known height.
+	ef.lastKnownHeight = height
+	ef.numBlocksRegistered++
+
+	// Randomly order txs in block.
```
Roasbeef: Why should the transaction here be ordered a pseudo-random manner? Is a randomized iteration order critical to the generating correct fee estimates?

estimatefee.go
```diff
@@ -0,0 +1,434 @@
...
+	for _, t := range block.Transactions() {
+		transactions[t] = struct{}{}
+	}
+
+	// Count the number of replacements we make per bin so that we don't
+	// replace too many.
+	var replacementCounts [estimateFeeBins]int
+
+	// Keep track of which txs were dropped in case of an orphan block.
+	dropped := registeredBlock{
```
Roasbeef: It may just be a front-end inconsistency on my end, but, it looks like some of this file hasn't yet been run through `gofmt`.

estimatefee.go
```diff
@@ -0,0 +1,434 @@
...
+			continue
+		}
+
+		// Put the observed tx in the oppropriate bin.
+		o.mined = height
+
+		blocksToConfirm := height - o.observed - 1
+
+		// Make sure we do not replace too many transactions per min.
+		if replacementCounts[blocksToConfirm] == ef.maxReplacements {
```
Roasbeef: This array index as well as the one below this within the loop will cause a **panic** if a transaction took more than 25 blocks to confirm. A possible solution would be to increase the declared size of the array. However, I recommend using a map instead in order to be on the safe side.

estimatefee.go
```diff
@@ -0,0 +1,434 @@
...
+		return errors.New("Max rollbacks reached.")
+	}
+
+	ef.numBlocksRegistered--
+
+	dropped := ef.dropped[last]
+	ef.dropped = ef.dropped[0:last]
+
+	// where we are in each bin as we replace txs?
+	var replacementCounters [estimateFeeBins]int
```
Roasbeef: My comment above about the use of a fixed-size array here possibly triggering a `panic` applies here also.

estimatefee.go
```diff
@@ -0,0 +1,434 @@
...
+
+	// Go through the txs in the dropped box.
```
Roasbeef: s/dropped box/dropped **block**/

estimatefee.go
```diff
@@ -0,0 +1,434 @@
...
+		blocksToConfirm := o.mined - o.observed - 1
+
+		bin := ef.bin[blocksToConfirm]
+
+		var counter = replacementCounters[blocksToConfirm]
+
+		// Continue to go through that bin where we left off.
+		for {
+			if counter >= len(bin) {
+				panic("Illegal state: cannot rollback dropped transaction!")
```
Roasbeef: This panic should be removed in place of simply returning an error.

estimatefee.go
```diff
@@ -0,0 +1,434 @@
...
+			if j >= l {
+				break
+			}
+
+			prev := ef.bin[i][j]
+
+			if prev.mined == ef.lastKnownHeight {
+				prev.mined = mempoolHeight
+
+				ef.bin[i] = append(ef.bin[i][0:j], ef.bin[i][j+1:l]...)
```
Roasbeef: In order to avoid an inadvertent memory leak, `ef.bin[i][j]` should be set to `nil` below this line.

estimatefee.go
```diff
@@ -0,0 +1,434 @@
...
+func (b *estimateFeeSet) Less(i, j int) bool {
+	return b.feeRate[i] > b.feeRate[j]
+}
+
+func (b *estimateFeeSet) Swap(i, j int) {
+	b.feeRate[i], b.feeRate[j] = b.feeRate[j], b.feeRate[i]
+}
+
+// EstimateFee returns the estimated fee for a transaction
+// to confirm in confirmations blocks from now, given from
```
Roasbeef: The `from` at the end of this line should be deleted, it's unnecessary.

estimatefee.go
```diff
@@ -0,0 +1,434 @@
...
+// the data set we have collected.
+func (b *estimateFeeSet) EstimateFee(confirmations int) float64 {
+	if confirmations <= 0 {
+		return math.Inf(1)
```
Roasbeef: Rather than reporting an estimated fee of `infinity`, I think an error should be returned here instead. 

An alternative is to modify the `confirmations` parameter to be a `uint` rather than an `int`.

mempool.go
```diff
@@ -469,7 +480,7 @@ func (mp *txMemPool) FetchTransaction(txHash *chainhash.Hash) (*btcutil.Tx, erro
 	return nil, fmt.Errorf("transaction is not in the pool")
 }
 
-// maybeAcceptTransaction is the internal function which implements the public
+// AcceptTransaction is the internal function which implements the public
```
Roasbeef: This comment change should be reverted.

Roasbeef:
Thanks for the PR! 

Great work, we've been missing a fee estimation model within `btcd` for too long now. I've been meaning to revisit a project of mine which predicts transaction fees, but never got around to doing it. The method is a bit more complex (it uses Gradient Boosted Regression Trees), and hasn't yet been compared properly to the existing fee estimation methods so I've held off on proposing integration of the project into `btcd`. 

I've done an initial pass through the PR, but haven't closely examined the tests yet nor verified the correctness of the model. 

One recommendation: I think fees should be stored and predicated in `sat/byte` rather than `btc/kilobyte`. The former unit should yield more accurate fee estimations as its a more precise unit. Additionally, it avoids the tendency of a `fee/KB` model to cause users to `over pay` due to the rounding involved in the calculations. 


> btcd doesn't seem to have a similar file, so fee estimation starts over if the program restarts.

No such file currently exists, however it should be rather straight forward to persist the state of the fee estimation model to the [database](https://godoc.org/github.com/btcsuite/btcd/database). A [bucket](https://godoc.org/github.com/btcsuite/btcd/database#Bucket) within the `metadata` storage can be dedicated to storing the fee estimation data so it persists between restarts. 


Also if we decide to move forward with this method, then it'll need a corresponding PR within [`btcrpcclient`](https://github.com/btcsuite/btcrpcclient) so users can programmatically gain access to the new RPC call.

Roasbeef:
Does this model incorporate [Alex Morcos' modifications to Gavin's initial model](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2014-October/006824.html)? At first glance, it doesn't appear that it does.

blockmanager.go
```diff
@@ -1386,6 +1396,7 @@ func (b *blockManager) Pause() chan<- struct{} {
 // newBlockManager returns a new bitcoin block manager.
 // Use Start to begin processing asynchronous block and inv updates.
 func newBlockManager(s *server, indexManager blockchain.IndexManager) (*blockManager, error) {
+
```
davecgh: Looks like an extra blank line snuck in here.

estimatefee.go
```diff
@@ -0,0 +1,434 @@
+// Copyright (c) 2013-2016 The btcsuite developers
```
davecgh: This is a new file.  2016.

estimatefee.go
davecgh: I agree.  The use of `btcutil.Amount` will be required before the PR can be merged.

estimatefee.go
```diff
@@ -0,0 +1,434 @@
...
+	mined    int32 // The block in which this tx was mined.
+}
+
+// registeredBlock
+type registeredBlock struct {
+	hash         *chainhash.Hash
+	transactions []*observedTransaction
+}
+
+// feeEstimator is a class that manages the data necessary to create
```
davecgh: nit: Go doesn't have classes.

```Go
feeEstimator manages the data....
```

davecgh:
This will require a rebase now that #737 went in.  I apologize for the churn, but it was important to bite the bullet and get the `mempool` separated.

estimatefee.go
```diff
@@ -0,0 +1,434 @@
...
+		if replacementCounts[blocksToConfirm] == ef.maxReplacements {
+			continue
+		}
+
+		replacementCounts[blocksToConfirm]++
+
+		bin := ef.bin[blocksToConfirm]
+
+		// Remove a random element and replace it with this new tx.
+		if len(bin) == int(ef.binSize) {
```
davecgh: ` ef.binSize` is already an int.  No need to cast it.

estimatefee.go
```diff
@@ -0,0 +1,434 @@
...
+			l := int(ef.binSize - replacementCounts[blocksToConfirm])
```
davecgh: Same here.

estimatefee.go
```diff
@@ -0,0 +1,434 @@
...
+			size:     size,
+			fee:      fee,
+			feePerKb: float64(1000*fee) / float64(size),
+			observed: t.Height,
+			mined:    mempoolHeight,
+		}
+	}
+}
+
+// RegisterBlock informs the fee estimator of a new block to take into account.
```
davecgh: Comment doesn't match function name.

mempool.go
```diff
@@ -396,6 +400,13 @@ func (mp *txMemPool) addTransaction(utxoView *blockchain.UtxoViewpoint, tx *btcu
 		},
 		StartingPriority: calcPriority(tx.MsgTx(), utxoView, height),
 	}
+
+	// If the fee estimator is not null, record this tx for fee estimation.
+	if mp.cfg.FeeEstimator != nil {
```
davecgh: Please put this code after `atomic.StoreInt64(&mp.lastUpdated, time.Now().Unix())`.  Putting it here makes the above comment no longer apply to the block and also it's an optional thing just like the `mp.cfg.AddrIndex != nil` below so it fits better there.

Also, for the comment, please just say `Record this tx for fee estimation if enabled.` so it matches the style of the `mp.cfg.AddrIndex != nil` that will be right next to it after it's moved.

rpcserver.go
```diff
@@ -172,6 +172,7 @@ var rpcHandlersBeforeInit = map[string]commandHandler{
 	"validateaddress":       handleValidateAddress,
 	"verifychain":           handleVerifyChain,
 	"verifymessage":         handleVerifyMessage,
+	"estimatefee":           handleEstimateFee,
```
davecgh: Please alphabetize this to be consistent with the rest of the handlers.  After `decdeScript`.

rpcserver.go
```diff
@@ -3683,6 +3683,17 @@ func handleVerifyMessage(s *rpcServer, cmd interface{}, closeChan <-chan struct{
 	return address.EncodeAddress() == c.Address, nil
 }
 
+// handleEstimateFee handles estimatefee commands.
```
davecgh: And move this up to the after the `handleDecodeScript` function so it is alphabetized as the rest of them are.

rpcserverhelp.go
```diff
@@ -526,6 +526,15 @@ var helpDescsEnUS = map[string]string{
 	"verifymessage-message":   "The signed message",
 	"verifymessage--result0":  "Whether or not the signature verified",
 
+	// EstimateFeeCmd help.
```
davecgh: Alphabetize too.

rpcserverhelp.go
```diff
@@ -621,6 +630,7 @@ var rpcResultTypes = map[string][]interface{}{
 	"validateaddress":       {(*btcjson.ValidateAddressChainResult)(nil)},
 	"verifychain":           {(*bool)(nil)},
 	"verifymessage":         {(*bool)(nil)},
+	"estimatefee":           {(*float64)(nil)},
```
davecgh: Alphabetize.

estimatefee.go
DanielKrawisz: It's because only we only make a maximum of estimateFeeMaxReplacements (default 10) per bin per registered block. If the transactions were not randomly ordered, we would bias the transactions toward those which are in the block first. There is no reason to assume that the first transactions in a block can be treated like a random sample.

estimatefee.go
DanielKrawisz: That shouldn't happen because all transactions that have been observed more than 25 blocks ago are removed after a block is registered. Once a new block comes in, there shouldn't be any which observedTransactions which have taken > 25 blocks to confirm. I'll add a check in there anyway.

mempool.go
DanielKrawisz: Oops. I don't know how THAT happened!

estimatefee.go
DanielKrawisz: I've changed the function so that it is unexported. It is only used internally. Errors for invalid queries are handled at a higher level.

DanielKrawisz:
Thanks guys! I have addressed the comments here and rebased this pr with the new master. FeeEstimator is now an exported type that resides in package mempool. However, I have yet to address the bigger suggestions. In particular, 

* use sat/byte rather than btc/kilobyte. 
* store and restore the FeeEstimator state in the database. 
* I was unaware of Alex Morcos' modifications to Gavin's initial model, so I'll incorporate that next chance I get.

mempool/estimatefee.go
```diff
@@ -0,0 +1,699 @@
...
+	// estimator before it will provide answers.
+	minRegisteredBlocks uint32
+
+	// The last known height.
+	lastKnownHeight int32
+
+	// The number of blocks that have been registered.
+	numBlocksRegistered uint32
+
+	sync.RWMutex
```
dajohi: unexport the mutex.

server.go
```diff
@@ -2478,6 +2489,31 @@ func newServer(listenAddrs []string, db database.DB, chainParams *chaincfg.Param
...
+		// same thing again somehow.
+		metadata.Delete(key)
+
+		// If there is an error just make a new fee estimator.
+		s.feeEstimator, _ = mempool.RestoreFeeEstimator(feeEstimationData)
+	}
+	tx.Commit()
+
+	if s.feeEstimator == nil {
+		s.feeEstimator = mempool.NewFeeEstimator(2, 5)
```
dajohi: magic numbers.  please add consts with a comment.

server.go
```diff
@@ -2478,6 +2489,31 @@ func newServer(listenAddrs []string, db database.DB, chainParams *chaincfg.Param
...
+	metadata := tx.Metadata()
+	feeEstimationData := metadata.Get(key)
+	if feeEstimationData != nil {
+
+		// delete it from the database so that we don't try to restore the
+		// same thing again somehow.
+		metadata.Delete(key)
+
+		// If there is an error just make a new fee estimator.
+		s.feeEstimator, _ = mempool.RestoreFeeEstimator(feeEstimationData)
```
dajohi: Please log the error.

mempool/estimatefee.go
```diff
@@ -0,0 +1,699 @@
...
+	// The block height when it was observed.
+	observed int32
+
+	// The height of the block in which it was mined.
+	// If the transaction has not yet been mined, it is zero.
+	mined int32
+}
+
+func (o *observedTransaction) Serialize(w io.Writer) {
+
```
dajohi: empty line

mempool/estimatefee.go
```diff
@@ -0,0 +1,699 @@
...
+	binary.Write(w, binary.BigEndian, o.hash)
+	binary.Write(w, binary.BigEndian, o.feeRate)
+	binary.Write(w, binary.BigEndian, o.observed)
+	binary.Write(w, binary.BigEndian, o.mined)
+}
+
+func deserializeObservedTransaction(r io.Reader) (*observedTransaction, error) {
+
```
dajohi: empty line

dajohi:
Excellent work.  I;ve been wanting this for a while.

Some more tests are needed to check the actual API that will be used.   Here is what i found with coverprofile:
```
github.com/btcsuite/btcd/mempool/estimatefee.go:43:     ToSatoshiPerKb                  0.0%
github.com/btcsuite/btcd/mempool/estimatefee.go:54:     Fee                             0.0%
github.com/btcsuite/btcd/mempool/estimatefee.go:65:     NewSatoshiPerByte               100.0%
github.com/btcsuite/btcd/mempool/estimatefee.go:86:     Serialize                       100.0%
github.com/btcsuite/btcd/mempool/estimatefee.go:94:     deserializeObservedTransaction  100.0%
github.com/btcsuite/btcd/mempool/estimatefee.go:120:    serialize                       0.0%
github.com/btcsuite/btcd/mempool/estimatefee.go:165:    NewFeeEstimator                 0.0%
github.com/btcsuite/btcd/mempool/estimatefee.go:178:    ObserveTransaction              87.5%
github.com/btcsuite/btcd/mempool/estimatefee.go:202:    RegisterBlock                   91.1%
github.com/btcsuite/btcd/mempool/estimatefee.go:315:    Rollback                        0.0%
github.com/btcsuite/btcd/mempool/estimatefee.go:342:    rollback                        97.2%
github.com/btcsuite/btcd/mempool/estimatefee.go:435:    Len                             100.0%
github.com/btcsuite/btcd/mempool/estimatefee.go:437:    Less                            100.0%
github.com/btcsuite/btcd/mempool/estimatefee.go:441:    Swap                            100.0%
github.com/btcsuite/btcd/mempool/estimatefee.go:448:    estimateFee                     87.5%
github.com/btcsuite/btcd/mempool/estimatefee.go:481:    newEstimateFeeSet               100.0%
github.com/btcsuite/btcd/mempool/estimatefee.go:508:    estimates                       100.0%
github.com/btcsuite/btcd/mempool/estimatefee.go:521:    EstimateFee                     81.8%
github.com/btcsuite/btcd/mempool/estimatefee.go:555:    deserializeRegisteredBlock      0.0%
github.com/btcsuite/btcd/mempool/estimatefee.go:580:    Save                            92.0%
github.com/btcsuite/btcd/mempool/estimatefee.go:629:    RestoreFeeEstimator             70.5%
github.com/btcsuite/btcd/mempool/estimatefee.go:315:    Rollback                        0.0%
```

DanielKrawisz:
Hey, I think my pr is ready for another review. Test coverage is over 90% and I'm pretty sure I've addressed all the comments here.

dajohi:
@DanielKrawisz can you rebase one more time and i will run/test it.

davecgh:
I haven't reviewed the updated code yet, but this is failing the CI builds.

```
++go fmt ./addrmgr/... ./blockchain/... ./btcec/... ./btcjson/... ./chaincfg/... ./cmd/... ./connmgr/... ./database/... ./integration/... ./limits/... ./mempool/... ./mining/... ./peer/... ./rpctest/... ./txscript/... ./wire/... .
mempool/mempool.go
server.go
```

Looks like you need to run `go fmt`.

DanielKrawisz:
I forgot to run goclean.sh after the last rebase!

Roasbeef:
Testing this PR on mainnet now. Will report back with findings/graphs/comparisons.

stevenroose:
Fixes #139

jlopp:
*Bump* this is a pretty important PR; it's sad to see it languish.

Roasbeef:
@DanielKrawisz can you squash those additional commits down into a core set? Thanks!

Roasbeef:
I've been running this on testnet for a few days now, and hit this earlier today: 
```
2017-12-01 03:34:47.900 [INF] CHAN: REORGANIZE: Block 00000000000033e8d1f0118d3504b2e0580f5ed603a76de8d5979106585c602a is causing a reorganize.                                                                                                 [1442/1788]
fatal error: sync: Unlock of unlocked RWMutex

goroutine 26 [running]:
runtime.throw(0xab12ad, 0x20)
        /usr/lib/go-1.8/src/runtime/panic.go:596 +0x95 fp=0xc43db28f48 sp=0xc43db28f28
sync.throw(0xab12ad, 0x20)
        /usr/lib/go-1.8/src/runtime/panic.go:585 +0x35 fp=0xc43db28f68 sp=0xc43db28f48
sync.(*RWMutex).Unlock(0xc4201fc1fc)
        /usr/lib/go-1.8/src/sync/rwmutex.go:118 +0xa7 fp=0xc43db28f98 sp=0xc43db28f68
runtime.call32(0x0, 0xbee280, 0xc43c07e1f0, 0x800000008)
        /usr/lib/go-1.8/src/runtime/asm_amd64.s:514 +0x48 fp=0xc43db28fc8 sp=0xc43db28f98
panic(0x9ecc40, 0xeadc70)
        /usr/lib/go-1.8/src/runtime/panic.go:489 +0x2cf fp=0xc43db29060 sp=0xc43db28fc8
runtime.panicindex()
        /usr/lib/go-1.8/src/runtime/panic.go:28 +0x5e fp=0xc43db29080 sp=0xc43db29060
github.com/roasbeef/btcd/mempool.(*FeeEstimator).RegisterBlock(0xc4204f2b00, 0xc43c3b3c70, 0x0, 0x0)
        /root/go/src/github.com/roasbeef/btcd/mempool/estimatefee.go:283 +0xc11 fp=0xc43db29460 sp=0xc43db29080
github.com/roasbeef/btcd/netsync.(*SyncManager).handleBlockchainNotification(0xc42013e370, 0xc43b9a5220)
        /root/go/src/github.com/roasbeef/btcd/netsync/manager.go:1257 +0x210 fp=0xc43db29530 sp=0xc43db29460
github.com/roasbeef/btcd/netsync.(*SyncManager).(github.com/roasbeef/btcd/netsync.handleBlockchainNotification)-fm(0xc43b9a5220)
        /root/go/src/github.com/roasbeef/btcd/netsync/manager.go:1456 +0x34 fp=0xc43db29550 sp=0xc43db29530
github.com/roasbeef/btcd/blockchain.(*BlockChain).sendNotification(0xc4201fc180, 0x1, 0xa4a2c0, 0xc43c3b3c70)
        /root/go/src/github.com/roasbeef/btcd/blockchain/notifications.go:78 +0xc0 fp=0xc43db29598 sp=0xc43db29550
github.com/roasbeef/btcd/blockchain.(*BlockChain).connectBlock(0xc4201fc180, 0xc42a85aaf0, 0xc43c3b3c70, 0xc43c562240, 0xc43062ca00, 0x23, 0x23, 0x0, 0x0)
        /root/go/src/github.com/roasbeef/btcd/blockchain/chain.go:679 +0x402 fp=0xc43db29640 sp=0xc43db29598
github.com/roasbeef/btcd/blockchain.(*BlockChain).reorganizeChain(0xc4201fc180, 0xc42ac66480, 0xc42ac66450, 0xc4335f42f0, 0x1)
        /root/go/src/github.com/roasbeef/btcd/blockchain/chain.go:997 +0x152f fp=0xc43db29818 sp=0xc43db29640
github.com/roasbeef/btcd/blockchain.(*BlockChain).connectBestChain(0xc4201fc180, 0xc42a85b110, 0xc43bff8460, 0xc400000000, 0x0, 0x0, 0x79e49b)
        /root/go/src/github.com/roasbeef/btcd/blockchain/chain.go:1113 +0x9ea fp=0xc43db29908 sp=0xc43db29818
github.com/roasbeef/btcd/blockchain.(*BlockChain).maybeAcceptBlock(0xc4201fc180, 0xc43bff8460, 0xc400000000, 0xc43c3102e8, 0x79d47b, 0xc4201fc1fc)
        /root/go/src/github.com/roasbeef/btcd/blockchain/accept.go:78 +0x220 fp=0xc43db299d0 sp=0xc43db29908
github.com/roasbeef/btcd/blockchain.(*BlockChain).processOrphans(0xc4201fc180, 0xc43c4c7240, 0xffffffff00000000, 0x0, 0x0)
```

Roasbeef:
Here's another instance with a full paste (still haven't pin pointed if it's related to this PR): https://pastebin.com/Qx0t6HZk

stevenroose:
@DanielKrawisz you still aware of this?

Roasbeef:
I've been running this for some time now on my fork of `btcd`. I haven't hit that panic I listed above in weeks. However, we've added the following commits in order to make the fee estimation more accurate: 
 * https://github.com/Roasbeef/btcd/commit/2fc9884a2ee75eccffbd7e4c28a7fb1cbf2800ea
 * https://github.com/Roasbeef/btcd/commit/fe8e9472df015c742f08759a2633da9e50d2b02b
 * https://github.com/Roasbeef/btcd/commit/0dcaa7f9b47c3b541fa4270ad57fb3d13b518061

stevenroose:
Hmm, what is the quality of the estimation (I didn't read what algorithm it uses)?

What happened to the ML algos for online estimation that you wrote about? (Read that draft paper this weekend.)

Roasbeef:
A slightly modified version of this was recently merged in as part of #1180. It should serve as a starting point to improve the fee estimation algo in the future.

PR #736, opened 2016-08-15
server: Remove superfluous check in OnMemPool.

davecgh:
This reduces the mempool lock contention by removing an unnecessary check when responding to a `mempool` request.

In particular, the code first gets a list of all transactions from the mempool and then iterates them in order to construct the inventory vectors and apply bloom filtering if it is enabled.  Since it is possible that the transaction was removed from the mempool by another thread while that list is being iterated, the code was checking if each transaction was still in the mempool.  This is a pointless check because the transaction might still be removed at any point after the check anyways.  For example, it might be removed after the mempool response has been sent to the remote peer or even while the loop is still iterating.


dajohi:
OK.  I've had this in my tree for a while.

PR #737, opened 2016-08-15
mempool: Refactor mempool code to its own package.

davecgh:
This does the minimum work necessary to refactor the mempool code into its own package.  The idea is that separating this code into its own package will greatly improve its testability, allow independent
benchmarking and profiling, and open up some interesting opportunities for future development related to the memory pool.

There are likely some areas related to policy that could be further refactored, however it is better to do that in future commits in order to keep the changeset as small as possible during this refactor.

Overview of the major changes:

- Create the new package
- Move several files into the new package:
  - `mempool.go` -> `mempool/mempool.go`
  - `mempoolerror.go` -> `mempool/error.go`
  - `policy.go` -> `mempool/policy.go`
  - `policy_test.go` -> `mempool/policy_test.go`
- Update mempool logging to use the new mempool package logger
- Rename `mempoolPolicy` to `Policy` (so it's now `mempool.Policy`)
- Rename `mempoolConfig` to `Config` (so it's now `mempool.Config`)
- Rename `mempoolTxDesc` to `TxDesc` (so it's now `mempool.TxDesc`)
- Rename `txMemPool` to `TxPool` (so it's now `mempool.TxPool`)
- Move `defaultBlockPrioritySize` to the new package and export it
- Export `DefaultMinRelayTxFee` from the `mempool` package
- Export the `CalcPriority` function from the `mempool` package
- Introduce a new `RawMempoolVerbose` function on the `TxPool` and update
  the RPC server to use it
- Update all references to the `mempool` to use the package
- Add a skeleton `README.md`

### DIFF
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -2,7 +2,7 @@
 // Use of this source code is governed by an ISC
 // license that can be found in the LICENSE file.
 
-package main
+package mempool
 
 import (
 	"container/list"
@@ -16,6 +16,8 @@ import (
 
 	"github.com/btcsuite/btcd/blockchain"
 	"github.com/btcsuite/btcd/blockchain/indexers"
+	"github.com/btcsuite/btcd/btcjson"
+	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/mining"
 	"github.com/btcsuite/btcd/txscript"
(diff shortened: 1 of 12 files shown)


dajohi:
ok

mempool/mempool.go
```diff
@@ -982,23 +997,69 @@ func (mp *txMemPool) MiningDescs() []*mining.TxDesc {
...
 }
 
-// newTxMemPool returns a new memory pool for validating and storing standalone
+// New returns a new memory pool for validating and storing standalone
 // transactions until they are mined into a block.
-func newTxMemPool(cfg *mempoolConfig) *txMemPool {
-	memPool := &txMemPool{
+func New(cfg *Config) *TxPool {
+	return &TxPool{
 		cfg:           *cfg,
```
Roasbeef: Minor nit: why does the `New()` function accept a `*Config`, but then immediately deference it? 

It seems that either `New()` should be changed to accept a `Config`, or the `TxPool`'s `cfg` attribute be changed to be a `*Config` rather than a `Config`. 

I know this behavior has been in place since the mempool was slightly refactored to use the previous `mempoolConfig`, but it's been bugging me since then... 🤓

davecgh: It's to avoid two copies although it's so small it probably really doesn't matter.

The intent is to make sure once a config is provided it can't be changed by the caller hence the deref to make a copy.  Accepting it as a parameter means that parameter is a copy and then setting the field in the type is another copy.

davecgh: I don't mind changing it if it really bugs folks.

Roasbeef: I see, I hadn't considered possible mutation of the config after it had been passed to the mempool. 

With that knowledge, it no longer bugs me 🙂

Roasbeef:
The changes read well. I like how the commit is rather minimal in its refactorings. 

This is a great step towards future mempool extensions such as shared distributed `TxPool` or a fully on-disk `TxPool`! Extracting the mempool into a separate package will much such changes easier in the future as pointed out in the PR description.

mempool/mempool.go
jrick: The config is not safe for concurrent access though, it should be treated as immutable.

PR #738, opened 2016-08-16
blockchain: Add median time to state snapshot.

davecgh:
This adds a new field to the best chain state snapshot for the calculated past median time as returned by the `calcPastMedianTime` function.  This is useful since it provides fast access to it without having to acquire the chain lock which is needed to recalculate it.

This will ultimately allow the associated exported function to be removed since it only exists to be able to calculate this exact value, however this commit only introduces the new field in order to keep the changes minimal.

### DIFF
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -128,23 +128,25 @@ func removeChildNode(children []*blockNode, node *blockNode) []*blockNode {
 // However, the returned snapshot must be treated as immutable since it is
 // shared by all callers.
 type BestState struct {
-	Hash      *chainhash.Hash // The hash of the block.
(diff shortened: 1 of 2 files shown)


Roasbeef:
Changes look good to me. This was something I was considering adding into by BIP 113 PR as I was writing it initially. Inclusion of this PR will simplify (reducing lines of code) #712 a bit. 

I also support later removing the exported function, as callers would either re-use an existing `blockchain.BestState` with the local scope or obtain a new instance. Either way lock usage is reduced as it no longer need to be re-calculated each time before usage. 

This will also speed up block validation once #712 is in.

PR #744, opened 2016-08-19
wire: Lower MaxUserAgentLen to 256.

### LINKED ISSUE #743: Why max user agent length in btcd is different from bitcoin?
The MaxUserAgentLen of btcd defined in msgversion.go is 2000 where MAX_SUBVERSION_LENGTH of bitcoin  defined in net.h is 256, why are they different? And if btcd node sends a version message which user agent length greater than 256, other bitcoin nodes will reject the message?

davecgh:
The protocol was silently upgrade in Core some time ago to enforce a limit of 256 for the user agent in version messages.  This updates wire to coincide with that change and consequently includes a wire version bump to 0.4.1.

Fixes #743.

### DIFF
--- a/wire/msgversion.go
+++ b/wire/msgversion.go
@@ -15,10 +15,10 @@ import (
 
 // MaxUserAgentLen is the maximum allowed length for the user agent field in a
 // version message (MsgVersion).
-const MaxUserAgentLen = 2000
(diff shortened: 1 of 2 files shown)


dajohi:
OK

PR #745, opened 2016-08-19
rpctest: Prevent bogus test fail in mempool join.

davecgh:
This modifies the new rpctest package test which ensures the `JoinNodes` function blocks until two nodes have synced mempools to first wait until a transaction is seen in the mempool of one of the nodes before
attempting to join.

This is necessary since slow systems, such as the CI build environment, might take longer for the transaction to be processed into the mempool than it does to connect and retrieve an empty mempool which leads a false positive test failure.


davecgh:
While the logic in this PR is correct and needs to happen, I found a few more bugs in the new `rpctest` package `JoinNodes` code as well as the test logic, so I'm going to incorporate this change into a separate PR which addresses the various issues.

PR #746, opened 2016-08-20
rpctest: Correct several issues in tests and joins.

davecgh:
This corrects several issues with the new `rpctest` package.

The following is an overview of the issues that have been corrected:

- The `JoinNodes` code was incorrect for both mempool and blocks.
  - For mempool it was only checking the first node against itself
  - For blocks it was only testing the height instead of hash and height which means the chains could be different and it would claim they're synced
- The test for mempool joins was inaccurate in a few ways:
  - It was generating separate chains such that the generated transaction was invalid (an orphan) on one chain, but not the other
  - Mempools are not automatically synced when nodes are connected and there is a large random delay before any transaction rebroadcast happens, so it can't be relied on for the purposes of this test
- The test for block joins was generating two independent chains of the same height with the same difficulty and was only passing due to the aforementioned bug in `JoinNodes`
- All of the `ConnectNode` calls were connecting the main harness outbound to the local test harness instances
  - This is not correct because `ConnectNode` makes the outbound connection persistent, which means once the local test harness is gone, it would keep trying to connect for the remainder of the tests to a node that is never coming back and instead ends up connecting to an independent test harness.

NOTE: This includes the changes from #745 as well.

### DIFF
--- a/rpctest/utils.go
+++ b/rpctest/utils.go
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcrpcclient"
 )
 
@@ -48,8 +49,8 @@ func JoinNodes(nodes []*Harness, joinType JoinType) error {
 func syncMempools(nodes []*Harness) error {
 	poolsMatch := false
 
+retry:
 	for !poolsMatch {
-	retry:
 		firstPool, err := nodes[0].Node.GetRawMempool()
 		if err != nil {
 			return err
@@ -58,7 +59,7 @@ func syncMempools(nodes []*Harness) error {
 		// If all nodes have an identical mempool with respect to the
 		// first node, then we're done. Otherwise, drop back to the top
 		// of the loop and retry after a short wait period.
-		for _, node := range nodes[:1] {
+		for _, node := range nodes[1:] {
 			nodePool, err := node.Node.GetRawMempool()
 			if err != nil {
 				return err
@@ -66,7 +67,7 @@ func syncMempools(nodes []*Harness) error {
 
 			if !reflect.DeepEqual(firstPool, nodePool) {
 				time.Sleep(time.Millisecond * 100)
-				goto retry
+				continue retry
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


rpctest/utils.go
```diff
@@ -76,24 +77,25 @@ func syncMempools(nodes []*Harness) error {
...
 		for _, node := range nodes {
-			blockHeight, err := node.Node.GetBlockCount()
+			blockHash, blockHeight, err := node.Node.GetBestBlock()
 			if err != nil {
 				return err
 			}
 
-			blockHeights[blockHeight] = struct{}{}
-			if len(blockHeights) > 1 {
+			existingHeight, ok := blocksByHash[*blockHash]
```
Roasbeef: Seems like the two-value read here is unnecessary. It's never used within the loop, and the `existingHeight` value can safely be used if the node's `blockHash` isn't found within the map since the value is simply an `int32` rather than a pointer.

davecgh: It's actually used below in the `if len(blocksByHash) > 1 || ok &&` statement.

However, I think you're right that it doesn't really need to be checked since the only time the actual `blockHeight` will be 0 is when it's the genesis block, and if you're on two different networks, the hashes won't be the same so it'll fail the `if len(blocksByHash) > 1` check.

davecgh: Updated to remove the `ok` parameter and check in the `if` statement.

davecgh: Actually, after thinking about it (and seeing the test fail upon changing it), it is required.  The reason is because you only want to fail the check when another node actually has the given block hash as at given height that is not the current node's `blockHeight`.

If it's removed, the test fails like so:
```
--- FAIL: TestHarness (61.71s)
	rpc_harness_test.go:324: blocks never detected as synced
```

Roasbeef: Ahh, yeah you're correct, the check is in-fact needed. I came to conclusion a little too quickly there without considering the full context.

davecgh: I think this code was a bit too clever (and hence hard to follow), so I changed it to something that I think should be much simpler.  It also has the benefit of being faster since it avoids the need for a map.

Roasbeef:
Excellent. Thanks for the fix ups! This should completely eliminate the flakiness of the `JoinNodes` mempool test, amongst the other bugs this PR squashes. 

I also like the elimination of that "clever" bit of code in `syncBlocks`. At a glance, it's now much easier to understand exactly what's going. 

OK.

PR #747, opened 2016-08-21
mempool: Break dependency on chain instance.

davecgh:
This modifies the config for the new `mempool` package such that it takes a callback function to obtain the best chain height instead of requiring a fully initialized `blockchain.BlockChain` instance.

This will make it much easier to test the `mempool` since the tests will be able to provide their own height function to test various functionality without having create and manipulate full blocks and chain instances.

### DIFF
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -55,9 +55,9 @@ type Config struct {
 	// transaction output information.
 	FetchUtxoView func(*btcutil.Tx) (*blockchain.UtxoViewpoint, error)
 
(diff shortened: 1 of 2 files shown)


jrick:
looks like an improvement. ok

alexlyp:
great move for further decoupling.  OK

jrick:
although I will say I wish it was more explicit that the function is a required dependency instead of it being part of a Config struct (where zeroed fields are usually perfectible acceptable).

dajohi:
It is an improvement.  OK

PR #748, opened 2016-08-22
rpctest: prevent leaking processes in integration tests due to panics

Roasbeef:
This PR adds a `defer` statement at the top of  `TestRpcServer`
which will attempt a `recover` which tears down all active harnesses in
the event that one of the tests causes a panic in the main goroutine.

Before this PR, if a buggy test caused a panic while all
integration tests were being executed, then any active harnesses would
fail to be properly torn down. This would cause the running btcd
processes to be leaked, possibly interfering with future test runs until
the process was manually killed. This PR fixes such behavior.

In order to aid in debugging, when a test panics, the test # is 
printed out along with a full stack-trace from the start of the test 
to the panic point.

### DIFF
--- a/rpcserver_test.go
+++ b/rpcserver_test.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"runtime/debug"
 	"testing"
 
 	"github.com/btcsuite/btcd/chaincfg"
@@ -109,17 +110,18 @@ func TestMain(m *testing.M) {
 	// Initialize the primary mining node with a chain of length 125,
 	// providing 25 mature coinbases to allow spending from for testing
 	// purposes.
-	if err = primaryHarness.SetUp(true, 25); err != nil {
+	if err := primaryHarness.SetUp(true, 25); err != nil {
 		fmt.Println("unable to setup test chain: ", err)
 		os.Exit(1)
 	}
 
 	exitCode := m.Run()
 
-	// Clean up the primary harness created above. This includes removing
-	// all temporary directories, and shutting down any created processes.
-	if err := primaryHarness.TearDown(); err != nil {
-		fmt.Println("unable to setup test chain: ", err)
+	// Clean up any active harnesses that are still currently running.This
+	// includes removing all temporary directories, and shutting down any
+	// created processes.
+	if err := rpctest.TearDownAll(); err != nil {
+		fmt.Println("unable to tear down all harnesses: ", err)
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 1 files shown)


rpcserver_test.go
```diff
@@ -109,25 +110,44 @@ func TestMain(m *testing.M) {
...
 }
 
 func TestRpcServer(t *testing.T) {
+	var currentTestNum int
+	defer func() {
+		// If one of the integration tests caused a panic within the main
+		// goroutine, then tear down all the harnesses in order to avoid
+		// any leaked btcd processes.
+		if r := recover(); r != nil {
+			fmt.Println("recovering from test panic: ", r)
```
davecgh: This should probably be `t.Log` along with the other prints in here.

rpcserver_test.go
```diff
@@ -91,9 +92,9 @@ func testGetBlockHash(r *rpctest.Harness, t *testing.T) {
 }
 
 var rpcTestCases = []rpctest.HarnessTestCase{
-	testGetBestBlock,
-	testGetBlockCount,
-	testGetBlockHash,
+	0: testGetBestBlock,
```
davecgh: Was this left over from your testing?  Unless I'm missing something, I don't see any benefit to this as it isn't valid to specify them sparsely.

rpcserver_test.go
```diff
@@ -109,25 +110,44 @@ func TestMain(m *testing.M) {
...
 
-	// Clean up the primary harness created above. This includes removing
-	// all temporary directories, and shutting down any created processes.
-	if err := primaryHarness.TearDown(); err != nil {
-		fmt.Println("unable to setup test chain: ", err)
-		os.Exit(1)
+	// Clean up any active harnesses that are still currently running.This
+	// includes removing all temporary directories, and shutting down any
+	// created processes.
+	if len(rpctest.ActiveHarnesses()) > 0 {
```
davecgh: Any reason to even check this?  I believe `TearDownAll` will simply be a noop if there are no more running.

rpcserver_test.go
Roasbeef: Using `t.Log` would only print the statements if the `-v` is enabled when running `go test`. This first statement prints the nature of the panic itself which can be very valuable when debugging the test.

rpcserver_test.go
Roasbeef: Yeah this was left over from exploring an idea. I've removed it.

rpcserver_test.go
Roasbeef: Correct. If there aren't any harnesses running then the call to `TearDownAll` will be a no-op. I've removed the conditional guarding it.

davecgh:
Should be `rpctest:` prefix and I don't know if you care, but there are a couple of typos in the commit message:

to be properly teared down -> to be properly **torn** down
possibly interring with -> possibly **interfering** with

rpcserver_test.go
davecgh: Yeah fair point.  I always have `-v` set when running tests and so does Travis, so I didn't really consider that.

Roasbeef:
Thanks, I've correct the typos in the commit message and switched to the `rpctest:` prefix.

Roasbeef:
Another small addition to `rpctest` I think would be valuable is possibly dumping the logs of some/all (or just the main harness) in the case that one of the tests fails.

While writing tests for the upcoming soft-forks I've found myself doing this manually in order to find the bug in either my test, or within `btcd.

davecgh:
I edit the PR description to match the updated commit message as well.

OK after squash.

Roasbeef:
Squashed 👍

davecgh:
As far as your comment regarding dumping the logs on failure, that sounds like a reasonable thing to do as I can see it being rather manual to hunt them down otherwise (off in some random temp directory).

PR #749, opened 2016-08-23
config: introduce new flags to accept/reject non-std transactions

Roasbeef:
This PR adds two new cli flags: one for accepting non-std transactions, and the other for rejecting non-std transactions.

The two flag are rejected when used concurrently. Config parsing is set up such that, the desired policy expressed via the config always overrides the policy set by default for a particular chain.

The `mempool`'s `Policy` config has been modified to add a new `RelayNonStd` bool which is checked when validating transactions rather than the policy as dicated by the current `ChainParams` instance. 

Finally, the current set of integration tests has been modified to ensure the main harness always rejects non-standard transactions. With this change, although we're using `simnet` parameters, the harness' transaction validation/acceptance should reflect that policy on mainnet,

### DIFF
--- a/config.go
+++ b/config.go
@@ -141,6 +141,8 @@ type config struct {
 	DropTxIndex        bool          `long:"droptxindex" description:"Deletes the hash-based transaction index from the database on start up and then exits."`
 	AddrIndex          bool          `long:"addrindex" description:"Maintain a full address-based transaction index which makes the searchrawtransactions RPC available"`
 	DropAddrIndex      bool          `long:"dropaddrindex" description:"Deletes the address-based transaction index from the database on start up and then exits."`
+	RelayNonStd        bool          `long:"relaynonstd" description:"Relay non-standard transactions regardless of the default settings for the active network."`
+	RejectNonStd       bool          `long:"rejectnonstd" description:"Reject non-standard transactions regardless of the default settings for the active network."`
 	onionlookup        func(string) ([]net.IP, error)
 	lookup             func(string) ([]net.IP, error)
 	oniondial          func(string, string) (net.Conn, error)
@@ -478,6 +480,26 @@ func loadConfig() (*config, []string, error) {
 		fmt.Fprintln(os.Stderr, usageMessage)
 		return nil, nil, err
 	}
+
+	// Set the default policy for relaying non-standard transactions
+	// according to the default of the active network. The set
+	// configuration value takes precedence over the default value for the
+	// selected network.
+	relayNonStd := activeNetParams.RelayNonStdTxs
+	switch {
+	case cfg.RelayNonStd && cfg.RejectNonStd:
+		str := "%s: rejectnonstd and relaynonstd cannot be used " +
+			"together -- choose only one"
+		err := fmt.Errorf(str, funcName)
+		fmt.Fprintln(os.Stderr, err)
+		fmt.Fprintln(os.Stderr, usageMessage)
+		return nil, nil, err
+	case cfg.RejectNonStd:
+		relayNonStd = false
+	case cfg.RelayNonStd:
+		relayNonStd = true
+	}
+	cfg.RelayNonStd = relayNonStd
 
 	// Append the network type to the data directory so it is "namespaced"
 	// per network.  In addition to the block database, there are other
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -78,6 +78,12 @@ type Policy struct {
 	// transactions that do not have enough priority to be relayed.
 	DisableRelayPriority bool
 
+	// RelayNonStd defines whether to relay non-standard transactions. If
+	// true, non-standard transactions will be accepted into the mempool
+	// and relayed. Otherwise, all non-standard transactions will be
+	// rejected.
+	RelayNonStd bool
+
 	// FreeTxRelayLimit defines the given amount in thousands of bytes
 	// per minute that transactions with no fee are rate limited to.
 	FreeTxRelayLimit float64
@@ -537,7 +543,7 @@ func (mp *TxPool) maybeAcceptTransaction(tx *btcutil.Tx, isNew, rateLimit bool) 
 
 	// Don't allow non-standard transactions if the network parameters
 	// forbid their relaying.
-	if !mp.cfg.ChainParams.RelayNonStdTxs {
+	if !mp.cfg.Policy.RelayNonStd {
 		err := checkTransactionStandard(tx, nextBlockHeight,
 			mp.cfg.TimeSource, mp.cfg.Policy.MinRelayTxFee)
 		if err != nil {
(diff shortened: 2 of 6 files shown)


davecgh:
Would you mind modifying the first commit which adds the policy config option to the mempool to also set it in `newServer`?

```diff
diff --git a/server.go b/server.go
index 60e3e2f..f78d2a3 100644
--- a/server.go
+++ b/server.go
@@ -2519,6 +2519,7 @@ func newServer(listenAddrs []string, db database.DB, chainParams *chaincfg.Param
        txC := mempool.Config{
                Policy: mempool.Policy{
                        DisableRelayPriority: cfg.NoRelayPriority,
+                       RelayNonStd:          chainParams.RelayNonStdTxs,
                        FreeTxRelayLimit:     cfg.FreeTxRelayLimit,
                        MaxOrphanTxs:         cfg.MaxOrphanTxs,
                        MaxOrphanTxSize:      defaultMaxOrphanTxSize,
```

I realize the next commit introduces the config options and sets it accordingly, however I'd prefer the first commit not to break the existing functionality in case it gets cherry-picked.

config.go
```diff
@@ -479,6 +481,24 @@ func loadConfig() (*config, []string, error) {
...
+	// selected network.
+	switch {
+	case cfg.RelayNonStd && cfg.RejectNonStd:
+		str := "%s: rejectnonstd and relaynonstd cannot be used " +
+			"together -- choose only one"
+		err := fmt.Errorf(str, funcName)
+		fmt.Fprintln(os.Stderr, err)
+		fmt.Fprintln(os.Stderr, usageMessage)
+		return nil, nil, err
+	case cfg.RejectNonStd && activeNetParams.RelayNonStdTxs:
```
davecgh: The logic in this section is seems a little hard to follow in my opinion.  How about the following which I think is a little clearer.  Start with the value configured by the network params, then override it if one of the flags is set.

```diff
diff --git a/config.go b/config.go
index 1b04b2f..b3af9c5 100644
--- a/config.go
+++ b/config.go
@@ -485,6 +485,7 @@ func loadConfig() (*config, []string, error) {
        // according to the default of the active network. The set
        // configuration value takes precedence over the default value for the
        // selected network.
+       relayNonStd := activeNetParams.RelayNonStdTxs
        switch {
        case cfg.RelayNonStd && cfg.RejectNonStd:
                str := "%s: rejectnonstd and relaynonstd cannot be used " +
@@ -493,11 +494,12 @@ func loadConfig() (*config, []string, error) {
                fmt.Fprintln(os.Stderr, err)
                fmt.Fprintln(os.Stderr, usageMessage)
                return nil, nil, err
-       case cfg.RejectNonStd && activeNetParams.RelayNonStdTxs:
-               cfg.RelayNonStd = false
-       case cfg.RelayNonStd && !activeNetParams.RelayNonStdTxs:
-               cfg.RelayNonStd = true
+       case cfg.RejectNonStd:
+               relayNonStd = false
+       case cfg.RelayNonStd:
+               relayNonStd = true
        }
+       cfg.RelayNonStd = relayNonStd

        // Append the network type to the data directory so it is "namespaced"
        // per network.  In addition to the block database, there are other
```

config.go
```diff
@@ -141,6 +141,8 @@ type config struct {
 	DropTxIndex        bool          `long:"droptxindex" description:"Deletes the hash-based transaction index from the database on start up and then exits."`
 	AddrIndex          bool          `long:"addrindex" description:"Maintain a full address-based transaction index which makes the searchrawtransactions RPC available"`
 	DropAddrIndex      bool          `long:"dropaddrindex" description:"Deletes the address-based transaction index from the database on start up and then exits."`
+	RelayNonStd        bool          `long:"relaynonstd" description:"Denotes if non-standard transactions should be relayed by btcd."`
```
davecgh: I'm not sure I would figure out the meaning of these from the descriptions.

What do you think of the following?

> Relay non-standard transactions regardless of the default settings for the active network.
> Reject non-standard transactions regardless of the default settings for the active network.

mempool/mempool.go
```diff
@@ -78,6 +78,12 @@ type Policy struct {
 	// transactions that do not have enough priority to be relayed.
 	DisableRelayPriority bool
 
+	// RelayNonStd defines whether to relay non-standard transactions. If
+	// true, non-standard transactions will be accepted into the mempool
+	// and relayed. Otherwise, all non-standard transactions will be
+	// ignored.
```
davecgh: I believe it will be rejected as opposed to ignored.

davecgh:
I like the addition of the mempool policy config option in the first commit.  I was highly considering making that same change in my mempool test harness infrastructure PR #750, so I'm glad to see it here.

Also, aside from the fact the new CLI params are needed to be able to properly run integration tests, I think it's a good idea to expose more policy controls such as this anyways.

mempool/mempool.go
Roasbeef: Correct. I've updated the comment.

config.go
Roasbeef: Agreed. I originally had something almost identical to that, but somehow convinced myself that the logic was incorrect (a bit of overthinking). The logic has been updated the clearer fragment you posted above.

config.go
Roasbeef: I like your suggested descriptions, they're clear and to the point. I've changed the descriptions in the PR to mirror your suggestions above.

Roasbeef:
Thanks. I really like the slight refactoring that have been getting in master over the past few weeks. In isolation, the changes seem insignificant, but as a whole they allow a new level of testability within the project as demonstrated by your mempool test harness PR. 

I'll update the first commit to also properly set set the config option in the server once we get to the squashing stage with this PR 👍 .

davecgh:
Changes look good.  I'll give it a spin later tonight to test the various combinations.  I know the test harness already tests that `--rejectnonstd`+`--simnet` is working, but I'd like verify the other combinations as well.

davecgh:
I noticed that `doc.go` was not updated with the new options.  That is required so they show up on on the [btcd godoc page](https://godoc.org/github.com/btcsuite/btcd).

Aside from that, I modified this slightly to print out the configured value and tried what I believe are all options.  Everything seems to be working correctly.  Results:

```bash
$ ./btcd
relay nonstd? false
$ ./btcd --testnet
relay nonstd? true
$ ./btcd --regtest
relay nonstd? true
$ ./btcd --simnet
relay nonstd? true
$ ./btcd --relaynonstd
relay nonstd? true
$ ./btcd --relaynonstd --testnet
relay nonstd? true
$ ./btcd --relaynonstd --regtest
relay nonstd? true
$ ./btcd --relaynonstd --simnet
relay nonstd? true
$ ./btcd --rejectnonstd
relay nonstd? false
$ ./btcd --rejectnonstd --testnet
relay nonstd? false
$ ./btcd --rejectnonstd --regtest
relay nonstd? false
$ ./btcd --rejectnonstd --simnet
relay nonstd? false
$ ./btcd --relaynonstd --rejectnonstd
loadConfig: rejectnonstd and relaynonstd cannot be used together -- choose only one
Use btcd -h to show usage
```

Roasbeef:
Great. I've just pushed a commit which updates the `doc.go` and additionally adds the new flag under the `Mempool Settings` header within the `sample-btcd.conf`

davecgh:
Looks good.  :+1:   Ready for squash then.  Please don't forget to update that first commit as previously discussed.

Roasbeef:
Alrighty, just pushed a squashed version of all the commits, with the first commit being amended to set the new `mempool` config option within the server. The second commit has also been modified to replace the `chaincfg` default with the new configuration option.

Please have a look over yourself to verify the correctness of the latest set of commits. After such a check, I think this is ready to be merged now 🎉

davecgh:
Reviewed squashed commits and retested options as per above.

OK

PR #750, opened 2016-08-23
mempool: Add basic test harness infrastructure.

davecgh:
This adds a basic test harness infrastructure for the mempool package which aims to make writing tests for it much easier.

The harness provides functionality for creating and signing transactions as well as a fake chain that provides utxos for use in generating valid transactions and allows an arbitrary chain height to be set.  In order to simplify transaction creation, a single signing key and payment address is used throughout and a convenience type for spendable outputs is provided.

The harness is initialized with a spendable coinbase output by default and the fake chain height set to the maturity height needed to ensure the provided output is in fact spendable as well as a policy that is suitable for testing.

Since tests are in the same package and each harness provides a unique pool and fake chain instance, the tests can safely reach into the pool policy, or any other state, and change it for a given harness without affecting the others.

In order to be able to make use of the existing `blockchain.Viewpoint` type, a `Clone` method has been added to the UtxoEntry type which allows the fake chain instance to keep a single view with the actual available unspent utxos while the mempool ends up fetching a subset of the view with the specifically requested entries cloned.

To demo the harness, this also contains a couple of tests which make use of it:

- `TestSimpleOrphanChain` -- Ensures an entire chain of orphans is properly accepted and connects up when the missing parent transaction is added
- `TestOrphanRejects` -- Ensure orphans are actually rejected when the flag on `ProcessTransactions` is set to reject them

### DIFF
--- a/mempool/mempool_test.go
+++ b/mempool/mempool_test.go
@@ -1,0 +1,463 @@
+// Copyright (c) 2016 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package mempool
+
+import (
+	"encoding/hex"
+	"reflect"
+	"sync"
+	"testing"
+
+	"github.com/btcsuite/btcd/blockchain"
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+)
+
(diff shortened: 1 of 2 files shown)


mempool/mempool_test.go
```diff
@@ -0,0 +1,463 @@
...
+	// orphans in the pool are linked.
+	acceptedTxns, err := harness.txPool.ProcessTransaction(chainedTxns[0],
+		false, false)
+	if err != nil {
+		t.Fatalf("ProcessTransaction: failed to accept valid "+
+			"orphan %v", err)
+	}
+	if len(acceptedTxns) != len(chainedTxns) {
+		t.Fatalf("ProcessTransaction: reported accepted transactions "+
+			"lengh does not match expected -- got %d, want %d",
```
davecgh: length

dajohi:
OK

PR #753, opened 2016-08-25
wire: implement feefilter message

dajohi:
feefilter is used to request the receiving peer does not announce any
transactions below the specified minimum fee rate.

NOTE: This only implements the new message type in the `wire` package.  Further PRs will be needed to the `peer` package and `btcd` itself to implement and respond to the message.

### DIFF
--- a/wire/protocol.go
+++ b/wire/protocol.go
@@ -12,7 +12,7 @@ import (
 
 const (
 	// ProtocolVersion is the latest protocol version this package supports.
-	ProtocolVersion uint32 = 70012
+	ProtocolVersion uint32 = 70013
 
 	// MultipleAddressVersion is the protocol version which added multiple
 	// addresses per message (pver >= MultipleAddressVersion).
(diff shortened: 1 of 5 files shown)


dajohi:
See https://github.com/bitcoin/bitcoin/commit/9e072a6e

davecgh:
Nice.  Please call out that this is BIP0133 and add it to the list of supported BIPs in `doc.go`.

dajohi:
updated.

stevenroose:
Interesting to see how new message types are to be introduced in btcd!
I suppose corresponding logic is to be introduced in a separate PR?

davecgh:
@stevenroose Correct.  The btcd codebase is intentionally modular.  The `wire` package just handles the protocol, so any other projects that make use of it (SPV wallets for example) can independently implement and make use of the message.

Changes will also be necessary to the `peer` package and btcd itself in order to actually recognize the new message type and implement the handling logic.

wire/protocol.go
```diff
@@ -39,6 +39,10 @@ const (
 	// service flag.
 	BIP0111Version uint32 = 70011
 
+	// FeeFilterVersion is the protocol version which added a new
+	// feefilter message.
+	FeeFilterVersion uint32 = 70013
```
davecgh: Mind putting this after `SendHeadersVersion` so they are in increasing order?  I also see that `RejectVersion` managed to sneak in out of order too.

dajohi: Sorted both.

PR #755, opened 2016-08-28
Implement infrastructure for BIP0009.

davecgh:
This adds all of the infrastructure needed to support BIP0009 soft forks.

- Add new configuration options to the chaincfg package which allows the rule deployments to be defined per chain
- Implement code to calculate the threshold state as required by BIP0009
  - Use threshold state caches that are stored to the database in order to accelerate startup time
  - Remove caches that are invalid due to definition changes in the params including additions, deletions, and changes to existing entries
- Detect and warn when a new unknown rule is about to activate or has been activated in the block connection code
- Detect and warn when 50% of the last 100 blocks have unexpected versions.
- Remove the latest block version from wire since it no longer applies
- Add a version parameter to the `wire.NewBlockHeader` function since the default is no longer available
- Update the miner block template generation code to use the calculated block version based on the currently defined rule deployments and their threshold states as of the previous block
- Add tests for new error type
- Add tests for threshold state cache
- Modify the `rpctest` package to provide its own constant for the block version since it is no longer available via the wire package
- Add integration tests to ensure the miner generates blocks according to the BIP0009 rules.

Outstanding items:
- [x] Implement integration tests to ensure the state transitions work as expected using the `rpctest` framework.

### DIFF
--- a/blockchain/chainio.go
+++ b/blockchain/chainio.go
@@ -11,6 +11,7 @@ import (
 	"math/big"
 	"sort"
 
+	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/database"
 	"github.com/btcsuite/btcd/wire"
@@ -37,6 +38,27 @@ var (
 	// utxoSetBucketName is the name of the db bucket used to house the
 	// unspent transaction output set.
 	utxoSetBucketName = []byte("utxoset")
+
+	// thresholdBucketName is the name of the db bucket used to house cached
+	// threshold states.
+	thresholdBucketName = []byte("thresholdstate")
+
+	// numDeploymentsKeyName is the name of the db key used to store the
+	// number of saved deployment caches.
+	numDeploymentsKeyName = []byte("numdeployments")
+
+	// deploymentBucketName is the name of the db bucket used to house the
+	// cached threshold states for the actively defined rule deployments.
+	deploymentBucketName = []byte("deploymentcache")
+
+	// deploymentStateKeyName is the name of the db key used to store the
+	// deployment state associated with the threshold cache for a given rule
+	// deployment.
+	deploymentStateKeyName = []byte("deploymentstate")
+
+	// warningBucketName is the name of the db bucket used to house the
+	// cached threshold states for unknown rule deployments.
+	warningBucketName = []byte("warningcache")
 
 	// byteOrder is the preferred byte order used for serializing numeric
 	// fields for storage in the database.
@@ -73,6 +95,13 @@ func (e errDeserialize) Error() string {
 func isDeserializeErr(err error) bool {
 	_, ok := err.(errDeserialize)
 	return ok
+}
+
+// isDbBucketNotFoundErr returns whether or not the passed error is a
+// database.Error with an error code of database.ErrBucketNotFound.
+func isDbBucketNotFoundErr(err error) bool {
+	dbErr, ok := err.(database.Error)
+	return ok && dbErr.ErrorCode == database.ErrBucketNotFound
 }
 
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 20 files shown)


stevenroose:
Nice to see BIP9 support! 

Question: Will there be support for querying BIP9 information via RPC? I'd be very interested in a way to retrieve both the current state and historical data.

blockchain/chainio.go
```diff
@@ -38,6 +39,27 @@ var (
 	// unspent transaction output set.
 	utxoSetBucketName = []byte("utxoset")
 
+	// thresholdBucketName is the name of the db bucked used to house cached
```
Roasbeef: bucked -> bucket

The same typo is repeated in several of the variable comments below.

blockchain/chainio.go
```diff
@@ -1403,3 +1432,501 @@ func (b *BlockChain) HeightRange(startHeight, endHeight int32) ([]chainhash.Hash
 	})
 	return hashList, err
 }
+
+// -----------------------------------------------------------------------------
+// The threshold state consists of individual threshold cache buckets for each
+// cache id under one main threshold state bucket.  Each threshold cache bucket
+// contains entries keyed by the block hash for the final block in each window
+// and their associated threshold states as well as the assocaited deployment
```
Roasbeef: assocaited -> associated

blockchain/chainio.go
```diff
@@ -1403,3 +1432,501 @@ func (b *BlockChain) HeightRange(startHeight, endHeight int32) ([]chainhash.Hash
...
+	// the warning cache associated with their bits.
+	numStoredDeployments, err := dbFetchNumDeployments(cachesBucket)
+	if err != nil {
+		return err
+	}
+	definedDeployments := uint32(len(deployments))
+	for i := definedDeployments; i < numStoredDeployments; i++ {
+		// Nothing to do when nothing is stored for the deployment.
+		deployBucketKey := thresholdCacheBucket(deploymentBucketName, i)
+		deployBucket := cachesBucket.Bucket(deployBucketKey)
```
Roasbeef: From my reading, and the above comment, it seems that there should be a conditional `continue` here to avoid passing around a possibly `nil` bucket (which may lead to a panic): 
```go
if depolyBucket == nil {
    continue
}
```

blockchain/chainio.go
```diff
@@ -1403,3 +1432,501 @@ func (b *BlockChain) HeightRange(startHeight, endHeight int32) ([]chainhash.Hash
...
+
+		// Load the deployment details the cache was created for from
```
Roasbeef: Minor nit, but this comment is inaccurate: this loop deletes all deployment info+caches whose index is higher than the highest currently defined deployment. This deletion is unconditional. Changes in defined deployments are checked below and possibly invalidated accordingly.

mining.go
```diff
@@ -20,13 +20,9 @@ import (
...
-	// generatedBlockVersion is the version of the block being generated.
-	// It is defined as a constant here rather than using the
-	// wire.BlockVersion constant since a change in the block version
-	// will require changes to the generated block.  Using the wire constant
-	// for generated block version could allow creation of invalid blocks
-	// for the updated version.
-	generatedBlockVersion = 4
+	// minHighPriority is the minimum priority value that allows a
+	// transaction to be considered high priority.
+	minHighPriority = btcutil.SatoshiPerBitcoin * 144.0 / 250
```
Roasbeef: It appears that this definition was mistakenly added. This currently re-defines the `mempool.MinHighPriority` value.

Roasbeef:
Happy to see this PR finally land! 

I really like the `conditionChecker` interface. It's abstract nature will make to possibly extend the current simple bit-counting condition thresholds to much complex activation thresholds in the future. 

I'll get started on a PR which includes the deployment parameters for the CSV-package soft-fork on all chains, and which also adds logic to the BIP 68, 112, and 113 to PR's to enforce the new rules during block validation and mining.

wire/blockheader.go
```diff
@@ -104,7 +101,7 @@ func NewBlockHeader(prevHash *chainhash.Hash, merkleRootHash *chainhash.Hash,
 	// Limit the timestamp to one second precision since the protocol
 	// doesn't support better.
 	return &BlockHeader{
-		Version:    BlockVersion,
+		Version:    0,
```
davecgh: Fair point.  I thought about that, but I wasn't really sure it's worth breaking the public API over since the version will have to be set regardless (typically after the header is created) by looking at the various threshold states after this goes in.  I'll go ahead and change it though since the only thing that actually even calls this right now are the tests in the peer package.

davecgh:
I´ve updated this with 2 new commits.  The first adds a new integration package along with tests to ensure the mining code is correctly creating blocks that use the new BIP0009 functionality.  The second moves the RPC server integration tests into the new package.

I´ll be addressing all review items next.

blockchain/thresholdstate.go
```diff
@@ -0,0 +1,322 @@
+// Copyright (c) 2016 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package blockchain
```
davecgh: This might be something reasonable to consider for the future, but given the current code is already complete, reviewed, and tested, I don´t think it would make sense to do it for this PR.

davecgh:
@stevenroose: This PR doesn't have support for querying BIP9 information via RPC, but it is definitely something I'd like to add.  That functionality would allow additional integration testing to be performed as well since the test code then query if the state changed as expected.

The current integration tests use blocks generated by the miner and checks the bits to detect the state changes worked as intended, which is also really useful, but the ability to query the state from the RPC server would definitely allow more fine grained testing as well.

mining.go
davecgh: Nice catch.  It was added before the `mempool` package existed and I missed removing it when rebasing over it.

blockchain/chainio.go
davecgh: Fixed.

blockchain/chainio.go
davecgh: Fixed.

blockchain/chainio.go
```diff
@@ -1403,3 +1432,501 @@ func (b *BlockChain) HeightRange(startHeight, endHeight int32) ([]chainhash.Hash
...
+func (b *BlockChain) markThresholdCachesFlushed() {
+	for i := 0; i < len(b.deploymentCaches); i++ {
+		b.deploymentCaches[i].MarkFlushed()
+	}
+	for i := 0; i < len(b.warningCaches); i++ {
+		b.warningCaches[i].MarkFlushed()
+	}
+}
+
+// dbFetchTresholdCaches uses an existing database transaction to retrieve
```
davecgh: Fixed

blockchain/chainio.go
davecgh: Yep, you're correct that it should have the check.  Added.

blockchain/chainio.go
davecgh: Updated.  Looks like a case of copy paste that I forgot to finish (the code to check for changes was written first).

davecgh:
All review items have been addresses.  This is ready for a final review.

dajohi:
Tested on mainnet and testnet without issue.  OK

davecgh:
Since everyone has now finished their reviews, I squashed commits that addressed the review items into the original commit.

This is now ready for master, but I'm going to hold off merging it until the PRs for the entire CSV softfork package are reviewed and ready for merge.  The reason for that is because this PR adds code which complains about the fact an unknown soft fork is active (in particular the CSV softfork package), so I'd rather avoid having a bunch of issues opened about something that is already being worked on when people see the warning and think something is wrong.

davecgh:
Rebased to latest master and resolved conflicts.

chaincfg/params.go
```diff
@@ -60,6 +61,37 @@ type DNSSeed struct {
...
+const (
+	// DeploymentTestDummy defines the rule change deployment ID for testing
+	// purposes.
+	DeploymentTestDummy = iota
+
+	// NOTE: MaxDeployments must always come last since it is used to
+	// determine how many defined deployments there currently are.
+
+	// MaxDeployments is how many defined deployments there currently are.
+	MaxDeployments
```
stevenroose: Max suggests this is the maximum number there can ever be.. Wouldn't `NbDeployments` or `NbDefinedDeployments` be more suitable?

davecgh: Fair point.  It actually is used as a max at some places in the code as in `if deploymentID > ....`, however, that usage doesn't need to influence the name here and it wouldn't be unclear to read `if deploymentID > cfg.DefinedDeployments { ....`.

I'll change it to `DefinedDeployments` even though it's longer!

davecgh: Done.

davecgh:
Rebased to latest master.

davecgh:
I've added a commit which only warns about unknown rules when the chain is already consider current which is a slight variation of a suggestion by @Roasbeef.

dajohi:
OK

Firescar96:
Can this get merged in?

davecgh:
Yes, I will be squashing and getting this merged tomorrow.  I'm getting to a stopping point on another project at the moment, but this is next.

PR #756, opened 2016-08-29
peer: Implement feefilter p2p message (bip0133)

dajohi:
Requires #753 

NOTE: This only implements the new message type in the `peer` package. Further PRs will be needed to btcd itself to implement and respond to the message.

### DIFF
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -27,7 +27,7 @@ import (
 
 const (
 	// MaxProtocolVersion is the max protocol version the peer supports.
-	MaxProtocolVersion = wire.SendHeadersVersion
+	MaxProtocolVersion = wire.FeeFilterVersion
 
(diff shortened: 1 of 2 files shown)


davecgh:
Looks good overall assuming the underlying wire commit is accurate which I haven't reviewed closely yet.

However, I did notice there is no test to ensure the `feeFilter` field value is actually set properly and can be retrieved via the `FeeFilter` function.  It only appears to be testing that the callback is invoked.

davecgh:
I would suggest the peer shouldn't be tracking the fee filter, rather just offer the callback handler so the consumer of the peer package can choose whether or not they want to store and make use it or not.  For example, an SPV client clearly would not have any need to store it.

dajohi:
OK, updated.

PR #758, opened 2016-09-09
config: Replace log outputs with fmt.Fprintln.

### LINKED ISSUE #757: Remove btcdLog.Warn before logging is initialized

marcopeereboom:
Fixes #757.

### DIFF
--- a/config.go
+++ b/config.go
@@ -834,8 +834,8 @@ func loadConfig() (*config, []string, error) {
 
 		if cfg.TorIsolation &&
 			(cfg.ProxyUser != "" || cfg.ProxyPass != "") {
-			btcdLog.Warn("Tor isolation set -- overriding " +
-				"specified proxy user credentials")
+			fmt.Fprintln(os.Stderr, "Tor isolation set -- "+
+				"overriding specified proxy user credentials")
 		}
 
 		proxy := &socks.Proxy{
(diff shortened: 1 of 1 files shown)


davecgh:
Nice catch.  OK.

config.go
```diff
@@ -833,8 +834,8 @@ func loadConfig() (*config, []string, error) {
 
 		if cfg.TorIsolation &&
 			(cfg.ProxyUser != "" || cfg.ProxyPass != "") {
-			btcdLog.Warn("Tor isolation set -- overriding " +
-				"specified proxy user credentials")
+			fmt.Fprintln(os.Stderr, "Tor isolation set -- "+
```
davecgh: It's safe to use logging at this point since it has been initialized.  Any reason to switch it?

config.go
```diff
@@ -871,8 +872,9 @@ func loadConfig() (*config, []string, error) {
 
 		if cfg.TorIsolation &&
 			(cfg.OnionProxyUser != "" || cfg.OnionProxyPass != "") {
-			btcdLog.Warn("Tor isolation set -- overriding " +
-				"specified onionproxy user credentials ")
+			fmt.Fprintln(os.Stderr, "Tor isolation set -- "+
```
davecgh: Same as previous.

config.go
marcopeereboom: That is correct, however, people tend to copy paste and now by having everything identical there is no excuse to do it wrong.  Larting can then ensue appropriately.

config.go
marcopeereboom: Ditto.

config.go
davecgh: Fair enough.  I have no objection to that.

davecgh:
OK

PR #759, opened 2016-09-12
mining: fix duplicate transactions in prio heap

### LINKED ISSUE #691: Transactions with the same dependent tx can be readded to prio queue multiple times

tuxcanfly:
This prevents the ability for duplicate transactions to be added to the mining priority heap when a transaction spends multiple outputs from the same input transaction by converting the dependency tracking to use a map keyed by the transaction hash instead of a linked list.

It follows up the discussion in #692 and updates the diff provided in the comments there to use the `chainhash` package.

Fixes #691

### DIFF
--- a/mining.go
+++ b/mining.go
@@ -6,7 +6,6 @@ package main
 
 import (
 	"container/heap"
-	"container/list"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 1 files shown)


davecgh:
Mind removing the PR numbers from the commit and putting it in the PR description instead?  Been trying to make an effort to make the commits more portable to other projects like Decred.

PR #760, opened 2016-09-12
config: Reconcile differences between btcd/dcrd

marcopeereboom:


dajohi:
OK, with a new commit message

marcopeereboom:
Don't merge this yet.  Found one more bug.

marcopeereboom:
This time is for reals!!!

marcopeereboom:
I want to have a discussion about this particular tool.  I don't like the defaults as is.

If there is no btcd.conf it aborts; but everything could have been passed on the command line.  At the very least we need to ignore tbtcd.conf not existing error if rpcuser and rpcpassword are passed on the command line.  This is currently a bit of a mess.

Any ideas?

jrick:
I agree that there should be no application abort when there is no btcd.conf in the default location.  A warning for this is fine.  All options can be passed either as command line arguments or settings in the dcrctl.conf.

davecgh:
The current master does not abort if there is no `btcd.conf` when the `btcctl.conf` file exists:

> $ mv $LOCALAPPDATA/Btcd/btcd.conf $LOCALAPPDATA/Btcd/btcd.conf.bak
> $ btcctl getblockcount
429540

davecgh:
In fact, the current master doesn't abort when there is neither a `btcd.conf` nor a `btcctl.conf`.  It just shows an error message and then succeeds:

> $ btcctl --rpcuser=**** --rpcpass=**** getblockcount
> Error creating a default config file: open C:\Users\davec\AppData\Local\Btcd\btcd.conf: The system cannot find the file specified.
> 429541

davecgh:
What is the purpose of this PR?  I can't tell from the PR title or commit description, and as I pointed out previously, master already works correctly.

marcopeereboom:
For now this is a placeholder to make btcd/dcrd as identical as possible in config.go.  More to come and I'll make the title pretty and everything.

davecgh:
This does not build on windows:

># github.com/btcsuite/btcd
> .\service_windows.go:40: undefined: btcdHomeDir

marcopeereboom:
Oops, that's embarrassing.  I'll push a fix shortly.

PR #761, opened 2016-09-13
Remove extra backtick in README.md

chris-martin:


davecgh:
Nice catch.  Thanks.

OK

Issue #697, opened 2016-05-19
Two btcd instanced on same machine keep losing connection


stevenroose:
I setup one daemon that was synced up and listening to 8335 and another one with `connect=127.0.0.1:8335` that started from scratch.

However, they kept losing connection:

```
00:02:56 2016-05-19 [INF] BMGR: Syncing to block height 412025 from peer localhost:8335
00:02:56 2016-05-19 [INF] BMGR: Downloading headers for blocks 1 to 11111 from peer localhost:8335
00:04:41 2016-05-19 [INF] BMGR: Lost peer localhost:8335 (outbound)
```

I found this in the debug logs and the probable cause:

```
00:56:30 2016-05-19 [DBG] PEER: Peer 127.0.0.1:8335 (outbound) appears to be stalled or misbehaving, headers timeout -- disconnecting
```

That message occurred on the new (connecting) instance.

I talked about it with @davecgh and he could not immediately identify the issue. The receiving instance does not mention anything special about the other instance, just that it received a `getheaders` message.

Here are some more possibly relevant logs:
https://gist.github.com/stevenroose/5b69ed22d8fd35d8c04c564c0f3733b7
https://gist.github.com/stevenroose/a06184820c7b44c96d9728072ae2886a
https://gist.github.com/stevenroose/7e6265c06b505b4e51733b6b15eae883

dajohi:
using different datadir's?

dajohi:
Can you run with this diff on the server that has the blockchain?  See if any more logs pop up...

```
diff --git a/server.go b/server.go
index 008bb9a..d974ad5 100644
--- a/server.go
+++ b/server.go
@@ -691,6 +691,7 @@ func (sp *serverPeer) OnGetBlocks(p *peer.Peer, msg *wire.MsgGetBlocks) {
 func (sp *serverPeer) OnGetHeaders(p *peer.Peer, msg *wire.MsgGetHeaders) {
        // Ignore getheaders requests if not in sync.
        if !sp.server.blockManager.IsCurrent() {
+               peerLog.Warnf("Peer %v sent getheaders, but we are not current", p)
                return
        }
 
@@ -709,6 +710,7 @@ func (sp *serverPeer) OnGetHeaders(p *peer.Peer, msg *wire.MsgGetHeaders) {
                // to do.  Just return.  This behavior mirrors the reference
                // implementation.
                if endIdx == math.MaxInt32 {
[...]

PR #769, opened 2016-09-21
config: Add --whitelist support.

dajohi:
--whitelist specifies IP networks and IPs of remote peers that will not
be banned for misbehaving.

### DIFF
--- a/server.go
+++ b/server.go
@@ -223,6 +223,7 @@ type serverPeer struct {
 	relayMtx       sync.Mutex
 	disableRelayTx bool
 	sentAddrs      bool
+	isWhitelisted  bool
 	filter         *bloom.Filter
 	knownAddresses map[string]struct{}
 	banScore       connmgr.DynamicBanScore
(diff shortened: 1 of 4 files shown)


stevenroose:
This would fix #697.

server.go
```diff
@@ -1569,11 +1575,41 @@ func (s *server) listenHandler(listener net.Listener) {
...
+			sp.isWhitelisted = s.isWhitelisted(addr)
+		}
 	}
 	s.wg.Done()
 	srvrLog.Tracef("Listener handler done for %s", listener.Addr())
 }
 
+// isWhitelisted returns whether the IP address is included in the whitelisted
+// networks and IPs.
+func (s *server) isWhitelisted(addr net.Addr) bool {
```
davecgh: Any reason this is defined on the server type as opposed to being a standalone function?  The code doesn't appear to use anything from the server instance.

davecgh:
I'm fairly certain this will not work well with tor because `.RemoteAddr()` on a proxy connection is the proxy itself.  Thus, if you were to do the most common proxy setup of using tor (which will be on 127.0.0.1) and whitelist 127.0.0.1, you'd effectively be whitelisting all nodes as this is currently written.

dajohi:
The user should run Tor on a different IP then.

davecgh:
This is a NACK until the localhost issue via proxy can be resolved.

davecgh:
After thinking about this some more, I think there should be check to see if both a proxy address and a whitelist addresses overlap and, if they do, disallow it.  That would solve the above concerns.

dajohi:
rebased.

dajohi:
Rebased

PR #770, opened 2016-09-21
Update dependencies and API

dajohi:


davecgh:
dependencies

glide.lock
```diff
@@ -1,19 +1,19 @@
-hash: 9b8ff781a12daad991983e9a421a320e905eecc4e9ff0b0643e790f472bc78c8
-updated: 2016-08-16T23:18:48.976672741-05:00
+hash: 8f89ba1940f8798b8a0449810532e3f0bc552ed97d907bc3447a8941a2ea3a4a
```
davecgh: Do you have changes to your local `glide.yml`?  The hash shouldn't change if `glide.yml` didn't change.

davecgh: Alright, so I guess `glide.lock` was manually updated previously and the hash wasn't updated properly, so this is fine.

PR #773, opened 2016-09-26
rpctest: Gate rpctest-based behind a build tag.

### LINKED ISSUE #772: Unable to successfully execute btcd tests.
With commit 99165eb using Go version 1.6.2 I always get the following error:

```
$ cd $GOPATH/src/github.com/btcsuite/btcd
$ go test -v
$ unable to setup test chain:  connection timeout

$ cd $GOPATH/src/github.com/btcsuite/btcd/rpctest
$ go test -v
$ unable to setup test chain:  connection timeout
```

I tried increasing [`maxConnRetries` in `rpctest/rpc_harness.go`](https://github.com/btcsuite/btcd/blob/99165eb5580b0cb3af1401a9dd641af867852300/rpctest/rpc_harness.go#L170) as a quick guess to a very high value but still no luck; the command just blocks.

I see tests in all the other packages pass. Any help on what am I doing wrong will be appreciated.

davecgh:
This adds a new build tag named `rpctest` which must be set in order for `rpctest`-based tests to be executed.  The new build tag is also added to the `goclean.sh` script which is executed by Travis during continuous integration builds.

This change is being made because the `rpctest` framework requires additional careful user configuration to ensure the version of `btcd` under test can be programmatically launched from the system path with
all of the necessary ports open whereas all of the other tests are self-contained within the test binary itself.

Since said additional configuration is typically not done, it leads to a lot of false positives.  Putting the tests behind a build tag allows them to remain to be available and run during continuous integration without imposing the additional configuration requirements on users.

Closes #772

### DIFF
--- a/goclean.sh
+++ b/goclean.sh
@@ -12,7 +12,7 @@ set -ex
 test -z "$(go fmt $(glide novendor) | tee /dev/stderr)"
 test -z "$(for package in $(glide novendor); do golint $package; done | grep -v 'ALL_CAPS\|OP_\|NewFieldVal' | tee /dev/stderr)"
 test -z "$(go vet $(glide novendor) 2>&1 | tee /dev/stderr)"
-env GORACE="halt_on_error=1" go test -v -race $(glide novendor)
+env GORACE="halt_on_error=1" go test -v -race -tags rpctest $(glide novendor)
 
 # Run test coverage on each subdirectories and merge the coverage profile.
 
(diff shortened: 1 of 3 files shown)


dajohi:
OK

cjepson:
OK

Roasbeef:
Posthumous OK. 

I had planned to suggest this given that the Travis environment seemed stable w.r.t to proper execution of the tests while we were altering either `rpctest` or the execution of the tests to accommodate difficult to reproduce bug repots.

PR #628, opened 2016-02-16
txscript: Implement CheckSequenceVerify (BIP0112)

dajohi:
This implements [BIP 112](https://github.com/bitcoin/bips/blob/master/bip-0112.mediawiki) which introduces a new `OP_CHECKSEQUENCEVERIFY` opcode  to `txscript` that replaces `OP_NOP3`.

This PR currently only enforces the behavior when accepting candidate transactions to the mempool.  It relies on [BIP 9](https://github.com/bitcoin/bips/blob/master/bip-0009.mediawiki) (btcd PR #755) for deployment as part of the CSV package soft fork (btcd PR #775).

The new opcode is ultimately intended to be combined with [BIP 68](https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki) (btcd PR #713) which repurposes the sequence number of a transaction's input (`wire.TxIn.Sequence)` as a consensus-enforced *relative* locktime which together allows execution pathways of a script to be restricted based on the age of the output being spent

Bitcoin Core commit https://github.com/bitcoin/bitcoin/commit/53e53a33c939949665f60d5eeb82abbb21f97128

PR #775, opened 2016-09-27
Implement the CSV-Package soft-fork via BIP0009 VersionBits

Roasbeef:
This commit fully implements the CSV-package soft-fork. The CSV-package soft-fork bundles the deployment of the following BIPs: [68](https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki), [112](https://github.com/bitcoin/bips/blob/master/bip-0112.mediawiki), and [113](https://github.com/bitcoin/bips/blob/master/bip-0113.mediawiki). The activation logic of the soft-fork is governed by [BIP 9](https://github.com/bitcoin/bips/blob/master/bip-0009.mediawiki), VersionBits. 

This PR currently is built upon a combined branch includes #713, #628, #712, and #755. Therefore, **only the last 3 commits** should be examined during code-review. As the aforementioned pending PR's re merged into master, this PR will be updated accordingly to increasingly reflect the final diff. 

A summary of the proposed changes follows: 
- A new `ConsensusDeployment` has been added to the chain parameters for mainnet, testnet, simnet, and regtestnet. This new deployment defines the BIP0009 parameters (bit #, start time, end time) for the CSV-package soft-fork. A new constant, `DeploymentCSV` has been added which uniquely identifies the new deployment within the `Deployments` array for each of the target networks. 
- During non-contextual, and contextual block validation, we now examine the `ThresholdState` for the `CSV` deployment, activating the observance of the new consensus rules once the soft-fork switches to the `ThresholdActive` state. 
  - If the soft-fork is active, then MTP is used for transaction finality checks, the CSV op-code is recognized and validated during Script execution, and a transaction can only be included within a block if the sequence locks for _all_ its inputs have been met. 
- Finally, when selecting transactions for inclusion into the current block template, the version bits state is again checked in order to guard the observance of the new consensus rules. If the soft-fork is active, then transactions can only be included into the to-be-constructed block iff they don't violate any of the new consensus rules.

### DIFF
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -351,17 +351,15 @@ func (b *BlockChain) CalcSequenceLock(tx *btcutil.Tx, utxoView *UtxoViewpoint,
 	b.chainLock.Lock()
 	defer b.chainLock.Unlock()
 
-	return b.calcSequenceLock(tx, utxoView, mempool)
+	return b.calcSequenceLock(b.bestNode, tx, utxoView, mempool)
 }
 
 // calcSequenceLock computes the relative lock-times for the passed
 // transaction. See the exported version, CalcSequenceLock for further details.
 //
 // This function MUST be called with the chain state lock held (for writes).
-func (b *BlockChain) calcSequenceLock(tx *btcutil.Tx, utxoView *UtxoViewpoint,
-	mempool bool) (*SequenceLock, error) {
-
-	mTx := tx.MsgTx()
+func (b *BlockChain) calcSequenceLock(node *blockNode, tx *btcutil.Tx,
+	utxoView *UtxoViewpoint, mempool bool) (*SequenceLock, error) {
 
 	// A value of -1 for each relative lock type represents a relative time
 	// lock value that will allow a transaction to be included in a block
@@ -370,20 +368,42 @@ func (b *BlockChain) calcSequenceLock(tx *btcutil.Tx, utxoView *UtxoViewpoint,
 	// activated.
 	sequenceLock := &SequenceLock{Seconds: -1, BlockHeight: -1}
 
+	// The sequence locks semantics are always active for transactions
+	// within the mempool.
+	csvSoftforkActive := mempool
+
+	// If we're performing block validation, then we need to query the BIP9
+	// state.
+	if !csvSoftforkActive {
+		prevNode, err := b.index.PrevNodeFromNode(node)
+		if err != nil {
+			return nil, err
+		}
+
+		// Obtain the latest BIP9 version bits state for the
+		// CSV-package soft-fork deployment. The adherence of sequence
+		// locks depends on the current soft-fork state.
+		csvState, err := b.deploymentState(prevNode, chaincfg.DeploymentCSV)
+		if err != nil {
+			return nil, err
+		}
+		csvSoftforkActive = csvState == ThresholdActive
+	}
+
 	// If the transaction's version is less than 2, and BIP 68 has not yet
 	// been activated then sequence locks are disabled. Additionally,
 	// sequence locks don't apply to coinbase transactions Therefore, we
 	// return sequence lock values of -1 indicating that this transaction
 	// can be included within a block at any given height or time.
-	// TODO(roasbeef): check version bits state or pass as param
-	// * true should be replaced with a version bits state check
-	sequenceLockActive := mTx.Version >= 2 && (mempool || true)
+	mTx := tx.MsgTx()
+	sequenceLockActive := mTx.Version >= 2 && csvSoftforkActive
 	if !sequenceLockActive || IsCoinBase(tx) {
 		return sequenceLock, nil
 	}
 
-	// Grab the next height to use for inputs present in the mempool.
-	nextHeight := b.BestSnapshot().Height + 1
+	// Grab the next height from the PoV of the passed blockNode to use for
+	// inputs present in the mempool.
+	nextHeight := node.height + 1
 
 	for txInIndex, txIn := range mTx.TxIn {
 		utxo := utxoView.LookupEntry(&txIn.PreviousOutPoint.Hash)
@@ -421,9 +441,8 @@ func (b *BlockChain) calcSequenceLock(tx *btcutil.Tx, utxoView *UtxoViewpoint,
 			// compute the past median time for the block prior to
 			// the one which included this referenced output.
 			// TODO: caching should be added to keep this speedy
-			inputDepth := uint32(b.bestNode.height-inputHeight) + 1
-			blockNode, err := b.index.RelativeNode(b.bestNode,
-				inputDepth)
+			inputDepth := uint32(node.height-inputHeight) + 1
+			blockNode, err := b.index.RelativeNode(node, inputDepth)
 			if err != nil {
 				return sequenceLock, err
 			}
--- a/blockchain/chain_test.go
+++ b/blockchain/chain_test.go
@@ -122,7 +122,7 @@ func TestCalcSequenceLock(t *testing.T) {
 	netParams := &chaincfg.SimNetParams
 
 	// Create a new database and chain instance to run tests against.
-	chain, teardownFunc, err := chainSetup("haveblock", netParams)
+	chain, teardownFunc, err := chainSetup("calcseqlock", netParams)
 	if err != nil {
 		t.Errorf("Failed to setup chain instance: %v", err)
 		return
@@ -190,7 +190,8 @@ func TestCalcSequenceLock(t *testing.T) {
 		prevBlock = block
 	}
 
-	// Create with all the utxos within the create created above.
+	// Create a utxo view with all the utxos within the blocks created
+	// above.
 	utxoView := blockchain.NewUtxoViewpoint()
 	for blockHeight, blockWithMTP := range blocksWithMTP {
 		for _, tx := range blockWithMTP.block.Transactions() {
@@ -200,8 +201,9 @@ func TestCalcSequenceLock(t *testing.T) {
 	utxoView.SetBestHash(blocksWithMTP[len(blocksWithMTP)-1].block.Hash())
 
 	// The median time calculated from the PoV of the best block in our
-	// test chain. For unconfirmed inputs, this value will be used since th
-	// MTP will be calculated from the PoV of the yet-to-be-mined block.
+	// test chain. For unconfirmed inputs, this value will be used since
+	// the MTP will be calculated from the PoV of the yet-to-be-mined
+	// block.
 	nextMedianTime := int64(1401292712)
 
 	// We'll refer to this utxo within each input in the transactions
--- a/blockchain/chain_test.go
+++ b/blockchain/chain_test.go
@@ -5,11 +5,15 @@
 package blockchain_test
 
 import (
+	"bytes"
 	"testing"
+	"time"
 
 	"github.com/btcsuite/btcd/blockchain"
+	"github.com/btcsuite/btcd/btcec"
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/rpctest"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcutil"
 )
@@ -115,15 +119,10 @@ func TestHaveBlock(t *testing.T) {
 // combinations of inputs to the CalcSequenceLock function in order to ensure
 // the returned SequenceLocks are correct for each test instance.
 func TestCalcSequenceLock(t *testing.T) {
-	fileName := "blk_0_to_4.dat.bz2"
-	blocks, err := loadBlocks(fileName)
-	if err != nil {
-		t.Errorf("Error loading file: %v\n", err)
-		return
-	}
+	netParams := &chaincfg.SimNetParams
 
 	// Create a new database and chain instance to run tests against.
-	chain, teardownFunc, err := chainSetup("haveblock", &chaincfg.MainNetParams)
+	chain, teardownFunc, err := chainSetup("haveblock", netParams)
 	if err != nil {
 		t.Errorf("Failed to setup chain instance: %v", err)
 		return
@@ -134,45 +133,90 @@ func TestCalcSequenceLock(t *testing.T) {
 	// maturity to 1.
 	chain.TstSetCoinbaseMaturity(1)
 
-	// Load all the blocks into our test chain.
-	for i := 1; i < len(blocks); i++ {
-		_, isOrphan, err := chain.ProcessBlock(blocks[i], blockchain.BFNone)
+	// Create a test mining address to use for the blocks we'll generate
+	// shortly below.
+	k := bytes.Repeat([]byte{1}, 32)
+	_, miningPub := btcec.PrivKeyFromBytes(btcec.S256(), k)
+	miningAddr, err := btcutil.NewAddressPubKey(miningPub.SerializeCompressed(),
+		netParams)
+	if err != nil {
+		t.Fatalf("unable to generate mining addr: %v", err)
+	}
+
+	// We'll keep track of the previous block for back pointers in blocks
+	// we generated, and also the generated blocks along with the MTP from
+	// their PoV to aide with our relative time lock calculations.
+	var prevBlock *btcutil.Block
+	var blocksWithMTP []struct {
+		block *btcutil.Block
+		mtp   time.Time
+	}
+
+	// We need to activate CSV in order to test the processing logic, so
+	// manually craft the block version that's used to signal the soft-fork
+	// activation.
+	csvBit := netParams.Deployments[chaincfg.DeploymentCSV].BitNumber
+	blockVersion := int32(0x20000000 | (uint32(1) << csvBit))
+
+	// Generate enough blocks to activate CSV, collecting each of the
+	// blocks into a slice for later use.
+	numBlocksToActivate := (netParams.MinerConfirmationWindow * 3)
+	for i := uint32(0); i < numBlocksToActivate; i++ {
+		block, err := rpctest.CreateBlock(prevBlock, nil, blockVersion,
+			time.Time{}, miningAddr, netParams)
 		if err != nil {
-			t.Errorf("ProcessBlock fail on block %v: %v\n", i, err)
-			return
+			t.Fatalf("unable to generate block: %v", err)
+		}
+
+		mtp := chain.BestSnapshot().MedianTime
+
+		_, isOrphan, err := chain.ProcessBlock(block, blockchain.BFNone)
+		if err != nil {
+			t.Fatalf("ProcessBlock fail on block %v: %v\n", i, err)
 		}
 		if isOrphan {
-			t.Errorf("ProcessBlock incorrectly returned block %v "+
+			t.Fatalf("ProcessBlock incorrectly returned block %v "+
 				"is an orphan\n", i)
-			return
 		}
+
+		blocksWithMTP = append(blocksWithMTP, struct {
+			block *btcutil.Block
+			mtp   time.Time
+		}{
+			block: block,
+			mtp:   mtp,
+		})
+
+		prevBlock = block
 	}
 
 	// Create with all the utxos within the create created above.
 	utxoView := blockchain.NewUtxoViewpoint()
-	for blockHeight, block := range blocks {
-		for _, tx := range block.Transactions() {
+	for blockHeight, blockWithMTP := range blocksWithMTP {
+		for _, tx := range blockWithMTP.block.Transactions() {
 			utxoView.AddTxOuts(tx, int32(blockHeight))
 		}
 	}
-	utxoView.SetBestHash(blocks[len(blocks)-1].Hash())
+	utxoView.SetBestHash(blocksWithMTP[len(blocksWithMTP)-1].block.Hash())
 
-	// The median past time from the point of view of the second to last
-	// block in the chain.
-	medianTime := blocks[2].MsgBlock().Header.Timestamp.Unix()
-
-	// The median past time of the *next* block will be the timestamp of
-	// the 2nd block due to the way MTP is calculated in order to be
-	// compatible with Bitcoin Core.
-	nextMedianTime := blocks[2].MsgBlock().Header.Timestamp.Unix()
+	// The median time calculated from the PoV of the best block in our
+	// test chain. For unconfirmed inputs, this value will be used since th
+	// MTP will be calculated from the PoV of the yet-to-be-mined block.
+	nextMedianTime := int64(1401292712)
 
 	// We'll refer to this utxo within each input in the transactions
-	// created below. This block that includes this UTXO has a height of 4.
-	targetTx := blocks[4].Transactions()[0]
+	// created below. This utxo has an age of 4 blocks and was mined within
+	// block 297
+	targetTx := blocksWithMTP[len(blocksWithMTP)-4].block.Transactions()[0]
 	utxo := wire.OutPoint{
 		Hash:  *targetTx.Hash(),
 		Index: 0,
 	}
+
+	// Obtain the median time past from the PoV of the input created above.
+	// The MTP for the input is the MTP from the PoV of the block *prior*
+	// to the one that included it.
+	medianTime := blocksWithMTP[len(blocksWithMTP)-5].mtp.Unix()
 
 	// Add an additional transaction which will serve as our unconfirmed
 	// output.
@@ -187,6 +231,7 @@ func TestCalcSequenceLock(t *testing.T) {
 		Hash:  unConfTx.TxHash(),
 		Index: 0,
 	}
+
 	// Adding a utxo with a height of 0x7fffffff indicates that the output
 	// is currently unmined.
 	utxoView.AddTxOuts(btcutil.NewTx(unConfTx), 0x7fffffff)
@@ -283,24 +328,24 @@ func TestCalcSequenceLock(t *testing.T) {
 					Sequence:         blockchain.LockTimeToSequence(true, 2560),
 				}, {
 					PreviousOutPoint: utxo,
-					Sequence: blockchain.LockTimeToSequence(false, 3) |
+					Sequence: blockchain.LockTimeToSequence(false, 5) |
 						wire.SequenceLockTimeDisabled,
 				}, {
 					PreviousOutPoint: utxo,
-					Sequence:         blockchain.LockTimeToSequence(false, 3),
+					Sequence:         blockchain.LockTimeToSequence(false, 4),
 				}},
 			}),
 			view: utxoView,
 			want: &blockchain.SequenceLock{
 				Seconds:     medianTime + (5 << wire.SequenceLockTimeGranularity) - 1,
-				BlockHeight: 6,
+				BlockHeight: 299,
 			},
 		},
 		// Transaction has a single input spending the genesis block
 		// transaction. The input's sequence number is encodes a
 		// relative lock-time in blocks (3 blocks). The sequence lock
 		// should have a value of -1 for seconds, but a block height of
-		// 6 meaning it can be included at height 7.
+		// 298 meaning it can be included at height 299.
 		{
 			tx: btcutil.NewTx(&wire.MsgTx{
 				Version: 2,
@@ -312,7 +357,7 @@ func TestCalcSequenceLock(t *testing.T) {
 			view: utxoView,
 			want: &blockchain.SequenceLock{
 				Seconds:     -1,
-				BlockHeight: 6,
+				BlockHeight: 298,
 			},
 		},
 		// A transaction with two inputs with lock times expressed in
@@ -337,8 +382,9 @@ func TestCalcSequenceLock(t *testing.T) {
 		},
 		// A transaction with two inputs with lock times expressed in
 		// seconds. The selected sequence lock value for blocks should
-		// be the height further in the future, so a height of 10
-		// indicating in can be included at height 7.
+		// be the height further in the future. The converted absolute
+		// block height should be 302, meaning it can be included in
+		// block 303.
 		{
 			tx: btcutil.NewTx(&wire.MsgTx{
 				Version: 2,
@@ -353,7 +399,7 @@ func TestCalcSequenceLock(t *testing.T) {
 			view: utxoView,
 			want: &blockchain.SequenceLock{
 				Seconds:     -1,
-				BlockHeight: 10,
+				BlockHeight: 302,
 			},
 		},
 		// A transaction with multiple inputs. Two inputs are time
@@ -380,14 +426,15 @@ func TestCalcSequenceLock(t *testing.T) {
 			view: utxoView,
 			want: &blockchain.SequenceLock{
 				Seconds:     medianTime + (13 << wire.SequenceLockTimeGranularity) - 1,
-				BlockHeight: 12,
+				BlockHeight: 304,
 			},
 		},
 		// A transaction with a single unconfirmed input. As the input
 		// is confirmed, the height of the input should be interpreted
-		// as the height of the *next* block. So the relative block
-		// lock should be based from a height of 5 rather than a height
-		// of 4.
+		// as the height of the *next* block. The current block height
+		// is 300, so the lock time should be calculated using height
+		// 301 as a base. A 2 block relative lock means the transaction
+		// can be included after block 302, so in 303.
 		{
 			tx: btcutil.NewTx(&wire.MsgTx{
 				Version: 2,
@@ -399,7 +446,7 @@ func TestCalcSequenceLock(t *testing.T) {
 			view: utxoView,
 			want: &blockchain.SequenceLock{
 				Seconds:     -1,
-				BlockHeight: 6,
+				BlockHeight: 302,
 			},
 		},
 		// A transaction with a single unconfirmed input. The input has
--- a/blockchain/thresholdstate.go
+++ b/blockchain/thresholdstate.go
@@ -309,14 +309,33 @@ func (b *BlockChain) thresholdState(prevNode *blockNode, checker thresholdCondit
 //
 // This function is safe for concurrent access.
 func (b *BlockChain) ThresholdState(deploymentID uint32) (ThresholdState, error) {
+	b.chainLock.Lock()
+	state, err := b.deploymentState(b.bestNode, deploymentID)
+	b.chainLock.Unlock()
+
+	return state, err
+}
+
+// deploymentState returns the current rule change threshold for a given
+// deploymentID. The threshold is evaluated from the point of view of the block
+// node passed in as the first argument to this method.
+//
+// It is important to note that, as the variable name indicates, this function
+// expects the block node prior to the block for which the deployment state is
+// desired.  In other words, the returned deployment state is for the block
+// AFTER the passed node.
+//
+// This function MUST be called with the chain state lock held (for writes).
+func (b *BlockChain) deploymentState(prevNode *blockNode,
+	deploymentID uint32) (ThresholdState, error) {
+
 	if deploymentID > uint32(len(b.chainParams.Deployments)) {
 		return ThresholdFailed, DeploymentError(deploymentID)
 	}
+
 	deployment := &b.chainParams.Deployments[deploymentID]
 	checker := deploymentChecker{deployment: deployment, chain: b}
 	cache := &b.deploymentCaches[deploymentID]
-	b.chainLock.Lock()
-	state, err := b.thresholdState(b.bestNode, checker, cache)
-	b.chainLock.Unlock()
-	return state, err
+
+	return b.thresholdState(prevNode, checker, cache)
 }
--- a/blockchain/validate.go
+++ b/blockchain/validate.go
@@ -752,6 +752,27 @@ func (b *BlockChain) checkBlockContext(block *btcutil.Block, prevNode *blockNode
 
 	fastAdd := flags&BFFastAdd == BFFastAdd
 	if !fastAdd {
+		// Obtain the latest state of the deployed CSV soft-fork in
+		// order to properly guard the new validation behavior based on
+		// the current BIP 9 version bits state.
+		csvState, err := b.deploymentState(prevNode, chaincfg.DeploymentCSV)
+		if err != nil {
+			return err
+		}
+
+		// Once the CSV soft-fork is fully active, we'll switch to
+		// using the current median time past of the past block's
+		// timestamps for all lock-time based checks.
+		blockTime := header.Timestamp
+		if csvState == ThresholdActive {
+			medianTime, err := b.index.CalcPastMedianTime(prevNode)
+			if err != nil {
+				return err
+			}
+
+			blockTime = medianTime
+		}
+
 		// The height of this block is one more than the referenced
 		// previous block.
 		blockHeight := prevNode.height + 1
@@ -759,7 +780,7 @@ func (b *BlockChain) checkBlockContext(block *btcutil.Block, prevNode *blockNode
 		// Ensure all transactions in the block are finalized.
 		for _, tx := range block.Transactions() {
 			if !IsFinalizedTransaction(tx, blockHeight,
-				header.Timestamp) {
+				blockTime) {
 
 				str := fmt.Sprintf("block contains unfinalized "+
 					"transaction %v", tx.Hash())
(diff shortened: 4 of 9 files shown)


Roasbeef:
I plan to cherry-pick the integration tests I wrote for BIPs 112 and 113 which were extracted from the final version of the PR's in favor of reducing the scope of them by making them mempool only. I'm also working on an integration test for BIP 68 (CSV) will also be added to this PR once complete. 

@davecgh you mentioned earlier that you'd prefer the integration tests for soft-fork be segmented into a distinct package as they have stricter requirements on the pre-conditions of the test-chain. Once the new package is added to the BIP 9 PR, I'll move my integration tests to the new package.

davecgh:
This can be rebased as  #713, #628, #712 have all been merged.  Also, #755 is ready for master, but it is waiting on this one so they can be deployed at the same time.

blockchain/validate.go
```diff
@@ -774,14 +774,34 @@ func (b *BlockChain) checkBlockContext(block *btcutil.Block, prevNode *blockNode
 
 	fastAdd := flags&BFFastAdd == BFFastAdd
 	if !fastAdd {
+		medianTime := b.BestSnapshot().MedianTime
+
+		// Obtain the latest state of the deployed CSV soft-fork in
+		// order to properly guard the new validation behavior based on
+		// the current BIP 9 version bits state.
+		csvState, err := b.ThresholdState(chaincfg.DeploymentCSV)
```
davecgh: I believe this will end up dead locking because the chain state lock is already held and the exported `ThresholdState` takes the lock.  Given the usefulness of being able to check the state based on a deployment as the exported version does, I suspect we might need to refactor it slightly such that there is a `thresholdHoldForDeployment` or something similar.

blockchain/validate.go
```diff
@@ -774,14 +774,34 @@ func (b *BlockChain) checkBlockContext(block *btcutil.Block, prevNode *blockNode
...
+			return err
+		}
+
+		// Once the CSV soft-fork is fully active, we'll switch to
+		// using the current median time past of the past block's
+		// timestamps for all lock-time based checks.
+		var blockTime time.Time
+		if csvState == ThresholdActive {
+			blockTime = medianTime
+		} else {
```
davecgh: nit: Perhaps it's just a matter of taste, but I find the set and override more readable since it's shorter:

```Go
blockTime := header.Timestamp
if csvState == ThresholdActive {
	blockTime = medianTime
}
```

blockchain/validate.go
```diff
@@ -803,6 +823,40 @@ func (b *BlockChain) checkBlockContext(block *btcutil.Block, prevNode *blockNode
...
+		if csvState != ThresholdActive {
+			return nil
+		}
+
+		// Load all of the utxos referenced by the inputs for all
+		// transactions in the block don't already exist in the utxo
+		// view from the database. This view is required in order to
+		// properly evalulate the sequence locks (relative lock-time)
+		// for each input referenced within the block.
+		utxoView := NewUtxoViewpoint()
```
davecgh: Since we don't have a utxo cache this will end up double loading the utxos.  Also, this assumes the block is extended the main chain which may or may not be true.  We'll need to figure out a better place to do these checks and/or get the view earlier and pass it in to the function as well as down the call chain.

mining.go
```diff
@@ -446,6 +447,24 @@ func NewBlockTemplate(policy *mining.Policy, server *server, payToAddress btcuti
...
+	// lock-time, and CSV depends on the current soft-fork state.
+	csvState, err := blockManager.chain.ThresholdState(chaincfg.DeploymentCSV)
+	if err != nil {
+		return nil, err
+	}
+
+	// Once the CSV soft-fork is fully active, we'll switch to using the
+	// current median time past of the past block's timestamps for all
+	// lock-time based checks.
+	var blockTime time.Time
```
davecgh: Same comment here regarding setting and overriding.

Roasbeef: Fixed.

blockchain/validate.go
Roasbeef: Agreed. I've modified this fragment to eliminate the else cause by setting the `blockTime` value to `header.Timestamp` initially. As a result, the code is a bit shorter and easier to read as you noted.

blockchain/validate.go
Roasbeef: I've carried out your suggestion and have created a new internal version of  `ThresholdState` which assumes that the caller already has the `chainLock` acquired. This function is now used in this fragment in order to prevent the deadlock condition you pointed out.

blockchain/thresholdstate.go
```diff
@@ -0,0 +1,333 @@
...
+
+	return state, err
+}
+
+// thresholdHoldForDeployment returns the current rule change threshold for a
+// given deploymentID. The threshold is evaluated from the point of view of the
+// tip of the current best chain.
+//
+// This function MUST be called with the chain state lock held (for writes).
+func (b *BlockChain) thresholdHoldForDeployment(deploymentID uint32) (ThresholdState, error) {
```
davecgh: A couple of points here.  First one is a nit.  Everything else calls it `threshold` as one word, so this should probably be `thresholdForDeployment` (lowercase h).

Second point is the function should take the `*blockNode` to operate on:

```Go
func (b *BlockChain) thresholdholdForDeployment(node *blockNode, deploymentID uint32) (ThresholdState, error) {
	...
	return b.thresholdState(node, checker, cache)
}
```

blockchain/thresholdstate.go
```diff
@@ -0,0 +1,333 @@
...
+	return state, nil
+}
+
+// ThresholdState returns the current rule change threshold state of the given
+// deployment ID for the end of the current best chain.
+//
+// This function is safe for concurrent access.
+func (b *BlockChain) ThresholdState(deploymentID uint32) (ThresholdState, error) {
+	b.chainLock.Lock()
+	state, err := b.thresholdHoldForDeployment(deploymentID)
```
davecgh: With the below mentioned changes this would become:

```Go
b.thresholdForDeployment(b.bestNode, deploymentID)
```

blockchain/validate.go
```diff
@@ -1133,6 +1217,16 @@ func (b *BlockChain) checkConnectBlock(node *blockNode, block *btcutil.Block, vi
 		scriptFlags |= txscript.ScriptVerifyCheckLockTimeVerify
 	}
 
+	// Enforce CHECKSEQUENCEVERIFY during all block validation checks once
+	// the soft-fork deployment is fully active.
+	csvState, err := b.thresholdHoldForDeployment(chaincfg.DeploymentCSV)
```
davecgh: And this would need to be:

```Go
csvState, err := b.thresholdHoldForDeployment(prevNode, chaincfg.DeploymentCSV)
```

so it is from the point of view of the block being checked instead of the main chain.

blockchain/validate.go
davecgh: After looking at this again, this section isn't even needed.  The code later on in `checkConnectBlock` (with my other suggested corrections), will enable the script flag `txscript.ScriptVerifyCheckSequenceVerify` which instructs the script engine to enforce the locks.

blockchain/thresholdstate.go
```diff
@@ -0,0 +1,336 @@
...
+
+	return state, err
+}
+
+// thresholdForDeployment returns the current rule change threshold for a
+// given deploymentID. The threshold is evaluated from the point of view of the
+// blockNode passed in as the first argument to this method.
+//
+// This function MUST be called with the chain state lock held (for writes).
+func (b *BlockChain) thresholdForDeployment(blockNode *blockNode,
```
davecgh: I don't want to bikeshed too much since it's important we get this PR finished and landed, but I wonder if this might be better named `deploymentState` since I noticed that `thresholdForDeployment` is quite long and most of the callers end up using same variant of the name `state` in the variable anyways.  Definitely not something that has to be changed, but if you're in there making modifications anyways to fix the actual code issues pointed out regarding reorgs, and you think it's a good idea, go for it.

Roasbeef:
I've successfully fully synced both testnet and mainnet with this PR without any issues. The version bits state guarded consensus enforcements properly flipped over perfectly without any issues 🎉

davecgh:
Thanks for following up with test results.  I'm going to do some testing myself.  I did want to point out that the RPC server help for the new `getblockchaininfo` command is missing as reported by Travis:

```
rpcserverhelp_test.go:17: RPC handler defined for method 'getblockchaininfo' without also specifying result types
rpcserverhelp_test.go:44: Failed to generate help for method 'getblockchaininfo': no result types specified for method getblockchaininfo
```

blockchain/thresholdstate.go
```diff
@@ -0,0 +1,336 @@
...
+// consensus changes.
+type ThresholdState byte
+
+// These constants are used to identify specific threshold states.
+//
+// NOTE: This section specifically does not use iota for the individual states
+// since these values are serialized and must be stable for long-term storage.
+const (
+	// ThresholdDefined is the first state for each deployment and is the
+	// state for the genesis block has by defintion for all deployments.
```
stevenroose: `defintion`

definition*

blockchain/thresholdstate.go
```diff
@@ -0,0 +1,336 @@
...
+func (c *thresholdStateCache) Update(hash chainhash.Hash, state ThresholdState) {
+	if existing, ok := c.entries[hash]; ok && existing == state {
+		return
+	}
+
+	c.dbUpdates[hash] = state
+	c.entries[hash] = state
+}
+
+// MarkFlushed marks all of the current udpates as flushed to the database.
```
stevenroose: `udpates`

updates*

rpcserver.go
```diff
@@ -1120,6 +1120,101 @@ func handleGetBlock(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (i
...
+				Status bool `json:"status"`
+			}{
+				Status: height >= activeNetParams.BIP0065Height,
+			},
+		},
+	}
+
+	// Finally, query the BIP0009 verion bits state fo all currently
+	// defined BIP0009 soft-fork deployments.
+	for deployment, depolymentDetails := range activeNetParams.Deployments {
```
Roasbeef: Fixed.

blockchain/thresholdstate.go
```diff
@@ -0,0 +1,336 @@
...
+// This function is safe for concurrent access.
+func (b *BlockChain) ThresholdState(deploymentID uint32) (ThresholdState, error) {
+	b.chainLock.Lock()
+	state, err := b.deploymentState(b.bestNode, deploymentID)
+	b.chainLock.Unlock()
+
+	return state, err
+}
+
+// thresholdForDeployment returns the current rule change threshold for a
```
davecgh: s/thresholdForDeployment/deploymentState/

davecgh:
The following diff should take care of the test failure:

```diff
diff --git a/rpctest/rpc_harness.go b/rpctest/rpc_harness.go
index d09b472..dc3853e 100644
--- a/rpctest/rpc_harness.go
+++ b/rpctest/rpc_harness.go
@@ -273,7 +273,9 @@ func (h *Harness) TearDown() error {
                return err
        }

+       harnessStateMtx.Lock()
        delete(testInstances, h.testNodeDir)
+       harnessStateMtx.Unlock()

        return nil
 }
```

davecgh:
Looks like `TearDownAll` grabs the mutex and then calls `harness.TearDown` which grabs it again.  So, I believe there needs to be an unexported `func (h *Harness) tearDown() error {` which has to be called with the `harnessStateMtx` held that is called from `TearDownAll`.  Then `Teardown` would just become:

```Go
// TearDown stops the running rpc test instance. All created processes are
// killed, and temporary directories removed.
//
// NOTE: This method and SetUp should always be called from the same goroutine
// as they are not concurrent safe.
func (h *Harness) TearDown() error {
	harnessStateMtx.Lock()
	defer harnessStateMtx.Unlock()

	return h.teardown()
}
```

rpcserver.go
```diff
@@ -1120,6 +1120,101 @@ func handleGetBlock(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (i
...
+			continue
+		}
+
+		// Map the integer deployment ID into a human readable
+		// fork-name.
+		var forkName string
+		switch deployment {
+		case chaincfg.DeploymentCSV:
+			forkName = "csv"
+		}
```
davecgh: I think there should be a default case here which results in an `internalError` so tests will end up failing if a new deployment is added without also covering its fork name.

Alternatively, the name could be added to the `chainparams.ConsensusDeployment` struct.

rpcserver.go
```diff
@@ -1120,6 +1120,101 @@ func handleGetBlock(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (i
...
+
+		// Query the chain for the current status of the deployment as
+		// identified by its deployment ID.
+		deploymentStatus, err := s.chain.ThresholdState(uint32(deployment))
+		if err != nil {
+			return nil, err
```
davecgh: This should be converted to a `btcjson.RPCError` because this is in an RPC handler.  All error returns should be of that type.

Perhaps something like the following:

```Go
			context := "Failed to obtain deployment status"
			return nil, internalRPCError(err.Error(), context)
```

rpcserver.go
```diff
@@ -1120,6 +1120,101 @@ func handleGetBlock(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (i
...
+		}
+
+		// The string method for deployment status has a "Threshold"
+		// prefix for all the states, so we trim off this section in
+		// order to display it properly.
+		statusString := strings.TrimPrefix(deploymentStatus.String(),
```
davecgh: I'm a little bit concerned about this because it means changing the internal variable names would end up returning a different status string.

I suspect there should be a conversion function which returns the appropriate status string given a `ThresholdState` and uses a switch with a default case that returns an error.  That way any additional threshold states would be properly caught.

Then this would become something like:

```Go
statusString, err := softForkStatus(deploymentStatus)
if err != nil {
		context := "Failed to obtain soft fork status description"
		return nil, internalRPCError(err.Error(), context)
}
```

blockchain/validate.go
```diff
@@ -1126,6 +1156,46 @@ func (b *BlockChain) checkConnectBlock(node *blockNode, block *btcutil.Block, vi
...
+		scriptFlags |= txscript.ScriptVerifyCheckSequenceVerify
+
+		medianTime, err := b.calcPastMedianTime(prevNode)
+		if err != nil {
+			return err
+		}
+
+		// Additionally, if the CSV soft-fork package is now active,
+		// then we also enforce the relative sequence number based
+		// lock-times within the inputs of all transacitons in this
```
davecgh: s/transacitons/transactions/

davecgh:
Looks like this also relies on the addition of a `GetBlockChainInfo` command to `btcrpcclient`.  Do you have local changes for it that weren't PR'd to its repo?

```
# github.com/btcsuite/btcd/integration
integration\csv_fork_test.go:150: r.Node.GetBlockChainInfo undefined (type *btcrpcclient.Client has no field or method GetBlockChainInfo)
integration\csv_fork_test.go:200: r.Node.GetBlockChainInfo undefined (type *btcrpcclient.Client has no field or method GetBlockChainInfo)
integration\csv_fork_test.go:366: r.Node.GetBlockChainInfo undefined (type *btcrpcclient.Client has no field or method GetBlockChainInfo)
```

rpcserverhelp.go
```diff
@@ -156,6 +156,32 @@ var helpDescsEnUS = map[string]string{
...
+	"getblockchaininforesult-bip9_softforks":        "JSON object describing active BIP0009 deployments",
+	"getblockchaininforesult-bip9_softforks--key":   "bip9_softforks",
+	"getblockchaininforesult-bip9_softforks--value": "An object describing a particular BIP009 deployment",
+	"getblockchaininforesult-bip9_softforks--desc":  "The status of any defined BIP0009 soft-fork deployments",
+
+	// SoftForkDescription help.
+	"softforkdescription-reject":  "The current activation status of the softfork",
+	"softforkdescription-version": "The block version that signals enforcement of this softfork",
+	"softforkdescription-id":      "The string identifier for the soft fork",
+	"-status":                     "A bool which indicates if the soft fork is active",
```
Roasbeef: Fixed.

rpcserver.go
Roasbeef: Good point, I've opted to go with the first option.

rpcserver.go
Roasbeef: Nice catch, fixed.

rpcserver.go
Roasbeef: Good point, switching to a function like `softForkStatus` is more robust and will catch errors much sooner. I've modified this fragment to use such a function instead.

chaincfg/params.go
```diff
@@ -270,6 +275,11 @@ var MainNetParams = Params{
 			StartTime:  1199145601, // January 1, 2008 UTC
 			ExpireTime: 1230767999, // December 31, 2008 UTC
 		},
+		DeploymentCSV: {
+			BitNumber:  0,
+			StartTime:  1462060800, // March 1st, 2016
```
Roasbeef: Yep, I noticed that a bit ago and fixed it. Looks like this comment was made from an outdated diff.

blockchain/chain.go
```diff
@@ -766,20 +764,29 @@ func (b *BlockChain) calcSequenceLock(tx *btcutil.Tx, utxoView *UtxoViewpoint,
 	// activated.
 	sequenceLock := &SequenceLock{Seconds: -1, BlockHeight: -1}
 
+	// Obtain the latest BIP9 version bits state for the CSV-package
+	// soft-fork deployment. The adherence of sequence locks depends on the
+	// current soft-fork state.
+	csvState, err := b.deploymentState(node, chaincfg.DeploymentCSV)
```
Roasbeef: Nice, I like the suggestion. This'll also indirectly speed up mempool acceptance/processing since it doesn't need to _unconditionally_ query the the version bit state.

blockchain/validate.go
Roasbeef: Fixed.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,687 @@
...
+		return nil, nil, nil, err
+	}
+	output := &wire.TxOut{PkScript: selfAddrScript, Value: 1e8}
+
+	// Next, create and broadcast a transaction paying to the output.
+	fundTx, err := r.CreateTransaction([]*wire.TxOut{output}, 10)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	txid, err := r.Node.SendRawTransaction(fundTx, true)
```
davecgh: Please make all of these `txid` vars `txHash` instead.  We have been very careful to avoid mischaracterizing the hash as an ID in general.  Improperly treating the transaction hash as a unique ID is one of the main reason malleability led to so many issues in various software.  I do realize a big reason for segwit is stable hashes so they can be (ab)used as an ID, but I'd really prefer to avoid the notion altogether in the code since non-segwit transactions still exist and it applies there even less.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,687 @@
...
+		// input has a relative time-lock of 1 blocks, but the disable
+		// bit it set. The transaction should be rejected as a result.
+		{
+			tx: makeTxCase(
+				blockchain.LockTimeToSequence(false, 1)|wire.SequenceLockTimeDisabled,
+				2,
+			),
+			accept: false,
+		},
+		// A v2 transaction with a single input having a 10 lock
```
Roasbeef: Fixed.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,687 @@
...
+		// A valid transaction with a single input a sequence number
+		// creating a 100 block relative time-lock. This transaction
+		// should be rejected as its version number is 1, and only tx
+		// of version > 2 will trigger the CSV behavior.
+		{
+			tx:     makeTxCase(blockchain.LockTimeToSequence(false, 100), 1),
+			accept: false,
+		},
+		// A transaction of version 2 spending a single input. The
+		// input has a relative time-lock of 1 blocks, but the disable
```
Roasbeef: Fixed.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,687 @@
...
+	// age of 11. Space out each block 10 minutes after the previous block.
+	prevBlockHash, err := r.Node.GetBestBlockHash()
+	if err != nil {
+		t.Fatalf("unable to get prior block hash: %v", err)
+	}
+	prevBlock, err := r.Node.GetBlock(prevBlockHash)
+	if err != nil {
+		t.Fatalf("unable to get block: %v", err)
+	}
+	for i := 0; i < 10; i++ {
```
Roasbeef: Fixed.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,687 @@
...
+	// three outputs which use the above time based sequence lock.
+	for i := 0; i < numTests; i++ {
+		var timeLock int32
+		var isSeconds bool
+
+		if i < 3 {
+			timeLock = relativeBlockLock
+			isSeconds = false
+		} else {
+			timeLock = relativeTimeLock
```
Roasbeef: Great suggestion, altered to go with the 6 line version.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,687 @@
...
+
+	// This transaction should be rejected from the mempool since CSV
+	// validation is already mempool policy pre-fork.
+	_, err = r.Node.SendRawTransaction(spendingTx, true)
+	if err == nil {
+		t.Fatalf("transaction should have been rejected, but was " +
+			"instead accepted")
+	}
+
+	// However, this transaction should be accepted in a custom generate
```
Roasbeef: Fixed.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,687 @@
...
+	// Create a trivially spendable output with a CSV lock-time of 10
+	// relative blocks.
+	redeemScript, testUTXO, tx, err := createCSVOutput(r, t, outputAmt, 10,
+		false)
+	if err != nil {
+		t.Fatalf("unable to create CSV encumbered output: %v", err)
+	}
+
+	// As the transaction is p2sh it should be accepted into the mempool
+	// and fond within the next generated block.
```
Roasbeef: Fixed.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,687 @@
...
+	harnessScript, err := txscript.PayToAddrScript(harnessAddr)
+	if err != nil {
+		t.Fatalf("unable to generate pkScript: %v", err)
+	}
+
+	const outputAmt = btcutil.SatoshiPerBitcoin
+
+	// Create a trivially spendable output with a CSV lock-time of 10
+	// relative blocks.
+	redeemScript, testUTXO, tx, err := createCSVOutput(r, t, outputAmt, 10,
```
Roasbeef: Fixed.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,687 @@
...
+//    - See the cases exercised within the table driven tests towards the end
+//    of this test.
+func TestBIP0068AndBIP0112Activation(t *testing.T) {
+	t.Parallel()
+
+	// We'd like the test proper evaluation and validation of the BIP 68
+	// (sequence locks) and BIP 112 rule-sets which add input-age based
+	// relative lock times.
+
+	// Initialize the primary mining node with only the genesis block.
```
Roasbeef: Fixed.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,687 @@
...
+	}
+
+	for _, txn := range block.Transactions {
+		txHash := txn.TxHash()
+		if (&txHash).IsEqual(txid) {
+			return
+		}
+	}
+
+	t.Fatalf("txid %v was not found in block %v", txid, blockHash)
```
Roasbeef: Fixed, also added similar caller/line context to the `isSoftForkActive` function.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,687 @@
...
+	if err != nil {
+		t.Fatalf("unable to get block: %v", err)
+	}
+	if len(block.Transactions) < 2 {
+		t.Fatal("target transaction was not mined")
+	}
+
+	for _, txn := range block.Transactions {
+		txHash := txn.TxHash()
+		if (&txHash).IsEqual(txid) {
```
Roasbeef: Fixed.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,687 @@
...
+		PreviousOutPoint: *csvUTXO,
+		Sequence:         sequence,
+	})
+	tx.AddTxOut(targetOutput)
+
+	b := txscript.NewScriptBuilder()
+	b.AddOp(txscript.OP_TRUE)
+	b.AddData(redeemScript)
+
+	sigScript, err := b.Script()
```
Roasbeef: Fixed.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,687 @@
...
+	// Finally create a valid transactions which creates the output crafted
+	// above.
+	tx, err := r.CreateTransaction([]*wire.TxOut{output}, 10)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	var outputIndex uint32
+	if bytes.Equal(tx.TxOut[0].PkScript, p2shScript) {
+		outputIndex = 0
```
Roasbeef: Nice catch, fixed.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,687 @@
...
+	p2shScript, err := txscript.PayToAddrScript(p2shAddr)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	output := &wire.TxOut{
+		PkScript: p2shScript,
+		Value:    int64(numSatoshis),
+	}
+
+	// Finally create a valid transactions which creates the output crafted
```
Roasbeef: Fixed.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,687 @@
...
+	numSatoshis btcutil.Amount, timeLock int32,
+	isSeconds bool) ([]byte, *wire.OutPoint, *wire.MsgTx, error) {
+
+	// Convert the time-lock to the proper sequence lock based according to
+	// if the lock is seconds or time based.
+	sequenceLock := blockchain.LockTimeToSequence(isSeconds,
+		uint32(timeLock))
+
+	// Our CSV script is simply: <sequenceLock> OP_CSV OP_DROP
+	b := txscript.NewScriptBuilder()
```
Roasbeef: Fixed.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,687 @@
...
+//
+//  - Post soft-fork:
+//    - Transactions with non-final lock-times from the PoV of MTP should be
+//    rejected from the mempool and when found within otherwise valid blocks.
+//    - Transactions with final lock-times from the PoV of MTP should be
+//    accepted to the mempool and mined in future block.
+func TestBIP0113Activation(t *testing.T) {
+	t.Parallel()
+
+	// Initialize the primary mining node with only the genesis block.
```
Roasbeef: Fixed.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,687 @@
...
+
+// TestBIP0113Activation tests for proper adherance of the BIP 113 rule
+// constraint which requires all transaction finality tests to use the MTP of
+// the last 11 blocks, rather than the timestamp of the block which includes
+// them.
+//
+// Overview:
+//  - Pre soft-fork:
+//    - Transactions with non-final lock-times from the PoV of MTP should be
+//    rejected from the mempool.
```
Roasbeef: Fixed.

integration/csv_fork_test.go
Roasbeef: Good point, fixed to instead read `txHash` in the latest diff.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,687 @@
...
+	if err != nil {
+		t.Fatalf("unable to submit block: %v", err)
+	}
+	txid := tx.TxHash()
+	assertTxInBlock(r, t, block.Hash(), &txid)
+
+	// Next, mine enough blocks to ensure that the soft-fork becomes
+	// activated. Assert that the block version of the second-to-last block
+	// in the final range is active.
+	numBlocks := (r.ActiveNet.MinerConfirmationWindow * 2) + 1
```
Roasbeef: Hmm, yes I think you're correct, CSV should be active at 300, but _not_ at height 299. 

As we discussed offline, I'll separate out the `getblockchaininfo` portion of this PR into a new one, so we can use it to write some additional integration tests for bip9 so we can drill down into the issue an locate where the off-by-one error is.

davecgh:
Running this PR with the new tests in #873, and adding a new entry for CSV as follows shows there is definitely an off by one in this PR:

```diff
diff --git a/integration/bip0009_test.go b/integration/bip0009_test.go
index b841ebf..8aa6f57 100644
--- a/integration/bip0009_test.go
+++ b/integration/bip0009_test.go
@@ -294,6 +294,7 @@ func TestBIP0009(t *testing.T) {
        t.Parallel()

        testBIP0009(t, "dummy")
+       testBIP0009(t, "csv")
 }

 // TestBIP0009Mining ensures blocks built via btcd's CPU miner follow the rules
```

```
$ go test -tags=rpctest -run=TestBIP0009 -v
=== RUN   TestBIP0009
=== RUN   TestBIP0009Mining
--- PASS: TestBIP0009Mining (1.22s)
--- FAIL: TestBIP0009 (3.20s)
        bip0009_test.go:117: assertion failed at line 235: softfork status for "csv" is started instead of expected lockedin
FAIL
exit status 1
FAIL    github.com/btcsuite/btcd/integration    4.477s
```

blockchain/thresholdstate.go
```diff
@@ -309,14 +309,28 @@ func (b *BlockChain) thresholdState(prevNode *blockNode, checker thresholdCondit
...
+
+	return state, err
+}
+
+// depolymentState returns the current rule change threshold for a given
+// deploymentID. The threshold is evaluated from the point of view of the
+// blockNode passed in as the first argument to this method.
+//
+// This function MUST be called with the chain state lock held (for writes).
+func (b *BlockChain) deploymentState(blockNode *blockNode,
```
davecgh: The argument here really should be named `prevNode` because `thresholdState` accepts a prevNode.  That is to say it returns the threshold state for the block _after_ the provided block node (hence the name `prevNode`).

I would also suggest the following diff to make the comments call this out more clearly.

```diff
 // deploymentState returns the current rule change threshold for a given
 // deploymentID. The threshold is evaluated from the point of view of the
-// blockNode passed in as the first argument to this method.
+// block node passed in as the first argument to this method.
+//
+// It is important to note that, as the variable name indicates, this function
+// expects the block node prior to the block for which the deployment state is
+// desired.  In other words, the returned deployment state is for the block
+// AFTER the passed node.
 //
 // This function MUST be called with the chain state lock held (for writes).
-func (b *BlockChain) deploymentState(blockNode *blockNode,
-       deploymentID uint32) (ThresholdState, error) {
-
+func (b *BlockChain) deploymentState(prevNode *blockNode, deploymentID uint32) (ThresholdState, error) {
```

blockchain/validate.go
```diff
@@ -752,14 +752,35 @@ func (b *BlockChain) checkBlockContext(block *btcutil.Block, prevNode *blockNode
...
+		if err != nil {
+			return err
+		}
+
+		// Once the CSV soft-fork is fully active, we'll switch to
+		// using the current median time past of the past block's
+		// timestamps for all lock-time based checks.
+		blockTime := header.Timestamp
+		if csvState == ThresholdActive {
+			medianTime, err := b.calcPastMedianTime(prevNode)
```
davecgh: I believe this is incorrect.  The `calcPastMedianTime` takes the block node you want to the calulate the median time for as opposed to its parent.  However, I don't think it's available in this function at the current time.  It'll probably need to be modified to pass in the current block node and get the previous one for the places that need it such as `checkBlockHeaderContext` and `deploymentState`.

This also likely means the block node creation for the block being validated will need to move up before the call to `checkBlockContext` in `accept.go`.

blockchain/validate.go
```diff
@@ -1126,6 +1147,46 @@ func (b *BlockChain) checkConnectBlock(node *blockNode, block *btcutil.Block, vi
 		scriptFlags |= txscript.ScriptVerifyCheckLockTimeVerify
 	}
 
+	// Enforce CHECKSEQUENCEVERIFY during all block validation checks once
+	// the soft-fork deployment is fully active.
+	csvState, err := b.deploymentState(node, chaincfg.DeploymentCSV)
```
davecgh: I believe this should be the previous node since `deploymentState` calculates the state for the block _after_ the passed node.

blockchain/chain.go
```diff
@@ -766,20 +764,37 @@ func (b *BlockChain) calcSequenceLock(tx *btcutil.Tx, utxoView *UtxoViewpoint,
...
+	// within the mempool.
+	csvSoftforkActive := mempool
+
+	// If we're performing block validation, then we need to query the BIP9
+	// state.
+	if !csvSoftforkActive {
+		// Obtain the latest BIP9 version bits state for the
+		// CSV-package soft-fork deployment. The adherence of sequence
+		// locks depends on the current soft-fork state.
+		csvState, err := b.deploymentState(node, chaincfg.DeploymentCSV)
```
davecgh: I believe this should be the previous node since `deploymentState` calculates the state for the block _after_ the passed node.

We really need some tests that check for off-by-one conditions in the sequence locks.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,719 @@
...
+
+// isSoftForkActive returns true if the soft-fork as defined by the passed
+// string is currently active.
+func isSoftForkActive(r *rpctest.Harness, t *testing.T, bip9ForkName string) bool {
+	chainInfo, err := r.Node.GetBlockChainInfo()
+	if err != nil {
+		t.Fatalf("unable to query for chain info: %v", err)
+	}
+
+	fmt.Println("height: ", chainInfo.Blocks)
```
davecgh: Leftover debug.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,719 @@
...
+
+	csvForkState, ok := chainInfo.Bip9SoftForks[bip9ForkName]
+	fmt.Println("status: ", csvForkState)
```
davecgh: Leftover debug.

blockchain/thresholdstate.go
Roasbeef: Indeed, makes sense to ensure that all the documentation around these functions is sufficiently explicit in order to avoid future off-by-one errors like the ones we discovered in earlier PR's and this one itself.

integration/csv_fork_test.go
Roasbeef: Fixed.

blockchain/chain.go
Roasbeef: Fixed.

blockchain/validate.go
Roasbeef: Fixed.

blockchain/validate.go
Roasbeef: As I noted on IRC, I believe that this is actually _correct_. The MTP used for a block to decide transaction finality is the median of the _past_ 11 blocks. So this would mean using `prevNode` everywhere during transaction finality tests as a part of block validation. 

This meshes with the mempool behavior at it uses the current MTP (of the chain tip) to decide if the transaction can be accepted into the mempool based on its eligibility for the _next_ block. The latest set of commits includes a boundary condition tests for the adherence of MTP.

integration/csv_fork_test.go
```diff
@@ -457,88 +457,103 @@ func TestBIP0068AndBIP0112Activation(t *testing.T) {
...
-	// transation.
+	// At this point, the block height should be 107: we started at height
+	// 101, then generated 2 blocks in each loop iteration above.
 	_, height, err := r.Node.GetBestBlock()
 	if err != nil {
 		t.Fatalf("unable to get best block: %v", err)
 	}
-	if height != 103 {
-		t.Fatalf("height mismatch: expected %v got %v", 102, height)
+	if height != 107 {
```
Roasbeef: Modified the tests to use the newly available testing helper funcs.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,742 @@
...
+	r, err := rpctest.New(&chaincfg.SimNetParams, nil, btcdCfg)
+	if err != nil {
+		t.Fatal("unable to create primary harness: ", err)
+	}
+	if err := r.SetUp(true, 1); err != nil {
+		t.Fatalf("unable to setup test chain: %v", err)
+	}
+	defer r.TearDown()
+
+	if isSoftForkActive(r, t, "csv") {
```
Roasbeef: Modified the tests to use `assertSoftForkStatus` when applicable.

integration/csv_fork_test.go
```diff
@@ -0,0 +1,742 @@
...
+		assertTxInBlock(r, t, block.Hash(), &txid)
+	}
+
+	// At this point, the block height should be 107: we started at height
+	// 101, then generated 2 blocks in each loop iteration above.
+	_, height, err := r.Node.GetBestBlock()
+	if err != nil {
+		t.Fatalf("unable to get best block: %v", err)
+	}
+	if height != 107 {
```
Roasbeef: Fixed.

blockchain/chain_test.go
```diff
@@ -116,15 +120,10 @@ func TestHaveBlock(t *testing.T) {
...
-	blocks, err := loadBlocks(fileName)
-	if err != nil {
-		t.Errorf("Error loading file: %v\n", err)
-		return
-	}
+	netParams := &chaincfg.SimNetParams
 
 	// Create a new database and chain instance to run tests against.
-	chain, teardownFunc, err := chainSetup("haveblock", &chaincfg.MainNetParams)
+	chain, teardownFunc, err := chainSetup("haveblock", netParams)
```
Roasbeef: Haven't seen any signs of negative interactions yet myself, but changing to `calcseqlock` just to be safe :)

blockchain/chain_test.go
```diff
@@ -136,46 +135,91 @@ func TestCalcSequenceLock(t *testing.T) {
...
+			mtp   time.Time
+		}{
+			block: block,
+			mtp:   mtp,
+		})
+
+		prevBlock = block
 	}
 
 	// Create with all the utxos within the create created above.
```
Roasbeef: Fixed.

blockchain/chain.go
```diff
@@ -766,20 +764,37 @@ func (b *BlockChain) calcSequenceLock(tx *btcutil.Tx, utxoView *UtxoViewpoint,
...
+	// within the mempool.
+	csvSoftforkActive := mempool
+
+	// If we're performing block validation, then we need to query the BIP9
+	// state.
+	if !csvSoftforkActive {
+		// Obtain the latest BIP9 version bits state for the
+		// CSV-package soft-fork deployment. The adherence of sequence
+		// locks depends on the current soft-fork state.
+		csvState, err := b.deploymentState(node.parent, chaincfg.DeploymentCSV)
```
Roasbeef: Nice catch, fixed to use `getPrevFromNode` to ensure the parent node is accessed/loaded safely.

rpctest/rpc_harness.go
```diff
@@ -261,6 +261,17 @@ func (h *Harness) SetUp(createTestChain bool, numMatureOutputs uint32) error {
...
+	defer harnessStateMtx.Unlock()
+
+	return h.tearDown()
+}
+
+// tearDown stops the running rpc test instance. See the publicly exported
+// TearDown method for more details.
+//
+// NOTE: This MUST be called with the harnessStateMtx held.
+func (h *Harness) tearDown() error {
```
Roasbeef: No worries, I removed my commit fixing the race condition as a part of my latest rebase.

Roasbeef:
Rebased over the recent changes to the indexing within the `blockchain` package.

afk11:
Hey all, me again. CSV has been active on mainnet for over a year now, what would be the holdup on this PR?

davecgh:
Other than that final comment typo and squashing the review commits down into the respective commits, this is ready to go.  I've extensively tested it and reviewed it multiple times.  Fantastic work!

Roasbeef:
Thanks @davecgh!

I've squashed down the review commits. Very excited to see this get in!

PR #778, opened 2016-10-13
blockchain: Add block validation infrastructure.

davecgh:
**This PR requires #777**

This adds a full-blown testing infrastructure in order to test consensus validation rules.  It is built around the idea of dynamically generating full blocks that target specific rules linked together to form a block chain.  In order to properly test the rules, each test instance starts with a valid block that is then modified in the specific way needed to test a specific rule.

Blocks which exercise the following rules have been added for this initial version.  These tests were largely ported from the original Java-based 'official' block acceptance tests as well as some additional tests available in the Core python port.  It is expected that further tests can be added over time as consensus rules change.

* Enough valid blocks to have a stable base of mature coinbases to spend for futher tests
* Basic forking and chain reorganization
* Double spends on forks
* Too much proof-of-work coinbase (extending main chain, in block that forces a reorg, and in a valid fork)
* Max and too many signature operations via various combinations of OP_CHECKSIG, OP_MULTISIG, OP_CHECKSIGVERIFY, and OP_MULTISIGVERIFY
* Too many and max signature operations with offending sigop after invalid data push
* Max and too many signature operations via pay-to-script-hash redeem scripts
* Attempt to spend tx created on a different fork
* Attempt to spend immature coinbase (on main chain and fork)
* Max size block and block that exceeds the max size
* Children of rejected blocks are either orphans or rejected
* Coinbase script too small and too large
* Max length coinbase script
* Attempt to spend tx in blocks that failed to connect
* Valid non-coinbase tx in place of coinbase
* Block with no transactions
* Invalid proof-of-work
* Block with a timestamp too far in the future
* Invalid merkle root
* Invalid proof-of-work limit (bits header field)
* Negative proof-of-work limit (bits header field)
* Two coinbase transactions
* Duplicate transactions
* Spend from transaction that does not exist
* Timestamp exactly at and one second after the median time
* Blocks with same hash via merkle root tricks
* Spend from transaction index that is out of range
* Transaction that spends more that its inputs provide
* Transaction with same hash as an existing tx that has not been fully spent (BIP0030)
* Non-final coinbase and non-coinbase txns
* Max size block with canonical encoding which exceeds max size with non-canonical encoding
* Spend from transaction earlier in same block
* Spend from transaction later in same block
* Double spend transaction from earlier in same block
* Coinbase that pays more than subsidy + fees
* Coinbase that includes subsidy + fees
* Invalid opcode in dead execution path
* Reorganization of txns with OP_RETURN outputs
* Spend of an OP_RETURN output
* Transaction with multiple OP_RETURN outputs
* Large max-sized block reorganization test (disabled by default since it takes a long time and a lot of memory to run)

Finally, the `README.md` files in the main and `docs` directories have been updated to reflect the use of the new testing framework.


stevenroose:
epic

davecgh:
Added another test:

* Negative proof-of-work limit (bits header field)

davecgh:
Updated the `README.md` in the main and `docs` directories to reflect the use of the new testing framework.  The commit message and original PR description have been updated accordingly.

PR #782, opened 2016-10-19
btcec: Consolidate tests into the btcec package.

davecgh:
Putting the test code in the same package makes it easier for forks since they don't have to change the import paths as much and it also gets rid of the need for `internal_test.go` to bridge.

Also, remove the exception from the lint checks about returning the unexported type since it is no longer required.


jcvernaleo:
I haven't review the code (just a glance) but I'm a very big fan of this idea (from the fork perspective).

jcvernaleo:
Now I really did look at the code.
OK

PR #784, opened 2016-10-19
btcjson: Add preciousblock

dajohi:


dajohi:
See https://github.com/bitcoin/bitcoin/commit/5127c4f21ca3f092b3a853833631dfc7f9c9456b for reference.

PR #785, opened 2016-10-19
blockchain: Remove isMajorityVersion code.

davecgh:
**NOTE: This depends on #792**

Now that all softforking is done via BIP0009 versionbits, replace the old `isMajorityVersion` deployment mechanism with hard coded historical block heights at which they became active.

Since the activation heights vary per network, this adds new parameters to the `chaincfg.Params` struct for them and sets the correct heights at which each softfork became active on each chain.

It should be noted that this is a technically hard fork since the behavior of alternate chain history is different with these hard-coded activation heights as opposed to the old `isMajorityVersion` code.  In
particular, an alternate chain history could activate one of the soft forks earlier than these hard-coded heights which means the old code would reject blocks which violate the new soft fork rules whereas this
new code would not.

However, all of the soft forks this refers to were activated so far in the chain history there is there is no way a reorg that long could happen and checkpoints reject alternate chains before the most recent checkpoint anyways.  Furthermore, the same change was made in Bitcoin Core so this needs to be changed to be consistent anyways.

### DIFF
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -654,35 +654,6 @@ func (b *BlockChain) removeBlockNode(node *blockNode) error {
 	return nil
 }
 
-// isMajorityVersion determines if a previous number of blocks in the chain
-// starting with startNode are at least the minimum passed version.
-//
-// This function MUST be called with the chain state lock held (for writes).
-func (b *BlockChain) isMajorityVersion(minVer int32, startNode *blockNode, numRequired uint64) bool {
-	numFound := uint64(0)
-	iterNode := startNode
-	for i := uint64(0); i < b.chainParams.BlockUpgradeNumToCheck &&
-		numFound < numRequired && iterNode != nil; i++ {
-		// This node has a version that is at least the minimum version.
-		if iterNode.version >= minVer {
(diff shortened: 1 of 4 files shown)


davecgh:
Rebased on top of latest #755.

Roasbeef:
So as you may know there's recently been a bit of controversy strewn up the bitcoin-dev mailing list related to this change. My point-of-view regarding the manner is similar to yours: while this change is technically/theoretically a hard-fork, any re-orgs that far back would cause havoc within the ecosystem, and the checkpoint in place already disallow such deep re-orgs. Additionally, by moving forward with this change we're able to remove a good bit of dead-code as seen from the diff caused by the latest commit (+38/-143). 

Therefore, I have no problem going forward with this change. 

The changes read well. I'm performing a mainnet and testnet sync with the total diff applied. I'll approve the changes once the syncs finish without any issues.

davecgh:
Thanks for the review.  Aye, I've seen the controversy on the mailing list and I certainly understand the complaints, but aside from the checkpoints preventing it to being with, if there were ever a reorg that large, there are much bigger problems!

PR #786, opened 2016-10-19
blockchain: Only enforce BIP0030 prior to BIP0034.

davecgh:
**This PR depends on #785**

This modifies the code to only enforce the fairly expensive BIP0030 (duplicate transcactions) checks when the chain has not yet reached the BIP0034 activation height (and is not one of the 2 special historical blocks that break the rule and prompted BIP0034 to being with) since that BIP made it impossible to create duplicate coinbases and thus removed the possibility of creating transactions that overwrite older ones.

This is a rather large optimization because the check is expensive due to involving a ton of cache misses in the utxoset.  For example, the following are times it took to perform the BIP0030 check on blocks 425490 - 425502 on a system with a relatively old Hitachi spinner HDD:

block 425490: 674.5857ms
block 425491: 726.5923ms
block 425492: 827.6051ms
block 425493: 680.0863ms
block 425494: 722.0917ms
block 425495: 700.0889ms
block 425496: 647.5823ms
block 425497: 445.0565ms
block 425498: 602.5765ms
block 425499: 375.0476ms
block 425500: 771.0979ms
block 425501: 461.5586ms
block 425502: 603.0766ms

As can be seen from these numbers, this reduces the block validation time by an average of just over half a second for the given representative data set and hardware.

### DIFF
--- a/blockchain/validate.go
+++ b/blockchain/validate.go
@@ -979,11 +979,18 @@ func (b *BlockChain) checkConnectBlock(node *blockNode, block *btcutil.Block, vi
 	// spent.  See the documentation for checkBIP0030 for more details.
 	//
 	// There are two blocks in the chain which violate this rule, so the
-	// check must be skipped for those blocks. The isBIP0030Node function is
-	// used to determine if this block is one of the two blocks that must be
-	// skipped.
-	enforceBIP0030 := !isBIP0030Node(node)
-	if enforceBIP0030 {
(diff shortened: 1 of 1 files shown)


davecgh:
Rebased on top of latest #785.

davecgh:
Rebased to lasest master.

davecgh:
@Roasbeef: Did your sync completed successfully?

Issue #709, opened 2016-06-10
btcec: recovers certain signature without error


fjl:
I am trying to use package btcec in go-ethereum for the sake of having a pure-go build.
During testing, I discovered that btcec.RecoverCompact accepts certain signatures which
are rejected by both libsecp256k1 and the ECDSA_SIG_recover_key_GFp function previously
used in bitcoind (before they switched to libsecp256k1).

The signature and message in question are:

``` text
msg = 00c547e4f7b0f325ad1e56f57e26c745b09a3e503d86e00e5255ff7f715d3d1c
sig = 1c00b1693892219d736caba55bdb67216e485557ea6b6af75f37096c9aa6a5a75f00b940b1d03b21e36b0e47e79769f095fe2ab855bd91e3a38756b7d75a9c4549
```

See https://github.com/fjl/btcec-issue for code that reproduces the issue.
Output from `make test` in that repo (the signature above is sig 2):

``` text
== sig 1 libsecp256k1:
pubkey: 04E32DF42865E97135ACFB65F3BAE71BDC86F4D49150AD6A440B6F15878109880A0A2B2667F7E725CEEA70C673093BF67663E0312623C8E091B13CF2C0F11EF652
== sig 2 libsecp256k1:
recover failed
== sig 1 openssl:
pubkey: 04E32DF42865E97135ACFB65F3BAE71BDC86F4D49150AD6A440B6F15878109880A0A2B2667F7E725CEEA70C673093BF67663E0312623C8E091B13CF2C0F11EF652
== sig 2 openssl:
recover failed
== sig 1 btcec:
pubkey: 04E32DF42865E97135ACFB65F3BAE71BDC86F4D49150AD6A440B6F15878109880A0A2B2667F7E725CEEA70C673093BF67663E0312623C8E091B13CF2C0F11EF652
== sig 2 btcec:
pubkey: 04D808F6C95B99D15DC39CC452D5B1A37633455A3CC0B96B191AA1873A6174B00B4A58079675094B2459AA57E09505BF9A22DD6EB988FBB3A471B1F13D8D74CFF7
```

It might also be that I am using btcec.RecoverCompact incorrectly.

fjl:
I have debugged this a little bit and the issue seems to be in the decompressPoint function.

fjl:
The following diff makes recovery work for the signature above, but disables the sqrt optimization in decompressPoint. Not sure whether that's a good fix.

```diff
[...]

PR #789, opened 2016-10-19
txscript: Add null data script creator

DanielKrawisz:
NullDataScript takes a byte array and constructs a null data output script. This is used for constructing transactions that contain extra data in provably prunable outputs. 

There is also a bug fix in isNullData.

### DIFF
--- a/txscript/standard.go
+++ b/txscript/standard.go
@@ -358,12 +358,7 @@ func NullDataScript(data []byte) ([]byte, error) {
 		return nil, ErrStackLongScript
 	}
 
-	script := make([]byte, 3+len(data))
-	script[0] = OP_RETURN
-	script[1] = OP_PUSHDATA1
-	script[2] = byte(len(data))
-	copy(script[3:3+len(data)], data)
-	return script, nil
+	return NewScriptBuilder().AddOp(OP_RETURN).AddData(data).Script()
 }
 
 // MultiSigScript returns a valid script for a multisignature redemption where
--- a/txscript/standard.go
+++ b/txscript/standard.go
@@ -351,6 +351,24 @@ func PayToAddrScript(addr btcutil.Address) ([]byte, error) {
 	return nil, ErrUnsupportedAddress
 }
 
+// NullDataScript creates a provably prunable script
+// containing OP_RETURN followed by the passed data.
+func NullDataScript(data []byte) ([]byte, error) {
+	if len(data) > MaxDataCarrierSize {
+		return nil, ErrStackLongScript
+	}
+
+	return unparseScript([]parsedOpcode{
+		parsedOpcode{
+			opcode: &opcodeArray[OP_RETURN],
+		},
+		parsedOpcode{
+			opcode: &opcodeArray[OP_PUSHDATA1],
+			data:   data,
+		},
+	})
+}
+
 // MultiSigScript returns a valid script for a multisignature redemption where
 // nrequired of the keys in pubkeys are required to have signed the transaction
 // for success.  An ErrBadNumRequired will be returned if nrequired is larger
--- a/txscript/standard.go
+++ b/txscript/standard.go
@@ -358,15 +358,12 @@ func NullDataScript(data []byte) ([]byte, error) {
 		return nil, ErrStackLongScript
 	}
 
-	return unparseScript([]parsedOpcode{
-		parsedOpcode{
-			opcode: &opcodeArray[OP_RETURN],
-		},
-		parsedOpcode{
-			opcode: &opcodeArray[OP_PUSHDATA1],
-			data:   data,
-		},
-	})
+	script := make([]byte, 3+len(data))
+	script[0] = OP_RETURN
+	script[1] = OP_PUSHDATA1
+	script[2] = byte(len(data))
+	copy(script[3:3+len(data)], data)
+	return script, nil
 }
 
 // MultiSigScript returns a valid script for a multisignature redemption where
--- a/txscript/standard_test.go
+++ b/txscript/standard_test.go
@@ -1028,3 +1028,28 @@ func TestStringifyClass(t *testing.T) {
 		}
 	}
 }
+
+// TestNullDataScript tests whether NullDataScript returns a valid script.
+func TestNullDataScript(t *testing.T) {
+	data := []byte{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16,
+		17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31, 32, 33,
+		34, 35, 36, 37, 38, 39, 40, 41, 42, 43, 44, 45, 46, 47, 48, 49, 50,
+		51, 52, 53, 54, 55, 56, 57, 58, 59, 60, 61, 62, 63, 64, 65, 66, 67,
+		68, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, 79}
+
+	tooLong := append(data, byte(80))
+
+	script, err := txscript.NullDataScript(data)
+	if err != nil {
+		t.Error("Error returned creating script: ", err)
+	}
+	scriptType := txscript.GetScriptClass(script)
+	if scriptType != txscript.NullDataTy {
+		t.Error("Expected a NullDataTy, got ", scriptType)
+	}
+
+	_, tooLongErr := txscript.NullDataScript(tooLong)
+	if tooLongErr == nil || tooLongErr != txscript.ErrStackLongScript {
+		t.Error("Expected ErrStackLongScript, got ", tooLongErr)
+	}
+}
--- a/txscript/standard_test.go
+++ b/txscript/standard_test.go
@@ -1031,25 +1031,86 @@ func TestStringifyClass(t *testing.T) {
 
 // TestNullDataScript tests whether NullDataScript returns a valid script.
 func TestNullDataScript(t *testing.T) {
-	data := []byte{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16,
-		17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31, 32, 33,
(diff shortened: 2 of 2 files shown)


txscript/standard.go
```diff
@@ -350,6 +350,21 @@ func PayToAddrScript(addr btcutil.Address) ([]byte, error) {
...
 }
 
+// NullDataScript creates a provably prunable script
+// containing OP_RETURN followed by the passed data.
+func NullDataScript(data []byte) ([]byte, error) {
+	if len(data) > MaxDataCarrierSize {
+		return nil, ErrStackLongScript
+	}
+
+	script := make([]byte, 3+len(data))
```
davecgh: This will produce OP_RETURN scripts with non-canonical data pushes that would end up being rejected by the majority of the network.

Please use this instead:

````Go
	builder := txscript.NewScriptBuilder()
	return builder.AddOp(txscript.OP_RETURN).AddData(data).Script()
```

txscript/standard_test.go
```diff
@@ -1013,3 +1013,28 @@ func TestStringifyClass(t *testing.T) {
 		}
 	}
 }
+
+// TestNullDataScript tests whether NullDataScript returns a valid script.
+func TestNullDataScript(t *testing.T) {
+	data := []byte{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16,
```
davecgh: Please make this table driven so different data can be tested.

Something like the following

```Go
tests := []struct{
	name string
	data []byte
	valid bool
	expected []byte
}{
	{
		name: "more than max allowed",
		data: hexToBytes("000102....80"),
		valid: false,
		expected: nil,
	},
	{
		name: "small data",
		data: hexToBytes("01"),
		valid: false,
		expected: []byte{OP_RETURN, 0x51},
	},
	...
}

for i, test := range tests {
	... // Call func, check for err against valid, bytes.Compare(result, test.expected), etc
}
````

File: txscript/standard.go (hunk 1 above)
Comment (DanielKrawisz) [inline txscript/standard.go]: Sorry, but what's noncononical about what I did?

Comment (davecgh) [inline txscript/standard.go]: Another case would be:

`txscript.NullDataScript(bytes.Repeat([]byte{0x01, 24))`

Your code would produce `OP_RETURN OP_PUSHDATA1 0x18 0x01010101...` instead of `OP_RETURN OP_DATA_24 0x01010101...`.

File: txscript/standard_test.go (hunk 2 above)
Comment (davecgh) [inline txscript/standard_test.go]: Another test:

    {
        name: "small data",
        data: hexToBytes("0102030405060708090a0b0c0d0e0f101112131415161718"),
        valid: false,
        expected: hexToBytes("6a180102030405060708090a0b0c0d0e0f101112131415161718"),
    },

File: txscript/standard.go (hunk 1 above)
Comment (DanielKrawisz) [inline txscript/standard.go]: I guess I just don't understand bitcoin scripts well enough.

Comment (davecgh) [inline txscript/standard.go]: In case he above wasn't clear, here is some code using both that should illustrate better:

```Go
func main() {
	data := hexToBytes("0102030405060708091a1b1c1d1e1f101112")
	script, _ := txscript.NewScriptBuilder().AddOp(txscript.OP_RETURN).AddData(data).Script()
	disasm, _ := txscript.DisasmString(script)
	fmt.Printf("%x\n", script)
	fmt.Println(disasm)

	script, _= NullDataScript(data)
	disasm, _ = txscript.DisasmString(script)
	fmt.Printf("%x\n", script)
	fmt.Println(disasm)}
```

Output:
```
6a120102030405060708091a1b1c1d1e1f101112
OP_RETURN 0102030405060708091a1b1c1d1e1f101112
6a4c120102030405060708091a1b1c1d1e1f101112
OP_RETURN 0102030405060708091a1b1c1d1e1f101112
```

Notice how they both disassemble to the same thing, but the script builder version is the shorter, canonical form.

txscript/standard_test.go
DanielKrawisz: Thanks for all your advice but I'm still pretty confused. Your example "01" => "6a51" is what the code you suggested produces, but the result is not recognized as a NullDataTy when I use GetScriptClass. It returns NonStandardTy. Is that correct? If so, then NullDataScript should probably return an error rather than 6a51.

davecgh: Nice.  We just found a bug in the `GetScriptClass` function.  I confirmed that 6a51 is indeed showing up as `nonstandard`, but it most certainly is a `nulldata` script.

```
{
  "asm": "OP_RETURN 1",
  "type": "nulldata",
  "p2sh": "35zfU45S1vbHDhfY3ibN8XgCafBUzTvT3a"
}
```

DanielKrawisz: Great, well that would explain why my original function was wrong because I just followed the existing isNullData function to make something that would fit it. I have updated my pr, and one of the tests doesn't pass yet because of this bug. It will pass once isNullData is fixed though.

davecgh:
I pushed #790 to fix the `GetScriptClass` nulldata with small integer detection issue discussed in the review.

DanielKrawisz:
I thiiiink I figured out how to correct isNullData and I added a new test case to scriptClassTests. 

As for whether this pr is necessary, all I wanted to do was add an OP_RETURN output to a tx and that was a pretty big job for someone who is just learning the code base! It sure would have been convenient for me to have had this func available from the start.

txscript/standard_test.go
```diff
@@ -1013,3 +1019,89 @@ func TestStringifyClass(t *testing.T) {
...
+
+	tests := []struct {
+		name     string
+		data     []byte
+		valid    bool
+		expected []byte
+	}{
+		{
+			name:     "small data",
+			data:     tiny,
```
davecgh: You can use `data: hexToBytes("01"),` to do the scripts inline.

txscript/standard_test.go
```diff
@@ -1013,3 +1019,89 @@ func TestStringifyClass(t *testing.T) {
...
+		if err != nil {
+			t.Errorf("NullDataScript test %v: Unexpected error returned %v ",
+				test.name, err)
+
+			continue
+		}
+
+		// Check that the expected result was returned.
+		if !bytes.Equal(script, test.expected) {
+			t.Errorf("NullDataScript test %v: Expected %v, got %v",
```
davecgh: You can use %x in the `Errorf` format string to get hex.

```Go
t.Errorf("%s: expected %x, got %x`, test.name, test.expected, script)
```

Also, as above, please include a `continue` after the `t.Errorf` so it continues to the next test.

txscript/standard_test.go
```diff
@@ -1013,3 +1019,89 @@ func TestStringifyClass(t *testing.T) {
...
+			name:     "too big",
+			data:     tooLong,
+			valid:    false,
+			expected: nil,
+		},
+	}
+
+	for _, test := range tests {
+		script, err := txscript.NullDataScript(test.data)
+
```
davecgh: Please combine the checks here and add a continue after the error so it skips to the next test:

```Go
if test.valid && err != nil {
	t.Errorf("....")
	continue
} else if !test.valid && err == nil {
	t.Errorf("....")
	continue
}
```

txscript/standard.go
```diff
@@ -132,14 +132,23 @@ func isNullData(pops []parsedOpcode) bool {
...
 	}
 
+	// Otherwise, OP_RETURN followed by some data is allowed.
+	// This can either be in the form of an OP_PUSHDATA, an
+	// OP_DATA, or an OP_0 through OP_16.
 	return l == 2 &&
 		pops[0].opcode.value == OP_RETURN &&
-		pops[1].opcode.value <= OP_PUSHDATA4 &&
-		len(pops[1].data) <= MaxDataCarrierSize
+		((pops[1].opcode.value <= OP_PUSHDATA4 &&
```
davecgh: I pushed PR #709 which does this via `isSmallInt` which is designed for this purpose and adds a bunch of tests for the various edge conditions.  Please remove this and rebase the PR on top of 709.

davecgh:
Yes, I agree it's more convenient to do:

```Go
script, err := txscript.NullDataScript(data)
```
versus
```Go
script, err := txscript.NewScriptBuilder().AddOp(txscript.OP_RETURN).AddData(data).Script()
```

I appreciate the PR.

txscript/standard.go
```diff
@@ -350,6 +359,17 @@ func PayToAddrScript(addr btcutil.Address) ([]byte, error) {
...
 
+// NullDataScript creates a provably prunable script
+// containing OP_RETURN followed by the passed data.
+func NullDataScript(data []byte) ([]byte, error) {
+	if len(data) > MaxDataCarrierSize {
+		return nil, ErrStackLongScript
+	}
+
+	builder := NewScriptBuilder()
+	return builder.AddOp(OP_RETURN).AddData(data).Script()
```
davecgh: I know I originally showed this on two lines, but I forgot it was in the `txscript` package itself, so it'll fit on a single line.

Mind making it the following single line?

```Go
return NewScriptBuilder().AddOp(OP_RETURN).AddData(data).Script()
```

txscript/standard_test.go
DanielKrawisz: standard_test.go is part of package txscript_test, so hexToBytes isn't available, as it is in package txscript.

davecgh: Ok, I see it's `decodeHex` in the test package.  I'll need to combine those when doing aforementioned PR.

txscript/standard_test.go
DanielKrawisz: Ok, I did this but I think it makes the rest of the test look a little weird. I updated the pr so you can see for yourself.

davecgh:
Hmm, not sure how you rebased it there, but it somehow ended up with an extra `btcjson` commit.

There should only end up being two commits:
`* txscript: Correct nulldata standardness check.`  (the one from PR #790)
`* txscript: Add NullDataScript function.`  (your commit)

The git command should look something like:

```bash
$ git rebase -i --onto 4d4612 f78f2fb^
```

Where 4d4612 is the `txscript: Correct nulldata standardness check` commit and f78f2fb is your first commit that you want to squash into.

davecgh:
Both #790 and #791 have been merged to master now, so you should just be able to do a normal rebase to master now.

DanielKrawisz:
Ok I fixed the rebasing... I don't know how that happened, sorry!

PR #794, opened 2016-10-21
docs: Add fullblocktests entry and minor consistency updates

davecgh:


blockchain/fullblocktests/doc.go
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2013-2014 The btcsuite developers
```
davecgh: Nice catch.  Fixed.

PR #795, opened 2016-10-21
txscript: Cleanup and improve NullDataScript tests.

davecgh:
This modifies the recently-added `NullDataScript` function in several ways in an effort to make them more consistent with the tests in the rest of the code base and improve/correct the logic:

- Use the `hexToBytes` and `mustParseShortForm` functions
- Consistently format the test errors
- Replace the `valid` bool flag with an expected error and test against it
- Ensure the returned script type is the expected type in all cases

### DIFF
--- a/txscript/standard.go
+++ b/txscript/standard.go
@@ -351,8 +351,8 @@ func PayToAddrScript(addr btcutil.Address) ([]byte, error) {
 	return nil, ErrUnsupportedAddress
 }
 
-// NullDataScript creates a provably prunable script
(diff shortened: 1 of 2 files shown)


dajohi:
I think we should have a MaxDataCarrierSize+1 test.

davecgh:
@dajohi: There is one.  It's the final test:

```Go
		name: "too big",
		...
		err: ErrStackLongScript,
		class:    NonStandardTy,
```

PR #797, opened 2016-10-21
txscript: Remove OP_SMALLDATA

dajohi:


txscript/standard.go
```diff
@@ -129,7 +129,7 @@ func isMultiSig(pops []parsedOpcode) bool {
 // false otherwise.
 func isNullData(pops []parsedOpcode) bool {
 	// A nulldata transaction is either a single OP_RETURN or an
-	// OP_RETURN SMALLDATA (where SMALLDATA is a data push up to
+	// OP_RETURN with any sequence of data pushes up to
```
davecgh: While this will eventually be the case, it's not true as of this PR.  We should leave this with the correct comment until the behavior is actually changed.

PR #798, opened 2016-10-21
Do not add inbound peers to the address manager.

dajohi:


dajohi:
This matches https://github.com/bitcoin/bitcoin/pull/8594

PR #799, opened 2016-10-22
chaincfg: Introduce new type DNSSeed

dajohi:
DNSSeed defines a DNS Seed with a hostname and whether it supports
filtering by service flag bits.

### DIFF
--- a/chaincfg/params.go
+++ b/chaincfg/params.go
@@ -50,6 +50,16 @@ type Checkpoint struct {
 	Hash   *chainhash.Hash
 }
 
+// DNSSeed identifies a DNS seed.
+type DNSSeed struct {
+	// Host defines the hostname of the seed.
+	Host string
+
+	// HasFiltering defines whether the seed supports filtering
+	// by service flags (wire.ServiceFlag).
+	HasFiltering bool
+}
+
 // Params defines a Bitcoin network by its parameters.  These parameters may be
 // used by Bitcoin applications to differentiate networks as well as addresses
 // and keys for one network from those intended for use on another network.
@@ -65,7 +75,7 @@ type Params struct {
 
 	// DNSSeeds defines a list of DNS seeds for the network that are used
 	// as one method to discover peers.
-	DNSSeeds []string
+	DNSSeeds []DNSSeed
 
 	// GenesisBlock defines the first block of the chain.
 	GenesisBlock *wire.MsgBlock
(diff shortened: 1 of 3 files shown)


dajohi:
No functional change intended.

chaincfg/params.go
```diff
@@ -50,6 +50,16 @@ type Checkpoint struct {
 	Hash   *chainhash.Hash
 }
 
+// DNSSeed identifies a DNS seed.
+type DNSSeed struct {
+	// Host defines the hostname of the seed.
+	Host string
+
+	// HasFiltering defines whether the seed supports filtering
+	// by wire.ServiceFlag's.
```
davecgh: nit: I don't think there is any reason for this to possessive.  Perhaps just `filtering by service flags (wire.ServiceFlag).`

chaincfg/params.go
```diff
@@ -155,14 +165,29 @@ var MainNetParams = Params{
...
-		"seed.bitcoin.sipa.be",
-		"dnsseed.bluematt.me",
-		"dnsseed.bitcoin.dashjr.org",
-		"seed.bitcoinstats.com",
-		"seed.bitnodes.io",
-		"bitseed.xf2.org",
-		"seed.bitcoin.jonasschnelli.ch",
+	DNSSeeds: []DNSSeed{
+		{
+			"seed.bitcoin.sipa.be",
```
davecgh: I think these will comfortably fit on a single line each and would look cleaner.

```Go
DNSSeeds: []DNSSeed{
	{"seed.bitcoin.sipa.be", true},
	{"dnsseed.bluematt.me", true},
	{"dnsseed.bitcoin.dashjr.org", false},
	{"seed.bitcoinstats.com", true},
	...
}

File: chaincfg/params.go [hunk 3]
```diff
@@ -288,10 +313,20 @@ var TestNet3Params = Params{
 	Name:        "testnet3",
 	Net:         wire.TestNet3,
 	DefaultPort: "18333",
-	DNSSeeds: []string{
-		"testnet-seed.bitcoin.schildbach.de",
-		"testnet-seed.bitcoin.petertodd.org",
-		"testnet-seed.bluematt.me",
+	DNSSeeds: []DNSSeed{
+		{
+			"testnet-seed.bitcoin.jonasschnelli.ch",
```
davecgh: Same here.

connmgr/seed.go
```diff
@@ -29,8 +30,17 @@ type OnSeed func(addrs []*wire.NetAddress)
...
+
 	for _, seeder := range chainParams.DNSSeeds {
+		var host string
+		if !seeder.HasFiltering || reqServices == wire.SFNodeNetwork {
+			host = seeder.Host
+		} else {
+			host = fmt.Sprintf("x%x.%s", reqServices, seeder.Host)
+		}
+
 		go func(seeder string) {
```
davecgh: Should this make the argument match the name passed to it or choose a different name altogether?  It seems to me like it might be rather easier to confuse this with the new `DNSSeed` struct in the code below due to it having the same name as the variable in the outer scope introduced by the for loop over `chainParams.DNSSeeds`.

chaincfg/params.go
dajohi: updated

connmgr/seed.go
dajohi: yes.  changed a few variable names..

PR #802, opened 2016-10-24
mempool: Add docs.go and flesh out README.md.

davecgh:
This consists of two commits.  The first adds the mempool to the [Developer Resources](https://github.com/btcsuite/btcd/blob/master/docs/README.md#5-developer-resources) section of [docs/README.md](https://github.com/btcsuite/btcd/blob/master/docs/README.md).  The second adds a `docs.go`file to the [mempool](https://github.com/btcsuite/btcd/blob/master/mempool) package and fleshes out its `README.md` which previously still had that it was a work in progress.

### DIFF
--- a/mempool/doc.go
+++ b/mempool/doc.go
@@ -1,0 +1,81 @@
+// Copyright (c) 2016 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+/*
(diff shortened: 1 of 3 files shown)


mempool/doc.go
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) 2013-2014 The btcsuite developers
```
davecgh: Fixed.

PR #808, opened 2016-10-25
integration: Initial implementation.

davecgh:
This introduces a new package to house tests which use the `rpctest` package to programmatically drive end-to-end integration testing and move the existing RPC integration tests into the new package.


davecgh:
This was merged even though it doesn't show as such.

PR #809, opened 2016-10-25
bmgr: Remove block manager chain state.

davecgh:
This removes the block manager chain state in favor of using the `blockchain.BlockChain` instance now that it is safe for concurrency.


dajohi:
Tested OK.

PR #810, opened 2016-10-25
mining: Introduce a block template generator.

davecgh:
**This requires PR #809**

This introduces a new type named `BlkTmplGenerator` which encapsulates the various state needed to generate block templates.

This is useful since it means code that needs to generate block templates can simply accept the generator rather than needing access to all of the additional state which in turn will ultimately make it easier to split the mining code into its own package.


davecgh:
I've tested this using both calls to `getblocktemplate` and the woefully decprecated `getwork`.

PR #816, opened 2016-10-26
mining: Add basic priority calculation tests.

davecgh:
This adds some basic tests for the priority calculation code in the mining policy.


dajohi:
Could you add a test to hit
```
        if overhead >= serializedTxSize {
                return 0.0
        }
```
which would bring it to 100%

davecgh:
As mentioned, it is not currently possible for that condition to be triggered.  When segwit introduces "weighted" transaction serialize sizes, it could be possible and so the check is needed to safeguard against future updates even though it's not possible at the current time.

PR #817, opened 2016-10-27
wire: Make NewMsgTx accept the tx version.

davecgh:
**Requires PR btcsuite/btcutil#82**

This modifies the `NewMsgTx` function to accept the transaction version as a parameter and updates all callers.

The reason for this change is so the transaction version can be bumped in wire without breaking existing tests and to provide the caller with the flexibility to create the specific transaction version they desire.

### DIFF
--- a/blockchain/validate_test.go
+++ b/blockchain/validate_test.go
@@ -111,8 +111,7 @@ func TestCheckBlockSanity(t *testing.T) {
 func TestCheckSerializedHeight(t *testing.T) {
 	// Create an empty coinbase template to be used in the tests below.
 	coinbaseOutpoint := wire.NewOutPoint(&chainhash.Hash{}, math.MaxUint32)
-	coinbaseTx := wire.NewMsgTx()
-	coinbaseTx.Version = 2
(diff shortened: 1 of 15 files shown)


blockchain/validate_test.go
```diff
@@ -111,8 +111,7 @@ func TestCheckBlockSanity(t *testing.T) {
 func TestCheckSerializedHeight(t *testing.T) {
 	// Create an empty coinbase template to be used in the tests below.
 	coinbaseOutpoint := wire.NewOutPoint(&chainhash.Hash{}, math.MaxUint32)
-	coinbaseTx := wire.NewMsgTx()
-	coinbaseTx.Version = 2
+	coinbaseTx := wire.NewMsgTx(1)
```
stevenroose: Here the version is changed, is that correct?

davecgh: Yes, it should've been version 1 all along.  The serialized height in the coinbase is required as of version 2 _blocks_ as opposed to transactions.  Version 2 transactions did not exist at that point in time.

PR #820, opened 2016-10-28
server: Notify connmgr of available connection requests.

davecgh:
This corrects a few issues introduced with the connection manager where the server was not notifying the connection manager when a connection request is available again.

The cases resolved are:
- Unable to initialize a server peer instance in response to the connection
- Failure to associate the connection with the server peer instance
- Disconnection of a non-persistent outbound peer

It also changes the log message to a debug in the former case because it's not something that should be shown to the user as an error given it's not due to anything the user has misconfigured nor is it even unexpected if an invalid address is provided.

Finally, it also changes the log message in peer for inability to start a peer to debug since the inability for a peer to negotiate is not something that should be a warning which implies something is wrong.  On the contrary, it is quite expected that various peers will connect (or be connected to) that are unable to properly negotiate for a variety of reasons.  One example would be a peer that is too old.

### DIFF
--- a/server.go
+++ b/server.go
@@ -1245,7 +1245,7 @@ func (s *server) handleDonePeerMsg(state *peerState, sp *serverPeer) {
 		if !sp.Inbound() && sp.VersionKnown() {
 			state.outboundGroups[addrmgr.GroupKey(sp.NA())]--
 		}
-		if sp.persistent && sp.connReq != nil {
+		if !sp.Inbound() && sp.connReq != nil {
 			s.connManager.Disconnect(sp.connReq.ID())
 		}
 		delete(list, sp.ID())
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -1246,7 +1246,7 @@ func (s *server) handleDonePeerMsg(state *peerState, sp *serverPeer) {
 		if !sp.Inbound() && sp.VersionKnown() {
 			state.outboundGroups[addrmgr.GroupKey(sp.NA())]--
 		}
-		if sp.persistent && sp.connReq != nil {
+		if (!sp.Inbound() || sp.persistent) && sp.connReq != nil {
```
davecgh: Good point.  I'll update it shortly.  I assume you meant if `sp.persistent` is true it implies `!sp.Inbound()` is true.

PR #825, opened 2016-10-28
btcd: support feefilter requests.

dajohi:
This only adds support for handling remote peer requests.

### DIFF
--- a/mining/mining.go
+++ b/mining/mining.go
@@ -55,6 +55,9 @@ type TxDesc struct {
 
 	// Fee is the total fee the transaction associated with the entry pays.
 	Fee int64
+
+	// FeePerKB is the fee the transaction pays in Satoshi per 1000 bytes.
+	FeePerKB int64
 }
 
 // TxSource represents a source of transactions to consider for inclusion in
@@ -575,8 +578,7 @@ mempoolLoop:
 			nextBlockHeight)
 
 		// Calculate the fee in Satoshi/kB.
-		txSize := tx.MsgTx().SerializeSize()
-		prioItem.feePerKB = (txDesc.Fee * 1000) / int64(txSize)
+		prioItem.feePerKB = txDesc.FeePerKB
 		prioItem.fee = txDesc.Fee
 
 		// Add the transaction to the priority queue to mark it ready
--- a/server.go
+++ b/server.go
@@ -181,6 +181,9 @@ type server struct {
 // serverPeer extends the peer to maintain state shared by the server and
 // the blockmanager.
 type serverPeer struct {
+	// The following variables must only be used atomically
+	feeFilter int64
+
 	*peer.Peer
 
 	connReq         *connmgr.ConnReq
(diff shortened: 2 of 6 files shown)


server.go
```diff
@@ -782,6 +785,18 @@ func (sp *serverPeer) enforceNodeBloomFlag(cmd string) bool {
 	return true
 }
 
+func (sp *serverPeer) OnFeeFilter(p *peer.Peer, msg *wire.MsgFeeFilter) {
```
stevenroose: Out of curiosity, where is this method called?

davecgh: Given it's still a WIP, I would guess @dajohi just hasn't tied it in yet, but there will need to be an entry in `newPeerConfig` to register this handler to the peer [here](https://github.com/btcsuite/btcd/blob/83432785f3df1f0b6f8e6926dfe0987a34ea5b3c/server.go#L1530).

dajohi:
This is ready for review

server.go
```diff
@@ -782,6 +785,18 @@ func (sp *serverPeer) enforceNodeBloomFlag(cmd string) bool {
...
+	// Check that the passed minimum fee is a valid amount.
+	if msg.MinFee < 0 || msg.MinFee > btcutil.MaxSatoshi {
+		peerLog.Debugf("Peer %v sent an invalid feefilter '%v' -- "+
+			"disconnecting", sp, msg.MinFee)
```
davecgh: I would suggest this should use ` btcutil.Amount(msg.MinFee)` so that the log will have it nicely formatted with the unit as opposed to just being a raw number.

mining/mining.go
```diff
@@ -55,6 +55,13 @@ type TxDesc struct {
 
 	// Fee is the total fee the transaction associated with the entry pays.
 	Fee int64
+
+	// FeePerKB is the fee the transaction pays in Satoshi per 1000 bytes.
+	FeePerKB int64
+
+	// SerializeSize is the cached number of bytes of the underlying
+	// transaction.
+	SerializeSize int
```
davecgh: I would prefer this not be saved.  It is super cheap to compute and, unlike `FeePerKB`, it's not needed anywhere in a hot path.  This would add 8 bytes per transaction which, in turn, reduces the number that can fit into a given amount of memory.  I don't believe the trade off is worth it.

server.go
```diff
@@ -1316,20 +1331,28 @@ func (s *server) handleRelayInvMsg(state *peerState, msg relayMsg) {
...
-				}
-
 				if !sp.filter.MatchTxAndUpdate(txD.Tx) {
 					return
 				}
 			}
+
+			// Don't relay the transaction if the transaction fee-per-kb
+			// is less than the peer's feefilter.
+			feeFilter := atomic.LoadInt64(&sp.feeFilter)
```
davecgh: Should this come before the bloom filter check?  I mention it because it's cheaper, and, as a general rule, it's typically better to save the more expensive checks to last so they can be avoided when the cheaper checks don't pass.

dajohi: Good catch.

PR #827, opened 2016-10-28
mempool:  Return type TxDesc instead of type btcutil.Tx

dajohi:
This will provide callers more information on the accepted transaction.

### DIFF
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -481,10 +481,10 @@ func (mp *TxPool) RemoveDoubleSpends(tx *btcutil.Tx) {
 // helper for maybeAcceptTransaction.
 //
 // This function MUST be called with the mempool lock held (for writes).
-func (mp *TxPool) addTransaction(utxoView *blockchain.UtxoViewpoint, tx *btcutil.Tx, height int32, fee int64) {
+func (mp *TxPool) addTransaction(utxoView *blockchain.UtxoViewpoint, tx *btcutil.Tx, height int32, fee int64) *TxDesc {
 	// Add the transaction to the pool and mark the referenced outpoints
 	// as spent by the pool.
-	mp.pool[*tx.Hash()] = &TxDesc{
+	txD := &TxDesc{
 		TxDesc: mining.TxDesc{
 			Tx:     tx,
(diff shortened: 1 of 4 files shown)


server.go
```diff
@@ -972,18 +972,18 @@ func (s *server) RemoveRebroadcastInventory(iv *wire.InvVect) {
...
 	// Generate and relay inventory vectors for all newly accepted
 	// transactions into the memory pool due to the original being
 	// accepted.
-	for _, tx := range newTxs {
+	for _, txD := range newTxs {
 		// Generate the inventory vector and relay it.
-		iv := wire.NewInvVect(wire.InvTypeTx, tx.Hash())
-		s.RelayInventory(iv, tx)
+		iv := wire.NewInvVect(wire.InvTypeTx, txD.Tx.Hash())
+		s.RelayInventory(iv, txD)
```
davecgh: This needs to be `txD.Tx` or `handleRelayInvMsg` that [type asserts](https://github.com/btcsuite/btcd/blob/1a69eb0617283de0abf9fab0bf874adb4c8cf0f2/server.go#L1322) the data to a `*btcutil.Tx` will fail.

mempool/mempool.go
```diff
@@ -831,18 +824,28 @@ func (mp *TxPool) maybeAcceptTransaction(tx *btcutil.Tx, isNew, rateLimit, rejec
...
 	if err != nil {
 		if cerr, ok := err.(blockchain.RuleError); ok {
-			return nil, chainRuleError(cerr)
+			return nil, nil, chainRuleError(cerr)
 		}
-		return nil, err
+		return nil, nil, err
+	}
+
+	txD := &TxDesc{
```
davecgh: I think this would be cleaner and less code movement to just have `addTransaction` return the TxDesc.

```Go
txD := mp.addTransaction(utxoView, tx, bestHeight, txFee)
```

davecgh:
I think the concept here is sound.  There are various cases where having access to the information without having to acquire the mempool lock to lookup what was just added is useful.

server.go
dajohi: I changed the type assertion since more changes will be coming to that function.

PR #829, opened 2016-10-31
server: Cap max outbound at max peers

tuxcanfly:
The number of max outbound connections cannot be more max peers, so we cap that at max peers.

### DIFF
--- a/server.go
+++ b/server.go
@@ -2440,9 +2440,13 @@ func newServer(listenAddrs []string, db database.DB, chainParams *chaincfg.Param
 	}
 
 	// Create a connection manager.
+	maxOutbound := defaultMaxOutbound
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -1205,16 +1205,6 @@ func (s *server) handleAddPeerMsg(state *peerState, sp *serverPeer) bool {
 
 	// TODO: Check for max peers from a single IP.
 
-	// Limit max number of total peers.
-	if state.Count() >= cfg.MaxPeers {
```
davecgh: Removing this would allow unlimited inbound peers.

tuxcanfly: Yup, removed. Working on adding the max check to `connmgr`

tuxcanfly:
Limiting this PR to fix the cap for max outbound.

PR #831, opened 2016-11-02
Switch to gometalint in goclean

jcvernaleo:
This switches goclean to use gometalint instead of running each
checking tool one at a time.

This runs them concurrently.

More importantly it will allow us to easily add additional
linters as desired.

### DIFF
--- a/goclean.sh
+++ b/goclean.sh
@@ -5,13 +5,18 @@
 # 3. go vet        (http://golang.org/cmd/vet)
 # 4. race detector (http://blog.golang.org/race-detector)
 # 5. test coverage (http://blog.golang.org/cover)
+#
+# gometaling (github.com/alecthomas/gometalinter) is used to run each each
+# static checker.
 
 set -ex
 
 # Automatic checks
-test -z "$(go fmt $(glide novendor) | tee /dev/stderr)"
-test -z "$(for package in $(glide novendor); do golint $package; done | grep -v 'ALL_CAPS\|OP_' | tee /dev/stderr)"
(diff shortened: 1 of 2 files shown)


jcvernaleo:
Really I want this for decred but I think it might be easier to try it out here first.

goclean.sh
```diff
@@ -5,13 +5,18 @@
...
 
 # Automatic checks
-test -z "$(go fmt $(glide novendor) | tee /dev/stderr)"
-test -z "$(for package in $(glide novendor); do golint $package; done | grep -v 'ALL_CAPS\|OP_' | tee /dev/stderr)"
-test -z "$(go vet $(glide novendor) 2>&1 | tee /dev/stderr)"
+test -z "$(gometalinter --disable-all \
+--enable=gofmt \
+--enable=golint \
+--enable=vet \
+--vendor --deadline=20s ./... | grep -v 'ALL_CAPS\|OP_' 2>&1 | tee /dev/stderr)"
```
jcvernaleo: According to the help --vendor both sets that variable and skips the vendor directory which seems to match what it seems to be doing.

jrick: hrm, not sure I care for that (since there is `glide novendor`) but I guess that is fine then.

davecgh:
Nice update.  This should definitely allow the builds to complete faster and also make it easier to add new linters if desired.

goclean.sh
davecgh: I don't really have a super strong opinion either since it's effectively the same, however, I do have a slight preference for `glide novendor` since we use glide and that is the "glide way".

PR #837, opened 2016-11-03
multi: Simplify code per gosimple linter.

davecgh:
This simplifies the code based on the recommendations of the [gosimple](https://github.com/dominikh/go-simple) lint tool.

### DIFF
--- a/blockchain/indexers/manager.go
+++ b/blockchain/indexers/manager.go
@@ -271,10 +271,7 @@ func (m *Manager) Init(chain *blockchain.BlockChain) error {
 		err := m.db.View(func(dbTx database.Tx) error {
 			idxKey := indexer.Key()
 			hash, height, err = dbFetchIndexerTip(dbTx, idxKey)
-			if err != nil {
-				return err
-			}
-			return nil
+			return err
 		})
 		if err != nil {
 			return err
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -2232,7 +2232,7 @@ func handleGetPeerInfo(s *rpcServer, cmd interface{}, closeChan <-chan struct{})
 			SyncNode:       p == syncPeer,
 		}
 		if p.LastPingNonce() != 0 {
-			wait := float64(time.Now().Sub(statsSnap.LastPingTime).Nanoseconds())
+			wait := float64(time.Since(statsSnap.LastPingTime).Nanoseconds())
 			// We actually want microseconds.
 			info.PingWait = wait / 1000
 		}
@@ -3740,11 +3740,7 @@ func (s *rpcServer) writeHTTPResponseHeaders(req *http.Request, headers http.Hea
 	}
 
 	_, err = io.WriteString(w, "\r\n")
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return err
 }
 
 // Stop is used by server.go to stop the rpc listener.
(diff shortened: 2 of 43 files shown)


blockchain/indexers/manager.go
```diff
@@ -271,10 +271,7 @@ func (m *Manager) Init(chain *blockchain.BlockChain) error {
 		err := m.db.View(func(dbTx database.Tx) error {
 			idxKey := indexer.Key()
 			hash, height, err = dbFetchIndexerTip(dbTx, idxKey)
-			if err != nil {
-				return err
-			}
-			return nil
+			return err
```
jrick: I've mentioned this simplification before (and I like it, and use it a lot in my own code) but when asking about this as part of the btcd codebase a while ago, I was told this is not preferable since it "hides" code coverage reporting of when this function is successful vs errored.

I'm all for this change (the code behaves identically), just pointing out the inconsistency.

rpcserver.go
```diff
@@ -3948,8 +3944,7 @@ func (s *rpcServer) jsonRPCRead(w http.ResponseWriter, r *http.Request, isAdmin
 	if err != nil {
 		errMsg := fmt.Sprintf("error reading JSON message: %v", err)
 		errCode := http.StatusBadRequest
-		http.Error(w, strconv.FormatInt(int64(errCode), 10)+" "+errMsg,
-			errCode)
+		http.Error(w, strconv.Itoa(errCode)+" "+errMsg, errCode)
```
jrick: Are either of these preferable to using fmt.Sprintf or should we switch to that?

rpcserver.go
```diff
@@ -3964,16 +3959,14 @@ func (s *rpcServer) jsonRPCRead(w http.ResponseWriter, r *http.Request, isAdmin
 		errMsg := "webserver doesn't support hijacking"
 		rpcsLog.Warnf(errMsg)
 		errCode := http.StatusInternalServerError
-		http.Error(w, strconv.FormatInt(int64(errCode), 10)+" "+errMsg,
-			errCode)
+		http.Error(w, strconv.Itoa(errCode)+" "+errMsg, errCode)
```
jrick: see above comment

rpctest/rpc_harness.go
```diff
@@ -240,16 +240,13 @@ func (h *Harness) SetUp(createTestChain bool, numMatureOutputs uint32) error {
...
 	ticker := time.NewTicker(time.Millisecond * 100)
-out:
-	for {
-		select {
-		case <-ticker.C:
-			walletHeight := h.wallet.SyncedHeight()
-			if walletHeight == height {
-				break out
-			}
+	for range ticker.C {
```
jrick: This is an odd simplification since it actually is not semantically equivalent.  The range version will break out of the loop when the channel closes, while the old code would continuously read the zero value (and spike CPU usage).  For this particular case it looks good, but need to be careful that these static tools are actually suggesting changes that are semantically equivalent.

blockchain/indexers/manager.go
davecgh: Yes, you are correct that was the original reason for not doing the simplification (and it is still true), however, since the greater community (and even yourself as you noted) prefers this simplification and changing it means we can add gosimple to the list of linters run on every PR, I think the tradeoff is worth it.

rpcserver.go
davecgh: I'd have to benchmark to verify, but I seem to recall reading that `fmt.Sprintf` is _much_ slower than the above.  I don't have a strong opinion on changing it however.  If you (and others) would prefer `fmt.Sprintf` instead there, I can change it.

jrick: I haven't tested either but i'd expect this way to be faster than fmt (avoids reflection and a vararg function)

jrick: Um actually after looking at the full context, it's already using fmt.Sprintf.  Can we do it all at once there instead?

davecgh: This is the only case that applies to.  Changed.

Issue #581, opened 2015-12-14
[addrmgr] Update to make use of upcoming database


davecgh:
The address manager package ([addrmgr](https://github.com/btcsuite/btcd/tree/master/addrmgr)) for handling available peers currently uses a serialized JSON file for persistence and keeps all known addresses in memory.  This is not ideal since, as the number of known good addresses grows, so too does memory consumption (albeit at a pretty slow rate).  For example, recent profiling showed it using almost 50 MB on my system currently.  Another issue with the current approach is the periodic serialization of the address states has to write a lot more data than is necessary.

It would be ideal to modify the package to take allow it to take advantage of the new bucket capabilities provided by the upcoming redesigned database (#380).  However, since the package is also quite useful for non btcd-specific applications, tying it to the new database directly should be avoided.  Thus, I propose the package should introduce a new interface which it uses to load and store the relevant data.  With the interface in place, callers can provide any concrete implementation of it they need to.  For example, btcd could simply implement the interface using the new database capabilities once they're in place and btcwallet could do the same using its database.

In order to provide a smooth transition and default functionality for callers that don't care, as well as tide the package over until the new database lands, I also propose implement taking the current JSON code as a default concrete implementation of the interface.  With this in place, the package can be updated now and existing callers can simply use the provided default implementation without any change in behavior.  Then, once the new database lands, it would be a simple matter of providing a different concrete implementation to make use of it.

PR #840, opened 2016-11-03
connmgr: switch to using net.Addr interface throughout for addresses

Roasbeef:
This commit modifies the ConnManager to use the `net.Addr interface through the package instead of a plain string to represent and manipulate addresses. This change makes the package much more general as users of the package can possibly utilize custom implementations of the net.Addr interface to establish connections.

More precisely, the `ConnReq` struct has been modified to use a net.Addr instance explicitly, and the `DialFunc` type has also been modified to take a `net.Addr` directly. This latter change gives functions that adhere to the `DialFunc` type more flexibility as to exactly how the connection is established.

My motivation for this change is that I'd like to use the `connmgr` package within [`lnd`](https://github.com/lightningnetwork/lnd) to manage attempting to establish persistent outbound connections to channel counter parties. Within the Lightning Network, [all p2p connections are encrypted+authenticated by default](https://github.com/lightningnetwork/lnd/blob/master/brontide/noise.go). 

Before this commit, this additional constraint meant that I couldn't directly use the `connmgr` because all the functions/structs referred to addresses explicitly by a string. This is rather limiting as in LN addresses also are tied to a long-term identity public key. By modifying the package to use a `net.Addr` throughout, I'm utilize our implementation of `net.Addr` whose concrete struct contains the additional information required to establish an encrypted connection to peers on the network.

### DIFF
--- a/connmgr/connmanager.go
+++ b/connmgr/connmanager.go
@@ -57,7 +57,7 @@ type ConnReq struct {
 	// The following variables must only be used atomically.
 	id uint64
 
-	Addr      string
+	Addr      net.Addr
 	Permanent bool
 
 	conn       net.Conn
@@ -88,7 +88,7 @@ func (c *ConnReq) State() ConnState {
 
 // String returns a human-readable string for the connection request.
 func (c *ConnReq) String() string {
-	if c.Addr == "" {
+	if c.Addr.String() == "" {
 		return fmt.Sprintf("reqid %d", atomic.LoadUint64(&c.id))
 	}
 	return fmt.Sprintf("%s (reqid %d)", c.Addr, atomic.LoadUint64(&c.id))
@@ -137,10 +137,10 @@ type Config struct {
 
 	// GetNewAddress is a way to get an address to make a network connection
 	// to.  If nil, no new connections will be made automatically.
-	GetNewAddress func() (string, error)
+	GetNewAddress func() (net.Addr, error)
 
 	// Dial connects to the address on the named network. It cannot be nil.
-	Dial func(string, string) (net.Conn, error)
+	Dial func(net.Addr) (net.Conn, error)
 }
 
 // handleConnected is used to queue a successful connection.
@@ -281,14 +281,18 @@ func (cm *ConnManager) NewConnReq() {
 	if cm.cfg.GetNewAddress == nil {
 		return
 	}
+
 	c := &ConnReq{}
 	atomic.StoreUint64(&c.id, atomic.AddUint64(&cm.connReqCount, 1))
+
 	addr, err := cm.cfg.GetNewAddress()
 	if err != nil {
 		cm.requests <- handleFailed{c, err}
 		return
 	}
+
 	c.Addr = addr
+
 	cm.Connect(c)
 }
 
@@ -302,7 +306,7 @@ func (cm *ConnManager) Connect(c *ConnReq) {
 		atomic.StoreUint64(&c.id, atomic.AddUint64(&cm.connReqCount, 1))
 	}
 	log.Debugf("Attempting to connect to %v", c)
-	conn, err := cm.cfg.Dial("tcp", c.Addr)
+	conn, err := cm.cfg.Dial(c.Addr)
 	if err != nil {
 		cm.requests <- handleFailed{c, err}
 	} else {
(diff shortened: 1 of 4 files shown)


davecgh:
I haven't looked at this super closely yet, but a couple of initial comments:

1) I generally like the 
2) I'm fairly certain the current implementation will leak the DNS resolution over Tor because it's calling `net.ResolveTCPAddr` instead of using `cfg.lookup` and `cfg.onionlookup` which are setup in `config.go` based upon the tor parameters.

Roasbeef:
Excellent observation. I've pushed a rebased and slightly modified version of the original commit. Rather than using `net.ResolveTCPAddr` everywhere due to the change, the `connmgr.Config` struct has gained a new attribute: a closure for performing DNS lookups. This is used throughout rather than potentially incorrectly performing DNS lookup as you detailed.

connmgr/connmanager.go
```diff
@@ -288,7 +292,30 @@ func (cm *ConnManager) NewConnReq() {
 		cm.requests <- handleFailed{c, err}
 		return
 	}
-	c.Addr = addr
+
+	host, strPort, err := net.SplitHostPort(addr)
```
davecgh: Since everything else in the package now deals with `net.Addr`, I think this should be modified instead such that the `GetNewAddress` callback returns a `net.Addr` and then the `newAddressFunc` closure in `server.go` which is provided would pick up the parsing, lookup, and `net.Addr` creation.  This would work as is, but it seems to go against the spirit of the PR which seems to be aimed at making it agnostic to the specific address type.  In fact, I believe that would get rid of the need to even pass in a lookup function at all and allow the code you added in `server.go` under the `connectNodeMsg` case and `newServer` function to be refactored into a common code used in all three places (`connectNodeMsg`, `newServer`, and `newAddressFunc`).  What do you think?

I expect `addrmgr.GetAddress` itself should probably also be modified at some point to return `net.Addr` too, but that can be saved for another PR.

Roasbeef: Great suggestion! Modifying `GetNewAddress` to return a `net.Addr` directly allows for a good bit of code consolidation and allows this PR to keep the bulk of the parsing/lookup logic in a single area. 

I've just pushed a commit which makes the modification you mentioned in your comment above. 

> I expect addrmgr.GetAddress itself should probably also be modified at some point to return net.Addr too, but that can be saved for another PR.

One way to implement this would be to make `wire.NetAddress` adhere to the `net.Addr` interface by adding a `.Network()` method. This would allow the addresses returned by `addrmgr.GetAddress` to be directly passed into this context. 

As an aside, I think the `addrmgr` packaged is currently one of the most neglected packages within the project. The code could use a bit of cleaning up, and we could also dump the json file dump format in favor of creating a dedicated bucket for the data within the database. Additionally, it'd be nice to implement some of the counter measures for the attacks described [here](https://www.usenix.org/system/files/conference/usenixsecurity15/sec15-paper-heilman.pdf) and [here](https://arxiv.org/abs/1605.07524).

server.go
```diff
@@ -2432,9 +2438,13 @@ func newServer(listenAddrs []string, db database.DB, chainParams *chaincfg.Param
 					activeNetParams.DefaultPort && tries < 50 {
 					continue
 				}
-				return addrmgr.NetAddressKey(addr.NetAddress()), nil
+
+				addrString := addrmgr.NetAddressKey(addr.NetAddress())
+
```
davecgh: Super minor nit: This blank line probably shouldn't be here since it's grouped with the return in terms of logic.

connmgr/connmanager.go
davecgh: Nice.  Yes, I completely agree about `addrmgr`.  I've mentioned a few times as well that it's probably the package that needs the most improvement/cleanup in the entire repository.

That seems like a reasonable idea to make it implement `net.Addr`.  Another approach would be to create a new type in addrmgr which embeds a `wire.NetAddress` and adds the `Network` method so it implements `net.Addr`.  That way the wire type wouldn't have to be directly modified and additional metadata could be added to the new type if desired as well.

Also, there is an open issue (#581) for converting it to make use of the database as you mentioned.

server.go
Roasbeef: Fixed.

Roasbeef:
Thanks! I've addressed your final review item and squashed down the review commits into the original PR commit. I've also updated the commit message to reflect the new approach taken w.r.t `GetNewAddress`.

PR #843, opened 2016-11-04
connmgr: Implement inbound connection handling.

davecgh:
This modifies the connection manager to provide support for accepting inbound connections on a caller-provided set of listeners and notify the caller via a callback.

This is only the minimum work necessary to get inbound support into the connection manager.  The intention for future commits is to move more connection-related logic such as limiting the maximum number of overall connections and banned peer tracking into the connection manager.

It is split into 3 commits:

* The first commit removes the type definitions for the callback functions in favor of declaring them directly in the `Config` struct. since it is more consistent with the rest of the code base and is preferred since it means callers reviewing the documentation don't have to follow another level of indirection to figure out the signature.

* The second commit modifies the connection manager to add `Listeners` and `OnAccept` fields to the `Config` struct, introduces a `listenHandler` function that is run as a goroutine for each caller-provided listener in order to accept the connections and invoke the `OnAccept` callback with the connections, and adds the code needed to start and stop the listeners with the connection manager.

* The third commit adds tests for the new code by introducing a `mockListener` to the test code and making use of it to ensure connections are accepted and passed along to the `OnAccept` callback.

### DIFF
--- a/connmgr/connmanager.go
+++ b/connmgr/connmanager.go
@@ -96,6 +96,29 @@ func (c *ConnReq) String() string {
 
 // Config holds the configuration options related to the connection manager.
 type Config struct {
+	// Listeners defines a slice of listeners for which the connection
+	// manager will take ownership of and accept connections.  When a
+	// connection is accepted, the OnAccept handler will be invoked with the
+	// connection.  Since the connection manager takes ownership of these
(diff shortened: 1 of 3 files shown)


davecgh:
Rebased to deal with merge conflicts caused by #832.

PR #844, opened 2016-11-04
server: Ensure callbacks use the server peer.

davecgh:
This modifies the signatures of all `serverPeer` callbacks that are provided as `peer.Listeners` to use `_` for the first parameter name which ensures the passed peer can't be used within the function and updates
all references to the server peer.

This helps ensure any overridden methods that might be defined on a `serverPeer` will be invoked where directly calling methods on the passed `peer.Peer` would not.

Also, while here, add a comment to the `OnFeeFilter` function.

### DIFF
--- a/server.go
+++ b/server.go
@@ -314,10 +314,10 @@ func (sp *serverPeer) addBanScore(persistent, transient uint32, reason string) {
 // OnVersion is invoked when a peer receives a version bitcoin message
 // and is used to negotiate the protocol version details as well as kick start
 // the communications.
-func (sp *serverPeer) OnVersion(p *peer.Peer, msg *wire.MsgVersion) {
+func (sp *serverPeer) OnVersion(_ *peer.Peer, msg *wire.MsgVersion) {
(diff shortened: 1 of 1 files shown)


stevenroose:
Why is the peer parameter passed then? The only instantiation of `MessageListeners` I can find is for `serverPeer`, which doesn't use the parameter..

davecgh:
The `peer` package is agnostic to btcd and is useful for other things such as crawlers like bitnodes to use to gather statistics on the network.  In those cases, they are able to just make use of the peer passed to each callback, and thus the callbacks provided to the `peer` have to include the parameter even if the caller doesn't intent to use it.

In the case of btcd, additional state is kept with each peer (hence the `serverPeer` type) and so it's binding closures which encapsulate all the state for the callbacks using shorthand syntax instead and therefore it doesn't need the parameter.

The following example illustrates what I mean by binding closures if it's not clear: https://play.golang.org/p/21XrQHoiNc

PR #845, opened 2016-11-06
rpc: add new btcd extension commands: addminingaddr, delminingaddr, listminingaddrs

Roasbeef:
This PR adds three new `btcd` `JSON-RPC` extension commands: `addminingaddr`, `delminingaddr`, and `listminingaddrs` .  These RPC allows the caller to dynamically add, remove and list mining addresses to `btcd'`s list of generation addresses. Before this RPC, doing such a thing would've required a full restart of `btcd`. 

**NOTE:** This PR depends on btcsuite/btcrpcclient#105. 

The addition of this RPC also adds a new degree of flexibility to automated deployments of `btcd` using orchestration tools like `Docker`. The original inspiration for this RPC was spurred by some workarounds I had to implement when attempting to package `btcd` and `lnd` into a development environment using `docker-compose`.

A breakdown of the changes is as follows: 
 * New RPC's: `addminingaddr`, `delminingaddr`, and `listminingaddrs` have been added. In order to ensure thread-safety, a mutex `miningAddrMTx` has been added to the `config` struct. This mutex must be held at all times when reading/modifying the `miningAddr` slice. 
 * The `cpuminer.Config` struct has been modified to use a closure function for obtaining new mining addresses. This decouple the current logic of selecting a mining address (randomly select from a list) from the internals of the `cpuminer` package. In the future if/when alternate selection logic is added, then code within the package shouldn't need to change at all. 
* The `Config` struct has been modified to change the `MiningAddr` paramter from a `[]string` to a `string`. When writing integration tests for these new RPC's, I discovered that the address set via the `--miningaddr` flag was being duplicated. This change fixes the duplication. 

An interesting extension of this RPC was briefly discussed on IRC between myself and `davec`: the call could be extended to possibly accept a `BIP0032` extended public key rather than a series of addresses. Within the daemon, when generating new blocks, `btcd` could instead iterate through the HD keychain grabbing fresh addresses for each block rather than repeating through a statically coded list of addresses. Taking it a set further, further logic would be added to generate multi-sig `p2sh` addresses from addresses plucked from the keychain. 

For simplicity of this patch set, I've only implemented the most basic behavior round dynamically setting generation addresses, however the two extensions I mentioned above my be worth pursuing.

### DIFF
--- a/btcjson/btcdextcmds.go
+++ b/btcjson/btcdextcmds.go
@@ -89,6 +89,41 @@ func NewGetCurrentNetCmd() *GetCurrentNetCmd {
 	return &GetCurrentNetCmd{}
 }
 
+// AddMiningAddrCmd defines the addminingaddr JSON-RPC command.
+type AddMiningAddrCmd struct {
+	Address string `json:"address"`
+}
+
+// NewAddMiningAddrCmd returns a new instance which can be used to issue a
+// addminingaddr JSON-RPC command.
+func NewAddMiningAddrCmd(addr string) *AddMiningAddrCmd {
+	return &AddMiningAddrCmd{
+		Address: addr,
+	}
+}
+
+// DelMiningAddrCmd defines the delminingaddr JSON-RPC command.
+type DelMiningAddrCmd struct {
+	Address string `json:"address"`
+}
+
+// NewDelMiningAddrCmd returns a new instance which can be used to issue a
+// delminingaddr JSON-RPC command.
+func NewDelMiningAddrCmd(addr string) *DelMiningAddrCmd {
+	return &DelMiningAddrCmd{
+		Address: addr,
+	}
+}
+
+// ListMiningAddrsCmd defines the listminingaddrs JSON-RPC command.
+type ListMiningAddrsCmd struct{}
+
+// NewListMiningAddrsCmd returns a new instance which can be used to issue a
+// listminingaddrs JSON-RPC command.
+func NewListMiningAddrsCmd() *ListMiningAddrsCmd {
+	return &ListMiningAddrsCmd{}
+}
+
 func init() {
 	// No special flags for commands in this file.
 	flags := UsageFlag(0)
@@ -98,4 +133,7 @@ func init() {
 	MustRegisterCmd("generate", (*GenerateCmd)(nil), flags)
 	MustRegisterCmd("getbestblock", (*GetBestBlockCmd)(nil), flags)
 	MustRegisterCmd("getcurrentnet", (*GetCurrentNetCmd)(nil), flags)
+	MustRegisterCmd("addminingaddr", (*AddMiningAddrCmd)(nil), flags)
+	MustRegisterCmd("delminingaddr", (*DelMiningAddrCmd)(nil), flags)
+	MustRegisterCmd("listminingaddrs", (*ListMiningAddrsCmd)(nil), flags)
 }
--- a/config.go
+++ b/config.go
@@ -764,7 +764,7 @@ func loadConfig() (*config, []string, error) {
 
 	// Ensure there is at least one mining address when the generate flag is
 	// set.
-	if cfg.Generate && len(cfg.MiningAddrs) == 0 {
+	if cfg.Generate && len(cfg.miningAddrs) == 0 {
 		str := "%s: the generate flag is set, but there are no mining " +
 			"addresses specified "
 		err := fmt.Errorf(str, funcName)
--- a/config.go
+++ b/config.go
@@ -151,8 +151,8 @@ type config struct {
 	lookup               func(string) ([]net.IP, error)
 	oniondial            func(string, string) (net.Conn, error)
 	dial                 func(string, string) (net.Conn, error)
-	miningAddrs          []btcutil.Address
 	minRelayTxFee        btcutil.Amount
+	miningAddrs          []btcutil.Address
 }
 
 // serviceOptions defines the configuration options for the daemon as a service on
@@ -369,7 +369,9 @@ func loadConfig() (*config, []string, error) {
 	// Pre-parse the command line options to see if an alternative config
 	// file or the version flag was specified.  Any errors aside from the
 	// help message error can be ignored here since they will be caught by
-	// the final parse below.
+	// the final parse below. A copy here is made on purpose. Making this a
+	// pointer causes it to set the same fields in the same config struct
+	// on the second parse which can cause duplicate entries in slices.
 	preCfg := cfg
 	preParser := newConfigParser(&preCfg, &serviceOpts, flags.HelpFlag)
 	_, err := preParser.Parse()
--- a/docs/json_rpc_api.md
+++ b/docs/json_rpc_api.md
@@ -583,6 +583,9 @@ The following is an overview of the RPC methods which are implemented by btcd, b
 |4|[searchrawtransactions](#searchrawtransactions)|Y|Query for transactions related to a particular address.|None|
 |5|[node](#node)|N|Attempts to add or remove a peer. |None|
 |6|[generate](#generate)|N|When in simnet or regtest mode, generate a set number of blocks. |None|
+|7|[addminingaddr](#addminingaddr)|N|Add a new address to the list of generation (coinbase payout) addresses. |None|
+|8|[delminingaddr](#delminingaddr)|N|Delete an address from the list of generation (coinbase payout) addresses. |None|
+|9|[listminingaddrs](#listminingaddrs)|N|List all the active generation (coinbase payout) addresses. |None|
 
 
 <a name="ExtMethodDetails" />
@@ -660,6 +663,44 @@ The following is an overview of the RPC methods which are implemented by btcd, b
 |Parameters|1. numblocks (int, required) - The number of blocks to generate |
 |Description|When in simnet or regtest mode, generates `numblocks` blocks. If blocks arrive from elsewhere, they are built upon but don't count toward the number of blocks to generate. Only generated blocks are returned. This RPC call will exit with an error if the server is already CPU mining, and will prevent the server from CPU mining for another command while it runs. |
 |Returns|`[ (json array of strings)` <br/>&nbsp;&nbsp; `"blockhash", ... hash of the generated block` <br/>`]` |
+[Return to Overview](#MethodOverview)<br />
+
+***
+
+<a name="addminingaddr"/>
+
+|   |   |
+|---|---|
+|Method|addminingaddr|
+|Parameters|1. address (string, required) - The address to add to the list of generation addresses|
+|Description|Attempts to add the specified address to the list of mining generation addresses. Attempting to add a duplicated address results in an error. All generation addresses must be unique.|
+|Returns|Nothing|
+[Return to Overview](#MethodOverview)<br />
+
+***
+
+<a name="delminingaddr"/>
+
+|   |   |
+|---|---|
+|Method|delminingaddr|
+|Parameters|1. address (string, required) - The address to delete from the list of mining addresses|
+|Description|Attempts to remove the passed address from the list of mining addresses. Attempting to delete a non-existent address results in an error.|
+|Returns|Nothing|
+[Return to Overview](#MethodOverview)<br />
+
+***
+
+***
+
+<a name="listminingaddrs"/>
+
+|   |   |
+|---|---|
+|Method|listminingaddrs|
+|Parameters|None|
+|Description|Returns a list of all the currently active generation (coinbase payout) addresses.|
+|Returns|Nothing|
 [Return to Overview](#MethodOverview)<br />
 
 ***
--- a/integration/rpcserver_test.go
+++ b/integration/rpcserver_test.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/rpctest"
+	"github.com/btcsuite/btcutil"
 )
 
 func testGetBestBlock(r *rpctest.Harness, t *testing.T) {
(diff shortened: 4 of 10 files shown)


dajohi:
I'd rather have this be called addminingaddr since 'set' to me implies overwriting the current settings.  and then perhaps a delete as well?

Roasbeef:
Sure, what do you think about also throwing in a companion to the `add`, and `delete` calls: `list`?

stevenroose:
What happens when a node has multiple mining addresses?

Roasbeef:
@stevenroose the current algorithm pseudorandomly selects a mining address from the current set

davecgh:
I think having ``{add,delete}miningaddress` and `listminingaddresses` makes fine sense.  I haven't reviewed the code yet, but should be able to get it reviewed later this week.

Roasbeef:
I've pushed a new version of the PR which adds three new RPC's: `addminingaddr`, `delminingaddr` and `listminingaddrs`. I've also added a new integration tests to exercise the behavior and have also included a commit which fixes (what I consider) to be an existing bug in the configuration parsing.

davecgh:
Thanks for the updates.  One thing I noticed is the second commit converts the MiningAddr from `[]string` to `string` and I definitely don't think that should be changed.  There are already configurations out there that rely on the behavior it provides and changing it would break existing setups.

I know from IRC that you were seeing duplicates, but the reason for that is due to the following change:
```diff
-	preCfg := cfg
+	preCfg := &cfg
```

That was intentionally a copy. Making it a pointer causes it to set the same fields in the same config struct on the second parse which leads to appending entries in any slices twice.

rpcserver.go
```diff
@@ -3495,6 +3516,87 @@ func handleSetGenerate(s *rpcServer, cmd interface{}, closeChan <-chan struct{})
...
+			cfg.miningAddrMtx.Unlock()
+			return nil, nil
+		}
+	}
+	cfg.miningAddrMtx.Unlock()
+
+	// If we've reached this point, then the targeted mining address wasn't
+	// found so we return an error indicating so.
+	return nil, &btcjson.RPCError{
+		Code:    btcjson.ErrRPCInternal.Code,
```
davecgh: This isn't an internal error.  It was an invalid address provided by the caller.  Thus, it should be `btcjson.ErrRPCInvalidAddressOrKey`.

rpcserverhelp.go
```diff
@@ -488,6 +488,17 @@ var helpDescsEnUS = map[string]string{
 	"setgenerate-generate":     "Use true to enable generation, false to disable it",
 	"setgenerate-genproclimit": "The number of processors (cores) to limit generation to or -1 for default",
 
+	// AddMiningAddrCmd help.
```
davecgh: Please alphabetize these too so it's all in the same order it's defined in the other file(s).

rpcserverhelp.go
```diff
@@ -488,6 +488,17 @@ var helpDescsEnUS = map[string]string{
...
+	"addminingaddr--synopsis": "Adds an additional available mining address to the set of mining addresses for the daemon",
+	"addminingaddr-address":   "The address to append to the set of mining addresses",
+
+	// ListMiningAddrsCmd help.
+	"listminingaddrs--synopsis": "Returns a list of the current active mining addresses",
+
```
davecgh:
```Go
	"listminingaddrs--result0":  "List of current active mining addresses",
```

Otherwise:

```
$ btcctl help listminingaddrs
...
Result:
Nothing
```

rpcserverhelp.go
```diff
@@ -618,6 +629,9 @@ var rpcResultTypes = map[string][]interface{}{
 	"searchrawtransactions": {(*string)(nil), (*[]btcjson.SearchRawTransactionsResult)(nil)},
 	"sendrawtransaction":    {(*string)(nil)},
 	"setgenerate":           nil,
+	"addminingaddr":         nil,
+	"delminingaddr":         nil,
+	"listminingaddrs":       nil,
```
davecgh: This returns a slice of strings, so it should be:

```Go
{(*[]string)(nil)},
```

That would also trigger a test failure due to the missing `listminingaddrs--result0` help key I commented on above.

davecgh (changes requested):

davecgh:
I forgot to mention in my review that I really like the idea here.  It will also simplify setting up simulation networks since you currently need btcd running first to get wallet to generate an address that you then have to stop btcd to configure with it.

It also allows miners to avoid downtime when changing things which is quite important for them.

rpcserver.go
```diff
@@ -169,6 +169,9 @@ var rpcHandlersBeforeInit = map[string]commandHandler{
 	"searchrawtransactions": handleSearchRawTransactions,
 	"sendrawtransaction":    handleSendRawTransaction,
 	"setgenerate":           handleSetGenerate,
+	"addminingaddr":         handleAddMiningAddr,
```
Roasbeef: Fixed.

rpcserver.go
```diff
@@ -3495,6 +3516,87 @@ func handleSetGenerate(s *rpcServer, cmd interface{}, closeChan <-chan struct{})
 	return nil, nil
 }
 
+func handleAddMiningAddr(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (interface{}, error) {
```
Roasbeef: Fixed.

rpcserver.go
```diff
@@ -3495,6 +3516,87 @@ func handleSetGenerate(s *rpcServer, cmd interface{}, closeChan <-chan struct{})
...
+	c := cmd.(*btcjson.AddMiningAddrCmd)
+
+	// Decode the addresses encoded as a string into an in-memory object.
+	addr, err := btcutil.DecodeAddress(c.Address, activeNetParams.Params)
+	if err != nil {
+		return nil, err
```
Roasbeef: Fixed.

rpcserver.go
```diff
@@ -3495,6 +3516,87 @@ func handleSetGenerate(s *rpcServer, cmd interface{}, closeChan <-chan struct{})
...
+	}
+
```
Roasbeef: Nice catch! Fixed.

rpcserver.go
```diff
@@ -3495,6 +3516,87 @@ func handleSetGenerate(s *rpcServer, cmd interface{}, closeChan <-chan struct{})
...
+	// If adding this new address will not introduce any duplicate mining
+	// addresses, then append it to the list of available mining addresses.
+	cfg.miningAddrMtx.Lock()
+	cfg.miningAddrs = append(cfg.miningAddrs, addr)
+	cfg.miningAddrMtx.Unlock()
+
+	return nil, nil
+}
+
+func handleDelMiningAddr(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (interface{}, error) {
```
Roasbeef: Fixed.

rpcserver.go
```diff
@@ -3495,6 +3516,87 @@ func handleSetGenerate(s *rpcServer, cmd interface{}, closeChan <-chan struct{})
...
+	c := cmd.(*btcjson.DelMiningAddrCmd)
+
+	// Decode the addresses encoded as a string into an in-memory object.
+	addr, err := btcutil.DecodeAddress(c.Address, activeNetParams.Params)
+	if err != nil {
+		return nil, err
```
Roasbeef: Fixed.

rpcserver.go
```diff
@@ -3495,6 +3516,87 @@ func handleSetGenerate(s *rpcServer, cmd interface{}, closeChan <-chan struct{})
...
+	}
+
```
Roasbeef: Fixed.

rpcserver.go
```diff
@@ -3495,6 +3516,87 @@ func handleSetGenerate(s *rpcServer, cmd interface{}, closeChan <-chan struct{})
...
+	cfg.miningAddrMtx.Lock()
+	for i := 0; i < len(cfg.miningAddrs); i++ {
```
Roasbeef: Changed to use the single value range syntax instead. I always seems to forget about the semantics of the single value range syntax over slices!

rpcserver.go
```diff
@@ -3495,6 +3516,87 @@ func handleSetGenerate(s *rpcServer, cmd interface{}, closeChan <-chan struct{})
...
+		if cfg.miningAddrs[i].String() == addr.String() {
```
Roasbeef: Hmm, yeah I agree with your analysis. Although I don't think any miners would be using p2pk address in this age, deletion should be more precise, hence the usage of `String` here.

rpcserver.go
Roasbeef: Fixed.

config.go
```diff
@@ -133,7 +134,7 @@ type config struct {
 	NoRelayPriority      bool          `long:"norelaypriority" description:"Do not require free or low-fee transactions to have high priority for relaying"`
 	MaxOrphanTxs         int           `long:"maxorphantx" description:"Max number of orphan transactions to keep in memory"`
 	Generate             bool          `long:"generate" description:"Generate (mine) bitcoins using the CPU"`
-	MiningAddrs          []string      `long:"miningaddr" description:"Add the specified payment address to the list of addresses to use for generated blocks -- At least one address is required if the generate option is set"`
+	MiningAddr           string        `long:"miningaddr" description:"Add the specified payment address to the list of addresses to use for generated blocks -- At least one address is required if the generate option is set"`
```
Roasbeef: Created a new commit to implement this bug fix.

config.go
```diff
@@ -153,6 +154,7 @@ type config struct {
 	dial                 func(string, string) (net.Conn, error)
 	miningAddrs          []btcutil.Address
```
Roasbeef: Fixed.

config.go
```diff
@@ -370,8 +372,8 @@ func loadConfig() (*config, []string, error) {
 	// file or the version flag was specified.  Any errors aside from the
 	// help message error can be ignored here since they will be caught by
 	// the final parse below.
-	preCfg := cfg
-	preParser := newConfigParser(&preCfg, &serviceOpts, flags.HelpFlag)
+	preCfg := &cfg
```
Roasbeef: Added

btcjson/btcdextcmds.go
```diff
@@ -89,6 +89,41 @@ func NewGetCurrentNetCmd() *GetCurrentNetCmd {
 	return &GetCurrentNetCmd{}
 }
 
+// AddMiningAddrCmd defines the setminingaddr JSON-RPC command.
+type AddMiningAddrCmd struct {
```
Roasbeef: Fixed. Must've been left over from the name change from `setminingaddr` to `addminingaddr`.

btcjson/btcdextcmds.go
```diff
@@ -89,6 +89,41 @@ func NewGetCurrentNetCmd() *GetCurrentNetCmd {
...
+	Address string `json:"address"`
+}
+
+// NewAddMiningAddrCmd returns a new instance which can be used to issue a
+// setminingaddr JSON-RPC command.
```
Roasbeef: Fixed.

mining/cpuminer/cpuminer.go
```diff
@@ -334,9 +333,13 @@ out:
 			continue
 		}
 
-		// Choose a payment address at random.
-		rand.Seed(time.Now().UnixNano())
-		payToAddr := m.cfg.MiningAddrs[rand.Intn(len(m.cfg.MiningAddrs))]
+		// Obtain a new payment address selected according to the logic
+		// of the current mining address generation function.
+		payToAddr, err := m.cfg.GetMiningAddr()
```
Roasbeef: Fixed.

mining/cpuminer/cpuminer.go
```diff
@@ -588,9 +591,13 @@ func (m *CPUMiner) GenerateNBlocks(n uint32) ([]*chainhash.Hash, error) {
 		m.submitBlockLock.Lock()
 		curHeight := m.g.BestSnapshot().Height
 
-		// Choose a payment address at random.
-		rand.Seed(time.Now().UnixNano())
-		payToAddr := m.cfg.MiningAddrs[rand.Intn(len(m.cfg.MiningAddrs))]
+		// Obtain a new payment address selected according to the logic
+		// of the current mining address generation function.
+		payToAddr, err := m.cfg.GetMiningAddr()
```
Roasbeef: Fixed.

integration/rpcserver_test.go
```diff
@@ -94,10 +95,80 @@ func testGetBlockHash(r *rpctest.Harness, t *testing.T) {
...
+	// Attempting to add the same address twice should be rejected due to
+	// duplicates.
+	if err := r.Node.AddMiningAddr(addrs[0]); err == nil {
+		t.Fatalf("duplicate mining address wasn't rejected")
+	}
+
+	// Next, try to delete a single address, then immediately try to delete
+	// the same address. The second call should result in an error as the
+	// address is no longer within the set of mining addresses.
+	if err := r.Node.DelMiningAddr(addrs[0]); err != nil {
```
Roasbeef: Excellent, point. I've changed the duplicate deletion detection portion in the integration test in order to exercise deletion from near the mid-point of the slice.

integration/rpcserver_test.go
```diff
@@ -94,10 +95,80 @@ func testGetBlockHash(r *rpctest.Harness, t *testing.T) {
...
+		t.Fatalf("unable to delete mining address: %v", err)
+	}
+	if err := r.Node.DelMiningAddr(addrs[0]); err == nil {
+		t.Fatalf("deletion of non-existant address should fail")
```
Roasbeef: Fixed.

rpcserverhelp.go
Roasbeef: Fixed.

rpcserverhelp.go
Roasbeef: Fixed.

rpcserverhelp.go
Roasbeef: Fixed.

Roasbeef:
Yep, after this PR portions of the initialization of the `rpctest.Harness` can also be re-written to utilize the RPC call rather than manually setting the flag on the command line.

davecgh:
So, I believe I know why you changed the `preCfg := cfg` copy to `preCfg := &cfg` originally which is the following vet error:
```
config.go:377::error: assignment copies lock value to preCfg: main.config contains sync.RWMutex (vet)
```

Unfortunately, as we've already discussed, it can't be a pointer because of the double parse.  Looks like the mutex is going to have to be refactored out of the config struct.  That could be a blessing as I suspect it will make the code a little cleaner anyways since the mutex won't have to be littered all throughout the code if it's encapsulated in a concurrent safe type.

Roasbeef:
I've just pushed a new version of the commit with the addressed review items squashed into place. I've also modified the implementation of the RPC's to utilize a new concurrent-safe struct which houses the mining address. As a result, the implementation is much cleaner as the mutex accesses are no longer strewn across the RPC calls. However, I wasn't quite sure where to put the new struct, so atm it lives in `server.go`.

docs/json_rpc_api.md
```diff
@@ -664,6 +667,44 @@ The following is an overview of the RPC methods which are implemented by btcd, b
...
+|---|---|
+|Method|delminingaddr|
+|Parameters|1. address (string, required) - The address to delete from the list of mining addresses|
+|Description|Attempts to remove the passed address from the list of mining addresses. Attempting to delete a non-existent address results in an error.|
+|Returns|Nothing|
+[Return to Overview](#MethodOverview)<br />
+
+***
+
+***
```
davecgh: Duplicate separator and extra blank line.

docs/json_rpc_api.md
```diff
@@ -664,6 +667,44 @@ The following is an overview of the RPC methods which are implemented by btcd, b
...
+
+<a name="listminingaddrs"/>
+
+|   |   |
+|---|---|
+|Method|listminingaddrs|
+|Parameters|None|
+|Description|Returns a list of all the currently active generation (coinbase payout) addresses.|
+|Returns|Nothing|
```
davecgh: It returns the list of addresses, so this is not correct.

davecgh (changes requested):

davecgh:
Nice job overall as usual!  Also, this needs a rebase to correct the conflicts.

davecgh:
Now the segwit has been merged, would you please rebase and update this to take care of the final review comments?  Thanks @Roasbeef!

jakesyl:
@jcvernaleo (as per #1530)

- Low priority
- Enhancement 

Should be closed immediately in favor of #1441

yashbhutwala:
I have rebased off of the latest master and updated the approach on #1441.  Please re-review, hoping to get it merged 🙏

allocz (commented):
Looks like this PR is abandoned. Also, I prefer the approach proposed in #1441. I think at least one of the two PR's should be closed.

PR #847, opened 2016-11-08
peer: No error log on unexpected EOF.

davecgh:
This modifies the error handling path in the peer read loop such that it will no longer log an error when the error is `io.ErrUnexpectedEOF`.  This is being done because that error is almost always the result of a peer
being remotely disconnected and thus it isn't useful information to log.

However, since it might actually be due to a malformed message, a reject message is still queued up to be sent back to the peer (which will simply be discarded if the peer disconnected) before it is disconnected.

While it would be ideal to only log if it's not due to a disconnect, and the code already attempts to handle that, it's not 100% possible to detect upon the read returning an error without attempting to read again
which will not happen until the next loop iteration.


alexlyp:
OK

PR #848, opened 2016-11-08
Add support for defining additional checkpoints

stevenroose:
Adds a config option and cmd line flag `--checkpoint` that allows for defining custom checkpoints that will be used for verifying the active chain.

It also changes the way `--nocheckpoints` is handled: instead of keeping a flag to indicate checkpoint support, the built-in checkpoints and possible custom checkpoints are copied into the blockchain struct. This allows for overwriting the built-in ones as well.

I did not add tests for testing the flag, I wouldn't know where to start, but I can give it a shot if it's preferred!

### DIFF
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -158,6 +158,7 @@ type BlockChain struct {
 	// The following fields are set when the instance is created and can't
 	// be changed afterwards, so there is no need to protect them with a
 	// separate mutex.
+	checkpoints         []chaincfg.Checkpoint
 	checkpointsByHeight map[int32]*chaincfg.Checkpoint
 	db                  database.DB
 	chainParams         *chaincfg.Params
@@ -188,8 +189,7 @@ type BlockChain struct {
 
 	// These fields are configuration parameters that can be toggled at
 	// runtime.  They are protected by the chain lock.
-	noVerify      bool
-	noCheckpoints bool
+	noVerify bool
 
 	// These fields are related to the memory block index.  They are
 	// protected by the chain lock.
@@ -1569,7 +1569,7 @@ func (b *BlockChain) connectBestChain(node *blockNode, block *btcutil.Block, fla
 func (b *BlockChain) isCurrent() bool {
 	// Not current if the latest main (best) chain height is before the
 	// latest known good checkpoint (when checkpoints are enabled).
-	checkpoint := b.latestCheckpoint()
+	checkpoint := b.LatestCheckpoint()
 	if checkpoint != nil && b.bestNode.height < checkpoint.Height {
 		return false
 	}
@@ -1640,6 +1640,12 @@ type Config struct {
 	// This field is required.
 	ChainParams *chaincfg.Params
 
+	// Checkpoints hold caller-defined checkpoints that should be added to the
+	// default checkpoints in ChainParams. Checkpoints must be sorted by height.
+	//
+	// This field can be nil if the caller did not specify any checkpoints.
+	Checkpoints []chaincfg.Checkpoint
+
 	// TimeSource defines the median time source to use for things such as
 	// block processing and determining whether or not the chain is current.
 	//
@@ -1688,20 +1694,21 @@ func New(config *Config) (*BlockChain, error) {
 	}
 
 	// Generate a checkpoint by height map from the provided checkpoints.
-	params := config.ChainParams
 	var checkpointsByHeight map[int32]*chaincfg.Checkpoint
-	if len(params.Checkpoints) > 0 {
+	if len(config.Checkpoints) > 0 {
 		checkpointsByHeight = make(map[int32]*chaincfg.Checkpoint)
-		for i := range params.Checkpoints {
-			checkpoint := &params.Checkpoints[i]
+		for i := range config.Checkpoints {
+			checkpoint := &config.Checkpoints[i]
 			checkpointsByHeight[checkpoint.Height] = checkpoint
 		}
 	}
 
+	params := config.ChainParams
 	targetTimespan := int64(params.TargetTimespan)
 	targetTimePerBlock := int64(params.TargetTimePerBlock)
 	adjustmentFactor := params.RetargetAdjustmentFactor
 	b := BlockChain{
+		checkpoints:         config.Checkpoints,
 		checkpointsByHeight: checkpointsByHeight,
 		db:                  config.DB,
 		chainParams:         params,
--- a/blockchain/chain_test.go
+++ b/blockchain/chain_test.go
@@ -43,9 +43,7 @@ func TestHaveBlock(t *testing.T) {
 	}
 	defer teardownFunc()
 
-	// Since we're not dealing with the real block chain, disable
-	// checkpoints and set the coinbase maturity to 1.
-	chain.DisableCheckpoints(true)
+	// Since we're not dealing with the real block chain, set the coinbase maturity to 1.
 	chain.TstSetCoinbaseMaturity(1)
 
 	for i := 1; i < len(blocks); i++ {
@@ -131,9 +129,7 @@ func TestCalcSequenceLock(t *testing.T) {
 	}
 	defer teardownFunc()
 
-	// Since we're not dealing with the real block chain, disable
-	// checkpoints and set the coinbase maturity to 1.
-	chain.DisableCheckpoints(true)
+	// Since we're not dealing with the real block chain, set the coinbase maturity to 1.
 	chain.TstSetCoinbaseMaturity(1)
 
 	// Load all the blocks into our test chain.
--- a/blockchain/checkpoints.go
+++ b/blockchain/checkpoints.go
@@ -27,66 +27,39 @@ func newHashFromStr(hexStr string) *chainhash.Hash {
 	return hash
 }
 
-// DisableCheckpoints provides a mechanism to disable validation against
-// checkpoints which you DO NOT want to do in production.  It is provided only
-// for debug purposes.
-//
-// This function is safe for concurrent access.
-func (b *BlockChain) DisableCheckpoints(disable bool) {
-	b.chainLock.Lock()
-	b.noCheckpoints = disable
-	b.chainLock.Unlock()
-}
-
 // Checkpoints returns a slice of checkpoints (regardless of whether they are
-// already known).  When checkpoints are disabled or there are no checkpoints
-// for the active network, it will return nil.
+// already known).
+// When there are no checkpoints for the chain, it will return nil.
 //
 // This function is safe for concurrent access.
 func (b *BlockChain) Checkpoints() []chaincfg.Checkpoint {
-	b.chainLock.RLock()
-	defer b.chainLock.RUnlock()
-
-	if b.noCheckpoints || len(b.chainParams.Checkpoints) == 0 {
-		return nil
-	}
-
-	return b.chainParams.Checkpoints
+	return b.checkpoints
 }
 
-// latestCheckpoint returns the most recent checkpoint (regardless of whether it
-// is already known).  When checkpoints are disabled or there are no checkpoints
-// for the active network, it will return nil.
+// HasCheckpoints returns whether this BlockChain has checkpoints defined.
 //
-// This function MUST be called with the chain state lock held (for reads).
-func (b *BlockChain) latestCheckpoint() *chaincfg.Checkpoint {
-	if b.noCheckpoints || len(b.chainParams.Checkpoints) == 0 {
-		return nil
-	}
-
-	checkpoints := b.chainParams.Checkpoints
-	return &checkpoints[len(checkpoints)-1]
+// This function is safe for concurrent access.
+func (b *BlockChain) HasCheckpoints() bool {
+	return len(b.checkpoints) > 0
 }
 
-// LatestCheckpoint returns the most recent checkpoint (regardless of whether it
-// is already known).  When checkpoints are disabled or there are no checkpoints
-// for the active network, it will return nil.
+// LatestCheckpoint returns the most recent checkpoint (regardless of whether they
+// are already known). When there are no defined checkpoints for the active chain
+// instance, it will return nil.
 //
 // This function is safe for concurrent access.
 func (b *BlockChain) LatestCheckpoint() *chaincfg.Checkpoint {
-	b.chainLock.RLock()
-	checkpoint := b.latestCheckpoint()
-	b.chainLock.RUnlock()
-	return checkpoint
+	if !b.HasCheckpoints() {
+		return nil
+	}
+	return &b.checkpoints[len(b.checkpoints)-1]
 }
 
 // verifyCheckpoint returns whether the passed block height and hash combination
-// match the hard-coded checkpoint data.  It also returns true if there is no
-// checkpoint data for the passed block height.
-//
-// This function MUST be called with the chain lock held (for reads).
+// match the checkpoint data. It also returns true if there is no checkpoint
+// data for the passed block height.
 func (b *BlockChain) verifyCheckpoint(height int32, hash *chainhash.Hash) bool {
-	if b.noCheckpoints || len(b.chainParams.Checkpoints) == 0 {
+	if !b.HasCheckpoints() {
 		return true
 	}
 
@@ -112,14 +85,14 @@ func (b *BlockChain) verifyCheckpoint(height int32, hash *chainhash.Hash) bool {
 //
 // This function MUST be called with the chain lock held (for reads).
 func (b *BlockChain) findPreviousCheckpoint() (*btcutil.Block, error) {
-	if b.noCheckpoints || len(b.chainParams.Checkpoints) == 0 {
+	if !b.HasCheckpoints() {
 		return nil, nil
 	}
 
-	// No checkpoints.
-	checkpoints := b.chainParams.Checkpoints
+	checkpoints := b.checkpoints
 	numCheckpoints := len(checkpoints)
 	if numCheckpoints == 0 {
+		// No checkpoints.
 		return nil, nil
 	}
 
@@ -260,11 +233,6 @@ func isNonstandardTransaction(tx *btcutil.Tx) bool {
 func (b *BlockChain) IsCheckpointCandidate(block *btcutil.Block) (bool, error) {
 	b.chainLock.RLock()
 	defer b.chainLock.RUnlock()
-
-	// Checkpoints must be enabled.
-	if b.noCheckpoints {
-		return false, fmt.Errorf("checkpoints are disabled")
-	}
 
 	var isCandidate bool
 	err := b.db.View(func(dbTx database.Tx) error {
--- a/blockchain/reorganization_test.go
+++ b/blockchain/reorganization_test.go
@@ -52,9 +52,7 @@ func TestReorganization(t *testing.T) {
 	}
 	defer teardownFunc()
 
-	// Since we're not dealing with the real block chain, disable
-	// checkpoints and set the coinbase maturity to 1.
-	chain.DisableCheckpoints(true)
+	// Since we're not dealing with the real block chain set the coinbase maturity to 1.
 	chain.TstSetCoinbaseMaturity(1)
 
 	expectedOrphans := map[int]struct{}{5: {}, 6: {}}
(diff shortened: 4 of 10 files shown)


stevenroose:
> Rather than passing in the additional checkpoints, I would much prefer the caller specify all checkpoints.

I agree with this because in an ideal situation, only the latest checkpoint should be relevant. However, someone told me on IRC (don't remember who) that for some reason intermediate checkpoints would still improve sync time. If the latter is still true, that would mean that a user that wants to use an additional checkpoint to improve sync time, should also cherry pick other checkpoints in the past, which is somewhat cumbersome.

sample-btcd.conf
```diff
@@ -153,6 +153,11 @@
 ; Disable peer bloom filtering.  See BIP0111.
 ; nopeerbloomfilters=1
 
+; Add a custom checkpoints. Format: '<height>:<hash>'
+; checkpoint=<height>:<hash>
+; Disable built-in checkpoints. Don't do this unless you know what you're doing.
+; nocheckpoints=1
```
stevenroose: That makes sense

davecgh:
@stevenroose Correct, you do want the intermediate checkpoints as well.  That is not prevented by passing them in however.  You would just concatenate them in server prior to creating the chain so they are passed in.

What I'm suggesting is to create the final concatenated list of checkpoints in `newBlockManager` just before the call to `blockchain.New`.  I would also allow the additional checkpoints to override for testing purposes.

Something like the following:

```diff
diff --git a/blockmanager.go b/blockmanager.go
index c85e64d..d7122f6 100644
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -9,6 +9,7 @@ import (
        "net"
        "os"
        "path/filepath"
+       "sort"
        "sync"
        "sync/atomic"
        "time"
@@ -1324,6 +1325,40 @@ func (b *blockManager) Pause() chan<- struct{} {
        return c
 }

+// checkpointSorter implements sort.Interface to allow a slice of checkpoints to
+// be sorted.
+type checkpointSorter []chaincfg.Checkpoint
+
+// Len returns the number of checkpoints in the slice.  It is part of the
+// sort.Interface implementation.
+func (s checkpointSorter) Len() int {
+       return len(s)
+}
+
+// Swap swaps the checkpoints at the passed indices.  It is part of the
+// sort.Interface implementation.
+func (s checkpointSorter) Swap(i, j int) {
+       s[i], s[j] = s[j], s[i]
+}
+
+// Less returns whether the checkpoint with index i should sort before the
+// checkpoint with index j.  It is part of the sort.Interface implementation.
+func (s checkpointSorter) Less(i, j int) bool {
+       return s[i].Height < s[j].Height
+}
+
+// mergeCheckpoints returns two slices of checkpoints merged into one slice
+// such that the checkpoints are sorted by height.  In the case the additional
+// checkpoints contain a checkpoint with the same height as a checkpoint in the
+// default checkpoints, the additional checkpoint will take precedence and
+// overwrite the default one.
+func mergeCheckpoints(defaultCheckpoints, additional []chaincfg.Checkpoint) []chaincfg.Checkpoint {
+       // Create a map of the additional checkpoint heights to detect
+       // duplicates.
+       additionalHeights := make(map[int32]struct{})
+       for _, checkpoint := range additional {
+               additionalHeights[checkpoint.Height] = struct{}{}
+       }
+
+       // Add all default checkpoints that do not have an override in the
+       // additional checkpoints.
+       numDefault := len(defaultCheckpoints)
+       checkpoints := make([]chaincfg.Checkpoint, 0, numDefault+len(additional))
+       for _, checkpoint := range defaultCheckpoints {
+               if _, exists := additionalHeights[checkpoint.Height]; !exists {
+                       checkpoints = append(checkpoints, checkpoint)
+               }
+       }
+
+       // Append the additional checkpoints and return the sorted results.
+       checkpoints = append(checkpoints, additional...)
+       sort.Sort(checkpointSorter(checkpoints))
+       return checkpoints
+}
+
 // newBlockManager returns a new bitcoin block manager.
 // Use Start to begin processing asynchronous block and inv updates.
 func newBlockManager(s *server, indexManager blockchain.IndexManager) (*blockManager, error) {
@@ -1341,8 +1376,10 @@ func newBlockManager(s *server, indexManager blockchain.IndexManager) (*blockMan
        // Create a new block chain instance with the appropriate configuration.
        var err error
        bm.chain, err = blockchain.New(&blockchain.Config{
-               DB:            s.db,
-               ChainParams:   s.chainParams,
+               DB:          s.db,
+               ChainParams: s.chainParams,
+               Checkpoints: mergeCheckpoints(s.chainParams.Checkpoints,
+                       cfg.checkpoints),
                TimeSource:    s.timeSource,
                Notifications: bm.handleNotifyMsg,
                SigCache:      s.sigCache,
```

With that, you can change your `parseAndSortCheckpoints` to just parse (since they will all be sorted together during the merge step) as well as removing all of the extra logic in `blockchain.New`.  I suspect the parse function should error if the user specifies duplicates (within the user-supplied checkpoints themselves).

Finally, notice how this uses `[]chaincfg.Checkpoint` instead of `[]*chaincfg.Checkpoint` as previously mentioned.

stevenroose:
@davecgh Something like this?

blockchain/chain_test.go
```diff
@@ -43,9 +43,9 @@ func TestHaveBlock(t *testing.T) {
 	}
 	defer teardownFunc()
 
-	// Since we're not dealing with the real block chain, disable
+	// Since we're not dealing with the real block chain, remove
 	// checkpoints and set the coinbase maturity to 1.
-	chain.DisableCheckpoints(true)
+	chain.RemoveCheckpoints()
```
davecgh: This should be removed as mentioned in my first bullet point in the original review. Allowing the checkpoints to be modified after the fact means the entire checkpoints slice would need to be protected by a mutex. Part of the benefit of passing the checkpoints in an init time is the fact you can then treat them as immutable and then you don't need a mutex.

Instead, they should be set to nil during the initial setup via `chainSetup` (which incidentally they already are because you forgot to modify the call to `blockchain.New` in `common_test.go` to add the field and nil is the zero value for slices).  I would prefer that be made more explicit like so:

```diff
diff --git a/blockchain/common_test.go b/blockchain/common_test.go
index aec85bf..7ba3d7c 100644
--- a/blockchain/common_test.go
+++ b/blockchain/common_test.go
@@ -116,6 +116,7 @@ func chainSetup(dbName string, params *chaincfg.Params) (*blockchain.BlockChain,
        chain, err := blockchain.New(&blockchain.Config{
                DB:          db,
                ChainParams: &paramsCopy,
+               Checkpoints: nil,
                TimeSource:  blockchain.NewMedianTime(),
                SigCache:    txscript.NewSigCache(1000),
        })
```

Then the various callsites of `RemoveCheckpoints` should be removed like so:

```diff
diff --git a/blockchain/chain_test.go b/blockchain/chain_test.go
index 45732d6..860bdbd 100644
--- a/blockchain/chain_test.go
+++ b/blockchain/chain_test.go
@@ -43,9 +43,8 @@ func TestHaveBlock(t *testing.T) {
        }
        defer teardownFunc()

-       // Since we're not dealing with the real block chain, remove
-       // checkpoints and set the coinbase maturity to 1.
-       chain.RemoveCheckpoints()
+       // Since we're not dealing with the real block chain, set the coinbase
+       // maturity to 1.
        chain.TstSetCoinbaseMaturity(1)
```

blockchain/chain_test.go
```diff
@@ -131,9 +131,9 @@ func TestCalcSequenceLock(t *testing.T) {
...
```
davecgh: As mentioned elsewhere, should be removed in favor of initial setup with no checkpoints.

blockchain/checkpoints.go
```diff
@@ -27,31 +27,33 @@ func newHashFromStr(hexStr string) *chainhash.Hash {
...
 // This function is safe for concurrent access.
 func (b *BlockChain) Checkpoints() []chaincfg.Checkpoint {
 	b.chainLock.RLock()
-	defer b.chainLock.RUnlock()
-
-	if b.noCheckpoints || len(b.chainParams.Checkpoints) == 0 {
-		return nil
-	}
+	checkpoints := b.checkpoints
+	b.chainLock.RUnlock()
```
davecgh: No need for lock because `b.checkpoints` is being made immutable.

```Go
return b.checkpoints
```

blockchain/checkpoints.go
```diff
@@ -27,31 +27,33 @@ func newHashFromStr(hexStr string) *chainhash.Hash {
...
+	return checkpoints
+}
 
-	return b.chainParams.Checkpoints
+// HasCheckpoints returns whether this BlockChain has checkpoints defined.
```
davecgh: Comment doesn't match function name.

blockchain/checkpoints.go
```diff
@@ -27,31 +27,33 @@ func newHashFromStr(hexStr string) *chainhash.Hash {
...
-	b.noCheckpoints = disable
+	b.checkpoints = make([]chaincfg.Checkpoint, 0)
+	b.checkpointsByHeight = make(map[int32]*chaincfg.Checkpoint)
 	b.chainLock.Unlock()
 }
 
-// Checkpoints returns a slice of checkpoints (regardless of whether they are
-// already known).  When checkpoints are disabled or there are no checkpoints
-// for the active network, it will return nil.
+// Checkpoints returns a slice of checkpoints.
```
davecgh: Why was the parenthesized bit of the comment changed?  That information is important for callers to know.

blockchain/checkpoints.go
```diff
@@ -27,31 +27,33 @@ func newHashFromStr(hexStr string) *chainhash.Hash {
...
-	}
+	checkpoints := b.checkpoints
+	b.chainLock.RUnlock()
+	return checkpoints
+}
 
-	return b.chainParams.Checkpoints
+// HasCheckpoints returns whether this BlockChain has checkpoints defined.
+//
+// This function MUST be called with the chain state lock held (for reads).
```
davecgh: Not true since `b.checkpoints` is being made immutable.

blockchain/checkpoints.go
```diff
@@ -60,12 +62,10 @@ func (b *BlockChain) Checkpoints() []chaincfg.Checkpoint {
 //
 // This function MUST be called with the chain state lock held (for reads).
```
davecgh: No longer true since `b.checkpoints` is being made immutable.

blockchain/checkpoints.go
```diff
@@ -86,7 +86,7 @@ func (b *BlockChain) LatestCheckpoint() *chaincfg.Checkpoint {
 //
 // This function MUST be called with the chain lock held (for reads).
```
davecgh: No longer true since `b.checkpoints` is being made immutable.

blockmanager.go
```diff
@@ -179,10 +180,7 @@ func (b *blockManager) resetHeaderState(newestHash *chainhash.Hash, newestHeight
 func (b *blockManager) findNextHeaderCheckpoint(height int32) *chaincfg.Checkpoint {
 	// There is no next checkpoint if checkpoints are disabled or there are
 	// none for this current network.
-	if cfg.DisableCheckpoints {
-		return nil
-	}
-	checkpoints := b.server.chainParams.Checkpoints
+	checkpoints := b.chain.Checkpoints()
```
davecgh: I think the check for disable should be kept.  You're making assumptions about the state of chain's checkpoints here that could change in a refactor.  The code is more robust with the check of `cfg.DisableCheckpoints ` (and it keeps the comment accurate).

blockmanager.go
```diff
@@ -1337,12 +1386,17 @@ func newBlockManager(s *server, indexManager blockchain.IndexManager) (*blockMan
 		headerList:      list.New(),
 		quit:            make(chan struct{}),
 	}
-
+	// merge given checkpoints with the default ones unless they are disabled
```
davecgh: Separate this block of code with blank lines for consistency please.

```Go
...
}

// Merge given checkpoints with the default ones unless they are disabled.
var checkpoints []chaincfg.Checkpoint
if !cfg.DisableCheckpoints {
	checkpoints = mergeCheckpoints(s.chainParams.Checkpoints, cfg.checkpoints)
}

// Create a new block chain instance with the appropriate configuration.
...
```

Also, notice I made the comment style consistent.

config.go
```diff
@@ -304,6 +308,43 @@ func normalizeAddresses(addrs []string, defaultPort string) []string {
...
+		if err != nil {
+			return nil, err
+		}
+		checkpoints[i] = checkpoint
+	}
+	return checkpoints, nil
+}
+
+// newCheckpointFromStr parses checkpoints in the '<height>:<hash>' format.
+func newCheckpointFromStr(checkpoint string) (chaincfg.Checkpoint, error) {
```
davecgh: Please define the function before you use it.  I know the compiler doesn't care, but the vast majority of the code base aims to be consistent in this regard although I know a few cases otherwise have snuck in.

config.go
```diff
@@ -304,6 +308,43 @@ func normalizeAddresses(addrs []string, defaultPort string) []string {
...
+	parts := strings.Split(checkpoint, ":")
+	if len(parts) != 2 {
+		return chaincfg.Checkpoint{}, errors.New("Checkpoints must use the syntax '<height>:<hash>'")
+	}
+	height, err := strconv.ParseInt(parts[0], 10, 32)
```
davecgh: Blank line before each logical block, please.

config.go
```diff
@@ -304,6 +308,43 @@ func normalizeAddresses(addrs []string, defaultPort string) []string {
...
+	if err != nil {
+		return chaincfg.Checkpoint{}, errors.New("Checkpoints must use the syntax '<height>:<hash>'")
+	}
+	hash, err := chainhash.NewHashFromStr(parts[1])
```
davecgh: Blank line before each logical block, please.

config.go
```diff
@@ -304,6 +308,43 @@ func normalizeAddresses(addrs []string, defaultPort string) []string {
...
+	if err != nil {
+		return chaincfg.Checkpoint{}, errors.New("Checkpoints must use the syntax '<height>:<hash>'")
+	}
+	return chaincfg.Checkpoint{
```
davecgh: Blank line before each logical block, please.

config.go
```diff
@@ -818,6 +859,12 @@ func loadConfig() (*config, []string, error) {
 	cfg.ConnectPeers = normalizeAddresses(cfg.ConnectPeers,
 		activeNetParams.DefaultPort)
 
+	// Check the checkpoints for syntax errors
```
davecgh: Missing period.

blockchain/reorganization_test.go
```diff
@@ -52,9 +52,9 @@ func TestReorganization(t *testing.T) {
 	}
 	defer teardownFunc()
 
-	// Since we're not dealing with the real block chain, disable
+	// Since we're not dealing with the real block chain, remove
 	// checkpoints and set the coinbase maturity to 1.
-	chain.DisableCheckpoints(true)
+	chain.RemoveCheckpoints()
```
stevenroose: Oh yeah, that should be possible if the tests don't want to change that mid-testing, I'll take a look

blockchain/checkpoints.go
stevenroose: Oh, I interpreted that line incorrectly!

blockchain/checkpoints.go
stevenroose: Yeah I know, forgot to update the comments, on it.

stevenroose:
Thanks for the in-depth review. Although I have the feeling you put more effort in this than I do :) Will fix all comments!

blockchain/checkpoints.go
stevenroose: You prefer to have the explicit `This function is safe for concurrent access.` on fields that are treated as immutable?

blockchain/chain.go
```diff
@@ -1684,21 +1691,19 @@ func New(config *Config) (*BlockChain, error) {
...
-		checkpointsByHeight = make(map[int32]*chaincfg.Checkpoint)
-		for i := range params.Checkpoints {
-			checkpoint := &params.Checkpoints[i]
-			checkpointsByHeight[checkpoint.Height] = checkpoint
-		}
+
+	// Create checkpoint variables. Leave empty when disabled.
+	checkpointsByHeight := make(map[int32]*chaincfg.Checkpoint)
+	for _, checkpoint := range config.Checkpoints {
+		checkpointsByHeight[checkpoint.Height] = &checkpoint
```
stevenroose: Hmm, that's interesting, I did not know that!

blockchain/checkpoints.go
davecgh: Yes, the caller otherwise has no way of knowing the concurrency safety characteristics (or lack thereof) of the function.  How the function goes about achieving that, whether it be via a mutex, a channel, CAS, or use of immutable data structures doesn't change that.

davecgh:
Sure thing.  It's true that I always try to do very thorough reviews including manual testing and verification of all code changes and paths (unless I'm positive all paths are already covered extremely well by the tests).  In the case of checkpoints, unfortunately there currently isn't very good test coverage (which is something that @jcvernaleo is working on for a downstream project that will likely get back ported), so it's important to be extremely careful when messing with them.

The intense level of review is why it usually takes me a while to get through them.

blockchain/chain.go
```diff
@@ -1640,6 +1640,12 @@ type Config struct {
 	// This field is required.
 	ChainParams *chaincfg.Params
 
+	// Checkpoints hold caller-defined checkpoints that should be added to the
+	// default checkpoints in ChainParams. Checkpoints must be sorted by height.
+	//
+	// This field can be nil is the caller did not specify any checkpoints.
```
davecgh: s/is the/if the/

blockchain/chain.go
```diff
@@ -1687,21 +1693,22 @@ func New(config *Config) (*BlockChain, error) {
 		return nil, AssertError("blockchain.New timesource is nil")
 	}
 
-	// Generate a checkpoint by height map from the provided checkpoints.
-	params := config.ChainParams
+	// Create checkpoint variables. Leave empty when disabled.
```
davecgh: Any reason this comment was changed?  It seems like it more clearly stated the intention before.  "Create checkpoint variables" doesn't actually tell me what the intent is in my opinion.

config.go
```diff
@@ -123,6 +125,7 @@ type config struct {
 	TestNet3             bool          `long:"testnet" description:"Use the test network"`
 	RegressionTest       bool          `long:"regtest" description:"Use the regression test network"`
 	SimNet               bool          `long:"simnet" description:"Use the simulation test network"`
+	Checkpoints          []string      `long:"checkpoint" description:"Add a custom checkpoint. Format: '<height>:<hash>'"`
```
davecgh: I wonder if the option should be `--addcheckpoint` so it is more consistent with `addpeer` and also makes it a little more obvious that it is adding one or more checkpoints as opposed to replacing them?

stevenroose: Makes sense.

blockchain/chain.go
stevenroose: Yeah I changed it first to not only make the map there but also the slice. Forgot to change the comment back I guess.

stevenroose:
Should be fine now

davecgh:
Still has a conflict with ` config.go`.  Needs a rebase to the latest master.

config.go
```diff
@@ -796,6 +843,12 @@ func loadConfig() (*config, []string, error) {
 	cfg.ConnectPeers = normalizeAddresses(cfg.ConnectPeers,
 		activeNetParams.DefaultPort)
 
+	// Check the checkpoints for syntax errors.
+	cfg.addCheckpoints, err = parseCheckpoints(cfg.AddCheckpoints)
+	if err != nil {
+		return nil, nil, err
```
davecgh: This will not print anything to the console.  Returning an error just lets btcd know to exist with a non-zero exit code.  You'll need to follow what the other errors in here do:

```Go
if err != nil {
 		err := fmt.Errorf("%s: %s ", funcName, err)
 		fmt.Fprintln(os.Stderr, err)
 		return nil, nil, err
```

stevenroose:
Did not notice that. Updated

PR #854, opened 2016-11-14
btcjson: add ErrRPCClientNodeNotAdded

dajohi:


stevenroose:
What are these values based on? Could use a line of documentation :p

dajohi:
https://github.com/bitcoin/bitcoin/blob/master/src/rpc/protocol.h

PR #856, opened 2016-11-16
txscript: reduce allocs in calcSignatureHash

dajohi:


txscript/script.go
```diff
@@ -356,6 +356,7 @@ func calcSignatureHash(script []parsedOpcode, hashType SigHashType, tx *wire.Msg
 	// transaction and the hash type (encoded as a 4-byte little-endian
 	// value) appended.
 	var wbuf bytes.Buffer
+	wbuf.Grow(txCopy.SerializeSize() + 4)
```
davecgh: In order to be more consistent with the rest of the codebase, I would suggest:
```
wbuf := bytes.NewBuffer(make([]byte, 0, txCopy.SerializeSize() + 4))
```

dajohi:
updated.

PR #857, opened 2016-11-16
wire: reduce allocs in BlockHash()

dajohi:


jrick:
This is fine but there are other places in the codebase that do

```
buf := bytes.NewBuffer(make([]byte, 0, size))
```

PR #859, opened 2016-11-19
btcd: only allow one getaddr request per peer

dajohi:


dajohi:
See https://github.com/bitcoin/bitcoin/commit/66b07247a7a9e48e082502338176cc06edf61474

dajohi:
Rebased.

PR #860, opened 2016-11-19
btcd: disable mempool requests if bloom filtering is disabled

dajohi:


dajohi:
See https://github.com/bitcoin/bitcoin/commit/beceac9bbf14bf4a81f6f63b9cca2a64157054ae

stevenroose:
After #769 is merged, we could also allow whitelisted peers to do this.

PR #863, opened 2016-11-21
addrmgr: Remove unused param from GetAddress()

dskloet:
addrmgr.GetAddress() had a parameter `class string` originally intended
to support looking up addresses according to some type of filter such as
IPv4, IPv6, and only those which support specific wire.ServiceFlags
(full nodes, nodes that support bloom filters, nodes that support
segwit, etc). But currently the parameter is unused and also has an
inappropriate type `string`.

If it would ever be used, it's easy to add back and should then get an
appropriate type such as something that allows bitflags to be set so
that the caller could request combinations such as peers that support
IPv6, are full nodes, and support bloom filters.

### DIFF
--- a/addrmgr/addrmanager.go
+++ b/addrmgr/addrmanager.go
@@ -739,7 +739,7 @@ func NetAddressKey(na *wire.NetAddress) string {
 // random one from the possible addresses with preference given to ones that
 // have not been used recently and should not pick 'close' addresses
 // consecutively.
-func (a *AddrManager) GetAddress(class string) *KnownAddress {
+func (a *AddrManager) GetAddress() *KnownAddress {
 	// Protect concurrent access.
 	a.mtx.Lock()
 	defer a.mtx.Unlock()
--- a/addrmgr/addrmanager_test.go
+++ b/addrmgr/addrmanager_test.go
@@ -221,7 +221,7 @@ func TestAttempt(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Adding address failed: %v", err)
 	}
-	ka := n.GetAddress("any")
+	ka := n.GetAddress()
 
 	if !ka.LastAttempt().IsZero() {
 		t.Errorf("Address should not have attempts, but does")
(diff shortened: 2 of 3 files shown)


dskloet:
Sure, but should the prefix be addr (as you ask) or addrmgr (as you did in
the PR title)?

(Will do tomorrow as it's bed time in Europe)

davecgh:
It should be `addrmgr`.  I edited my initial comment right after, but I'm sure github had already emailed it.  Thanks.

dskloet:
Thanks. Done.

dskloet:
Weird. In my branch, the commit message has
> Author: David de Kloet <dskloet@google.com>
but when I pull from btcsuite, the commit message has
> Author: dskloet <dskloet@gmail.com>

Any idea how this could have changed?

My other commit doesn't have the same difference.

davecgh:
I used the github rebase feature.  I suspect it must have been that?

dskloet:
Huh? I check and it did fit in the 50 characters with "parameter".

    addrmgr: Remove unused parameter from GetAddress()
    01234567890123456789012345678901234567890123456789

I guess github changed the author based on this account when you changed the commit message. Somehow github also thinks that the commits belong to my other github account dskloetg, based on email address, I guess.

davecgh:
Hmm, I see that you're right.  I originally added the prefix and noticed it was over 50, but didn't remove the `addrmgr.` part and that's why.  It was fine as it was.  Sorry about that.

dskloet:
I'd like to report a bug to github. Can you explain what you did to rebase on github so I can make an accurate report? I'm assuming you didn't change the author of the commit manually?

davecgh:
Correct.

I used the "Squash and merge" option which allows you to effectively perform a `git rebase -i` in order to combine commits and/or modify the commit message(s).  When you select that option, it provides two text fields.  One is for the commit title (summary), and the other is for the commit body.  The github interface does not provide any options to change anything else about the commit such as the author's email.

dskloet:
Thanks. I got an answer. When you use "Squash and merge" on a pull request, the result is attributed to the github account that made the pull request. I mentioned that it was only a single commit and the you only changed the commit message so that it wouldn't make sense to change the Author field. She said she would pass it on to the team.

PR #867, opened 2016-11-24
wire: Treat NetAddress more like immutable

dskloet:
Replace assignments to individual fields of wire.NetAddress with
creating the entire object at once, as one would do if the type was
immutable.

### DIFF
--- a/connmgr/seed.go
+++ b/connmgr/seed.go
@@ -60,14 +60,13 @@ func SeedFromDNS(chainParams *chaincfg.Params, reqServices wire.ServiceFlag,
 			// if this errors then we have *real* problems
 			intPort, _ := strconv.Atoi(chainParams.DefaultPort)
 			for i, peer := range seedpeers {
-				addresses[i] = new(wire.NetAddress)
-				addresses[i].SetAddress(peer, uint16(intPort))
-				// bitcoind seeds with addresses from
-				// a time randomly selected between 3
-				// and 7 days ago.
-				addresses[i].Timestamp = time.Now().Add(-1 *
-					time.Second * time.Duration(secondsIn3Days+
-					randSource.Int31n(secondsIn4Days)))
+				addresses[i] = wire.NewNetAddressTimestamp(
+					// bitcoind seeds with addresses from
+					// a time randomly selected between 3
+					// and 7 days ago.
+					time.Now().Add(-1*time.Second*time.Duration(secondsIn3Days+
+						randSource.Int31n(secondsIn4Days))),
+					0, peer, uint16(intPort))
 			}
 
 			seedFn(addresses)
--- a/wire/netaddress.go
+++ b/wire/netaddress.go
@@ -57,20 +57,21 @@ func (na *NetAddress) AddService(service ServiceFlag) {
 	na.Services |= service
 }
 
-// SetAddress is a convenience function to set the IP address and port in one
-// call.
-func (na *NetAddress) SetAddress(ip net.IP, port uint16) {
-	na.IP = ip
-	na.Port = port
-}
-
 // NewNetAddressIPPort returns a new NetAddress using the provided IP, port, and
 // supported services with defaults for the remaining fields.
 func NewNetAddressIPPort(ip net.IP, port uint16, services ServiceFlag) *NetAddress {
+	return NewNetAddressTimestamp(time.Now(), services, ip, port)
+}
+
+// NewNetAddressTimestamp returns a new NetAddress using the provided
+// timestamp, IP, port, and supported services. The timestamp is rounded to
+// single second precision.
+func NewNetAddressTimestamp(
+	timestamp time.Time, services ServiceFlag, ip net.IP, port uint16) *NetAddress {
 	// Limit the timestamp to one second precision since the protocol
 	// doesn't support better.
 	na := NetAddress{
-		Timestamp: time.Unix(time.Now().Unix(), 0),
+		Timestamp: time.Unix(timestamp.Unix(), 0),
 		Services:  services,
 		IP:        ip,
 		Port:      port,
@@ -88,7 +89,6 @@ func NewNetAddress(addr *net.TCPAddr, services ServiceFlag) *NetAddress {
 // version and whether or not the timestamp is included per ts.  Some messages
 // like version do not include the timestamp.
 func readNetAddress(r io.Reader, pver uint32, na *NetAddress, ts bool) error {
-	var services ServiceFlag
 	var ip [16]byte
 
 	// NOTE: The bitcoin protocol uses a uint32 for the timestamp so it will
@@ -101,7 +101,7 @@ func readNetAddress(r io.Reader, pver uint32, na *NetAddress, ts bool) error {
 		}
 	}
 
-	err := readElements(r, &services, &ip)
+	err := readElements(r, &na.Services, &ip)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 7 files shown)


dskloet:
In tests I saw many more cases where it uses time.Now() with high precision. Would you like me to change those as well?

davecgh:
I personally don't have a super strong opinion on whether the tests should be changed or not, but I'm a huge fan of consistency unless there is a very specific reason not to be.  So, given that, I would say the tests probably should be converted as well.

wire/netaddress.go
```diff
@@ -90,12 +83,13 @@ func NewNetAddress(addr *net.TCPAddr, services ServiceFlag) *NetAddress {
 func readNetAddress(r io.Reader, pver uint32, na *NetAddress, ts bool) error {
 	var services ServiceFlag
 	var ip [16]byte
+	timestamp := na.Timestamp
```
davecgh: I would prefer this section be left the way it is.  The reason is that doing it this way forces another allocation for every single `NetAddress` that is reads.  There was a recent commit that intentionally removed this allocation (and eventually it should do the same for the other two fields as well).

Before this PR:
```
$ go test -run=NONE -bench=BenchmarkDecodeAddr -benchmem
BenchmarkDecodeAddr-6            300           6304134 ns/op        97728 B/op    2002 allocs/op
```
After:

```
$ go test -run=NONE -bench=BenchmarkDecodeAddr -benchmem
BenchmarkDecodeAddr-6            100          10371317 ns/op       129756 B/op    3002 allocs/op
```

We can see the additional allocation is indeed this line here:

```
$ go test -run=NONE -bench=BenchmarkDecodeAddr -memprofile=mem.out -memprofilerate=1
$ go tool pprof -alloc_objects ./wire.test ./mem.out
Entering interactive mode (type "help" for commands)
(pprof) list readNetAddress
Total: 263037
ROUTINE ======================== github.com/btcsuite/btcd/wire.readNetAddress in D:/Projects/mygo/src/github.com/btcsuite/btcd/wire/netaddress.go
    252502     252502 (flat, cum) 95.99% of Total
         .          .     79:
         .          .     80:// readNetAddress reads an encoded NetAddress from r depending on the protocol
         .          .     81:// version and whether or not the timestamp is included per ts.  Some messages
         .          .     82:// like version do not include the timestamp.
         .          .     83:func readNetAddress(r io.Reader, pver uint32, na *NetAddress, ts bool) error {
     50502      50502     84:   var services ServiceFlag
    101000     101000     85:   var ip [16]byte
    101000     101000     86:   timestamp := na.Timestamp
         .          .     87:
         .          .     88:   // NOTE: The bitcoin protocol uses a uint32 for the timestamp so it will
         .          .     89:   // stop working somewhere around 2106.  Also timestamp wasn't added until
         .          .     90:   // protocol version >= NetAddressTimeVersion
         .          .     91:   if ts && pver >= NetAddressTimeVersion {
(pprof)

Comment (dskloet) [inline wire/netaddress.go]: Do you mean this is allocated on the heap? Why isn't it on the stack?

Comment (davecgh) [inline wire/netaddress.go]: Yes, it is allocated on the heap (as can be seen in the profile I pasted).  Interfaces in Go (of which the readers all are) always force the contained thing to be boxed.

In the case of an `addr` message, there is a slice of `NetAddress` which is allocated as a single contiguous block on the heap.  This function then reads directly into those structs that are already on the heap and thus doesn't need another allocation.  If you, however, introduce a local stack variable as is being done here and read into it, the read forces it to the heap because of the boxing.

Comment (dskloet) [inline wire/netaddress.go]: I've removed the allocation of the timestamp and the service flag. The benchmark is now better than before.
I thought I could also avoid the IP allocation in the case where the capacity is already 16, but adding that code makes things worse so I left that for now.

$ go test -run=NONE -bench=BenchmarkDecodeAddr -benchmem
BenchmarkDecodeAddr-4   	    2000	    517959 ns/op	   89729 B/op	    1002 allocs/op
PASS
ok  	github.com/btcsuite/btcd/wire	1.105s

File: connmgr/seed.go [hunk 2]
```diff
@@ -60,14 +60,16 @@ func SeedFromDNS(chainParams *chaincfg.Params, reqServices wire.ServiceFlag,
...
-				// a time randomly selected between 3
-				// and 7 days ago.
-				addresses[i].Timestamp = time.Now().Add(-1 *
-					time.Second * time.Duration(secondsIn3Days+
-					randSource.Int31n(secondsIn4Days)))
+				addresses[i] = &wire.NetAddress{
+					// bitcoind seeds with addresses from
+					// a time randomly selected between 3
+					// and 7 days ago.
+					Timestamp: time.Now().Add(-1 *
```
davecgh: You're correct.  It's already doing a heap alloc anyways, so it certainly wouldn't hurt to refactor it into a func.  If you do add the func, please make sure you maintain the test coverage by ensuring the new func is covered by the tests as well.

On a somewhat related note, one of the things we'd like to ultimately do is reduce allocs by using `[]wire.X` instead of `[]*wire.X` where appropriate.  I believe `[]wire.NetAddress` versus `[]*wire.NetAddress` is one such case.  The first one is a single, contiguous allocation which is better for cache locality and reduces the burden on the GC, and translates into less memory usage over all since there is less tracking overhead needed for the GC for a bunch of small allocations.

Most of the `[]*wire.X` stuff is legacy from when we were new to Go and didn't have as clear of an understanding of the way memory was handled by the GC and also an artifact of how the `range` statement works.

dskloet:
I introduced wire.NewNetAddressTimestamp, which takes all 4 fields and rounds the timestamp to single second precision. I used it in connmgr.seed.go. I wanted to add a test and planned to model it after the test for wire.NewNetAddressIPPort but it turns out there is no test for that either. Both are called by wire.NewNetAddress which is tested so the code is covered. Is that enough?

I've also changed the tests to use single second precision but I'm not convinced it's an improvement and it's still not clear why we need the single second precision in the NetAddress anyway. And if we do, why not store a single int64 rather than an entire time.Time object? If you want, I can revert this part.

What do you think? If you think it's ready, I'll rebase and force push.

dskloet:
How can we move forward with this pull request?

davecgh:
Thanks for following up.  I'll take a closer look and respond tomorrow.  I just wanted to reply tonight, so you aren't left hanging.

davecgh:
Sorry for the delay in getting back to this.

To answer your question about the test coverage,  It is covered indirectly by other tests, so there isn't one directly for it.  I can see that your new function is also covered indirectly since it is invoked by `NewNetAddressIPPort` which is already covered.  The following test coverage report also shows that:

```Go
$ cd $GOPATH/src/github.com/btcsuite/btcd/wire
$ go test -coverprofile=cov.out
$ go tool cover -func=cov.out | grep NewNetAddress
github.com\btcsuite\btcd\wire\netaddress.go:62:         NewNetAddressIPPort     100.0%
github.com\btcsuite\btcd\wire\netaddress.go:69:         NewNetAddressTimestamp  100.0%
github.com\btcsuite\btcd\wire\netaddress.go:84:         NewNetAddress           100.0%
```

As to whether or not to use an `int64`, I could go either way to be honest, but I sort of lean towards it being the standard Go type of `time.Time` since it provides extra functions for dealing with time and it's only a `uint32` on the wire anyways.  That means if it were made an `int64` you couldn't read directly into the struct field without either a local variable or a cast anyways.

I do agree with you that it doesn't really matter much in the tests, but I think it's good to be consistent, and you already did the work, so probably not much point in reverting it.

I'm going to review the actual code now and unless I see anything that I think warrants further discussion or change, I'll get it merged after the review.

davecgh:
This looks good to go.  If you would please squash and rebase it to the latest master, I'll get it merged.  I would do it, but I remember you asked to let you do it (because of github changing your e-mail).

dskloet:
Great. I believe it's ready to merge. But feel free to let me know if anything is still off.

davecgh:
Merged it, but looks like github is having issues right now and isn't showing it as merged yet.

dskloet:
I pull the master branch and it's there so it worked.

dskloet:
But I can't make a new pull request without it looking like a mess so I guess I'll wait for the github issues to go away.

davecgh:
Looks like github noticed the merge after the latest push.  All set.

PR #871, opened 2016-12-02
rpc: implement getblockchaininfo 

Roasbeef:
This commit implements the `getblockchaininfo` as it's defined in BC to the best of btcd's current capabilities. Due to some differences in header storage, and `BIP0009` state storage between `btcd` and BC, some of the information within the call isn't readily or easily accessible to `btcd`, therefore they are currently blank within the PR.

PR #873, opened 2016-12-03
integration: Add BIP0009 state transition tests

davecgh:
** This PR requires #871 **

This adds tests which exercise the BIP0009 state transitions using the newly available soft fork status information in the `getblockchaininfo` RPC.

The following is an overview of the tests added:
- Assert the chain height is 0 and the state is `ThresholdDefined`
- Generate 1 fewer blocks than needed to reach the first state transition
  - Assert chain height is expected and state is still `ThresholdDefined`
- Generate 1 more block to reach the first state transition
  - Assert chain height is expected and state moved to `ThresholdStarted`
- Generate enough blocks to reach the next state transition window, but only signal support in 1 fewer than the required number to achieve `ThresholdLockedIn`
  - Assert chain height is expected and state is still `ThresholdStarted`
- Generate enough blocks to reach the next state transition window with only the exact number of blocks required to achieve locked in status signalling support
  - Assert chain height is expected and state moved to `ThresholdLockedIn`
- Generate 1 fewer blocks than needed to reach the next state transition
  - Assert chain height is expected and state is still `ThresholdLockedIn`
- Generate 1 more block to reach the next state transition
  - Assert chain height is expected and state moved to `ThresholdActive`

In addition, it updates the existing BIP0009 mining tests to include extra assertions that the chain height is at the expected height in addition to checking that the bits are correctly set according to the expected state.

### DIFF
--- a/integration/bip0009_test.go
+++ b/integration/bip0009_test.go
@@ -8,12 +8,25 @@
 package integration
 
 import (
+	"fmt"
 	"runtime"
 	"testing"
+	"time"
 
+	"github.com/btcsuite/btcd/blockchain"
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/rpctest"
+)
+
+const (
+	// vbLegacyBlockVersion is the highest legacy block version before the
+	// version bits scheme became active.
+	vbLegacyBlockVersion = 4
+
+	// vbTopBits defines the bits to set in the version to signal that the
+	// version bits scheme is being used.
+	vbTopBits = 0x20000000
 )
 
 // assertVersionBit gets the passed block hash from the given test harness and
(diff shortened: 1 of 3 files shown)


btcjson/chainsvrresults.go
```diff
@@ -75,16 +75,41 @@ type GetAddedNodeInfoResult struct {
...
+	Blocks               int32                               `json:"blocks"`
+	Headers              int32                               `json:"headers"`
+	BestBlockHash        string                              `json:"bestblockhash"`
+	Difficulty           float64                             `json:"difficulty"`
+	MedianTime           int64                               `json:"mediantime"`
+	VerificationProgress float64                             `json:"verificationprogress,omitempty"`
+	Pruned               bool                                `json:"pruned"`
+	PruneHeight          int32                               `json:"pruneheight,omitempty"`
+	ChainWork            string                              `json:"chainwork,omitempty"`
+	Softforks            []*SoftForkDescription              `json:"softforks"`
```
jongillham: Would this be better as `SoftForks` instead of `Softforks`?

davecgh: Nice catch.  Technically that is part of PR #871 upon which this one depends, but I'll go ahead and comment on that PR accordingly.

integration/bip0009_test.go
```diff
@@ -37,6 +50,255 @@ func assertVersionBit(r *rpctest.Harness, t *testing.T, hash *chainhash.Hash, bi
...
+	// NOTE: This is two blocks before the confirmation window because the
+	// getblockchaininfo RPC reports the status for the block AFTER the
+	// current one.  All of the heights below are thus offset by one to
+	// compensate.
+	//
+	// Assert the chain height is the expected value and soft fork status is
+	// still defined and did NOT move to started.
+	confirmationWindow := r.ActiveNet.MinerConfirmationWindow
+	for i := uint32(0); i < confirmationWindow-2; i++ {
+		_, err := r.GenerateAndSubmitBlock(nil, vbLegacyBlockVersion,
```
Roasbeef: Shouldn't the passed block version here (and within the rest of this integration test func) be the converted version bits block version? Meaning `vbTopBits | bitNum`.

Roasbeef: Realized this was just testing the defind->started, so nevermind :)

integration/bip0009_test.go
```diff
@@ -37,6 +50,255 @@ func assertVersionBit(r *rpctest.Harness, t *testing.T, hash *chainhash.Hash, bi
...
+			time.Time{})
+		if err != nil {
+			t.Fatalf("failed to generated block %d: %v", i, err)
+		}
+	}
+	assertChainHeight(r, t, (confirmationWindow*3)-1)
+	assertSoftForkStatus(r, t, forkKey, blockchain.ThresholdLockedIn)
+
+	// *** ThresholdLockedIn part 2 -- 1 block prior to ThresholdStarted ***
```
Roasbeef: I think the end of the comment should read: 
> 1 block prior to ThresholdActive

integration/bip0009_test.go
```diff
@@ -37,6 +50,255 @@ func assertVersionBit(r *rpctest.Harness, t *testing.T, hash *chainhash.Hash, bi
...
+	}
+	assertChainHeight(r, t, (confirmationWindow*4)-2)
+	assertSoftForkStatus(r, t, forkKey, blockchain.ThresholdLockedIn)
+
+	// *** ThresholdActive ***
+	//
+	// Generate another block to reach the next window without continuing to
+	// signal support since it is already locked in.
+	//
+	//  the chain height is the expected value and the soft fork status
```
Roasbeef: Missing the leading word in the sentence here: "_Assert_ the chain..."

davecgh: Fixed

btcjson/chainsvrresults.go
davecgh: Rebased over latest master (to which #871 has been merged).  Fix included.

integration/bip0009_test.go
davecgh: I modified the comment a bit to specifically call out why it's using a version that does not signal support.

dskloet:
Since I pulled this comment, goclean.sh started failing for me

    integration/bip0009_test.go:99: r.Node.GetBlockChainInfo undefined (type *btcrpcclient.Client has no field or method GetBlockChainInfo)

It seems to introduce a dependency on btcrpcclient that wasn't there before and it breaks because I haven't cloned that other repo? I don't think that's quite it but I'm not sure what exactly is the problem.

davecgh:
Have you updated your dependencies with `glide install`?

dskloet:
I had no idea. Do I (potentially) need to run glide install after every git pull?

davecgh:
Yes, since it ensures all deps are up-to-date.  That said, the deps usually don't change all that often.  The real key is whether or not the `glide.lock` file changed since your last pull.  If it hasn't, everything is still up-to-date.  If it has, you need to run `glide install` (or just `glide i`) to update them.

davecgh:
I use the following script on my boxes that I don't do merging from.  You're welcome to use it and/or tweak it if you find it useful.

```sh
#!/bin/sh

set -e

cd $GOPATH/src/github.com/btcsuite/btcd
git checkout master
cp glide.lock glide.lock.bak
git pull
if ! cmp glide.lock glide.lock.bak; then
        glide i
fi
rm glide.lock.bak
```

dskloet:
Thanks!

dskloet:
Now I get a different error

    --- FAIL: TestBIP0009 (1.50s)
	bip0009_test.go:101: failed to retrieve chain info: -1: Command unimplemented
    --- FAIL: TestBIP0009Mining (1.61s)
	bip0009_test.go:45: assertion failed at line 366: block 08c711462ac7ca8cbd46529d44094d2d9acd0318063033284d870d57115caf9e, version 0x4 does not have bit 2

davecgh:
The integration tests use the installed version of btcd.  It looks like it's still running the old version.  Make sure you also do `go install` so the latest btcd executable is installed to your `$GOPATH`.

dskloet:
OK, but what's the point of testing the installed version? Shouldn't it test the actual changes in the repo?

Sorry, for cluttering your PR. Is there a better place to discuss? Or you don't mind?

davecgh:
It's really just a relic of the fact the integration tests are relatively new and were written specifically to run in the CI build environment where the latest version is always installed as part of the setup.  Thus, when you submit a PR, it is always tested against the latest.

I'm in favor of changing the `rpctest` framework to use the version in the parent path (`rpctest` lives at `$GOPATH/src/github.com/btcsuite/btcd/rpctest`, hence I mention the parent `..`), just haven't got around to doing it since it works properly in the CI environment and there are much higher priority things.  In the mean time, if you or someone else wants to take a stable at getting it working properly from the repo directory, I'd be happy to review and merge the PR.

PR #879, opened 2016-12-08
server.go: Optimize newAddressFunc

dskloet:
Change the order of conditions to avoid calling fmt.Sprintf
unnecessarily.


dskloet:
Yes, we did discuss it. And now I got around to it :)

PR #880, opened 2016-12-09
Drop btcsuite/go-flags in favor of upstream

dajohi:


davecgh:
I'm not really a big fan of this change.  It means that the upstream repo can be deleted outside of our control and suddenly you won't be able to build anymore.  It's true that `glide.lock` could be updated to point to a different repo in that event, but I'm not really comfortable relying on a bunch of repos we don't directly control.

jrick:
obviously I like it, or I would not have approved it.

i'm not too paranoid about the repo disappearing, since this is 1) probably not going to happen and 2) trivial to fix if it does by pointing it to our own fork.  If we still can't accept the possibility of it disappearing for a short time, then we can switch over the upstream import path but pull the code from our own fork.

jrick:
Can we decide what to do with this? merge it or close the pr. I don't like it just sitting here without any activity.

stevenroose:
As it would be a bit strange if everyone would have to fork repo's in order to depend on them, I'd try to avoid that scenario as much as possible. If you really want to make sure that this repo builds regardless of dependency availability, you need a vendor directory. However I think Glide is the ideal way in between.

Issue #886, opened 2016-12-14
Port version RPC from dcrd


jrick:
In dcrd the json-rpc API is versioned and can be queried with the version rpc.  I'd like that same functionality in btcd so that btcwallet and other clients can test for compatibility with the server.

https://github.com/decred/dcrd/blob/da8a0334fa1f08b9d7e37a0ba3355c45e7694f30/rpcserver.go#L5731

aakselrod:
I'm taking this issue.

jrick:
Thanks. Please do this before the other backports, and then increment the version for each of those.

jrick:
This has been implemented on master.

PR #888, opened 2016-12-17
Add `version` command to JSON-RPC API for API versioning

aakselrod:
This pull request is a backport of the `version` JSON-RPC API command from dcrd. It permits the JSON-RPC API to have its own semantic versioning scheme separate from the underlying software. The initial version is 1.0.0. This pull request resolves issue #886.

### DIFF
--- a/docs/json_rpc_api.md
+++ b/docs/json_rpc_api.md
@@ -583,6 +583,7 @@ The following is an overview of the RPC methods which are implemented by btcd, b
 |4|[searchrawtransactions](#searchrawtransactions)|Y|Query for transactions related to a particular address.|None|
 |5|[node](#node)|N|Attempts to add or remove a peer. |None|
 |6|[generate](#generate)|N|When in simnet or regtest mode, generate a set number of blocks. |None|
+|7|[version](#version)|Y|Returns the JSON-RPC API version.|
 
 
 <a name="ExtMethodDetails" />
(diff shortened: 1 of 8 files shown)


jrick:
This reads fine (well, I wrote it originally :p) but I think any touched files should also have "The Decred developers" added to the copyrights in the license header.  I think that the LICENSE file in the project root also needs to be updated similarly.

aakselrod:
Sounds like a good idea to me. @davecgh, are you OK with that? If so, I'll go ahead and update.

docs/json_rpc_api.md
```diff
@@ -566,6 +567,18 @@ Example Return|`{`<br />&nbsp;&nbsp;`"bytes": 310768,`<br />&nbsp;&nbsp;`"size":
...
+***
+<a name="version"/>
+
+|   |   |
+|---|---|
+|Method|version|
+|Parameters|None|
+|Description|Returns the version of the JSON-RPC API built into this release of btcd.|
+|Returns|string|
+|Example Return|1.0.0|
```
jrick: Should be the actual json response that is returned, for example:

```
{
    "btcdjsonrpcapi": {
        "versionstring": "1.0.0,
        "major": 1,
        "minor": 0,
        "patch": 0,
        "prerelease": "",
        "buildmetadata": ""
    }
}
```

aakselrod: ACK, fixing.

aakselrod: Fixed.

jrick: Still missing the key for "btcdjsonrpcapi".  The idea with this object format is that we can extend it to provide versions of other systems or programs (for example, when wallet adds this, it can report both api versions, because that matters to clients who are using rpc passthrough).

aakselrod: ACK. I *think* I've got it this time.

btcjson/walletsvrcmds.go
```diff
@@ -640,6 +641,19 @@ func NewWalletPassphraseChangeCmd(oldPassphrase, newPassphrase string) *WalletPa
 	}
 }
 
+// VersionCmd defines the version JSON-RPC command.
+//
+// NOTE: This is a btcsuite extension ported from
+// github.com/decred/dcrd/dcrjson.
+type VersionCmd struct{}
```
davecgh: This is a btcd-specific extension command, so I think it should be in the `btcdextcmds.go` file.  This file is for wallet-specific commands that are compatible with Core.

btcjson/walletsvrcmds.go
```diff
@@ -1,4 +1,5 @@
-// Copyright (c) 2014 The btcsuite developers
+// Copyright (c) 2014-2016 The btcsuite developers
+// Copyright (c) 2015-2016 The Decred developers
```
davecgh: Once the command definition and supporting code is moved (see below), there would no longer be any need to update the copyright header in this file since it will have no changes.

btcjson/walletsvrcmds_test.go
```diff
@@ -1153,6 +1154,17 @@ func TestWalletSvrCmds(t *testing.T) {
 			},
 		},
 		{
+			name: "version",
```
davecgh: Same thing here.  This should all go in `btcdextcmds_test.go` since it is a btcd-specific extension.

docs/json_rpc_api.md
```diff
@@ -566,6 +567,18 @@ Example Return|`{`<br />&nbsp;&nbsp;`"bytes": 310768,`<br />&nbsp;&nbsp;`"size":
 |Example Return|`true`|
 [Return to Overview](#MethodOverview)<br />
 
+***
```
davecgh: Along the same lines, this doco should move to section 6 where the extension methods are defined.

btcjson/walletsvrcmds.go
aakselrod: ACK, fixed.

btcjson/walletsvrcmds.go
aakselrod: ACK, fixed.

btcjson/walletsvrcmds_test.go
aakselrod: ACK, fixed.

docs/json_rpc_api.md
aakselrod: ACK, fixed.

PR #889, opened 2016-12-19
Add MerkleBlock deadline to CmdGetData

cpacia:


Roasbeef:
Seems like a straight forward change to me, and it reads well. 


Have you been able to test this patch out in the wild yet from the PoV of an SPV node that's using the `peer` library? If so, has this patched improved any past symptoms? 

Another thing I'm curious about is if you've run into any false positives related to the added stall detection from the PoV of your SPV peer. Currently, due to lack of a context within the `stallDetector` itself, at times `btcd` erroneously disconnects a peer for stalling since depending on the chain state of both nodes, certain messages won't _always_ warrant a reply.

cpacia:
@Roasbeef I have been using the peer class in my spv wallet I've been working on. This was needed to make sure I dump peers who don't respond to a merkleblock getdata. So far it seems to be working well as I haven't noticed any issues, though I haven't yet put my code through the full paces.

dajohi:
@cpacia make sure your SPV client is only sending to peers that support it.  You probably already have this.

```go
if peer.ProtocolVersion() >= wire.BIP0037Version &&
    peer.Services()&wire.SFNodeNetwork == wire.SFNodeNetwork {

    // request merkleblock
} else {
    // request block
}
```

as for the PR, OK.

cpacia:
Thanks @dajohi. Yeah it checks the services for nodeBloom and disconnects from peers that don't support it.

Issue #883, opened 2016-12-12
Port getheaders RPC from dcrd


jrick:
This will be used by upcoming changes to btcwallet (backported from dcrwallet).

The block locator hashes in dcrd's getheaders RPC are also passed as a single concatenated non-byte-reversed hexadecimal string. We should discuss whether to continue using this or to switch to using a JSON array of byte-reversed block hash strings.

jrick:
Discussed with @davecgh on zkc and will use a JSON array of strings for the hashes.

aakselrod:
I'm taking this issue.

davecgh:
This has been resolved by #890.

PR #890, opened 2016-12-22
Port `getheaders` JSON-RPC command from dcrd

aakselrod:
This pull request is the backport of the `getheaders` JSON-RPC API command from Decred's dcrd. It allows a client to download headers for an arbitrary range of blocks.

This PR resolves issue #883.

### DIFF
--- a/btcjson/btcdextresults.go
+++ b/btcjson/btcdextresults.go
@@ -1,5 +1,5 @@
-// Copyright (c) 2016 The btcsuite developers
-// Copyright (c) 2015-2016 The Decred developers
+// Copyright (c) 2016-2017 The btcsuite developers
+// Copyright (c) 2015-2017 The Decred developers
 // Use of this source code is governed by an ISC
 // license that can be found in the LICENSE file.
 
--- a/docs/json_rpc_api.md
+++ b/docs/json_rpc_api.md
@@ -584,6 +584,7 @@ The following is an overview of the RPC methods which are implemented by btcd, b
 |5|[node](#node)|N|Attempts to add or remove a peer. |None|
 |6|[generate](#generate)|N|When in simnet or regtest mode, generate a set number of blocks. |None|
 |7|[version](#version)|Y|Returns the JSON-RPC API version.|
+|8|[getheaders](#getheaders)|Y|Returns block headers starting with the first known block hash from the request.|
 
 
 <a name="ExtMethodDetails" />
@@ -674,6 +675,19 @@ The following is an overview of the RPC methods which are implemented by btcd, b
 |Description|Returns the version of the JSON-RPC API built into this release of btcd.|
 |Returns|`{ (json object)`<br />&nbsp;&nbsp;`"btcdjsonrpcapi": {`<br />&nbsp;&nbsp;&nbsp;&nbsp;`"versionstring": "x.y.z",  (string) the version of the JSON-RPC API`<br />&nbsp;&nbsp;&nbsp;&nbsp;`"major": x,  (numeric) the major version of the JSON-RPC API`<br />&nbsp;&nbsp;&nbsp;&nbsp;`"minor": y,  (numeric) the minor version of the JSON-RPC API`<br />&nbsp;&nbsp;&nbsp;&nbsp;`"patch": z,  (numeric) the patch version of the JSON-RPC API`<br />&nbsp;&nbsp;&nbsp;&nbsp;`"prerelease": "",  (string) prerelease info for the JSON-RPC API`<br />&nbsp;&nbsp;&nbsp;&nbsp;`"buildmetadata": ""  (string) metadata about the server build`<br />&nbsp;&nbsp;`}`<br />`}`|
 |Example Return|`{`<br />&nbsp;&nbsp;`"btcdjsonrpcapi": {`<br />&nbsp;&nbsp;&nbsp;&nbsp;`"versionstring": "1.0.0",`<br />&nbsp;&nbsp;&nbsp;&nbsp;`"major": 1,  `<br />&nbsp;&nbsp;&nbsp;&nbsp;`"minor": 0,`<br />&nbsp;&nbsp;&nbsp;&nbsp;`"patch": 0,`<br />&nbsp;&nbsp;&nbsp;&nbsp;`"prerelease": "",`<br />&nbsp;&nbsp;&nbsp;&nbsp;`"buildmetadata": ""`<br />&nbsp;&nbsp;`}`<br />`}`|
+[Return to Overview](#MethodOverview)<br />
+
+***
+
+<a name="getheaders"/>
+
+|   |   |
+|---|---|
+|Method|getheaders|
+|Parameters|1. Block Locators (JSON array, required)<br />&nbsp;`[ (json array of strings)`<br />&nbsp;&nbsp;`"blocklocator", (string) the known block hash`<br />&nbsp;&nbsp;`...`<br />&nbsp;`]`<br />2. hashstop (string) - last desired block's hash|
+|Description|Returns block headers starting with the first known block hash from the request.|
+|Returns|`[ (json array of strings)`<br />&nbsp;&nbsp;`"blockheader",`<br />&nbsp;&nbsp;`...`<br />`]`|
+|Example Return|`[`<br />&nbsp;&nbsp;`"0000002099417930b2ae09feda10e38b58c0f6bb44b4d60fa33f0e000000000000000000d53...",`<br />&nbsp;&nbsp;`"000000203ba25a173bfd24d09e0c76002a910b685ca297bd09a17b020000000000000000702..."`<br />`]`|
 [Return to Overview](#MethodOverview)<br />
 
 ***
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -1,5 +1,5 @@
-// Copyright (c) 2013-2016 The btcsuite developers
-// Copyright (c) 2015-2016 The Decred developers
+// Copyright (c) 2013-2017 The btcsuite developers
+// Copyright (c) 2015-2017 The Decred developers
 // Use of this source code is governed by an ISC
 // license that can be found in the LICENSE file.
 
@@ -45,9 +45,9 @@ import (
 
 // API version constants
 const (
-	jsonrpcSemverString = "1.0.0"
+	jsonrpcSemverString = "1.1.0"
 	jsonrpcSemverMajor  = 1
-	jsonrpcSemverMinor  = 0
+	jsonrpcSemverMinor  = 1
 	jsonrpcSemverPatch  = 0
 )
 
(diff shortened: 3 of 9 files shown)


rpcserverhelp.go
```diff
@@ -360,6 +360,12 @@ var helpDescsEnUS = map[string]string{
 	"infowalletresult-relayfee":        "The minimum relay fee for non-free transactions in BTC/KB",
 	"infowalletresult-errors":          "Any current errors",
 
+	// GetHeadersCmd help.
+	"getheaders--synopsis":     "Returns block headers starting with the first known block hash from the request",
+	"getheaders-blocklocators": "Concatenated hashes of blocks.  Headers are returned starting from the first known hash in this list",
```
jrick: Description needs to be updated for the new type, in btcd it's a JSON array of strings

rpcserver.go
```diff
@@ -2148,6 +2150,67 @@ func handleGetHashesPerSec(s *rpcServer, cmd interface{}, closeChan <-chan struc
...
+		if err != nil {
+			return nil, &btcjson.RPCError{
+				Code: btcjson.ErrRPCInvalidParameter,
+				Message: "Failed to decode block locator: " +
+					err.Error(),
+			}
+		}
+		pBlockLocators[i] = pBlockLocator
+	}
+	var hashStop *chainhash.Hash
```
jrick: This isn't quite right.  The stop hash for locateBlocks is not optional.  In the dcrd RPC method handler, it passes a pointer to a zero'd hash when the request's parameter is missing.

server.go
```diff
@@ -640,6 +641,85 @@ func (sp *serverPeer) OnGetBlocks(_ *peer.Peer, msg *wire.MsgGetBlocks) {
 	}
 }
 
+// locateBlocks returns the hashes of the blocks after the first known block in
+// locators, until hashStop is reached, or up to a max of
+// wire.MaxBlockHeadersPerMsg block hashes.  This implements the search
+// algorithm used by getheaders.
+//
+// NOTE: This is a btcsuite extension ported from
+// github.com/decred/dcrd.
```
jrick: i'd remove this part of the comment. this isn't a new feature, just some code reorganization that dcrd did to make this feature available to call by the rpc server.

server.go
```diff
@@ -640,6 +641,85 @@ func (sp *serverPeer) OnGetBlocks(_ *peer.Peer, msg *wire.MsgGetBlocks) {
...
+
+	// Fetch the inventory from the block database.
+	return chain.HeightRange(startIdx, endIdx)
+}
+
+// fetchHeaders fetches and decodes headers from the db for each hash in
+// blockHashes.
+//
+// NOTE: This is a btcsuite extension ported from
+// github.com/decred/dcrd.
```
jrick: see above.

jrick:
If you look at the dcrd commit that implemented this, you can see that the codepaths used by wire protocol have also been updated for some of the more generic block location and header fetching apis.  We should make the same changes to btcd.

https://github.com/decred/dcrd/commit/70c5a4957ae3913f50d50ac9711485011499818b

rpcserverhelp.go
aakselrod: ACK. Fixed in new commit (will squash when ready).

server.go
aakselrod: ACK. Fixed in new commit (will squash when ready).

server.go
aakselrod: ACK. Fixed in new commit (will squash when ready).

rpcserver.go
aakselrod: ACK. Looking at this.

aakselrod:
ACK. Looking at this.

rpcserver.go
aakselrod: So the reason I did it this way is that btcd's `chainhash` package has no `Decode` function. The `NewHashFromStr` method returns a `*Hash` and an `error`. I've changed this code to make an empty stop hash into an all-zero stop hash.

aakselrod:
Updated OnGetHeaders to use locateBlocks and fetchHeaders.

docs/json_rpc_api.md
```diff
@@ -678,6 +679,19 @@ The following is an overview of the RPC methods which are implemented by btcd, b
...
 
+<a name="getheaders"/>
+
+|   |   |
+|---|---|
+|Method|getheaders|
+|Parameters|1. Block Locators (JSON array, required)<br />&nbsp;`[ (json array of strings)`<br />&nbsp;&nbsp;`"blocklocator", (string) the known block hash`<br />&nbsp;&nbsp;`...`<br />&nbsp;`]`<br />2. hashstop (string) - last desired block's hash|
+|Description|Returns block headers starting with the first known block hash from the request.|
+|Returns|`{ (json object)`<br />&nbsp;&nbsp;`[ (json array of strings)`<br/>&nbsp;&nbsp;&nbsp;&nbsp;`"blockheader",`<br />&nbsp;&nbsp;&nbsp;&nbsp;`...`<br />&nbsp;&nbsp;`]`<br />`}`|
+|Example Return|`{`<br />&nbsp;&nbsp;`[`<br/>&nbsp;&nbsp;&nbsp;&nbsp;`"0123456789abcdef",`<br />&nbsp;&nbsp;&nbsp;&nbsp;`"fedcba9876543210"`<br />&nbsp;&nbsp;`]`<br />`}`|
```
aakselrod: ACK. Added "headers" key to Returns and Example Return sections.

rpcserver.go
```diff
@@ -2148,6 +2150,66 @@ func handleGetHashesPerSec(s *rpcServer, cmd interface{}, closeChan <-chan struc
 	return int64(s.server.cpuMiner.HashesPerSecond()), nil
 }
 
+// handleGetHeaders implements the getheaders command.
+//
+// NOTE: This is a btcsuite extension ported from
+// github.com/decred/dcrd.
+func handleGetHeaders(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (interface{}, error) {
+	c := cmd.(*btcjson.GetHeadersCmd)
+	pBlockLocators := make([]*chainhash.Hash, len(c.BlockLocators))
```
aakselrod: ACK. Removed Hungarian notation.

rpcserver.go
```diff
@@ -2148,6 +2150,66 @@ func handleGetHashesPerSec(s *rpcServer, cmd interface{}, closeChan <-chan struc
...
+			return nil, &btcjson.RPCError{
+				Code: btcjson.ErrRPCInvalidParameter,
+				Message: "Failed to decode block locator: " +
+					err.Error(),
+			}
+		}
+		pBlockLocators[i] = pBlockLocator
+	}
+	if c.HashStop != "" {
+		c.HashStop = strings.Repeat("0", chainhash.HashSize*2)
```
aakselrod: ACK. Created `chainhash.Decode` function and updated `chainhash.NewHashFromStr` and `handleGetHeaders` to use it like dcrd does.

server.go
```diff
@@ -700,52 +673,75 @@ func (sp *serverPeer) OnGetHeaders(_ *peer.Peer, msg *wire.MsgGetHeaders) {
...
+	blockHashes, err := sp.server.locateBlocks(msg.BlockLocatorHashes,
+		&msg.HashStop)
 	if err != nil {
-		peerLog.Warnf("Failed to build headers: %v", err)
+		peerLog.Errorf("OnGetHeaders: failed to fetch hashes: %v", err)
 		return
 	}
+	blockHeaders, err := fetchHeaders(sp.server.db, blockHashes)
+	if err != nil {
+		peerLog.Errorf("OnGetHeaders: failed to fetch block headers: "+
```
davecgh: Same here.

server.go
```diff
@@ -700,52 +673,75 @@ func (sp *serverPeer) OnGetHeaders(_ *peer.Peer, msg *wire.MsgGetHeaders) {
...
+			"%v", err)
```
davecgh: Looks like a `return` is missing here.

server.go
```diff
@@ -700,52 +673,75 @@ func (sp *serverPeer) OnGetHeaders(_ *peer.Peer, msg *wire.MsgGetHeaders) {
...
+	// Ignore getheaders requests if not in sync.
+	if !sp.server.blockManager.IsCurrent() {
+		return
+	}
+
+	blockHashes, err := sp.server.locateBlocks(msg.BlockLocatorHashes,
+		&msg.HashStop)
 	if err != nil {
-		peerLog.Warnf("Failed to build headers: %v", err)
+		peerLog.Errorf("OnGetHeaders: failed to fetch hashes: %v", err)
```
jrick: I disagree.  Part of running a full node is being able to serve other nodes.  In this case, the input received by the other nodes is not incorrect, and the failure to serve them lies with us rather than them.  In my mind that should be an error and not a warning.  For invalid, unexpected, or deprecated input received from other nodes and and the local administrator (e.g. RPC client), a warning is acceptable iff the program can continue executing.

davecgh: After discussing this with @jrick, I think it's safe to go ahead leave this as a `.Errorf`.  In practice it should never be hit anyways unless something in the code is wrong.

server.go
davecgh: Same reasoning as above here.  @jrick convincement me it's reasonable to leave as `.Errorf`.

btcjson/btcdextresults.go
```diff
@@ -5,6 +5,15 @@
 
 package btcjson
 
+// GetHeadersResult models the data returned by the chain server getheaders
+// command.
+//
+// NOTE: This is a btcsuite extension ported from
+// github.com/decred/dcrd/dcrjson.
+type GetHeadersResult struct {
```
aakselrod: ACK. I've made the change you requested.

server.go
aakselrod: ACK. Left as is.

server.go
aakselrod: ACK. Left as is.

server.go
aakselrod: ACK. Added missing `return`.

docs/json_rpc_api.md
```diff
@@ -678,6 +679,19 @@ The following is an overview of the RPC methods which are implemented by btcd, b
...
 
+<a name="getheaders"/>
+
+|   |   |
+|---|---|
+|Method|getheaders|
+|Parameters|1. Block Locators (JSON array, required)<br />&nbsp;`[ (json array of strings)`<br />&nbsp;&nbsp;`"blocklocator", (string) the known block hash`<br />&nbsp;&nbsp;`...`<br />&nbsp;`]`<br />2. hashstop (string) - last desired block's hash|
+|Description|Returns block headers starting with the first known block hash from the request.|
+|Returns|`[ (json array of strings)`<br />&nbsp;&nbsp;`"blockheader",`<br />&nbsp;&nbsp;`...`<br />`]`|
+|Example Return|`[`<br />&nbsp;&nbsp;`"0123456789abcdef",`<br />&nbsp;&nbsp;`"fedcba9876543210"`<br />`]`|
```
davecgh: Can we use truncated real headers here for the examples?  The other entries all have valid examples.  For example consider the `hex` field in `getrawtransaction`.

Here are the headers for 446447 and 446448 truncated to the same length as the other long hex fields:

> 0000002099417930b2ae09feda10e38b58c0f6bb44b4d60fa33f0e000000000000000000d53...
> 000000203ba25a173bfd24d09e0c76002a910b685ca297bd09a17b020000000000000000702...

aakselrod: ACK. Updated the examples.

PR #894, opened 2017-01-11
mining: drop getwork support.

dajohi:
Since the Midstate is no longer needed, switch to using
crypto/sha256.

Details from @davecgh:

For some additional details regarding this change, the `getwork` RPC has been deprecated for quite a while now because it simply can't serve ASICs fast enough to be useful.  This is because `getwork` only provided for the caller to manipulate the 32-bit nonce field in the header to find a solution.  Given it's a 32-bit field, that means it only supported 2^32 variations, which typically do not result in a solution due to the ever increasing difficulty requirements.  Further, ASICs are able to try every variation within that range so fast that even if `getwork` were changed to supply a new template on every invocation, it would still severely throttle them.

The `getblocktemplate` RPC, on the other hand, does not have this issue since it provides miners with control over the coinbase transaction (among many other things such as transaction selection) which allows them to modify it (typically via a 64-bit extra nonce) which in turn changes the 256-bit merkle root thus preventing ASIC throttling.


dajohi:
Removed from docs.

docs/json_rpc_api.md
```diff
@@ -170,7 +170,6 @@ the method name for further details such as parameter and return information.
 |20|[getpeerinfo](#getpeerinfo)|N|Returns information about each connected network peer as an array of json objects.|
 |21|[getrawmempool](#getrawmempool)|Y|Returns an array of hashes for all of the transactions currently in the memory pool.|
 |22|[getrawtransaction](#getrawtransaction)|Y|Returns information about a transaction given its hash.|
-|23|[getwork](#getwork)|N|Returns formatted hash data to work on or checks and submits solved data.<br /><font color="orange">NOTE: Since btcd does not have the wallet integrated to provide payment addresses, btcd must be configured via the `--miningaddr` option to provide which payment addresses to pay created blocks to for this RPC to function.</font>|
 |24|[help](#help)|Y|Returns a list of all commands or help for a specified command.|
```
davecgh: These need to be renumbered since 23 was removed.

sample-btcd.conf
```diff
@@ -293,8 +293,7 @@
 ; worth your while.
 ; generate=false
 
-; Add addresses to pay mined blocks to for CPU mining and the block templates
-; generated for the getwork RPC as desired.  One address per line.
+; Add addresses to pay mined blocks to for CPU mining.  One address per line.
```
davecgh: The bit about block templates is still true it's just for `getblocktemplate` instead of `getwork`.

davecgh: I should note that it is **potentially** true though.  It depends on the flags in the getblocktemplate request.  It is possible to request templates with or without coinbase details.  When requested with coinbase details, a mining address is required.

dajohi: fixed.

docs/json_rpc_api.md
dajohi: done.

PR #900, opened 2017-01-16
btcd: fix rebroadcasting of local txs.

dajohi:


rpcserver.go
```diff
@@ -3251,12 +3251,17 @@ func handleSendRawTransaction(s *rpcServer, cmd interface{}, closeChan <-chan st
 		}
 	}
 
+	if len(acceptedTxs) == 0 || !acceptedTxs[0].Tx.Hash().IsEqual(tx.Hash()) {
+		s.server.txMemPool.RemoveTransaction(tx, true)
+		return nil, btcjson.ErrRPCInternal
```
davecgh: The error here should be using `internalRPCError`.  Also, let's go ahead and add a comment about why this is being done:

```Go
	// When the transaction was accepted it should be the first item in the
	// returned array of accepted transactions.  The only way this will not
	// be true is if the API for ProcessTransaction changes and this code is
	// not properly updated, but ensure the condition holds as a safeguard.
	//
	// Also, since an error is being returned to the caller, ensure the
	// transaction is removed from the memory pool.
	if len(acceptedTxs) == 0 || !acceptedTxs[0].Tx.Hash().IsEqual(tx.Hash()) {
		s.server.txMemPool.RemoveTransaction(tx, true)

		errStr := fmt.Sprintf("transaction %v is not in accepted list",
			tx.Hash())
		return nil, internalRPCError(errStr, "")
	}
```

rpcserver.go
```diff
@@ -3251,12 +3251,17 @@ func handleSendRawTransaction(s *rpcServer, cmd interface{}, closeChan <-chan st
...
+	}
+
 	s.server.AnnounceNewTransactions(acceptedTxs)
 
 	// Keep track of all the sendrawtransaction request txns so that they
 	// can be rebroadcast if they don't make their way into a block.
-	iv := wire.NewInvVect(wire.InvTypeTx, tx.Hash())
-	s.server.AddRebroadcastInventory(iv, tx)
+	iv := wire.NewInvVect(wire.InvTypeTx, acceptedTxs[0].Tx.Hash())
```
davecgh: No reason to change this to another array lookup.  The transaction hash is the same.

If you want them to be the same, that is acceptable, but please only do the lookup once.  i.e
```Go
	txD := acceptedTxs[0]
	iv := wire.NewInvVect(wire.InvTypeTx, txD.Tx.Hash())
	s.server.AddRebroadcastInventory(iv, txD)
```

PR #903, opened 2017-01-23
build: Exclude .glide from test coverage.

davecgh:
The most recent version of glide creates a working directory named `.glide` that includes .go files and thus must be excluded from the find operation that generates merged test coverage of all packages.


davecgh:
I will work up a second PR that uses a specific glide version instead of the latest master branch, but I wanted to get this fix in as well since it will be required when updating to the next glide release anyways.

PR #904, opened 2017-01-23
TravisCI: Use release version of glide.

davecgh:
This modifies the `.travis.yml` file to install a release version of glide instead of using its latest master branch.  This will prevent upstream changes from inadvertently breaking the CI builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,10 @@ go:
   - 1.7.x
 sudo: false
 install:
-  - go get -v github.com/Masterminds/glide
+  - GLIDE_TAG=v0.12.3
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,7 +4,10 @@ go:
   - 1.7.x
 sudo: false
 install:
-  - go get -v github.com/Masterminds/glide
+  - GLIDE_TAG=v0.12.3
+  - GLIDE_PKG=github.com/Masterminds/glide
+  - go get -v $GLIDE_PKG
```
jrick: if master doesn't build, this will fail

davecgh: Good call.  I forget to put `-d` on there so it's download only.  I'll do that now.

davecgh: I went ahead and switched it all over to just download the release binary.

PR #905, opened 2017-01-23
btcjson: Add getmempoolentry API

### LINKED ISSUE #908: btcjson: Add getmempoolentry API

dajohi:
Fixes #908

### DIFF
--- a/btcjson/chainsvrcmds.go
+++ b/btcjson/chainsvrcmds.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2014 The btcsuite developers
+// Copyright (c) 2014-2017 The btcsuite developers
 // Use of this source code is governed by an ISC
 // license that can be found in the LICENSE file.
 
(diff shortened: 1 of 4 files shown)


FrozenPrincess:
<3

BTW the struct `GetRawMempoolVerboseResult` and `GetMempoolEntryResult` should actually be the same thing (although at the moment GetRawMempoolVerboseResult is missing fields), so it might be best to typedef them or something

dajohi:
Thanks for pointing that out.  I'll do that after btcd gets support for the new fields in getrawmempool verbose.

dajohi:
Fixes #908

btcjson/chainsvrcmds.go
```diff
@@ -333,6 +333,19 @@ func NewGetInfoCmd() *GetInfoCmd {
 	return &GetInfoCmd{}
 }
 
+// GetMempoolEntryCmd defines the getmempoolentry JSON-RPC command.
+type GetMempoolEntryCmd struct {
+	TxID string
```
jrick: doesn't need it actually. json-rpc requests are positional, not keyed in an object.

davecgh: Ah right, I noticed it was on one of the others (`GetHeadersCmd`).  Given it's positional though, you're right it isn't needed.

jrick: Yep.. I added it to that on accident.  We can remove it but it doesn't really matter, it just gets ignored.

PR #910, opened 2017-01-29
blockchain: Store side chain blocks in database.

davecgh:
This modifies the `blockchain` code to store all blocks that have passed proof-of-work and contextual validity tests in the database even if they may ultimately fail to connect.

This eliminates the need to store those blocks in memory, allows them to be available as orphans later even if they were never part of the main chain, and helps pave the way toward being able to separate the download logic from the connection logic.

### DIFF
--- a/blockchain/accept.go
+++ b/blockchain/accept.go
@@ -4,7 +4,10 @@
 
 package blockchain
 
-import "github.com/btcsuite/btcutil"
+import (
+	"github.com/btcsuite/btcd/database"
+	"github.com/btcsuite/btcutil"
+)
 
 // maybeAcceptBlock potentially accepts a block into the block chain and, if
 // accepted, returns whether or not it is on the main chain.  It performs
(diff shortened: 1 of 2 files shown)


stevenroose:
So this will store them indefinitely? There is a timestamp attached to the block, so I would prefer those blocks not to stay forever. I suppose block age is checked upon validation as well, so blocks could be removed from storage when they are too old to be accepted again (this would prevent a node from broadcasting the same old transaction and have btcd store it again).

davecgh:
Correct.  It's worth noting that this behavior already happens for side chain blocks that were once part of the main chain from the point of view of the node in question and this PR just makes it more consistently apply.  For example, take a look at [orphaned blocks](https://blockchain.info/orphaned-blocks).

I agree that eventually removing old blocks is definitely something that makes sense, however I think that should be done as part of overall block pruning.

This is really just the first step in the many steps that are needed to get to multi-peer syncs.

stevenroose:
@davecgh It does indeed make more sense to do this in the more general context of pruning.

PR #911, opened 2017-01-31
blockchain: Use int64 timestamps in block nodes.

davecgh:
This modifies the block nodes used in the `blockchain` package for keeping track of the block index to use `int64` for the timestamps instead of `time.Time`.

This is being done because a `time.Time` takes 24 bytes while an `int64` only takes 8 and the plan is to eventually move the entire block index into memory instead of the dynamically-loaded version, so cutting the number of bytes used for the timestamp by a third is highly desirable.

Also, the consensus code requires working with unix-style timestamps anyways, so switching over to them in the block node does not seem unreasonable.

Finally, this does not go so far as to change all of the `time.Time` references, particularly those that are in the public API, so it is purely an internal change.

### DIFF
--- a/blockchain/difficulty.go
+++ b/blockchain/difficulty.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2013-2016 The btcsuite developers
+// Copyright (c) 2013-2017 The btcsuite developers
 // Use of this source code is governed by an ISC
 // license that can be found in the LICENSE file.
 
@@ -158,14 +158,16 @@ func CalcWork(bits uint32) *big.Int {
 // known good checkpoint.
 func (b *BlockChain) calcEasiestDifficulty(bits uint32, duration time.Duration) uint32 {
 	// Convert types used in the calculations below.
-	durationVal := int64(duration)
+	durationVal := int64(duration / time.Second)
 	adjustmentFactor := big.NewInt(b.chainParams.RetargetAdjustmentFactor)
 
 	// The test network rules allow minimum difficulty blocks after more
 	// than twice the desired amount of time needed to generate a block has
 	// elapsed.
 	if b.chainParams.ReduceMinDifficulty {
-		if durationVal > int64(b.chainParams.MinDiffReductionTime) {
+		reductionTime := int64(b.chainParams.MinDiffReductionTime /
+			time.Second)
+		if durationVal > reductionTime {
 			return b.chainParams.PowLimitBits
 		}
 	}
@@ -243,9 +245,10 @@ func (b *BlockChain) calcNextRequiredDifficulty(lastNode *blockNode, newBlockTim
 		if b.chainParams.ReduceMinDifficulty {
 			// Return minimum difficulty when more than the desired
 			// amount of time has elapsed without mining a block.
-			reductionTime := b.chainParams.MinDiffReductionTime
-			allowMinTime := lastNode.timestamp.Add(reductionTime)
-			if newBlockTime.After(allowMinTime) {
+			reductionTime := int64(b.chainParams.MinDiffReductionTime /
+				time.Second)
+			allowMinTime := lastNode.timestamp + reductionTime
+			if newBlockTime.Unix() > allowMinTime {
 				return b.chainParams.PowLimitBits, nil
 			}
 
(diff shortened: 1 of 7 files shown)


jrick:
why does this use int64 and not uint32?

davecgh:
Mainly because all unix timestamps are `int64` and thus it simplifies the code and is more natural.  However, there are cases where the timestamps are subtracted from one another, so using uint32 without first checking which one is higher would lead to incorrect results.

I assume you're asking about `uint32` because that is what is sent on the wire (which incidentally will have to change by 2106 anyways).  I'd prefer to use a signed type because of the above and `int32` isn't enough, so that leaves `int64`.

Any particular reason you'd like it to be `uint32`?

jrick:
I wouldn't like it to be uint32, just curious why it was given the format on the wire.

What does core do here?  It sounds like if they use uint32_t then their math is (or maybe might be) wrong during subtraction, but that int32 also "isn't enough" (or are you only talking about the 2106 problem there?).

(and most, not all unix timestamps are 64-bit, but that is irrelevant to this discussion)

davecgh:
I just double checked and Core uses `int64` for all of the timestamp math operations as expected.  It would have to be that way, otherwise the consensus timestamp math would be wrong, and are quite a few tests which exercise it.  It only goes onto the wire as a `uint32` (and thus gets serialized in order to form the block hash, which is likely why that hasn't been changed).

davecgh:
I just double checked and Core uses `int64` for all of the timestamp math operations as expected.  It would have to be that way, otherwise the consensus timestamp math would be wrong, and are quite a few tests which exercise it.  It only goes onto the wire as a `uint32` (and thus gets serialized in order to form the block hash, which is likely why that hasn't been changed).

```C++
int64_t GetBlockTime() const
```

jrick:
cool, that sounds reasonable and makes this change a bit more consistent

blockchain/difficulty.go
```diff
@@ -244,7 +244,8 @@ func (b *BlockChain) calcNextRequiredDifficulty(lastNode *blockNode, newBlockTim
 			// Return minimum difficulty when more than the desired
 			// amount of time has elapsed without mining a block.
 			reductionTime := b.chainParams.MinDiffReductionTime
```
jrick: change this too so that we don't need to use time.Time for math?

davecgh: Fair point.  I can change the others.  I mainly didn't do it because they are defined in terms of `time.Duration` which is nanosecond precision, so I was trying to touch as little as possible.  That is also why there is:

```Go
actualTimespan *= int64(time.Nanosecond)`
```

blockchain/difficulty.go
```diff
@@ -286,7 +287,8 @@ func (b *BlockChain) calcNextRequiredDifficulty(lastNode *blockNode, newBlockTim
 
 	// Limit the amount of adjustment that can occur to the previous
 	// difficulty.
-	actualTimespan := lastNode.timestamp.UnixNano() - firstNode.timestamp.UnixNano()
```
jrick: this removed code actually looks pretty scary considering it could overflow if the nanoseconds exceeds the max int64 value.  We sure this isn't a fork? (were we even correct before?)

davecgh: IIRC I checked that a long time ago and no overflows would happen until 2262, which is well past the point in 2106 where the network would break without changing the timestamp in the header anyways.

davecgh: All of this math is now done in terms of seconds, so even though it wasn't wrong before, it will now last much further into the future.

blockchain/difficulty.go
davecgh: I changed all of the other references and the math is now all performed in terms of seconds.

blockchain/difficulty.go
stevenroose: Assuming the int64 holds milliseconds instead of nanoseconds, doesn't this change the value of actualTimespan with a factor of 1000? It is later compared with `b.minRetargetTimespan`, which is not changed in this PR.

stevenroose: The line later `time.Duration(actualTimespan)*time.Second` suggests that the int64 holds seconds instead of milliseconds.

davecgh: @stevenroose:  They all hold seconds.  The end result is not changed because it's a ratio between the adjusted timespam and the target timespan and both are in seconds.

I'm not sure what you mean that `b.minRetargetTimespan` isn't changed.  It is defined as:

```Go
minRetargetTimespan: targetTimespan / adjustmentFactor,
```
 And `targetTimespan` was updated to:

```Go
targetTimespan := int64(params.TargetTimespan / time.Second)
```

stevenroose: Oh yeah, I should have been less lazy and checked the definition of `minRetargetTimespan`, but I go through the changes via e-mail in GitHub, instead of checking them out.

No problem, I really think this project is great. I should try to make more time to contribute, though :)

PR #916, opened 2017-02-01
blockchain: Refactor main block index logic.

davecgh:
This refactors the block index logic into a separate struct and introduces an individual lock for it so it can be queried independent of the chain lock.


dajohi:
OK

PR #917, opened 2017-02-01
blockchain: Use hash values in structs.

davecgh:
**This PR requires #916**

This modifies the `blockNode` and `BestState` structs in the `blockchain` package to store hashes directly instead of pointers to them and updates callers to deal with the API change in the exported `BestState` struct.

In general, the preferred approach for hashes moving forward is to store hash values in complex data structures, particularly those that will be used for cache entries, and accept pointers to hashes in arguments to functions.

Some of the reasoning behind making this change is:

- It is generally preferred to avoid storing pointers to data in cache objects since doing so can easily lead to storing interior pointers into other structs that then can't be GC'd
- Keeping the hash values directly in the block node provides better cache locality

### DIFF
--- a/blockchain/blockindex.go
+++ b/blockchain/blockindex.go
@@ -31,13 +31,13 @@ type blockNode struct {
 	children []*blockNode
 
 	// hash is the double sha 256 of the block.
-	hash *chainhash.Hash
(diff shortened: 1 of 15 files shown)


dajohi:
OK

PR #919, opened 2017-02-02
blockchain: Convert to full block index in mem.

davecgh:
**This requires PR #917**

This reworks the block index code such that it loads all of the headers in the main chain at startup and constructs the full block index accordingly.

Since the full index from the current best tip all the way back to the genesis block is now guaranteed to be in memory, this also removes all code related to dynamically loading the nodes and updates some of the logic to take advantage of the fact traversing the block index can longer potentially fail.  There are also more optimizations and simplifications that can be made in the future as a result of this.

Due to removing all of the extra overhead of tracking the dynamic state, and ensuring the block node structs are aligned to eliminate extra padding, the end result of a fully populated block index now takes quite a bit less memory than the previous dynamically loaded version.

The main downside is that it now takes a while to start whereas it was nearly instant before, however, it is much better to provide more efficient runtime operation since that is its ultimate purpose and the benefits far outweigh this downside.

Some benefits are:

- Since every block node is in memory, the recent code which reconstructs headers from block nodes means that all headers can always be served from memory which is important since the majority of the network has moved to header-based semantics
- Several of the error paths can be removed since they are no longer necessary
- It is no longer expensive to calculate CSV sequence locks or median times of blocks way in the past
- It will be possible to create much more efficient iteration and simplified views of the overall index
- The entire threshold state database cache can be removed since it is cheap to construct it from the full block index as needed

An overview of the logic changes are as follows:

- Move `AncestorNode` from `blockIndex` to `blockNode` and greatly simplify since it no longer has to deal with the possibility of dynamically loading nodes and related failures
- Rename `RelativeNode` to `RelativeAncestor`, move to `blockNode`, and redefine in terms of `AncestorNode`
- Move `CalcPastMedianTime` from `blockIndex` to `blockNode` and remove no longer necessary test for nil
- Change `calcSequenceLock` to use `Ancestor` instead of `RelativeAncestor` since it reads more clearly

### DIFF
--- a/blockchain/chainio.go
+++ b/blockchain/chainio.go
@@ -1081,7 +1081,7 @@ func (b *BlockChain) createChainState() error {
 	// Create a new node from the genesis block and set it as the best node.
 	genesisBlock := btcutil.NewBlock(b.chainParams.GenesisBlock)
 	header := &genesisBlock.MsgBlock().Header
-	node := newBlockNode(header, genesisBlock.Hash(), 0)
+	node := newBlockNode(header, 0)
 	node.inMainChain = true
 	b.bestNode = node
 
@@ -1168,6 +1168,44 @@ func (b *BlockChain) initChainState() error {
 			return err
 		}
 
+		// Load all of the headers from the data for the known best
+		// chain and construct the block index accordingly.  Since the
+		// number of nodes are already known, perform a single alloc
+		// for them versus a whole bunch of little ones to reduce
+		// pressure on the GC.
+		log.Infof("Loading block index.  This might take a while...")
+		bestHeight := int32(state.height)
+		blockNodes := make([]blockNode, bestHeight+1)
+		for height := int32(0); height <= bestHeight; height++ {
+			header, err := dbFetchHeaderByHeight(dbTx, height)
+			if err != nil {
+				return err
+			}
+
+			// Initialize the block node for the block, connect it,
+			// and add it to the block index.
+			node := &blockNodes[height]
+			initBlockNode(node, header, height)
+			if parent := b.bestNode; parent != nil {
+				node.parent = parent
+				node.workSum = node.workSum.Add(parent.workSum,
+					node.workSum)
+			}
+			node.inMainChain = true
+			b.index.AddNode(node)
+
+			// This node is now the end of the best chain.
+			b.bestNode = node
+		}
+
+		// Ensure the resulting best node matches the stored best state
+		// hash.
+		if b.bestNode.hash != state.hash {
+			return AssertError(fmt.Sprintf("initChainState: block "+
+				"index chain tip %s does not match stored "+
+				"best state %s", b.bestNode.hash, state.hash))
+		}
+
 		// Load the raw block bytes for the best block.
 		blockBytes, err := dbTx.FetchBlock(&state.hash)
 		if err != nil {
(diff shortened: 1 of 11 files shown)


blockchain/chainio.go
```diff
@@ -1167,6 +1167,44 @@ func (b *BlockChain) initChainState() error {
 			return err
 		}
 
+		// Load all of the headers from the data for the known best
+		// chain and construct the block index accordingly.  Since the
+		// number of nodes are already known, perform a single alloc
+		// for them versus a whole bunch of little ones to reduce
+		// pressure on the GC.
```
Roasbeef: Nice memory optimization! Packing them the raw values into a single slice (rather than using a slice of pointers) will also indirectly improve performance as accesses will have better cache locality.

Roasbeef: A similar memory optimization (reducing allocations) could also be done within the index which maps a blockhash to a `blockNode`.

davecgh: Thanks!  I'm not following what you mean since the index is using a map where the entries are a pointer to an existing block node.  One of the big things this PR aims to do is make it so there is only ever one single block node per block in memory.  That will allow them to be compared by pointer in other code I have in mind which is significantly faster than eating derefs + hash comparisons.

davecgh:
@Roasbeef:  Thanks for the initial review.  The main bug that I was referring to with that is the difficulty retarget isn't being calculated right.  If you try to do an IBD on this PR with checkpoints disabled you'll notice it'll incorrectly rejecting blocks.  I also wrote this very quickly and haven't gone through it with a fine tooth comb yet, so I'm not confident there aren't others.

The main reason I wanted to get a PR open is to get feedback on the direction.  The slower startup time is a pretty significant downside, however the runtime gains are quite substantial both in performance and reduced memory usage as compared to both cases having all nodes loaded.  It seems to me that runtime performance is much more important than that startup time even though, as developers, those longer startup times are really not fun to deal with.

davecgh:
Updated to resolve the aforementioned bug regarding the difficulty retarget.  This still requires more vetting, but it should be bug free now.

dan-da:
> the full index from the current best tip all the way back to the genesis block is now guaranteed to be in memory,

Just curious, but doesn't this mean that the amount of memory required to run btcd will increase approx linearly with the number of historical transactions?   Is that not a concern, especially for memory constrained devices?

davecgh:
@dan-da: It will increase linearly for each block, not the number of txns.  Since there is only one block per 10 minutes, which equates to about 52560 blocks per year, and a block header is 80 bytes, it would only grow by roughly 4-4.5MB per year.

However, I mentioned in the description, the fully loaded index with this PR actually takes less memory than it does currently "due to removing all of the extra overhead of tracking the dynamic state, and ensuring the block node structs are aligned to eliminate extra padding."

dan-da:
ah sorry, I didn't read carefully enough.  I was thinking of the addrindex.

davecgh:
Rebased to latest master.

jimpo:
@davecgh What is the behavior when a reorg happens when the node is off but has some of the side blocks saved to the DB? Like, the node is running, it saves some side blocks, then shuts off and loses the in-memory index. Then a reorg happens and those side blocks become part of the longest chain. When the node boots back up, it loads the (old) active chain into the index, but not the side blocks.

From what I can tell, the newer blocks on the main chain will appear as orphans because `maybeAcceptBlock` only checks the in-memory index, and the side blocks that have already been validated will be re-requested, whereas previously they would have been lazily loaded into the index.

davecgh:
@jimpo That behavior is unchanged because the block index in the database has only ever had the main chain currently.  That is to say that any reorgs remove the entries for side chain blocks from the block index.    If we take a look at the old `loadBlockNode` function, one of the first things it does is `dbFetchHeightByHash` which won't be able to find any entries for side chain blocks since they were removed.

You are correct to note that this behavior is not as efficient as it should be and in fact there is a comment that alludes to this in the code [here](https://github.com/btcsuite/btcd/blob/bc36cf51c6aec42ddbfc4585107a87ca26ba9815/blockchain/process.go#L59-L67).

```Go
		// Ignore side chain blocks in the database.  This is necessary
		// because there is not currently any record of the associated
		// block index data such as its block height, so it's not yet
		// possible to efficiently load the block and do anything useful
		// with it.
		//
		// Ultimately the entire block index should be serialized
		// instead of only the current main chain so it can be consulted
		// directly.
```

PR #920, opened 2017-02-02
blockchain: Remove threshold state db cache.

davecgh:
**This PR requires #919**

This completely removes the threshold state database caching code since it can very quickly be calculated at startup now that the entire block index is loaded first.


davecgh:
Rebased.

PR #922, opened 2017-02-04
blockchain: Optimize checkpoint handling.

davecgh:
**This requires PR #919**

This modifies the code that determines the most recently known checkpoint to take advantage of recent changes which make the entire block index available in memory by only storing a reference to the specific node in the index that represents the latest known checkpoint.

Previously, the entire block was stored and new checkpoints required loading it from the database.


davecgh:
Rebased.

PR #924, opened 2017-02-09
multi: Implement block size hard fork demo voting.

davecgh:
** This requires PR decred/dcrd#542**

This implements a new voting agenda for the testnet and simnet networks for increasing the maximum block size from 1MiB to 1.25MB.

- Bumps the maximum protocol size limit to 1.25MB
- Updates the `MaxSigOpsPerBlock` constant to maintain existing value regardless of the raised max protocol block size
- Decouples the maximum tx size from the block size by creating a chain parameter for it with the current sizes so it is not a part of the hard fork vote
- Adds definition for new version 4 stake vote along with agenda to vote on block size to the testnet and simnet networks
- Converts the `MaximumBlockSize` chain parameter to a slice in order to hold the different possible sizes
- Adds a new function that returns the maximum block size based upon the result of the vote
- Changes the existing context-free block size check to enforce the max protocol size and introduces a context-specific check which enforces a restricted size based upon the network consensus parameters and the result of the vote
- Sets the max allowed size in generated block templates based upon the network params and result of the vote


davecgh:
Wrong repository.

PR #931, opened 2017-03-21
main: Improve comment for headers peer message.

alexbosworth:


davecgh:
NOTE: The Travis failure is a false positive due to a lint issue unrelated to this PR.

stevenroose:
Yeah I also noticed the comment and that it was a bit weird that it mentioned when the method was used.

PR #933, opened 2017-03-22
TravisCI: Increase linting deadline.

davecgh:
This increases the deadline for the linters to run to 10 minutes and also reduces the number of threads that is uses.  This is being done because the Travis environment has become increasingly slower and it
also seems to be hampered by too many threads running concurrently.


davecgh:
Closing since I combined this with #935 in order to ensure both the deadline change and the lint failure are addressed.

PR #936, opened 2017-04-14
config: Add user agent comments flag --uacomment

stevenroose:
Just like Core's -uacomment, this flag allows to specify user agent
comments like defined in BIP 14.

### DIFF
--- a/config.go
+++ b/config.go
@@ -141,6 +141,7 @@ type config struct {
 	BlockMinSize         uint32        `long:"blockminsize" description:"Mininum block size in bytes to be used when creating a block"`
 	BlockMaxSize         uint32        `long:"blockmaxsize" description:"Maximum block size in bytes to be used when creating a block"`
 	BlockPrioritySize    uint32        `long:"blockprioritysize" description:"Size in bytes for high-priority/low-fee transactions when creating a block"`
+	UserAgentComments    []string      `long:"uacomment" description:"Comment to add to the user agent -- See BIP 14 for more information."`
 	NoPeerBloomFilters   bool          `long:"nopeerbloomfilters" description:"Disable bloom filtering support"`
 	SigCacheMaxSize      uint          `long:"sigcachemaxsize" description:"The maximum number of entries in the signature verification cache"`
 	BlocksOnly           bool          `long:"blocksonly" description:"Do not accept transactions from remote peers."`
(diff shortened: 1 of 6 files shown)


peer/peer.go
```diff
@@ -803,7 +808,9 @@ func (p *Peer) localVersionMsg() (*wire.MsgVersion, error) {
 
 	// Version message.
 	msg := wire.NewMsgVersion(ourNA, theirNA, nonce, blockNum)
-	msg.AddUserAgent(p.cfg.UserAgentName, p.cfg.UserAgentVersion)
+	msg.AddUserAgent(p.cfg.UserAgentName,
+		p.cfg.UserAgentVersion,
```
davecgh: Please keep this on the original line and add the new parameter on this new line so it is consistent with the rest of the code base.

```Go
msg.AddUserAgent(p.cfg.UserAgentName, p.cfg.UserAgentVersion,
	p.cfg.UserAgentComments...)
```

sample-btcd.conf
```diff
@@ -156,6 +156,9 @@
 ; Add additional checkpoints. Format: '<height>:<hash>'
 ; addcheckpoint=<height>:<hash>
 
+; Add comments to the User Agent that is advertised to peers.
+; Should not include characters '/', ':', '(' and ')'.
```
davecgh: Since this produces an error in the code, it should say "Must not include...".  Should implies it is optional to do so which is not the case.

config.go
```diff
@@ -736,6 +737,18 @@ func loadConfig() (*config, []string, error) {
 	cfg.BlockPrioritySize = minUint32(cfg.BlockPrioritySize, cfg.BlockMaxSize)
 	cfg.BlockMinSize = minUint32(cfg.BlockMinSize, cfg.BlockMaxSize)
 
+	// Look for illegal characters in the user agent comments.
+	for _, uaComment := range cfg.UserAgentComments {
+		if strings.ContainsAny(uaComment, "/:()") {
+			err := fmt.Errorf("%s: The following characters should not "+
```
davecgh: s/should not/must not/

config.go
```diff
@@ -736,6 +737,18 @@ func loadConfig() (*config, []string, error) {
...
+				"appear in User Agent comments: '/', ':', '(', ')'",
```
davecgh: s/User Agent/user agent/

stevenroose: I changed it to lower case everywhere.

stevenroose:
Hmm, an error from my part. The characters `: ; ( )` are the general reserved characters for the whole user agent. The comments are allowed to have : and ; and actually ( and ) too as long as they don't interfere with the surrounding ones. But I'll just change this to disallowing brackets.

stevenroose:
Oh mistake from my part, I thought for a second the semicolon was in that list too. The only one that we might allow is the colon, the specification is not clear about that. I will leave it for now, let me know if you think we should allow the colon in the comments.

davecgh:
I'd prefer to leave it the way it and and disallow them.  It's ok, and desirable, in my opinion, to be more strict.

doc.go
```diff
@@ -77,6 +77,8 @@ Application Options:
       --addcheckpoint=      Add a custom checkpoint.  Format: '<height>:<hash>'
       --nocheckpoints       Disable built-in checkpoints.  Don't do this unless
                             you know what you're doing.
+      --aucomment=          Comments to add to the user agent --
```
davecgh: uacomment

stevenroose: Whoops, that's a typo, good catch!

stevenroose: I also misspelled (plural) it in config.go. All are not --uacomment, like the bitcoind flag. (It's a multi-flag, so can have single comment per flag.)

davecgh:
I'm going to test this out shortly.  The code reads properly now.

stevenroose:
Yeah maybe a user wants to have direct control over the whole comment string (f.e. the order in which they occur) and wants to provide all the semicolon-separated parts.

PR #939, opened 2017-05-03
Fix docs/README.md

esemplastic:
Invalid tokens: github had made some changes the past months, the old style is not rendering at all, so I fixed that for you. 

Other contributors can do the same for all of the project's documents.

**Thanks.**


esemplastic:
You're welcome @davecgh and thanks because you did a very good job for the project, although I'm not using btcd yet but I have future plans using it:)

PR #956, opened 2017-05-30
Fix warnings from ineffassign

stevenroose:
Some suggestions by [ineffassign](https://github.com/gordonklaus/ineffassign) that looks for unnecessary assignments.

I left two at the end of fullblocktest, since I suspected the unused
variable assignments there were set for the possibility of extending the
tests.

```
btcd/connmgr/tor.go
Line 107: warning: err assigned and not used (ineffassign)
btcd/server.go
Line 2227: warning: err assigned and not used (ineffassign)
btcd/txscript/script.go
Line 352: warning: idx assigned and not used (ineffassign)
btcd/txscript/sign.go
Line 193: warning: err assigned and not used (ineffassign)
btcd/addrmgr/knownaddress.go
Line 45: warning: lastSeen assigned and not used (ineffassign)
btcd/blockchain/fullblocktests/generate.go
Line 2131: warning: chain1TipName assigned and not used (ineffassign)
```

### DIFF
--- a/connmgr/tor.go
+++ b/connmgr/tor.go
@@ -99,7 +99,7 @@ func TorLookupIP(host, proxy string) ([]net.IP, error) {
 		return nil, ErrTorInvalidProxyResponse
 	}
 	if buf[1] != 0 {
-		if int(buf[1]) > len(torStatusErrors) {
+		if int(buf[1]) >= len(torStatusErrors) {
 			return nil, ErrTorInvalidProxyResponse
 		} else if err := torStatusErrors[buf[1]]; err != nil {
 			return nil, err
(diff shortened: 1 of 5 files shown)


connmgr/tor.go
```diff
@@ -100,14 +100,11 @@ func TorLookupIP(host, proxy string) ([]net.IP, error) {
 	}
 	if buf[1] != 0 {
 		if int(buf[1]) > len(torStatusErrors) {
```
stevenroose: On second review, should this be `>=`?

davecgh: No.  Slice indexes are zero based.

jrick: it should be >=.  Indexing the len(slice)'th element always panics.

stevenroose: That's what I was thinking. This allows to do `slice[len(slice)]` which will panic.

PR #958, opened 2017-05-31
all: Remove seelog logger.

jrick:
The btclog package has been changed to defining its own logging
interface (rather than seelog's) and provides a default implementation
for callers to use.

There are two primary advantages to the new logger implementation.

First, all log messages are created before the call returns.  Compared
to seelog, this prevents data races when mutable variables are logged.

Second, the new logger does not implement any kind of artifical rate
limiting (what seelog refers to as "adaptive logging").  Log messages
are outputted as soon as possible and the application will appear to
perform much better when watching standard output.

Because log rotation is not a feature of the btclog logging
implementation, it is handled by the main package by importing a file
rotation package that provides an io.Reader interface for creating
output to a rotating file output.  The rotator has been configured
with the same defaults that btcd previously used in the seelog config
(10MB file limits with maximum of 3 rolls) but now compresses newly
created roll files.  Due to the high compressibility of log text, the
compressed files typically reduce to around 15-30% of the original
10MB file.


jrick:
I don't know what is causing those test errors but i see them as well locally on master.

jrick:
Also, this PR currently is using my forks for the btclog and btcrpcclient packages and those should be merged first before this PR lands.

jrick:
Found the cause of the failure, and not sure why i thought i saw it on master, because it does look related to the logger change:

```
failed to create file rotator: open C:\Users\jrick\AppData\Local\Temp\rpctest-logs233740733\simnet\btcd.log: The system cannot find the path specified.
```

(btw it would have been great if stdout/stderr was redirected or at least read for the btcd process in the integration tests)

PR #959, opened 2017-06-07
Add human-readable part for segwit addresses to chaincfg.Params

halseth:
To be used together with https://github.com/btcsuite/btcutil/pull/100

### DIFF
--- a/chaincfg/params.go
+++ b/chaincfg/params.go
@@ -190,6 +190,10 @@ type Params struct {
 	// Mempool parameters
 	RelayNonStdTxs bool
 
+	// Human-readable part for Bech32 encoded segwit addresses, as defined in
+	// BIP 173.
+	Bech32HRPSegwit string
+
 	// Address encoding magics
 	PubKeyHashAddrID byte // First byte of a P2PKH address
 	ScriptHashAddrID byte // First byte of a P2SH address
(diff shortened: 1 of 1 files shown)


stevenroose:
I would prefer a more declarative name than `Hrp`. First of all, abbreviations are supposed to be all upper case, so it'd be `HRP`, but I'm more in favor of something like `Bech32SegwitHRP`.

halseth:
Good point! I wonder if maybe `SegwitBech32HRP` would be even better, to emphasize that this is for segwit addresses? (and since `hrp` is Bech32 terminology I think it makes sense to have it next to `Bech32` 😄 )

stevenroose:
Well, if there will be more prefixes in the future, it's nice to have them looking similar, that's why those are often named hierarchical. Bech32HRPSegWit is hierarchical

halseth:
Agreed, done.

stevenroose:
(You can squash both commits into the original one like this:
```
git reset HEAD~1
git commit --amend
git push -f
```
So we don't need two commits for this small a change :) )

halseth:
Done 👍

chaincfg/params.go
```diff
@@ -476,6 +492,10 @@ var SimNetParams = Params{
 	// Mempool parameters
 	RelayNonStdTxs: true,
 
+	// Human-readable part for Bech32 encoded segwit addresses, as defined in
+	// BIP 173.
+	Bech32HRPSegwit: "tb", // always tb for test net
```
davecgh: Can we make this "sb" (and adjust the comments accordingly)?  We don't want simnet to be able to use testnet addresses.  Hence why the p2pkh addrs start with S (just below).

halseth: I guess you are referring to `SimNetParams`? Just pushed the requested change there :)

davecgh: I was, yep.  I commented on that specific line.  Thanks for the update!

PR #960, opened 2017-06-07
btcec: Optimize normalize and NAF, correct normalize, and add tests.

### LINKED ISSUE #709: btcec: recovers certain signature without error

davecgh:
This modifies the normalize function of the internal field value to both optimize it and address an issue where the reduction could lead to an incorrect result with a small range of values along with some tests to ensure the behavior is correct.

It also contains contributions from @jimmysong to optimize NAF to avoid returning the unneeded bit when there is not a carry and adds a bunch of unit tests.

The following benchmark shows the relative speedups as a result of the optimization on my system.  In particular, the changes result in approximately a 14% speedup in Normalize, which ultimately translates to a 2% speedup in signature verifies.

```
benchmark                        old ns/op     new ns/op     delta
--------------------------------------------------------------------
BenchmarkAddJacobian             1364          1289          -5.50%
BenchmarkAddJacobianNotZOne      3150          3091          -1.87%
BenchmarkScalarBaseMult          134117        132816        -0.97%
BenchmarkScalarBaseMultLarge     135067        132966        -1.56%
BenchmarkScalarMult              411218        402217        -2.19%
BenchmarkSigVerify               671585        657833        -2.05%
BenchmarkFieldNormalize          36.0          31.0          -13.89%
```

Fixes #709.

### DIFF
--- a/btcec/bench_test.go
+++ b/btcec/bench_test.go
@@ -111,3 +111,13 @@ func BenchmarkSigVerify(b *testing.B) {
 		sig.Verify(msgHash.Bytes(), &pubKey)
 	}
 }
+
+// BenchmarkFieldNormalize benchmarks how long it takes the internal field
+// to perform normalization (which includes modular reduction).
+func BenchmarkFieldNormalize(b *testing.B) {
+	// The normalize function is constant time so default value is fine.
+	f := new(fieldVal)
+	for i := 0; i < b.N; i++ {
+		f.Normalize()
+	}
+}
--- a/btcec/btcec.go
+++ b/btcec/btcec.go
@@ -747,9 +747,9 @@ func NAF(k []byte) ([]byte, []byte) {
 	}
 	if carry {
 		retPos[0] = 1
+		return retPos, retNeg
 	}
-
-	return retPos, retNeg
+	return retPos[1:], retNeg[1:]
 }
 
 // ScalarMult returns k*(Bx, By) where k is a big endian integer.
--- a/btcec/btcec_test.go
+++ b/btcec/btcec_test.go
@@ -65,7 +65,6 @@ func TestAddJacobian(t *testing.T) {
 			"131c670d414c4546b88ac3ff664611b1c38ceb1c21d76369d7a7a0969d61d97d",
 			"1",
 		},
-
 		// Addition with z1=z2=1 different x values.
 		{
 			"34f9460f0e4f08393d192b3c5133a6ba099aa0ad9fd54ebccfacdfa239ff49c6",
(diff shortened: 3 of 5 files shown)


btcec/bench_test.go
```diff
@@ -111,3 +111,13 @@ func BenchmarkSigVerify(b *testing.B) {
...
 	}
 }
+
+// BenchmarkFieldNormalize benchmarks how long it takes the internal field
+// to perform normalization (which includes modular reduction).
+func BenchmarkFieldNormalize(b *testing.B) {
+	// The normalize function is constant time so default value is fine.
+	f := new(fieldVal)
+	for i := 0; i < b.N; i++ {
+		_ = f.Normalize()
```
jimmysong: I don't think you need the `_ = `

davecgh: Correct.  I'll nuke it.

PR #968, opened 2017-07-07
chaincfg: Add IsBech32SegwitPrefix utility method.

halseth:
The IsBech32SegwitPrefix method takes a string prefix and
determines if it is a valid prefix for a Bech32 encoded segwit
address for any of the default or registered networks.


halseth:
To be used by https://github.com/btcsuite/btcutil/pull/100

PR #973, opened 2017-08-02
multi: Optimize by removing defers

davecgh:
Upstream commit b1621332cc6d5860f58a7526088569e92af05ae0.

---

Defers are nice for readability but they do add overhead.  This gets rid of defers where it is just as easy as not to use one.


davecgh:
Wrong repo.

Issue #978, opened 2017-08-10
Move blockManager into it's own sync package


jimpo:
blockManager implements the blockchain syncing protocol in addition to adding/removing transactions from the mempool in response to peer notifications. This is a lot of important logic that in contained in the main package without unit tests.

I propose creating a `sync` package that is responsible for keeping the blockchain and transaction mempool in sync with the peers. It will essentially have the responsibilities of the current block manager, though some logic will be moved around to decouple it from `server`. A primary goal of the refactor is to add more test coverage.

Once this logic is encapsulated in a package, further improvements should be easier such as:
- Headers first syncing (past checkpoints)
- Parallel block downloading
- Accepting BIP 130 direct header announcements
- BIP 152 support

Steps:
- Remove dependency of blockManager on server
- Either remove references in blockManager interface to serverPeer or make public ServerPeer interface
- Move blockmanager.go and blocklogger.go into sync package
- Reify sync package public interface
- Add unit test coverage
- ...
- Profit

davecgh:
So, my first though is that I completely agree that block manager could use a bit of decoupling.  As an aside, the same is true of the rpcserver package (and I have a local branch that has some work done towards that end, but it woefully out of date now).

Something that I have wanted to get to for quite some time now, but just have never had a chance is multi-peer syncing.  That is precisely parallel block downloading, full headers first syncing, BIP130 and BIP152.  It should be noted that it already does do some header-first syncing when before the last checkpoint where it downloads the headers up to the next checkpoint, then the blocks, then the next set of headers up to the next checkpoint, etc.  It would indeed be better to switch over to a full headers first sync where it doesn't rely on checkpoint.

That said, there are quite a few things that need to happen for that to be viable.  Specifically in regards to how the database and blockchain packages handle the block index.  A little more concretely, in order to support parallel block download there needs to be the ability to keep track of different "states" for blocks in the database which is something that does not currently exist.  The database/chain package always assumes that the blocks are connected in order.

I'm definitely on board with the idea.

jrick:
I had the exact same idea of improving the btcwallet code by using sync packages (there would be at least two there, for rpc and spv), so I'm also in favor of this.

PR #979, opened 2017-08-11
blockmanager: Decouple from server.

jimpo:
This PR removes some unnecessary references from blockManager to server.

The blockManager uses the server to obtain a reference to the txMemPool, which it interacts with many times. Since one of the responsibilities of the blockManager is to keep the mempool in sync, it should have store its own reference. 

The blockManager also sends notifications to the RPC server, in response to chain events. I don't think the blockManager should know anything about the RPC server, so instead the RPC server will listen directly to notifications emitted from the BlockChain.

https://github.com/btcsuite/btcd/issues/978

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -9,7 +9,6 @@ import (
 	"net"
 	"os"
 	"path/filepath"
-	"sort"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -623,14 +622,6 @@ func (b *blockManager) handleBlockMsg(bmsg *blockMsg) {
 
 		// Clear the rejected transactions.
 		b.rejectedTxns = make(map[chainhash.Hash]struct{})
-
-		// Allow any clients performing long polling via the
-		// getblocktemplate RPC to be notified when the new block causes
-		// their old block template to become stale.
-		rpcServer := b.server.rpcServer
-		if rpcServer != nil {
-			rpcServer.gbtWorkState.NotifyBlockConnected(blockHash)
-		}
 	}
 
 	// Update the block height for this peer. But only send a message to
@@ -1143,14 +1134,6 @@ out:
 					}
 				}
 
-				// Allow any clients performing long polling via the
-				// getblocktemplate RPC to be notified when the new block causes
-				// their old block template to become stale.
-				rpcServer := b.server.rpcServer
-				if rpcServer != nil {
-					rpcServer.gbtWorkState.NotifyBlockConnected(msg.block.Hash())
-				}
-
 				msg.reply <- processBlockResponse{
 					isOrphan: isOrphan,
 					err:      nil,
@@ -1177,9 +1160,10 @@ out:
 	bmgrLog.Trace("Block handler done")
 }
 
-// handleNotifyMsg handles notifications from blockchain.  It does things such
-// as request orphan block parents and relay accepted blocks to connected peers.
-func (b *blockManager) handleNotifyMsg(notification *blockchain.Notification) {
+// handleBlockchainNotification handles notifications from blockchain.  It does
+// things such as request orphan block parents and relay accepted blocks to
+// connected peers.
+func (b *blockManager) handleBlockchainNotification(notification *blockchain.Notification) {
 	switch notification.Type {
 	// A block has been accepted into the block chain.  Relay it to other
 	// peers.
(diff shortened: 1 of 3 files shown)


server.go
```diff
@@ -1048,6 +1050,13 @@ func (s *server) AnnounceNewTransactions(newTxs []*mempool.TxDesc) {
 	}
 }
 
+// Transaction has one confirmation on the main chain. Now we can mark it as no
+// longer needing rebroadcasting.
+func (s *server) TransactionConfirmed(tx *btcutil.Tx) {
+	iv := wire.NewInvVect(wire.InvTypeTx, tx.Hash())
```
davecgh: This doesn't need to be done unless the RPC server is active because the only time rebroadcasting is needed is for transactions that were submitted via RPC. That check seems to have been lost in the shuffle.

I think it's quite reasonable to have this call separate to decouple that logic, but really should have the check for a nil RPC server (aka no RPC server is running) moved into it.

The following should do the trick nicely:


```Go
// No need to work with rebroadcasting when the RPC server is not active.
if s.rpcServer == nil {
	return
}
```

blockmanager.go
```diff
@@ -1151,26 +1136,14 @@ func (b *blockManager) handleNotifyMsg(notification *blockchain.Notification) {
...
 		// valid.
 		for _, tx := range block.Transactions()[1:] {
-			b.server.txMemPool.RemoveTransaction(tx, false)
-			b.server.txMemPool.RemoveDoubleSpends(tx)
-			b.server.txMemPool.RemoveOrphan(tx)
-			acceptedTxs := b.server.txMemPool.ProcessOrphans(tx)
+			b.txMemPool.RemoveTransaction(tx, false)
+			b.txMemPool.RemoveDoubleSpends(tx)
+			b.txMemPool.RemoveOrphan(tx)
+			b.server.TransactionConfirmed(tx)
```
davecgh: In the comment for `newBlockManager`, I highly recommended adding a config struct for the block manager.  This function would become a callback in said config struct in order to break the need to directly access `server`.

blockmanager.go
```diff
@@ -1151,26 +1136,14 @@ func (b *blockManager) handleNotifyMsg(notification *blockchain.Notification) {
...
+			acceptedTxs := b.txMemPool.ProcessOrphans(tx)
 			b.server.AnnounceNewTransactions(acceptedTxs)
```
davecgh: Along the same lines, this should also be a callback in the newly created config struct.

davecgh:
Also, for reference, it would be useful to take a look at #981 that I just pushed.  In particular, that PR changes the `AnnounceNewTransactions` to split the logic up such that the peer relay and RPC notification portions are separate.

server.go
jimpo: Ah, I didn't think that was intentional. I'll add the check back.

jimpo:
Thanks for the feedback!

With regards the remaining calls to `server`, you are right that the four methods invoked on server are `AnnounceNewTransactions`, `RelayInventory`, `UpdatePeerHeights`, and `TransactionConfirmed`. My intention was to create a new interface in the upcoming `netsync` package that the blockManager would know about called `PeerNotifier` with those methods exposed. Then we could continue passing in the server to the blockManager as an instance of that interface. I didn't think of the config approach. Which is the preferred way to go?

There's one more point of coupling, which is that methods on blockManager accept and return `serverPeer`s. I thought of using a similar approach and, for the time being until things can be untangled more, creating a ServerPeer interface in the new package that `serverPeer` implements and which exposes only the methods that blockManager needs.

davecgh:
In regards to the block manager still taking the server, perhaps I misunderstood, but I thought you ultimately goal was to replace it with the `netsync` package.  Since the `server` is not itself a separate package, you won't be able to directly pass it.  The idea of a `PeerNotifier` is sound.  I was thinking you'd first take the callbacks in via the config and then later refactor it out into an interface in the `netsync` package, but I see you've already went ahead and updated this PR with that change accordingly.

In regards to the peers, I do think that probably warrants a larger discussion, and one that perhaps we should open a separate discussion for.  I had a similar issue with it in my `rpcserver` separation PR #981.  I introduced an `rpcserverPeer` interface there (that will ultimately become `rpcserver.Peer` when it move into a separate package), in order to get the ball rolling in terms of separation, but I'm not convinced my approach there is the best way to ultimately go.  On the plus side, it ensures that the RPC server expectations of the information available from a peer are well defined and can be stubbed out as necessary.  On the negative side, having to create an adapter specifically for that  seems a bit worky.

jimpo:
Yes, my plan is still to create a separate `netsync` package, and I will do that in a future PR. You would know better that I, but I assumed that if there was a public `PeerNotifier` interface defined in the `netsync` package and the config struct has one, then from `server.go`, I could create a config using a `*server` as an instance of `PeerNotifier`, even though it is private to the main package.

blockmanager.go
```diff
@@ -134,13 +133,33 @@ type headerNode struct {
 	hash   *chainhash.Hash
 }
 
+// The PeerNotifier exposes methods to notify peers of status changes to
```
davecgh: All comments should start with the name of the thing they are commenting per [Effective Go](https://golang.org/doc/effective_go.html) which this project follows.  In other words `PeerNotifier ...`.

blockmanager.go
```diff
@@ -134,13 +133,33 @@ type headerNode struct {
...
+	AnnounceNewTransactions(newTxs []*mempool.TxDesc)
+
+	UpdatePeerHeights(latestBlkHash *chainhash.Hash, latestHeight int32, updateSource *serverPeer)
+
+	RelayInventory(invVect *wire.InvVect, data interface{})
+
+	TransactionConfirmed(tx *btcutil.Tx)
+}
+
+// Configuration structure used to initialize a new blockManager.
```
davecgh: Same deal here.  This comment should start with `blockManagerConfig `.

davecgh:
Would you rebase this please?  It has conflicts right now and won't merge cleanly.

PR #980, opened 2017-08-11
blockchain: Allow multiple subscribers to notifications.

jimpo:
The BlockChain struct emits notifications for various events, but it is only possible to register one listener. This changes the interface and implementations to allow multiple listeners.

https://github.com/btcsuite/btcd/pull/979/files#diff-34c6b408d72845d076d47126c29948d1R2419

### DIFF
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -81,7 +81,6 @@ type BlockChain struct {
 	db                  database.DB
 	chainParams         *chaincfg.Params
 	timeSource          MedianTimeSource
-	notifications       NotificationCallback
 	sigCache            *txscript.SigCache
 	indexManager        IndexManager
 	hashCache           *txscript.HashCache
@@ -168,6 +167,11 @@ type BlockChain struct {
 	// being mined.
 	unknownRulesWarned    bool
 	unknownVersionsWarned bool
+
+	// The notifications field stores a slice of callbacks to be executed on
+	// certain blockchain events.
+	notificationsLock sync.RWMutex
+	notifications     []NotificationCallback
 }
 
 // DisableVerify provides a mechanism to disable transaction script validation
@@ -1308,15 +1312,6 @@ type Config struct {
 	// time is adjusted to be in agreement with other peers.
 	TimeSource MedianTimeSource
 
-	// Notifications defines a callback to which notifications will be sent
-	// when various events take place.  See the documentation for
-	// Notification and NotificationType for details on the types and
-	// contents of notifications.
-	//
-	// This field can be nil if the caller is not interested in receiving
-	// notifications.
-	Notifications NotificationCallback
-
 	// SigCache defines a signature cache to use when when validating
 	// signatures.  This is typically most useful when individual
 	// transactions are already being validated prior to their inclusion in
@@ -1385,7 +1380,7 @@ func New(config *Config) (*BlockChain, error) {
 		db:                  config.DB,
 		chainParams:         params,
 		timeSource:          config.TimeSource,
-		notifications:       config.Notifications,
+		notifications:       make([]NotificationCallback, 0),
 		sigCache:            config.SigCache,
 		indexManager:        config.IndexManager,
 		minRetargetTimespan: targetTimespan / adjustmentFactor,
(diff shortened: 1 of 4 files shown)


blockchain/chain.go
```diff
@@ -1304,15 +1304,6 @@ type Config struct {
...
 	TimeSource MedianTimeSource
 
-	// Notifications defines a callback to which notifications will be sent
-	// when various events take place.  See the documentation for
-	// Notification and NotificationType for details on the types and
-	// contents of notifications.
-	//
-	// This field can be nil if the caller is not interested in receiving
-	// notifications.
-	Notifications NotificationCallback
```
davecgh: I would much rather this be left as an option in the config struct and changed to an array.  That allows it to be treated as immutable and thus safe for concurrent access without the need to introduce a mutex.

The changes in this PR, as they are, are not safe for concurrent access since it would be possible to subscribe new listeners at any time and there is no mutex to protect it.  Not only would you need the mutex on the subscribe side, but you'd have to take it every time you want to range the `s.notifications` too.

blockchain/notifications_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	_, _, err = chain.ProcessBlock(blocks[1], blockchain.BFNone)
+	if err != nil {
+		t.Fatalf("ProcessBlock fail on block 1: %v\n", err)
+	}
+
+	select {
+	case notification := <-notifications:
+		if notification.Type != blockchain.NTBlockAccepted {
+			t.Error("Expected NTBlockAccepted notification, got %v", notification.Type)
```
davecgh: Errorf

blockchain/chain.go
jimpo: Happy to change it if you prefer. I did it this way because it makes more sense to me to have each listener subscribe individually that have something central configure them. For example, it doesn't seem necessary to me that the server know whether the RPC server wants to respond to blockchain events. Furthermore, there would never really be contention on a read-write lock since all of the writes to notifications are made at setup time and after the server is actually started, it's all reads. But it doesn't make a big difference, and I'm fine switching to the configuration approach.

davecgh:
Changes look good.  Thanks.  Ready for squash and I'll get it merged.

jimpo:
Actually, now that I think about it more, without this `Subscribe` functionality, this change is pretty pointless. If all callbacks are passed into the config, it is just as easy to aggregate multiple callbacks into a single function as it is to aggregate them into an array.

Again, the benefit of the `Subscribe` instead of the configuration approach is  that each listener register independently, without having a central module knowing. And as I noted in [this comment](https://github.com/btcsuite/btcd/pull/980#discussion_r132849752), the overhead of a read-write lock around the notifications array should be very low, since it is almost all reads.

I would say we should either revert to the `Subscribe` interface (and I'll make it concurrency-safe) or scrap this PR and leave as it is. Or I'm open to the idea of creating a new `events` package to replace the notifications emitted from `BlockChain`, which could also fire mempool events for and other interesting state changes.

davecgh:
I have no big objections to `Subscribe` as long as it's concurrent safe.  I do prefer the more immutable single shot setup so no locks are needed in general, but we aren't really talking about a lot of evens here anyways, so if you really prefer that approach, feel free to go for it.

davecgh:
Oh, in regards to an events package, I generally like to try and keep the packages as easily testable as possible.  That is why most of them take a config struct with callbacks and/or interfaces.  The blockchain package is actually one of the worst examples in this regard.  The mempool is a better example there as you can see in the tests that it creates a pool harness with stubbed out functions for things such as fetching utxos, the current best chain height, and the current median time.  This allows the tests to avoid needing to actually create a full blown chain instance (which in turn needs a database), having to actually make a bunch of real blocks and populate the chain, etc.

blockchain/chain.go
```diff
@@ -1385,7 +1380,7 @@ func New(config *Config) (*BlockChain, error) {
 		db:                  config.DB,
 		chainParams:         params,
 		timeSource:          config.TimeSource,
-		notifications:       config.Notifications,
+		notifications:       make([]NotificationCallback, 0),
```
davecgh: For future reference, there is no need to perform an extra unnecessary allocation here.  The `range` and `append` statements in Go work correctly with nil slices.  I'm not going to hold this PR up over it as there are a few other minor nitpicks that I figure I'll just push in a separate PR rather than bugging you about them (`Fatalf` in tests, making test function comment match the function name, test print exceeding col 80, avoiding deferred unlocks in non-trival cases since it has a pretty high overhead)

PR #656, opened 2016-04-08
Implement BIPs 141, 142, 143, 145 and 147 (Segregated Witness) 

Roasbeef:
This PR makes several modifications across `btcd`, and many of the sub-packages in order to add compatibility for the upcoming Segregated Witness soft-fork. As the set of changes is rather expansive, this PR has been split up into several commits to in order to facilitate thorough code review of each change. 

This PR currently implements the changes dictated by the following BIPS: 
- [BIP 141](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki)
- [BIP 142](https://github.com/bitcoin/bips/blob/master/bip-0142.mediawiki)
- [BIP 143](https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki)
- [BIP 145](https://github.com/bitcoin/bips/blob/master/bip-0145.mediawiki) 
- [BIP 147](https://github.com/bitcoin/bips/blob/master/bip-0147.mediawiki)
- [BIP 173](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki) 

As of the writing of this PR description, with this patch set `btcd` is able to successful sync up, and correctly validate all blocks and witness commitments on Bitcoin's`testnet`. This PR is still being extensively tested in order to ensure all new consensus critical decision related to transactions, blocks, and script validation are functioning properly. 

The changes across all packages are briefly summarized below:
- `blockchain`:
  - New functions have been added to calculate block, and transaction "cost". 
  - The previous `BuildMerkleTreeStore` function is now able to calculate the witness merkle root. This allowed the creation of new functions to extract, and validate the witness commitment. 
  - The new "cost" metrics are now used when validating blocks, and transactions. 
  - The `txscript.ScriptVerifyWitness` is now used by default when validating transaction scripts. Once BIP9 support is in place, this will instead be gated by the appropriate `versionbits` stage threshold. 
  - Script validation performed by `txValidator` now also uses the new `HashCache` in order to more efficiently validate inputs spending witness program outputs. 
  - When checking BIP34 for blocks below height 16, we now attempt to extract just a small integer. This behavior never came when BIP34 was rolled out, as it was deployed around height 100k or so.
- `chaincfg`: 
  - The parameters for `segnet4` have been added, along with the new address bytes for the new standard output types.
- `database`: 
  - Blocks, and transactions are always stored using the witness encoding if they have witness data. Additionally, all blocks are attempted to be decoded using the witness encoding. 
- `peer`: 
[...]

PR #981, opened 2017-08-13
rpcserver: Decouple from server.

davecgh:
**NOTE:  This will most certainly conflict with #656 and #979, so it will need to be rebased and updated accordingly after those are merged**

This decouples the RPC server from the internal btcd server to move closer to being able to split it out into a separate package.

In order to accomplish this, it introduces an `rpcserverConfig` type and several new interfaces, named `rpcserverPeer`, `rpcserverConnManager`, and `rpcserverBlockManager`, which are necessary to break the direct dependencies on the main server and block manager instances.

It also adds concrete implementations of the new interfaces and uses them to configure the RPC server.

Ultimately, the RPC server should ideally be decoupled even more such that all of the types in the configuration struct use interfaces instead of the concrete types.  Doing this would make the RPC server much easier to internally test since it would allow creating lightweight stubs for the various pieces.

### DIFF
--- a/server.go
+++ b/server.go
@@ -1072,28 +1072,28 @@ func (s *server) RemoveRebroadcastInventory(iv *wire.InvVect) {
 	s.modifyRebroadcastInv <- broadcastInventoryDel(iv)
 }
 
+// relayTransactions generates and relays inventory vectors for all of the
+// passed transactions to all connected peers.
+func (s *server) relayTransactions(txns []*mempool.TxDesc) {
+	for _, txD := range txns {
+		iv := wire.NewInvVect(wire.InvTypeTx, txD.Tx.Hash())
+		s.RelayInventory(iv, txD)
+	}
+}
+
 // AnnounceNewTransactions generates and relays inventory vectors and notifies
 // both websocket and getblocktemplate long poll clients of the passed
 // transactions.  This function should be called whenever new transactions
 // are added to the mempool.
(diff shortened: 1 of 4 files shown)


server.go
```diff
@@ -1023,28 +1023,28 @@ func (s *server) RemoveRebroadcastInventory(iv *wire.InvVect) {
...
 
-			// Potentially notify any getblocktemplate long poll clients
-			// about stale block templates due to the new transaction.
-			s.rpcServer.gbtWorkState.NotifyMempoolTx(
-				s.txMemPool.LastUpdated())
-		}
+	// Notify both websocket and getblocktemplate long poll clients of all
+	// newly accepted transactions.
+	if s.rpcServer != nil {
+		s.rpcServer.NotifyNewTransactions(txns)
```
jimpo: What do you think of inverting control here and having the RPC server listen for events from the mempool instead of the blockManager telling the server to tell the RPC server about the new transactions? I imagine there's a way along those lines to remove the `AnnounceNewTransactions` methods from `server` altogether.

I know the mempool does not currently emit events, so I'm not suggesting that be done as part of this change, but I'm interested for your thoughts on the concept.

davecgh: Yes, I agree that eventually it would probably be better to emit event from mempool.  My main goal with this PR was to do just enough to break the dependency on server.  There are still more dependencies that have to be broken before it can be split into a separate package too such as what #982 does and direct dependency on the CLI config.  I generally prefer to do these in stages to ease the reviews.

PR #989, opened 2017-08-15
blockmanager: Remove serverPeer references.

jimpo:
The purpose is to remove the dependency of blockmanager on serverPeer, which is defined in the main package. Instead, we split out some of the fields from serverPeer into a separate struct called peerSyncState in blockmanager.go. While they are in the same package now, this change makes it easier to move blockManager into its own package along with peerSyncState. The blockManager tracks a map of Peer pointers to the peer state and keeps it updated as peers connect and disconnect.

This further decouples blockManager from server on the path to https://github.com/btcsuite/btcd/issues/978.

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -16,6 +16,7 @@ import (
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/database"
 	"github.com/btcsuite/btcd/mempool"
+	peerpkg "github.com/btcsuite/btcd/peer"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcutil"
 )
@@ -49,14 +50,14 @@ var zeroHash chainhash.Hash
 
 // newPeerMsg signifies a newly connected peer to the block handler.
 type newPeerMsg struct {
-	peer *serverPeer
+	peer *peerpkg.Peer
 }
 
 // blockMsg packages a bitcoin block message and the peer it came from together
 // so the block handler has access to that information.
 type blockMsg struct {
 	block *btcutil.Block
-	peer  *serverPeer
+	peer  *peerpkg.Peer
 	reply chan struct{}
 }
 
@@ -64,33 +65,33 @@ type blockMsg struct {
 // so the block handler has access to that information.
 type invMsg struct {
 	inv  *wire.MsgInv
-	peer *serverPeer
+	peer *peerpkg.Peer
 }
 
 // headersMsg packages a bitcoin headers message and the peer it came from
 // together so the block handler has access to that information.
 type headersMsg struct {
 	headers *wire.MsgHeaders
-	peer    *serverPeer
+	peer    *peerpkg.Peer
 }
 
 // donePeerMsg signifies a newly disconnected peer to the block handler.
 type donePeerMsg struct {
-	peer *serverPeer
+	peer *peerpkg.Peer
 }
 
 // txMsg packages a bitcoin tx message and the peer it came from together
 // so the block handler has access to that information.
 type txMsg struct {
 	tx    *btcutil.Tx
-	peer  *serverPeer
+	peer  *peerpkg.Peer
 	reply chan struct{}
 }
 
 // getSyncPeerMsg is a message type to be sent across the message channel for
 // retrieving the current sync peer.
 type getSyncPeerMsg struct {
-	reply chan *serverPeer
+	reply chan int32
 }
 
 // processBlockResponse is a response sent to the reply channel of a
@@ -138,7 +139,7 @@ type headerNode struct {
 type PeerNotifier interface {
 	AnnounceNewTransactions(newTxs []*mempool.TxDesc)
 
-	UpdatePeerHeights(latestBlkHash *chainhash.Hash, latestHeight int32, updateSource *serverPeer)
+	UpdatePeerHeights(latestBlkHash *chainhash.Hash, latestHeight int32, updateSource *peerpkg.Peer)
 
 	RelayInventory(invVect *wire.InvVect, data interface{})
 
@@ -157,23 +158,35 @@ type blockManagerConfig struct {
 	MaxPeers           int
 }
 
+// peerSyncState stores additional information that the blockManager tracks
+// about a peer.
+type peerSyncState struct {
+	syncCandidate   bool
+	requestQueue    []*wire.InvVect
+	requestedTxns   map[chainhash.Hash]struct{}
+	requestedBlocks map[chainhash.Hash]struct{}
+}
+
 // blockManager provides a concurrency safe block manager for handling all
 // incoming blocks.
 type blockManager struct {
-	peerNotifier    PeerNotifier
-	started         int32
-	shutdown        int32
-	chain           *blockchain.BlockChain
-	txMemPool       *mempool.TxPool
-	chainParams     *chaincfg.Params
+	peerNotifier   PeerNotifier
+	started        int32
+	shutdown       int32
+	chain          *blockchain.BlockChain
+	txMemPool      *mempool.TxPool
+	chainParams    *chaincfg.Params
+	progressLogger *blockProgressLogger
+	msgChan        chan interface{}
+	wg             sync.WaitGroup
+	quit           chan struct{}
+
+	// These fields should only be accessed from the blockHandler thread
 	rejectedTxns    map[chainhash.Hash]struct{}
 	requestedTxns   map[chainhash.Hash]struct{}
 	requestedBlocks map[chainhash.Hash]struct{}
-	progressLogger  *blockProgressLogger
-	syncPeer        *serverPeer
-	msgChan         chan interface{}
-	wg              sync.WaitGroup
-	quit            chan struct{}
+	syncPeer        *peerpkg.Peer
+	peerStates      map[*peerpkg.Peer]*peerSyncState
 
 	// The following fields are used for headers-first mode.
 	headersFirstMode bool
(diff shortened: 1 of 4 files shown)


davecgh:
This is failing Travis current.  I believe that would probably be resolved by a rebase though.  I believe you'll need to modify the new `rpcserverPeer` interface to be `SyncPeerID` instead of `SyncPeer` and update the adapter and consumers accordingly.

davecgh:
After looking more closely, I see the issue is that `getSyncPeerMsg` is now doing `b.syncPeer.ID()` which will cause a panic when there is no `syncPeer`.  It will be nil when there is no peer.

davecgh:
Also, for an update, the first two commits look good.  I'll need to have a closer look at the next two.  If you would prefer to split out the first two into separate PRs we can get those in more quickly while figuring out the rest.

jimpo:
I fixed up the RPC adapter interface, but I'm not able to replicate the test failure on Travis locally.

davecgh:
@jimpo: The test failure only happens when you run the RPC tests.  Make sure you have btcd installed (the RPC tests use the version available in your path) and then `cd $GOPATH/src/github.com/btcsuite/btcd/integration/rpctest; go test -tags=rpctest` and you'll hit the failure.  That won't help you very much though since you want be able to see the panic, only the test failure.

There is at least one other panic when you actually try to run btcd normally though too in regards to the nil peer map.

You can more easily reproduce the panics by just starting two btcd instances such that the first one is `btcd --connect=0.0.0.0 --listen=127.0.0.1 --norpc` and the second one at the first one with `btcd --datadir=somewhere_else --logdir=somewhere_else --rpcuser=test --rpcpass=test --connect 127.0.0.1:8333` and doing `btcctl --rpcuser=test --rpcpass=test getpeerinfo`

davecgh:
To follow up, just running btcd normally produces:

```
panic(0x99c000, 0xc042004390)
        go/src/runtime/panic.go:500 +0x1af
main.(*blockManager).handleNewPeerMsg(0xc04209c140, 0xc04200e580)
        src/github.com/btcsuite/btcd/blockmanager.go:397 +0x2a8
main.(*blockManager).blockHandler(0xc04209c140)
        src/github.com/btcsuite/btcd/blockmanager.go:1166 +0x636
created by main.(*blockManager).Start
        src/github.com/btcsuite/btcd/blockmanager.go:1376 +0x12f
```

davecgh:
```
2017-08-15 21:40:21.876 [WRN] BMGR: Received done peer message for unknown peer 1
...
2017-08-15 21:40:26.878 [WRN] BMGR: No sync peer candidates available
2017-08-15 21:40:26.878 [WRN] BMGR: Received done peer message for unknown peer 2
2017-08-15 21:40:31.880 [WRN] BMGR: Received done peer message for unknown peer 3
2017-08-15 21:40:36.880 [WRN] BMGR: Received done peer message for unknown peer 4
2017-08-15 21:40:41.883 [WRN] BMGR: Received done peer message for unknown peer 5
2017-08-15 21:40:46.885 [WRN] BMGR: Received done peer message for unknown peer 6
2017-08-15 21:40:51.886 [WRN] BMGR: Received done peer message for unknown peer 7
```

This is a main part of the reason the whole server peer approach was used to begin with over keeping a separate map of peers.  It is easy for them to get out of sync and then peer lookups fail.

jimpo:
OK, I fixed the issue with the unknown peer messages with https://github.com/btcsuite/btcd/pull/989/commits/fcc4d6effaf2336153bab5f141b5786ac5b27095.

davecgh:
How's the testing going?  I'm going to take a closer look at this today.  Also, I believe the PR description needs to be updated since some of that is already merged via other PRs now.

jimpo:
I let it run for a while and spun up another instance to disconnect and reconnect a few times. Couldn't get it to panic anymore after 952f8064671fc24e2a1df8bbcedf7898a95aada2.

blockmanager.go
```diff
@@ -367,70 +381,80 @@ func (b *blockManager) isSyncCandidate(sp *serverPeer) bool {
...
-		b.startSync(peers)
+		b.startSync()
 	}
 }
 
 // handleTxMsg handles transaction messages from all peers.
 func (b *blockManager) handleTxMsg(tmsg *txMsg) {
+	peer, exists := b.peers[tmsg.peerID]
+	if !exists {
+		bmgrLog.Errorf("Received tx message from unknown peer %d", tmsg.peerID)
```
davecgh: Should be a `Warnf` since it very like means the calling code is broken.

blockmanager.go
```diff
@@ -515,18 +538,25 @@ func (b *blockManager) current() bool {
 
 // handleBlockMsg handles block messages from all peers.
 func (b *blockManager) handleBlockMsg(bmsg *blockMsg) {
+	peer, exists := b.peers[bmsg.peerID]
+	if !exists {
+		bmgrLog.Errorf("Received block message from unknown peer %d",
```
davecgh: Should be a `Warnf` since it very like means the calling code is broken.

blockmanager.go
```diff
@@ -764,13 +794,20 @@ func (b *blockManager) fetchHeaderBlocks() {
 // handleHeadersMsg handles block header messages from all peers.  Headers are
 // requested when performing a headers-first sync.
 func (b *blockManager) handleHeadersMsg(hmsg *headersMsg) {
+	peer, exists := b.peers[hmsg.peerID]
+	if !exists {
+		bmgrLog.Errorf("Received headers message from unknown peer %d",
```
davecgh: Should be a `Warnf` since it very like means the calling code is broken.

rpcadaptors.go
```diff
@@ -262,8 +262,8 @@ func (b *rpcSyncMgr) Pause() chan<- struct{} {
 //
 // This function is safe for concurrent access and is part of the
 // rpcserverSyncManager interface implementation.
-func (b *rpcSyncMgr) SyncPeer() rpcserverPeer {
-	return (*rpcPeer)(b.blockMgr.SyncPeer())
+func (b *rpcSyncMgr) SyncPeerID() int32 {
```
davecgh: s/SyncPeer/SyncPeerID for the comment.

blockmanager.go
```diff
@@ -902,6 +939,13 @@ func (b *blockManager) haveInventory(invVect *wire.InvVect) (bool, error) {
 // handleInvMsg handles inv messages from all peers.
 // We examine the inventory advertised by the remote peer and act accordingly.
 func (b *blockManager) handleInvMsg(imsg *invMsg) {
+	peer, exists := b.peers[imsg.peerID]
+	if !exists {
+		bmgrLog.Errorf("Received inv message from unknown peer %d (%d)",
```
jimpo: Oops, this is leftover from when I was debugging the unknown peer warnings. Will remove.

rpcadaptors.go
jimpo: Done

jimpo:
Thank you very much for the review!

If you are concerned about overflow, I could change it to use a `*Peer` as the map key type instead of the peer ID. If peer IDs do overflow, might there be other issues in the peering code? I haven't audited it extensively, but that should probably be made clear as an assumption.

davecgh:
I not 100% sure if the pointer as a key would work correctly with the way persistent peers are handled.  They maintain the same ID and connection state, but create new instances on reconnect.  I suppose as long as everything handles the removal and addition of the changed instance it would work.  That seems like the preferable approach to me since then it won't depend on the ID at all.

I also think changing it to a uint64 is probably a good idea too.  Even though the problem would technically still exist, the chances of a server running long enough to hit that would be extremely unlikely.

jimpo:
@davecgh I changed the map in blockManager to track a map with peer pointers as keys in https://github.com/btcsuite/btcd/pull/989/commits/06916071448d5d584c08bbe326ff637be5819be3. AFAICT, the peer ID changes if a persistent peer disconnects then gets reconnected and a new Peer struct is created, so I don't see any issues with persistent peers. I tested it locally with two nodes where one `addpeer`s the other and it seems to work fine.

If you like this approach more than the map by ID, then I'll squash and update the commit comments.

davecgh:
Thanks for the updates.  I'm going to test this out a bit, but I do like the peer pointers as keys better since even if the persistent peer IDs do end up changing, you coud still, in super rare corner cases, end up with a duplicate ID.  This should avoid any potential errors that altogether.

blockmanager.go
```diff
@@ -232,30 +245,30 @@ func (b *blockManager) findNextHeaderCheckpoint(height int32) *chaincfg.Checkpoi
...
+	var bestPeer *peerpkg.Peer
+	for peer, state := range b.peerStates {
+		if !state.syncCandidate {
 			continue
 		}
-		if segwitActive && !sp.IsWitnessEnabled() {
-			bmgrLog.Infof("peer %v not witness enabled, skipping", sp)
+
+		if segwitActive && !peer.IsWitnessEnabled() {
+			bmgrLog.Warnf("peer %v not witness enabled, skipping", peer)
```
davecgh: Looks like this got reverted in the shuffle.  Previous Comment:

This should be a Debugf. I know it was already an Infof, but I obviously missed that on the segwit review. There is no reason to notify users at an info level about this.

Since segwit isn't active this isn't being hit yet, but once it does, this could get quite noisy.

blockmanager.go
```diff
@@ -1348,9 +1396,9 @@ func (b *blockManager) Stop() error {
 	return nil
 }
 
-// SyncPeer returns the current sync peer.
-func (b *blockManager) SyncPeer() *serverPeer {
-	reply := make(chan *serverPeer)
+// SyncPeer returns the ID of the current sync peer, or 0 if there is none.
```
davecgh: SyncPeerID

PR #1007, opened 2017-08-16
Show info when JSON-RPC is not available

abitrolly:


config.go
```diff
@@ -684,6 +684,7 @@ func loadConfig() (*config, []string, error) {
 	if (cfg.RPCUser == "" || cfg.RPCPass == "") &&
 		(cfg.RPCLimitUser == "" || cfg.RPCLimitPass == "") {
 		cfg.DisableRPC = true
+		btcdLog.Infof("JSON-RPC interface is disabled")
```
davecgh: I think this would be better guarded by an of on the ` DisableFlag` instead so that it is displayed regardless of why it was disabled whether it was via this particular case, the user specifying `--norpc`, or any other future conditions which might disable it.  Also, it's not just JSON-RPC.  It's the entire RPC server which also supports WebSockets, so the log message is inaccurate.

I would suggest:

```Go
if cfg.DisableRPC {
	btcdLog.Infof("RPC server is disabled")
}
```

abitrolly:
@davecgh done. I also think about moving all RPC related config processing into the else section.

PR #1008, opened 2017-08-16
Add VSize() for a transaction

FrozenPrincess:
See: https://bitcoincore.org/en/segwit_wallet_dev/#transaction-fee-estimation

### DIFF
--- a/wire/msgtx.go
+++ b/wire/msgtx.go
@@ -828,6 +828,13 @@ func (msg *MsgTx) SerializeSizeStripped() int {
 	return msg.baseSize()
 }
 
+// The virtual size of the the transaction
+// should be used to for the purpose of fee rates and provides a smooth migration from transaction size
+func (msg *MsgTx) VSize() int {
+	return (msg.SerializeSizeStripped() * 3 + msg.SerializeSize() + 3) / 4
+}
+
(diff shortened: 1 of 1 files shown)


davecgh:
This is already available via the [mempool package as part of policy](http://godoc.org/github.com/btcsuite/btcd/mempool#GetTxVirtualSize).

FrozenPrincess:
Ah, sorry. Didn't see it. Seems like a little bit of an awkward place to put it, but I'll close this as it already exists

davecgh:
@Roasbeef can chime in as he put it there, but I think it makes sense because the virtual size is entirely a policy-related thing.  It's not something that is part of consensus nor required to serialize/deserialize at the protocol level.

FrozenPrincess:
Yeah, fair enough. It's just that I had a `*MsgTx` and need to know it's vsize, it wasn't obvious that I'd need to convert to a `*Tx` and then use the mempool.  Now that I know about it however, it's not a big deal

davecgh:
I would be in favor that function (and the one it calls in `blockchain`) over to accepting a `*wire.MsgTx` since there really isn't any reason for them to require a `*btcutil.Tx`.

PR #1010, opened 2017-08-17
blockchain: Implement new chain view.

davecgh:
This implements a new type in the `blockchain` package that takes advantage of the fact that all block nodes are now in memory to provide a flat view of a specific chain of blocks (a specific branch of the
overall block tree) from a given tip all the way back to the genesis block along with several convenience functions such as efficiently comparing two views, quickly finding the fork point (if any) between two
views, and O(1) lookup of the node at a specific height.

The view is not currently used, but the intent is that the code will be refactored to make use of these views to simplify and optimize several areas such as best chain selection and reorg logic and finding successor nodes.  They will also greatly simplify the process of disconnecting the download logic from the connection logic.

A comprehensive suite of tests is provided to ensure the chain views behave correctly.

### DIFF
--- a/blockchain/chainview.go
+++ b/blockchain/chainview.go
@@ -1,0 +1,316 @@
+// Copyright (c) 2017 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package blockchain
+
+import (
+	"sync"
+)
+
+// approxNodesPerWeek is an approximation of the number of new blocks there are
+// in a week on average.
+const approxNodesPerWeek = 6 * 24 * 7
+
+// chainView provides a flat view of a specific branch of the block chain from
+// its tip back to the genesis block and provides various convenience functions
+// for comparing chains.
+//
+// For example, assume a block chain with a side chain as depicted below:
+//   genesis -> 1 -> 2 -> 3 -> 4  -> 5 ->  6  -> 7  -> 8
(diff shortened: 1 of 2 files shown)


blockchain/chainview.go
```diff
@@ -0,0 +1,316 @@
...
+// tips is efficient.
+//
+// This function is safe for concurrent access.
+func (c *chainView) SetTip(node *blockNode) {
+	c.mtx.Lock()
+	c.setTip(node)
+	c.mtx.Unlock()
+}
+
+// height returns the height of hte tip of the chain view.  It will return -1 if
```
jimpo: typo: the

blockchain/chainview_test.go
```diff
@@ -0,0 +1,423 @@
...
+				t.Errorf("%s: unexpected view tip -- got %v, "+
+					"want %v", test.name, test.view.Tip(),
+					tip)
+				continue
+			}
+		}
+	}
+}
+
+// TestChainViewNil ensurs that creating and accessing a nil chain view behaves
```
jimpo: typo: ensures

blockchain/chainview_test.go
```diff
@@ -0,0 +1,423 @@
...
+// as expected.
+func TestChainViewNil(t *testing.T) {
+	// Ensure to unininitialized view are considered equal.
```
jimpo: typo: two uninitialized views?

PR #1011, opened 2017-08-17
blockmanager: Remove dependency on cfg in main package.

jimpo:
Continue working towards https://github.com/btcsuite/btcd/issues/978.

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -7,8 +7,6 @@ package main
 import (
 	"container/list"
 	"net"
-	"os"
-	"path/filepath"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -153,6 +151,10 @@ type blockManagerConfig struct {
 	PeerNotifier PeerNotifier
 	Chain        *blockchain.BlockChain
 	TxMemPool    *mempool.TxPool
+	ChainParams  *chaincfg.Params
+
+	DisableCheckpoints bool
+	MaxPeers           int
 }
 
 // blockManager provides a concurrency safe block manager for handling all
@@ -163,6 +165,7 @@ type blockManager struct {
 	shutdown        int32
 	chain           *blockchain.BlockChain
 	txMemPool       *mempool.TxPool
+	chainParams     *chaincfg.Params
 	rejectedTxns    map[chainhash.Hash]struct{}
 	requestedTxns   map[chainhash.Hash]struct{}
 	requestedBlocks map[chainhash.Hash]struct{}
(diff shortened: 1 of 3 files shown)


blockmanager.go
```diff
@@ -309,10 +307,7 @@ func (b *blockManager) startSync(peers *list.List) {
 		// and fully validate them.  Finally, regression test mode does
 		// not support the headers-first approach so do normal block
 		// downloads when in regression test mode.
-		if b.nextCheckpoint != nil &&
-			best.Height < b.nextCheckpoint.Height &&
-			!cfg.RegressionTest && !cfg.DisableCheckpoints {
```
jimpo: If either `cfg.RegressionTest` or `cfg.DisableCheckpoints`, then `b.nextCheckpoint` is guaranteed to be `nil`, so this check is redundant.

blockmanager.go
```diff
@@ -452,9 +447,8 @@ func (b *blockManager) handleTxMsg(tmsg *txMsg) {
 
 	// Process the transaction to include validation, insertion in the
 	// memory pool, orphan handling, etc.
-	allowOrphans := cfg.MaxOrphanTxs > 0
 	acceptedTxs, err := b.txMemPool.ProcessTransaction(tmsg.tx,
-		allowOrphans, true, mempool.Tag(tmsg.peer.ID()))
```
jimpo: This changes the behavior slightly. Previously, if `MaxOrphanTxs` was configured to be 0 and the node received an orphan transaction, it would respond with a reject (duplicate) packet. Now it will just not add the tx to the orphan pool and silently carry on. This seems more like the right behavior.

There are ways to modify the logic to achieve parity with Core (eg. request parent txs), but that can be added later.

blockmanager.go
davecgh: I agree about `cfg.DisableCheckpoints`, but I believe there can in fact be checkpoints for `cfg.RegressionTest` because you can manually specify them via `--addcheckpoint`.  Granted it wouldn't make sense to do that since the whole point of the regression test network is generating a new chain, so it's probably wouldn't be that big of a deal, but it would mean you can't use the regression network to detect intentionally bad checkpoints.

blockmanager.go
davecgh: It's not a big deal, but I'm not sure I agree here.  The intention of a reject message is to inform the remote peer that you rejected the transaction.  It's true that nothing is really done about that by an peers on the network right now, but it would be possible to notice that your txns are consistently being rejected by a peer and decide to disconnect and go find greener pastures.

blockmanager.go
jimpo: I'll add back the RegressionNet check.

blockmanager.go
jimpo: I did this for two reasons. 1) This is not the same behavior as Bitcoin Core and 2) it seems odd that we don't send a reject if MaxOrphanTxs is configured to 1 and we already have 1 orphan tx, whereas we do send a reject if MaxOrphanTxs is configured to 0. Essentially, we are saying "this may legitimately be an orphan tx, but we don't want to spend the spend storing it", whereas a reject(duplicate) indicates that we don't think the peer should have broadcast it to us (I think?).

davecgh: That sounds very reasonable.  I have no objections to the change then.

davecgh:
I should add that I also ran this PR overnight along with #989 with debug logging on and didn't encounter any issues nor see anything out of the ordinary in the debug log.

PR #1012, opened 2017-08-17
blockchain: Optimize block locator generation.

davecgh:
This significantly optimizes and simplifies the generation of block locators by making use of the fact that all block nodes are now in memory and therefore it is no longer necessary to consult the database
for the hashes or worry about issues related to dynamic loading of nodes.

Also, it slightly modifies the algorithm so that the doubling doesn't start for one additional iteration in order to mirror other prominent clients on the network.  Due to the way block locators are used, this
does not change any semantics in terms of requesting and locating blocks.

Finally, the semantics of `BlockLocatorFromHash` have been changed to return a locator for the current tip in the case the hash is unknown. This is far preferable since only including the passed block hash, when
it isn't known, could end up leading to causing a redownload of the entire chain under certain circumstances.

### DIFF
--- a/blockchain/blocklocator.go
+++ b/blockchain/blocklocator.go
@@ -6,17 +6,34 @@ package blockchain
 
 import (
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
-	"github.com/btcsuite/btcd/database"
-	"github.com/btcsuite/btcd/wire"
 )
+
+// log2FloorMasks defines the masks to use when quickly calculating
+// floor(log2(x)) in a constant log2(32) = 5 steps, where x is a uint32, using
+// shifts.  They are derived from (2^(2^x) - 1) * (2^(2^x)), for x in 4..0.
+var log2FloorMasks = []uint32{0xffff0000, 0xff00, 0xf0, 0xc, 0x2}
+
+// fastLog2Floor calculates and returns floor(log2(x)) in a constant 5 steps.
+func fastLog2Floor(n uint32) uint8 {
+	rv := uint8(0)
+	exponent := uint8(16)
+	for i := 0; i < 5; i++ {
+		if n&log2FloorMasks[i] != 0 {
+			rv += exponent
+			n >>= exponent
+		}
+		exponent >>= 1
+	}
+	return rv
+}
 
 // BlockLocator is used to help locate a specific block.  The algorithm for
 // building the block locator is to add the hashes in reverse order until
 // the genesis block is reached.  In order to keep the list of locator hashes
-// to a reasonable number of entries, first the most recent previous 10 block
-// hashes are added, then the step is doubled each loop iteration to
-// exponentially decrease the number of hashes as a function of the distance
-// from the block being located.
+// to a reasonable number of entries, first the most recent 12 block hashes are
+// added, then the step is doubled each loop iteration to exponentially decrease
+// the number of hashes as a function of the distance from the block being
+// located.
 //
 // For example, assume you have a block chain with a side chain as depicted
 // below:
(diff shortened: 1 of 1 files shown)


blockchain/blocklocator.go
```diff
@@ -6,139 +6,94 @@ package blockchain
...
+		// final node is the genesis block.
+		height := node.height - step
+		if height < 0 {
+			height = 0
 		}
 
-		return nil
-	})
+		// Walk backwards through the nodes to the correct ancestor.
+		node = node.Ancestor(height)
```
jimpo: Can you check if `bi.Contains(node)`, in which case you can look up the ancestor from the index? Otherwise, `bi` is not needed as a receiver?

davecgh:
@jimpo: Do you mean elaborate in the code?  The commit message and PR description have:

Finally, the semantics of BlockLocatorFromHash have been changed to return a locator for the current tip in the case the hash is unknown. This is far preferable since only including the passed block hash, when
it isn't known, could end up leading to causing a redownload of the entire chain under certain circumstances.

blockchain/blocklocator.go
davecgh: This suggestion gets implemented on the chain view in the next PR #1013 which does take advantage of the O(1) lookups they provide.  I could certainly do an index lookup here too, but I figured since #1014 is already done (switch over to using the chain view) and #1013 has the optimization, I'd just let it go.

Great observation!

blockchain/blocklocator.go
```diff
@@ -6,139 +6,94 @@ package blockchain
...
-				if n.inMainChain {
-					forkHeight = n.height
-					break
-				}
-			}
-		}
+		// Requested hash itself + previous 10 entries + genesis block.
+		// Then floor(log2(height-10)) entries for the skip portion.
+		adjustedHeight := uint32(node.height) - 10
+		maxEntries = 12 + fastLog2Floor(adjustedHeight)
```
jimpo: I've calculated `maxEntries` as 11 + floor(log_2(node.height - 9)), for height > 9. For example, if calculating the locator for node with height 12, the locator is 12, 11, ..., 2, 0, which has 12 elements, whereas `12 + floor(log(12 - 10))` would calculate it as 13.

Say node.height is `n` and you fill the first 10 elements of the block locator. The 10th element has height `n' = n - 9`. The remaining nodes in the locator are at heights `n' - 1`, `n' - 1 - 2`, `n' - 1 - 2 - 4`, `n' - 1 - 2 - 4 - 8`, ... or `n' + 1 - 2^(i+1)`. The number of such elements in this sequence of remaining nodes (after the first 10) is `ceil(log_2(n' + 1))` or `1 + floor(log_2(n'))` for `n' > 1`. Thus the total locator length `11 + floor(log_2(n - 9))`.

blockchain/blocklocator.go
jimpo: Ah, right. I got blockIndex confused with chainView.

davecgh:
Thanks for checking the math.  I'm pretty sure it's right as is because the calc only runs for `height >= 13`, so it doesn't apply to anything less than that.  In terms of `heights >= 13`, here are the expected locators:

```
height 13: 13 12 11 10 9 8 7 6 5 4 3 2 genesis == len 13
height 14: 14 13 12 11 10 9 8 7 6 5 4 3 1 genesis == len 14
height 15: 15 14 13 12 11 10 9 8 7 6 5 4 2 genesis == len 14
height 16: 16 15 14 13 12 11 10 9 8 7 6 5 3 genesis == len 14
height 17: 17 16 15 14 13 12 11 10 9 8 7 6 4 genesis == len 14
height 18: 18 17 16 15 14 13 12 11 10 9 8 7 5 1 genesis == len 15
height 19: 19 18 17 16 15 14 13 12 11 10 9 8 6 2 genesis == len 15
height 25: 25 24 23 22 21 20 19 18 17 16 15 14 12 8 genesis == len 15
height 26: 26 25 24 23 22 21 20 19 18 17 16 15 13 9 1 genesis == len 16
height 27: 27 26 25 24 23 22 21 20 19 18 17 16 14 10 2 genesis == len 16
height 32: 32 31 30 29 28 27 26 25 24 23 22 21 19 15 7 genesis == len 16
height 41: 41 40 39 38 37 36 35 34 33 32 31 30 28 24 16 genesis == len 16
height 42: 42 41 40 39 38 37 36 35 34 33 32 31 29 25 17 1 genesis == len 17
height 44: 44 43 42 41 40 39 38 37 36 35 34 33 31 27 19 3 genesis == len 17
height 73: 73 72 71 70 69 68 67 66 65 64 63 62 60 56 48 32 genesis == len 17
height 74: 74 73 72 71 70 69 68 67 66 65 64 63 61 57 49 33 1 genesis == len 18
height 40000: 400000 399999 399998 399997 399996 399995 399994 399993 399992 399991 399990 399989 399987 399983 399975 399959 399927 399863 399735 399479 398967 397943 395895 391799 383607 367223 334455 268919 137847 genesis == len 30
```

So, checking those same values with the existing formula gives:
```Go
	for _, height := range []uint32{13, 14, 15, 16, 17, 18, 19, 25, 26, 27, 32, 41, 42, 44, 73, 74, 400000} {
		fmt.Printf("height %d: floor(log2(%d-10) = %d\n", height,
			height, 12+fastLog2Floor(height-10))
	}
```

```
height 13: 12+floor(log2(13-10) = 13
height 14: 12+floor(log2(14-10) = 14
height 15: 12+floor(log2(15-10) = 14
height 16: 12+floor(log2(16-10) = 14
height 17: 12+floor(log2(17-10) = 14
height 18: 12+floor(log2(18-10) = 15
height 19: 12+floor(log2(19-10) = 15
height 25: 12+floor(log2(25-10) = 15
height 26: 12+floor(log2(26-10) = 16
height 27: 12+floor(log2(27-10) = 16
height 32: 12+floor(log2(32-10) = 16
height 41: 12+floor(log2(41-10) = 16
height 42: 12+floor(log2(42-10) = 17
height 44: 12+floor(log2(44-10) = 17
height 73: 12+floor(log2(73-10) = 17
height 74: 12+floor(log2(74-10) = 18
height 400000: 12+floor(log2(400000-10) = 30
```
If we were instead to use `11+ floor(log_2(node.height - 9))`, the results would be:
```
height 13: 11+floor(log2(13-9) = 13
height 14: 11+floor(log2(14-9) = 13
height 15: 11+floor(log2(15-9) = 13
height 16: 11+floor(log2(16-9) = 13
height 17: 11+floor(log2(17-9) = 14
height 18: 11+floor(log2(18-9) = 14
height 19: 11+floor(log2(19-9) = 14
height 25: 11+floor(log2(25-9) = 15
height 26: 11+floor(log2(26-9) = 15
height 27: 11+floor(log2(27-9) = 15
height 32: 11+floor(log2(32-9) = 15
height 41: 11+floor(log2(41-9) = 16
height 42: 11+floor(log2(42-9) = 16
height 44: 11+floor(log2(44-9) = 16
height 73: 11+floor(log2(73-9) = 17
height 74: 11+floor(log2(74-9) = 17
height 400000: 11+floor(log2(400000-9) = 29
```

Notice that is not correct for heights 14, 15, 16, 18, 19, 26, 27, etc

jimpo:
Yep, I was off by one. :-/

Not really looking for a code comment. I was just curious under what circumstances we would try to generate a locator for a block that's not in the active chain. I'm guessing [here](https://github.com/btcsuite/btcd/blob/master/blockmanager.go#L1029)?

davecgh:
Yes, that is the primary reason for the change to use tip on invalid hashes.  It really shouldn't happen under normal circumstances with the current code since it already checked if the inventory is known (aka we already have the block) prior to reaching that code in question.  However, it is technically possible there could be a reorg in between those two checks and in that cause it would end up redownloading the entire chain.  That would obviously be an extremely rare and hard to hit corner case, but I still wanted to protect against it.

Also, not related to the current code, but I think it's smart to safeguard against a potential incorrect usage that very well could crop up when moving over to parallel block downloads.  Due to semantics of how they are used to locate blocks, if a block locator only consists of a single hash and the remote peer doesn't know about that hash (perhaps it's on a side chain and the peer was offline when the block was originally broadcast), the remote peer will just give you back inventory starting from the genesis block which would trigger redownloading the entire chain.

blockchain/blocklocator.go
davecgh: For reference, I did change this over to be a standalone function since it doesn't need the block index receiver.

PR #1014, opened 2017-08-18
blockchain: Refactor to use new chain view.

davecgh:
**This PR requires #1012 and #1013**.

- Remove `inMainChain` from block nodes since that can now be efficiently determined by using the chain view
- Track the best chain via a chain view instead of a single block node
  - Use the tip of the best chain view everywhere `bestNode` was used
  - Update chain view tip instead of updating best node
- Change reorg logic to use more efficient chain view fork finding logic
- Change block locator code over to use more efficient chain view logic
  - Remove now unused block-index-based block locator code
  - Move `BlockLocator` definition to `chain.go`
  - Move `BlockLocatorFromHash` and `LatestBlockLocator` to `chain.go`
    - Update both to use more efficient chain view logic
- Rework `IsCheckpointCandidate` to use block index and chain view
- Optimize `MainChainHasBlock` to use chain view instead of hitting db
      - Move to `chain.go` since it no longer involves database I/O
      - Removed error return since it can no longer fail
- Optimize `BlockHeightByHash` to use chain view instead of hitting db
      - Move to `chain.go` since it no longer involves database I/O
      - Removed error return since it can no longer fail
- Optimize `BlockHashByHeight` to use chain view instead of hitting db
      - Move to `chain.go` since it no longer involves database I/O
      - Removed error return since it can no longer fail
- Optimize `HeightRange` to use chain view instead of hitting db
      - Move to `chain.go` since it no longer involves database I/O
- Optimize `BlockByHeight` to use chain view for main chain check
- Optimize `BlockByHash` to use chain view for main chain check


blockchain/chain.go
```diff
@@ -1309,6 +1412,7 @@ func New(config *Config) (*BlockChain, error) {
 		blocksPerRetarget:   int32(targetTimespan / targetTimePerBlock),
 		index:               newBlockIndex(config.DB, params),
 		hashCache:           config.HashCache,
+		bestChain:           newChainView(nil),
```
jimpo: Is it maybe worth having a chainView constructor that accepts a capacity since that can be looked up from the DB at startup?

blockchain/chainio.go
```diff
@@ -1164,24 +1164,24 @@ func (b *BlockChain) initChainState() error {
...
-			b.bestNode = node
+			b.bestChain.SetTip(node)
+			tip = node
 		}
 
-		// Ensure the resulting best node matches the stored best state
+		// Ensure the resulting best chain matches the stored best state
 		// hash.
-		if b.bestNode.hash != state.hash {
+		if b.bestChain.Tip().hash != state.hash {
```
jimpo: nit: Use `tip` variable?

blockchain/chainview.go
```diff
@@ -12,6 +12,25 @@ import (
 // in a week on average.
 const approxNodesPerWeek = 6 * 24 * 7
 
+// log2FloorMasks defines the masks to use when quickly calculating
+// floor(log2(x)) in a constant log2(32) = 5 steps, where x is a uint32, using
+// shifts.  They are derived from (2^(2^x) - 1) * (2^(2^x)), for x in 4..0.
+var log2FloorMasks = []uint32{0xffff0000, 0xff00, 0xf0, 0xc, 0x2}
+
+// fastLog2Floor calculates and returns floor(log2(x)) in a constant 5 steps.
+func fastLog2Floor(n uint32) uint8 {
```
jimpo: Is there like a math util package/file this could go in instead?

blockchain/chain.go
davecgh: I'm fairly certain that is what happens already.  Please double check me, but this will just result in a chain view with a nil backing array (no allocations).  Then, when the initialization happens, it calls `.SetTip` which will notice that the cap of the nil slice is less than the needed amount and perform the single allocation then (along with and additional `approxNodesPerWeek`) for future expansion before a realloc is needed.

davecgh: After looking more closely, I see that `initChainState` is calling `.SetTip` on every loop iteration.  I will move this out of the loop and after the assertion that the tip is correct (also addressing your other comment about using the local) and that will ensure only a single allocation of the right size.  Agree?

blockchain/chainio.go
davecgh: Good suggestion.  I was specifically trying to ensure the `bestChain` was set to tip here, but in conjunction with your other comment, I think it does indeed make more sense to check the tip local here and then `b.bestChain.SetTip(tip)` after it.

blockchain/chainview.go
davecgh: There isn't one.  This really is only used in this one place though (it looks like an addition because I moved it out of `blocklocator.go`), so I don't think creating a separate package for just this one use is worth it here.

blockchain/chain.go
davecgh: Changed per latest comment.

davecgh:
Rebased

PR #1019, opened 2017-08-20
blockchain: Refactor inv discovery for chain view.

davecgh:
**This relies on several outstand PRs.  #1017, #1018, #1012, #1013, #1014**.

This refactors the code that locates blocks (inventory discovery) out of server and into `blockchain` where it can make use of the new much more efficient chain view and more easily be tested.  As an aside, it really belongs in `blockchain` anyways since it's purely dealing with the block index and best chain.

Since the majority of the network has moved to header-based semantics, this also provides an additional optimization to allow headers to be located directly versus needing to first discover the hashes and then fetch the headers.

The new functions are named `LocateBlocks` and `LocateHeaders`.  The former returns a slice of located hashes and the latter returns a slice of located headers.

Finally, it also updates the RPC server `getheaders` call and related plumbing to use the new `LocateHeaders` function.

A comprehensive suite of tests is provided to ensure both functions behave correctly for both correct and incorrect block locators.

### DIFF
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -1290,6 +1290,159 @@ func (b *BlockChain) HeightRange(startHeight, endHeight int32) ([]chainhash.Hash
 	return hashes, nil
 }
 
+// locateInventory returns the node of the block after the first known block in
+// the locator along with the number of subsequent nodes needed to either reach
+// the provided stop hash or the provided max number of entries.
+//
+// In addition, there are two special cases:
+//
+// - When no locators are provided, the stop hash is treated as a request for
+//   that block, so it will either return the node associated with the stop hash
+//   if it is known, or nil if it is unknown
+// - When locators are provided, but none of them are known, nodes starting
+//   after the genesis block will be returned
+//
+// This is primarily a helper function for the locateBlocks and locateHeaders
+// functions.
+//
+// This function MUST be called with the chain state lock held (for reads).
+func (b *BlockChain) locateInventory(locator BlockLocator, hashStop *chainhash.Hash, maxEntries uint32) (*blockNode, uint32) {
+	// There are no block locators so a specific block is being requested
+	// as identified by the stop hash.
+	stopNode := b.index.LookupNode(hashStop)
+	if len(locator) == 0 {
+		if stopNode == nil {
+			// No blocks with the stop hash were found so there is
+			// nothing to do.
+			return nil, 0
+		}
+		return stopNode, 1
+	}
+
+	// Find the most recent locator block hash in the main chain.  In the
+	// case none of the hashes in the locator are in the main chain, fall
+	// back to the genesis block.
+	startNode := b.bestChain.Genesis()
+	for _, hash := range locator {
+		node := b.index.LookupNode(hash)
+		if node != nil && b.bestChain.Contains(node) {
+			startNode = node
+			break
+		}
+	}
+
+	// Start at the block after the most recently known block.  When there
+	// is no next block it means the most recently known block is the tip of
+	// the best chain, so there is nothing more to do.
+	startNode = b.bestChain.Next(startNode)
+	if startNode == nil {
+		return nil, 0
+	}
+
+	// Calculate how many entries are needed.
+	total := uint32((b.bestChain.Tip().height - startNode.height) + 1)
+	if stopNode != nil && b.bestChain.Contains(stopNode) &&
+		stopNode.height >= startNode.height {
+
+		total = uint32((stopNode.height - startNode.height) + 1)
+	}
+	if total > maxEntries {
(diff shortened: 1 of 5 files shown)


blockchain/chain.go
```diff
@@ -1290,6 +1290,176 @@ func (b *BlockChain) HeightRange(startHeight, endHeight int32) ([]chainhash.Hash
...
+		if node != nil && b.bestChain.Contains(node) {
+			knownNode = node
+			break
+		}
+	}
+
+	// Start at the block after the most most recently known block.  When
+	// there is no next block it means the most recently known block is
+	// the tip of the best chain, so there is nothing more to do.
+	knownNode = b.bestChain.Next(knownNode)
```
jimpo: nit: `knownNode` seems like an inappropriate variable name since it's the first block not known by the peer. `startNode` or something seems better.

blockchain/chain.go
```diff
@@ -1290,6 +1290,176 @@ func (b *BlockChain) HeightRange(startHeight, endHeight int32) ([]chainhash.Hash
...
+}
+
+// locateHeaders returns the headers of the blocks after the first known block
+// in the locator until the provided stop hash is reached, or up to the provided
+// max number of block headers.
+//
+// See the comment on the exported function for more details on special cases.
+//
+// This function MUST be called with the chain state lock held (for reads).
+func (b *BlockChain) locateHeaders(locator BlockLocator, hashStop *chainhash.Hash, maxHeaders uint32) []wire.BlockHeader {
```
jimpo: There's a chance to refactor `locateHeaders` and `locateBlocks` logic into a shared function `locateNodes` which returns a `[]*blockNode`, then `locateHeaders` can just map the blockNodes to headers and `locateBlocks` would map nodes to hashes. That would reduce duplication at the expense of an additional temporary slice allocation.

blockchain/chain.go
davecgh: Great suggestion.  I was thinking of it from the point of view of the local peer, but I I agree your name is much clearer.  Thanks!  Changed.

blockchain/chain.go
davecgh: I'm a bit torn on this one.  That is originally how I had it, but I decided to change it over as an optimization to avoid the temporary slice allocation.  As the commit description says `Since the majority of the network has moved to header-based semantics, this also provides an additional optimization to allow headers to be located directly versus needing to first discover the hashes and the fetch the headers.`

I can certainly change it if there are any serious reservations about the duplication, but my goal was really to optimize for the most common case, and given that realistically almost every peer on the network now uses headers-first semantics, it is the most common case.

blockchain/chain.go
```diff
@@ -1187,6 +1177,289 @@ func (b *BlockChain) FetchHeader(hash *chainhash.Hash) (wire.BlockHeader, error)
...
+	}
+
+	// Populate and return the found hashes.
+	hashes := make([]chainhash.Hash, 0, total)
+	for node := startNode; node != nil; node = b.bestChain.Next(node) {
+		hashes = append(hashes, node.hash)
+
+		// Stop once either the number of hashes exceeds the maximum or
+		// the stop hash has been reached.
+		if uint32(len(hashes)) >= maxHashes || node == stopNode {
```
jimpo: You can just check `uint32(len(hashes)) >= total` since total was calculated using the height of `stopNode`.

blockchain/chain.go
jimpo: That makes sense. Another way to remove the duplication would be to have a helper function that just returns `startNode` and `total`, then `LocateHeaders` could be implemented as

```go
func (b *BlockChain) locateHeaders(locator BlockLocator, hashStop *chainhash.Hash, maxHeaders uint32) (*blockNode, uint32) {
    node, total := locateHelper(locator, hashStop, maxHeaders)
    headers := make([]wire.BlockHeader, 0, total)
    for i := 0; i < total; i++ {
        headers[i] = node.Header()
        node = b.bestChain.Next(node)
    }
    return headers
}
```

My concern with the duplication (as with all duplication) is that someone changes one function and not the other. But since there's only two, it's not a big deal, so I'll leave it up to you.

davecgh: I like that idea.  I'll introduce a `locateInventory` that does that and `locateBlocks` and `locateHeaders` to use it.  It will need to be slightly different in that we want it to return nil when there are no entries to return instead of a slice with zero entries to avoid the allocation (and also make sure the tests pass).

davecgh: Done

blockchain/chain.go
davecgh: Done as part of the refactor to `locateInventory`.

davecgh:
Rebased and squashed.

PR #1022, opened 2017-08-21
blockchain: Correct chainview locator comment.

davecgh:
Pointed out by @ysangkok.

### DIFF
--- a/blockchain/chainview.go
+++ b/blockchain/chainview.go
@@ -390,10 +390,10 @@ func (c *chainView) blockLocator(node *blockNode) BlockLocator {
 
 // BlockLocator returns a block locator for the passed block node.  The passed
 // node can be nil in which case the block locator for the current tip
-// associated with the view will be returned.  This only differs from the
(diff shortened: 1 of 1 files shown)


blockchain/chainview.go
```diff
@@ -390,8 +390,7 @@ func (c *chainView) blockLocator(node *blockNode) BlockLocator {
 
 // BlockLocator returns a block locator for the passed block node.  The passed
 // node can be nil in which case the block locator for the current tip
-// associated with the view will be returned.  This only differs from the
-// exported version in that it is up to the caller to ensure the lock is held.
+// associated with the view will be returned.
 //
 // See BlockLocator for details on the algorithm used to create a block locator.
```
jrick: this should be referring to the unexported function.

jrick: actually my bad, it is referring to the type not the function.  perhaps that can be cleared up too but this is still ok.

davecgh: I modified the comment to clarify the BlockLocator type.

PR #1027, opened 2017-08-24
[netsync] Move blockmanager to new netsync package.

jimpo:
The blockManager implements the blockchain syncing protocol in addition to adding/removing transactions from the mempool in response to peer notifications. This is a lot of important logic that in contained in the main package without unit tests.

Moving this into its own package makes it easier to add unit tests and the additional encapsulation aids in the development of new features.

Possible improvements are:
- Adding unit tests
- Headers first syncing (past checkpoints) + multi-peer downloading
- Accepting BIP 130 direct header announcements
- BIP 152 support

Regarding the package name `netsync`:

- `blockmgr` is inappropriate because it also keeps the mempool in sync
- `sync` conflicts with the standard library of the same name
- "net" alludes to the fact that this is similar to some of the net_processing code in Bitcoin Core

Closes https://github.com/btcsuite/btcd/issues/978.


netsync/manager.go
```diff
@@ -1207,55 +1181,55 @@ out:
...
-	bmgrLog.Trace("Block handler done")
+	sm.wg.Done()
+	log.Trace("Block handler done")
 }
 
 // handleBlockchainNotification handles notifications from blockchain.  It does
 // things such as request orphan block parents and relay accepted blocks to
 // connected peers.
-func (b *blockManager) handleBlockchainNotification(notification *blockchain.Notification) {
+func (sm *SyncManager) handleBlockchainNotification(notification *blockchain.Notification) {
```
davecgh: This is definitely not something that needs to be addressed in this PR, but I wonder if this wouldn't make more sense to have in the mempool package now with it having a way to query if the sync manager believes it is current.  I believe that is really the only reason it requires the sync manager.  Everything else is related to updating the mempool and relay.

davecgh:
I'm testing with this applied as well.  I'll update with results.

netsync/manager.go
jimpo: Yeah, I'll keep that in mind as I refactor this package.

PR #1029, opened 2017-08-25
connmgr: Preventing duplicate call in the concurrent context, which could cause program crash

lonnng:
Close a closed channel will cause panic in the program, it should be eliminated.

### DIFF
--- a/connmgr/connmanager.go
+++ b/connmgr/connmanager.go
@@ -398,7 +398,14 @@ func (cm *ConnManager) Stop() {
 		_ = listener.Close()
 	}
 
-	close(cm.quit)
+	// Preventing duplicate call in the concurrent context, which could cause program crash.
+	select {
+	case <-cm.quit:
+		// expect
+	default:
+		close(cm.quit)
+	}
+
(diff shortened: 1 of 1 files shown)


davecgh:
Thanks for the PR, but there is no need do this because the `Stop` function is protected by the atomic add to `cm.stop` above it where it exits the function on any subsequent calls.  So, even if you call it concurrently, it's not possible to hit a second channel close.

Here is some sample code to illustrate:

```Go
package main

import (
	"fmt"
	"sync"
	"sync/atomic"
)

type connmgr struct {
	stop int32
	quit chan struct{}
}

func (cm *connmgr) Stop() {
	if atomic.AddInt32(&cm.stop, 1) != 1 {
		return
	}

	fmt.Println("Stopping")
	close(cm.quit)

}

func main() {
	cm := &connmgr{quit: make(chan struct{})}
	var wg sync.WaitGroup
	for i := 0; i < 1000; i++ {
		wg.Add(1)
		go func() {
			cm.Stop()
			wg.Done()
		}()
	}
	wg.Wait()
}
```

Output:

```
Stopping
```

PR #1030, opened 2017-08-25
travis: Update to go 1.8 and 1.9.

davecgh:
Also, make the `gosimple` linter happy while here, and update `README.md` to call out the Go 1.8 minimum version requirement.


davecgh:
Travis has now been updated to support building on Go 1.9.

PR #1034, opened 2017-08-28
blockchain: Load all block headers into the block index on init.

jimpo:
Currently only the blocks in the active chain are loaded into the block index on initialization. This instead iterates over the entire block index bucket in LevelDB and loads all nodes.

In order to do so efficiently, we switch the key format for the blocks in LevelDB to be prefixed with the block height encoded as big-endian, so when iterating over keys, we always see blocks in order of height.

This adds a database migration at startup to sync the old block index bucket with the new experimental one.

### DIFF
--- a/blockchain/chainio.go
+++ b/blockchain/chainio.go
@@ -1058,6 +1058,7 @@ func dbPutBestState(dbTx database.Tx, snapshot *BestState, workSum *big.Int) err
 func (b *BlockChain) createChainState() error {
 	// Create a new node from the genesis block and set it as the best node.
 	genesisBlock := btcutil.NewBlock(b.chainParams.GenesisBlock)
+	genesisBlock.SetHeight(0)
 	header := &genesisBlock.MsgBlock().Header
 	node := newBlockNode(header, 0)
 	b.bestChain.SetTip(node)
@@ -1131,6 +1132,8 @@ func (b *BlockChain) initChainState() error {
 	// Attempt to load the chain state from the database.
 	var isStateInitialized bool
 	err := b.db.View(func(dbTx database.Tx) error {
+		start := time.Now()
+
 		// Fetch the stored chain state from the database metadata.
 		// When it doesn't exist, it means the database hasn't been
 		// initialized for use with chain yet, so break out now to allow
@@ -1151,40 +1154,61 @@ func (b *BlockChain) initChainState() error {
 		// for them versus a whole bunch of little ones to reduce
 		// pressure on the GC.
 		log.Infof("Loading block index.  This might take a while...")
-		bestHeight := int32(state.height)
-		blockNodes := make([]blockNode, bestHeight+1)
-		var tip *blockNode
-		for height := int32(0); height <= bestHeight; height++ {
-			header, err := dbFetchHeaderByHeight(dbTx, height)
+
+		blockCount, err := dbTx.GetBlockCount()
+		if err != nil {
+			return err
+		}
+
+		blockNodes := make([]blockNode, blockCount)
+		var lastNode *blockNode
+		var i int32
+		dbTx.ForEachBlockHeader(func(headerBytes []byte) error {
+			var header wire.BlockHeader
+			err := header.Deserialize(bytes.NewReader(headerBytes))
 			if err != nil {
 				return err
 			}
 
 			// Initialize the block node for the block, connect it,
 			// and add it to the block index.
-			node := &blockNodes[height]
-			initBlockNode(node, header, height)
-			if tip != nil {
-				node.parent = tip
-				node.workSum = node.workSum.Add(tip.workSum,
-					node.workSum)
+			node := &blockNodes[i]
+			initBlockNode(node, &header, 0)
+			if lastNode == nil {
+				if header.BlockHash() != *b.chainParams.GenesisHash {
+					return AssertError(fmt.Sprintf("initChainState: Expected "+
+						"first entry in block index to be genesis block, "+
+						"found %s", header.BlockHash()))
+				}
+			} else {
+				// Since we iterate block headers in order of height, if the
+				// blocks are mostly linear there is a very good chance the
+				// previous header processed is the parent.
+				parent := lastNode
+				if header.PrevBlock != parent.hash {
+					parent = b.index.LookupNode(&header.PrevBlock)
+				}
+				if parent == nil {
+					return AssertError(fmt.Sprintf("initChainState: Could "+
+						"not find parent for block %s", header.BlockHash()))
+				}
+
+				node.parent = parent
+				node.height = parent.height + 1
+				node.workSum = node.workSum.Add(parent.workSum, node.workSum)
 			}
+
 			b.index.AddNode(node)
+			lastNode = node
+			i++
 
-			// This node is now the end of the best chain.
-			tip = node
-		}
+			return nil
+		})
 
-		// Ensure the resulting best chain matches the stored best state
-		// hash and set the best chain view accordingly.
-		if tip == nil || tip.hash != state.hash {
-			var tipHash chainhash.Hash
-			if tip != nil {
-				tipHash = tip.hash
-			}
-			return AssertError(fmt.Sprintf("initChainState: block "+
-				"index chain tip %s does not match stored "+
-				"best state %s", tipHash, state.hash))
+		tip := b.index.LookupNode(&state.hash)
+		if tip == nil {
+			return AssertError(fmt.Sprintf("initChainState: cannot find "+
+				"chain tip %s in block index", state.hash))
 		}
 		b.bestChain.SetTip(tip)
 
@@ -1206,6 +1230,9 @@ func (b *BlockChain) initChainState() error {
 		b.stateSnapshot = newBestState(tip, blockSize, blockWeight,
 			numTxns, state.totalTxns, tip.CalcPastMedianTime())
 		isStateInitialized = true
+
+		elapsed := time.Since(start)
+		log.Debugf("initChainState took %d seconds", int64(elapsed/time.Second))
 
 		return nil
 	})
--- a/database/ffldb/db.go
+++ b/database/ffldb/db.go
@@ -74,6 +74,10 @@ var (
 	// writeLocKeyName is the key used to store the current write file
 	// location.
 	writeLocKeyName = []byte("ffldb-writeloc")
+
+	// blockIdxBucketName is the bucket used internally to track block
+	// metadata.
+	blockIdx2BucketName = []byte("ffldb-blockidx-v2")
 )
 
 // Common error strings.
@@ -945,21 +949,23 @@ func (b *bucket) Delete(key []byte) error {
 // pendingBlock houses a block that will be written to disk when the database
 // transaction is committed.
 type pendingBlock struct {
-	hash  *chainhash.Hash
-	bytes []byte
+	hash   *chainhash.Hash
+	bytes  []byte
+	height uint32
 }
 
 // transaction represents a database transaction.  It can either be read-only or
 // read-write and implements the database.Bucket interface.  The transaction
 // provides a root bucket against which all read and writes occur.
 type transaction struct {
-	managed        bool             // Is the transaction managed?
-	closed         bool             // Is the transaction closed?
-	writable       bool             // Is the transaction writable?
-	db             *db              // DB instance the tx was created from.
-	snapshot       *dbCacheSnapshot // Underlying snapshot for txns.
-	metaBucket     *bucket          // The root metadata bucket.
-	blockIdxBucket *bucket          // The block index bucket.
+	managed         bool             // Is the transaction managed?
+	closed          bool             // Is the transaction closed?
+	writable        bool             // Is the transaction writable?
+	db              *db              // DB instance the tx was created from.
+	snapshot        *dbCacheSnapshot // Underlying snapshot for txns.
+	metaBucket      *bucket          // The root metadata bucket.
+	blockIdxBucket  *bucket          // The v1 block index bucket.
+	blockIdx2Bucket *bucket          // The v2 block index bucket.
 
 	// Blocks that need to be stored on commit.  The pendingBlocks map is
 	// kept to allow quick lookups of pending data by block hash.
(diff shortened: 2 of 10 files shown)


database/ffldb/db.go
```diff
@@ -1172,6 +1187,12 @@ func (tx *transaction) StoreBlock(block *btcutil.Block) error {
 		return makeDbErr(database.ErrDriverSpecific, str, err)
 	}
 
+	if block.Height() == btcutil.BlockHeightUnknown {
```
davecgh: I believe this is going to be a problem since the block is inserted before its contextual position is known.  See [here](https://github.com/btcsuite/btcd/blob/3cb87afa2fc078efc2b12d16458cce15daab311c/blockchain/accept.go#L45-L59).

I understand that once everything is fully headers first, versus the partial headers-first approach that is used now, this probably won't be an issue since the block's position will be known by virtue of its header already having been processed.

database/ffldb/db.go
```diff
@@ -2082,3 +2147,21 @@ func openDB(dbPath string, network wire.BitcoinNet, create bool) (database.DB, e
 	// well as database initialization, if needed.
 	return reconcileDB(pdb, create)
 }
+
+func serializeBlockIndexKey(blockHash *chainhash.Hash, blockHeight int32) ([]byte, error) {
+	var blockIdxKey bytes.Buffer
+	blockIdxKey.Grow(4 + 32)
+
+	err := binary.Write(&blockIdxKey, binary.BigEndian,
```
davecgh: I know this is just PoC right now, but I would really prefer this section create an appropriately sized slice and avoid `bytes.Bufffer` and, in particular, `binary.Write`, since it uses reflection and is extremely slow, plus it needs an error.  Consequently, using `binary.BigEndian.PutUin32`, as done elsewhere in the package is highly preferred.

Something like:
```Go
func serializeBlockIndexKey(blockHash *chainhash.Hash, blockHeight int32) []byte {
	// The serialized block index key format is:
	//   <blockhash><blockheight>
	indexKey := make([]byte, len(blockHash)+4)
	copy(indexKey, blockHash[:])
	binary.BigEndian.PutUint32(indexKey[len(blockHash):], uint32(blockHeight))
	return indexKey
}
```

database/ffldb/migrate.go
```diff
@@ -0,0 +1,170 @@
+// Copyright (c) 2015-2016 The btcsuite developers
```
davecgh: 2017

blockchain/chainio.go
```diff
@@ -1058,6 +1058,7 @@ func dbPutBestState(dbTx database.Tx, snapshot *BestState, workSum *big.Int) err
 func (b *BlockChain) createChainState() error {
 	// Create a new node from the genesis block and set it as the best node.
 	genesisBlock := btcutil.NewBlock(b.chainParams.GenesisBlock)
+	genesisBlock.SetHeight(0)
```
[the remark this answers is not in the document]
jimpo: `NewBlock` initializes the height to `BlockHeightUnknown`, which is -1.

database/ffldb/db.go
jimpo: Correct me if I'm wrong, but from what I can tell `maybeAcceptBlock` is never called with an orphan block. The call to `maybeAcceptBlock` is skipped [here](https://github.com/btcsuite/btcd/blob/3cb87afa2fc078efc2b12d16458cce15daab311c/blockchain/process.go#L225) if the previous block is not known. So by the time the block is stored, the height is definitely known I believe.

blockchain/chainio.go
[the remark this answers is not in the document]
davecgh: Ah, nice catch.  It didn't really matter before since the genesis block isn't really ever used from the db since it's available via the chain params.

database/ffldb/db.go
davecgh: You're right.  Sorry, I've been looking at like 4 different code bases that are all really similar, so I got this one confused with another.  I should've looked before commenting.  It does indeed set the block height about six lines above the code I linked.  Doh!

davecgh:
I need to give this some more review, but here are the results on a fully synced chain on a 7200 RPM HDD for reference:

```
...
2017-08-30 18:48:28.559 [INF] BTCD: Block database loaded
18:48:28.593 [INF] BCDB: Re-indexing block information in the database. This might take a while...
2017-08-30 18:48:35.155 [INF] BCDB: Block database upgrade complete
2017-08-30 18:48:35.228 [INF] CHAN: Loading block index.  This might take a while...
2017-08-30 18:48:38.029 [INF] CHAN: initChainState took 2 seconds
2017-08-30 18:48:42.607 [INF] CHAN: Chain state (height 482709, hash 000000000000000000e14d588e7cb6ef0ad082bfe565dc2e3a7e84df6c944691, totaltx 250677677, work 156563225709273107132928742)
...
```

Future starts remained at 2 seconds (and didn't migrate again of course).

database/ffldb/db.go
```diff
@@ -980,6 +986,19 @@ type transaction struct {
 // Enforce transaction implements the database.Tx interface.
 var _ database.Tx = (*transaction)(nil)
 
+// blockIdx2Bucket returns the new block index bucket.
+func (tx *transaction) blockIdx2Bucket() (*bucket, error) {
```
davecgh: I would suggest leaving this caching out since it slightly increases the complexity, and unless you have future plans to make more use of it, it seems like the reality is that it is only ever used to get the initial block count, to foreach the headers, and when writing new blocks to the database.  The first two only happen once, and the last one can easily lookup the bucket once outside of the write loop.

It is also per transaction, so even if more uses are added in the future, it won't be cached across transactions and not really be used then either.

database/ffldb/db.go
```diff
@@ -1708,11 +1786,25 @@ func (tx *transaction) writePendingAndCommit() error {
...
 		blockHdr := blockData.bytes[0:blockHdrSize]
 		blockRow := serializeBlockRow(location, blockHdr)
+
 		err = tx.blockIdxBucket.Put(blockData.hash[:], blockRow)
 		if err != nil {
 			rollback()
 			return err
 		}
+
+		blockIdx2Bucket, err := tx.blockIdx2Bucket()
```
davecgh: I would suggest moving this outside of the loop.  Once you've looked it up, you can reuse it inside the loop.

database/ffldb/db.go
```diff
@@ -1331,6 +1357,58 @@ func (tx *transaction) FetchBlockHeaders(hashes []chainhash.Hash) ([][]byte, err
 	return headers, nil
 }
 
+// GetBlockCount counts all blocks stored in the database. This function has to
+// iterate a cursor over all blocks and thus is relatively slow.
+//
```
davecgh: The comment on both the interface method and here should call out that it will return `ErrTxClosed`.

jimpo:
Yes, my intention was to fully deprecate the current `blockIndexBucket`. If you want to be able to look up blocks by hash from the DB, we can just have a bucket storing a hash -> height index. In fact, that already exists in `chainio` as the `hashidx` bucket. We could just move it over to be managed by `ffldb` and backfill during the migration to ensure it has a height for all blocks.

davecgh:
I was thinking about the migration a bit last night and think it probably should be done in the database itself as opposed to asking the caller to call a method.  The reason I suggest this is that the block index is specifically managed by the database itself.  Migrations of the utxoset and spend journal, which an upcoming PR of mine modify, have to be done in `blockchain` since they are managed by that package and hence couldn't call `Migrate` either.  What do you think?

jimpo:
@davecgh I moved the migration into `openDB` and removed from the public interface.

database/interface.go
```diff
@@ -389,6 +389,38 @@ type Tx interface {
...
 
+	// GetBlockCount counts all blocks stored in the database.
+	//
+	// Depending on the backend implementation, this function may have to
+	// iterate a cursor over all blocks and thus be relatively slow.
+	//
+	// The interface contract guarantees at least the following errors will
+	// be returned (other implementation-specific errors are possible):
+	//   - ErrTxClosed if the transaction has already been closed
+	GetBlockCount() (uint, error)
```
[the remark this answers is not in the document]
jimpo: I sort of think uint32 is more appropriate. We encode block heights as uint32 or int32 everywhere and with checkpoints, the number of blocks should never really come close to 4 billion. Also, if it did we'd have big problems because we use this number to allocate a slice of block nodes.

blockchain/chainio.go
```diff
@@ -1151,41 +1154,105 @@ func (b *BlockChain) initChainState() error {
...
+				"chain tip %s in block index", state.hash))
 		}
 
-		// Ensure the resulting best chain matches the stored best state
-		// hash and set the best chain view accordingly.
-		if tip == nil || tip.hash != state.hash {
-			var tipHash chainhash.Hash
-			if tip != nil {
-				tipHash = tip.hash
+		/*
```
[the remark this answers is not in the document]
jimpo: Yeah, I left it commented to compare performance during review, but I'll remove now.

database/interface.go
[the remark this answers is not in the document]
davecgh: Agreed.  I was thinking about it in terms of the fact that this is all blocks as opposed to only those in the main chain so there will always be more than whatever the current height is, but after thinking about it, `uint32` will provide more than enough space (2^31) for those extra blocks.

database/ffldb/db.go
```diff
@@ -2080,5 +2246,42 @@ func openDB(dbPath string, network wire.BitcoinNet, create bool) (database.DB, e
...
 	// well as database initialization, if needed.
-	return reconcileDB(pdb, create)
+	if _, err := reconcileDB(pdb, create); err != nil {
+		return nil, err
+	}
+
+	if !create {
+		// Run all required data migrations
+		if err := migrateBlockIndex(pdb); err != nil {
+			return nil, err
```
davecgh: While I was going back and forth between this branch and another, I added some code to delete the migration key and new buckets, but I noticed some strange behavior where they weren't actually being stored to the database when there was an error, despite the underlying transactions having completed successfully.

After thinking about it for a bit, I realized the problem is that the database needs to be closed at any point after the `pdb` is created above when there are any errors before returning or it won't close the underlying cache and flush to disk.  This is true for the `reconcileDB` error path above, this one, and the lookup path below.  The fact it wasn't doing it before for the `reconcileDB` error return was also a bug.

```Go
		// Run all required data migrations.
		if err := migrateBlockIndex(pdb); err != nil {
			// Ignore error on close since no further recovery could
			// be done anyways.
			_ = pdb.Close()
			return nil, err
		}
```

database/ffldb/db.go
```diff
@@ -2080,5 +2246,42 @@ func openDB(dbPath string, network wire.BitcoinNet, create bool) (database.DB, e
...
+		}
+	}
+
+	// Lookup bucket IDs for internal buckets
+	err = pdb.View(func(dbTxUncast database.Tx) error {
```
davecgh: Since there are tests and this certainly works, this isn't a big deal, but I wonder if this would be better using the standard bucket machinery to get a bucket and then copy the ID so this doesn't really need to understand the underlying details?

Something like the following:

```Go
	// Lookup bucket IDs for internal buckets
	err = pdb.View(func(dbTx database.Tx) error {
		meta := dbTx.Metadata()
		b := meta.Bucket(blockIdx2BucketName)
		if b == nil {
			message := "Database is missing internal bucket: blockIdx2Bucket"
			return makeDbErr(database.ErrBucketNotFound, message, nil)
		}
		pdb.blockIdx2BucketID = b.(*bucket).id

		b = meta.Bucket(blockHashBucketName)
		if b == nil {
			message := "Database is missing internal bucket: blockHashBucket"
			return makeDbErr(database.ErrBucketNotFound, message, nil)
		}
		pdb.blockHashBucketID = b.(*bucket).id

		return nil
	})
	if err != nil {
		// Ignore error on close since no further recovery could be done
		// anyways.
		_ = pdb.Close()
		return nil, err
	}
```

database/ffldb/db.go
```diff
@@ -1331,6 +1350,99 @@ func (tx *transaction) FetchBlockHeaders(hashes []chainhash.Hash) ([][]byte, err
...
+//   - ErrTxClosed if the transaction has already been closed
+//
+// NOTE: The slices returned by this function are only valid during a
+// transaction.  Attempting to access them after a transaction has ended
+// results in undefined behavior.  Additionally, the slices must NOT
+// be modified by the caller.  These constraints prevent additional data
+// copies and allows support for memory-mapped database implementations.
+//
+// This function is part of the database.Tx interface implementation.
+func (tx *transaction) ForEachBlockHeader(fn func(headerBytes []byte) error) error {
```
davecgh: Everything here looks sound, but seeing this complexity makes me wonder if the code (in a future PR) should be updated to store the partial block rows at the time `StoreBlock` is called along with the header, some dummy values for the on-disk location, and a flag to indicate if the block data is available, and then have `writePendingAndCommit` update the record accordingly once the block is written to disk and the location is known.

Since it would be in the underlying treaps at that point, a simple `.ForEach` could be used in here.  I think it's also going to have to do something similar to that anyways since I can foresee the need for a `StoreHeader` with full blown headers-first sync.

jimpo:
Thinking about this even more, what if we stop storing headers in ffldb and instead put them in a bucket managed by the `blockchain` package? So ffldb just stores the raw block data (including the header of course, but not in LevelDB) indexed by block hash. Then `blockchain` would have a bucket where we store all block headers indexed by (height + hash), along with a bitfield of the validation state and possibly any other chain-contextual things we might need in the future. That bucket would essentially serve as the backing store for the block index. The only thing that would break AFAICT is the `loadheaders` db command, which just times how long it takes to load headers from ffldb, so we could just remove that.

I think this model would work better with headers-first sync and (looking ahead) probably pruning as well.

davecgh:
We probably should open up an issue to further discuss this, but yes, I have been thinking similarly.  However, I was trying not to get too far ahead since, as this is, it's already a huge improvement.

Previously it wasn't possible to get headers only out of memory, but now that all block nodes are in memory and they can be reconstructed from the nodes, we never have to touch disk anymore to get headers.  Yay!

One of the challenges I've run into when working with implementing a utxo cache is that some parts of blockchain still believe the database always represents the best chain.  My recent `chainview` work helped solve a lot of these issues (and even was a primary motivator for it), but the block index is an area where this is still problematic, in large part because it's managed by the database and not blockchain.

PR #1037, opened 2017-08-29
server: Split some newServer logic into new methods.

jimpo:
Just moves the listener setup logic out of `newServer` and cleans it up a bit.

### DIFF
--- a/server.go
+++ b/server.go
@@ -93,6 +93,28 @@ func (oa *onionAddr) Network() string {
 // Ensure onionAddr implements the net.Addr interface.
 var _ net.Addr = (*onionAddr)(nil)
 
+// onionAddr implements the net.Addr interface with two struct fields
+type simpleAddr struct {
+	net, addr string
+}
+
+// String returns the address.
+//
+// This is part of the net.Addr interface.
+func (a simpleAddr) String() string {
+	return a.addr
+}
+
+// Network returns the network.
+//
+// This is part of the net.Addr interface.
+func (a simpleAddr) Network() string {
+	return a.net
+}
+
+// Ensure simpleAddr implements the net.Addr interface.
+var _ net.Addr = simpleAddr{}
+
 // broadcastMsg provides the ability to house a bitcoin message to be broadcast
 // to all connected peers except specified excluded peers.
 type broadcastMsg struct {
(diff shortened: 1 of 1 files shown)


davecgh:
If I recall correctly, the parsing code was done that way because `net.Listen` with plain `tcp` doesn't work as you might expect.  It will only listen on a single interface, IPv4, or IPv6, and the behavior is not predictable nor the same between Operating Systems.

For example, launching btcd without this PR results in the p2p server listening on both the IPv4 and IPv6 interfaces, as expected.

```
...
2017-08-29 23:34:56.887 [INF] CMGR: Server listening on 0.0.0.0:8333
2017-08-29 23:34:56.887 [INF] CMGR: Server listening on [::]:8333
...
```

However, when launching it with this PR results in only listening on the IPv6 interface:

```
2017-08-29 23:35:58.208 [INF] RPCS: RPC server listening on 127.0.0.1:8334
2017-08-29 23:35:58.208 [INF] RPCS: RPC server listening on [::1]:8334
...
2017-08-29 23:35:58.708 [INF] CMGR: Server listening on [::]:8333
```

For more details on the various expected listen behavior, take a look at the docs [here](https://github.com/btcsuite/btcd/blob/master/docs/configure_rpc_server_listen_interfaces.md).  Now, if you try those things with this PR on various Operating Systems, you'll notice they don't work the same.

davecgh:
With the above said, aside from the issues around removing `parseListeners`, I'm definitely in favor of the concept of splitting the logic out of `newServer` into a separate function, preferably one that could even potentially have tests added for it.

jimpo:
OK, thanks. I wasn't 100% sure if that was a safe change, but I figured I'd throw it up for review.

Reverted the change to remove `parseListeners`, but I changed it to return a `net.Addr` with the `tcp4` or `tcp6` net included instead of returning two slices. This reduces duplication in the calling code.

server.go
```diff
@@ -1973,28 +1995,23 @@ func (s *server) ScheduleShutdown(duration time.Duration) {
...
-// listeners on the correct interface "tcp4" and "tcp6".  It also properly
-// detects addresses which apply to "all interfaces" and adds the address to
-// both slices.
-func parseListeners(addrs []string) ([]string, []string, bool, error) {
-	ipv4ListenAddrs := make([]string, 0, len(addrs)*2)
-	ipv6ListenAddrs := make([]string, 0, len(addrs)*2)
-	haveWildcard := false
-
+// parseListeners determines whether each listen address is IPv4 and IPv6 and
+// returns a listen of appropriate net.Addrs to listen on with TCP. It also
```
davecgh: s/listener/list/

server.go
```diff
@@ -93,6 +93,28 @@ func (oa *onionAddr) Network() string {
...
 
+// onionAddr implements the net.Addr interface with two struct fields
+type simpleAddr struct {
+	net, addr string
+}
+
+// String returns the address.
+//
+// This is part of the net.Addr interface.
+func (a *simpleAddr) String() string {
```
davecgh: If you change this to be a value receiver, it will allow it work with anything that uses a `fmt.Stringer` interface (e.g. `fmt.Printf`, `log.Warnf`, etc) so don't need to modify the logging statements.

As an aside, `fmt.Stringers` should pretty much always be value receives because Go will automatically deference pointers, so it works with both.

Example:
```Go
package main

import(
	"net"
	"fmt"
)

type foo struct {
	addr string
}

func (f foo) String() string {
	return f.addr
}

func (f foo) Network() string {
	return "moo"
}


func main(){
	addrs := []net.Addr{&foo{"1"}, foo{"2"}}
	for _, addr := range addrs {
		fmt.Println(addr)
	}
}
```
Output:
```
1
2
```

server.go
```diff
@@ -2529,6 +2424,88 @@ func newServer(listenAddrs []string, db database.DB, chainParams *chaincfg.Param
...
+	netAddrs, err := parseListeners(listenAddrs)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	listeners := make([]net.Listener, 0, len(listenAddrs))
+	for _, addr := range netAddrs {
+		listener, err := net.Listen(addr.Network(), addr.String())
+		if err != nil {
+			srvrLog.Warnf("Can't listen on %s: %v", addr.String(), err)
```
davecgh: Don't need the `.String()` here since it's a `fmt.Stringer`.

server.go
```diff
@@ -2529,6 +2424,88 @@ func newServer(listenAddrs []string, db database.DB, chainParams *chaincfg.Param
 	return &s, nil
 }
 
+// initListeners initializes the configured net listeners and adds any bound
+// addresses to the address manager. Returns the listeners and a NAT interface,
+// which is non-nil if UPnP is in use.
+func initListeners(amgr *addrmgr.AddrManager, listenAddrs []string, services wire.ServiceFlag) ([]net.Listener, NAT, error) {
+	var nat NAT
+	discover := true
```
davecgh: I know it was already this way, but it seems to me like this would be cleaner as:

```Go
discover := len(cfg.ExternalIPs) == 0
if !discover {
	defaultPort, err := ...
	...
} else {
	...
}
```

server.go
```diff
@@ -2529,6 +2424,88 @@ func newServer(listenAddrs []string, db database.DB, chainParams *chaincfg.Param
...
+		var err error
+		nat, err = Discover()
+		if err != nil {
+			srvrLog.Warnf("Can't discover upnp: %v", err)
+		}
+		// nil nat here is fine, just means no upnp on network.
+	}
+
+	// Listen for TCP connections at the configured addresses
+	netAddrs, err := parseListeners(listenAddrs)
```
davecgh: I think this (and the section below to actually listen) should be at the top of the function.  The most likely failure point is going to be when either parsing the user-provided listen addresses, or when attempting to listen due to the port already being used, etc.

Also, the code above this point does not result in returning any failures.  Therefore, I don't see any benefit to doing all the work above only to end up more likely failing out here after the fact.

PR #1044, opened 2017-09-01
netsync: Tests for SyncManager.

jimpo:
This is far from full coverage, but it provides some basic coverage and a starting place to add more checks. It doesn't make sense to fully test the current implementation, but I hope to flesh out this test suite as I work on the headers-first sync.

I think it would be worthwhile to pull a bunch of the code from `common_test.go` (both here and from other packages) into a package that can be shared by the tests of other packages. For example, some of the peer setup with the pipe connections was stolen from peer_test.go.


davecgh:
I think they're worthwhile since I suspect the RPC integration tests, aside from the fact they're not very comprehensive currently, would not really be able to do a good job of exercising all of the possible corner cases.  I'm also thinking this could come in very handy when moving to parallel sync.

You could certainly create multiple nodes and link them up in various ways to make sure they end up all in agreement (and that is definitely something that would be nice to have), but I think it would be really difficult for it to test a lot of the negative conditions such as misbehaving peers that intentionally don't answer after advertising they have inventory, purposely sending blocks at insanely slow speeds (like 1 byte per second), etc.

Having unit-style tests with mock peers will allow all sorts of chicanery to be thrown at the sync code which I believe will really help improve its robustness in the face of bad actors.

netsync/manager_test.go
```diff
@@ -0,0 +1,776 @@
...
+
+	select {
+	case call := <-ctx.peerNotifier.relayInventoryChan:
+		if call.invVect.Type != wire.InvTypeBlock ||
+			call.invVect.Hash != *block.Hash() {
+			t.Fatalf("PeerNotifier received unexpected RelayInventory call: "+
+				"%v", call.invVect)
+		}
+	default:
+		t.Fatalf("Expected SyncManager to make " +
```
davecgh: `t.Fatal`.  Also, looks like the wrapping is off here.

netsync/manager_test.go
```diff
@@ -0,0 +1,776 @@
...
+	instance.invChan = make(chan *wire.MsgInv)
+	instance.headersChan = make(chan *wire.MsgHeaders)
+	instance.getDataChan = make(chan *wire.MsgGetData)
+	instance.getBlocksChan = make(chan *wire.MsgGetBlocks)
+	instance.getHeadersChan = make(chan *wire.MsgGetHeaders)
+	instance.rejectChan = make(chan *wire.MsgReject)
+	return &instance
+}
+
+func buildBlockInv(block *wire.MsgBlock) *wire.MsgInv {
```
davecgh: I think this would be more useful as `blocks ...*btcutil.Block`, so you could use it for the multi-block calls too.

Something like:

```Go
func buildBlockInv(blocks ...*btcutil.Block) *wire.MsgInv {
	msg := wire.NewMsgInv()
	for _, block := range blocks {
		invVect := wire.NewInvVect(wire.InvTypeBlock, block.Hash())
		msg.AddInvVect(invVect)
	}
	return msg
}
```

Then above:

```Go
...
	// Remote node replies to getblocks with an inv.
	syncMgr.QueueInv(buildBlockInv(blocks...), localNode)
...
	// SyncManager should send a getdata message requesting blocks.
	syncMgr.QueueInv(buildBlockInv(block), localNode)
...
	// Send invalid block with timestamp in the far future.
	...
	syncMgr.QueueInv(buildBlockInv(block), localNode)
```

netsync/manager_test.go
```diff
@@ -0,0 +1,776 @@
...
+	}
+
+	localPeerCfg := peer.Config{
+		UserAgentName:    "btcdtest",
+		UserAgentVersion: "1.0",
+		ChainParams:      &chainParams,
+		Services:         wire.SFNodeNetwork,
+	}
+
+	_, localNode, err :=
```
davecgh: Same as above regarding func on same line.

netsync/manager_test.go
```diff
@@ -0,0 +1,776 @@
...
+	select {
+	case <-syncChan:
+	case <-time.After(time.Second):
+		t.Fatal("Timeout waiting for sync manager to process transaction 3")
+	}
+
+	// AnnounceNewTransactions should be called with no transactions
+	select {
+	case call := <-ctx.peerNotifier.announceNewTransactionsChan:
+		if len(call.newTxs) != 0 {
```
davecgh: This feels like a (rather benign) bug to me (not in your tests, rather in the code and your tests have found it which is awesome).  `AnnounceNewTransactions` is intended to announce transactions that are actually added to the mempool.  Orphan transactions are _not_ added to the mempool until/if their parent(s) show(s) up.  Also, there is little use to receiving and empty slice since the caller can't even tell what tx was added as an orphan if that were the intent (which it isn't as previously noted).

Since the code that is called in `server` and `rpcserver` both just do a for loop over the slice, which is empty, nothing actually happens, so the bug doesn't actually have any negative behavior.  Nevertheless, I feel like `handleTxMsg` should be doing the following to avoid sending the notification when no transactions were actually added:

```Go
if len(acceptedTxs) > 0 {
	sm.peerNotifier.AnnounceNewTransactions(acceptedTxs)
}
```

Naturally that means this test would need to be updated.

netsync/manager_test.go
```diff
@@ -0,0 +1,776 @@
...
+
+	// Attempt to process orphan transaction
+	syncMgr.QueueTx(tx4, localNode, syncChan)
+	select {
+	case <-syncChan:
+	case <-time.After(time.Second):
+		t.Fatal("Timeout waiting for sync manager to process transaction 4")
+	}
+
+	// AnnounceNewTransactions should not be called for invalid transaction
```
davecgh: Given the discussion above, if that is changed as suggested, this would be `for an invalid non-orphan transaction` to indicate that it is indeed testing the desired property.

netsync/manager_test.go
```diff
@@ -0,0 +1,776 @@
...
+
+	// An already rejected transaction should not get a reject response
+	syncMgr.QueueTx(tx4, localNode, syncChan)
+	select {
+	case <-syncChan:
+	case <-time.After(time.Second):
+		t.Fatal("Timeout waiting for sync manager to process transaction 4")
+	}
+
+	// Expect node to send reject in response to invalid transaction
```
davecgh: This comment doesn't seem accurate.  The one above does.

peer/peer.go
```diff
@@ -82,11 +82,6 @@ var (
 	// sentNonces houses the unique nonces that are generated when pushing
 	// version messages that are used to detect self connections.
 	sentNonces = newMruNonceMap(50)
-
-	// allowSelfConns is only used to allow the tests to bypass the self
-	// connection detecting and disconnect logic since they intentionally
-	// do so for testing purposes.
-	allowSelfConns bool
```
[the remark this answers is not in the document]
jimpo: With `TstAllowSelfConns` defined in `export_test.go` in the `peer` package, it can't be called from tests in the `netsync` package. If I run `go test ./netsync/...`, I get `undefined: peer.TstAllowSelfConns`.

If you really don't like having it in the peer config, I can move `TstAllowSelfConns()` into the `peer` package proper (from export_test). However, this pattern still feels kind of wrong to me. First, the global doesn't get unset currently, so if it is running multiple tests the change of this global setting could leak into other tests I believe. What might be better is on a per-test basis to call:

```go
func TestSomething(t *testing.T) {
  peer.TstAllowSelfConns()
  defer peer.TstDisallowSelfConns()

  ...
}
```

jimpo:
Handled all comments without an explicit reply.

jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Enhancement
- Outdated (very)

PR #1047, opened 2017-09-04
blockchain/indexers: Allow interrupts.

davecgh:
This propagates the interrupt channel through to `blockchain` and the indexers so that it is possible to interrupt long-running operations such as catching up indexes.


jimpo:
LGTM

PR #1050, opened 2017-09-06
blockchain: Track block validation status in block index.

jimpo:
Each node in the block index records some flags about its validation state. This is just stored in memory for now, but can save effort if attempting to reconnect a block that failed validation or was
disconnected. In a subsequent change, we will store this in the block index bucket created by https://github.com/btcsuite/btcd/pull/1049.

This field provides little value now, is required for headers-first syncing.

### DIFF
--- a/blockchain/blockindex.go
+++ b/blockchain/blockindex.go
@@ -16,6 +16,29 @@ import (
 	"github.com/btcsuite/btcd/wire"
 )
 
+// blockStatus is a bit field representing the validation state of the block.
+type blockStatus byte
+
+const (
+	// statusDataStored indicates that the block's payload is stored on disk.
+	statusDataStored blockStatus = 1 << iota
+
+	// statusValid indicates that the block has been fully validated.
+	statusValid
+
+	// statusValidateFailed indicates that the block has failed validation.
+	statusValidateFailed
+
+	// statusInvalidAncestor indicates that one of the block's ancestors has
+	// has failed validation, thus the block is also invalid.
+	statusInvalidAncestor
+
+	// statusNone indicates that the block has no validation state flags set.
+	//
+	// NOTE: This must be defined last in order to avoid influencing iota.
+	statusNone blockStatus = 0
+)
+
 // blockNode represents a block within the block chain and is primarily used to
 // aid in selecting the best chain to be the main chain.  The main chain is
 // stored into the block database.
@@ -49,6 +72,9 @@ type blockNode struct {
 	nonce      uint32
 	timestamp  int64
 	merkleRoot chainhash.Hash
+
+	// status is a bitfield representing the validation state of the block
+	status blockStatus
 }
 
 // initBlockNode initializes a block node from the given header and height.  The
@@ -165,6 +191,20 @@ func (node *blockNode) CalcPastMedianTime() time.Time {
 	// even number, this code will be wrong.
 	medianTimestamp := timestamps[numNodes/2]
 	return time.Unix(medianTimestamp, 0)
+}
+
+// KnownValid returns whether the block is known to be valid. This will return
+// false for a valid block that has not been fully validated yet.
+func (node *blockNode) KnownValid() bool {
+	return node.status&statusValid != 0
+}
+
+// KnownInvalid returns whether the block is known to be invalid. This may be
+// because the block itself failed validation or any of its ancestors is
+// invalid. This will return false for invalid blocks that have not been proven
+// invalid yet.
+func (node *blockNode) KnownInvalid() bool {
+	return node.status&(statusValidateFailed|statusInvalidAncestor) != 0
 }
 
 // blockIndex provides facilities for keeping track of an in-memory index of the
--- a/blockchain/blockindex.go
+++ b/blockchain/blockindex.go
@@ -39,6 +39,27 @@ const (
 	statusNone blockStatus = 0
 )
 
+// HaveData returns whether the full block data is stored in the database. This
+// will return false for a block node where only the header is downloaded or
+// kept.
+func (status blockStatus) HaveData() bool {
+	return status&statusDataStored != 0
+}
+
+// KnownValid returns whether the block is known to be valid. This will return
+// false for a valid block that has not been fully validated yet.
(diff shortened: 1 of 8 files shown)


blockchain/blockindex.go
```diff
@@ -49,6 +67,9 @@ type blockNode struct {
 	nonce      uint32
 	timestamp  int64
 	merkleRoot chainhash.Hash
+
+	// status is a bitfield representing the validation state of the block
+	status blockStatus
```
jimpo: With padding, this additional byte takes up a whole word. If we switch timestamp to be a uint32 (which is reasonable given that that is how it is serialized in the block message), we could avoid using the additional 8 bytes. However, that requires adding a bunch of casting in various parts of the code where timestamp is converted to/from a `time.Time`.

blockchain/blockindex.go
```diff
@@ -16,6 +16,24 @@ import (
 	"github.com/btcsuite/btcd/wire"
 )
 
+const (
+	// statusDataStored indicates that the block's payload is stored on disk.
+	statusDataStored = 1 << iota
```
davecgh: The first one should specify the type so that `iota` will carry it through.  Otherwise these will all be untyped constants.

```Go
statusDataStored blockStatus = 1 << iota
statusValid
...
```

Also, please define the type before the `const` block (which will use it) for consistency.

blockchain/blockindex.go
davecgh: I personally would prefer to just eat the extra 4 bytes.  I realize that currently it does get serialized to the wire as a uint32 (and is indeed part of the consensus rules because the hash is committing to that uint32 serialized value).  However, working with timestamps that aren't `int64` is not very natural in my opinion.  There are places in the code that do it, but it is much more natural to work with `int64` timestamps since that is really sort of the standard when dealing with unix-style timestamps.  It is also more natural to be able to subtract two timestamps and result in a negative time instead of worrying about overflowing.

blockchain/chain.go
```diff
@@ -503,16 +503,37 @@ func (b *BlockChain) getReorganizeNodes(node *blockNode) (*list.List, *list.List
...
 	}
 
+	// If any of the node's ancestors are invalid, unwind attachNodes, marking
+	// each one as invalid for future reference.
+	if invalidChain {
+		var next *list.Element
+		for e := attachNodes.Front(); e != nil; e = next {
+			next = e.Next()
+			n := attachNodes.Remove(e).(*blockNode)
+			n.status |= statusInvalidAncestor
```
davecgh: What do you think about having a `func (node *blockNode) SetStatus(status blockStatus)` that accepts the status flag(s)?  My reasoning is that when this is persisted, chances are there will need to be some type of map to track the modified nodes for a later database flush.  This will be particularly important with a utxocache since the states have to be kept in sync in memory.  Having a setter means it would only need to marked modified once in the setter versus having to remember to update the modified map after setting it all over the place.

blockchain/error.go
```diff
@@ -212,6 +208,13 @@ const (
 	// included in the block's coinbase transaction doesn't match the
 	// manually computed witness commitment.
 	ErrWitnessCommitmentMismatch
+
+	// ErrPreviousBlockUnknown indicates that the previous block is not known.
+	ErrPreviousBlockUnknown
+
+	// ErrPreviousBlockInvalid indicates that an ancestor of this block has
+	// already failed validation.
+	ErrPreviousBlockInvalid
```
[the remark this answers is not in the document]
jimpo: The intention is that a block with an invalid ancestor is invalid itself, so the previous block is also invalid. But if you think `ErrInvalidAncestor` is more clear, I'll change it.

blockchain/chain.go
jimpo: As you can see in https://github.com/btcsuite/btcd/pull/1066, I decided to to explicitly call `MarkDirty` where required. This was because there were times (in initChainState) when we would want to add nodes without marking the blocks dirty. I guess that could be handled differently though, by clearing the dirty set or somehow having it inactive until `initChainState` completes.

Would `SetStatus` directly replace the status or do the bitwise or? Most of the time we edit the status we want to do the bitwise or, so I think `SetStatusFlags` and `UnsetStatusFlags` would be more convenient.

jimpo:
I realized that block status access is not concurrent-safe for reads since the status could be changed by another goroutine. Do we need a `GetStatus(node *blockNode) blockStatus` method on blockIndex?

blockchain/chain.go
```diff
@@ -500,7 +500,9 @@ func (b *BlockChain) getReorganizeNodes(node *blockNode) (*list.List, *list.List
 	// Nothing to detach or attach if there is no node.
 	attachNodes := list.New()
 	detachNodes := list.New()
-	if node == nil {
+
```
davecgh: Extraneous newline.  The comment applies to the entire block.

blockchain/blockindex.go
```diff
@@ -225,13 +272,33 @@ func (bi *blockIndex) AddNode(node *blockNode) {
 	bi.Unlock()
 }
 
-// RemoveNode removes the provided node from the block index.  There is no check
-// whether another node in the index depends on this one, so it is up to caller
-// to avoid that situation.
+// GetStatus provides concurrent-safe access to the status field of a node.
+//
+// This function is safe for concurrent access.
+func (bi *blockIndex) GetStatus(node *blockNode) blockStatus {
```
davecgh: Please call this `NodeStatus` instead of `GetStatus`.  Naming functions `GetX` is highly discouraged by the [Effective Go Guidelines](https://golang.org/doc/effective_go.html#Getters) which this project follows.

blockchain/blockindex.go
```diff
@@ -49,6 +93,9 @@ type blockNode struct {
 	nonce      uint32
 	timestamp  int64
 	merkleRoot chainhash.Hash
+
+	// status is a bitfield representing the validation state of the block
```
davecgh: Should we call out this this is protected by the block index mutex?  The rest of the fields are effectively immutable since they aren't changed after the initial creation and thus don't require a mutex.

However, this field is unique in that it is mutable and changed depending on the state.

PR #1052, opened 2017-09-10
btcjson: Implement addwitnessaddress functionality.

sturd:
These commits implement objects and functions to build addwitnessaddress JSON-RPC commands within the btcjson package.  I have added a test to prove out the marshalling of the object.


sturd:
Thank you, and apologies for the commit mess.  I _had_ tried to squash the two original commits down to save you the bother.  I'll make sure to keep future commits within guidelines, too.

PR #1053, opened 2017-09-12
rpcclient: implement addwitnessaddress

jsgoyette:
This PR adds the necessary type and functions for calling the addwitnessaddress method from the RPC client.

### DIFF
--- a/rpcclient/wallet.go
+++ b/rpcclient/wallet.go
@@ -943,6 +943,44 @@ func (c *Client) GetRawChangeAddress(account string) (btcutil.Address, error) {
 	return c.GetRawChangeAddressAsync(account).Receive()
 }
 
+// FutureAddWitnessAddressResult is a future promise to deliver the result of
+// a AddWitnessAddressAsync RPC invocation (or an applicable error).
+type FutureAddWitnessAddressResult chan *response
+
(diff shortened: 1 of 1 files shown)


rpcclient/wallet.go
```diff
@@ -943,6 +943,44 @@ func (c *Client) GetRawChangeAddress(account string) (btcutil.Address, error) {
...
+	}
+
+	// Unmarshal result as a string.
+	var addr string
+	err = json.Unmarshal(res, &addr)
+	if err != nil {
+		return nil, err
+	}
+
+	return btcutil.DecodeAddress(addr, &chaincfg.MainNetParams)
```
dajohi: What about testnet?

ysangkok:
should this be closed now that https://github.com/btcsuite/btcd/commit/91e5ba1a80082d436cc35965d74e0a41be7f2d99 is in?

jsgoyette:
This is different from 91e5ba1 and in fact builds upon it. While 91e5ba1 adds addwitnessaddress functionality to btcjson, this PR exposes the addwitnessaddress method to the rpcclient (which uses the new function from btcjson). So I would think that you wouldn't want to close it for that reason.

PR #1055, opened 2017-09-15
blockchain: Remove BFDryRun behaviour flag.

jimpo:
This is only used to test block proposals. That has been changed to instead use a new(ish) CheckConnectBlockTemplate function in the blockchain package. The flag complicated the implementation of some chain processing routines and would be difficult to implement with headers-first syncing. Block proposals now may only connect to the tip of the main chain, which is the behavior implemented in Bitcoin Core, bcoin, and possibly other implementations of getblocktemplate.

### DIFF
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -785,12 +785,8 @@ func countSpentOutputs(block *btcutil.Block) int {
 // the chain) and nodes the are being attached must be in forwards order
 // (think pushing them onto the end of the chain).
 //
-// The flags modify the behavior of this function as follows:
-//  - BFDryRun: Only the checks which ensure the reorganize can be completed
-//    successfully are performed.  The chain is not reorganized.
-//
 // This function MUST be called with the chain state lock held (for writes).
-func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List, flags BehaviorFlags) error {
+func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List) error {
 	// All of the blocks to detach and related spend journal entries needed
 	// to unspend transaction outputs in the blocks being disconnected must
 	// be loaded from the database during the reorg check phase below and
@@ -883,12 +879,6 @@ func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List, flags 
 		}
 	}
 
-	// Skip disconnecting and connecting the blocks when running with the
-	// dry run flag set.
-	if flags&BFDryRun == BFDryRun {
-		return nil
-	}
-
 	// Reset the view for the actual connection code below.  This is
 	// required because the view was previously modified when checking if
 	// the reorg would be successful and the connection code requires the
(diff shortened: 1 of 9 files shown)


blockchain/chain.go
```diff
@@ -785,10 +785,6 @@ func countSpentOutputs(block *btcutil.Block) int {
 // the chain) and nodes the are being attached must be in forwards order
 // (think pushing them onto the end of the chain).
 //
-// The flags modify the behavior of this function as follows:
-//  - BFDryRun: Only the checks which ensure the reorganize can be completed
-//    successfully are performed.  The chain is not reorganized.
-//
 // This function MUST be called with the chain state lock held (for writes).
 func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List, flags BehaviorFlags) error {
```
davecgh: I would suggest removing the `flags` parameter altogether since it is no longer used in the function.

blockchain/validate_test.go
```diff
@@ -63,23 +63,88 @@ func TestSequenceLocksActive(t *testing.T) {
 	}
 }
 
-// TestCheckConnectBlock tests the CheckConnectBlock function to ensure it
-// fails.
-func TestCheckConnectBlock(t *testing.T) {
+// TestCheckConnectBlock tests the CheckConnectBlockTemplate function to ensure
```
davecgh: `TestCheckConnectBlockTemplate`

blockchain/validate_test.go
```diff
@@ -63,23 +63,88 @@ func TestSequenceLocksActive(t *testing.T) {
...
+			t.Fatalf("CheckConnectBlockTemplate: Received unexpected error "+
+				"processing block %d: %v", i, err)
+		}
+		if !isMainChain {
+			t.Fatalf("CheckConnectBlockTemplate: Expected block %d to connect "+
+				"to main chain", i)
+		}
+	}
+
+	// The block 3 should fail to connect since it's already inserted.
```
davecgh: s/The block 3/Block 3/

blockchain/validate.go
```diff
@@ -1243,27 +1247,44 @@ func (b *BlockChain) checkConnectBlock(node *blockNode, block *btcutil.Block, vi
 	return nil
 }
 
-// CheckConnectBlock performs several checks to confirm connecting the passed
-// block to the main chain does not violate any rules.  An example of some of
-// the checks performed are ensuring connecting the block would not cause any
-// duplicate transaction hashes for old transactions that aren't already fully
```
davecgh: Please leave the meat of the description here.  Unexported functions don't show up on godoc, so when looking at https://godoc.org/github.com/btcsuite/btcd/blockchain, the information will not be visible.  That is typically why the unexported comments refer to the exported comments for more details.

blockchain/chain.go
jimpo: Good find.

blockchain/validate.go
jimpo: This doesn't seem like a good function to leave a docstring enumerating all block validation rules (since now the function performs more checks than just these). I feel like it is sufficient to say "CheckConnectBlockTemplate fully validates that connecting the passed block to the main chain does not violate any consensus rules, aside from the proof of work requirement", which it does.

But if you feel like  it belongs on here, I'm willing to move it back.

davecgh: Fair enough.  I'm not overly concerned about it here.  I just noticed that the bulk of the comment was moved to the unexported function.  I agree that calling out it validates per the consensus rules is probably good enough.

PR #1060, opened 2017-09-20
btcec: Prevent static initialization of S256

### LINKED ISSUE #1057: Lazily initialize S256 please
By importing btcec our program calls S256 because of the static initializers.

https://github.com/btcsuite/btcd/blob/1bdb71328511a5225d6300c3c08a79172f667996/btcec/signature.go#L33

This causes our program to use 1MB of memory even if we never end up using btcec (in our case we were importing IPFS).

It would be better to allocate all that memory the first time it is needed instead of doing it in a static initializer.

stevenroose:
This is achieved by introducing a new variable `halfOrder` on the
KoblitzCurve struct that is half the order.

Fixes #1057


btcec/signature.go
```diff
@@ -51,8 +47,8 @@ var (
 func (sig *Signature) Serialize() []byte {
 	// low 'S' malleability breaker
 	sigS := sig.S
-	if sigS.Cmp(halforder) == 1 {
-		sigS = new(big.Int).Sub(order, sigS)
+	if sigS.Cmp(S256().N) == 1 {
```
davecgh: You changed from comparing against the half order to the order.

stevenroose: Oh whoops. Late night programming. Weird that the tests didn't catch that though.

dajohi: want to add a test? :)

stevenroose: Done

btcec/signature_test.go
```diff
@@ -418,6 +418,24 @@ func TestSignatureSerialize(t *testing.T) {
 			},
 		},
 		{
+			"valid 4 - s is bigger than half order",
```
davecgh: This test doesn't actually test the condition.  You can provide it by changing `if s.Cmp(halfOrder) == 1 {` to `if s.Cmp(N) == 1 {` and notice it doesn't fail.

stevenroose: That's strange. With this diff
```diff
diff --git a/btcec/signature.go b/btcec/signature.go
index 4392ab41..1e6f2af8 100644
--- a/btcec/signature.go
+++ b/btcec/signature.go
@@ -47,7 +47,7 @@ var (
 func (sig *Signature) Serialize() []byte {
        // low 'S' malleability breaker
        sigS := sig.S
-       if sigS.Cmp(S256().halfOrder) == 1 {
+       if sigS.Cmp(S256().N) == 1 {
                sigS = new(big.Int).Sub(S256().N, sigS)
        }
        // Ensure the encoded bytes for the r and s values are canonical and
```
I get this error when I run the tests:
```
--- FAIL: TestSignatureSerialize (0.00s)
	signature_test.go:453: Serialize #3 (valid 4 - s is bigger than half order) unexpected result:
		got:  3046022100a196ed0e7ebcbe7b63fe1d8eecbdbde03a67ceba4fc8f6482bdcb9606a911404022100971729c7fa944b465b35250c6570a2f31acbb14b13d1565fab7330dcb2b3dfb1
		want: 3045022100a196ed0e7ebcbe7b63fe1d8eecbdbde03a67ceba4fc8f6482bdcb9606a911404022068e8d638056bb4b9a4cadaf39a8f5d0b9fe32b9b9b7749dc145f2db01d826190
```

davecgh: Oh, you're right.  I was thinking this was a test for `signRFC6979` and changed that one to test it.  Looking at the actual test data now though, I can see it is indeed accurate.

PR #1063, opened 2017-09-29
rpctest: Compile current version of btcd and run that.

jimpo:
Previously, rpctest would start a btcd node using the btcd executable in the environment PATH. This caused difficult-to-find issues where the code would be tested against an older version of btcd, or another fork entirely. Now it compiles btcd the first time it is needed and uses that fresh version when launching nodes.

Testing by running in the btcsuite/btcd project as well as in lnd, where roasbeef/btcd is installed as a vendored package.

### DIFF
--- a/integration/rpctest/btcd.go
+++ b/integration/rpctest/btcd.go
@@ -1,0 +1,73 @@
+// Copyright (c) 2017 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package rpctest
+
+import (
+	"fmt"
+	"go/build"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"sync"
+)
+
+var (
+	// compileMtx guards access to the executable path so that the project is
+	// only compiled once.
+	compileMtx sync.Mutex
+
+	// executablePath is the path to the compiled executable. This is the empty
+	// string until btcd is compiled. This should not be accessed directly;
+	// instead use the function btcdExecutablePath().
+	executablePath string
+)
+
+// btcdExecutablePath returns a path to the btcd executable to be used by
+// rpctests. To ensure the code tests against the most up-to-date version of
+// btcd, this method compiles btcd the first time it is called. After that, the
+// generated binary is used for subsequent test harnesses. The executable file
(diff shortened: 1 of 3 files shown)


integration/rpctest/btcd.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+var (
+	// btcdCompileMtx guards access to the btcd path so that it is only compiled
+	// once.
+	btcdCompileMtx sync.Mutex
+
+	// _btcdFilepath is the path to the btcd executable. This is the empty
+	// string until btcd is compiled.
+	_btcdFilepath string
```
davecgh: Can we call this just `compilePath` (no underscore and agnostic for forks)?

integration/rpctest/btcd.go
```diff
@@ -0,0 +1,71 @@
...
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"sync"
+)
+
+var (
+	// btcdCompileMtx guards access to the btcd path so that it is only compiled
+	// once.
+	btcdCompileMtx sync.Mutex
```
davecgh: Similarly, I think `compileMtx` would be better here since it's agnostic for forks.

integration/rpctest/btcd.go
```diff
@@ -0,0 +1,71 @@
...
+func btcdExecutablePath() (string, error) {
+	btcdCompileMtx.Lock()
+
+	// If btcd has already been compiled, just use that.
+	if len(_btcdFilepath) != 0 {
+		btcdCompileMtx.Unlock()
+		return _btcdFilepath, nil
+	}
+
+	defer btcdCompileMtx.Unlock()
```
davecgh: Minor nitpick, but might as well just put the defer after the lock and remove the explicit unlock above.  This is only test code, so saving the defer overhead doesn't seem worth the readability hit to me.  I actually missed this on my first read through and though the variable was used below without the mutex!

davecgh:
Thanks for the udpates.  Looks good.  I've confirmed it runs properly on Windows now as well.  Once it's squashed, I'll get it merged.

jimpo:
Squashed.

Issue #1048, opened 2017-09-04
RFC: Block persistence changes to support headers-first sync


jimpo:
This describes some options for implementing headers-first syncing and how it relates to block persistence. The new blockindex and chainview structures in the `blockchain` package go a long way towards enabling this, but there are some more possible changes that would make it for more efficient.

First of all, the block index is currently only initialized with nodes from the main chain, partly due to how block data is structured in buckets. Instead, we should efficiently load all stored headers into the block index on init. If we ensure that we only store connected headers that satisfy PoW and checkpoint checks, there is very little risk of populating the in-memory index with a ton of invalid or orphaned blocks. As demonstrated in https://github.com/btcsuite/btcd/pull/1034, having a bucket that stores all block headers keyed by `<uint32(height><hash>`, allows the block index to be populated very quickly.

Each `blockNode` in the index should store a bitfield representing its validation state. To support headers-first syncing, we should store with each node: if we have the block data in ffldb, if the block was fully validated, if the block failed validation, and if the block is a descendent of another invalid block. More bits may be used in the future, for example if pruning is implemented, we may want to store whether block info has been pruned from the spend journal. This bit field can be stored alongside the header.

While it is possible to implement headers-first sync with the headers only in memory, they would have to be resynced if the node restarts during IBD. Instead, it seems preferable to store headers independently of block data.

I propose using ffldb to only store/fetch full blocks and creating a bucket with headers & validation state managed by in `chainio.go`. To save storage space, we would drop the block headers from the block index bucket in LevelDB and each row in the bucket would just have the block location. A one-time migration is necessary to populate the new bucket and drop header data from the current block index bucket.

I'm sure there are some complications I am not aware of, so I appreciate any feedback.

davecgh:
Thanks for taking the time to write all this up.  I know we've pretty much discussed this already between various PRs and IRC, but it's good to have it all in one place.

I think moving the management of the block index into the blockchain and adding a field of bitflags for tracking the validation state is definitely the way to go here.  As mentioned, there are certainly going to need to be the ability to do a few things that really aren't very easy to do right now with the block index being managed by `database`.  In addition to the aforementioned cases, I can foresee some additional flags identifying specific portions of the verification process that have been done and performing multi-block validation in parallel.  Obviously there are a lot of considerations in something like that due to the ordering requirements, but nevertheless there are a lot of opportunities here.

[...]

PR #1066, opened 2017-10-09
Track and persist block statuses in block index and bucket

jimpo:
As discussed in https://github.com/btcsuite/btcd/issues/1048, for headers first syncing, we need to track additional state about each block such as whether the full block data is stored or just the header. This adds a status bitfield to each node in the block index and writes the statuses to the persistent block index bucket in LevelDB when necessary.

Depends on https://github.com/btcsuite/btcd/pull/1055, https://github.com/btcsuite/btcd/pull/1050, and https://github.com/btcsuite/btcd/pull/1049.

This modifies the bucket format of the block index bucket. Given that https://github.com/btcsuite/btcd/pull/1049 has not been merged yet, it might be worth holding off on it until this is ready so we only need to support one migration path.


davecgh:
I think it's worth only having a single migration path, although, aside from that final comment on #1049, it's ready to be merged.  I should be able to get the review on #1050 done later today, but I realize it depends on #1049 as well.

blockchain/chainio.go
```diff
@@ -1331,40 +1346,31 @@ func dbFetchBlockByNode(dbTx database.Tx, node *blockNode) (*btcutil.Block, erro
...
 
-// dbStoreBlock stores the provided block in the database. The block header is
-// written to the block index bucket and full block data is written to ffldb.
-func dbStoreBlockHeader(dbTx database.Tx, blockHeader *wire.BlockHeader, height uint32) error {
-	// Serialize block data to be stored. This is just the serialized header.
+// dbStoreBlockNode stores the block header and validation status to the block
+// index bucket. This overwrites the current entry if there exists one.
+func dbStoreBlockNode(dbTx database.Tx, node *blockNode) error {
+	// Serialize block data to be stored.
 	w := bytes.NewBuffer(make([]byte, 0, blockHdrSize))
```
davecgh: This should be `blockHdrSize+1` to account for the status byte as well.  Otherwise, two allocations will have to be done because it will serialize the header and then need more space which will force a realloc.

blockchain/chainio.go
```diff
@@ -1331,40 +1346,31 @@ func dbFetchBlockByNode(dbTx database.Tx, node *blockNode) (*btcutil.Block, erro
...
+	key := blockIndexKey(&node.hash, uint32(node.height))
 	return blockIndexBucket.Put(key, value)
 }
 
-// dbStoreBlock stores the provided block in the database. The block header is
-// written to the block index bucket and full block data is written to ffldb.
+// dbStoreBlock stores the provided block in the database if it is not already
+// there. The full block data is written to ffldb.
 func dbStoreBlock(dbTx database.Tx, block *btcutil.Block) error {
-	if block.Height() == btcutil.BlockHeightUnknown {
```
davecgh: I agree with this change since the database itself no longer case about the height, but it seems to me like this should be in #1049 which switches over to using the chainio bucket for headers.

dajohi:
rebase needed.

dajohi:
rebase needed.

dajohi:
nvm, still waiting on PRs :)

blockchain/accept.go
```diff
@@ -52,28 +52,24 @@ func (b *BlockChain) maybeAcceptBlock(block *btcutil.Block, flags BehaviorFlags)
...
+	newNode.status |= statusDataStored
 
-	// Undo changes to the block index when running in dry run mode.
-	if dryRun {
-		defer func() {
-			b.index.RemoveNode(newNode)
-		}()
+	b.index.AddNode(newNode)
+	b.index.MarkDirty(newNode)
+	err = b.index.flushToDB()
```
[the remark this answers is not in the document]
jimpo: It's here because `flushToDB` creates its own DB transaction. I don't think there's much harm in leaving it here because if the flush fails either 1) a later flush will succeed or 2) in the worst case, dbStoreBlock is called again but is a no-op because it checks if the block is already in ffldb.

jimpo:
@davecgh Can you take another look?

jimpo:
@davecgh I think this is good to go. This is the merged PR with a single migration path.

davecgh:
Thanks for the update.  I'll give it another pass later tonight.

blockchain/upgrade.go
```diff
@@ -92,11 +123,12 @@ func migrateBlockIndex(db database.DB) error {
 	return nil
 }
 
-// readBlockTree reads the old block index bucket and constructs a mapping of
-// each block to all child blocks. This mapping represents the full tree of
-// blocks.
-func readBlockTree(v1BlockIdxBucket database.Bucket) (map[chainhash.Hash][]*chainhash.Hash, error) {
-	childBlocksMap := make(map[chainhash.Hash][]*chainhash.Hash)
+// readBlockTree reads the old block index bucket and construct a mapping of
```
davecgh: constructs

blockchain/upgrade.go
```diff
@@ -65,16 +86,26 @@ func migrateBlockIndex(db database.DB) error {
 
 			var hash chainhash.Hash
 			copy(hash[:], hashBytes[0:chainhash.HashSize])
+			chainContext := blocksMap[hash]
 
-			height, exists := blockHeights[hash]
-			if !exists {
+			if chainContext.height == -1 {
```
davecgh: This will panic if there is no entry for the hash in `blocksMap`.  It probably should either retain the exists check or check it for nil as well.

```Go
if chainContext == nil || chainContext.height == -1 {
...
}
```

blockchain/upgrade.go
```diff
@@ -107,41 +139,60 @@ func readBlockTree(v1BlockIdxBucket database.Bucket) (map[chainhash.Hash][]*chai
...
 		}
 	}
-	return blockHeights
+}
+
+// determineMainChainBlocks traverses the block graph down from the tip to
+// determine which block hashes that are part of the main chain. This function
+// modifies the mainChain field on the blocksMap entries.
+func determineMainChainBlocks(blocksMap map[chainhash.Hash]*blockChainContext,
+	tip *chainhash.Hash) {
```
davecgh: Please keep the function definitions on the same line for consistency.

blockchain/upgrade.go
```diff
@@ -107,41 +139,60 @@ func readBlockTree(v1BlockIdxBucket database.Bucket) (map[chainhash.Hash][]*chai
...
 	queue := list.New()
 
-	// The genesis block is included in childBlocksMap as a child of the zero
-	// hash because that is the value of the PrevBlock field in the genesis
-	// header.
-	for _, genesisHash := range childBlocksMap[zeroHash] {
-		blockHeights[*genesisHash] = 0
+	// The genesis block is included in blocksMap as a child of the zero hash
+	// because that is the value of the PrevBlock field in the genesis header.
+	for _, genesisHash := range blocksMap[zeroHash].children {
```
davecgh: Same here regarding panic if the hash doesn't exist since it will attempt to dereference nil.  It previously wouldn't panic because the range was over the entry and range over nil produce no crash or iterations.

There are several more instances of this below as well.

I will concede that it _shouldn't_ happen that these hashes don't exist because it was built from a supposedly valid chain, so it might not be worth worrying about.

jimpo: Added a check for this case.

blockchain/upgrade.go
jimpo: I feel like this should panic. It should be impossible given that `readBlockTree` iterates over all entries in the bucket and puts them in `blocksMap`.

PR #1078, opened 2017-11-18
txscript: export calcSignatureHash

benma:
This is a useful function for users of this library, and deserves to
be public.


benma:
Requesting a review :smile:

PR #1087, opened 2017-12-03
btcec: Optimize field inversion.

bmperrea:
This uses the addition chain from [Brian Smith's website](https://briansmith.org/ecc-inversion-addition-chains-01#secp256k1_field_inversion) to speed up
the inversion algorithm for btcec/field.go.

### DIFF
--- a/btcec/field.go
+++ b/btcec/field.go
@@ -1119,105 +1119,118 @@ func (f *fieldVal) SquareVal(val *fieldVal) *fieldVal {
 // The field value is returned to support chaining.  This enables syntax like:
 // f.Inverse().Mul(f2) so that f = f^-1 * f2.
 func (f *fieldVal) Inverse() *fieldVal {
-	// Fermat's little theorem states that for a nonzero number a and prime
-	// prime p, a^(p-1) = 1 (mod p).  Since the multipliciative inverse is
-	// a*b = 1 (mod p), it follows that b = a*a^(p-2) = a^(p-1) = 1 (mod p).
-	// Thus, a^(p-2) is the multiplicative inverse.
-	//
-	// In order to efficiently compute a^(p-2), p-2 needs to be split into
-	// a sequence of squares and multipications that minimizes the number of
-	// multiplications needed (since they are more costly than squarings).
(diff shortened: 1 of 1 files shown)


bmperrea:
It gets back the sig-verify speed lost in #1084 😄 .

benchmark                          old ns/op     new ns/op     delta
------------------------------------------------------------------
BenchmarkAddJacobian-8             550           552           +0.36%
BenchmarkAddJacobianNotZOne-8      1124          1128          +0.36%
BenchmarkScalarBaseMult-8          49320         48393         -1.88%
BenchmarkScalarBaseMultLarge-8     50016         48749         -2.53%
BenchmarkScalarMult-8              160763        158805        -1.22%
BenchmarkNAF-8                     867           870           +0.35%
BenchmarkSigVerify-8               256511        251033        -2.14%
BenchmarkFieldNormalize-8          13.8          13.7          -0.72%

bmperrea:
@davecgh Any chance we can get this into 0.13.0?

jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Enhancement

onyb (commented):
@bmperrea Great work on this one!  👏 

While the implementation is correct, I would like  to suggest some updates to the code comments:
* The Haskell function `secp256k1FieldInverseSquaredExponent` is actually calculating the inverse directly, and not the squared inverse. This is because of the additional `andThen add x1` added by you at the end.

  I think we should copy the algorithm verbatim, and our code should closely follow the same, to help future readers.

* Following the above point,  we should explain the addition chain for `prime - 3`, as opposed to `prime - 2`, and only present the final multiplication as an extra step to compute the inverse.

* Finally, the explanation from "It follows the analogy..." until the end of the sentence could be framed in a better way to make it more understandable.

Here's [a diff](https://github.com/bmperrea/btcd/compare/speedup-inversion...onyb:speedup-inversion-v2) containing all the changes I proposed. Let me know if they make sense to you.

allocz (commented):
As of now the field implementation is `dcrec/secp256k1`, I think this PR can be closed.

PR #1108, opened 2018-01-15
Validate PKCS7 padding correctly

AdamISZ:
The removePKCSPadding function did not valide the contents
of the padding; the format is defined in RFC2315.
This commit corrects that and some extra test cases which
trigger padding validation failure are provided.

I realise that this is most likely not important. And I am not aware of anyone using this code right now. And it actually is kind of a corner case as to whether it'd ever matter.

For back-story you might find interesting [this](https://github.com/JoinMarket-Org/joinmarket/pull/191#issuecomment-131111979) example - we were using an old slowaes.py module which likewise did not verify padding; and it turned out that because of the way we were using it, there was indeed a corner case here where it led to something bad happening. I think that'd be rare.

### DIFF
--- a/btcec/ciphering.go
+++ b/btcec/ciphering.go
@@ -208,9 +208,14 @@ func addPKCSPadding(src []byte) []byte {
 func removePKCSPadding(src []byte) ([]byte, error) {
 	length := len(src)
 	padLength := int(src[length-1])
-	if padLength > aes.BlockSize || length < aes.BlockSize {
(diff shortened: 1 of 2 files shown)


jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Bug(ish)

allocz (commented):
The code in question does not exists anymore, I think we should close this PR.

PR #1128, opened 2018-02-20
change string to address type in rpcclient interface

### LINKED ISSUE #1127: rpcclient: Use btcutil.Address type in all interfaces
Right now ImportAddress and some other functions take address as a string while ListUnspentMinMaxAddressesAsync and most other functions take it as btcutil.Address.

Everything function should take btcutil.Address in its interface for type safety.

See also 
* https://github.com/btcsuite/btcd/blob/fe786c93b6f4e0c96d6b19157b72963a9b50293d/rpcclient/wallet.go
* https://github.com/btcsuite/btcutil/blob/fd898ec77a5fdc8feac21487fe115fcdbc67cf7b/address.go#L105

gracenoah:
fixes #1127

### DIFF
--- a/rpcclient/wallet.go
+++ b/rpcclient/wallet.go
@@ -970,14 +970,14 @@ func (r FutureAddWitnessAddressResult) Receive() (btcutil.Address, error) {
 // the returned instance.
 //
 // See AddWitnessAddress for the blocking version and more details.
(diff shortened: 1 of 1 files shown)


Roasbeef:
Needs a rebase to fix the conflict!

gracenoah:

PR #1129, opened 2018-02-20
fix NewImportAddressCmd rescan wrong parameters

### LINKED ISSUE #1126: rpcclient: NewImportAddressCmd doesn't encode rescan: true correctly
NewImportAddressCmd constructs a command with the arguments:

* Address
* Rescan

Bitcoind expects:

* Address
* Account
* Rescan

When you try to make a request with `Rescan == true`, bitcoin throws a type error trying to parse `Rescan` (`true`) as a `string` (`account`). Account is optional, but it's not optional if rescan is set.

See the docs at https://bitcoin.org/en/developer-reference#importprivkey

gracenoah:
fix #1126


gracenoah:
fixed the conflict so this can be merged before #1128 PTAL

PR #1130, opened 2018-02-20
Migrate from glide to dep

stevenroose:
The Go team is pushing it's own dependency tool, [dep](https://github.com/golang/dep).

Since the development of Glide is discontinued and it doesn't currently fulfill all our needs, it makes sense to move to dep.

All binaries build and all tests run fine.

There was an explicit version tag of `1679536dcc895411a9f5848d9a0250be7856448c` for [`go-flags`](github.com/jessevdk/go-flags), introduced [in this commit](https://github.com/btcsuite/btcd/commit/4f12c97d0f198ee99794cadfe7b547fd75d88847). I didn't include this in the first run, and everything seems to work as it is. (Unless I'm missing something.)


stevenroose:
Whoops, forgot Travis of course. On it.

goclean.sh
```diff
@@ -11,21 +11,14 @@
 
 set -ex
 
-# Make sure glide is installed and $GOPATH/bin is in your path.
```
stevenroose: The script doesn't rely on glide nor dep. So I thought it was not necessary to explicitly check if dep was in the gopath.

davecgh:
Thanks for the PR.  I'll check all the deps and get this reviewed over the next few days.

stevenroose:
@davecgh ping

abitrolly:
Looks like a useful step forward.

stevenroose:
I'm not sure what's gonna happen with `dep` and `vgo`. They might both continue to exist with different purposes, or `vgo` might replace `dep` altogether. Might be worthwhile waiting a bit.

jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Outdated

This can be closed immediately - we're using go modules

jcvernaleo:
No longer relevant/outdated.  Closing.

PR #1133, opened 2018-02-24
goclean: Nit that made go vet complain

stevenroose:
I know it's silly to make a 1-character PR, but `go vet` in `goclean.sh` complains about this (I guess with new Go version).


stevenroose:
This has already been changed in master, apparently.

PR #1134, opened 2018-02-26
avoid wasting space in parseScriptTemplate() and fix some typo

qshuai:
PTAL

### DIFF
--- a/txscript/script.go
+++ b/txscript/script.go
@@ -275,7 +275,9 @@ func parseScriptTemplate(script []byte, opcodes *[256]opcode) ([]parsedOpcode, e
 		retScript = append(retScript, pop)
 	}
 
-	return retScript, nil
+	ret := make([]parsedOpcode, len(retScript))
+	copy(ret, retScript)
+	return ret, nil
 }
 
 // parseScript preparses the script in bytes into a list of parsedOpcodes while
--- a/mempool/policy.go
+++ b/mempool/policy.go
@@ -19,7 +19,7 @@ const (
 	// that are considered standard in a pay-to-script-hash script.
 	maxStandardP2SHSigOps = 15
 
-	// maxStandardTxCost is the max weight permitted by any transaction
+	// maxStandardTxWeight is the max weight permitted by any transaction
 	// according to the current default policy.
 	maxStandardTxWeight = 400000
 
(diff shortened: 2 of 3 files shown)


txscript/script.go
```diff
@@ -275,7 +275,9 @@ func parseScriptTemplate(script []byte, opcodes *[256]opcode) ([]parsedOpcode, e
 		retScript = append(retScript, pop)
 	}
 
-	return retScript, nil
+	ret := make([]parsedOpcode, len(retScript))
```
stevenroose: This doesn't fall under the description of the PR.. I also don't really think it's necessary.

qshuai: This is a real P2SH transaction sample:
txid: 5da708e9a903ac5d41a0f3b0731bedf3a2f99522d9978ec4d7d93412157296ce
P2SH script:a914f323f8af174e6829a9c8acf1013f65f18370380987

I will output origin script length and final length of the  result slice in parseScriptTemplate() function. as following:

```
func parseScriptTemplate(script []byte, opcodes *[256]opcode) ([]parsedOpcode, error) {
fmt.Println("script length:", len(script))
         ....
fmt.Println("retScript length:", len(retScript))
return retScript, nil
```

And I using a unit testing here:
```
func TestRealScript(t *testing.T) {
	b, _ := hex.DecodeString("a914f323f8af174e6829a9c8acf1013f65f18370380987")
	_, err := parseScript(b)
	if err != nil {
		t.Error(err)
	}
}
```

And run the unit testing, output:
   script length: 23
   retScript length: 3

So I think there are the memory space of twenty structs(parsedOpcode) wasting!

If there is any error, please let me know.

qshuai: @stevenroose PTAL

stevenroose: @qshuai I'm saying that the PR says "error annotation", while this addresses something else.

Looking at the entire method, it does seem to make sense to do the copy, but please don't mix several things in one PR.

qshuai: @stevenroose Yeah,  you are right. This PR include several commits. Github will merge all my commits in one PR. 
Is there a good practice? 

I am a newer bitcoin coder, and my english is poor. But I just want learn more. Please pardon my misnomer if has.🙂

stevenroose:
@qshuai You can squash these 4 commits as follows:
```
git rebase -i HEAD~4
```
Then in all the lines except the top one (the earliest commit), replace `pick` with `squash` (or `s`) and save. Then an editor will open to let you write a commit message for your squashed commit.

Then you can overwrite your branch here like this:
```
git push --force
```

qshuai:
@stevenroose Thanks for your directions!

qshuai:
@Roasbeef I close this pr first. If necessary, will reopen!

PR #1138, opened 2018-03-04
btcec: Format with gofmt for Go 1.10

stevenroose:
Somehow `goclean.sh` complained about this. It's in a comment section, though, but I guess it's better to have it not break Travis when we add Go 1.10 support.


stevenroose:
If I'm not mistaken, they were also removed in recent changes by @Roasbeef.

PR #1075, opened 2017-11-01
[WIP] Adding gRPC support

stevenroose:
One of the main design differences is that the `grpcServer` takes (only) a `server` as argument, which significantly eases the implementation. The current RPC server uses indirects for all the `server` functionality, sometimes even just [putting top level methods in one of the indirects](https://github.com/btcsuite/btcd/blob/master/rpcadapters.go#L278).

### Progress update: 

TODOs:

 - [x] implement listening
 - [ ] add GRPC dependenies (waiting for dep PR)
 - [ ] correct error handling
 - [ ] REST proxy? (cfr lnd)
 - [ ] in case it will replace the JSON-RPC interface: find a solution for
       getblocktemplate (maybe separate small HTTP server in the mining package 
	   behind `--getblocktemplate <interface>`)
 - [x] update rpcclient: there is a `WrappedBtcdClient` in the `btcrpc` package
	 that uses the respective Go structs like `chainhash.Hash`,
	 `btcutil.Tx|Block`, `wire.MsgTg`,...
 - [ ] update btcctl

Some questions:
 - is using `btcutil.Tx|Block.Serialize` safe instead of `BtcEncode`?

All current methods and the new methods that cover them:

 - [x] addnode: ConnectPeer
 - [x] createrawtransaction: deprecated
 - [x] debuglevel: SetDebugLevel
 - [x] decoderawtransaction: deprecated
 - [x] decodescript: deprecated
 - [x] generate: GenerateBlocks
 - [x] getaddednodeinfo: GetPeers
 - [x] getbestblock: GetBestBlockInfo
 - [x] getbestblockhash: GetBestBlockInfo
 - [x] getblock: GetBlock
 - [ ] **getblockchaininfo: could add to GetNetworkInfo**
 - [x] getblockcount: GetBestBlockInfo
 - [x] getblockhash: GetBlockInfo
 - [x] getblockheader: GetBlockInfo
 - [ ] **getblocktemplate**: see below
 - [x] getconnectioncount: GetSystemInfo
 - [x] getcurrentnet: GetNetworkInfo
 - [x] getdifficulty: GetNetworkInfo or GetMiningInfo
 - [x] getgenerate: GetGenerate
 - [x] gethashespersec: GetHashRate (or GetNetworkInfo or GetMiningInfo for network HR)
 - [ ] **getheaders**
 - [x] getinfo: GetSystemInfo and GetNetworkInfo
 - [x] getmempoolinfo: GetMempoolInfo
[...]

PR #1145, opened 2018-03-13
rpc: Implement GetUnspentTransactionOutputSet

stevenroose:
Requires #1075. Will also be simpler with https://github.com/btcsuite/btcd/pull/1045.

Implements a new asynchronous method to retrieve the entire UTXO set for the best chain (at the moment of calling).

I'm not entirely happy with the way the hash of the chain tip of the moment the UTXO snapshot is taken is communicated. Because gRPC allows methods to return streams, the one value could not be returned in a call. Instead, it is sent by the caller (who sends the best hash at the moment of calling) and the server returns an error if a new block was added to the blockchain in the meantime.


stevenroose:
Idea: `ScanUnspentOutputSet`. Cfr the way Core implemented `scantxoutset`.

stevenroose:
Yeah this is based on the gRPC implementation that I worked on in the past. I'm not doing much work on btcd anymore, but the gRPC impl is quite finished tbh. Whether it will ever land is the question :)

I'd be very happy to guide you through it if you would decide to pick it up.

jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Enhancement
- Outdated (very)

grpcserver.go
```diff
@@ -1261,6 +1261,71 @@ func (s *grpcServer) SubmitTransaction(ctx context.Context, req *btcrpc.SubmitTr
 	return &btcrpc.SubmitTransactionResponse{Hash: tx.Hash()[:]}, nil
 }
 
+func (s *grpcServer) GetUnspentTransactionOutputSet(req *btcrpc.GetUnspentTransactionOutputSetRequest, stream btcrpc.Btcd_GetUnspentTransactionOutputSetServer) error {
```
allocz, line 1264: This feature is still needed? The PR has no activity for some years, should we close this PR?

Issue #1146, opened 2018-03-14
rpcclient: Add estimatesmartfee support


gracenoah:
The JSON-RPC method `estimatesmartfee` was added to bitcoind on version 0.12 (https://github.com/bitcoin/bitcoin/pull/6134) but there is currently no way of reaching it through btcd.

hiromaily:
+1

hiromaily:
`EstimateFee()` returns error.
```
-32: estimatefee is deprecated and will be fully removed in v0.17. To use estimatefee in v0.16, restart bitcoind with -deprecatedrpc=estimatefee.
Projects should transition to using estimatesmartfee before upgrading to v0.17
```
So this should be fixed.

johandroz:
Got the same issue. You can use a RawRequest to get the `estimatesmartfee`.
I did something like:
```go

// RPCClient is the interface of the RPC client.
type RPCClient interface {
	RawRequest(method string, params []json.RawMessage) (json.RawMessage, error)
}

// EstimateMode encodes the possible estimation modes of estimatesmartfee
type EstimateMode string

const (
	// UnsetEstimeMode identifies the UNSET estimation strategy used by estimatesmartfee
	UnsetEstimeMode EstimateMode = "UNSET"
	// ConservativeEstimeMode identifies the CONSERVATIVE estimation strategy used by estimatesmartfee
	ConservativeEstimeMode EstimateMode = "CONSERVATIVE"
	// EconomicalEstimeMode identifies the ECONOMICAL estimation strategy used by estimatesmartfee
	EconomicalEstimeMode EstimateMode = "ECONOMICAL"
)

// EstimateSmartFeeReply is the result of the EstimateSmartFee call.
type EstimateSmartFeeReply struct {
	FeeRate float64 `json:"feerate"`
	Blocks  int64   `json:"blocks"`
}

// EstimateSmartFee estimates the approximate fee per kilobyte needed for a transaction to begin
// confirmation within 'numBlocks' blocks if possible and return the number of blocks
// for which the estimate is valid.
// The fee estimate ("ECONOMICAL", "CONSERVATIVE" or "" (default to CONSERVATIVE)) mode defines whether
// to return a more conservative estimate which also satisfies a longer history. A conservative estimate
// potentially returns a higher feerate and is more likely to be sufficient for the desired target, but
// is not as responsive to short term drops in the prevailing fee market.
[...]

PR #1147, opened 2018-03-14
Add method to bridge the estimatesmartfee JSON-RPC method

gracenoah:
fix https://github.com/btcsuite/btcd/issues/1146


gracenoah:

PR #1148, opened 2018-03-16
server: Fix incorrect log message format

stevenroose:


server.go
```diff
@@ -862,15 +862,15 @@ func (sp *serverPeer) OnGetAddr(_ *peer.Peer, msg *wire.MsgGetAddr) {
 	// Do not accept getaddr requests from outbound peers.  This reduces
 	// fingerprinting attacks.
 	if !sp.Inbound() {
-		peerLog.Debugf("Ignoring getaddr request from outbound peer ",
+		peerLog.Debugf("Ignoring getaddr request from outbound peer "+
 			"%v", sp)
```
red010b37: Shouldnt this be

peerLog.Debugf("Ignoring getaddr request from outbound peer %v", sp)

server.go
```diff
@@ -862,15 +862,15 @@ func (sp *serverPeer) OnGetAddr(_ *peer.Peer, msg *wire.MsgGetAddr) {
...
 		return
 	}
 
 	// Only allow one getaddr request per connection to discourage
 	// address stamping of inv announcements.
 	if sp.sentAddrs {
-		peerLog.Debugf("Ignoring repeated getaddr request from peer ",
+		peerLog.Debugf("Ignoring repeated getaddr request from peer "+
 			"%v", sp)
```
red010b37: Same as above

peerLog.Debugf("Ignoring getaddr request from outbound peer %v", sp)

server.go
stevenroose: Yes, but then the line is too long, so we break it.

server.go
stevenroose: We adhere to a ling length of 100 chars I think. (Vim shows when the line gets too long.) That's why it's broken.

stevenroose: The actual change is the comma for the plus.

red010b37: Yea I saw the comma replacement.  

Excuse my ignorance, but why 100 char length? As you are adding in code complexity for the sake of convention, - sure its simple - just trying to understand as I want to contribute over time to this project

red010b37: I thought the 100 chars was more of a guide

stevenroose: As far as I came across in the code (been reading it a lot lately), it's applied quite rigorously everywhere except for function declarations.

red010b37: ok i think that should looked into as that the 100 char limit can add complexity where it is not needed

jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Enhancement

PR #1150, opened 2018-03-21
Wrong BOLT11 prefix in regtest

vapopov:
Initially was observed in `lnd` project for this issue https://github.com/lightningnetwork/lnd/issues/882

if to try to generate invoice for `regtest` net, we get encoded invoice with wrong prefix in human readable part

### DIFF
--- a/chaincfg/params.go
+++ b/chaincfg/params.go
@@ -379,7 +379,7 @@ var RegressionNetParams = Params{
 
 	// Human-readable part for Bech32 encoded segwit addresses, as defined in
 	// BIP 173.
-	Bech32HRPSegwit: "tb", // always tb for test net
+	Bech32HRPSegwit: "bcrt", // always bcrt for reg test net
 
 	// Address encoding magics
 	PubKeyHashAddrID: 0x6f, // starts with m or n


stevenroose:
I can't find any mention of this prefix in either [BOLT11](https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md) or [BIP 173 (bech32)](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki).

vapopov:
сс @DanielWeigl

DanielWeigl:
@stevenroose yes, thats true - but that also makes it impossible to test payments to invoices across different implementations in regtest

I also already proposed to add a clarification to bolt11 https://github.com/lightningnetwork/lightning-rfc/pull/397

I think it would make sense to have a common format in regtest, or?

DanielWeigl:
FYI, current state:

 - bitcoin nodes:
   - btcd: `lntb...`
   - bitcoinCore: `lnbcrt...`

 - lightning nodes:
   - eclair: `lntb...`
   - lnd: `lntb...`
   - c-lightning:`lnbcrt...`

(all nodes run in regtest) 

So, `lnbcrt` is currently in minority, but i still think it would make more sense to use it - but whatever the decision is, i guess it would be really needed to have the same.

PR #1151, opened 2018-03-26
Add progress percentage to sync log, based on timestamps

### LINKED ISSUE #1092: Display Sync Percentage
It would be nice if "SYNC" log messages displayed a sync progress percentage like the reference implementation does.

For context, the reference implentation's UpdateTip message looks like this:
`2014-06-27 12:59:10 UpdateTip: new best=00000000000000002a03b175f183c3bbc440f5c2a171e958bfd504582f711f70  height=308004  log2_work=79.405055  tx=41447698  date=2014-06-27 00:04:17 progress=0.998105`

Having a sync percentage makes it somewhat more clear how long a blockchain sync will take

If there is interest, I'd be happy to work on this

svarogg:
This adds `progress xx.xx%` clause to the SYNC log messages, making it a bit clearer how long will sync take.
Bases the calculation on time differences between now, first block and current block. It is not perfect, but I'm not sure a better approach exists and is easily achievable.

This fixes #1092.

### DIFF
--- a/netsync/blocklogger.go
+++ b/netsync/blocklogger.go
@@ -8,6 +8,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/btcsuite/btcd/blockchain"
 	"github.com/btcsuite/btclog"
 	"github.com/btcsuite/btcutil"
 )
@@ -19,6 +20,7 @@ type blockProgressLogger struct {
 	receivedLogBlocks int64
 	receivedLogTx     int64
 	lastBlockLogTime  time.Time
+	firstBlockTime    time.Time
 
 	subsystemLogger btclog.Logger
 	progressAction  string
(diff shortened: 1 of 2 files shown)


stevenroose:
I don't think this makes much sense. Block times are really not an indicator for how far the sync is progressed. The first 3/4 of the chain is very sparse. Then blocks get full gradually and start to take a lot more time. Then the last checkpoint is gone and btcd start validating txs so it again slows down. Then there is segwit with even bigger blocks, ... 

Whatever new thing comes next, it will always end up being a "windows load bar" like I call them, going very fast in the beginning, and slowing down a ton at the end. (A typical "windows load bar" is characterized by going smooth and hanging at 99%, though, but you get the gist.)

stevenroose:
Besides, if some GUI wants to show a progress, we all know the first block is 3/1/2009.

svarogg:
First block is 3/1/2009 in mainnet, but what about testnets, etc? 

Anyway, I did this mostly to get acquainted with the code, according to suggestion in the issue. I don't see a better approach to approximating progress, and I must say that while syncing my node, these prints were useful to get some sense on where we're standing.

If there's anything that can make this code mergeable please let me know and I'll try to do it, otherwise feel free to decline this PR.

petrhejna:
Jut to reference for those who googled this issue. Here is some snippet how to calculate percentage of blocks downloaded:

> **Note**: It uses `jq` which is not usually present on linux distros and need to be installed. (https://linuxhint.com/bash_jq_command/)

```bash
 jq -n $(btcctl --rpcuser=REPLACE --rpcpass=REPLACE getinfo | jq -r '.blocks')/$(curl -s https://blockchain.info/q/getblockcount)*100
```

jakesyl:
@jcvernaleo (as per #1530)

- Low priority
- Enhancement

By the way- thanks @petrhejna

jakesyl:
I think we need to rebase this as per [this comment](https://github.com/btcsuite/btcd/pull/1558#issuecomment-633754000). Would love to see this merged

netsync/blocklogger.go
```diff
@@ -66,9 +74,13 @@ func (b *blockProgressLogger) LogBlockHeight(block *btcutil.Block) {
 	if b.receivedLogTx == 1 {
 		txStr = "transaction"
 	}
-	b.subsystemLogger.Infof("%s %d %s in the last %s (%d %s, height %d, %s)",
+
+	lastBlockTime := block.MsgBlock().Header.Timestamp
+	progress := 100 - (float64(now.Sub(lastBlockTime)) / float64(now.Sub(b.firstBlockTime)) * 100)
```
allocz, line 79: Wouldn't this be more readable using the following?
```
progress := 100 - (float64(time.Since(lastBlockTime)) / float64(time.Since(b.firstBlockTime)) * 100)
```

PR #1155, opened 2018-04-07
Add SubtractFeeFromAmount params to SendToAddressCmd

void-main:
As per [sendtoaddress documentation](https://bitcoin.org/en/developer-reference#sendtoaddress), there is a parameter named `Subtract Fee From Amount` which is a boolean value.

This could be useful especially when the sender wants to pay the fee from the transaction amount. For example, if you're building a bitcoin exchange, user wants to withdraw all her bitcoins, say 0.5 BTC. There should be some fee to be paid to the miners, which should be subtracted from this 0.5 BTC.

Without this parameter, one have to calculate this fee and extract it from the balance manually, which is a waste of time and logic.


void-main:
@stevenroose totally agreed. but i'm using v1 myself, so i have to change this code anyways in my project.

nmarley:
@stevenroose Would this move to gRPC mean breaking compatibility w/the Bitcoin Core jsonrpc server?

stevenroose:
@nmarley that compatibility is broken already anyways. It also doesn't really make sense to have the btcd client compatible with Core. And besides, it would be very easy to create a "fork" of the client code that *is* compatible with Core. Call it `corerpcclient` ;)

nmarley:
From my personal spot testing it's working, but I haven't tested 100% of the features. I would think you'd want them to be compatible -- why doesn't that make sense? The daemon strives for compatibility with the same consensus rules, right?

The client code is also a library which enables other interfaces written in Go, and I don't think the server implementation should matter. This would be a much looser coupling than switching over to gRPC, although I can see that as a viable option. What would the benefits of moving to gRPC?

Sorry, this is starting to hi-jack the PR thread. Where should this discussion go?

void-main:
I'm with @nmarley for this compatibility thing. I'm running `bitcoind` as daemon service, and using `btcd`'s rpcclient to build my app (using golang).

Personally I agreed using gRPC might be more universal, the code could be better structured, loosely coupled, but for my personal use, I'll have to stick to this rpcclient and make some changes to its code.

With that said, any one of you guys willing to help me reviewing the code. Thanks. :)

jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Enhancement

Issue #1152, opened 2018-03-26
Syncing with full hard disk is not handled gracefully


svarogg:
I've been syncing the mainnet blockchain to a hard-drive that apparently didn't have enough space to hold the whole blockchain.

At a certain point I saw this log:
```
2018-03-26 11:08:38.107 [ERR] SYNC: Failed to process block 00000000000005159e0de79c2a877a994d9a74509b2019a2051aa727b0fc44be: failed to write block to file 4 at offset 238509918: write /home/mike/.btcd/data/mainnet/blocks_ffldb/000000004.fdb: no space left on device: write /home/mike/.btcd/data/mainnet/blocks_ffldb/000000004.fdb: no space left on device
```

Which is good.
However, after that btcd started spamming the following logs very quickly:
```
2018-03-26 11:08:38.109 [INF] CHAN: Adding orphan block 00000000000006a2c0eb97ce99a2f3d749d2758feb600500b8dc6b15a4c21f79 with parent 00000000000005159e0de79c2a877a994d9a74509b2019a2051aa727b0fc44be  
2018-03-26 11:08:38.110 [INF] CHAN: Adding orphan block 00000000000000a438d46a1b52484b51cd90f4037409f6754c4943736bb7eb3d with parent 00000000000006a2c0eb97ce99a2f3d749d2758feb600500b8dc6b15a4c21f79
etc
```

Which isn't that good. Making it very hard to understand what the core problem was (I had to restart a couple of times and stop the moment spamming starts to find the first (real) error message).

I would expect btcd to stop sync until some hard disk space is free, or alternatively stop altogether.

I can try solving this myself if you could tell me which approach is preferred.

abitrolly:
I remember we've run into the same issue while running `btcd` on Google Cloud instance.

PR #1156, opened 2018-04-07
database: Prevent write transaction with low disk space

stevenroose:
I know it's kind of a hack, but it solves https://github.com/btcsuite/btcd/issues/1152.

In principle it does make sense to not try to operate with very little disk space, since it can cause all kinds of unexpected problems.

### DIFF
--- a/database/ffldb/disk_windows.go
+++ b/database/ffldb/disk_windows.go
@@ -1,0 +1,33 @@
+// +build windows
+
+// Copyright (c) 2013-2018 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package ffldb
+
+import (
+	"os"
+	"syscall"
+	"unsafe"
+)
+
+// getAvailableDiskSpace returns the number of bytes of available disk space.
+func getAvailableDiskSpace() (uint64, error) {
+	wd, err := os.Getwd()
+	if err != nil {
+		return 0, err
+	}
+
+	h := syscall.MustLoadDLL("kernel32.dll")
+	c := h.MustFindProc("GetDiskFreeSpaceExW")
+
+	var freeBytes int64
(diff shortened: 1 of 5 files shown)


zquestz:
@stevenroose you might want to look at https://github.com/gcash/bchd/pull/76

It would be better to use the db path so you are assured you are measuring the correct volume.

jakesyl:
@jcvernaleo (as per #1530)

- Low priority
- Bug

Linked Issues:
#1152

database/ffldb/disk_windows.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"os"
+	"syscall"
+	"unsafe"
+)
+
+// getAvailableDiskSpace returns the number of bytes of available disk space.
+func getAvailableDiskSpace() (uint64, error) {
+	wd, err := os.Getwd()
```
allocz, line 17: I think we should use the database directory here, right?

allocz (commented):
The patch seems to be a bit buggy, tested on linux here, and when the storage gets full, I get the following output:
```
2026-09-16 17:56:39.774 [ERR] BCDB: write /mnt/tmpmount/.btcd/mainnet/blocks_ffldb/000000001.fdb: no space left on device. Cannot save any more blocks due to the disk being full -- exiting
```

When starting the node again, this is the output:
```
2026-09-16 18:06:13.928 [ERR] BTCD: write /mnt/tmpmount/.btcd/mainnet/blocks_ffldb/metadata/MANIFEST-000162: no space left on device: write /mnt/tmpmount/.btcd/mainnet/blocks_ffldb/metadata/MANIFEST-000162: no space left on device
```

I think 25MiB is less than the data being flushed to disk, so I would instead require something like at least 1GiB, or N times the max cache size, so that we can be sure that the amount of data flushed to disk won't be more than the available space in the data mountpoint.

PR #1162, opened 2018-04-16
Update README.md

jraedisch:
Use internet archive for old conformal blog entry link.


totaloutput:
@maguayo totally agree on adding a small text, it took me a while to figure it out and probably a lot of people are spending their time to figure it out too...

PR #1165, opened 2018-04-17
Implemented the EOS canonical signature scheme.

abourget:
References:
* https://steemit.com/steem/@dantheman/steem-and-bitshares-cryptographic-security-update
* https://github.com/cryptonomex/secp256k1-zkp/blob/secp256k1-zkp/src/secp256k1.c#L187
* https://github.com/EOSIO/eos/blob/master/libraries/fc/src/crypto/elliptic_impl_priv.cpp#L96
* https://github.com/EOSIO/eosjs-ecc/blob/master/src/signature.js#L178


btcec/signature.go
```diff
@@ -352,6 +352,10 @@ func SignCompact(curve *KoblitzCurve, key *PrivateKey,
 	// bitcoind checks the bit length of R and S here. The ecdsa signature
 	// algorithm returns R and S mod N therefore they will be the bitsize of
 	// the curve, and thus correctly sized.
+	return makeCompact(curve, sig, key, hash, isCompressedKey)
+}
+
+func makeCompact(curve *KoblitzCurve, sig *Signature, key *PrivateKey, hash []byte, isCompressedKey bool) ([]byte, error) {
```
stevenroose: There is no documentation on this one.

And the method is used before it is declared.

btcec/signature.go
```diff
@@ -446,7 +450,13 @@ func signRFC6979(privateKey *PrivateKey, hash []byte) (*Signature, error) {
 
 // nonceRFC6979 generates an ECDSA nonce (`k`) deterministically according to RFC 6979.
 // It takes a 32-byte hash as an input and returns 32-byte nonce to be used in ECDSA algorithm.
-func nonceRFC6979(privkey *big.Int, hash []byte) *big.Int {
+func nonceRFC6979(privkey *big.Int, hash []byte, nonce int) *big.Int {
```
stevenroose: This diverges from RFC6979 spec, so new logic should move to another method.

btcec/privkey.go
```diff
@@ -71,3 +93,20 @@ func (p *PrivateKey) Serialize() []byte {
 	b := make([]byte, 0, PrivKeyBytesLen)
 	return paddedAppend(PrivKeyBytesLen, b, p.ToECDSA().D.Bytes())
 }
+
+func isCanonical(compactSig []byte) bool {
```
stevenroose: documentation missing

btcec/privkey.go
```diff
@@ -59,7 +60,28 @@ func (p *PrivateKey) ToECDSA() *ecdsa.PrivateKey {
 // is deterministic (same message and same key yield the same signature) and canonical
 // in accordance with RFC6979 and BIP0062.
 func (p *PrivateKey) Sign(hash []byte) (*Signature, error) {
-	return signRFC6979(p, hash)
+	return signRFC6979(p, hash, 0)
+}
+
+// SignCanonical goes through signatures and returns only a canonical
+// representations.  This matches the EOS blockchain expectations.
+func (p *PrivateKey) SignCanonical(curve *KoblitzCurve, hash []byte) ([]byte, error) {
```
stevenroose: `Canonical` seems quite general here, while it's just EOS-general. Should be renamed to reflect that.

Also, some references to the algo would be nice. F.e. the magic number 25 should be explained somehow.

PR #1174, opened 2018-04-27
netsync: Remove unnecessary buffered channel

stevenroose:
Somehow it was a buffered channel to not have the continue statement.

### DIFF
--- a/netsync/manager.go
+++ b/netsync/manager.go
@@ -1176,6 +1176,7 @@ out:
 						isOrphan: false,
 						err:      err,
 					}
+					continue out
 				}
 
 				msg.reply <- processBlockResponse{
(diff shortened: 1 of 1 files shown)


jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Enhancement

netsync/manager.go
```diff
@@ -1313,6 +1176,7 @@ out:
 						isOrphan: false,
 						err:      err,
 					}
+					continue out
```
allocz, line 1179: In fact, this is strange, as in the error case, we still send a second reply without the error, but the `ProcessBlock()` will only see the first reply and ignore the second one. I think you're right here, as I've ran the tests with the patch and also synced the node up to some height and everything seems to keep working.

If the buffered channel is really needed, I think we should at least document why, since this is the only reply channel in the file which is buffered.

@Roasbeef, what are your thoughts on this?

PR #1175, opened 2018-05-01
add gettxoutsetinfo JSON-RPC client method

gracenoah:


gracenoah:
Looks like lint is failing on master. I'll rebase once that's fixed.

gracenoah:

PR #1176, opened 2018-05-03
add watchonly argument to listtransactions api client

gracenoah:


gracenoah:

Issue #1170, opened 2018-04-23
shall we restrict the size of input in btcec.FieldVal.SetByteSlice(b []byte)? 


chaintechlab:
According to the comment, " Only the first 32-bytes are used.  As a result, it is up to the caller to ensure numbers of the appropriate size are used or the value will be truncated."
However, in the code:
 ```
for i := 0; i < len(b); i++ {
        if i < 32 {
            b32[i+(32-len(b))] = b[i]
        }
    }
```
This will crash if len(b) > 32. Shall we add 
`b = b[:32]`
?

PR #1178, opened 2018-05-05
btcec: Avoid panic in FieldVal.SetByteSlice

stevenroose:
When called with a value larger than 32 bytes.

Fixes https://github.com/btcsuite/btcd/issues/1170.

### DIFF
--- a/btcec/field.go
+++ b/btcec/field.go
@@ -220,6 +220,10 @@ func (f *fieldVal) SetBytes(b *[32]byte) *fieldVal {
 // The field value is returned to support chaining.  This enables syntax like:
 // f := new(fieldVal).SetByteSlice(byteSlice)
 func (f *fieldVal) SetByteSlice(b []byte) *fieldVal {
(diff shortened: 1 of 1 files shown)


jakesyl:
@jcvernaleo (as per #1530)
- High priority
- Bug

Linked Issues:
#1170

jcvernaleo:
Decred seems to have identical code as we do here:
https://github.com/decred/dcrd/blob/master/dcrec/secp256k1/field.go#L217
Does anyone know if this bug has been brought up there?  Might be worth checking with them to see if they think this is an issue as well.

jcvernaleo:
Also, I think this is a change that should really have a test case if possible.

onyb:
@jcvernaleo and @jakesyl: See https://github.com/btcsuite/btcd/pull/1602 for a slightly faster implementation, adapted from dcrd.

jcvernaleo:
Closing as outdated/dealt with elsewhere.

PR #1185, opened 2018-05-25
change addrtime stamp

ludete:
According to the comments, the timestamp should be after the now 24 hours, and the probability that the different nodes will differ by 10 minutes is too great. So i think this should be followed by the comments.


jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Enhancement

Should probably be closed

jakesyl (approved):

jakesyl:
Came across this in another test and does look to indeed be a bug. Can you rebase or force push so we can have the tests run?

PR #1186, opened 2018-05-25
logic error for OnFilterAdd

ludete:
according to the logic And errors news, when no fileteradd request, this peer should be closed. According to the current logic(	sp.filter.Add(msg.Data) )，in the following code snippet, it will not do anything but exit directly.


Roasbeef:
Fixed by #1169.

PR #1189, opened 2018-05-31
Change 'TxSha' parameter to 'TxID'/'TxHash' for the 'txaccepted' method within the JSON_RPC_AI

matadormel:
The 'txaccepted' method is currently showing an out of date paramater within the json_rpc_api. The first parameter is currently set to 'TxSha', however it should be set to either 'TxID' or 'TxHash' as per https://github.com/decred/dcrd/blob/ed7c45a3e114a4d09976aa5e7ed195412660bbae/dcrjson/chainsvrwsntfns.go


davecgh:
Wrong repo

PR #1201, opened 2018-06-22
chain: remove checking status for fetching cfheaders

vapopov:
Should be related to this issue: https://github.com/lightningnetwork/lnd/issues/511, in case of `neutrino` when we reach this error  -`btcd` sends nothing by request `getcfheaders` and then it breaks sequence of synced `headers` and `cfheaders`, by status command `getinfo` we see that it almost synced but actually we have no `cfheaders` on the filesystem

In our case after intensive load of btcd, it stops sending `cfheaders` because of status of the node, which is not `statusValid` but `statusDataStored`, somehow this status changes for many blocks, in my second case it even happened with block height 1. 

Probably this validation shouldn't be there, because for instance when we send block headers we don't do this kind of validation, only when we accept new blocks for best height.

I didn't investigate how it could that happened with switching statuses, might be some issue with flushing db. But anyway this fix the issue with sync of https://github.com/lightninglabs/neutrino 

cc @jimpo

### DIFF
--- a/blockchain/blockindex.go
+++ b/blockchain/blockindex.go
@@ -251,8 +251,9 @@ func newBlockIndex(db database.DB, chainParams *chaincfg.Params) *blockIndex {
 // This function is safe for concurrent access.
 func (bi *blockIndex) HaveBlock(hash *chainhash.Hash) bool {
 	bi.RLock()
+	defer bi.RUnlock()
+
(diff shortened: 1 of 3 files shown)


Roasbeef:
This is no longer needed as #1243 fixed the fundamental issue.

Issue #1194, opened 2018-06-14
I have a error  with  CHAN: REORGANIZE:


littlegirlpppp:
sunbodeMBP:bin sunbo$ ./btcd
2018-06-14 14:14:29.253 [INF] BTCD: Version 0.12.0-beta
2018-06-14 14:14:29.253 [INF] BTCD: Loading block database from '/Users/sunbo/Library/Application Support/Btcd/data/mainnet/blocks_ffldb'
2018-06-14 14:14:29.326 [INF] BTCD: Block database loaded
2018-06-14 14:14:29.338 [INF] INDX: cf index is enabled
2018-06-14 14:14:29.338 [INF] CHAN: Loading block index...
2018-06-14 14:14:31.483 [INF] CHAN: Chain state (height 384560, hash 000000000000000005a15c23c9f1d67a8e5d93944e73fb74a4177c0611992d02, totaltx 93547834, work 14942772535223485067948946)
2018-06-14 14:14:31.497 [INF] RPCS: RPC server listening on 127.0.0.1:8334
2018-06-14 14:14:31.497 [INF] RPCS: RPC server listening on [::1]:8334
2018-06-14 14:14:31.504 [INF] AMGR: Loaded 278 addresses from file '/Users/sunbo/Library/Application Support/Btcd/data/mainnet/peers.json'
2018-06-14 14:14:31.506 [INF] CMGR: 40 addresses found from DNS seed seed.bitcoin.sipa.be
2018-06-14 14:14:31.513 [INF] CMGR: Server listening on [::]:8333
2018-06-14 14:14:31.515 [INF] CMGR: 38 addresses found from DNS seed dnsseed.bitcoin.dashjr.org
2018-06-14 14:14:31.515 [INF] CMGR: 38 addresses found from DNS seed seed.bitcoin.jonasschnelli.ch
2018-06-14 14:14:31.517 [INF] CMGR: Server listening on 0.0.0.0:8333
2018-06-14 14:14:31.580 [INF] CMGR: 33 addresses found from DNS seed dnsseed.bluematt.me
2018-06-14 14:14:31.635 [INF] CMGR: 41 addresses found from DNS seed seed.bitcoinstats.com
2018-06-14 14:14:31.676 [INF] CMGR: 40 addresses found from DNS seed seed.bitnodes.io
2018-06-14 14:14:32.527 [INF] SYNC: New valid peer 64.99.224.88:8333 (outbound) (/Satoshi:0.16.0/)
[...]

PR #1205, opened 2018-06-26
fix: program crash when detachNodes is empty

liyue201:
fix issue： https://github.com/btcsuite/btcd/issues/1194


Roasbeef:
Before we merge this, we need to figure out exactly _why_ the `detchNodes` was empty in this case. Did this occur on testnet?

liyue201:
I've never met this on testnet.  it occur on mainnet randomly

Roasbeef:
Fixed by #1215

PR #1217, opened 2018-07-13
rpcclient: expose the Node command for RPC API.

halseth:
The `AddNode` command is not particularly easy to use for managing connections, this PR exposes the more useful `Node` command for the `rpcclient`.

### DIFF
--- a/rpcclient/net.go
+++ b/rpcclient/net.go
@@ -63,6 +63,40 @@ func (c *Client) AddNode(host string, command AddNodeCommand) error {
 	return c.AddNodeAsync(host, command).Receive()
 }
 
+// FutureNodeResult is a future promise to deliver the result of a NodeAsync
+// RPC invocation (or an applicable error).
+type FutureNodeResult chan *response
+
+// Receive waits for the response promised by the future and returns an error if
+// any occurred when performing the specified command.
+func (r FutureNodeResult) Receive() error {
+	_, err := receiveFuture(r)
+	return err
+}
+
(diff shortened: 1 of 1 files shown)


rpcclient/net.go
```diff
@@ -63,6 +63,40 @@ func (c *Client) AddNode(host string, command AddNodeCommand) error {
...
+	return c.sendCmd(cmd)
+}
+
+// Node attempts to perform the passed node command on the host.
+// For example, it can be used to add or a remove a persistent peer, or to do
+// connect or diconnect a non-persistent one.
+//
+// The connectSubCmd should be set either "perm" or "temp", depending on
+// whether we are targetting a persistent or non-persistent peer. Passing nil
+// will cause the default value to be used.
```
cfromknecht: what is the default value?

rpcclient/net.go
```diff
@@ -63,6 +63,40 @@ func (c *Client) AddNode(host string, command AddNodeCommand) error {
 	return c.AddNodeAsync(host, command).Receive()
 }
 
+// FutureNodeResult is a future promise to deliver the result of an NodeAsync
```
cfromknecht: s/an NodeAsync/a Node Async/

rpcclient/net.go
halseth: Fixed.

rpcclient/net.go
halseth: Fixed.

PR #1221, opened 2018-07-18
Add note about using gencerts when listening on specific interfaces

kulpreet:
Simple note in docs for using gencerts


jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Documentation

Just a comment, can be merged now if you think it's in scope

PR #1222, opened 2018-07-18
config/peer: make trickleTimeout configurable

halseth:
Useful for speeding up tests etc.


peer/peer.go
```diff
@@ -268,6 +264,10 @@ type Config struct {
 	// Listeners houses callback functions to be invoked on receiving peer
 	// messages.
 	Listeners MessageListeners
+
+	// trickleTimeout is the duration of the ticker which trickles down the
```
cfromknecht: s/trickleTimeout/TrickleTimeout

halseth: Fixed.

PR #1223, opened 2018-07-19
integration/rpctest: disable the txindex by default and allow sending txs without change outputs

wpaulino:
In this commit, we disable the txindex by default as the flag cannot be
overwritten by btcd once set.


integration/rpctest/node.go
```diff
@@ -115,10 +115,6 @@ func (n *nodeConfig) arguments() []string {
 	args = append(args, fmt.Sprintf("--rpccert=%s", n.certFile))
 	// --rpckey
 	args = append(args, fmt.Sprintf("--rpckey=%s", n.keyFile))
-	// --txindex
-	args = append(args, "--txindex")
```
halseth: Might be a bit reckless to change the defaults. Could you gate the change behind a flag in the nodeConfig instead?

PR #1229, opened 2018-07-27
rpcclient: add DisableTLSVerify to allow easy testing

ledzep2:
In testing environments we usually use self-generated certificates on servers (default settings), which must be shared beforehand to allow clients to connect properly. This is cumbersome sometimes. Adding this option (essentially tlsConfig.InsecureSkipVerify) allows some degree of security (no cleartext user/pass) without extra effort in setting up. It could be useful in dev and testing environments.

### DIFF
--- a/rpcclient/infrastructure.go
+++ b/rpcclient/infrastructure.go
@@ -1071,6 +1071,11 @@ type ConnConfig struct {
 	// the wire in cleartext.
 	DisableTLS bool
 
+	// DisableTLSVerify specifies whether TLS certificates should be checked.
+	// Disabling the check will allow potential man-in-the-middle attack. For
+	// testing only
+	DisableTLSVerify bool
+
 	// Certificates are the bytes for a PEM-encoded certificate chain used
 	// for the TLS connection.  It has no effect if the DisableTLS parameter
 	// is true.
@@ -1135,6 +1140,10 @@ func newHTTPClient(config *ConnConfig) (*http.Client, error) {
 			tlsConfig = &tls.Config{
 				RootCAs: pool,
 			}
+		} else {
+			tlsConfig = &tls.Config{
+				InsecureSkipVerify: config.DisableTLSVerify,
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Enhancement

@ledzep2 had the same issue, just commented out the verification on my fork

jakesyl:
@Roasbeef would like to get your input here. I know there are some security concerns around allowing users to do this and this is behavior we'd explicitly like to discourage

jcvernaleo:
While not dead set against this, I'm mostly opposed to this (but am willing to go with @Roasbeef's opinion here).  Testing should be as close to real world as possible so you probably should use tlsverify in testing situations as well.

jakesyl:
@jcvernaleo I'm pretty agnostic here. The counterargument here would be something like this:

btcd doesn't have a wallet API and is used pretty heavily inside private networks where SSL adds unnecessary overhead.

I do understand the counterargument here (see: defense in depth), and I think it's a good one. Just playing devils advocate. Completely comfortable with @Roasbeef's decision either way here

jakesyl:
As an aside, if we do merge this we'll need to do a force push to get the tests to pass

jcvernaleo:
I remain inclined to just close this since I'm not particularly in favor of it.

ledzep2:
With all due respect, I think it doesn't hurt to provide a little more flexibility for devs.  It's pretty common for network applications to have a similar option. And it doesn't introduce any overhead. On a higher level, this is not intended to change the recommendations or best practices on how to run btcd. This is to provide more tools for devs to work with, the same as more metrics and debug messages.

jakesyl:
I'm with @ledzep2 on this one. Since we don't have consensus/@jcvernaleo is opposed as it stands, this isn't getting merged. Would like to get more feedback from @Roasbeef/anyone else though. I think decreasing friction to get started is a huge plus for a project like this

jakesyl:
@jcvernaleo, just an idea here but I'm wondering about a middle zone after reading #1474. We _could_ actually add a helper command to generate the tls cert using golang.

jcvernaleo:
@jakesyl I like the idea of generating self-signed certs with some helped better than allowing disable.

I think there was a command at one point but it probably needs some love.
```
github.com/btcsuite/btcd/cmd/gencerts
```

jakesyl:
I'll take a look at this. @ledzep2 does this seem like a good solution?

ledzep2:
I have no problem with a helper generating self-signed certs. However, I'm not sure how this will make sharing self-signed certs with test clients easier.

allocz (commented):
As we don't have achieved consensus to approve the feature, I think we may close this PR.

PR #1243, opened 2018-08-09
blockchain: in initChainState mark ancestors of best tip as valid if …

Roasbeef:
…not marked

In this commit, we add an additional consistency check within the
`initChainState` method.  It has been observed that at times, a block
wil lbe written to disk (as it's valid), but then the block index isn't
updated to reflect this. This can cause btcd to fail to do things like
serve cfheaders for valid blocks.

To partially remedy this, when we're loading in the index, we assume
that all ancestors of the current chain tip are valid, and mark them as
such. At the very end, we'll flush the index to ensure the state is
fully consistent. Typically this will be a noop, as only dirty elements
are flushed.

### DIFF
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -1158,6 +1158,14 @@ func (b *BlockChain) connectBestChain(node *blockNode, block *btcutil.Block, fla
 			return false, err
 		}
 
+		// If this is fast add, or this block node isn't yet marked as
+		// valid, then we'll update its status and flush the state to
+		// disk again.
+		if fastAdd || !b.index.NodeStatus(node).KnownValid() {
+			b.index.SetStatusFlags(node, statusValid)
+			flushIndexState()
+		}
+
 		return true, nil
 	}
 	if fastAdd {
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -1085,6 +1085,17 @@ func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List) error 
 func (b *BlockChain) connectBestChain(node *blockNode, block *btcutil.Block, flags BehaviorFlags) (bool, error) {
 	fastAdd := flags&BFFastAdd == BFFastAdd
 
+	flushIndexState := func() {
+		// Intentionally ignore errors writing updated node status to DB. If
+		// it fails to write, it's not the end of the world. If the block is
+		// valid, we flush in connectBlock and if the block is invalid, the
+		// worst that can happen is we revalidate the block after a restart.
+		if writeErr := b.index.flushToDB(); writeErr != nil {
+			log.Warnf("Error flushing block index changes to disk: %v",
+				writeErr)
+		}
+	}
+
 	// We are extending the main (best) chain with a new block.  This is the
 	// most common case.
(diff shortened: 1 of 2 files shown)


blockchain/chainio.go
```diff
@@ -1221,6 +1221,25 @@ func (b *BlockChain) initChainState() error {
 			return err
 		}
 
+		// As a final consistency check, we'll run through all the
+		// nodes which are ancestors of the current chain tip, and mark
```
cfromknecht: Is it worth checking the opposite for orphaned blocks?

blockchain/chainio.go
```diff
@@ -1221,6 +1221,25 @@ func (b *BlockChain) initChainState() error {
...
+		// them as valid if they aren't already marked as such.  This
+		// is a safe assumption as all the block before the current tip
+		// are valid by definition.
+		for iterNode := tip; iterNode != nil; iterNode = iterNode.parent {
```
halseth: Seems expensive to go through the whole chain every time? Not sure if it matters, as it seems like we are reading the whole chain from disk above anyway.

blockchain/chain.go
```diff
@@ -1140,9 +1144,30 @@ func (b *BlockChain) connectBestChain(node *blockNode, block *btcutil.Block, fla
...
+			flushIndexState()
+
 			return false, err
 		}
 
+		// If this is fast add, or this block node isn't yet marked as
+		// valid, then we'll update its status and flush the state to
+		// disk again.
+		if fastAdd || !b.index.NodeStatus(node).KnownValid() {
+
```
davecgh: Extra blank line

blockchain/chain.go
```diff
@@ -1140,9 +1144,30 @@ func (b *BlockChain) connectBestChain(node *blockNode, block *btcutil.Block, fla
...
+			b.index.SetStatusFlags(node, statusValid)
+
```
davecgh: Extra blank line

blockchain/chainio.go
davecgh: Orphaned blocks may have never been validated and, as such, there are no guarantees as to their expected validity.

blockchain/chainio.go
Roasbeef: It's just in memory, so it should be pretty quick. This happens on start up only, as we're already reading the entire chain into memory.

blockchain/chain.go
Roasbeef: Fixed.

blockchain/chain.go
Roasbeef: Fixed.

blockchain/chain.go
```diff
@@ -1140,9 +1144,28 @@ func (b *BlockChain) connectBestChain(node *blockNode, block *btcutil.Block, fla
 		// Connect the block to the main chain.
 		err := b.connectBlock(node, block, view, stxos)
 		if err != nil {
+			// If we got hit with a rule error, then we'll mark
+			// that status of the block as invalid and flush the
+			// index state to disk before returning with the error.
+			if _, ok := err.(RuleError); ok {
```
someone235: Can `b.connectBlock` actually return a `RuleError`, or is it just a safety check?

PR #1248, opened 2018-08-11
peer: Add duplicate version message test.

davecgh:
**This relies on #1247**

Backported from decred/dcrd#1248.

This adds a test to ensure duplicate version messages are rejected.

PR #1249, opened 2018-08-11
peer: Rework version negotiation.

davecgh:
**This requires PR #1248**.

Backported from decred/dcrd#1250.

This modifies the negotiation logic to ensure the callback has the opportunity to see the message before the peer is disconnected and improves the error handling when reading the remote version message.

It also has the side effect of ensuring the protocol version is negotiated before sending reject messages with the exception of the first message not being a version message since negotiation is not possible in that case.

This is being changed because it is useful for the server to see the message regardless in order to have the opportunity to things such as update the address manager and reject peers that don't have desired services.


davecgh:
Rebased.

PR #1250, opened 2018-08-11
peer: Allow OnVersion callback to reject peer.

davecgh:
**This requires PR #1249**.

Backported from decred/dcrd#1251.

This modifies the `OnVersion` callback to allow a reject message to be returned in which case the message will be sent to the peer and the peer will be disconnected.


davecgh:
Rebased.

PR #1251, opened 2018-08-11
server: Reject outbound conns to non-full nodes.

davecgh:
**This requires PR #1250**.

Backported from decred/dcrd#1252.

This modifies the server connection code to reject outbound peers that do not offer full node services.


davecgh:
Rebased.

PR #1252, opened 2018-08-11
peer: Improve net address service adverts.

davecgh:
**This PR requires #1251**.

Backported from decred/dcrd#1253.

This modifies the peer code which deals with advertising service flags via the net address fields of the version message as follows:

- For outgoing connections:
  - Set the local netaddress services to what the local peer supports
  - Set the remote netaddress services to 0 to indicate no services as they are still unknown
- For incoming connections:
  - Set the local netaddress services to what the local peer supports
  - Set the remote netaddress services to the what was advertised by the remote peer in its version message


davecgh:
Rebased.

PR #1253, opened 2018-08-11
multi: Update addrmgr services on outbound conns.

davecgh:
**This is rebased on #1252**.

Backported from decred/dcrd#1254.

This consists of two commits.

The first commit exposes a new method named `SetServices` to the address manager which can be used to update the services for a known address.

The second updates server to add code which updates the address manager services for a known     address to the services advertised by peers when they are connected to via an outbound connection.  It is only done for outbound connections to help prevent malicious behavior from inbound connections.


davecgh:
Rebased.

PR #1254, opened 2018-08-11
server: Use local inbound var in version handler.

davecgh:
**This requires PR #1253**.

Backported from decred/dcrd#1255.

This modifies the `OnVersion` handler for server peers to use a local variable for the inbound status of the peer in order to avoid grabbing the mutex multiple times.

While here, it also does some light cleanup.  There are no functional changes.


davecgh:
Rebased.

PR #1255, opened 2018-08-11
server: Only advertise local addr when current.

davecgh:
**This requires PR #1254**.

Backported from decred/dcrd#1256.

This changes the server peers `OnVersion` handler to only advertise the server as a viable target for inbound connections when the server believes it is close the best known tip.


davecgh:
Rebased.

PR #1256, opened 2018-08-11
server: Use local addr var in version handler.

davecgh:
**This requires #1255**.

Backported from decred/dcrd#1258.

This modifies the `OnVersion` handler for server peers to use a local variable for the remote address of the peer in order to avoid grabbing the mutex multiple times.

There are no functional changes.


davecgh:
Rebased.

PR #1259, opened 2018-08-16
btcjson: use int64 for GetBlockHeader[Verbose] confirmations

cfromknecht:
This PR modifies the parsed fields in GetBlockVerbose
and GetBlockHeaderVerbose from a uint64 to an int64.
bitcoind will return -1 if the requested block is not in
the main chain, which causes a panic when trying to
serialize into a uint.


cfromknecht:
I've added a commit to mimic the bitcoind behavior, please see the commit message for rationale

cfromknecht:
I've removed the commit that mimcs the bitcoind behavior from this PR, as the downstream implications probably merit their own discussion beyond the serialization fix.

PR #1272, opened 2018-08-28
server: allow for read concurrency when serving cf checkpoints, fix bug in checkpoint serving 

Roasbeef:
In this commit, we modify the locking scheme when serving cf checkpoints
for peers to allow the server to serve multiple peers at the same time.
Before this commit, we would first grab the write lock, check to see for
expansion, then release the read lock. In this commit we reverse this
and instead will grab the read lock, and upgrade to the write lock if we
need to actually expand the size of the cache.

We also fix a panic bug that can arise when we attempt to
process a cf checkpoint message from a remote peer. Before this commit,
if the size of the checkpoint cache was large than the number of
checkpoints requested by the peer, we would panic with an out of bounds
error. In order to prevent, this we'll now use the size of the requested
set of hashes as our bound to ensure that we don't panic.

### DIFF
--- a/server.go
+++ b/server.go
@@ -1011,7 +1011,7 @@ func (sp *serverPeer) OnGetCFCheckpt(_ *peer.Peer, msg *wire.MsgGetCFCheckpt) {
 	// a re-org has occurred so items in the db are now in the main china
 	// while the cache has been partially invalidated.
 	var forkIdx int
-	for forkIdx = len(checkptCache); forkIdx > 0; forkIdx-- {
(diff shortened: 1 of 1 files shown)


JimmyMow:
Tested ACK 222a6da

Running BTCD to serve neutrino clients and was hitting the panic bug. No issues since I've been running this PR

JimmyMow:
Why was this closed?

wpaulino:
It's been merged as #1277.

PR #1276, opened 2018-09-03
syncmanager: pick the first candidate then break

detailyang:
Hello. 

It looks like we forget to break the loop after picking the first available candidate at [syncmanager#L263](https://github.com/btcsuite/btcd/blob/master/netsync/manager.go#L263).


detailyang:
ping @Roasbeef :)

detailyang:
It looks like pick the "last" available candidate.

PR #1283, opened 2018-09-09
server: Fix access knownAddresses map unsafely

qshuai:
add lock protection


jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Bug
- Outdated

allocz (commented):
Currently, master branch uses `lru.Cache` with access synchronized with a mutex, so I think we can close this PR.

PR #1289, opened 2018-09-13
Expose a transaction's weight via RPC

preminem:
This commit is for consistency with the latest version of bitcoin core rpc.


obarat:
any tests?

PR #1290, opened 2018-09-13
mempool,txscript: avoid big array copies in for-range

quasilyte:
When array is used in place of range expression,
whole array is copied. Using &array avoids this issue.

Found using https://go-critic.github.io/overview#rangeExprCopy-ref

Linter output:
```
$GOPATH/src/github.com/btcsuite/btcd/mempool/estimatefee.go:513:2: rangeExprCopy: copy of ef.bin (600 bytes) can be avoided with &ef.bin
$GOPATH/src/github.com/btcsuite/btcd/mempool/estimatefee.go:522:2: rangeExprCopy: copy of ef.bin (600 bytes) can be avoided with &ef.bin
$GOPATH/src/github.com/btcsuite/btcd/mempool/estimatefee.go:659:2: rangeExprCopy: copy of ef.bin (600 bytes) can be avoided with &ef.bin
$GOPATH/src/github.com/btcsuite/btcd/txscript/opcode.go:2440:2: rangeExprCopy: copy of opcodeArray (10240 bytes) can be avoided with &opcodeArray
```

### DIFF
--- a/mempool/estimatefee.go
+++ b/mempool/estimatefee.go
@@ -510,7 +510,7 @@ func (ef *FeeEstimator) newEstimateFeeSet() *estimateFeeSet {
 	set := &estimateFeeSet{}
 
 	capacity := 0
-	for i, b := range ef.bin {
+	for i, b := range &ef.bin {
 		l := len(b)
 		set.bin[i] = uint32(l)
 		capacity += l
@@ -519,7 +519,7 @@ func (ef *FeeEstimator) newEstimateFeeSet() *estimateFeeSet {
 	set.feeRate = make([]SatoshiPerByte, capacity)
 
 	i := 0
-	for _, b := range ef.bin {
+	for _, b := range &ef.bin {
 		for _, o := range b {
 			set.feeRate[i] = o.feeRate
 			i++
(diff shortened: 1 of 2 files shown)


jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Enhancement

jakesyl:
Can you rebase or force-push so CI can run?

allocz (commented):
Would be nice to have benchmarks here. So I wrote it just to test if this makes sense, turns out that without the patch, the performance is better.

The benchmark:
```
var EfsSink estimateFeeSet

func BenchmarkNewEstimateFeeSet(b *testing.B) {
        b.ReportAllocs()

        ef := newTestFeeEstimator(128, 3, 1)

        var efs estimateFeeSet
        for b.Loop() {
                efsTmp := ef.newEstimateFeeSet()
                efs = *efsTmp
        }

        EfsSink = efs
}
```

The comparison before and after:
```
goos: linux
goarch: amd64
pkg: github.com/btcsuite/btcd/mempool
                    │   before    │               after                │
NewEstimateFeeSet-8   177.7n ± 1%   186.3n ± 1%  +4.84% (p=0.000 n=11)

                    │   before   │             after              │
                    │    B/op    │    B/op     vs base            │
NewEstimateFeeSet-8   128.0 ± 0%   128.0 ± 0%  ~ (p=1.000 n=11) ¹
¹ all samples are equal

                    │   before   │             after              │
                    │ allocs/op  │ allocs/op   vs base            │
NewEstimateFeeSet-8   1.000 ± 0%   1.000 ± 0%  ~ (p=1.000 n=11) ¹
¹ all samples are equal
```

Thanks for the PR, but I think you can close it.

PR #1292, opened 2018-09-13
cmd/btcctl: use regexp.MustCompile for constant patterns

quasilyte:
Found using https://go-critic.github.io/overview#regexpMust-ref


jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Enhancement

PR #1297, opened 2018-09-13
simplify `x = x <op> ...` to `x <op>= ...`

ludweeg:
Found using https://go-critic.github.io/overview#assignOp-ref


Quasilyte:
Apart from gofmt errors, I think commit message also suffered from some kind of mistake.

jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Formatting

allocz (commented):
The patch to `field.go` was changing code that does not even exists anymore. Also, I don't think the code becomes more readable after the changes, so I would like to see this PR being closed.

PR #1303, opened 2018-09-22
Adding a method to send a pre-serialized raw transaction

romainPellerin:
- this is useful when we want to process a transaction that has been previously signed by another party without requiring the private key
- this can be used nicely on a node configured with txindex=1, so there's no need to import the other party wallet

### DIFF
--- a/btcjson/chainsvrresults.go
+++ b/btcjson/chainsvrresults.go
@@ -4,7 +4,9 @@
 
 package btcjson
 
-import "encoding/json"
+import (
+	"encoding/json"
+)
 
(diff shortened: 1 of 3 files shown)


jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Enhancement
- Outdated (tests failing)

allocz (commented):
Hey @romainPellerin, do you still want to get this merged? Would be nice to resolve this, as we have a lot of open pull requests with no activity for a considerable amount of time.

PR #1306, opened 2018-09-24
Don't panic if assertion failed

renlulu:


jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Bug

Merged in gcash/bchd#93 so probably safe

PR #1313, opened 2018-09-28
Fix test case bytes byteorder to little endian

imnisen:
I think this should use "little endian" format of integer 10


jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Bug

wire/msgfilterload_test.go
```diff
@@ -136,7 +136,7 @@ func TestFilterLoadWireErrors(t *testing.T) {
 	baseFilterLoad := NewMsgFilterLoad(baseFilter, 10, 0, BloomUpdateNone)
 	baseFilterLoadEncoded := append([]byte{0x04}, baseFilter...)
 	baseFilterLoadEncoded = append(baseFilterLoadEncoded,
-		0x00, 0x00, 0x00, 0x0a, // HashFuncs
```
allocz, line 139: Would be nice to see a new entry in the test vectors that fails when the suggested change is reversed.

allocz (changes requested):

PR #1321, opened 2018-10-10
fix taking the address of loop iter var problem

unixisevil:


jakesyl:
@jcvernaleo (as per #1530)

- Low priority
- Bug
- Testing

allocz (commented):
Go v1.22 already fixed the semantic bug (https://go.dev/blog/loopvar-preview), so I believe we should close this PR.

PR #1343, opened 2018-11-05
Bug Fix: Uptime was returning unix time now instead of number of seconds up

swdee:
The RPC `uptime` command was returning unix time now() instead of the number of seconds the server had been running.  This was due to a bug on the initialisation order of the startupTime variable occurring after the RPC server instance had been created by function newRPCServer().  Subsequently an uninitialised (zero) value was being passed to the RPC config https://github.com/btcsuite/btcd/blob/67e573d211ace594f1366b4ce9d39726c4b19bd0/server.go#L2835 resulting in the uptime calculation being `now() - 0`.

### DIFF
--- a/server.go
+++ b/server.go
@@ -2282,9 +2282,6 @@ func (s *server) Start() {
 
 	srvrLog.Trace("Starting server")
 
-	// Server startup time. Used for the uptime command for uptime calculation.
-	s.startupTime = time.Now().Unix()
-
 	// Start the peer handler which in turn starts the address and block
 	// managers.
 	s.wg.Add(1)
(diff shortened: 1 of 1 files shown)


jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Bug

allocz (commented):
As of now [server.go](https://github.com/btcsuite/btcd/blob/b48125d0a3565b1441522ee10422f7815db03216/server.go#L2710) `Start()` sets the internal rpc server startup time before starting it, so this bug should not exist anymore. Thanks for the PR but I think it can be closed now.

PR #1348, opened 2018-11-25
Follow RFC6979 guidance on bad R and S values

junderw:
Step H in the nonce function needs to loop if R or S is inappropriate for use in the signature.
This callback system allows for that.

### DIFF
--- a/btcec/signature.go
+++ b/btcec/signature.go
@@ -424,33 +424,47 @@ func signRFC6979(privateKey *PrivateKey, hash []byte) (*Signature, error) {
 	privkey := privateKey.ToECDSA()
 	N := S256().N
 	halfOrder := S256().halfOrder
-	k := nonceRFC6979(privkey.D, hash)
-	inv := new(big.Int).ModInverse(k, N)
-	r, _ := privkey.Curve.ScalarBaseMult(k.Bytes())
-	r.Mod(r, N)
-
-	if r.Sign() == 0 {
-		return nil, errors.New("calculated R is zero")
-	}
 
 	e := hashToInt(hash, privkey.Curve)
-	s := new(big.Int).Mul(privkey.D, r)
-	s.Add(s, e)
-	s.Mul(s, inv)
-	s.Mod(s, N)
 
-	if s.Cmp(halfOrder) == 1 {
-		s.Sub(N, s)
+	var r, s *big.Int
+
+	// using a callback to check (and store) r and s.
+	// When returning true, the loop in step H will stop.
+	callback := func (k *big.Int) bool {
+		r, _ = privkey.Curve.ScalarBaseMult(k.Bytes())
+		r.Mod(r, N)
+
+		if r.Sign() == 0 {
+			return false
+		}
+
+		invK := new(big.Int).ModInverse(k, N)
+		s = new(big.Int).Mul(privkey.D, r)
+		s.Add(s, e)
+		s.Mul(s, invK)
+		s.Mod(s, N)
+
+		if s.Cmp(halfOrder) == 1 {
+			s.Sub(N, s)
+		}
+
+		if s.Sign() == 0 {
+			return false
+		}
+
+		return true
 	}
-	if s.Sign() == 0 {
-		return nil, errors.New("calculated S is zero")
-	}
+	nonceRFC6979(privkey.D, hash, callback)
 	return &Signature{R: r, S: s}, nil
 }
 
 // nonceRFC6979 generates an ECDSA nonce (`k`) deterministically according to RFC 6979.
-// It takes a 32-byte hash as an input and returns 32-byte nonce to be used in ECDSA algorithm.
-func nonceRFC6979(privkey *big.Int, hash []byte) *big.Int {
+// It takes a private key, a 32-byte hash, and a callback func as an input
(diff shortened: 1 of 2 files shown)


junderw:
maybe instead of adding a callback arg, add an int arg x that will tell the step H loop to loop x times where it would have returned normally.

Basically if a bad R or S happened, you'd run nonceRFC6979 once, then run it a second time with 1 to tell it to loop once. So it does the pre-Step-H and initial loop twice, then the second loop once. since it's such a rare occurrence, the performance hit of having to loop an extra cycle is basically 0.
But with the upside of also making it easier to test

we could use the test cases from coinbin that I made to test badRS situations.

https://github.com/OutCast3k/coinbin/blob/739aee3c0ff76b6dbc328ad96b2578df481d6701/js/coin.js#L1377-L1437

Tests are here:

https://github.com/OutCast3k/coinbin/blob/739aee3c0ff76b6dbc328ad96b2578df481d6701/js/coin.js#L463-L575

bad00 is just normal k derivation
bad15 is passing a badrs arg of 15 so the step H will loop 15 times where it would have returned prior.

junderw:
Anyone with any preferences on which method to take?

Current state of this PR is callback based... looking back on it, it's kind of ugly and hard to read.

btcec/signature.go
```diff
@@ -424,33 +424,47 @@ func signRFC6979(privateKey *PrivateKey, hash []byte) (*Signature, error) {
...
-	if s.Cmp(halfOrder) == 1 {
-		s.Sub(N, s)
-	}
-	if s.Sign() == 0 {
-		return nil, errors.New("calculated S is zero")
+	var r, s *big.Int
+
+	// using a callback to check (and store) r and s.
+	// When returning true, the loop in step H will stop.
+	callback := func (k *big.Int) bool {
```
GuiltyMorishita: Is there any reason to implement it as a callback, not as a normal function?

btcec/signature.go
```diff
@@ -493,8 +507,8 @@ func nonceRFC6979(privkey *big.Int, hash []byte) *big.Int {
 
 		// Step H3
 		secret := hashToInt(t, curve)
-		if secret.Cmp(one) >= 0 && secret.Cmp(q) < 0 {
-			return secret
+		if secret.Cmp(one) >= 0 && secret.Cmp(q) < 0 && checkNonce(secret) {
```
GuiltyMorishita: I think you only need to implement the following function.

```
func checkNonce(k *big.Int) bool {
	// ...
	return true
}
```

btcec/signature_test.go
```diff
@@ -646,7 +646,14 @@ func TestRFC6979(t *testing.T) {
 		hash := sha256.Sum256([]byte(test.msg))
 
 		// Ensure deterministically generated nonce is the expected value.
-		gotNonce := nonceRFC6979(privKey.D, hash[:]).Bytes()
+		var gotNonce []byte
+		callback := func (k *big.Int) bool {
+			// not checking for r and s values
+			// just grab the nonce and return true to stop the internal loop
+			gotNonce = k.Bytes()
```
GuiltyMorishita: If you want to get nonce, I think that nonceRFC6979 should return nonce.

btcec/signature_test.go
```diff
@@ -646,7 +646,14 @@ func TestRFC6979(t *testing.T) {
...
+			return true
+		}
+		nonceRFC6979(privKey.D, hash[:], callback)
```
GuiltyMorishita: I do not know from this function name what nonceRFC6979 is a function to do.

junderw:
@Roasbeef @cfromknecht Any thoughts?

Technically the current implementation doesn't match RFC6979 exactly.

If using a counter instead of a callback is more desirable (imo it's less readable) I have a pull request on my repo here which is an alternative.

https://github.com/junderw/btcd/pull/1/files?utf8=%E2%9C%93&diff=unified&w=1

jakesyl:
@jcvernaleo (as per #1530)

- Low priority
- Enhancement

btcec/signature.go
```diff
@@ -424,33 +424,47 @@ func signRFC6979(privateKey *PrivateKey, hash []byte) (*Signature, error) {
...
 }
 
 // nonceRFC6979 generates an ECDSA nonce (`k`) deterministically according to RFC 6979.
-// It takes a 32-byte hash as an input and returns 32-byte nonce to be used in ECDSA algorithm.
-func nonceRFC6979(privkey *big.Int, hash []byte) *big.Int {
+// It takes a private key, a 32-byte hash, and a callback func as an input
+// and returns nothing. The 32-byte nonce to be used in ECDSA algorithm is passed
+// to the callback func and the nonce should be checked, r and s gathered, and
+// return true to indicate to this func to end the final loop in Step H.
+func nonceRFC6979(privkey *big.Int, hash []byte, checkNonce func (*big.Int) bool) {
```
allocz, line 467: The current file and code is not part of btcd anymore.

allocz (commented):
The current dependency that provides ECDSA signining calls the nonce generation again on failure, as you can see [here](https://github.com/decred/dcrd/blob/b6c67575ef22cf9e2d0c8600c4f58db14c35fc50/dcrec/secp256k1/ecdsa/signature.go#L715).

Thanks for the PR, but I think that it can be closed.

PR #1352, opened 2018-11-29
integration/rpctest: make exec path compatible with modules

halseth:
With the use of go modules, the btcdPkgPath will no longer be `github.com/btcsuite/btcd`, but end with a version string (e.g. btcd@v0.0.0-20181129140220-beb77e89572a).

This trips up build.ImportDir, which will return a ImportPath of ".", since this package cannot be found in the go path.

There are probably several ways of fixing this, by clever string magic, but seems easiest to just hardcode the btcd package name.

Relevant issue: https://github.com/golang/go/issues/26504

### DIFF
--- a/integration/rpctest/btcd.go
+++ b/integration/rpctest/btcd.go
@@ -6,7 +6,6 @@ package rpctest
 
 import (
 	"fmt"
-	"go/build"
 	"os/exec"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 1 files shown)


halseth:
an alternative is to use the `btcd` executable found in path, which I see `dcrd` is doing: https://github.com/decred/dcrd/blob/bd3a416cae89d63342479aae5fc4e1c9aeb46fb9/rpctest/node.go#L55

Roasbeef:
It seems we should just go ahead and switch `btcd` over to using modules as well along the way?

PR #1359, opened 2018-12-04
txscript+rpcwebsocket: detect script spend within rescans

wpaulino:
In this PR, we extend the `txscript` package to support re-deriving the PkScript of an output by looking at the input's signature script/witness attempting to spend it. So far, the only supported types are P2SH, v0 P2WSH, and v0 P2WPKH. This will serve useful to detect when a particular script has been spent on-chain. A set of test vectors has also been added for the supported script types to ensure its correctness.

We also modify the `rescanBlock` method to also determine whether a script (encoded as an address in its rescanKeys) has been spent. Upon detecting a spend, a btcjson.RedeemingTxNtfn is sent to the client who requested it.


txscript/pkscript.go
```diff
@@ -0,0 +1,194 @@
...
+)
+
+var (
+	// ErrUnsupportedScriptType ...
+	ErrUnsupportedScriptType = errors.New("unsupported script type")
+)
+
+// PkScript is a wrapper struct around a byte array, allowing it to be used
+// as a map index.
+type PkScript struct {
```
Roasbeef: We _could_ also extend this to cover a set of arbitrary scripts, but given that for our particular use case, we only care about segwit style scripts, I think we can defer that. However, since this only covers `v0` witness programs, I think we should update the name to reflect that into `txscript.WitnessProgram`.

txscript/pkscript.go
```diff
@@ -0,0 +1,194 @@
...
+	// array.
+	class ScriptClass
+
+	// script is the script contained within a byte array. If the script is
+	// smaller than the length of the byte array, it will be padded with 0s
+	// at the end.
+	script [maxLen]byte
+}
+
+// ParsePkScript parses an output script into the PkScript struct.
```
Roasbeef: Based on comment above, would be `txscript.ParseWitnessProgram`.

txscript/pkscript.go
Roasbeef: Would say the P2SH stuff falls into this bucket as well, since it's really just a witness program with a layer of indirection.

txscript/pkscript.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+}
+
+// Class returns the script type.
+func (s PkScript) Class() ScriptClass {
+	return s.class
+}
+
+// Script returns the script as a byte slice without any padding.
+func (s PkScript) Script() []byte {
```
Roasbeef: Perhaps we should make a copy instead, so the callers can't accidentally mutate our underlying script storage.

txscript/pkscript.go
```diff
@@ -0,0 +1,194 @@
...
+		return s.script[:witnessV0ScriptHashLen]
+	default:
+		// Unsupported script type.
+		return nil
+	}
+}
+
+// Address encodes the script into an address for the given chain.
+func (s PkScript) Address(chainParams *chaincfg.Params) (btcutil.Address, error) {
+	_, addrs, _, err := ExtractPkScriptAddrs(s.Script(), chainParams)
```
Roasbeef: We can memoize this function based on the `NewWitnessProgram` method since that's the only way that external callers can make a new instance of this struct. So we'd just cache a ref to the address, and re-use that with each call.

txscript/pkscript.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse address: %v", err)
+	}
+
+	return addrs[0], nil
+}
+
+// String returns a hex-encoded string representation of the script.
+func (s PkScript) String() string {
+	return fmt.Sprintf("%x", s.Script())
```
Roasbeef: If we use [`txscript.DisasmString`](https://godoc.org/github.com/btcsuite/btcd/txscript#DisasmString) then we can obtain output that's a bit prettier for the witness program. We'd see the `OP_0` and all, etc.

txscript/pkscript.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+// ComputePkScript computes the pkScript of an transaction output by looking at
+// the transaction input's signature script or witness.
+//
+// NOTE: Only P2SH, P2WSH, and P2WPKH redeem scripts are supported.
+func ComputePkScript(sigScript []byte, witness wire.TxWitness) (PkScript, error) {
```
Roasbeef: Naming change would also propagate here.

txscript/pkscript.go
```diff
@@ -0,0 +1,194 @@
...
+
+// String returns a hex-encoded string representation of the script.
+func (s PkScript) String() string {
+	return fmt.Sprintf("%x", s.Script())
+}
+
+// ComputePkScript computes the pkScript of an transaction output by looking at
+// the transaction input's signature script or witness.
+//
+// NOTE: Only P2SH, P2WSH, and P2WPKH redeem scripts are supported.
```
Roasbeef: Should read nested p2sh?

txscript/pkscript.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/btcsuite/btcutil"
+	"golang.org/x/crypto/ripemd160"
+)
+
+const (
+	// pubKeyHashSigScriptLen is the length of a signature script attempting
+	// to spend a P2PKH script. This length is determined by the following:
+	//   0x48 (72 byte data push) | <71 byte sig> | <1 byte sighash flag> |
+	//   0x21 (33 byte data push) | <33 byte compressed pubkey>
+	pubKeyHashSigScriptLen = 107
```
Roasbeef: The signature can either be 71 or 72 bytes, as a result this needs to be a bit fuzzier. Alternatively (in the check below) we can attempt to parse out the final stack items to determine if that's a public key, possibly using pubkey recovery to double check our guess.

txscript/pkscript.go
```diff
@@ -0,0 +1,194 @@
...
+// the transaction input's signature script or witness.
+//
+// NOTE: Only P2SH, P2WSH, and P2WPKH redeem scripts are supported.
+func ComputePkScript(sigScript []byte, witness wire.TxWitness) (PkScript, error) {
+	var pkScript PkScript
+
+	// If a signature script is provided, since we only support P2SH for
+	// non-segwit scripts, then we'll assume this is a P2SH redeem script
+	// based on its length not being that of a P2PKH redeem script.
+	if len(sigScript) > 0 && len(sigScript) != pubKeyHashSigScriptLen {
```
Roasbeef: We can make this a bit stricter, as we _know_ that the sigScript for a witness program _must_ be a canonical push of the witness program and nothing more. We can borrow some logic from this section of the main VM code: https://github.com/btcsuite/btcd/blob/master/txscript/engine.go#L965

txscript/pkscript.go
```diff
@@ -0,0 +1,194 @@
...
+		pubKeyHash := hash160(lastWitnessItem)
+		script, err := payToWitnessPubKeyHashScript(pubKeyHash)
+		if err != nil {
+			return pkScript, err
+		}
+
+		pkScript.class = WitnessV0PubKeyHashTy
+		copy(pkScript.script[:], script)
+
+	// For any other witnesses, we'll assume it's a P2WSH witness.
```
Roasbeef: Rather than just assuming this, we should SHA2 the last item and check that it produces the same hash as the raw witness program. If it doesn't , then we return an error as it's unknown.

rpcwebsocket.go
```diff
@@ -1978,12 +1977,8 @@ func deserializeOutpoints(serializedOuts []btcjson.OutPoint) ([]*wire.OutPoint,
 }
 
 type rescanKeys struct {
-	fallbacks           map[string]struct{}
-	pubKeyHashes        map[[ripemd160.Size]byte]struct{}
-	scriptHashes        map[[ripemd160.Size]byte]struct{}
-	compressedPubKeys   map[[33]byte]struct{}
-	uncompressedPubKeys map[[65]byte]struct{}
-	unspent             map[wire.OutPoint]struct{}
+	addrs   map[string]struct{}
```
Roasbeef: 👍

txscript/pkscript.go
wpaulino: If we do this, then we'll no longer be able to use this struct as as index in a map, which was the reason it was designed this way.

txscript/pkscript.go
```diff
@@ -0,0 +1,194 @@
...
+
+	// witnessV0ScriptHashLen is the length of a P2WSH script.
+	witnessV0ScriptHashLen = 34
+
+	// maxLen is the maximum script length supported by ParsePkScript.
+	maxLen = witnessV0ScriptHashLen
+)
+
+var (
+	// ErrUnsupportedScriptType ...
```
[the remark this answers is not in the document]
wpaulino: Fixed.

txscript/pkscript.go
wpaulino: Decided to keep as is after our discussion offline. We'll now also support P2PKH scripts.

txscript/pkscript.go
wpaulino: Fixed.

txscript/pkscript.go
wpaulino: Fixed.

txscript/pkscript.go
wpaulino: Fixed. Ended up making the check a bit more lossy and we'll now also attempt to parse the compressed public key found within it. If parsing it fails, we'll fall back to assuming it is a P2SH redeem script.

txscript/pkscript.go
wpaulino: If we intend to support regular P2SH, then we can't make this assumption as the redeem script can be arbitrary.

txscript/pkscript.go
wpaulino: Doesn't seem like this can be done without having the original output's script, which is what we're attempting to compute.

txscript/pkscript.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+
+		// If it isn't, we'll assume it is the P2SH redeem script.
+		fallthrough
+
+	// If we failed to parse a compressed public key from the script in the
+	// case above, or if the script is smaller in length than that of a
+	// P2PKH one, we'll assume it's P2SH.
+	case len(sigScript) > 0:
+		scriptHash := hash160(sigScript)
```
Roasbeef: This is incorrect. Instead we need to take _only_ the first data push in the script and hash that. Otherwise, if this is vanilla p2sh, then we'll hash the redeem script _and_ any witness data.

Roasbeef: So we need to parse the data pushes in order to do this properly.

txscript/pkscript_test.go
```diff
@@ -0,0 +1,390 @@
...
+				0x90, 0x1c, 0x86, 0x94, 0xc0, 0x3f, 0xaf, 0xd5,
+				0x52, 0x28, 0x10, 0xe0, 0x33, 0x0f, 0x26, 0xe6,
+				0x7a, 0x85, 0x33, 0xcd,
+				// OP_EQUAL
+				0x87,
+			},
+		},
+		// Valid P2SH.
+		{
+			sigScript: []byte{0x6b, 0x65, 0x6b}, // kek
```
Roasbeef: This isn't a valid sigScript for a P2SH output. It _must_ be canonical data push of the redeem script, and any other input script data. If this is updated to be an actual P2SH sigScript (can look on smartbit for examples), this test will fail.

txscript/pkscript_test.go
```diff
@@ -0,0 +1,390 @@
...
+			class:    WitnessV0PubKeyHashTy,
+			pkScript: nil,
+		},
+	}
+
+	for i, test := range tests {
+		valid := test.pkScript != nil
+		pkScript, err := ComputePkScript(test.sigScript, test.witness)
+		if err != nil && valid {
+			t.Fatalf("test #%d: unable to compute pkScript: %v", i, err)
```
cfromknecht: nit: use `t.Run()` and give each test case a name, that way we avoid having to prefix `"test #%d: "` to every failure message

txscript/pkscript.go
```diff
@@ -0,0 +1,245 @@
...
+	if len(sigScript) == 0 && len(witness) == 0 {
+		return pkScript, ErrUnsupportedScriptType
+	}
+
+	// We'll start by checking the input's signature script, if provided.
+	switch {
+	// If a signature script is provided with a length long enough to
+	// represent a P2PKH script, then we'll attempt to parse the compressed
+	// public key from it.
+	case len(sigScript) >= pubKeyHashSigScriptLen:
```
cfromknecht: constrain this to 106 or 107?

Roasbeef: 👍

wpaulino: Fixed.

txscript/pkscript_test.go
wpaulino: Nice, fixed.

txscript/pkscript_test.go
wpaulino: Fixed.

txscript/pkscript.go
wpaulino: Fixed.

txscript/pkscript.go
```diff
@@ -0,0 +1,260 @@
...
+	// a P2SH signature script.
+	case len(sigScript) > 0 && IsPushOnlyScript(sigScript):
+		// The redeem script will always be the last data push of the
+		// signature script, so we'll parse the script into opcodes to
+		// obtain it.
+		parsedOpcodes, err := parseScript(sigScript)
+		if err != nil {
+			return pkScript, err
+		}
+		redeemScript := parsedOpcodes[len(parsedOpcodes)-1].data
```
Roasbeef: 👍

PR #1361, opened 2018-12-08
Fix block boundary checking in FetchBlockRegion(s)

h4x3rotab:
In ffldb, a serialized block has some metadata attached. The corresponding `blockLocation.Len` is used to check if the fetching region is out of bunds. However it counts not only the block size but also the 12 bytes metadata, which is not considered accessible by the caller:

`[4 bytes: network version] [4 bytes: block size] [block] [4 bytes: checksum]`

This PR fixes the issue.

### DIFF
--- a/database/ffldb/interface_test.go
+++ b/database/ffldb/interface_test.go
@@ -1360,6 +1360,19 @@ func testFetchBlockIO(tc *testContext, tx database.Tx) bool {
 		if !checkDbError(tc.t, testName, err, wantErrCode) {
 			return false
 		}
+
+		// Ensure fetching a block region larger than the block returns
+		// the expected error.
+		testName = fmt.Sprintf("FetchBlockRegion(%s) out of block size",
+			blockHash)
+		wantErrCode = database.ErrBlockRegionInvalid
+		region.Hash = blockHash
+		region.Offset = 0
+		region.Len = uint32(len(blockBytes) + 1)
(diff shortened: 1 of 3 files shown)


jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Bug

database/ffldb/interface_test.go
```diff
@@ -1499,6 +1513,20 @@ func testFetchBlockIO(tc *testContext, tx database.Tx) bool {
 	}
 	wantErrCode = database.ErrBlockRegionInvalid
 	_, err = tx.FetchBlockRegions(badBlockRegions)
+	if !checkDbError(tc.t, testName, err, wantErrCode) {
+		return false
+	}
+
+	// Ensure fetching block regions larger than the block returns the
+	// expected error.
+	testName = "FetchBlockRegions out of bunds"
```
allocz, line 1522:
```suggestion
	testName = "FetchBlockRegions out of bounds"
```

allocz (changes requested):
Thanks for the bugfix, in fact reversing the fix causes the test to fail. Just requested the fix of a typo. Tested both the PR in your branch and the PR merged into master. Also performed IBD on mainnet for some minutes and everything looks good.

allocz:
@Roasbeef could you also take a look? Seems to be a real bug in the bounds verification, but I think it would require corruption of the database to be triggered.

PR #1363, opened 2018-12-19
rpcclient: Add  RPC command<generatetoaddress> for legacy bitcoin node

qshuai:
Support `generatetoaddress` RPC command for the legacy bitcoin node.


qshuai:
@Roasbeef PTAL

PR #1364, opened 2018-12-19
rpcclient: Replace the arbitrary chainparam with address string returned

qshuai:
The fixed chainparam `MainNetParams` will cause the error returned if the rpcclient is interacting with bitcoin test node.

### DIFF
--- a/rpcclient/wallet.go
+++ b/rpcclient/wallet.go
@@ -9,7 +9,6 @@ import (
 	"strconv"
 
 	"github.com/btcsuite/btcd/btcjson"
-	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcutil"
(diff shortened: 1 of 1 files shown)


qshuai:
@Roasbeef PTAL

rpcclient/wallet.go
```diff
@@ -1015,7 +1036,7 @@ func (c *Client) GetAccountAddressAsync(account string) FutureGetAccountAddressR
 
 // GetAccountAddress returns the current Bitcoin address for receiving payments
 // to the specified account.
-func (c *Client) GetAccountAddress(account string) (btcutil.Address, error) {
+func (c *Client) GetAccountAddress(account string) (string, error) {
```
Roasbeef: Instead, the chain params should be passed in.

qshuai: Thanks, good idea! Preparing an improving commit.

qshuai:
@Roasbeef Please review the new commit.

jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Enhancement

rpcclient/wallet.go
```diff
@@ -202,7 +201,9 @@ func (c *Client) ListUnspentMinMaxAsync(minConf, maxConf int) FutureListUnspentR
 // function on the returned instance.
 //
 // See ListUnspentMinMaxAddresses for the blocking version and more details.
-func (c *Client) ListUnspentMinMaxAddressesAsync(minConf, maxConf int, addrs []btcutil.Address) FutureListUnspentResult {
```
allocz, line 205: I think you should not format code that is not related with the PR. Because of this, reading the diff takes more time.

rpcclient/wallet.go
```diff
@@ -1041,17 +1032,13 @@ func (r FutureGetAccountAddressResult) Receive() (btcutil.Address, error) {
...
-	}
-	return result
+	return c.sendCmd(cmd)
 }
 
 // GetAccountAddress returns the current Bitcoin address for receiving payments
 // to the specified account.
-func (c *Client) GetAccountAddress(account string) (btcutil.Address, error) {
-	return c.GetAccountAddressAsync(account).Receive()
+func (c *Client) GetAccountAddress(account string, net *chaincfg.Params) (btcutil.Address, error) {
```
allocz, line 1040: Outdated when compared to master. The `chaincfg.Params` are inside `rpcclient.Client`, so this PR is not needed anymore.

allocz (commented):
Thanks for sending the patch, but as the PR is outdated and the improvement was also made in the master branch using another approach, I think that this PR can be closed.

PR #1163, opened 2018-04-16
Created support for watchOnly transaction listing

gertjaap:
I was missing a way of pulling transactions from the RPC including transactions from watchOnly addresses. I added it in the attached commit.

PR #1368, opened 2018-12-28
Extend the ListTransaction api to support queries to "watch only" wallets

zmanian:
The `ListTransactions` api doesn't support any methods where including watch only wallets are set to true. This PR adds an api call with that option.

### DIFF
--- a/rpcclient/wallet.go
+++ b/rpcclient/wallet.go
@@ -141,6 +141,25 @@ func (c *Client) ListTransactionsCountFrom(account string, count, from int) ([]b
 	return c.ListTransactionsCountFromAsync(account, count, from).Receive()
 }
 
+// ListTransactionsCountFromWatchOnlyAsync returns an instance of a type that can be used
+// to get the result of the RPC at some future time by invoking the Receive
+// function on the returned instance.
+//
+// See ListTransactionsCountFrom for the blocking version and more details.
+func (c *Client) ListTransactionsCountFromWatchOnlyAsync(account string, count, from int) FutureListTransactionsResult {
+	watchOnly := true
+	cmd := btcjson.NewListTransactionsCmd(&account, &count, &from, &watchOnly)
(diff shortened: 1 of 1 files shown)


rpcclient/wallet.go
```diff
@@ -141,6 +141,25 @@ func (c *Client) ListTransactionsCountFrom(account string, count, from int) ([]b
...
+// function on the returned instance.
+//
+// See ListTransactionsCountFrom for the blocking version and more details.
+func (c *Client) ListTransactionsCountFromWatchOnlyAsync(account string, count, from int) FutureListTransactionsResult {
+	watchOnly := true
+	cmd := btcjson.NewListTransactionsCmd(&account, &count, &from, &watchOnly)
+	return c.sendCmd(cmd)
+}
+
+// ListTransactionsCountFromWatchOnly returns a list of the most recent transactions up
```
Roasbeef: All columns should be wrapped to 80 characters.

jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Enhancement

rpcclient/wallet.go
```diff
@@ -141,6 +141,25 @@ func (c *Client) ListTransactionsCountFrom(account string, count, from int) ([]b
...
+// to the passed count while skipping the first 'from' transactions including transactions from "watch only" wallets.
+//
+// See the ListTransactions and ListTransactionsCount functions to use defaults.
+func (c *Client) ListTransactionsCountFromWatchOnly(account string, count, from int) ([]btcjson.ListTransactionsResult, error) {
```
Rjected, line 159: I think the signature should include the bool to be consistent with other commands we implement which have optional bools.

Rjected:
also, duplicate of #1505 _and_ #1163 ?

jcvernaleo:
Done with #1628

PR #1371, opened 2019-01-10
Fix issue with AddMultisigAddress

nkostoulas:
unmarshalling wrong object in AddMultisigAddress response

### DIFF
--- a/rpcclient/wallet.go
+++ b/rpcclient/wallet.go
@@ -764,14 +764,14 @@ func (r FutureAddMultisigAddressResult) Receive() (btcutil.Address, error) {
 		return nil, err
 	}
 
-	// Unmarshal result as a string.
-	var addr string
-	err = json.Unmarshal(res, &addr)
+	// Unmarshal result as a addmultisigaddress result object.
+	var addmultisigRes btcjson.AddMultisigAddressResult
+	err = json.Unmarshal(res, &addmultisigRes)
 	if err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Bug?

rpcclient/wallet.go
```diff
@@ -753,28 +753,25 @@ func (c *Client) SendManyComment(fromAccount string,
...
 
 // Receive waits for the response promised by the future and returns the
 // multisignature address that requires the specified number of signatures for
 // the provided addresses.
 func (r FutureAddMultisigAddressResult) Receive() (btcutil.Address, error) {
-	res, err := receiveFuture(r.responseChannel)
+	res, err := receiveFuture(r)
 	if err != nil {
 		return nil, err
 	}
```
allocz, line 769: In fact, this is still present in the [master](https://github.com/btcsuite/btcd/blob/b48125d0a3565b1441522ee10422f7815db03216/rpcclient/wallet.go#L864) branch. The bitcoin core RPC [reference](https://bitcoincore.org/en/doc/0.16.0/rpc/wallet/addmultisigaddress/) shows that bitcoind itself broke the API, as in the past, the same RPC command returned a different result, which was just the address string, as implemented in btcd.

allocz (commented):
The patch is outdated, but the issue is still present on master. The point is that if we update, we may break the client for people using older implementations of the RPC server.

PR #1168, opened 2018-04-21
[WIP] Implement a UTXO cache

stevenroose:
This PR requires https://github.com/btcsuite/btcd/pull/1045. Ignore the commit(s) from https://github.com/btcsuite/btcd/pull/1045 when reviewing this PR.

This PR adds an in-memory UTXO cache. `UtxoViewpoint` is trimmed of the block (dis)connection methods, as they now can be applied also to the cache (for reverting and forwarding blocks). 

The database keeps track of the consistency of the UTXO state with the rest of the chainstate (block index and blocks). When the node starts with the UTXO state lagging behind, it replays blocks to catch up. Since flushing the cache cannot be done in a single LevelDB transaction, it is done in batches. Before a flush starts, the db is updated to indicate a flush is ongoing. When the node shuts down in the middle of a flush and restarts, it does not know where it left of with flushing. Therefore, it first rolls back changes that had already been applied and then replays all blocks to catch up with the chain state.

Yet to be done:
 - [ ] Work out a flushing strategy. Flush inside cache when over limit?
 - [ ] Consider somehow merging `UtxoViewpoint` and `utxoCache` in a `cachedUtxoView`.

PR #1373, opened 2019-01-17
[WIP] Implement a UTXO cache 

Roasbeef:
This PR picks off where #1168 left off. I've back ported fixes from the `gcash` repo, ensuring all commits have their authors properly attributed. I've started to test this PR as is on both mainnet and testnet. We also need to examine the set of fixes and ensure they're absolutely necessary, as many of the back-ported `gcash` fixes didn't have accompanying unit tests that failed before the fix was applied. 

Additionally, I plan to fix what I view as two design flaws in the current UTXO set cache:
  1. When entries are added, we don't evict them in real time. As a result, the cache can grow very large beyond the targeted constrained sized. In the future, we'll likely also implement more advanced eviction logic, such as evicting the oldest present UTXO, as empirically, UTXOs die young. 
  2. When we flush the cache, we also remove all in memory entries from the cache. This causes additional G/C pressure, as the very next block, we then need to re-allocate the map. Additionally, this causes more cache hits, as we start with a cold cache for the next block which guarantees that we'll have a large set of cache misses. Instead, we should only seek to sync our in memory view with that what's on disk, only syncing the entires that have been modified. 

Thanks to everyone that had a part in this including @stevenroose, @cfromknecht and @cpacia.

### DIFF
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -85,6 +85,26 @@ func newBestState(node *blockNode, blockSize, blockWeight, numTxns,
 	}
 }
 
+// FlushMode is used to indicate the different urgency types for a flush.
+type FlushMode uint8
+
+const (
+	// FlushRequired is the flush mode that means a flush must be performed
+	// regardless of the cache state.  For example right before shutting down.
+	FlushRequired FlushMode = iota
+
+	// FlushPeriodic is the flush mode that means a flush can be performed
+	// when it would be almost needed.  This is used to periodically signal when
+	// no I/O heavy operations are expected soon, so there is time to flush.
+	FlushPeriodic
+
+	// FlushIfNeeded is the flush mode that means a flush must be performed only
+	// if the cache is exceeding a safety threshold very close to its maximum
+	// size.  This is used mostly internally in between operations that can
+	// increase the cache size.
+	FlushIfNeeded
+)
+
 // BlockChain provides functions for working with the bitcoin block chain.
 // It includes functionality such as rejecting duplicate blocks, ensuring blocks
 // follow all rules, orphan handling, checkpoint handling, and best chain
@@ -125,6 +145,12 @@ type BlockChain struct {
 	// efficient chain view into the block index.
 	index     *blockIndex
 	bestChain *chainView
+
+	// The UTXO state holds a cached view of the UTXO state of the chain.
+	//
+	// It has its own lock, however it is often also protected by the chain lock
+	// to help prevent logic races when blocks are being processed.
+	utxoCache *utxoCache
 
 	// These fields are related to handling of orphan blocks.  They are
 	// protected by a combination of the chain lock and the orphan lock.
@@ -622,14 +648,6 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
 			return err
 		}
 
-		// Update the utxo set using the state of the utxo view.  This
-		// entails removing all of the utxos spent and adding the new
-		// ones created by the block.
-		err = dbPutUtxoView(dbTx, view)
-		if err != nil {
-			return err
-		}
-
 		// Update the transaction spend journal by adding a record for
 		// the block that contains all txos spent by it.
 		err = dbPutSpendJournalEntry(dbTx, block.Hash(), stxos)
@@ -653,9 +671,9 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
 		return err
 	}
 
-	// Prune fully spent entries and mark all entries in the view unmodified
-	// now that the modifications have been committed to the database.
-	view.commit()
+	// Commit all modifications made to the view into the utxo state.  This also
+	// prunes these changes from the view.
+	b.utxoCache.Commit(view)
 
 	// This node is now the end of the best chain.
 	b.bestChain.SetTip(node)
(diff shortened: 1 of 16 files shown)


Roasbeef:
Tests fail as is, they also failed before of the cherry-picked "fix" commits. The set of extra commits actually seem to introduce even more test failures than without them.

Roasbeef:
Tests _should_ be passing now after that lastest commit. I don't like the fix in ee0b302 though, it should be _much_ more precise.

blockchain/chain.go
```diff
@@ -85,6 +85,26 @@ func newBestState(node *blockNode, blockSize, blockWeight, numTxns,
 	}
 }
 
+// FlushMode is used to indicate the different urgency types for a flush.
+type FlushMode uint8
+
+const (
+	// FlushRequired is the flush mode that means a flush must be performed
+	// regardless of the cache state.  For example right before shutting down.
```
maguayo:
```suggestion
	// regardless of the cache state. For example right before shutting down.
```

blockchain/chain.go
```diff
@@ -85,6 +85,26 @@ func newBestState(node *blockNode, blockSize, blockWeight, numTxns,
...
+	FlushRequired FlushMode = iota
+
+	// FlushPeriodic is the flush mode that means a flush can be performed
+	// when it would be almost needed.  This is used to periodically signal when
```
maguayo:
```suggestion
	// when it would be almost needed. This is used to periodically signal when
```

blockchain/chain.go
```diff
@@ -85,6 +85,26 @@ func newBestState(node *blockNode, blockSize, blockWeight, numTxns,
...
+	// no I/O heavy operations are expected soon, so there is time to flush.
+	FlushPeriodic
+
+	// FlushIfNeeded is the flush mode that means a flush must be performed only
+	// if the cache is exceeding a safety threshold very close to its maximum
+	// size.  This is used mostly internally in between operations that can
```
maguayo:
```suggestion
	// size. This is used mostly internally in between operations that can
```

txscript/script.go
```diff
@@ -848,7 +848,7 @@ func getWitnessSigOps(pkScript []byte, witness wire.TxWitness) int {
 }
 
 // IsUnspendable returns whether the passed public key script is unspendable, or
-// guaranteed to fail at execution.  This allows inputs to be pruned instantly
+// guaranteed to fail at execution.  This allows outputs to be pruned instantly
```
maguayo:
```suggestion
// guaranteed to fail at execution. This allows outputs to be pruned instantly
```

maguayo:
Any reason why the cache is limited to 250MB, just curious :)

blockchain/chain.go
```diff
@@ -772,9 +790,11 @@ func (b *BlockChain) disconnectBlock(node *blockNode, block *btcutil.Block, view
 		return err
 	}
 
-	// Prune fully spent entries and mark all entries in the view unmodified
-	// now that the modifications have been committed to the database.
-	view.commit()
+	// Commit all modifications made to the view into the utxo state.  This also
+	// prunes these changes from the view.
+	b.stateLock.Lock()
+	b.utxoCache.Commit(view)
```
yashbhutwala: Why not catch the error here as well? 😃 
```suggestion
	if err := b.utxoCache.Commit(view); err != nil {
		log.Errorf("error committing block %s(%d) to utxo cache: %s", block.Hash(), block.Height(), err.Error())
	}

```

blockchain/chain.go
```diff
@@ -1794,10 +1832,26 @@ func New(config *Config) (*BlockChain, error) {
 		return nil, err
 	}
 
-	bestNode := b.bestChain.Tip()
+	bestNode = b.bestChain.Tip()
```
yashbhutwala: Is this redundant after addition of line 1816?

blockchain/utxocache.go
```diff
@@ -0,0 +1,987 @@
...
+	s.mtx.Unlock()
+	return tmu
+}
+
+// seekAndCacheEntries attempts to fetch a series of entries from the
+// database.  In any aren't found, then nil is returned. All entries found are
+// cached in the process.
+//
+// This method should be called with the state lock held.
+func (s *utxoCache) seekAndCacheEntries(ops ...wire.OutPoint) (
```
yashbhutwala: Can we reduce code duplication between `seekAndCacheEntries` and `fetchAndCacheEntries`?  I'm not sure where `fetchAndCacheEntries` is used.

jakesyl:
@jcvernaleo (as per #1530)

- Low priority
- Outdated

jakesyl:
@Roasbeef can we make this a draft?

Roasbeef:
Wouldn't say the PR is outdated since it runs as is, and delivers a massive speed up in IBD. As mentioned in the recently closed (older) version of this PR, in combination with some of the other optimizations, `btcd` sync the chain in _under 24 hours_. This includes full verification after the past checkpoint, unlike `bitcoind`'s `assumevalid` which only does UTXO ingestion primarily.

jcvernaleo:
@Roasbeef I think we may have been using `outdated` to mean more like a catch-all for either not up to date or just not ready.  Not a great way to do it for sure, but we had a lot of triage to do :)

This PR is definitely super important so I'm really looking forward to when it can go in.

jakesyl:
Yep, that was the case. Will go ahead and run this on a full sync today to benchmark

kcalvinalvin:
Is the bug in `dbSeekUtxoEntry()` already known? In the tests, `dbSeekUtxoEntry()` will sometimes  return a tx that's new. This causes `checkBIP0030()` to fail as `seekAndCacheEntries()` will return a tx that hasn't been committed with `Commit()` method on `utxoCache`. Using `fetchAndCacheEntries()` is a workaround to this.

rstaudt2:
In case anyone is interested when picking up this work, we implemented a UTXO Cache in `dcrd` earlier this year, and portions of it could likely be pretty easily reused here.  It accounts for the design flaws mentioned in this PR description as well.  The PR that introduces it is https://github.com/decred/dcrd/pull/2591.

Roasbeef:
Closing in favor of https://github.com/btcsuite/btcd/pull/1955

PR #1375, opened 2019-01-18
txscript: optimize IsUnspendable by removing all allocations

Roasbeef:
During my work to optimize the IBD of btcd, I've been profiling heavily.
I've noticed that this call `IsUnspendable` shows up in several hot
spots during block validation, as we need to know if we should add a
UTXO to the view/cache, etc. The current version uses `parseScript` to
parse out all the pops and such. This is nice when we want to be able to
easily manipulate things like data pushes, but for this simple check,
it's unnecessary. The new version simply makes checks against the raw
script which saves us a ton of allocations and CPU cycles.

### DIFF
--- a/txscript/script.go
+++ b/txscript/script.go
@@ -847,14 +847,10 @@ func getWitnessSigOps(pkScript []byte, witness wire.TxWitness) int {
 	return 0
 }
 
-// IsUnspendable returns whether the passed public key script is unspendable, or
-// guaranteed to fail at execution.  This allows inputs to be pruned instantly
-// when entering the UTXO set.
+// IsUnspendable returns whether the passed public key script is unspendable,
+// or guaranteed to fail at execution.  This allows outputs to be pruned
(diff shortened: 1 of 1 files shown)


davecgh:
While I completely agree with the concept of this PR (in fact I'd ultimately like to see much less parsing of the scripts into pops in general), this is a change to the semantics and would result in entries that can't be parsed being admitted to the utxo set.  Any attempts to spend them would fail, so it shouldn't be a problem from a standpoint of consensus, but still not entirely desirable.

Perhaps it would make sense an `IsParsable` function which essentially does what `parseScriptTemplate` does, but rather than performing any allocations for the opcodes and data, it merely scans the script in the same way and errors.

I believe a function like that would be useful for more than just this.

jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Outdated

Roasbeef:
No longer needed anymore due to #1769

PR #1376, opened 2019-01-19
wire: cache the non-witness serialization of MsgTx to memoize part of…

Roasbeef:
… TxHash

In this commit, we add a new field to the `MsgTx` struct:
`cachedSeralizedNoWitness`. As we decode the main transaction, we use an
`io.TeeReader` to copy over the non-witness bytes into this new field.
As a result, we can fully cache all tx serialization when computing the
TxHash. This has been shown to show up on profiles during IBD. Caching
this value allows us to optimize TxHash calculation across the entire
daemon as a whole.

### DIFF
--- a/wire/msgtx.go
+++ b/wire/msgtx.go
@@ -877,6 +877,10 @@ func (msg *MsgTx) Serialize(w io.Writer) error {
 // Serialize, however even if the source transaction has inputs with witness
 // data, the old serialization format will still be used.
 func (msg *MsgTx) SerializeNoWitness(w io.Writer) error {
+	if msg.cachedSeralizedNoWitness != nil {
+		w.Write(msg.cachedSeralizedNoWitness)
+	}
+
 	return msg.BtcEncode(w, 0, BaseEncoding)
 }
 
--- a/wire/msgtx.go
+++ b/wire/msgtx.go
@@ -343,6 +343,12 @@ type MsgTx struct {
 	TxIn     []*TxIn
 	TxOut    []*TxOut
 	LockTime uint32
+
+	// cachedSeralizedNoWitness is a cached version of the serialization of
+	// this transaction without witness data. When we decode a transaction,
+	// we'll write out the non-witness bytes to this so we can quickly
+	// calculate the TxHash later if needed.
+	cachedSeralizedNoWitness []byte
 }
 
 // AddTxIn adds a transaction input to the message.
@@ -357,13 +363,19 @@ func (msg *MsgTx) AddTxOut(to *TxOut) {
 
 // TxHash generates the Hash for the transaction.
 func (msg *MsgTx) TxHash() chainhash.Hash {
-	// Encode the transaction and calculate double sha256 on the result.
-	// Ignore the error returns since the only way the encode could fail
-	// is being out of memory or due to nil pointers, both of which would
-	// cause a run-time panic.
-	buf := bytes.NewBuffer(make([]byte, 0, msg.SerializeSizeStripped()))
-	_ = msg.SerializeNoWitness(buf)
-	return chainhash.DoubleHashH(buf.Bytes())
+	if msg.cachedSeralizedNoWitness == nil {
+		// Encode the transaction and calculate double sha256 on the
+		// result.  Ignore the error returns since the only way the
+		// encode could fail is being out of memory or due to nil
+		// pointers, both of which would cause a run-time panic.
+		strippedSize := msg.SerializeSizeStripped()
+		buf := bytes.NewBuffer(make([]byte, 0, strippedSize))
+		_ = msg.SerializeNoWitness(buf)
+
+		msg.cachedSeralizedNoWitness = buf.Bytes()
+	}
+
+	return chainhash.DoubleHashH(msg.cachedSeralizedNoWitness)
 }
 
 // WitnessHash generates the hash of the transaction serialized according to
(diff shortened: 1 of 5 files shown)


Roasbeef:
There's a bug here in that it doesn't detect mutations in the underlying `*wire.MsgTx` so if the value is changed at all, then we don't update the serialization. In the current pipeline for block validation, this doesn't manifest, but some of our tests do this which introduces some of the new failures in Travis.

jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Outdated (needs updates to be merged)

Roasbeef:
Here's a recent run late in the chain: 

Screenshot 2023-03-13 at 4 40 00 PM

From this we can see that when checking signatures, we actually spend most of our time just encoding the tx again to compute the _sighash_. 

Here's a flame graph for another view: 

Screenshot 2023-03-13 at 4 42 30 PM

cc @kcalvinalvin

kcalvinalvin:
I need a refresher on this but if memory serves right, doing away with the channel stuff in binaryFreeList and using a sync.Pool got rid of a big chunk of the time taken.

> From this we can see that when checking signatures, we actually spend most of our time just encoding the tx again to compute the sighash.

But yeah if this is something that doesn't need to happen in the first place with memoization that'd be much better.

Roasbeef:
Recent observation of the impact here: 
Screenshot 2023-08-09 at 4 45 09 PM

Roasbeef:
Related PR re binary free list: https://github.com/btcsuite/btcd/pull/1426

Roasbeef:
Rebased!

Roasbeef:
Dug a bit, and I realize the issue is in the test itself, it mutates transactions, with stuff like `replaceSpendScript`. So what we need here (for the test's sake) is that we have a method to wipe the cached serialization if something is mutated. Alternatively, we have some internal map that uses the `MsgTx` as a value, which is consulted to see if the transaction has changed since last time.

Roasbeef:
Pushed a tx implementing the above, don't think it's ideal though...worried about weird edge cases where transaction generation code uses a tx as a scratch pad and makes multiple versions to sign, with this it'll give the wrong `txid` after the first instance. 

Also added a commit that'll re-use the non-witness serialization for `SerializeNoWitness`.

Crypt-iQ (commented):
I looked in both btcd and lnd to see if we modify `wire.MsgTx` after calculating the hash, but couldn't find any instances. The code looks good, but I think we need to be extra careful and see if there are any instances across our codebases

wire/msgtx.go
```diff
@@ -495,6 +517,15 @@ func (msg *MsgTx) BtcDecode(r io.Reader, pver uint32, enc MessageEncoding) error
 		if err != nil {
 			return err
 		}
+
+		hasWitneess = true
+	}
+
+	// Write out the actual number of inputs as this won't be the very byte
+	// series after the versino of segwit transactions.
+	if WriteVarInt(&rawTxBuf, pver, count); err != nil {
```
guggero, line 533: Need to add `err := WriteVarInt()`.

wire/msgtx.go
```diff
@@ -495,6 +517,15 @@ func (msg *MsgTx) BtcDecode(r io.Reader, pver uint32, enc MessageEncoding) error
 		if err != nil {
 			return err
 		}
+
+		hasWitneess = true
+	}
+
+	// Write out the actual number of inputs as this won't be the very byte
+	// series after the versino of segwit transactions.
```
guggero, line 532: nit: s/versino/version

wire/msgtx.go
```diff
@@ -343,6 +343,12 @@ type MsgTx struct {
 	TxIn     []*TxIn
 	TxOut    []*TxOut
 	LockTime uint32
+
+	// cachedSeralizedNoWitness is a cached version of the serialization of
+	// this transaction without witness data. When we decode a transaction,
+	// we'll write out the non-witness bytes to this so we can quickly
+	// calculate the TxHash later if needed.
+	cachedSeralizedNoWitness []byte
```
guggero, line 351: nit: s/cachedSeralizedNoWitness/cachedSerializedNoWitness/

wire/msgtx.go
```diff
@@ -378,6 +378,13 @@ func (msg *MsgTx) TxHash() chainhash.Hash {
 	return chainhash.DoubleHashH(msg.cachedSeralizedNoWitness)
 }
 
+// WipeCache removes the cached serialized bytes of the transaction. This is
+// useful to be able to get the correct txid after mutating a transaction's
+// state.
+func (msg *MsgTx) WipeCache() {
+	msg.cachedSeralizedNoWitness = nil
```
guggero, line 385: Don't we also need to call this in the helper methods like `AddTxIn()` and `AddTxOut()`? Or is the assumption that `TxHash()` would in practice only be called once the transaction is fully built, so the cache doesn't need to be invalidated?

I'm mostly worrying about uses of `MsgTx` outside of `btcd`, where I'm not sure we can 100% guarantee that we're always using this pattern...

guggero (changes requested):
While I like the elegance of using a `io.TeeReader` (TIL by the way), I'm not sure if this would break a bunch of things outside of `btcd`, namely in `lnd` and related projects.

What if we make the caching optional with an additional struct member boolean (or if memory footprint is a concern, use a bit within the `Version` field) that turns the caching on?
That way we could both benefit from the optimization within `btcd` while not breaking any assumptions for outside code.

Roasbeef:
> What if we make the caching optional with an additional struct member boolean (or if memory footprint is a concern, use a bit within the Version field) that turns the caching on?


Not a bad idea, agree that this could have a lot of unintended consequences. The other approach uses `btcutil.Tx` which gives another route to implement something like this.

Roasbeef:
Closing in favor of #2023, see this comment for some details: https://github.com/btcsuite/btcd/pull/2023#issuecomment-1858625556

TL;DR: I think we need to remove/optimize the binary free list instead.

PR #1381, opened 2019-01-27
btcd.go: Handle returned error for removing regression database

qshuai:
Improving


jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Enhancement

allocz (approved):
Thanks for the patch, LGTM. @Roasbeef we are not handling this error on [master](https://github.com/btcsuite/btcd/blob/b48125d0a3565b1441522ee10422f7815db03216/btcd.go#L389).

PR #1382, opened 2019-01-29
blockchain/indexers: Start a new line for long code

qshuai:


jakesyl:
@jcvernaleo (as per #1530)

- Low priority

@jcvernaleo we can merge this now if we want to, just a formatting change

PR #1383, opened 2019-01-29
addrmgr: store address' supported service bits

wpaulino:
In this commit, we update the serialized version of the AddressManager
to also store each address' service bits. This allows an address'
service bits to be properly set when read from disk, which allows
external callers to reliably filter out addresses which do not support
their required services.

### DIFF
--- a/addrmgr/addrmanager.go
+++ b/addrmgr/addrmanager.go
@@ -635,21 +635,9 @@ func (a *AddrManager) NeedMoreAddresses() bool {
 // AddressCache returns the current address cache.  It must be treated as
 // read-only (but since it is a copy now, this is not as dangerous).
 func (a *AddrManager) AddressCache() []*wire.NetAddress {
-	a.mtx.Lock()
-	defer a.mtx.Unlock()
+	allAddr := a.getAddresses()
 
-	addrIndexLen := len(a.addrIndex)
-	if addrIndexLen == 0 {
-		return nil
-	}
-
-	allAddr := make([]*wire.NetAddress, 0, addrIndexLen)
-	// Iteration order is undefined here, but we randomise it anyway.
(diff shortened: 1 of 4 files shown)


addrmgr/addrmanager.go
```diff
@@ -55,6 +55,8 @@ type serializedKnownAddress struct {
 	TimeStamp   int64
 	LastAttempt int64
 	LastSuccess int64
+	Services    wire.ServiceFlag
```
Roasbeef: Don't see these new fields being set in `savePeers`. Also it'd be nice to get some basic test coverage for this new addition to ensure that serialization version bump worked properly.

Roasbeef:
Also fails travis: 
```
addrmgr/addrmanager_test.go:265:42: not enough arguments in call to n.DeserializeNetAddress
```

wpaulino:
Tests added, PTAL @Roasbeef.

> Also fails travis:

I don't know what's causing the current travis failures. All tests passed when running `./goclean.sh` locally. This is also the case with the version I just pushed.

Roasbeef:
More faillures: 
```
addrmgr/addrmanager_internal_test.go:63:18: addrMgr.getAddresses undefined (type *AddrManager has no field or method getAddresses)
addrmgr/addrmanager_internal_test.go:138:9: addrMgr.version undefined (type *AddrManager has no field or method version)
addrmgr/addrmanager_internal_test.go:155:9: addrMgr.version undefined (type *AddrManager has no field or method version)
addrmgr/addrmanager_internal_test.go:163:18: addrMgr.getAddresses undefined (type *AddrManager has no field or method getAddresses)
addrmgr/addrmanager_internal_test.go:186:9: addrMgr.version undefined (type *AddrManager has no field or method version)
```

PR #1384, opened 2019-01-29
blockchain/indexers: Stop processing in advance when no requested items

qshuai:


jakesyl:
@jcvernaleo (as per #1530)
- Low priority
- Enhancement

blockchain/indexers/addrindex.go
allocz, line 288: Checked against master branch and this patch still makes sense, as the result will always be nil when `numRequested` is zero.

allocz (commented):
Thanks for the PR, the patch still makes sense, because we can return early without calling the database in case `numRequested` is zero, as zero entries was requested, I don't see problems in returning early. But the patch needs a rebase with conflict resolution.

PR #1393, opened 2019-02-17
wire: simplify *new(T) to T{}

quasilyte:
Dereferencing new(T) result is always identical to T zero value.
The compiler is smart enough not to generate `new` call in such cases,
but simple T literal is also better from code readability perspective.
`*new(T)` can also look like a suspicious code for static code analyzers,
since this form is quite unusual (redundant).

Signed-off-by: Iskander Sharipov <quasilyte@gmail.com>

### DIFF
--- a/wire/msgalert_test.go
+++ b/wire/msgalert_test.go
@@ -65,7 +65,7 @@ func TestMsgAlert(t *testing.T) {
 	// Test BtcEncode with Payload != nil
 	// note: Payload is an empty Alert but not nil
 	msg.Payload = new(Alert)
-	buf = *new(bytes.Buffer)
+	buf = bytes.Buffer{}
 	err = msg.BtcEncode(&buf, pver, encoding)
 	if err != nil {
 		t.Error(err.Error())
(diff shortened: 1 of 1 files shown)


wire/msgalert_test.go
```diff
@@ -65,7 +65,7 @@ func TestMsgAlert(t *testing.T) {
 	// Test BtcEncode with Payload != nil
 	// note: Payload is an empty Alert but not nil
 	msg.Payload = new(Alert)
-	buf = *new(bytes.Buffer)
+	buf = bytes.Buffer{}
```
cfromknecht: any reason this form was chosen over `var buf bytes.Buffer`?

jakesyl:
@jcvernaleo (as per #1530)

- Low priority
- Enhancement

Should be easily mergable

jcvernaleo:
@quasilyte any chance you could rebase this?

jcvernaleo:
As soon as we do that so action pass I'll merge it.

quasilyte:
It's not an important change and it was quite a long time ago (I don't have local copy right now and I don't want to pull it right now). Anyone else can open a new PR that performs that simplification if they feel compelled enough to do so. :)

PR #1399, opened 2019-03-03
Put best state with parent's workSum

imnisen:
When disconnecting block, why put best state with `node.workSum` not parent's worksum`node.parent.workSum`

### DIFF
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -722,7 +722,7 @@ func (b *BlockChain) disconnectBlock(node *blockNode, block *btcutil.Block, view
 
 	err = b.db.Update(func(dbTx database.Tx) error {
 		// Update best block state.
-		err := dbPutBestState(dbTx, state, node.workSum)
+		err := dbPutBestState(dbTx, state, node.parent.workSum)
 		if err != nil {
 			return err
 		}


jakesyl:
@jcvernaleo (as per #1530)
- Low priority

blockchain/chain.go
```diff
@@ -722,7 +722,7 @@ func (b *BlockChain) disconnectBlock(node *blockNode, block *btcutil.Block, view
 
 	err = b.db.Update(func(dbTx database.Tx) error {
 		// Update best block state.
-		err := dbPutBestState(dbTx, state, node.workSum)
+		err := dbPutBestState(dbTx, state, node.parent.workSum)
```
allocz, line 725: If fact, as we are disconnecting the current chain tip, the new best state should have the worksum of the previous block, as the tip was disconnected and we expect the total worksum to be reduced. This looks like a bug in block disconnection and is still present in the master branch.

allocz (commented):
Thanks for the PR. If this is a real bug, would be nice to add a test that fails with the bug and passes after the patch, preventing regressions. I've checked here and I could rebase this on top of master with no conflicts, also, after the patch, all the tests still passing, this means that we can save wrong `workSum` without any test catching the regression.

PR #1400, opened 2019-03-04
Fix HDCoinType for simnet

yura-pakhuchiy:
According to SLIP44[1] all test networks should use coin type 1. Coin type 115 is assigned to some altcoin.

1. https://github.com/satoshilabs/slips/blob/master/slip-0044.md

### DIFF
--- a/chaincfg/params.go
+++ b/chaincfg/params.go
@@ -565,7 +565,7 @@ var SimNetParams = Params{
 
 	// BIP44 coin type used in the hierarchical deterministic path for
 	// address generation.
-	HDCoinType: 115, // ASCII for s
+	HDCoinType: 1,
 }
 
 var (


jakesyl:
@jcvernaleo (as per #1530)
- Low priority

chaincfg/params.go
```diff
@@ -575,7 +565,7 @@ var SimNetParams = Params{
 
 	// BIP44 coin type used in the hierarchical deterministic path for
 	// address generation.
-	HDCoinType: 115, // ASCII for s
+	HDCoinType: 1,
```
allocz, line 568: Currently `HDCoinType` is not even being read in btcd source files, and I don't see any reasonably justification to change this.

Image

Thanks for the suggestion but unless you have a concrete problem, I think that the PR can be closed.

PR #1410, opened 2019-03-25
Improve error message about non-active segwit on simnet

darwin:
I started playing with btcd+lnd on simnet and was confronted with error message:

```
[ERR] FNDG: Unable to broadcast funding tx for ChannelPoint(<point>:0):
-22: TX rejected: transaction <tx> has witness data, but segwit isn't active yet
```

I wasn't aware of the activation period so I got quite puzzled.
Google helped. But I think the message could mention likely cause.

Newly it optionally prints something like:

```
(The threshold for segwit activation is 300 blocks on simnet, current best height is 113)
```


farukterzioglu:
I was banging my head to the wall why I can't use some tx as input. I wasn't aware of 300 block limitation. 
Thanks a lot @darwin. 

This commit can help a lot for simnet users.

jakesyl:
@jcvernaleo (as per #1530)

- Low priority
- Enhancement

Easily mergable, just changes an error message

PR #1413, opened 2019-04-09
Add signrawtransactionwithwallet rpc command

vctt94:
This PR adds signrawtransactionwithwallet command which was added to bitcoin at 0.17.0

https://bitcoincore.org/en/doc/0.17.0/rpc/wallet/signrawtransactionwithwallet/


tiger5226:
Why hasn't this been merged? Are there any issues with the implementation? Should I not use this code? Any information from the maintainers would be helpful. 

@vctt94 Thanks for the contribution. I guess I will need to make my own branch for now.

tiger5226:
@Roasbeef any idea?

rpcclient/rawtransactions.go
```diff
@@ -372,6 +372,72 @@ func (c *Client) SignRawTransaction(tx *wire.MsgTx) (*wire.MsgTx, bool, error) {
...
+}
+
+// SignRawTransactionWithWalletAsync returns an instance of a type that can be used to get
+// the result of the RPC at some future time by invoking the Receive function on
+// the returned instance.
+//
+// See SignRawTransaction for the blocking version and more details.
+func (c *Client) SignRawTransactionWithWalletAsync(tx *wire.MsgTx) FutureSignRawTransactionResult {
+	txHex := ""
+	if tx != nil {
```
torkelrogstad: If the user passes in nil here, we'll get an error response from the server. Wouldn't it be better if the RPC errored before making the network call?

tiger5226: @torkelrogstad I agree. Its best for the client to handle situations known to produce errors in advance. 

@vctt94 I can make the change to error if a nil `tx` is passed in. I can also rebase the change too, so we can get it merged. Let me know if you would rather do it.

jakesyl:
@jcvernaleo (as per #1530)

- Low priority
- Enhancement
- Outdated (needs to implement changes from @torkelrogstad's review)

rpcclient/rawtransactions.go
vctt94: Hey, sorry for the late reply.

Sure you can make the changes, no problems! Thanks for that

guggero:
WTF...

guggero:
Okay, I'm giving up. No idea why GitHub doesn't let me re-open the PR, sorry.

jcvernaleo:
@guggero I tried to reopen it too and it gets immediately closed as well.  Very confusing why github is doing that.

PR #1415, opened 2019-04-10
Switch AllowHighFee to MaxFeeRate

vctt94:
After commit https://github.com/bitcoin/bitcoin/commit/c033c4b5cef89a654e4d9d5c5f9bd823871b068b bitcoin switched allowHighFee to maxFeeRate on sendrawtransaction rpc command.

This PR switches AllowHighFees to MaxFeeRate as well.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -3299,14 +3299,6 @@ func handleSearchRawTransactions(s *rpcServer, cmd interface{}, closeChan <-chan
 // handleSendRawTransaction implements the sendrawtransaction command.
 func handleSendRawTransaction(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (interface{}, error) {
 	c := cmd.(*btcjson.SendRawTransactionCmd)
-	if s.cfg.TxIndex == nil {
-		return nil, &btcjson.RPCError{
-			Code: btcjson.ErrRPCNoTxInfo,
-			Message: "The transaction index must be " +
-				"enabled to query the blockchain " +
-				"(specify --txindex)",
-		}
-	}
 	// Deserialize and send off to tx relay
 	hexStr := c.HexTx
 	if len(hexStr)%2 != 0 {
@@ -3327,24 +3319,7 @@ func handleSendRawTransaction(s *rpcServer, cmd interface{}, closeChan <-chan st
 
 	// Use 0 for the tag to represent local node.
 	tx := btcutil.NewTx(&msgTx)
-
-	var totalSatoshiIn int64
-	inputs, err := fetchInputTxos(s, &msgTx)
-	for _, txIn := range inputs {
-		totalSatoshiIn += txIn.Value
-	}
-	var totalSatoshiOut int64
-	for _, txOut := range tx.MsgTx().TxOut {
-		totalSatoshiOut += txOut.Value
-	}
-	txFee := totalSatoshiIn - totalSatoshiOut
-	if txFee > int64(*c.MaxFeeRate) {
-		return nil, &btcjson.RPCError{
-			Code:    btcjson.ErrRPCDeserialization,
-			Message: "Fee is higher than the max fee allowed",
-		}
-	}
-	acceptedTxs, err := s.cfg.TxMemPool.ProcessTransaction(tx, false, false, 0)
+	acceptedTxs, err := s.cfg.TxMemPool.ProcessTransaction(tx, false, false, 0, c.MaxFeeRate)
 	if err != nil {
 		// When the error is a rule error, it means the transaction was
 		// simply rejected as opposed to something actually going wrong,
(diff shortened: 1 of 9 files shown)


Roasbeef:
Seems we also need to update the `btcd` server as well. Otherwise, we won't be able to respond to this new request.

vctt94:
Thanks for reviewing.

I updated as requested.

rpcserver.go
```diff
@@ -3299,6 +3299,14 @@ func handleSearchRawTransactions(s *rpcServer, cmd interface{}, closeChan <-chan
 // handleSendRawTransaction implements the sendrawtransaction command.
 func handleSendRawTransaction(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (interface{}, error) {
 	c := cmd.(*btcjson.SendRawTransactionCmd)
+	if s.cfg.TxIndex == nil {
```
Roasbeef: Seems like a pretty big downgrade if we now require the `txindex` just to _send_ a transaction over the interface. Instead, during validation in the mempool, we know the direct fee since it's stored as part of the mempool transaction entry. So we can put the logic deeper into the mempool, accepting an optional argument of the max fee that the resulting transaction should be allowed to have.

rpcserver.go
```diff
@@ -3319,6 +3327,23 @@ func handleSendRawTransaction(s *rpcServer, cmd interface{}, closeChan <-chan st
 
 	// Use 0 for the tag to represent local node.
 	tx := btcutil.NewTx(&msgTx)
+
+	var totalSatoshiIn int64
+	inputs, err := fetchInputTxos(s, &msgTx)
```
Roasbeef: Alternatively, we can just check the UTXO set here. As if the inputs are already spent, it'll be rejected within the mempool anyway.

rpcserver.go
```diff
@@ -3319,6 +3327,23 @@ func handleSendRawTransaction(s *rpcServer, cmd interface{}, closeChan <-chan st
...
+	for _, txIn := range inputs {
+		totalSatoshiIn += txIn.Value
+	}
+	var totalSatoshiOut int64
+	for _, txOut := range tx.MsgTx().TxOut {
+		totalSatoshiOut += txOut.Value
+	}
+	txFee := totalSatoshiIn - totalSatoshiOut
+	if txFee > int64(*c.MaxFeeRate) {
```
cfromknecht: this is comparing two different things. the limit is a _feerate_ in sats/kB, which is being compared to the _total fee_ in sats

rpcserver.go
```diff
@@ -3319,6 +3327,23 @@ func handleSendRawTransaction(s *rpcServer, cmd interface{}, closeChan <-chan st
...
+		return nil, &btcjson.RPCError{
+			Code:    btcjson.ErrRPCDeserialization,
```
cfromknecht: doesn't seem like the correct response code, since the request was successfully deserialized

rpcserverhelp.go
```diff
@@ -536,10 +536,10 @@ var helpDescsEnUS = map[string]string{
 	"searchrawtransactions--result0":    "Hex-encoded serialized transaction",
 
 	// SendRawTransactionCmd help.
-	"sendrawtransaction--synopsis":     "Submits the serialized, hex-encoded transaction to the local peer and relays it to the network.",
-	"sendrawtransaction-hextx":         "Serialized, hex-encoded signed transaction",
-	"sendrawtransaction-allowhighfees": "Whether or not to allow insanely high fees (btcd does not yet implement this parameter, so it has no effect)",
-	"sendrawtransaction--result0":      "The hash of the transaction",
+	"sendrawtransaction--synopsis":  "Submits the serialized, hex-encoded transaction to the local peer and relays it to the network.",
+	"sendrawtransaction-hextx":      "Serialized, hex-encoded signed transaction",
+	"sendrawtransaction-maxfeerate": "Reject transactions whose fee rate is higher than the specified value. The value is given in satoshis",
```
cfromknecht: if the intent is to match the bitcoind implementation, the `maxfeerate` is in sats/kB and not satoshis

https://github.com/bitcoin/bitcoin/pull/13541/files#diff-01aa7d1d32f1b9e5a836c9c411978918R1043

vctt94:
Thanks for pointing that Roasbeef. I moved the maxFeeRate to the `ProcessTransaction` method. Believe this looks better and there is no need of `txindex`, anymore. 

Also fixed the others comments

rpcserverhelp.go
vctt94: fixed

rpcserver.go
vctt94: returning `wire.RejectInvalid` now. I believe it is more appropriate

jakesyl:
@jcvernaleo (as per #1530)

- Low priority
- Bug (because RPC is outdated)
- Outdated

onyb:
@vctt94 Can you please rebase this PR on latest master? Let's try to have dual compatibility, following the fixes to the `sendrawtransaction` command done in #1659.

guggero:
Not sure why GitHub closed this upon merge of another PR. Re-opening, sorry.

guggero:
Okay, this seems to be a bug with the GitHub UI, will try again later.

guggero:
Okay, I'm giving up. No idea why GitHub doesn't let me re-open the PR, sorry.

PR #1377, opened 2019-01-19
blockchain: add new rolling merkle root computation to reduce allocations during IBD

Roasbeef:

PR #1421, opened 2019-04-18
blockchain: rolling merkle root

cfromknecht:
This PR adds a new `RollingMerkleTree` type for efficiently computing merkle roots. The leaves
of the tree are iteratively pushed onto the tree, and any complete substrees are evaluated and
stored so that subsequent pushes may consume them when their sibling tree is complete. As a 
result, this requires `O(log n)` additional memory be allocated. The current `BuildMerkleTreeStore`
allocates an array of up to 4 times the number of leaves, which is used to store internal nodes and the root. 

For 2000 leaves, the benchmarks show that this method is about 7% slower (~90 microseconds), but uses 99.04% less memory and 99.85% fewer allocations. These allocations reduce gc pressure, especially during IBD when the large slices of hashes currently used are almost immediately discarded.
```
BenchmarkMerkle/rolling-1000-8              2000            694382 ns/op             858 B/op          3 allocs/op
BenchmarkMerkle/rolling-2000-8              1000           1367763 ns/op             923 B/op          3 allocs/op
BenchmarkMerkle/rolling-4000-8               500           2900103 ns/op            1210 B/op          6 allocs/op
BenchmarkMerkle/rolling-8000-8               300           5799806 ns/op            2034 B/op         10 allocs/op
BenchmarkMerkle/rolling-16000-8              100          11572403 ns/op            1578 B/op          3 allocs/op
BenchmarkMerkle/rolling-32000-8               50          22949069 ns/op            1720 B/op          5 allocs/op
BenchmarkMerkle/old-1000-8                  2000            655435 ns/op           48416 B/op       1002 allocs/op
BenchmarkMerkle/old-2000-8                  1000           1277192 ns/op           96800 B/op       2002 allocs/op
BenchmarkMerkle/old-4000-8                   500           2533821 ns/op          193568 B/op       4002 allocs/op
BenchmarkMerkle/old-8000-8                   300           5029310 ns/op          387104 B/op       8002 allocs/op              
BenchmarkMerkle/old-16000-8                  200          10135631 ns/op          774176 B/op      16002 allocs/op                     
BenchmarkMerkle/old-32000-8                  100          20442194 ns/op         1548340 B/op      32002 allocs/op
```

alternative to #1377

### DIFF
--- a/blockchain/merkle_test.go
+++ b/blockchain/merkle_test.go
@@ -15,12 +15,20 @@ import (
 // TestMerkle tests the BuildMerkleTreeStore API.
 func TestMerkle(t *testing.T) {
 	block := btcutil.NewBlock(&Block100000)
-	merkles := BuildMerkleTreeStore(block.Transactions(), false)
-	calculatedMerkleRoot := merkles[len(merkles)-1]
+	calcMerkleRoot := CalcMerkleRoot(block.Transactions(), false)
+	merkleStoreTree := BuildMerkleTreeStore(block.Transactions(), false)
+	merkleStoreRoot := merkleStoreTree[len(merkleStoreTree)-1]
+
+	if calcMerkleRoot != *merkleStoreRoot {
+		t.Fatalf("BuildMerkleTreeStore root does not match "+
+			"CalcMerkleRoot, store: %v calc: %v", merkleStoreRoot,
+			calcMerkleRoot)
+	}
+
 	wantMerkle := &Block100000.Header.MerkleRoot
-	if !wantMerkle.IsEqual(calculatedMerkleRoot) {
+	if !wantMerkle.IsEqual(&calcMerkleRoot) {
 		t.Errorf("BuildMerkleTreeStore: merkle root mismatch - "+
-			"got %v, want %v", calculatedMerkleRoot, wantMerkle)
+			"got %v, want %v", calcMerkleRoot, wantMerkle)
 	}
 }
 
--- a/blockchain/merkle_test.go
+++ b/blockchain/merkle_test.go
@@ -24,6 +24,36 @@ func TestMerkle(t *testing.T) {
 	}
 }
 
+// TestMerkleBlock46559 computes the merkle root of block 46559. The number of
+// transactions in this blocks forces the case where a left leaf must be hashed
+// with itself, since the block only has three transactions.
+func TestRollingMerkleBlock46559(t *testing.T) {
+	merkle := NewRollingMerkleTree(3)
+
+	coinbase, _ := chainhash.NewHashFromStr(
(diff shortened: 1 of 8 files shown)


blockchain/rolling_merkle.go
```diff
@@ -0,0 +1,208 @@
...
+// RollingMerkleTree computes the double SHA256 merkle root over a set of leaf
+// hashes. Interior nodes are opportunistically computed as new leaves are
+// added to the tree, consolidating any nodes that have both a left and right
+// child before permitting another leaf to be added. As a result, a
+// RollingMerkleTree only requires O(log n) additional storage to compute the
+// final root.
+type RollingMerkleTree struct {
+	// nodes contains all nodes in the tree. The nodes may correspond to
+	// interior nodes if their children have already been pruned.
+	nodes map[uint32]chainhash.Hash
```
Roasbeef: Wouldn't this be better served as a slice? So we just index in rather than having the overhead of a map. We'd also gain memory locality for all the items as well.

blockchain/merkle_test.go
```diff
@@ -15,12 +15,20 @@ import (
 // TestMerkle tests the BuildMerkleTreeStore API.
 func TestMerkle(t *testing.T) {
 	block := btcutil.NewBlock(&Block100000)
-	merkles := BuildMerkleTreeStore(block.Transactions(), false)
-	calculatedMerkleRoot := merkles[len(merkles)-1]
+	calcMerkleRoot := CalcMerkleRoot(block.Transactions(), false)
+	merkleStoreTree := BuildMerkleTreeStore(block.Transactions(), false)
+	merkleStoreRoot := merkleStoreTree[len(merkleStoreTree)-1]
+
+	if calcMerkleRoot != *merkleStoreRoot {
```
Roasbeef: 👍

jakesyl:
@jcvernaleo (as per #1530)

- Low priority
- Enhancement
- Duplicate (of #1377, not sure if we have a tag for this)

Rjected:
@cfromknecht mind rebasing so the actions can run?

Roasbeef:
Superseded by https://github.com/btcsuite/btcd/pull/1979

moza88:
I noticed BuildMerkleTreeStore is still in the codebase even though it's replaced with RollingMerkleTree and CalcMerkleRoot. Should we remove it from [blockchain/merkle.go](https://github.com/btcsuite/btcd/blob/de4182589bfe0d1e2a6b45348ce283e7035b5d3d/blockchain/merkle.go#L107) and [blockchain/merkle_test.go](https://github.com/btcsuite/btcd/blob/de4182589bfe0d1e2a6b45348ce283e7035b5d3d/blockchain/merkle_test.go#L17)?

PR #1423, opened 2019-04-26
build: update to latest version of btcutil

Roasbeef:
In this commit, we update to the latest version of btcutil which
contains a bug fix for filter matching, and as optimizations speed up
both filter matching and constructions.


cfromknecht:
LGTB 💰

PR #1427, opened 2019-05-05
mempool: implement RBF signaling policy

wpaulino:
In this PR, we implement the [Replace-By-Fee (RBF) signaling policy](https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki) for `btcd`'s mempool. This allows unconfirmed transactions to be replaced with higher fee rate versions of themselves. `bitcoind`'s implementation can be cross-referenced [here](https://github.com/bitcoin/bitcoin/blob/master/src/validation.cpp#L571).


mempool/mempool.go
```diff
@@ -554,21 +564,123 @@ func (mp *TxPool) addTransaction(utxoView *blockchain.UtxoViewpoint, tx *btcutil
...
+		}
+		if !mp.signalsReplacement(conflict) {
 			str := fmt.Sprintf("output %v already spent by "+
 				"transaction %v in the memory pool",
-				txIn.PreviousOutPoint, txR.Hash())
-			return txRuleError(wire.RejectDuplicate, str)
+				txIn.PreviousOutPoint, conflict.Hash())
+			return false, txRuleError(wire.RejectDuplicate, str)
 		}
+		return true, nil
```
Roasbeef: What if there's two inputs on this new transaction and one of the signals replacement while the other doesn't?

mempool/mempool.go
```diff
@@ -554,21 +564,123 @@ func (mp *TxPool) addTransaction(utxoView *blockchain.UtxoViewpoint, tx *btcutil
...
+// This function MUST be called with the mempool lock held (for reads).
+func (mp *TxPool) txAncestors(tx *btcutil.Tx) map[chainhash.Hash]*btcutil.Tx {
+	ancestors := make(map[chainhash.Hash]*btcutil.Tx)
+	for _, txIn := range tx.MsgTx().TxIn {
+		parent, ok := mp.pool[txIn.PreviousOutPoint.Hash]
+		if !ok {
+			continue
+		}
+		ancestors[*parent.Tx.Hash()] = parent.Tx
+		for hash, ancestor := range mp.txAncestors(parent.Tx) {
```
Roasbeef: Recursion on transaction inputs is kinda of nitfy due to the fact that the "base case" is actually a built-in consensus rule.

mempool/mempool.go
```diff
@@ -631,6 +743,99 @@ func (mp *TxPool) FetchTransaction(txHash *chainhash.Hash) (*btcutil.Tx, error)
...
+	// exceed the maximum allowed.
+	conflicts := mp.txConflicts(tx)
+	if len(conflicts) > MaxReplacementEvictions {
+		str := fmt.Sprintf("replacement transaction %v evicts more "+
+			"transactions than permitted: max is %v, evicts %v",
+			tx.Hash(), MaxReplacementEvictions, len(conflicts))
+		return nil, txRuleError(wire.RejectNonstandard, str)
+	}
+
+	// The set of conflicts (transactions we'll replace) and ancestors don't
```
Roasbeef: don't -> should not

mempool/mempool.go
```diff
@@ -631,6 +743,99 @@ func (mp *TxPool) FetchTransaction(txHash *chainhash.Hash) (*btcutil.Tx, error)
...
+			continue
+		}
+		str := fmt.Sprintf("replacement transaction %v spends parent "+
+			"transaction %v", tx.Hash(), ancestorHash)
+		return nil, txRuleError(wire.RejectInvalid, str)
+	}
+
+	// The replacement should have a higher fee rate than each of the
+	// conflicting transactions and a higher absolute fee than the sum of
+	// all.
```
Roasbeef: conflicting transactions

mempool/mempool.go
```diff
@@ -631,6 +743,99 @@ func (mp *TxPool) FetchTransaction(txHash *chainhash.Hash) (*btcutil.Tx, error)
...
+			"insufficient absolute fee: needs %v, has %v",
+			tx.Hash(), conflictsFee+minFee, txFee)
+		return nil, txRuleError(wire.RejectInsufficientFee, str)
+	}
+
+	// Finally, It should not spend any new unconfirmed outputs, other than
+	// the ones already included in the parents of the conflicting
+	// transactions it'll replace.
+	for _, txIn := range tx.MsgTx().TxIn {
+		if _, ok := conflictsParents[txIn.PreviousOutPoint.Hash]; ok {
```
Roasbeef: Ah so the unconfirmed input does actually allow post-facto cut-through after all?

wpaulino: Yup! From the BIP:

> One or more transactions currently in the mempool (original transactions) will be replaced by a new transaction (replacement transaction) that spends one or more of the same inputs if,
>
> 2. The replacement transaction may only include an unconfirmed input if that input was included in one of the original transactions. (An unconfirmed input spends an output from a currently-unconfirmed transaction.)

So unconfirmed inputs are restricted to anything it replaces, otherwise any confirmed input is allowed.

mempool/mempool.go
wpaulino: That's fine as long as the other input not signaling replacement isn't introducing a conflict. So if any input introduces a conflict, then it _must_ signal replacement.

wpaulino:
cc @halseth

mempool/mempool.go
```diff
@@ -631,6 +743,99 @@ func (mp *TxPool) FetchTransaction(txHash *chainhash.Hash) (*btcutil.Tx, error)
 	return nil, fmt.Errorf("transaction is not in the pool")
 }
 
+// validateReplacement determines whether a transaction is deemed as a valid
+// replacement of all of its conflicts according to the RBF policy. If it valid,
```
halseth: nit: s/if it/if it is

mempool/mempool.go
```diff
@@ -911,7 +1127,15 @@ func (mp *TxPool) maybeAcceptTransaction(tx *btcutil.Tx, isNew, rateLimit, rejec
...
 
-	// Add to transaction pool.
+	// Now that we've deemed the transaction as valid, we can add it to the
+	// mempool. If it ended up replacing any transactions, we'll remove them
+	// first.
+	for _, conflict := range conflicts {
+		// The conflict set should already include the descendants for
+		// each one, so we don't need to remove the redeemers within
+		// this call as they'll be removed eventually.
+		mp.removeTransaction(conflict, false)
```
halseth: Should we log transactions replaced?

mempool/mempool_test.go
```diff
@@ -866,3 +930,782 @@ func TestCheckSpend(t *testing.T) {
 		t.Fatalf("Unexpeced spend found in pool: %v", spend)
 	}
 }
+
+// TestSignalsReplacement tests that transactions properly signal they can be
```
halseth: awesome set of tests! 👍

mempool/mempool.go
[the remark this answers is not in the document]
wpaulino: Ah yes, good catch! I think this is what @Roasbeef was referring to as well. Updated the test case to check for this as well.

mempool/mempool_test.go
```diff
@@ -866,3 +930,782 @@ func TestCheckSpend(t *testing.T) {
...
+// replaced using RBF.
+func TestSignalsReplacement(t *testing.T) {
```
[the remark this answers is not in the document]
wpaulino: Yeah the commit structure is not the best at the moment. I opted to not have the helper methods in their own commit to not have them introduce unused code.

mempool/mempool.go
```diff
@@ -554,21 +564,123 @@ func (mp *TxPool) addTransaction(utxoView *blockchain.UtxoViewpoint, tx *btcutil
...
+	descendants := make(map[chainhash.Hash]*btcutil.Tx)
+	op := wire.OutPoint{Hash: *tx.Hash()}
+	for i := range tx.MsgTx().TxOut {
+		op.Index = uint32(i)
+		descendant, ok := mp.outpoints[op]
+		if !ok {
+			continue
+		}
+		descendants[*descendant.Hash()] = descendant
+		for hash, descendant := range mp.txDescendants(descendant) {
```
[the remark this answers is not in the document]
wpaulino: I believe this is why `bitcoind` has a restriction of 100 descendants, which they actually track as transactions are added to the pool. IMO, doing so, or even introducing a cache, would complicate this logic a bit more. What if instead we keep the same approach as is now, but limit it to 100 as well?

mempool/mempool.go
wpaulino: Done.

wpaulino:
> I'm wondering, should we make RBF acceptance a command line flag? It could default to enabled.

Done!

mempool/mempool.go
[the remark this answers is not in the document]
halseth: I don't think limiting to 100 would work here, as that is just the max number of evictions? The problem is that the recursion could visit descendants multiple times.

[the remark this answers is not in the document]
wpaulino: Ahh I see what you mean now. So the cache would only live within this method, and not within the `TxPool`?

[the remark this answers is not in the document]
halseth: yes, something like 
```go
func txDescendants(cache, tx) {

        for ... {
              ....
              if cache[descendant] { return }
              else mp.txDescendants(cache, descendant)...
        }
}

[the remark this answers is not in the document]
Comment (wpaulino) [reply mempool/mempool.go]: Done.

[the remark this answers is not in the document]
Comment (halseth) [reply mempool/mempool.go]: think the cache needs to be passed to the `txDescendants` call for it to be effective. Also could be useful with some comments in this method :)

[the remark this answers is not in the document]
Comment (wpaulino) [reply mempool/mempool.go]: Duh, should be fixed now :)

File: mempool/mempool.go [hunk 11]
```diff
@@ -554,21 +569,153 @@ func (mp *TxPool) addTransaction(utxoView *blockchain.UtxoViewpoint, tx *btcutil
...
+func (mp *TxPool) signalsReplacement(tx *btcutil.Tx) bool {
+	for _, txIn := range tx.MsgTx().TxIn {
+		if txIn.Sequence <= MaxRBFSequence {
+			return true
+		}
+		unconfirmedAncestor, ok := mp.pool[txIn.PreviousOutPoint.Hash]
+		if !ok {
+			continue
+		}
+		if mp.signalsReplacement(unconfirmedAncestor.Tx) {
```
halseth: I'm starting to wonder if this has the same potential attack vector as in `txDescendants`. That we can fill the mempool with a tree of unconfirmed transactions, that would make this recursive algorithm exponential in running time. It would be more expensive to do in this case but I think it is still possible. 

Same for `txAncestors`.

mempool/mempool.go
```diff
@@ -554,21 +569,153 @@ func (mp *TxPool) addTransaction(utxoView *blockchain.UtxoViewpoint, tx *btcutil
...
+		moreDescendants, ok := cache[*descendant.Hash()]
+		if !ok {
+			moreDescendants = mp.txDescendants(descendant, cache)
+			cache[*descendant.Hash()] = moreDescendants
+		}
+
+		for _, moreDescendant := range moreDescendants {
+			descendants[*moreDescendant.Hash()] = moreDescendant
+		}
+		descendants[*descendant.Hash()] = descendant
```
halseth: nit: move this to above the descendants-of-descendant recursion.

mempool/mempool.go
```diff
@@ -554,21 +569,153 @@ func (mp *TxPool) addTransaction(utxoView *blockchain.UtxoViewpoint, tx *btcutil
...
+		conflict, ok := mp.outpoints[txIn.PreviousOutPoint]
+		if !ok {
+			continue
+		}
+		conflicts[*conflict.Hash()] = conflict
+		for hash, descendant := range mp.txDescendants(conflict, nil) {
+			conflicts[hash] = descendant
+		}
+	}
+	return conflicts
```
halseth: looks like this will be much more precise than what bitcoind does, since they intentionally overcount.

mempool/mempool.go
```diff
@@ -911,7 +1162,18 @@ func (mp *TxPool) maybeAcceptTransaction(tx *btcutil.Tx, isNew, rateLimit, rejec
 		return nil, nil, err
 	}
 
-	// Add to transaction pool.
+	// Now that we've deemed the transaction as valid, we can add it to the
+	// mempool. If it ended up replacing any transactions, we'll remove them
+	// first.
+	for _, conflict := range conflicts {
+		log.Debugf("Replacing transaction %v with %v", conflict.Hash(),
```
halseth: nit: maybe `"Replacing transaction %v  (feeRate=x) with %v (feeRate=Y)"`

mempool/mempool_test.go
```diff
@@ -866,3 +930,851 @@ func TestCheckSpend(t *testing.T) {
...
+
+				return tx, nil
+			},
+			err: "insufficient absolute fee",
+		},
+		{
+			// A transaction cannot replace another if it introduces
+			// a new unconfirmed input that was not already in any
+			// of the transactions it's directly replacing.
+			name: "spends new unconfirmed input",
```
halseth: This is interesting... Any idea what this rule achieves?

wpaulino: https://github.com/bitcoin/bitcoin/blob/03858b23fe1d17e681eb18d24602b276020295a0/src/validation.cpp#L709

mempool/mempool.go
wpaulino: Done.

mempool/mempool.go
wpaulino: Done.

mempool/mempool.go
wpaulino: `bitcoind` doesn't seem to have this issue since they limit chains of unconfirmed transactions to 25. In any case, it doesn't hurt to cover this case so these are also cached now.

mempool/mempool.go
```diff
@@ -554,21 +569,197 @@ func (mp *TxPool) addTransaction(utxoView *blockchain.UtxoViewpoint, tx *btcutil
...
+		hash := txIn.PreviousOutPoint.Hash
+		unconfirmedAncestor, ok := mp.pool[hash]
+		if !ok {
+			continue
+		}
+
+		// If we've already determined the transaction signals
+		// replacement, we can avoid visiting it again.
+		if _, ok := cache[hash]; ok {
+			continue
```
halseth: should return true?

mempool/mempool.go
```diff
@@ -554,21 +569,197 @@ func (mp *TxPool) addTransaction(utxoView *blockchain.UtxoViewpoint, tx *btcutil
...
+		}
+
+		if mp.signalsReplacement(unconfirmedAncestor.Tx, cache) {
+			cache[hash] = struct{}{}
```
halseth: We should also store a `false` result in the cache, as we might attempt to recalcalculate it as well.

halseth: On more thought, only false results is needed to be cached, as the moment we encounter a `true`, the we return?

wpaulino: Yup, fixed!

mempool/mempool_test.go
```diff
@@ -866,3 +930,853 @@ func TestCheckSpend(t *testing.T) {
...
+
+			// Assert the expected conflicts are returned.
+			txConflicts := ctx.harness.txPool.txConflicts(tx)
+			if len(txConflicts) != len(conflicts) {
+				ctx.t.Fatalf("expected %d conflicts, got %d",
+					len(conflicts), len(txConflicts))
+			}
+			for _, conflict := range conflicts {
+				conflictHash := *conflict.Hash()
+				if _, ok := txConflicts[conflictHash]; !ok {
```
halseth: This is not a totally accurate assertion, since technically `conflicts` could have duplicates.

mempool/mempool.go
```diff
@@ -554,21 +569,200 @@ func (mp *TxPool) addTransaction(utxoView *blockchain.UtxoViewpoint, tx *btcutil
...
+// This function MUST be called with the mempool lock held (for reads).
+func (mp *TxPool) txConflicts(tx *btcutil.Tx) map[chainhash.Hash]*btcutil.Tx {
+	conflicts := make(map[chainhash.Hash]*btcutil.Tx)
+	for _, txIn := range tx.MsgTx().TxIn {
+		conflict, ok := mp.outpoints[txIn.PreviousOutPoint]
+		if !ok {
+			continue
+		}
+		conflicts[*conflict.Hash()] = conflict
+		for hash, descendant := range mp.txDescendants(conflict, nil) {
```
halseth: simple comment right here would be cool

mempool/mempool.go
```diff
@@ -631,6 +825,99 @@ func (mp *TxPool) FetchTransaction(txHash *chainhash.Hash) (*btcutil.Tx, error)
...
+	conflicts := mp.txConflicts(tx)
+	if len(conflicts) > MaxReplacementEvictions {
+		str := fmt.Sprintf("replacement transaction %v evicts more "+
+			"transactions than permitted: max is %v, evicts %v",
+			tx.Hash(), MaxReplacementEvictions, len(conflicts))
+		return nil, txRuleError(wire.RejectNonstandard, str)
+	}
+
+	// The set of conflicts (transactions we'll replace) and ancestors
+	// should not overlap, otherwise the replacement would be spending an
```
halseth: was not obvious that these could overlap, but thinking about it I think that is indeed the case 👍

mempool/mempool.go
```diff
@@ -631,6 +825,99 @@ func (mp *TxPool) FetchTransaction(txHash *chainhash.Hash) (*btcutil.Tx, error)
...
+	}
+
+	// Finally, it should not spend any new unconfirmed outputs, other than
+	// the ones already included in the parents of the conflicting
+	// transactions it'll replace.
+	for _, txIn := range tx.MsgTx().TxIn {
+		if _, ok := conflictsParents[txIn.PreviousOutPoint.Hash]; ok {
+			continue
+		}
+		if _, ok := mp.pool[txIn.PreviousOutPoint.Hash]; !ok {
```
halseth: does this check `!ok` indicate it was a confirmed input? Maybe add a comment.

mempool/mempool.go
```diff
@@ -631,6 +825,99 @@ func (mp *TxPool) FetchTransaction(txHash *chainhash.Hash) (*btcutil.Tx, error)
 	return nil, fmt.Errorf("transaction is not in the pool")
 }
 
+// validateReplacement determines whether a transaction is deemed as a valid
+// replacement of all of its conflicts according to the RBF policy. If it is
+// valid, no error is returned. Otherwise, an error is returned indicating what
+// went wrong.
+//
+// This function MUST be called with the mempool lock held (for reads).
+func (mp *TxPool) validateReplacement(tx *btcutil.Tx,
```
halseth: Does this not have a unit test? Or is it covered by the `TestRBF` cases?

wpaulino: `TestRBF` is the unit test. You can see each test case failing the conditions one by one as you go through them.

mempool/mempool_test.go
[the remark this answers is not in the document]
wpaulino: The bulk of the PR is really just test code and I prefer including those in the same commit as the behavior its testing.

mempool/mempool.go
wpaulino: Already explained in the method's `godoc`.

> Any descendants of these transactions are also considered conflicts as they would no longer exist.

mempool/mempool.go
```diff
@@ -554,21 +569,200 @@ func (mp *TxPool) addTransaction(utxoView *blockchain.UtxoViewpoint, tx *btcutil
...
+	cache map[chainhash.Hash]struct{}) bool {
+
+	// If a cache was not provided, we'll initialize one now to use for the
+	// recursive calls.
+	if cache == nil {
+		cache = make(map[chainhash.Hash]struct{})
+	}
+
+	for _, txIn := range tx.MsgTx().TxIn {
+		if txIn.Sequence <= MaxRBFSequence {
```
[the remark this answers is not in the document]
wpaulino: Already explained in the method's `godoc`.

mempool/mempool_test.go
```diff
@@ -866,3 +930,853 @@ func TestCheckSpend(t *testing.T) {
...
+				outs := []spendableOutput{coinbaseOut1}
+				ctx.addSignedTx(outs, 1, 0, true, false)
+
+				coinbase2 := ctx.addCoinbaseTx(1)
+				coinbaseOut2 := txOutToSpendableOut(coinbase2, 0)
+				outs = []spendableOutput{coinbaseOut2}
+				ctx.addSignedTx(outs, 1, 0, false, false)
+
+				// Create a transaction that spends both
+				// coinbase outputs that were spent above. This
```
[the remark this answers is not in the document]
wpaulino: Spent is correct here. There are two transactions (`conflict1` and `conflict2`) above that spend the coinbase outputs.

mempool/mempool_test.go
wpaulino: It's assumed that `conflicts` won't include duplicates since the `txConflicts` method doesn't as well.

mempool/mempool.go
wpaulino: Done.

mempool/mempool_test.go
```diff
@@ -866,3 +930,853 @@ func TestCheckSpend(t *testing.T) {
...
+				outs = []spendableOutput{coinbaseOut1, coinbaseOut2}
+				tx, err := ctx.harness.CreateSignedTx(
+					outs, 1, 0, false,
+				)
+				if err != nil {
+					ctx.t.Fatalf("unable to create "+
+						"transaction: %v", err)
+				}
+
+				return tx, []*btcutil.Tx{conflict1, conflict2}
```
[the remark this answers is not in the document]
wpaulino: Done.

mempool/mempool_test.go
```diff
@@ -866,3 +930,853 @@ func TestCheckSpend(t *testing.T) {
...
+// processing new incoming transactions.
+func TestConflicts(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name  string
+		setup func(ctx *testContext) (*btcutil.Tx, []*btcutil.Tx)
+	}{
+		{
+			name: "no conflicts",
```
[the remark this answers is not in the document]
wpaulino: Done.

mempool/mempool_test.go
```diff
@@ -866,3 +930,853 @@ func TestCheckSpend(t *testing.T) {
...
+func TestCheckPoolDoubleSpend(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name          string
+		setup         func(ctx *testContext) *btcutil.Tx
+		isReplacement bool
+	}{
+		{
+			name: "no double spend",
```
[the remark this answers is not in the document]
wpaulino: Done.

PR #1428, opened 2019-05-12
btcjson+rpc: match bitcoind's RPC error codes for rejected transactions

wpaulino:
`bitcoind`'s behavior can be referenced [here](https://github.com/bitcoin/bitcoin/blob/master/src/node/transaction.cpp#L34).

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -3323,19 +3323,44 @@ func handleSendRawTransaction(s *rpcServer, cmd interface{}, closeChan <-chan st
 	if err != nil {
 		// When the error is a rule error, it means the transaction was
 		// simply rejected as opposed to something actually going wrong,
-		// so log it as such.  Otherwise, something really did go wrong,
-		// so log it as an actual error.  In both cases, a JSON-RPC
-		// error is returned to the client with the deserialization
-		// error code (to match bitcoind behavior).
-		if _, ok := err.(mempool.RuleError); ok {
-			rpcsLog.Debugf("Rejected transaction %v: %v", tx.Hash(),
-				err)
-		} else {
+		// so log it as such. Otherwise, something really did go wrong,
+		// so log it as an actual error and return.
+		ruleErr, ok := err.(mempool.RuleError)
+		if !ok {
 			rpcsLog.Errorf("Failed to process transaction %v: %v",
 				tx.Hash(), err)
+
+			return nil, &btcjson.RPCError{
+				Code:    btcjson.ErrRPCTxError,
+				Message: "TX rejected: " + err.Error(),
+			}
 		}
+
+		rpcsLog.Debugf("Rejected transaction %v: %v", tx.Hash(), err)
+
+		// We'll then map the rule error to the appropriate RPC error,
+		// matching bitcoind's behavior.
+		code := btcjson.ErrRPCTxError
+		if txRuleErr, ok := ruleErr.Err.(mempool.TxRuleError); ok {
+			errDesc := txRuleErr.Description
+			switch {
+			case strings.Contains(
+				strings.ToLower(errDesc), "orphan transaction",
+			):
+				code = btcjson.ErrRPCTxError
+
+			case strings.Contains(
+				strings.ToLower(errDesc), "transaction already exists",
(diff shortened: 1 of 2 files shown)


rpcserver.go
```diff
@@ -3324,18 +3324,33 @@ func handleSendRawTransaction(s *rpcServer, cmd interface{}, closeChan <-chan st
...
 				err)
 		} else {
 			rpcsLog.Errorf("Failed to process transaction %v: %v",
 				tx.Hash(), err)
 		}
+
+		// We'll then map the rule error to the appropriate RPC error,
+		// matching bitcoind's behavior.
+		code := btcjson.ErrRPCTxError
+		if txRuleErr, ok := ruleErr.Err.(mempool.TxRuleError); ok {
```
halseth: This might not work if `ok := err.(mempool.RuleError); !ok` above.

rpcserver.go
```diff
@@ -3324,18 +3324,33 @@ func handleSendRawTransaction(s *rpcServer, cmd interface{}, closeChan <-chan st
...
+			errDesc := txRuleErr.Description
+			switch {
+			case strings.Contains(errDesc, "orphan transaction"):
```
halseth: can we add error codes to `RuleError` instead of doing string matching here?

wpaulino: No unless we return more structured errors from `TxPool.MaybeAcceptTransaction`. The current errors use the `wire.RejectCode`s, which don't provide enough context as they can be used for different scenarios. For example, `wire.RejectDuplicate` is used for transactions that have already confirmed, transactions that are already in the mempool, and transactions double spending others in the mempool.

rpcserver.go
wpaulino: Nice catch!

rpcserver.go
```diff
@@ -3324,18 +3324,33 @@ func handleSendRawTransaction(s *rpcServer, cmd interface{}, closeChan <-chan st
...
+				code = btcjson.ErrRPCTxError
+			case strings.Contains(errDesc, "transaction already exists"):
+				code = btcjson.ErrRPCTxAlreadyInChain
+			default:
+				code = btcjson.ErrRPCTxRejected
+			}
+		}
+
 		return nil, &btcjson.RPCError{
-			Code:    btcjson.ErrRPCDeserialization,
```
Roasbeef: 🙈

rpcserver.go
Roasbeef: Perhaps we should go ahead and output better errors? Once again we find ourselves string matching here without any existing `btcd` level integration tests for this particular scenario for this call.

rpcserver.go
Roasbeef: Ah guess it was OK in the past since we directly parsed the error string itself. Fixing this now will let us rely solely on the error code in the future though, which is nice.

rpcserver.go
wpaulino: Agreed. Is it worth deferring until we overhaul `btcwallet` with concrete errors? Can also use the newest error additions that will come as part of `go 1.13`, which will make things a bit nicer.

rpcserver.go
```diff
@@ -3324,18 +3324,33 @@ func handleSendRawTransaction(s *rpcServer, cmd interface{}, closeChan <-chan st
...
+
+		// We'll then map the rule error to the appropriate RPC error,
+		// matching bitcoind's behavior.
+		code := btcjson.ErrRPCTxError
+		if txRuleErr, ok := ruleErr.Err.(mempool.TxRuleError); ok {
+			errDesc := txRuleErr.Description
+			switch {
+			case strings.Contains(errDesc, "orphan transaction"):
+				code = btcjson.ErrRPCTxError
+			case strings.Contains(errDesc, "transaction already exists"):
```
cfromknecht: is now a good time to add case insensitive matching?

rpcserver.go
wpaulino: Fixed.

rpcserver.go
wpaulino: Done.

PR #1429, opened 2019-05-15
btcec: optimize square root using fieldVal

cfromknecht:
As of #1193, decompressPoint now
validates that the point is on the curve. The x and y coordinates are
also implicitly <= `P`, since the modular reduction is applied to both
before the method returns. The checks are moved so that they are still
applied when parsing an uncompressed pubkey, as the checks are not
redundant in that path.

We also optimize the decompressPoint subroutine, used in extracting
compressed pubkeys and performing pubkey recovery. We do so by replacing
the use of big.Int.Exp with with square-and-multiply exponentiation of
btcec's more optimized fieldVals, reducing the overall latency and
memory requirements of decompressPoint.

Instead of operating on bits of Q = (P+1)/4, the exponentiation applies
the square-and-multiply operations on full bytes of Q.  Compared to the
original speedup. Compared the bit-wise version, the improvement is
roughly 10%.

A new pair fieldVal methods called Sqrt and SqrtVal are added, which
applies the square-and-multiply exponentiation using precomputed
byte-slice of the value Q.

Comparison against big.Int sqrt and SAM sqrt over bytes of Q:

```
benchmark                            old ns/op     new ns/op     delta
BenchmarkParseCompressedPubKey-8     35545         23119         -34.96%

benchmark                            old allocs     new allocs     delta
BenchmarkParseCompressedPubKey-8     35             6            -82.86%

benchmark                            old bytes     new bytes     delta
BenchmarkParseCompressedPubKey-8     2777          256           -90.78%
```

Finally the `*KoblitzCurve.QPlus1Div4()` method is deprecated, and superseded
by `*KoblitzCurve.Q()`, to convey that the returned value is `Q = (P+1)/4`, and
not `(((P+1)/4)+1)/4) `.

### DIFF
--- a/btcec/btcec.go
+++ b/btcec/btcec.go
@@ -36,9 +36,16 @@ var (
 // interface from crypto/elliptic.
 type KoblitzCurve struct {
 	*elliptic.CurveParams
-	q         *big.Int
+
+	// q is the value (P+1)/4 used to compute the square root of field
+	// elements.
+	q *big.Int
+
 	H         int      // cofactor of the curve.
 	halfOrder *big.Int // half the order N
+
+	// fieldB is the constant B of the curve as a fieldVal.
+	fieldB *fieldVal
 
 	// byteSize is simply the bit size / 8 and is provided for convenience
 	// since it is calculated repeatedly.
@@ -917,6 +924,7 @@ func initS256() {
 		big.NewInt(1)), big.NewInt(4))
 	secp256k1.H = 1
 	secp256k1.halfOrder = new(big.Int).Rsh(secp256k1.N, 1)
+	secp256k1.fieldB = new(fieldVal).SetByteSlice(secp256k1.B.Bytes())
 
 	// Provided for convenience since this gets computed repeatedly.
 	secp256k1.byteSize = secp256k1.BitSize / 8
--- a/btcec/btcec.go
+++ b/btcec/btcec.go
@@ -886,9 +886,19 @@ func (curve *KoblitzCurve) ScalarBaseMult(k []byte) (*big.Int, *big.Int) {
 	return curve.fieldJacobianToBigAffine(qx, qy, qz)
 }
 
-// QPlus1Div4 returns the Q+1/4 constant for the curve for use in calculating
-// square roots via exponention.
+// QPlus1Div4 returns the (P+1)/4 constant for the curve for use in calculating
+// square roots via exponentiation.
+//
+// DEPRECATED: The actual value returned is (P+1)/4, where as the original
+// method name implies that this value is (((P+1)/4)+1)/4. This method is kept
+// to maintain backwards compatibility of the API. Use Q() instead.
 func (curve *KoblitzCurve) QPlus1Div4() *big.Int {
+	return curve.q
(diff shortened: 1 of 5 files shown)


btcec/btcec.go
```diff
@@ -36,10 +36,21 @@ var (
 // interface from crypto/elliptic.
 type KoblitzCurve struct {
 	*elliptic.CurveParams
-	q         *big.Int
+
+	// q is the value (P+1)/4 use to compute the square root of field
```
yashbhutwala:
```suggestion
	// q is the value (P+1)/4, used to compute the square root of field
```

btcec/btcec.go
```diff
@@ -879,12 +890,33 @@ func (curve *KoblitzCurve) ScalarBaseMult(k []byte) (*big.Int, *big.Int) {
 	return curve.fieldJacobianToBigAffine(qx, qy, qz)
 }
 
-// QPlus1Div4 returns the Q+1/4 constant for the curve for use in calculating
-// square roots via exponention.
+// QPlus1Div4 returns the (P+1)/4 constant for the curve for use in calculating
+// square roots via exponentiation.
+//
+// DEPRECATED: The actual value returned is (P+1)/4, where as the original
+// method name implies that this value is ((P+1)/4)+1)/4. This method is kept to
```
yashbhutwala:
```suggestion
// method name implies that this value is (((P+1)/4)+1)/4. This method is kept to
```

btcec/pubkey.go
```diff
@@ -22,41 +22,139 @@ func isOdd(a *big.Int) bool {
...
+	//
+	//   Q = 3fffffffffffffffffffffffffffffffffffffffffffffffffffffffbfffff0c
+	//
+	// We can unroll the typical bit-wise interpretation of the
+	// exponentiation algorithm above to instead operate on bytes.
+	// This reduces the number of comparisons from 256*3 = 768 in the
+	// prior bit-wise algorithm, down to (32*2)+1+2+2 = 69, reducing the
+	// overhead of failed branch predictions and additional comparisons.
+	//
+	// Since the there are only 4 unique bytes of Q, this keeps the jump
```
yashbhutwala:
```suggestion
	// Since there are only 4 unique bytes of Q, this keeps the jump
```

cfromknecht:
thanks @yashbhutwala, comments addressed :)

btcec/field.go
```diff
@@ -1221,3 +1235,118 @@ func (f *fieldVal) Inverse() *fieldVal {
...
+	//
+	// The final value r_0 = x^n.
+	//
+	// See https://en.wikipedia.org/wiki/Exponentiation_by_squaring for more
+	// details.
+	//
+	// This can be further optimized, by observing that the value of Q in
+	// secp256k1 has the value:
+	//
+	//   Q = 3fffffffffffffffffffffffffffffffffffffffffffffffffffffffbfffff0c
```
Roasbeef: This is dope! Can really see how this presents the savings compared to an earlier version that _did_ do the bitwise comparison. Nice having a constant which has such a large Hamming weight.

PR #1431, opened 2019-05-31
txscript: handle variable length P2PKH signatures in ComputePkScript

wpaulino:
Since P2PKH signatures have variable lengths, we would attempt to parse P2PKH scripts as P2SH if they didn't fit the previous length constraints.


txscript/pkscript.go
```diff
@@ -13,13 +13,21 @@ import (
...
-	//   0x47 or 0x48 (71 or 72 byte data push) | <71 or 72 byte sig> |
-	//   0x21 (33 byte data push) | <33 byte compressed pubkey>
-	pubKeyHashSigScriptLen = 106
+	// minPubKeyHashSigScriptLen is the minimum length of a signature script
+	// that spends a P2PKH output. The length is composed of the following:
+	//   Signature length (1 byte)
+	//   Signature without variable R and S values (73 - 33 - 33 bytes)
+	//   Public key length (1 byte)
+	//   Public key (33 byte)
+	minPubKeyHashSigScriptLen = 1 + (73 - 33 - 33) + 1 + 33
```
Roasbeef: `(73 - 33 - 33)`? So just `7`? That isn't a valid signature length.

Roasbeef: In place, I think you just want `71` here.

wpaulino: I've seen cases where the R and/or S values of the signature end up being less than the usual 32 bytes, so the lower end of the range is meant to work around this.

cfromknecht: minimum signature length is 8: https://github.com/btcsuite/btcd/blob/master/btcec/signature.go#L91

wpaulino: Ah, that's exactly what I was looking for!

wpaulino: Fixed.

PR #1444, opened 2019-06-28
 rpc: skip rescan if client has no addresses or UTXOs

cfromknecht:
In this commit, we implement an optimization that will speed up clients
that attempt to perform a rescan in the past with no addresses. If the
client doesn't have any thing to search for, then we simply exit early
and send them a rescan finished notification with the final block in our
chain.

### DIFF
--- a/rpcwebsocket.go
+++ b/rpcwebsocket.go
@@ -2588,6 +2588,11 @@ func handleRescan(wsc *wsClient, icmd interface{}) (interface{}, error) {
 		}
 	}
 
+	var (
+		lastBlock     *btcutil.Block
+		lastBlockHash *chainhash.Hash
+	)
+	if len(lookups.addrs) != 0 || len(lookups.unspent) != 0 {
 		// With all the arguments parsed, we'll execute our chunked rescan
 		// which will notify the clients of any address deposits or output
 		// spends.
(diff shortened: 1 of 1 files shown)


rpcwebsocket.go
```diff
@@ -2321,85 +2321,20 @@ func descendantBlock(prevHash *chainhash.Hash, curBlock *btcutil.Block) error {
...
-			}
-		}
-	}
+// scanBlockChunks executes a rescan in chunked stages. We do this to limit the
+// amount of memory that we'll allocate to a given rescan. Every so often,
+// we'll send back a rescan progress notification to the websockets client. The
+// final block and block hash that we've scanned will be returned.
+func scanBlockChunks(wsc *wsClient, cmd *btcjson.RescanCmd, lookups *rescanKeys, minBlock,
+	maxBlock int32, chain *blockchain.BlockChain) (
+	*btcutil.Block, *chainhash.Hash, error) {
```
wpaulino: Nit: don't see the point in returning the block hash as well as you can derive it from the block.

rpcwebsocket.go
```diff
@@ -2565,23 +2503,112 @@ fetchRange:
...
+		maxBlock, err = chain.BlockHeightByHash(maxBlockHash)
+		if err != nil {
+			return nil, &btcjson.RPCError{
+				Code:    btcjson.ErrRPCBlockNotFound,
+				Message: "Error getting block: " + err.Error(),
+			}
+		}
+	}
+
+		// With all the arguments parsed, we'll execute our chunked rescan
```
wpaulino: Nit: weird indentation?

PR #1451, opened 2019-07-16
Added ListSinceBlockMinConfWatchOnly method.

nalcheg:
Added method for request ListSinceBlock with include_watchonly=true

### DIFF
--- a/rpcclient/wallet.go
+++ b/rpcclient/wallet.go
@@ -326,6 +326,7 @@ func (c *Client) ListSinceBlockAsync(blockHash *chainhash.Hash) FutureListSinceB
 // minimum confirmations as a filter.
 //
 // See ListSinceBlockMinConf to override the minimum number of confirmations.
+// See ListSinceBlockMinConfWatchOnly to override the minimum number of confirmations and watch only parameter.
 func (c *Client) ListSinceBlock(blockHash *chainhash.Hash) (*btcjson.ListSinceBlockResult, error) {
 	return c.ListSinceBlockAsync(blockHash).Receive()
 }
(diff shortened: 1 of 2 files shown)


jakesyl:
@jcvernaleo (as per #1530)

- Low priority
- Enhancement

rpcclient/wallet.go
```diff
@@ -312,6 +313,30 @@ func (c *Client) ListSinceBlockMinConf(blockHash *chainhash.Hash, minConfirms in
...
+// ListSinceBlockMinConfWatchOnlyAsync returns an instance of a type that can be used to
+// get the result of the RPC at some future time by invoking the Receive
+// function on the returned instance.
+//
+// See ListSinceBlockMinConfWatchOnly for the blocking version and more details.
+func (c *Client) ListSinceBlockMinConfWatchOnlyAsync(blockHash *chainhash.Hash, minConfirms int, watchOnly bool) FutureListSinceBlockResult {
+	var hash *string
+	if blockHash != nil {
+		hash = btcjson.String(blockHash.String())
+	}
```
onyb, line 325: This probably won't work if you want to do something like _list all watch-only transactions_ with the following call:
```
ListSinceBlockMinConfWatchOnlyAsync(nil, 1, true)
```

Here's a workaround, although I'm not sure if it's the best solution.

```suggestion
	} else {
	        hash = btcjson.String("")
	}
```

nalcheg:
did in https://github.com/btcsuite/btcd/pull/1163

onyb:
@nalcheg #1163 is not doing the same thing. You added `ListSinceBlockMinConfWatchOnlyAsync`. That one was about `ListTransactionsCountFromWatchOnlyAsync`.

nalcheg:
sorry, I was wrong

rpcclient/wallet.go
onyb: Thanks to https://github.com/btcsuite/btcd/pull/1594, we may not need to do this. Could you please verify this on your side?

onyb (changes requested):
Needs a rebase + the small test I suggested.

onyb (changes requested):
Could you please add the following unit-test to `walletsvrcmds_test.go`? (line 710)

```go
		{
			name: "listsinceblock pad null",
			newCmd: func() (interface{}, error) {
				return btcjson.NewCmd("listsinceblock", "null", 1, false)
			},
			staticCmd: func() interface{} {
				return btcjson.NewListSinceBlockCmd(nil, btcjson.Int(1), btcjson.Bool(false))
			},
			marshalled: `{"jsonrpc":"1.0","method":"listsinceblock","params":[null,1,false],"id":1}`,
			unmarshalled: &btcjson.ListSinceBlockCmd{
				BlockHash:           nil,
				TargetConfirmations: btcjson.Int(1),
				IncludeWatchOnly:    btcjson.Bool(false),
			},
		},
```

onyb (approved):
Looks great! Thanks for your contribution.

jcvernaleo (approved):
OK

PR #1453, opened 2019-07-19
multi: add gettxoutproof and verifytxoutproof RPCs

wpaulino:
In this PR, we extend btcd's RPC interface with support for `gettxoutproof` and `verifytxoutproof`. These RPCs are implemented so as to mimic [bitcoind's behavior](https://github.com/bitcoin/bitcoin/blob/536590f358dc3d3e5821eba7f1009452ea93b205/src/rpc/rawtransaction.cpp#L214) as much as possible, in order to prevent callers from special casing which backend is currently in use.

Depends on https://github.com/btcsuite/btcutil/pull/146 for proof verification.


jakesyl:
@wpaulino you mind updating the dependencies here?

jakesyl:
@jcvernaleo (as per #1530)

- Low priority
- Outdated

PR #1455, opened 2019-07-23
rpc server cert add host ip by cfg.Listeners

5dao:
rpc server cert olny contain local interface ip , but the cloud host most use NAT network , rpc server cert should add ip by cfg.Listeners


bjornoj:
LGTM

jakesyl:
@jcvernaleo (as per #1530)

- low priority
- enhancement

server.go
```diff
@@ -2531,7 +2510,7 @@ func setupRPCListeners() ([]net.Listener, error) {
 		// Generate the TLS cert and key file if both don't already
 		// exist.
 		if !fileExists(cfg.RPCKey) && !fileExists(cfg.RPCCert) {
-			err := genCertPair(cfg.RPCCert, cfg.RPCKey)
+			err := genCertPair(cfg.RPCCert, cfg.RPCKey, cfg.Listeners)
```
allocz: I don't understand how this is useful, as certificate generation adds all interface addresses. If you add a listener address that is not bound to an interface, btcd will fail to listen in the specified address and the node won't even start.

Issue #1054, opened 2017-09-14
rpcclient: Support cookie authentication


JeremyRand:
Bitcoin Core's RPC server allows clients to authenticate via a cookie file; it would be useful for rpcclient to support this authentication mode.

PR #1460, opened 2019-08-02
rpcclient: Add cookie auth

JeremyRand:
Fixes https://github.com/btcsuite/btcd/issues/1054

### DIFF
--- a/rpcclient/cookiefile.go
+++ b/rpcclient/cookiefile.go
@@ -1,0 +1,64 @@
+// Copyright (c) 2017 The Namecoin developers
+// Copyright (c) 2019 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package rpcclient
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+	"time"
+)
+
+func readCookieFile(path string) (username, password string, err error) {
+	b, err := ioutil.ReadFile(path)
+	if err != nil {
+		return
+	}
+
+	s := strings.TrimSpace(string(b))
+	parts := strings.SplitN(s, ":", 2)
+	if len(parts) != 2 {
+		err = fmt.Errorf("malformed cookie file")
+		return
+	}
+
+	username, password = parts[0], parts[1]
+	return
+}
+
+func cookieRetriever(path string) func() (username, password string, err error) {
+	lastCheckTime := time.Time{}
+	lastModTime := time.Time{}
+
+	curUsername, curPassword := "", ""
+	var curError error
+
+	doUpdate := func() {
+		if !lastCheckTime.IsZero() && time.Now().Before(lastCheckTime.Add(30*time.Second)) {
+			return
+		}
+
+		lastCheckTime = time.Now()
+
+		st, err := os.Stat(path)
+		if err != nil {
+			curError = err
+			return
+		}
+
+		modTime := st.ModTime()
+		if !modTime.Equal(lastModTime) {
+			lastModTime = modTime
+			curUsername, curPassword, curError = readCookieFile(path)
+		}
+	}
+
+	return func() (username, password string, err error) {
+		doUpdate()
+		return curUsername, curPassword, curError
+	}
+}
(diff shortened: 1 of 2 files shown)


rpcclient/infrastructure.go
```diff
@@ -1113,6 +1127,25 @@ type ConnConfig struct {
 	EnableBCInfoHacks bool
 }
 
+// getAuth returns the username and passphrase that will actually be used for
+// this connection.  This will be the result of checking the cookie if a cookie
+// path is configured; if not, it will be the user-configured username and
+// passphrase.
+func (config *ConnConfig) getAuth() (username, passphrase string, err error) {
```
Rjected: Bitcoin Core falls back to cookie authentication if a password is not found, we might want to replicate that functionality rather than fall back to username/password authentication if a cookie is not found.

Additionally, there is no example of cookie creation - this is a relevant feature, and it is currently undefined.

rpcclient/cookiefile.go
```diff
@@ -0,0 +1,64 @@
...
+	"time"
+)
+
+func readCookieFile(path string) (username, password string, err error) {
+	b, err := ioutil.ReadFile(path)
+	if err != nil {
+		return
+	}
+
+	s := strings.TrimSpace(string(b))
```
Rjected: Considering Core uses base64 encoding for their client cookie authentication, those libraries could be used instead of a binary => string cast

JeremyRand:
Thanks for the feedback @Rjected , I'll address the feedback ASAP.  Things are a bit busy here at the moment, so I might take a little while.

rpcclient/infrastructure.go
JeremyRand: > Additionally, there is no example of cookie creation - this is a relevant feature, and it is currently undefined.

Hmm, what functionality are you looking for exactly?  Making the btcd daemon support creating a cookie file?

rpcclient/cookiefile.go
JeremyRand: You're talking about https://github.com/bitcoin/bitcoin/blob/f4a0d27e85754d60804ffa36e415b67c263180b9/src/bitcoin-cli.cpp#L345 , right?  AFAICT from looking at the Bitcoin Core source code, while it's true that the Core cookie reader doesn't try to parse out a separate username and password (and instead simply submits the line extracted from the cookie file verbatim), the code that actually creates the cookie file does distinguish between the username and password components.

I think it's generally a good thing to try to parse out the username and password separately, for 2 reasons:

1. It simplifies the integration with existing code that expects authentication to use a username+password rather than a combined blob.
2. It will better detect cases where the cookie file doesn't look like a standard cookie file, which are probably an indication of some kind of error.

JeremyRand:
Updated PR to address feedback.

rpcclient/infrastructure.go
Rjected: Yup - something like in https://github.com/bitcoin/bitcoin/blob/master/src/rpc/request.cpp
Just so users don't have to generate using core first, and this can be a little more feature complete.

rpcclient/cookiefile.go
Rjected: That makes sense, I'm cool with the user/pass separation. I still have an issue with this line though - why are the spaces being trimmed? I just added a space in a cookie file with Core and it worked.
Parsing only the first line of the file makes sense, and it is what core does in https://github.com/bitcoin/bitcoin/blob/42d0eca725a83c999e2b67e33dfc7bcc96288dc3/src/rpc/request.cpp#L108, but spaces in general seem a little arbitrary

rpcclient/infrastructure.go
Rjected: Out of scope for this PR though

rpcclient/cookiefile.go
JeremyRand: Replaced `TrimSpace` with reading the first line.

torkelrogstad:
Tested this locally worked like a charm. 

LGTM

jakesyl:
@jcvernaleo (as per #1530)

- Low priority
- Enhancement

Well tested, mergable with minimal review

JeremyRand:
Rebased and fixed merge conflicts.

PR #1462, opened 2019-08-20
peer: include verack as part of version handshake

wpaulino:
It was possible for connections to bitcoind nodes to be closed from their side if the OnVersion callback queued a message to send. This is because bitcoind expects a verack message before attempting to process any other messages. To address this, we ensure the verack is sent as part of the handshake process, such that any queued messages happen after the fact.

### DIFF
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -1377,20 +1377,12 @@ out:
 			break out
 
 		case *wire.MsgVerAck:
-
-			// No read lock is necessary because verAckReceived is not written
-			// to in any other goroutine.
-			if p.verAckReceived {
-				log.Infof("Already received 'verack' from peer %v -- "+
-					"disconnecting", p)
-				break out
-			}
-			p.flagsMtx.Lock()
-			p.verAckReceived = true
-			p.flagsMtx.Unlock()
-			if p.cfg.Listeners.OnVerAck != nil {
-				p.cfg.Listeners.OnVerAck(p, msg)
-			}
+			// Limit to one verack message per peer.
+			p.PushRejectMsg(
+				msg.Command(), wire.RejectDuplicate,
+				"duplicate verack message", nil, true,
+			)
+			break out
 
 		case *wire.MsgGetAddr:
 			if p.cfg.Listeners.OnGetAddr != nil {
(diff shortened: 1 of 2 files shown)


peer/peer.go
```diff
@@ -2043,29 +2069,60 @@ func (p *Peer) writeLocalVersionMsg() error {
 		return err
 	}
 
-	return p.writeMessage(localVerMsg, wire.LatestEncoding)
+	if err := p.writeMessage(localVerMsg, wire.LatestEncoding); err != nil {
```
Roasbeef: Unnecessary change.

peer/peer.go
```diff
@@ -2043,29 +2069,60 @@ func (p *Peer) writeLocalVersionMsg() error {
...
-// then sends our version message. If the events do not occur in that order then
-// it returns an error.
+// negotiateInboundProtocol performs the negotiation protocol for an inbound
+// peer. The events should occur in the following order, otherwise an error is
+// returned:
+//
+//   1. Remote peer sends their version.
+//   2. We send our version.
+//   3. We send our verack.
+//   4. Remote peer sends their verack.
```
Roasbeef: We may want to add some grace to this ordering here, to ensure that both sides don't deadlock if we're connecting to an implementation with a slightly different ordering.

peer/peer.go
```diff
@@ -1974,6 +1966,40 @@ func (p *Peer) readRemoteVersionMsg() error {
...
+
+	// It should be a verack message, otherwise send a reject message to the
+	// peer explaining why.
+	msg, ok := remoteMsg.(*wire.MsgVerAck)
+	if !ok {
+		reason := "a verack message must follow version"
+		rejectMsg := wire.NewMsgReject(
+			msg.Command(), wire.RejectMalformed, reason,
+		)
+		_ = p.writeMessage(rejectMsg, wire.LatestEncoding)
```
Roasbeef: We may want to log this error at the very least.

wpaulino: The error will propagate back to `peer.AssociateConnection` which will log it.

peer/peer.go
wpaulino: If they send a message in the wrong order, then an error would be returned which would cause us to disconnect them. If they don't sent a message we expect, there's a timeout that'll also cause us to disconnect them. I believe that should cover everything?

peer/peer.go
wpaulino: Fixed.

peer/peer.go
Roasbeef: Yeah it's also already the case that we have an expected order as is.

PR #1463, opened 2019-08-27
blockchain: remove unknown block version warning

Rjected:
Received this warning while syncing, and this warning is being called when >50 of the last 100 blocks have unexpected version numbers. Due to AsicBoost this is triggering many false positives.
The warning was removed from Core for 0.18.0 in https://github.com/bitcoin/bitcoin/pull/15471/.


jakesyl:
@jcvernaleo (as per #1530)

- Low priority

PR #1465, opened 2019-09-05
Add Dockerfile to build and run btcd on Docker.

legacycode:
Hi team,

i created the following Dockerfile that builds and runs btcd in a small (55MB) Alpine container. The /root/.btcd folder is mounted to a named docker volume by default. Any additional btcd argument can be passed by adding it at the end of the docker command. Very simple.

Please check the comments in the Dockerfile on how to build and run the container. It would be great to see it in the official repository and to have an official container on Docker Hub.

Best regards

/legacycode

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,40 @@
+# This Dockerfile builds btcd from source and creates a small (55 MB) docker container based on alpine linux.
+#
+# Clone this repository and run the following command to build and tag a fresh btcd amd64 container:
+#
+# docker build . -t yourregistry/btcd
+#
+# You can use the following command to buid an arm64v8 container:
+#
+# docker build . -t yourregistry/btcd --build-arg ARCH=arm64v8
+#
+# For more information how to use this docker image visit:
+# https://github.com/btcsuite/btcd/tree/master/docs
+#
+# 8333  Mainnet Bitcoin peer-to-peer port
+# 8334  Mainet RPC port
+
+ARG ARCH=amd64
+
+FROM golang:1.14-alpine3.12 AS build-container
+
+ARG ARCH
+ENV GO111MODULE=on
+
+ADD . /app
+WORKDIR /app
+RUN set -ex \
+  && if [ "${ARCH}" = "amd64" ]; then export GOARCH=amd64; fi \
+  && if [ "${ARCH}" = "arm64v8" ]; then export GOARCH=arm64; fi \
+  && echo "Compiling for $GOARCH" \
+  && go install -v . ./cmd/...
+
+FROM $ARCH/alpine:3.12
+
+COPY --from=build-container /go/bin /bin
+
+VOLUME ["/root/.btcd"]
+
+EXPOSE 8333 8334
+
+ENTRYPOINT ["btcd"]
--- a/docs/using_docker.md
+++ b/docs/using_docker.md
@@ -1,0 +1,160 @@
+# Using Docker
+
+- [Using Docker](#using-docker)
+  - [Introduction](#introduction)
+  - [Docker volumes](#docker-volumes)
+  - [Known error messages when starting the btcd container](#known-error-messages-when-starting-the-btcd-container)
+  - [Examples](#examples)
+    - [Preamble](#preamble)
+    - [Full node without RPC port](#full-node-without-rpc-port)
+    - [Full node with RPC port](#full-node-with-rpc-port)
+    - [Full node with RPC port running on TESTNET](#full-node-with-rpc-port-running-on-testnet)
+
+## Introduction
+
+With Docker you can easily set up *btcd* to run your Bitcoin full node. You can find the official *btcd* Docker images on Docker Hub [btcsuite/btcd](https://hub.docker.com/r/btcsuite/btcd). The Docker source file of this image is located at [Dockerfile](https://github.com/btcsuite/btcd/blob/master/Dockerfile).
+
+This documentation focuses on running Docker container with *docker-compose.yml* files. These files are better to read and you can use them as a template for your own use. For more information about Docker and Docker compose visit the official [Docker documentation](https://docs.docker.com/).
+
+## Docker volumes
+
+**Special diskspace hint**: The following examples are using a Docker managed volume. The volume is named *btcd-data* This will use a lot of disk space, because it contains the full Bitcoin blockchain. Please make yourself familiar with [Docker volumes](https://docs.docker.com/storage/volumes/).
+
+The *btcd-data* volume will be reused, if you upgrade your *docker-compose.yml* file. Keep in mind, that it is not automatically removed by Docker, if you delete the btcd container. If you don't need the volume anymore, please delete it manually with the command:
+
+```bash
+docker volume ls
+docker volume rm btcd-data
+```
+
+For binding a local folder to your *btcd* container please read the [Docker documentation](https://docs.docker.com/). The preferred way is to use a Docker managed volume.
+
+## Known error messages when starting the btcd container
+
+We pass all needed arguments to *btcd* as command line parameters in our *docker-compose.yml* file. It doesn't make sense to create a *btcd.conf* file. This would make things too complicated. Anyhow *btcd* will complain with following log messages when starting. These messages can be ignored:
+
+```bash
+Error creating a default config file: open /sample-btcd.conf: no such file or directory
+...
+[WRN] BTCD: open /root/.btcd/btcd.conf: no such file or directory
+```
+
+## Examples
+
+### Preamble
+
+All following examples uses some defaults:
+
+- container_name: btcd
+  Name of the docker container that is be shown by e.g. ```docker ps -a```
+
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+RUN go install -v . ./cmd/...
+
+
+FROM alpine:3.10
+
+COPY --from=build-container /go/bin /bin
+
+VOLUME ["/root/.btcd"]
+
+EXPOSE 8333 8334 18333 18334
```
Rjected: Regtest & simnet would be super useful too!
Regtest P2P: 18444
Regtest RPC: 18334 (already added though)
Simnet P2P: 18555
Simnet RPC: 18556

Rjected: Sorry, actually regtest RPC (for bitcoin core) is 18443 - the `rpcPort` in `params.go` causes btcd to use 18334 instead of 18443

legacycode: I removed the testnet ports from default expose. It is not necessary to define all possible ports in the docker file. The dockerfile should only contain the minimal port requirements. It is possible to expose any other port in the docker-compose.yml file.

I also added some documentation with examples. I think that is a good start of point.

docs/using_docker.md
```diff
@@ -0,0 +1,160 @@
...
+  btcd:
+    container_name: btcd
+    hostname: btcd
+    image: btcsuite/btcd:latest
+    restart: unless-stopped
+    volumes:
+      - btcd-data:/root/.btcd
+    ports:
+      - 18883:18883
+      - 18332:18332
```
Rjected: same thing here

docs/using_docker.md
```diff
@@ -0,0 +1,160 @@
...
+  btcd:
+    container_name: btcd
+    hostname: btcd
+    image: btcsuite/btcd:latest
+    restart: unless-stopped
+    volumes:
+      - btcd-data:/root/.btcd
+    ports:
+      - 8883:8883
+      - 8332:8332
```
Rjected: nitpicks:
s/8883/8333
s/8332/8334

docs/using_docker.md
```diff
@@ -0,0 +1,160 @@
...
+    ports:
+      - 8883:8883
+
+volumes:
+  btcd-data:
+```
+
+### Full node with RPC port
+
+To use the RPC port of *btcd* you need to specify a *username* and a very strong *password*. If you want to connect to the RPC port from the internet, you need to expose port 8332(RPC) as well.
```
Rjected: nitpick: s/8332/8334

docs/using_docker.md
```diff
@@ -0,0 +1,160 @@
...
+services:
+  btcd:
+    container_name: btcd
+    hostname: btcd
+    image: btcsuite/btcd:latest
+    restart: unless-stopped
+    volumes:
+      - btcd-data:/root/.btcd
+    ports:
+      - 8883:8883
```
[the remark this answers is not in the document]
legacycode: You are absolutely right. Thanks for your review. Great work, man! Please double check my changes. With the port numbers i am absolutly unsure :-)

docs/using_docker.md
```diff
@@ -0,0 +1,160 @@
...
+        "--rpcpass=[CREATE_A_VERY_HARD_PASSWORD]"
+    ]
+
+volumes:
+  btcd-data:
+```
+
+### Full node with RPC port running on TESTNET
+
+To run a node on testnet, you need to provide the *--testnet* argument. The ports for testnet are 18883 (p2p) and 18332 (RPC):
```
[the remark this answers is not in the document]
legacycode: fixed

docs/using_docker.md
legacycode: fixed

docs/using_docker.md
legacycode: fixed

docs/using_docker.md
legacycode: fixed

docs/using_docker.md
```diff
@@ -0,0 +1,160 @@
...
+
+```docker logs btcd``` (shows the log )
+
+```docker-compose help``` (brings up some helpful information)
+
+### Full node without RPC port
+
+Let's start with an easy example. If you just want to create a full node without the need of using the RPC port, you can use the following example. This example will launch *btcd* and exposes only the default p2p port 8333 to the outside world:
+
+```dockerfile
```
olljanat: nit *yaml* would make this more readable.

docs/using_docker.md
```diff
@@ -0,0 +1,160 @@
...
+
+volumes:
+  btcd-data:
+```
+
+### Full node with RPC port
+
+To use the RPC port of *btcd* you need to specify a *username* and a very strong *password*. If you want to connect to the RPC port from the internet, you need to expose port 8334(RPC) as well.
+
+```dockerfile
```
olljanat: nit yaml

docs/using_docker.md
```diff
@@ -0,0 +1,160 @@
...
+
+volumes:
+  btcd-data:
+```
+
+### Full node with RPC port running on TESTNET
+
+To run a node on testnet, you need to provide the *--testnet* argument. The ports for testnet are 18333 (p2p) and 18334 (RPC):
+
+```dockerfile
```
olljanat: nit ```yaml```

olljanat:
> Hi team,
> 
> i created the following Dockerfile that builds and runs btcd in a small (55MB) Alpine container. The /root/.btcd folder is mounted to a named docker volume by default. Any additional btcd argument can be passed by adding it at the end of the docker command. Very simple.
> 
> Please check the comments in the Dockerfile on how to build and run the container. It would be great to see it in the official repository and to have an official container on Docker Hub.
> 
> Best regards
> 
> /legacycode

olljanat:
> It would be great to see it in the official repository and to have an official container on Docker Hub.

@legacycode yes it would and it would be nice to get those for ARM too. That why I think that Dockerfile should be updated to allow build for other platforms too. You can see from https://github.com/ruimarinho/docker-bitcoin-core/pull/90 how I implemented that for bitcoind.

On this one you basically need run go install commands with `GOARCH` value on first stage and then on second stage use `FROM $ARCH/alpine:3.10`

I also added some small nits about compose file styling on docs as those are yaml files. Not Dockerfile.

docs/using_docker.md
legacycode: fixed

docs/using_docker.md
legacycode: fixed

docs/using_docker.md
legacycode: fixed

legacycode:
@olljanat i implemented your suggestions and squashed my PR into one commit.

jakesyl:
@jcvernaleo (as per #1530)

- low priority
- enhancement

aureleoules:
I would recommend not using root as the user.

```dockerfile
RUN groupadd -g 999 btcd && useradd -r -u 999 -g btcd btcd
USER btcd
```

torkelrogstad:
> I would recommend not using root as the user.

Is there any real benefit to this? One significant downside is that `docker exec btcd-container <rpc-command-here>` does not work, instead you have to do `docker exec --user btcd btcd-container <rpc-command-here>`

aureleoules:
@torkelrogstad i realized this is not a priority.
Though, wouldn't it be better to mount the data volume as non root? Especially for volumes mounted from the host machine (instead of named volumes).

torkelrogstad:
Which user the container runs as does not matter for access to the volume from the outside. All access still has to go through the Docker daemon, which enforces that no on else can read it

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+ADD . /app
+WORKDIR /app
+RUN set -ex \
+  && if [ "${ARCH}" = "amd64" ]; then export GOARCH=amd64; fi \
+  && if [ "${ARCH}" = "arm64v8" ]; then export GOARCH=arm64; fi \
+  && echo "Compiling for $GOARCH" \
+  && go install -v . ./cmd/...
+
+FROM $ARCH/alpine:3.10
```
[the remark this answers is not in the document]
legacycode: I think it is a good practice to specify a image version. I have seen a lot of Docker files that did not build, because the image version uses `latest`. I will bump this version to 3.12 like the build container.

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+#
+# For more information how to use this docker image visit:
+# https://github.com/btcsuite/btcd/tree/master/docs
+#
+# 8333  Mainnet Bitcoin peer-to-peer port
+# 8334  Mainet RPC port
+
+ARG ARCH=amd64
+
+FROM golang:1.14-alpine3.12 AS build-container
```
jakesyl, line 19: This can be 1.15, but not worth holding this up over

jakesyl:
As an aside, would be nice to include a docker image in future releases, but again outside of scope. LGTM

PR #1467, opened 2019-09-07
rpcclient: Add net params to Client (fix 

Rjected:
Adds a chaincfg.Params to rpcclient.Client to allow for net-specific RPC commands to be called.

Removes a multiple `type ()` definition for consistency.

### DIFF
--- a/rpcclient/infrastructure.go
+++ b/rpcclient/infrastructure.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcd/btcjson"
+	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/go-socks/socks"
 	"github.com/btcsuite/websocket"
 )
@@ -120,6 +121,10 @@ type Client struct {
 
 	// config holds the connection configuration assoiated with this client.
 	config *ConnConfig
+
+	// chainParams holds the params for the chain that this client is using,
+	// and is used for many wallet methods.
+	chainParams *chaincfg.Params
 
 	// wsConn is the underlying websocket connection when not in HTTP POST
 	// mode.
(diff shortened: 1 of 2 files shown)


rpcclient/infrastructure.go
```diff
@@ -1065,6 +1068,9 @@ type ConnConfig struct {
 	// Pass is the passphrase to use to authenticate to the RPC server.
 	Pass string
 
+	// Params is the string represening the network that the server is running.
+	Params string
```
torkelrogstad: Mention that if not set, defaults to mainnet?

rpcclient/wallet.go
```diff
@@ -758,20 +757,20 @@ type FutureAddMultisigAddressResult chan *response
 // Receive waits for the response promised by the future and returns the
 // multisignature address that requires the specified number of signatures for
 // the provided addresses.
-func (r FutureAddMultisigAddressResult) Receive() (btcutil.Address, error) {
+func (r FutureAddMultisigAddressResult) Receive() (string, error) {
```
torkelrogstad: One option here (and the other places the return type got changed from `btcutil.Adress` to `string`) could be to do 

```golang
type FutureAddMultisigAddressResult  struct {
	response chan *response
	network *chaincfg.Params
}

func (r FutureAddMultisigAddressResult) Receive() (btcutil.Address, error) {
	res, err := receiveFuture(r.response)
	if err != nil {
		return nil, err
	}

	// Unmarshal result as a string.
	var addr string
	err = json.Unmarshal(res, &addr)
	if err != nil {
		return nil, err
	}

	return btcutil.DecodeAddress(addr, r.network)
}

func (c *Client) AddMultisigAddressAsync(requiredSigs int, addresses []btcutil.Address, account string) FutureAddMultisigAddressResult {
	addrs := make([]string, 0, len(addresses))
	for _, addr := range addresses {
		addrs = append(addrs, addr.String())
	}

	cmd := btcjson.NewAddMultisigAddressCmd(requiredSigs, addrs, &account)
	resp :=  c.sendCmd(cmd)
	return FutureAddMultisigAddressResult{network:c.chainParams, response:resp}
}
```

And in

rpcclient/infrastructure.go
Rjected: Added a note

rpcclient/wallet.go
Rjected: Good idea! I think I got all of the address commands with [`327749a`](https://github.com/btcsuite/btcd/pull/1467/commits/327749abd23118bf0daf4e117a7986cdba58b7c6).

nostdm:
What's the status of this change being marged into master?

tiger5226:
We still need it to @lbryio , we have to fork it...

torkelrogstad:
@Roasbeef this PR is fairly well reviewed, and looks like it is needed by quite a large number of people in the community. Would it be possible to get this merged?

Rjected:
@jcvernaleo I think this is an important PR, can you take a look?

jakesyl:
Agree - definitely a high priority issue

jakesyl:
@jcvernaleo 

- high priority
- bug

PR #1476, opened 2019-10-10
rpc: fix searchrawtransactionsverbose optional parameters

bitspill:
Previously if `includePrevOut == false` the omission of a `prevOut` value caused the subsequent `reverse` and `filterAddrs` parameters to be ignored by btcjson->marshalCmd->makeParams

### DIFF
--- a/rpcclient/rawtransactions.go
+++ b/rpcclient/rawtransactions.go
@@ -598,16 +598,18 @@ func (r FutureSearchRawTransactionsVerboseResult) Receive() ([]*btcjson.SearchRa
 //
 // See SearchRawTransactionsVerbose for the blocking version and more details.
 func (c *Client) SearchRawTransactionsVerboseAsync(address btcutil.Address, skip,
-	count int, includePrevOut, reverse bool, filterAddrs *[]string) FutureSearchRawTransactionsVerboseResult {
+	count int, includePrevOut, reverse bool, filterAddrs []string) FutureSearchRawTransactionsVerboseResult {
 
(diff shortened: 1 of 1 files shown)


bjornoj:
LGTM, but could you also add a test?

jakesyl:
@jcvernaleo (as per #1530)

- Low priority
- Bug
- Needs Test 

Also looks like this re

rpcclient/rawtransactions.go
```diff
@@ -598,16 +598,18 @@ func (r FutureSearchRawTransactionsVerboseResult) Receive() ([]*btcjson.SearchRa
 //
 // See SearchRawTransactionsVerbose for the blocking version and more details.
 func (c *Client) SearchRawTransactionsVerboseAsync(address btcutil.Address, skip,
-	count int, includePrevOut, reverse bool, filterAddrs *[]string) FutureSearchRawTransactionsVerboseResult {
+	count int, includePrevOut, reverse bool, filterAddrs []string) FutureSearchRawTransactionsVerboseResult {
```
onyb, line 601: This will be a breaking change, so it is better to introduce this in the next major release cycle. We'll need https://github.com/btcsuite/btcd/pull/1622 for `rpcclient` first though.

onyb (commented):
This bug should normally be resolved already. I added a test in https://github.com/btcsuite/btcd/pull/1635 showing exactly that. Let me know if it works for you.

onyb:
Closing this since #1635 is in. Please reopen if your issue is not resolved yet.

Issue #1471, opened 2019-09-25
Something in the wild started freezing my btcd


darwin:
I've been running btcd in production for a year. After [initial sync problems](https://github.com/btcsuite/btcd/issues/1024#issuecomment-360524638), I got it working and it was quite smooth sail. Most recently I've been running 130ea5bddde33df32b06a1cdb42a6316eb73cff5. About a week ago I noticed the node to fall out of sync and stayed stuck.

Restarting btcd and launching it again started syncing to catch up. It was behind couple hundred blocks.

The problem is that something triggers the freeze again and again I was unable to complete the syncing. I'm able to sync few blocks and the freeze happens again. I assume this is caused by some misbehaving peer reconnecting and feeding my btcd node some problematic data.

Here is a log just before the freeze happened (with redacted IPs):
```
btcd    | 2019-09-24 21:52:02.788 [INF] SYNC: New valid peer <IP1>:8333 (outbound) (/Satoshi:0.15.1/)
btcd    | 2019-09-24 21:52:03.940 [INF] SYNC: Lost peer <IP1>:8333 (outbound)
btcd    | 2019-09-24 21:52:04.103 [INF] SYNC: New valid peer <IP2>:8333 (outbound) (/Satoshi:0.13.1/)
btcd    | 2019-09-24 21:52:12.065 [INF] CHAN: Adding orphan block 0000000000000000000d11ede72158396bed291a784be023a51d7f812d41196d with parent 000000000000000000100abc79d25cc29e0a22800c53e90b7b9fceaeba991e1c
btcd    | 2019-09-24 21:52:16.297 [INF] SYNC: Lost peer <IP3>:59690 (inbound)
btcd    | 2019-09-24 21:52:16.298 [INF] SYNC: Lost peer <IP4>:57225 (inbound)
btcd    | 2019-09-24 21:52:16.299 [INF] SYNC: New valid peer <IP5>:60213 (inbound) (/bitnodes.earn.com:0.1/)
btcd    | 2019-09-24 21:52:18.802 [INF] SYNC: New valid peer <IP6>:34082 (inbound) (/bitcoin-seeder:0.01/)
btcd    | 2019-09-24 21:52:19.681 [INF] SYNC: Lost peer <IP6>:34082 (inbound)
btcd    | 2019-09-24 21:52:21.230 [INF] SYNC: Lost peer <IP5>:60213 (inbound)
[...]

PR #1480, opened 2019-10-11
 server: refactor OnVersion to instead add peers within OnVerAck

wpaulino:
This change is needed as part of requiring peers to also send a verack message following their version message during protocol negotiation. Peers were previously added to the SyncManager before their message queues were started, causing the server to stall if a peer didn't provide a timely verack response following their version. We now do this within OnVerAck, which happens shortly before peer message queues are started.

Fixes https://github.com/btcsuite/btcd/issues/1471.

### DIFF
--- a/server.go
+++ b/server.go
@@ -1563,7 +1563,7 @@ func (s *server) handleUpdatePeerHeights(state *peerState, umsg updatePeerHeight
 // handleAddPeerMsg deals with adding new peers.  It is invoked from the
 // peerHandler goroutine.
 func (s *server) handleAddPeerMsg(state *peerState, sp *serverPeer) bool {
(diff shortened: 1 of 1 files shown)


Roasbeef:
Needs a rebase.

server.go
```diff
@@ -1651,6 +1648,9 @@ func (s *server) handleAddPeerMsg(state *peerState, sp *serverPeer) bool {
 		}
 	}
 
+	// Signal the sync manager this peer is a new sync candidate.
+	s.syncManager.NewPeer(sp.Peer)
```
Roasbeef: 👍

wpaulino:
Rebased!

Issue #1464, opened 2019-09-04
Example of NewOutboundPeer doesn't work


arberiii:
Example written here https://godoc.org/github.com/btcsuite/btcd/peer#example-package--NewOutboundPeer doesn't work. The reason why is because self connection is not allowed. The example is also in peer_test package and it works there since when tests are init, function peer.TstAllowSelfConns() is called, which is placed in export_test.go file. I try to call same function in the main example but it cannot find the reference of the function because it is in a _test.go file. In my fork I resolved this issue by moving this function somewhere else.

I suggest to rename the file export_test.go to export.go and call the function in the main example.

Rjected:
Yes, there is a global `allowSelfConns` which is only modified in testing. The issue isn't actually in the `allowSelfConns`, but in `sentNonces`, which is also a global. This example would work if the `sentNonces` cache were moved to the `Peer` struct rather than be a global variable. Moving `sentNonces` would also remove the need for `allowSelfConns`, since that's used just for the example.

gtank:
I've also run into this: I'm trying to use just the peer library (but not the entire node) and would like to be able to generate mock peers. I opened #1481 with a fix.

PR #1481, opened 2019-10-12
peer: Move AllowSelfConns into peer config

gtank:
Currently, the `NewOutboundPeer` [example](https://godoc.org/github.com/btcsuite/btcd/peer#example-package--NewOutboundPeer) in the docs can't actually work outside of the `peer_test` package. This change allows external users of the peer library to designate peer configs as explicitly for testing that will allow self-connections.

Closes https://github.com/btcsuite/btcd/issues/1464


gtank:
Another approach to this would be to move `sentNonces` out of the global namespace. I'm not sure what approach the maintainers prefer, so I started with the one that seemed lowest impact.

bjornoj:
Add a test?

jakesyl:
@jcvernaleo (as per #1530)

- Low priority (doesn't affect node run as per #1464)
- Enhancement

conradoplg:
Note that this was merged later in https://github.com/btcsuite/btcd/pull/1698 and this PR could be closed.

jcvernaleo:
Thanks for point that out.  Closing this since the other PR covers it.

PR #1482, opened 2019-10-23
wire: decrease MaxGetCFiltersReqRange from 1000 to 100

Sjors:
BIP157 specificies a maximum of 100

Reverts ceb1caeb249cf3c7a8c1be2987af64bfffdd674d

This should prevent getting disconnected by Bitcoin Core nodes: https://github.com/bitcoin/bitcoin/pull/16442#issuecomment-545364727

### DIFF
--- a/wire/msgcfheaders.go
+++ b/wire/msgcfheaders.go
@@ -17,7 +17,7 @@ const (
 	MaxCFHeaderPayload = chainhash.HashSize
 
 	// MaxCFHeadersPerMsg is the maximum number of committed filter headers
-	// that can be in a single bitcoin cfheaders message.
+	// that can be in a single bitcoin cfheaders message according to BIP157.
 	MaxCFHeadersPerMsg = 2000
 )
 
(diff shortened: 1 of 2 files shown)


Sjors:
You may need a workaround on the server side to allow 1000 for a while, because existing (Lnd) users depend on that.

Roasbeef:
IMO the current value is superior as it allows for quicker restarts/rescans after the node has been offline for some time. It's able to fetch roughly a week's worth of filters in a single round trip. Using a value of 100 means we needs multiple round trips to fetch the filters for even a single day, which increase the latency from "open app" to "ready to rock". The PR linked above (the BIP PR) was merged w/o any comment from the OP nor any of the other co-authors of the BIP, and IMO was a mistake.

Sjors:
You mean https://github.com/bitcoin/bips/commit/f4948ddb4f4a2b536e774d7d80ad281b67c26a30 by @jimpo?

Roasbeef:
Yep, here's the original PR: https://github.com/bitcoin/bips/pull/699

Roasbeef:
BIP has been updated to move to 1k.

PR #1491, opened 2019-11-12
Implement 'getblockstats' JSON-RPC command

Jeiwan:
I had to implement `HashOrHeight` type because `getblockstats` RPC command takes either block height or hash and `interface{}` is not allowed in command arguments. The new type basically wraps an `interface{}` and defines JSON marshaling and unmarshaling rules.

### DIFF
--- a/btcjson/chainsvrcmds.go
+++ b/btcjson/chainsvrcmds.go
@@ -193,6 +193,50 @@ func NewGetBlockHeaderCmd(hash string, verbose *bool) *GetBlockHeaderCmd {
 	}
 }
 
+// HashOrHeight defines a type that can be used as hash_or_height value in JSON-RPC commands.
+type HashOrHeight struct {
+	Value interface{}
+}
+
+// MarshalJSON implements the json.Marshaler interface
+func (h HashOrHeight) MarshalJSON() ([]byte, error) {
+	return json.Marshal(h.Value)
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface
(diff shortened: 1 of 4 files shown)


btcjson/chainsvrcmds.go
```diff
@@ -193,6 +193,50 @@ func NewGetBlockHeaderCmd(hash string, verbose *bool) *GetBlockHeaderCmd {
...
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface
+func (h *HashOrHeight) UnmarshalJSON(data []byte) error {
+	var unmarshalled interface{}
+	if err := json.Unmarshal(data, &unmarshalled); err != nil {
+		return err
+	}
+
+	switch unmarshalled.(type) {
```
torkelrogstad: if you do `switch val := unmarshalled.(type) {` can assign `h.Value = val` directly on line 215 and 217 without the cast

torkelrogstad:
LGTM

Tested locally, worked with `bitcoind v0.18.1` :+1:

btcjson/chainsvrcmds.go
Jeiwan: Fixed.

bjornoj:
LGTM, tested locally, worked with bitcoind 0.19.99

jakesyl:
Tested locally, working find for me

jakesyl:
@jcvernaleo (as per #1530)

- low priority
- enhancement

Mergeable and tested locally by me, @bjornoj and @torkelrogstad

PR #1496, opened 2019-11-21
peer: remove high-mem trace logging

Crypt-iQ:
This PR removes trace level logging in the `UpdateLastBlockHeight` function. When profiling via `pprof`, this log alone showed 100MB+ memory usage when syncing a fresh neutrino node.

```
(pprof) list UpdateLastBlockHeight
Total: 23.58GB
ROUTINE ======================== github.com/btcsuite/btcd/peer.(*Peer).UpdateLastBlockHeight in /Users/nsa/go/pkg/mod/github.com/btcsuite/btcd@v0.20.1-beta/peer/peer.go
     101MB      101MB (flat, cum)  0.42% of Total
         .          .    492:// UpdateLastBlockHeight updates the last known block for the peer.
         .          .    493://
         .          .    494:// This function is safe for concurrent access.
         .          .    495:func (p *Peer) UpdateLastBlockHeight(newHeight int32) {
         .          .    496:	p.statsMtx.Lock()
   65.50MB    65.50MB    497:	log.Tracef("Updating last block height of peer %v from %v to %v",
   35.50MB    35.50MB    498:		p.addr, p.lastBlock, newHeight)
         .          .    499:	p.lastBlock = newHeight
         .          .    500:	p.statsMtx.Unlock()
         .          .    501:}
         .          .    502:
         .          .    503:// UpdateLastAnnouncedBlock updates meta-data about the last block hash this
```


bjornoj:
Is it necessary to remove the log? How about setting level to info mode or above when syncing the node?

Crypt-iQ:
> Is it necessary to remove the log? How about setting level to info mode or above when syncing the node?

I tried this and it didn't work since the allocations appear to be happening with the actual invocation to log

torkelrogstad:
What about this? 

```go
	if log.Level() <= btclog.LevelTrace {
		log.Tracef("Updating last block height of peer %v from %v to %v",
			p.addr, p.lastBlock, newHeight)
	}
```

Crypt-iQ:
> What about this?
> 
> ```go
> 	if log.Level() <= btclog.LevelTrace {
> 		log.Tracef("Updating last block height of peer %v from %v to %v",
> 			p.addr, p.lastBlock, newHeight)
> 	}
> ```

This should work, I was using a log closure instead. Thanks :)

torkelrogstad:
ACK

Roasbeef:
I find it odd that this actually reduces memory in practice, given that the code more or less does this inline: https://github.com/btcsuite/btclog/blob/master/log.go#L335. There's always a call to `Level`, but all that does is load an atomic integer.

Roasbeef:
Do y'all have a "deeper" profile available? So something like a flame graph so we can dig in a bit more here. Also would utilizing a log closure help here at all? Or always passing pointers rather than the raw values?

Crypt-iQ:
I don't have a deeper profile available. A log closure didn't help at all. IIRC the blowup was from passing `p.addr, p.lastBlock, newHeight` to the func. I think passing pointers would be a better sol'n here

Crypt-iQ:
> Do y'all have a "deeper" profile available? So something like a flame graph so we can dig in a bit more here. Also would utilizing a log closure help here at all? Or always passing pointers rather than the raw values?

Actually I'm not sure what you mean here by passing pointers instead of raw values. The bottleneck is just that the vars are being passed by value. So that's why the conditional avoided the mem usage

jakesyl:
@jcvernaleo (as per #1530)

- low priority
- enhancement

peer/peer.go
```diff
@@ -493,9 +494,12 @@ func (p *Peer) String() string {
 //
 // This function is safe for concurrent access.
 func (p *Peer) UpdateLastBlockHeight(newHeight int32) {
+	if log.Level() <= btclog.LevelTrace {
+		log.Tracef("Updating last block height of peer %v from %v to %v",
+			p.addr, p.lastBlock, newHeight)
+	}
+
 	p.statsMtx.Lock()
-	log.Tracef("Updating last block height of peer %v from %v to %v",
```
allocz: Performed some memory profiling performing IBD on mainnet up to height 200000.

Before changing the code (master branch), `UpdateLastBlockHeight` does not perform any allocation.

Moved the log to the top of the function to be sure it is called every time. ~20MiB of bytes allocated by the `UpdateLastBlockHeight` procedure (total):

Total memory allocated:
Image

Total allocations:
Image

With the provided patch:

Total memory allocated:
Image

Total allocations:
Image

As of now, the profile show better results without the patch. So my vote is to close this PR.

PR #1502, opened 2019-12-02
Try both TX serialization formats in createrawtransaction

torkelrogstad:
Currently, it is not possible to use the `rpcclient` package to create a raw TX with Bitcoin Core if the transaction you're creating has no inputs. This is a realistic scenario, as you would typically call `createrawtransaction` with your outputs specified, and then follow up with `fundrawtransaction`. 

The reason the RPC call fails is that we assume the returned transaction is encoded with the new witness format that was introduced with Segwit. We solve this by simply trying both serialization formats. If its possible to detect which format is used (so we avoid doing it twice) I'll update the PR.

### DIFF
--- a/rpcclient/rawtransactions.go
+++ b/rpcclient/rawtransactions.go
@@ -233,8 +233,13 @@ func (r FutureCreateRawTransactionResult) Receive() (*wire.MsgTx, error) {
 
 	// Deserialize the transaction and return it.
 	var msgTx wire.MsgTx
-	if err := msgTx.Deserialize(bytes.NewReader(serializedTx)); err != nil {
-		return nil, err
+	// we try both the new and old encoding format
+	witnessErr := msgTx.Deserialize(bytes.NewReader(serializedTx))
+	if witnessErr != nil {
+		legacyErr := msgTx.DeserializeNoWitness(bytes.NewReader(serializedTx))
+		if legacyErr != nil {
+			return nil, legacyErr
+		}
 	}
 	return &msgTx, nil
 }


bjornoj:
LGTM, but maybe leave a comment for context? Just a short version of what you commented here would be nice, for example `we try both serialization formats, to be able to create a transaction with no inputs`

jakesyl:
Tbh, @bjornoj I think the existing comment is clear enough, not sure we need anything more verbose

rpcclient/rawtransactions.go
```diff
@@ -233,9 +233,15 @@ func (r FutureCreateRawTransactionResult) Receive() (*wire.MsgTx, error) {
...
 	var msgTx wire.MsgTx
-	if err := msgTx.Deserialize(bytes.NewReader(serializedTx)); err != nil {
-		return nil, err
+	// we try both the new and old encoding format
+	witnessErr := msgTx.Deserialize(bytes.NewReader(serializedTx))
+	if witnessErr != nil {
+		legacyErr := msgTx.DeserializeNoWitness(bytes.NewReader(serializedTx))
+		if legacyErr != nil {
+			return nil, legacyErr
+		}
```
Rjected: If both return an error, would the call to `msgTx.Deserialize(...)` return something useful that isn't already covered by `msgTx.DeserializeNoWitness(...)`?

torkelrogstad: Sorry, I'm not sure I understand what you mean. Is this about which error to return?

Rjected: Sort of, yeah. Although I also just looked at the code and it's fine to do it this way. Let's say both `msgTx.Deserialize` and `msgTx.DeserializeNoWitness` both return an error. In this case, we go down the `if witnessErr != nil` branch and eventually return `(nil, legacyErr)`. But there is a `witnessErr` which is non-nil and I was wondering if it might make sense to return both error messages.
I just checked the code though and both just make calls to `msg.BtcDecode` so it should be fine.

jakesyl:
@jcvernaleo (as per #1530)

- high priority
- bug (technically an enhancement, but a clear compatibility issue with bitcoind rpc)

This has been pretty closely reviewed/run locally by myself and @Rjected so probably mergable

rpcclient/rawtransactions.go
```diff
@@ -233,9 +233,15 @@ func (r FutureCreateRawTransactionResult) Receive() (*wire.MsgTx, error) {
...
 	}
+
```
jcvernaleo: Super minor, but please remove the extraneous newline addition.

torkelrogstad: Fixed

PR #1509, opened 2020-01-02
BUG:dynamicbanscore deadlock

lochjin:
https://github.com/btcsuite/btcd/issues/1510

### DIFF
--- a/connmgr/dynamicbanscore_test.go
+++ b/connmgr/dynamicbanscore_test.go
@@ -66,3 +66,15 @@ func TestDynamicBanScoreReset(t *testing.T) {
 		t.Errorf("Failed to reset ban score.")
 	}
 }
+
+// TestDynamicBanScoreString
+func TestDynamicBanScoreString(t *testing.T) {
+	var bs DynamicBanScore
+	base := time.Now()
+
+	r := bs.increase(100, 50, base)
+	if r != 150 {
+		t.Errorf("Unexpected result %d after ban score increase.", r)
+	}
+	t.Log(bs.String())
+}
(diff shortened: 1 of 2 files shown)


torkelrogstad:
Nice find! Could you add a test for the `String()` method? So we make sure the deadlock isn't reintroduced later by someone finding the line you just changed and thinking it makes sense to replace it with the function call.

lochjin:
> Nice find! Could you add a test for the `String()` method? So we make sure the deadlock isn't reintroduced later by someone finding the line you just changed and thinking it makes sense to replace it with the function call.

OK, no problem.   https://github.com/btcsuite/btcd/pull/1509/commits/530ea00a48e54c762e08e4fbd4c7ddcc19de86ce

connmgr/dynamicbanscore_test.go
```diff
@@ -66,3 +67,15 @@ func TestDynamicBanScoreReset(t *testing.T) {
...
+// TestDynamicBanScoreString
+func TestDynamicBanScoreString(t *testing.T) {
+	var bs DynamicBanScore
+	base := time.Now()
+
+	r := bs.increase(100, 50, base)
+	if r != 150 {
+		t.Errorf("Unexpected result %d after ban score increase.", r)
+	}
+	fmt.Println(bs.String())
```
Rjected: Rather than using fmt here, you could do `t.Log` or `t.Logf` and I think it should still work.

lochjin: 👍OK. https://github.com/btcsuite/btcd/pull/1509/commits/c982d9b90643620b66f8758c4dcb48ef491c3547

jakesyl:
@jcvernaleo 

- high priority
- bug

Mergable

jcvernaleo:
@lochjin would you be able to squash this to a single commit and then I'll merge it?

lochjin:
> @lochjin would you be able to squash this to a single commit and then I'll merge it?



> @lochjin would you be able to squash this to a single commit and then I'll merge it?

OK

jakesyl:
Looks good!

PR #1513, opened 2020-01-07
BUG ： SignRawTransaction4 Params missing field "Amount"

kkkkking233:
the RawTxInput struct lost a field  "Amount" ,the RPC while return err


kkkkking233:
RPC always replies err before I change，after the change, everything is ok

PR #1514, opened 2020-01-07
BUG：RPC SignRawTransaction2-4 params missing field "Amount"

kkkkking233:
the RawTxInput struct missing a field "Amount",the RPC will return err.

### DIFF
--- a/btcjson/walletsvrcmds.go
+++ b/btcjson/walletsvrcmds.go
@@ -590,6 +590,7 @@ type RawTxInput struct {
 	Vout         uint32 `json:"vout"`
 	ScriptPubKey string `json:"scriptPubKey"`
 	RedeemScript string `json:"redeemScript"`
+	Amount       float64 `json:"amount"`
 }
 
 // SignRawTransactionCmd defines the signrawtransaction JSON-RPC command.


jakesyl:
@jcvernaleo (as per #1530)

- low priority
- outdated (needs to be updated so CI tests can passed and issues are deeper than #1537)

onyb:
This command has been removed in Bitcoin Core 0.18.0, and replaced by [`signrawtransactionwithwallet`](https://bitcoincore.org/en/doc/0.20.0/rpc/wallet/signrawtransactionwithwallet). There is already a pending pull request (#1561) adding support for this.

@jakesyl  I propose closing this and do a separate PR to nuke `signrawtransaction`.

btcjson/walletsvrcmds.go
```diff
@@ -590,6 +590,7 @@ type RawTxInput struct {
 	Vout         uint32 `json:"vout"`
 	ScriptPubKey string `json:"scriptPubKey"`
 	RedeemScript string `json:"redeemScript"`
+	Amount       float64 `json:"amount"`
```
allocz: btcwallet uses the structure without the amount, but in fact older versions of Bitcoin Core had the `signrawtransaction` command and the [docs](https://bitcoincore.org/en/doc/0.16.0/rpc/rawtransactions/signrawtransaction/) requires the amount when the optional `prevtxs` is passed.

Adding the field should not break btcd nor btcwallet and fixes the RPC for old Bitcoin Core nodes, I don't see any problem in merging this. LGTM.

allocz (approved):

PR #1516, opened 2020-01-10
build: use official go-leveldb

islishude:
syndtr/go-leveldb supports go module at now,we could migrate to latest go-leveldb libary to improve performance.


jakesyl:
@jcvernaleo (as per #1530)

- low priority
- enhancement

jakesyl:
@islishude is there a reason this isn't a good idea now?

PR #1521, opened 2020-01-19
testnet config for wallet, this is ugly, do not use

LiranCohen:


LiranCohen:
Did not mean to pull against upstream

PR #1526, opened 2020-01-31
make a block with an invalid orphan child a valid block

someone235:
There is a bug where a block is considered invalid if it has an invalid orphan child

### DIFF
--- a/blockchain/chain_test.go
+++ b/blockchain/chain_test.go
@@ -964,3 +964,62 @@ func TestIntervalBlockHashes(t *testing.T) {
 		}
 	}
 }
+
+func TestProcessOrphans(t *testing.T) {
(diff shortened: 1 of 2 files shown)


someone235:
I'm not really sure why it fails on travis, but in my machine the tests pass.

Roasbeef:
Haven't looked closely at the test, but perhaps there's a timing assumption that differs locally vs when it's run on Travis?

Rjected:
For this reason maybe we should add the `err` returned by `ProcessBlock` to the `t.Fatalf` in [fullblocks_test:195](https://github.com/btcsuite/btcd/blob/f3ec13030e4e828869954472cbc51ac36bee5c1d/blockchain/fullblocks_test.go#L195), not sure why it's not there in the first place but it would at least help in debugging this

jakesyl:
@jcvernaleo

PR #1528, opened 2020-01-31
Refactor GetBlockVerboseTx to reflect correct getblock RPC call parameters.

jalavosus:
Currently, rpcclient.GetBlockVerboseTx(...) does not return raw transaction data. This is due to how GetBlockCmd is implemented in btcjson; in upstream/master, `getblock` sends two boolean parameters (`verbose` and `verbosetx`), which throws an error. 

According to the bitcoin JSON RPC documentation, `getblock` accepts two parameters: `blockhash` (as a string) and `verbosity` (as an integer). `verbosity` can be any of `0`, `1`, or `2`, with `0` being the default. 

```
getblock "blockhash" ( verbosity ) 

If verbosity is 0, returns a string that is serialized, hex-encoded data for block 'hash'.
If verbosity is 1, returns an Object with information about block .
If verbosity is 2, returns an Object with information about block  and information about each transaction. 

Arguments:
1. "blockhash"          (string, required) The block hash
2. verbosity              (numeric, optional, default=1) 0 for hex encoded data, 1 for a json object, and 2 for json object with transaction data
```

This pull request aims to correctly implement `verbosity` as a parameter to `getblock` by refactoring `GetBlockCmd` (in `btcjson/chainsvrcmds.go`) to have two fields: `Hash` and `Verbosity`, with `Verbosity` being an optional int pointer. 

In order to implement correct behavior, three primary changes are necessary: the above refactor, an additional type specific to `getblock verbosity=2`, and a new `Future` type to enable proper (and error-free) unmarshalling of `verbosity=2` data. 

The new type (`GetBlockVerboseTxResult`) is almost entirely the same as  `GetBlockVerboseResult`, but with the minor modification of changing the `Tx` field from a `[]string` to a `[]TxRawResult` so as to enable proper unmarshalling. 

The new `Future` type (`FutureGetBlockVerboseTxResult`) returns a `GetBlockVerboseTxResult` rather than a `GetBlockVerboseResult` so that data can be properly returned to the user without error.

### DIFF
--- a/btcjson/chainsvrcmds.go
+++ b/btcjson/chainsvrcmds.go
@@ -130,8 +130,7 @@ func NewGetBestBlockHashCmd() *GetBestBlockHashCmd {
 // GetBlockCmd defines the getblock JSON-RPC command.
 type GetBlockCmd struct {
 	Hash      string
-	Verbose   *bool `jsonrpcdefault:"true"`
-	VerboseTx *bool `jsonrpcdefault:"false"`
+	Verbosity *int `jsonrpcdefault:"0"`
 }
 
 // NewGetBlockCmd returns a new instance which can be used to issue a getblock
@@ -139,11 +138,10 @@ type GetBlockCmd struct {
 //
 // The parameters which are pointers indicate they are optional.  Passing nil
 // for optional parameters will use the default value.
-func NewGetBlockCmd(hash string, verbose, verboseTx *bool) *GetBlockCmd {
+func NewGetBlockCmd(hash string, verbosity *int) *GetBlockCmd {
 	return &GetBlockCmd{
 		Hash:      hash,
-		Verbose:   verbose,
-		VerboseTx: verboseTx,
+		Verbosity: verbosity,
 	}
 }
 
--- a/btcjson/chainsvrresults.go
+++ b/btcjson/chainsvrresults.go
@@ -25,8 +25,11 @@ type GetBlockHeaderVerboseResult struct {
 }
 
 // GetBlockVerboseResult models the data from the getblock command when the
-// verbose flag is set.  When the verbose flag is not set, getblock returns a
-// hex-encoded string.
+// verbose flag is set to 1.  When the verbose flag is set to 0, getblock returns a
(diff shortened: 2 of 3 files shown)


jalavosus:
Additional data from the RPC documentation:

```
Result (for verbosity = 0):
"data"             (string) A string that is serialized, hex-encoded data for block 'hash'.

Result (for verbosity = 1):
{
  "hash" : "hash",     (string) the block hash (same as provided)
  "confirmations" : n,   (numeric) The number of confirmations, or -1 if the block is not on the main chain
  "size" : n,            (numeric) The block size
  "strippedsize" : n,    (numeric) The block size excluding witness data
  "weight" : n           (numeric) The block weight as defined in BIP 141
  "height" : n,          (numeric) The block height or index
  "version" : n,         (numeric) The block version
  "versionHex" : "00000000", (string) The block version formatted in hexadecimal
  "merkleroot" : "xxxx", (string) The merkle root
  "tx" : [               (array of string) The transaction ids
     "transactionid"     (string) The transaction id
     ,...
  ],
  "time" : ttt,          (numeric) The block time in seconds since epoch (Jan 1 1970 GMT)
  "mediantime" : ttt,    (numeric) The median block time in seconds since epoch (Jan 1 1970 GMT)
  "nonce" : n,           (numeric) The nonce
  "bits" : "1d00ffff", (string) The bits
  "difficulty" : x.xxx,  (numeric) The difficulty
  "chainwork" : "xxxx",  (string) Expected number of hashes required to produce the chain up to this block (in hex)
  "nTx" : n,             (numeric) The number of transactions in the block.
  "previousblockhash" : "hash",  (string) The hash of the previous block
  "nextblockhash" : "hash"       (string) The hash of the next block
}

Result (for verbosity = 2):
{
  ...,                     Same output as verbosity = 1.
  "tx" : [               (array of Objects) The transactions in the format of the getrawtransaction RPC. Different from verbosity = 1 "tx" result.
         ,...
  ],
  ,...                     Same output as verbosity = 1.
}
```

jalavosus:
Closing due to incorrect branching

PR #1531, opened 2020-01-31
add MoneySupply and StakingStatus to the InfoWalletResult type

olegzarubin:
Added MoneySupply and StakingStatus to the InfoWalletResult type.

### DIFF
--- a/btcjson/walletsvrresults.go
+++ b/btcjson/walletsvrresults.go
@@ -48,11 +48,13 @@ type InfoWalletResult struct {
 	Proxy           string  `json:"proxy"`
 	Difficulty      float64 `json:"difficulty"`
 	TestNet         bool    `json:"testnet"`
+	MoneySupply     float64 `json:"moneysupply"`
 	KeypoolOldest   int64   `json:"keypoololdest"`
 	KeypoolSize     int32   `json:"keypoolsize"`
 	UnlockedUntil   int64   `json:"unlocked_until"`
 	PaytxFee        float64 `json:"paytxfee"`
 	RelayFee        float64 `json:"relayfee"`
+	StakingStatus   string  `json:"staking status"`
 	Errors          string  `json:"errors"`
 }
 


btcjson/walletsvrresults.go
```diff
@@ -48,11 +48,13 @@ type InfoWalletResult struct {
 	Proxy           string  `json:"proxy"`
 	Difficulty      float64 `json:"difficulty"`
 	TestNet         bool    `json:"testnet"`
+	MoneySupply     float64 `json:"moneysupply"`
 	KeypoolOldest   int64   `json:"keypoololdest"`
 	KeypoolSize     int32   `json:"keypoolsize"`
 	UnlockedUntil   int64   `json:"unlocked_until"`
 	PaytxFee        float64 `json:"paytxfee"`
 	RelayFee        float64 `json:"relayfee"`
+	StakingStatus   string  `json:"staking status"`
```
torkelrogstad: This JSON tag doesn't look right... Also, is this a field returned by `btcd`?

olegzarubin: this is in the output of the getinfo command ... and it looks just like that ... and it works

example

jakesyl: I'd ran this as well with no issuese

torkelrogstad: Is "staking status" something `btcd` reports? I can't find it in the code. If this is something used by a non-Bitcoin chain I don't think `btcd` is the correct place to implement this.

olegzarubin: there are several thousand cryptocurrencies and, accordingly, *coind ... i do not think it is right for each cryptocurrency to write a separate rpcClient ...

torkelrogstad: I think that is absolutely right, that each crypto currency should maintain their own RPC client. If we added all possible fields to the `btcd/rpcclient` struct, that would make users of the Bitcoin client (which is what this package is made for) super confused. 

I'm strongly opposed to adding any altcoin functionality to this project. Forking it and maintaining a 1-commit difference is not hard.

olegzarubin: You have the right to your opinion, I have the right to my opinion ... I think that we do not need 10,000 cats that differ in the shape of the spots on the nose ...

torkelrogstad:
Strong NACK from me, btcd is not the appropriate place for altcoin functionality. @olegzarubin this is not meant as a personal attack on you, hope you don't take it as such.

olegzarubin:
I did not see any clear arguments why it cannot be here ... it’s just a package and make it one cryptocurrency ... it’s not right, it’s like a mathematical package only for functions of division of different types of numbers ...

olegzarubin:
P.S. I perceive this as a groundless denial ... nothing personal

jakesyl:
It's quite clearly aimed at being a bitcoin full node implementation. From the readme:

> btcd is an alternative full node bitcoin implementation written in Go (golang).

> It properly downloads, validates, and serves the block chain using the exact rules (including consensus bugs) for block acceptance as Bitcoin Core

olegzarubin:
What is the contradiction? Has Bitcoin stopped being a cryptocurrency? Why is there a special limited functionality of the package at such a low level as the rpcСlient? Its expansion does not allow to implement a full node?

P.S. Thank you. These questions are rhetorical, further I am not interested.

jcvernaleo:
Not commenting on this change in particular, but when I initially wrote all the json rpc stuff for btcd (back when it was the separate `btcjson` repo), I had hoped to make it generic enough for altcoins as long as it didn't negatively interact with the bitcoin functionality (which always had priority).

Then when we forked it for decred, I once again hoped to make things generic enough to use the same pieces for both project when possible.  Timing and other priorities thwarted me again there.

olegzarubin:
I like your approach, it’s a pity that you didn’t have enough time ... I am opposed to restricting the functionality of libraries at a low level if this does not interfere with the achievement of the final goal ...

Roasbeef:
These fields aren't relevant to Bitcoin as it is today.

jakesyl:
@jcvernaleo, go structs are also particularly poorly suited to that end because they're not (yet extensible) - probably could pull it off if they were

PR #1503, opened 2019-12-04
Add Batch JSON-RPC support (rpc client & server)

jakesyl:
This pr aims to address #329 which has been open for a while now by adding bulk request support. This is base on from [decred#841](https://github.com/decred/dcrd/pull/841) by @davecgh

PR #1537, opened 2020-02-10
Remove $GOPATH Caching

jakesyl:
$GOPATH caching has led to flaky tests as per #1503 and #1536. The speedup is marginal and while the false negatives are a headache, false positives are potentially dangerous.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: go
 cache:
   directories:
     - $GOCACHE
-    - $GOPATH
     - $GOPATH/pkg/mod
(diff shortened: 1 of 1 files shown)


jakesyl:
Looks like [LND](https://github.com/lightningnetwork/lnd) has this removed too. Actually, looks like @Roasbeef never added it there in the first place. I'm guessing because it's just error prone:

https://github.com/lightningnetwork/lnd/commit/f65b1a4d6e515e9237cfe5dfa71b8f46dc4081c6

Issue #1530, opened 2020-01-31
Project Velocity


jakesyl:
So I want to prefix this with saying the entire community is extremely thankful for @Roasbeef stepping up as a co-maintainer and leading this after the [decred](https://github.com/decred/dcrd) teams activity decreased and they focused on their altcoin. Your contributions here/leadership on lnd are awesome (and I'm not kidding, there's literally a [reddit thread](https://www.reddit.com/r/Bitcoin/comments/7qj0b4/why_is_lightning_network_being_developed_on_btcd/dspk5be/?utm_source=share&utm_medium=web2x) refusing to believe your one person)

With that being said, we've got a pretty substantial backlog of pull requests 100+, a lot of which are approved for review than going stale or just not getting merged after a few months. Would it make sense to add more maintainers or figure out a peer review/ci workflow where you don't have to personally merge in every PR?

Just trying to figure out if we can do anything to get some of these PR's in more quickly or think about a more long running fork

Roasbeef:
If you want to pitch in, I think you can begin by simply starting to help to review PRs in the backlog. W.r.t the number of open prs, it's typical for open source projects that some contributions may just never be merged/accepted.

jakesyl:
Will do. I've got a few code reviews out now (and know @torkelrogstad has been extremely helpful in reviewing some of mine, e.g. #1503)

My point on the stale contributions is that _generally_ 3/4-year-old stuff with a bunch of merge conflicts should generally be closed and the PR's that are opened/reviewed  (e.g. #1467 which fixes #1466 and #1519) should probably be merged.

The longer these things stay open, the more merge conflicts occur and the slower the subsequent release cycles. I will definitely start reviewing some more of these PR's though

panzarino:
@Roasbeef / @davecgh, as far as getting these things merged - can you tell us a bit about the process. I'm seeing the last release was November and nothing has been merged since then. Can you shine a little light on what the process of getting this stuff merged normally looks like?

jcvernaleo:
I just merged #1537 and and going to make some attempt to both triage issues/PRs and do some reviewing and merging where possible.  This was discussed with @Rjected and @someone235 on irc.

jakesyl:
@jcvernaleo I saw this conversation, but missed you guys buy a few hours. I'm generally on irc as `jakesyl` and will catch up with you guys at some point. I've stopped doing PR's here because of how stale a lot of my current ones were getting, but happy to begin again if things are getting merged

[...]

PR #1538, opened 2020-02-11
remove redundant sending to msg.reply in processBlockMsg in blockHandler

someone235:
In case of an error, msg.reply get two responses, but reads only one. To prevent it from blocking, msg.reply was created with a buffer.
This PR makes sure that msg.reply will get only one response, and then removes the redundant buffer.


jakesyl:
@jcvernaleo (as per #1530)

- high priority
- bug (implementation wise)

Mergable

PR #1539, opened 2020-02-11
Added testnet and simnet to rpcclient

kryptodirect:
You can declare the type of Netmode you want to use in config. Defaults to MainNet 

The options are:
-Mainnet
-Simnet
-Testnet

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,5 @@ _cgo_export.*
 _testmain.go
 
 *.exe
+
+*.idea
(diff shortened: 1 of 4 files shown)


Rjected:
How is this different from #1467?

jakesyl:
@jcvernaleo (as per #1530)

- high priority
- feature

Contained in #1467 (which also fixes an issue around `type()`

jcvernaleo:
@jakesyl @Rjected with #1467 in, can I close this one as obsolete?  Seems like it to me but want to confirm.

jakesyl:
Id day so- this just looks like a more explicit implementation

PR #1541, opened 2020-02-18
btcjson: add RPC_IN_WARMUP error code

sangaman:
This adds an error code for the `RPC_IN_WARMUP` error code defined at https://github.com/bitcoin/bitcoin/blob/master/src/rpc/protocol.h#L49 which is thrown when bitcoind has started but has not yet finished verifying recent blocks and being ready for rpc calls.


jakesyl:
@jcvernaleo 

- low priority
- bug

PR #1544, opened 2020-03-05
Peer queue

jakesyl:
As per @jongillham, I've gone ahead and did my best to rebase #629. A lot of stuff has moved around since 2016, but I think I did a good job at rebasing here. There are a few failing tests, but we're definitely getting there.

I'm recommending we close #629 in favor of this and continue work here since #629 will not be updated


jakesyl:
@jcvernaleo (as per #1530)

- Low priority
- Enhancement

jakesyl:
Has to be re-opened to get around travis ci issues

PR #1552, opened 2020-03-14
go get missing from readme

boufni95:
added go get command to obtain btcd and dependencies


jakesyl:
Hi,

I'm not sure we want to direct users here as opposed to the release page.

boufni95:
Oh, It was a bit strange to have the cd and go install command and not the go get, but I understand now, thank you for the clarification

PR #1555, opened 2020-03-26
Add getzmqnotifications RPC in rpcclient

torkelrogstad:
In this PR we implement the getzmqnotifications RPC 
in the RPC client.

### DIFF
--- a/rpcclient/chain.go
+++ b/rpcclient/chain.go
@@ -52,14 +52,61 @@ func (c *Client) GetBestBlockHash() (*chainhash.Hash, error) {
 	return c.GetBestBlockHashAsync().Receive()
 }
 
+// legacyGetBlockRequest constructs and sends a legacy getblock request which
+// contains two separate bools to denote verbosity, in contract to a single int
+// parameter.
+func (c *Client) legacyGetBlockRequest(hash string, verbose,
+	verboseTx bool) ([]byte, error) {
+
+	hashJSON, err := json.Marshal(hash)
+	if err != nil {
+		return nil, err
+	}
+	verboseJSON, err := json.Marshal(btcjson.Bool(verbose))
+	if err != nil {
+		return nil, err
+	}
+	verboseTxJSON, err := json.Marshal(btcjson.Bool(verboseTx))
+	if err != nil {
+		return nil, err
+	}
+	return c.RawRequest("getblock", []json.RawMessage{
+		hashJSON, verboseJSON, verboseTxJSON,
+	})
+}
+
+// waitForGetBlockRes waits for the response of a getblock request. If the
+// response indicates an invalid parameter was provided, a legacy style of the
+// request is resent and its response is returned instead.
+func (c *Client) waitForGetBlockRes(respChan chan *response, hash string,
+	verbose, verboseTx bool) ([]byte, error) {
+
+	res, err := receiveFuture(respChan)
+
+	// If we receive an invalid parameter error, then we may be
+	// communicating with a btcd node which only understands the legacy
+	// request, so we'll try that.
+	if err, ok := err.(*btcjson.RPCError); ok &&
+		err.Code == btcjson.ErrRPCInvalidParams.Code {
+		return c.legacyGetBlockRequest(hash, verbose, verboseTx)
+	}
+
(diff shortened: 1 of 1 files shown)


btcjson/chainsvrcmds.go
```diff
@@ -864,6 +873,7 @@ func init() {
 	MustRegisterCmd("gettxout", (*GetTxOutCmd)(nil), flags)
 	MustRegisterCmd("gettxoutproof", (*GetTxOutProofCmd)(nil), flags)
 	MustRegisterCmd("gettxoutsetinfo", (*GetTxOutSetInfoCmd)(nil), flags)
+	MustRegisterCmd("getzmqnotifications", (*GetZmqNotificationsCmd)(nil), flags)
```
onyb, line 876: Wondering if `getzmqnotifications` is really a _chain_ command. Bitcoin Core categorizes it separately under the [`zmq` section](https://bitcoincore.org/en/doc/0.19.0/rpc/zmq/getzmqnotifications/). Maybe `walletsvrcmds.go` is a better file for this?

What do you think, @jcvernaleo @jakesyl?

btcjson/chainsvrresults.go
```diff
@@ -379,6 +382,48 @@ type GetRawMempoolVerboseResult struct {
 	Depends          []string `json:"depends"`
 }
 
+// ZmqNotificationResult models the data returned from the zmqnotifications command.
+type ZmqNotificationResult struct {
```
onyb, line 386: Better stick to the standard naming convention, and use `GetZmqNotificationsResult`.

rpcclient/zmq.go
```diff
@@ -0,0 +1,39 @@
...
+type FutureGetZmqNotificationsResult chan *response
+
+// Receive waits for the response promised by the future and returns the unmarshalled
+// response, or an error if the request was unsuccessful.
+func (r FutureGetZmqNotificationsResult) Receive() ([]btcjson.ZmqNotificationResult, error) {
+	res, err := receiveFuture(r)
+	if err != nil {
+		return nil, err
+	}
+	var notifications []btcjson.ZmqNotificationResult
```
onyb, line 20: I think it's better if `btcjson.GetZmqNotificationsResult` models _exactly_ the response of the RPC server, since someone may want to just use `btcjson` outside the context of btcd's rpcclient. This means the result type (currently in `btcjson/chainsvrresults.go`) should be defined as:

```go
type GetZmqNotificationsResult []struct {
    ...
}
```

as opposed to

```go
type GetZmqNotificationResult struct {
    ...
}
```

btcjson/chainsvrresults.go
```diff
@@ -379,6 +382,48 @@ type GetRawMempoolVerboseResult struct {
...
+	})
+}
+
+// UnmarshalJSON satisfies the json.Unmarshaller interface
+func (z *ZmqNotificationResult) UnmarshalJSON(bytes []byte) error {
+	type basicNotification struct {
+		Type    string
+		Address string
+		Hwm     int
+	}
```
onyb, line 406: We can use this in the `MarshalJSON` method as well, as a replacement for `map[string]interface{}` type, by moving `basicNotification` outside.

What do you think?

btcjson/chainsvrresults.go
```diff
@@ -379,6 +382,48 @@ type GetRawMempoolVerboseResult struct {
 	Depends          []string `json:"depends"`
 }
 
+// ZmqNotificationResult models the data returned from the zmqnotifications command.
+type ZmqNotificationResult struct {
+	Type          string
+	Address       *url.URL
+	HighWaterMark int
```
onyb, line 389: What do you think about adding some comments for the struct fields? I think it's particularly useful when identifiers do not convey enough meaning.

```suggestion
	Type          string    // Type of notification
	Address       *url.URL  // Address of the publisher
	HighWaterMark int       // Outbound message high water mark
```

btcjson/chainsvrresults.go
torkelrogstad: A benefit of keeping `basicNotification` within the unmarshaling method is that it doesn't pollute the package namespace. So I think it's better to keep it where it is now

PR #1558, opened 2020-04-02
rpcclient: serialize nil inputs to empty list

torkelrogstad:
Prior to this, calling `CreateRawTransaction(nil, ...)` failed. We fix this, and add a test for the serialization.

### DIFF
--- a/btcjson/chainsvrcmds.go
+++ b/btcjson/chainsvrcmds.go
@@ -63,10 +63,15 @@ type CreateRawTransactionCmd struct {
 // NewCreateRawTransactionCmd returns a new instance which can be used to issue
 // a createrawtransaction JSON-RPC command.
 //
-// Amounts are in BTC.
+// Amounts are in BTC. Passing in nil and the empty slice as inputs is equivalent,
+// both gets interpreted as the empty slice.
 func NewCreateRawTransactionCmd(inputs []TransactionInput, amounts map[string]float64,
 	lockTime *int64) *CreateRawTransactionCmd {
-
+	// to make sure we're serializing this to the empty list and not null, we
+	// explicitly initialize the list
+	if inputs == nil {
+		inputs = []TransactionInput{}
+	}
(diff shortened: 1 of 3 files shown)


torkelrogstad:
Pushed some doc tweaks :+1:

btcjson/chainsvrcmds.go
```diff
@@ -63,10 +63,15 @@ type CreateRawTransactionCmd struct {
 // NewCreateRawTransactionCmd returns a new instance which can be used to issue
 // a createrawtransaction JSON-RPC command.
 //
-// Amounts are in BTC.
+// Amounts are in BTC. Passing in nil and the empty slice is equivalent, both
```
onyb, line 66: nit: It seems the remark about "passing nil" applies to `amounts`, instead of `inputs`.

torkelrogstad: Fixed

onyb:
@torkelrogstad CI is not happy. 🔴

torkelrogstad:
Seems like I was a bit trigger happy with pushing! Missed a `gofmt`

jakesyl:
We might have to update this branch from master to get rid of the travis health check

onyb:
> We might have to update this branch from master to get rid of the travis health check

@jakesyl At least on my PRs, rebasing on latest master doesn't get rid of the check. I think it applies to all PRs created before TravisCI was removed. Or perhaps, there are some settings on GitHub/TravisCI that need to be updated.

jakesyl:
I think @Roasbeef might have to remove the check

torkelrogstad:
> We might have to update this branch from master to get rid of the travis health check

I rebased this right before the weekend, without it removing the Travis check.

Rjected:
In #1575 the `.travis.yml` file was removed from the repo without the status check being removed as well, so unless we add remove the status check, nothing, including updating the branch, will get travis to run.

Rjected:
Should be good now!

PR #1559, opened 2020-04-09
invalid json field name

johnnyluo:


jakesyl:
`rawtx` is the incorrect marshall. Please [see here](https://bitcoincore.org/en/doc/0.19.0/rpc/blockchain/getblock/)

PR #1560, opened 2020-04-09
Change getblock default verbosity to 1

henryperson:
This change makes btcd's getblock command match bitcoind's. Previously
the default verbosity was 0, which caused errors when using the
rpcclient library to connect to a bitcoind node - getblock would
unmarshall incorrectly since it didn't expect a verbosity=1 result when
it did not specify verbosity.

Also includes a couple other related changes that I noticed were out of date after https://github.com/btcsuite/btcd/pull/1529.


rpcserverhelp.go
```diff
@@ -243,8 +243,8 @@ var helpDescsEnUS = map[string]string{
 	"getblockverboseresult-version":           "The block version",
 	"getblockverboseresult-versionHex":        "The block version in hexadecimal",
 	"getblockverboseresult-merkleroot":        "Root hash of the merkle tree",
-	"getblockverboseresult-tx":                "The transaction hashes (only when verbosetx=false)",
-	"getblockverboseresult-rawtx":             "The transactions as JSON objects (only when verbosetx=true)",
+	"getblockverboseresult-tx":                "The transaction hashes (only when verbosity=1)",
+	"getblockverboseresult-rawtx":             "The transactions as JSON objects (only when verbosity=2)",
```
Roasbeef, line 247: 2?

henryperson: When `verbosity=2`, the node returns information about each transaction in the block (as opposed to just the hashes of those transactions).

Per the `getblock` command in the [bitcoin RPC docs](https://bitcoincore.org/en/doc/0.19.0/rpc/blockchain/getblock/):

```
If verbosity is 0, returns a string that is serialized, hex-encoded data for block 'hash'.
If verbosity is 1, returns an Object with information about block <hash>.
If verbosity is 2, returns an Object with information about block <hash> and information about each transaction. 
```

PR #1561, opened 2020-04-11
Add signrawtransactionwithwallet command interface

henryperson:
Adds interface for issuing a [signrawtransactionwithwallet](https://bitcoincore.org/en/doc/0.19.0/rpc/wallet/signrawtransactionwithwallet/) command. It's heavily based on the code for signrawtransaction, which has been deprecated as of Bitcoin Core v0.17.0. Note that this does not add functionality for the btcd RPC server itself, it simply assumes that the RPC client one is connected to has this command already and gives an API for issuing the command. In other words, this works with a bitcoind client but not a btcd client.


jakesyl:
Hey, any chance you could force push here so we can get this merged? Github has an issue when required tests are swapped

henryperson:
Force push where?

jakesyl:
I'm good to merge once we rebase into a single commit

henryperson:
Sorry this is a workflow I'm not used to and I don't want to mess anything up. You want me to rebase everything onto the oldest commit here? Looks like the oldest commit on the PR is [e82643f](https://github.com/btcsuite/btcd/pull/1561/commits/e82643f48ab8997d444c48d8aa3e950e250c47f2) so given your request I would take its parent [57d44d0](https://github.com/btcsuite/btcd/commit/57d44d022ecb4edac34e1ea7dbd62c7a592d3d58) and run `git rebase -i 57d44d0`, which tries to rebase 66 commits. Let me know if that's right.

rpcclient/rawtransactions.go
```diff
@@ -582,6 +582,149 @@ func (c *Client) SignRawTransaction4(tx *wire.MsgTx,
...
+func (c *Client) SignRawTransactionWithWallet(tx *wire.MsgTx) (*wire.MsgTx, bool, error) {
+	return c.SignRawTransactionWithWalletAsync(tx).Receive()
+}
+
+// SignRawTransactionWithWallet2Async returns an instance of a type that can be
+// used to get the result of the RPC at some future time by invoking the Receive
+// function on the returned instance.
+//
+// See SignRawTransactionWithWallet2 for the blocking version and more details.
+func (c *Client) SignRawTransactionWithWallet2Async(tx *wire.MsgTx,
```
onyb, line 656: I see that coming up with the right function names is a bit tricky here. The issue with `SignRawTransactionWithWallet#` is that you need to read the documentation of all the variants to know which one to use.

I suggest using something more verbose, even if it makes the name inconveniently long. Something like:
* `SignRawTransactionWithWallet`
* `SignRawTransactionWithWalletUsingInputs`
* `SignRawTransactionWithWalletUsingInputsSighash`

onyb (changes requested):

onyb:
@henryperson Rebase won't be easy in your case, due to some unnecessary commits in your branch. I suggest you clean up your branch in the following way:

```
$ git checkout master  # make sure master is up-to-date
$ git branch -D add-signrawtransactionwithwallet
$ git checkout -b add-signrawtransactionwithwallet
$ git cherry-pick e0afe48fdcf96508fe33347230a24c28be69af81
$ git push -f
```

rpcclient/rawtransactions.go
onyb: Actually after seeing `SignRawTransaction#`, I'm kind of split. 🙂 So maybe it's better to follow the current convention, even if it's not so great.

PR #1567, opened 2020-04-24
Add addressType to GenNewAddress

michel-foucault:
- allow to create bech32 addresses


torkelrogstad:
Looks like this needs a rebase.

michel-foucault:
Sure, I have base my production code on tagged version of `btcd`
I see the conflict with the params for the client on the master.
If adding the optional argument is acceptable (and if no other methods are in the same case) I will rebase and test the changes

torkelrogstad:
I think is a nice change (note that I don't have merge permissions). One suggestion: add a `btcjson.AddressType` type that represents the different address types allowed (`legacy`, `p2sh-segwit` and `bech32`)

onyb:
@michel-foucault For the `btcjson.AddressType` suggestion, here's an example of [`EstimateSmartFeeMode`](https://github.com/btcsuite/btcd/blob/8b54b0b96418b33c1e584c84c0ec5141647306e0/btcjson/walletsvrcmds.go#L84-L98) to help you.

onyb:
@michel-foucault Any update on this?

michel-foucault:
Hi, we used direct rpc call for our need.
I can update the branch and finalize it if you need

btcjson/walletsvrcmds_test.go
```diff
@@ -241,13 +241,27 @@ func TestWalletSvrCmds(t *testing.T) {
...
 				Account: btcjson.String("acct"),
 			},
 		},
+		{
+			name: "getnewaddress optional addressType",
+			newCmd: func() (interface{}, error) {
+				return btcjson.NewCmd("getnewaddress", "acct", "bech32")
+			},
+			staticCmd: func() interface{} {
+				return btcjson.NewGetNewAddressCmd(btcjson.String("acct"), btcjson.String("bech32"))
```
onyb, line 257: Can you please add another test for the case when the label is `nil` and the address type is specified?

```go
btcjson.NewGetNewAddressCmd(nil, btcjson.String("bech32"))
```

onyb (changes requested):
It would be great to have your changes in since they're mostly ready.

**Summary of things to do for your reference:**
- [ ] Rebase and fix conflicts.
- [ ] Use an enum for `AddressType` field.
- [ ] Add a test for the case when account is unspecified, but address type is provided.

PR #1574, opened 2020-05-13
[connmgr] remove pending entry on failed connection

hxw:
fix memory leak in connmanager caused by:
* pending entries are never removed on error
* new connection always allocates new struct

Signed-off-by: Christopher Hall <hsw@bitmark.com>

### DIFF
--- a/connmgr/connmanager.go
+++ b/connmgr/connmanager.go
@@ -328,7 +328,11 @@ out:
 					connReq.updateState(ConnPending)
 					log.Debugf("Reconnecting to %v",
 						connReq)
-					pending[msg.id] = connReq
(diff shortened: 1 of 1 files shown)


hxw:
Initial testing was leaking memory in connmanager as >3MB/hour and was 250MB after a couple of days testing.
After fix only see 0.5M allocated in connmanager after 45 hours testing.

I am not currently sure how to fix the broken test.  I think handleFailedConnection needt to be refactored as it does bot retrying a permanent connection and allocating new items to the pending has map.

hxw:
I may have found a better solution, will know after a few hours testing

PR #1576, opened 2020-05-14
Fix memory leak in connmanager

hxw:
In the case of an error the connection struct is never deleted from the **pending** hash map.

In some testing I observed leaks of 3MB in an hour

The problem stem from the fact that the handleFailedConn does not reuse the connect but allocate a new one


jcvernaleo (changes requested):
Needs rebase and squash but otherwise looks good.

jcvernaleo (approved):
Looks good now!

OK

PR #1577, opened 2020-05-14
rpcclient: send legacy GetBlock request for backwards compatibility

wpaulino:
Without this, users of this library wouldn't be able to issue GetBlock requests to nodes which haven't updated to support the latest request format, namely the use of a single `int` parameter to denote verbosity instead of two `bool`s.

I've confirmed through manual testing that these changes, along with https://github.com/btcsuite/btcd/pull/1560/, provide full compatibility for bitcoind and btcd and provide backwards compatibility for btcd.

### DIFF
--- a/rpcclient/chain.go
+++ b/rpcclient/chain.go
@@ -52,14 +52,61 @@ func (c *Client) GetBestBlockHash() (*chainhash.Hash, error) {
 	return c.GetBestBlockHashAsync().Receive()
 }
 
+// legacyGetBlockRequest constructs and sends a legacy getblock request which
+// contains two separate bools to denote verbosity, in contract to a single int
+// parameter.
+func (c *Client) legacyGetBlockRequest(hash string, verbose,
+	verboseTx bool) ([]byte, error) {
+
+	hashJSON, err := json.Marshal(hash)
+	if err != nil {
+		return nil, err
+	}
+	verboseJSON, err := json.Marshal(btcjson.Bool(verbose))
+	if err != nil {
+		return nil, err
+	}
+	verboseTxJSON, err := json.Marshal(btcjson.Bool(verboseTx))
+	if err != nil {
+		return nil, err
+	}
+	return c.RawRequest("getblock", []json.RawMessage{
+		hashJSON, verboseJSON, verboseTxJSON,
+	})
+}
+
+// waitForGetBlockRes waits for the response of a getblock request. If the
+// response indicates an invalid parameter was provided, a legacy style of the
+// request is resent and its response is returned instead.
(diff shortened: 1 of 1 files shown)


Roasbeef:
Needs a rebase with the dependent PR merged.

rpcclient/chain.go
```diff
@@ -182,11 +245,18 @@ func (c *Client) GetBlockVerboseTxAsync(blockHash *chainhash.Hash) FutureGetBloc
 	if blockHash != nil {
 		hash = blockHash.String()
 	}
+
 	// From the bitcoin-cli getblock documentation:
-	// "If verbosity is 2, returns an Object with information about block  and information about each transaction."
+	//
+	// If verbosity is 2, returns an Object with information about block
+	// and information about each transaction."
```
Roasbeef, line 252: Dangling quote here.

wpaulino:
@Roasbeef done.

rpcclient/chain.go
```diff
@@ -155,10 +214,14 @@ func (c *Client) GetBlockVerbose(blockHash *chainhash.Hash) (*btcjson.GetBlockVe
 	return c.GetBlockVerboseAsync(blockHash).Receive()
 }
 
-type FutureGetBlockVerboseTxResult chan *response
+type FutureGetBlockVerboseTxResult struct {
```
Roasbeef, line 217: Missing `godoc` comment (lingering fwiw).

rpcclient/chain.go
```diff
@@ -52,14 +52,61 @@ func (c *Client) GetBestBlockHash() (*chainhash.Hash, error) {
...
+	return c.RawRequest("getblock", []json.RawMessage{
+		hashJSON, verboseJSON, verboseTxJSON,
+	})
+}
+
+// waitForGetBlockRes waits for the response of a getblock request. If the
+// response indicates an invalid parameter was provided, a legacy style of the
+// request is resent and its response is returned instead.
+func (c *Client) waitForGetBlockRes(respChan chan *response, hash string,
+	verbose, verboseTx bool) ([]byte, error) {
```
Roasbeef, line 82: Marginal, but we could collapse these two bools into an enum that describes target return type (the 4 options in this case, though we only use 3 of them rn). I guess non-verbose block, but verson tx isn't implemented on the RPC level for btcd/bitcoind?

wpaulino: > I guess non-verbose block, but verson tx isn't implemented on the RPC level for btcd/bitcoind?

Correct.

> Marginal, but we could collapse these two bools into an enum that describes target return type

We would still have to deconstruct them into two bools since that's what the server expects, so doesn't look like it'd help much?

PR #1579, opened 2020-05-16
rpcclient: Add getblockfilter JSON-RPC client command

lindlof:
Adds `getblockfilter` call to JSON-RPC client.

### DIFF
--- a/btcjson/chainsvrcmds.go
+++ b/btcjson/chainsvrcmds.go
@@ -163,6 +163,33 @@ func NewGetBlockCountCmd() *GetBlockCountCmd {
 	return &GetBlockCountCmd{}
 }
 
+// FilterTypeName defines the type used in the getblockfilter JSON-RPC command for the
+// filter type field.
+type FilterTypeName string
+
+const (
+	// FilterTypeBasic is the basic filter type defined in BIP0158.
+	FilterTypeBasic FilterTypeName = "basic"
+)
+
+// GetBlockFilterCmd defines the getblockfilter JSON-RPC command.
+type GetBlockFilterCmd struct {
+	BlockHash  string          // The hash of the block
+	FilterType *FilterTypeName // The type name of the filter, default=basic
+}
+
+// NewGetBlockFilterCmd returns a new instance which can be used to issue a
+// getblockfilter JSON-RPC command.
+//
+// The parameters which are pointers indicate they are optional.  Passing nil
+// for optional parameters will use the default value.
+func NewGetBlockFilterCmd(blockHash string, filterType *FilterTypeName) *GetBlockFilterCmd {
+	return &GetBlockFilterCmd{
+		BlockHash:  blockHash,
+		FilterType: filterType,
+	}
+}
+
 // GetBlockHashCmd defines the getblockhash JSON-RPC command.
 type GetBlockHashCmd struct {
 	Index int64
@@ -840,6 +867,7 @@ func init() {
 	MustRegisterCmd("getblock", (*GetBlockCmd)(nil), flags)
 	MustRegisterCmd("getblockchaininfo", (*GetBlockChainInfoCmd)(nil), flags)
 	MustRegisterCmd("getblockcount", (*GetBlockCountCmd)(nil), flags)
+	MustRegisterCmd("getblockfilter", (*GetBlockFilterCmd)(nil), flags)
 	MustRegisterCmd("getblockhash", (*GetBlockHashCmd)(nil), flags)
 	MustRegisterCmd("getblockheader", (*GetBlockHeaderCmd)(nil), flags)
 	MustRegisterCmd("getblockstats", (*GetBlockStatsCmd)(nil), flags)
--- a/btcjson/chainsvrresults.go
+++ b/btcjson/chainsvrresults.go
@@ -207,6 +207,13 @@ type GetBlockChainInfoResult struct {
 	*UnifiedSoftForks
 }
 
+// GetBlockFilterResult models the data returned from the getblockfilter
+// command.
+type GetBlockFilterResult struct {
+	Filter string `json:"filter"` // the hex-encoded filter data
+	Header string `json:"header"` // the hex-encoded filter header
+}
+
 // GetBlockTemplateResultTx models the transactions field of the
 // getblocktemplate command.
 type GetBlockTemplateResultTx struct {
--- a/btcjson/helpers.go
+++ b/btcjson/helpers.go
@@ -75,3 +75,11 @@ func String(v string) *string {
 	*p = v
 	return p
 }
+
+// NewFilterTypeName is a helper routine that allocates a new FilterTypeName value to store v and
+// returns a pointer to it.  This is useful when assigning optional parameters.
+func NewFilterTypeName(v FilterTypeName) *FilterTypeName {
+	p := new(FilterTypeName)
+	*p = v
+	return p
+}
--- a/rpcclient/chain.go
+++ b/rpcclient/chain.go
@@ -362,6 +362,44 @@ func (c *Client) GetBlockChainInfo() (*btcjson.GetBlockChainInfoResult, error) {
 	return c.GetBlockChainInfoAsync().Receive()
 }
 
+// FutureGetBlockFilterResult is a future promise to deliver the result of a
+// GetBlockFilterAsync RPC invocation (or an applicable error).
+type FutureGetBlockFilterResult chan *response
+
+// Receive waits for the response promised by the future and returns block filter
+// result provided by the server.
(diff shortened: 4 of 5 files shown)


btcjson/chainsvrcmds.go
```diff
@@ -163,6 +163,21 @@ func NewGetBlockCountCmd() *GetBlockCountCmd {
...
 }
 
+// GetBlockFilterCmd defines the getblockfilter JSON-RPC command.
+type GetBlockFilterCmd struct {
+	BlockHash  string
+	FilterType *string
+}
+
+// NewGetBlockFilterCmd returns a new instance which can be used to issue a
+// getblockfilter JSON-RPC command.
```
onyb, line 173: We generally add the following comment whenever there are optional arguments.

```suggestion
// NewGetBlockFilterCmd returns a new instance which can be used to issue a
// getblockfilter JSON-RPC command.
//
// The parameters which are pointers indicate they are optional. Passing nil
// for optional parameters will use the default value.
```

btcjson/chainsvrresults.go
```diff
@@ -207,6 +207,13 @@ type GetBlockChainInfoResult struct {
 	*UnifiedSoftForks
 }
 
+// GetBlockFilterResult models the data returned from the getblockfilter
+// command.
+type GetBlockFilterResult struct {
+	Filter string `json:"filter"`
+	Header string `json:"header"`
```
onyb, line 214: **Nit:** Whenever possible, adding extra comments always helps, since it shows up beautifully on [pkg.go.dev](https://pkg.go.dev/github.com/onyb/btcd/btcjson?tab=doc).
```suggestion
	Filter string `json:"filter"`  // the hex-encoded filter data
	Header string `json:"header"`  // the hex-encoded filter header
```

btcjson/chainsvrcmds.go
lindlof: Right on. Added these.

btcjson/chainsvrresults.go
lindlof: Good shout makes the result easier to consume. Didn't consider pkg.go.dev

Also added comments to `GetBlockFilterCmd` fields though they're not as visible through `rpcclient`.

btcjson/chainsvrcmds.go
```diff
@@ -163,6 +163,24 @@ func NewGetBlockCountCmd() *GetBlockCountCmd {
 	return &GetBlockCountCmd{}
 }
 
+// GetBlockFilterCmd defines the getblockfilter JSON-RPC command.
+type GetBlockFilterCmd struct {
+	BlockHash  string  // The hash of the block
```
torkelrogstad, line 168: I'd suggest using the `chainhash.Hash` type here.

btcjson/chainsvrcmds.go
```diff
@@ -163,6 +163,24 @@ func NewGetBlockCountCmd() *GetBlockCountCmd {
...
+	FilterType *string // The type name of the filter, default=basic
```
torkelrogstad, line 169: Looking at the Bitcoin Core codebase, "basic" is the only known filter type there. Are there any other known types out in the wild? Might make sense to create a custom type for this instead of passing in a string.

btcjson/chainsvrcmds.go
lindlof: These `btcjson.*Cmd` structs are generally using basic types assigned through NewCmd constructors. `btcjson.assignField` function makes a conversion from command-line input to the basic type that NewCmd takes.

While it's possible to implement JSON Unmarshaler to use a more specific type as you suggest I think the idea is to keep these types as stupid as possible.

In the JSON-RPC request the type will be string that the server has to unmarshal. Just taking the string from command-line and converting it to a JSON-RPC request without any semantic validation offloads the validation to server. Doing validation client-side would result in double validation which would be more fragile and more complex to consume.

btcjson/chainsvrcmds.go
lindlof: Haven't seen others than types `basic`. Not sure how standard the filter names are but first see 
my [previous response](https://github.com/btcsuite/btcd/pull/1579#discussion_r429965540) about doing validation only server-side.

rpcclient/chain.go
```diff
@@ -362,6 +362,42 @@ func (c *Client) GetBlockChainInfo() (*btcjson.GetBlockChainInfoResult, error) {
...
+// returned instance.
+//
+// See GetBlockFilter for the blocking version and more details.
+func (c *Client) GetBlockFilterAsync(blockHash string, filterType *string) FutureGetBlockFilterResult {
+	cmd := btcjson.NewGetBlockFilterCmd(blockHash, filterType)
+	return c.sendCmd(cmd)
+}
+
+// GetBlockFilter retrieves a BIP 157 content filter for a particular block.
+func (c *Client) GetBlockFilter(blockHash string, filterType *string) (*btcjson.GetBlockFilterResult, error) {
```
lindlof, line 397: @torkelrogstad This one should be using `chainhash.Hash`. I'm on it.

btcjson/chainsvrcmds.go
lindlof: I can see that for instance in `btcjson.AddNodeCmd` there is a custom type `AddNodeSubCmd` being used but in that case the options are [in the spec](https://bitcoincore.org/en/doc/0.19.0/rpc/network/addnode/).

In case of getblockfilter [the spec](https://bitcoincore.org/en/doc/0.19.0/rpc/blockchain/getblockfilter/) only mentions `basic` which is the default anyway.

btcjson/chainsvrcmds.go
lindlof: `rpcclient.GetBlockFilter` now takes `chainhash.Hash`. `rpcclient` is where to take the Golang call and transform it to JSON.

btcjson/chainsvrcmds.go
lindlof: [BIP 157](https://github.com/bitcoin/bips/blob/master/bip-0157.mediawiki#specification) describes that initial filter types are defined in [BIP 158](https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#block-filters) which defines only the `Basic` filter.

Bitcoin Core CLI seems to allow describing any `getblockfilter` in a CLI command and will send the request server which will reject any unknown filter type name. I'd have the similar behavior in `btcctl` and just handle this as a string.

In `rpcclient` the second parameter of `rpcclient.GetBlockFilter` could have a more specific type, for instance we can call it `FilterTypeName`. For the server-side implementation of `getblockfilter` to resolve the filter for the name we can map the basic `FilterTypeName` to `wire.GCSFilterRegular`.

I'll look at how to best implement `FilterTypeName`. It seems a good idea for `rpcclient`.

torkelrogstad: Good points about validation, as well as that in this specific scenario it might not be possible to construct a good type.

Just so you understand my reasoning: custom types are (at least for me) very beneficial when coding with `btcd` packages as a library, because they limit the scope of possibilities and place you on a narrower path of "valid actions". Does this make sense?

lindlof: Agreed definitely should have types for the `rpcserver` API. Sorry about the fuss - with the way you suggested the CLI mechanism works just as well (facepalm)

Please see the new commit 9d7bd87

btcjson/helpers.go
```diff
@@ -75,3 +75,11 @@ func String(v string) *string {
...
 	return p
 }
+
+// NewFilterTypeName is a helper routine that allocates a new FilterTypeName value to store v and
+// returns a pointer to it.  This is useful when assigning optional parameters.
+func NewFilterTypeName(v FilterTypeName) *FilterTypeName {
+	p := new(FilterTypeName)
+	*p = v
+	return p
+}
```
torkelrogstad, line 85: Was about to comment that it might make sense to make the `const FilterTypeBasic` a `var` so it would be possible to do `&FilterTypeBasic`, but seems like you already anticipated this and added a nice solution:-)

lindlof: Wasn't super happy about adding helper for this but it allows using `const` instead of `var` and is pretty consistent with how `rpcclient` API is used

lindlof:
@onyb @torkelrogstad Squashed ready to merge

lindlof:
> The rpcclient implementation looks good, but since our existing `getcfilter` does the exact same thing, I think we can just alias `getblockfilter` to the existing code for GetCFilterCmd, and skip rewriting filter types.

@Rjected `GetCFilterCmd` takes byte `filtertype` and `GetBlockFilterCmd` takes string `filtertype`. Not sure if there is any sensible way to reuse `GetCFilterCmd`. The response type is different too. Do miss something?

PR #1584, opened 2020-05-23
rpcclient: Implement importmulti JSON-RPC client command

onyb:
### Background reading

Bitcoin Core [**`importmulti` (0.19.0 RPC)**](https://bitcoincore.org/en/doc/0.19.0/rpc/wallet/importmulti)

### Design rationale

This RPC command is particularly complex, due to some parameters allowing values of different types, depending upon the expected outcome. Ones that are of interest are the `timestamp`, `range`, and `scriptPubKey` fields.

To solve the above problem, I drew inspiration from [**`HashOrHeight`**](https://github.com/btcsuite/btcd/blob/d38279ee7446d9ee17cb7242afb5ccb32c3775ec/btcjson/chainsvrcmds.go#L194-L221) in `btcjson` package, which seemed like a reasonably good solution to me.

### Miscelleneous

- I have added the documentation verbatim from Bitcoin Core, with slight changes wherever I felt necessary.
- I played around with the RPC command a bit, communicating to a Bitcoin Core regtest node, and although not exhaustive, it seems to be working as expected.
- I introduced a file called `rpcclient/example_test.go` in #1578, which hasn't been merged yet. I can add an example for `ImportMulti` as well, once the new file is available on master.

### DIFF
--- a/btcjson/walletsvrcmds.go
+++ b/btcjson/walletsvrcmds.go
@@ -7,6 +7,11 @@
 
 package btcjson
 
+import (
+	"encoding/json"
+	"fmt"
+)
+
 // AddMultisigAddressCmd defines the addmutisigaddress JSON-RPC command.
 type AddMultisigAddressCmd struct {
 	NRequired int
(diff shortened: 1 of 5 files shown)


btcjson/walletsvrcmds.go
```diff
@@ -677,6 +682,205 @@ func NewWalletPassphraseChangeCmd(oldPassphrase, newPassphrase string) *WalletPa
...
 }
 
+// Timestamp defines a type to represent a timestamp value in seconds,
+// since epoch.
+//
+// The value can either be a integer, or a string (ex, "now").
+//
+// NOTE: Interpretation of the timestamp value depends upon the specific
+// JSON-RPC command, where it is used.
+type Timestamp struct {
```
lindlof, line 692: Should it be called `TimestampOrNow` as this can either be an integer timestamp or the string "now"?

Other option would be the more generic `TimestampOrString` but I think it would be a good idea to specify that the string value must be "now" as required in the spec.

btcjson/walletsvrcmds.go
```diff
@@ -721,4 +925,5 @@ func init() {
 	MustRegisterCmd("walletlock", (*WalletLockCmd)(nil), flags)
 	MustRegisterCmd("walletpassphrase", (*WalletPassphraseCmd)(nil), flags)
 	MustRegisterCmd("walletpassphrasechange", (*WalletPassphraseChangeCmd)(nil), flags)
+	MustRegisterCmd("importmulti", (*ImportMultiCmd)(nil), flags)
```
lindlof, line 928: Other `MustRegisterCmd` lines in this function are in alphabetical order. This `MustRegisterCmd("importmulti"` line should be before the `MustRegisterCmd("importprivkey"` line

btcjson/walletsvrresults.go
```diff
@@ -159,3 +159,19 @@ type GetBestBlockResult struct {
...
+}
+
+// ImportMultiResults is a slice that models the result of the importmulti command.
+//
+// Each item in the slice contains the execution result corresponding to the input
+// requests of type btcjson.ImportMultiRequest, passed to the ImportMulti[Async]
+// function.
+type ImportMultiResults []struct {
+	Success  bool              `json:"success"`
+	Error    *importMultiError `json:"error,omitempty"`
```
lindlof, line 175: Looking at Bitcoin Core implementation this `error` element is the standard JSON-RPC Error `btcjson.RPCError`

rpcclient/wallet.go
```diff
@@ -2317,3 +2317,41 @@ func (c *Client) GetInfo() (*btcjson.InfoWalletResult, error) {
 // DUMP
 // importwallet (NYI in btcwallet)
 // dumpwallet (NYI in btcwallet)
+
+// FutureImportMultiResult is a future promise to deliver the result of an
+// ImportMultiAsync RPC invocation (or an applicable error).
+type FutureImportMultiResult chan *response
```
lindlof, line 2323: This is under the header `Miscellaneous Functions` in `wallet.go` file. I think `importmulti` should fall under the hear `Dump/Import Functions` which is lines 2060-2263.

Within a header each command seems to be in alphabetical order. In this fashion the `importmulti` functions will fall between the `importaddress` and `importprivkey` functions.

btcjson/walletsvrcmds.go
onyb: Indeed. I didn't pay attention to the order.

btcjson/walletsvrcmds.go
onyb: Totally agree. Renamed to `TimestampOrNow` and added a small check to make sure the string is always `"now"`.

rpcclient/wallet.go
onyb: Thanks. Will keep in mind in future.

btcjson/walletsvrresults.go
onyb: Fixed. 👍

btcjson/walletsvrcmds.go
lindlof: Awesome the check is a good idea 👌

onyb:
@jakesyl Added an example to `example_test.go`.  👍

jakesyl:
LGTM!

PR #1590, opened 2020-05-31
Add getnodeaddresses JSON-RPC support (rpc client & server)

lindlof:
Add NodeAddresses function to rpcserverConnManager
interface for fetching known node addresses.

### DIFF
--- a/rpcclient/net.go
+++ b/rpcclient/net.go
@@ -281,6 +281,43 @@ func (c *Client) GetNetworkInfo() (*btcjson.GetNetworkInfoResult, error) {
 	return c.GetNetworkInfoAsync().Receive()
 }
 
+// FutureGetNodeAddressesResult is a future promise to deliver the result of a
+// GetNodeAddressesAsync RPC invocation (or an applicable error).
+type FutureGetNodeAddressesResult chan *response
+
+// Receive waits for the response promised by the future and returns data about
+// known node addresses.
+func (r FutureGetNodeAddressesResult) Receive() ([]btcjson.GetNodeAddressesResult, error) {
+	res, err := receiveFuture(r)
+	if err != nil {
+		return nil, err
+	}
+
+	// Unmarshal result as an array of getnodeaddresses result objects.
+	var nodeAddresses []btcjson.GetNodeAddressesResult
+	err = json.Unmarshal(res, &nodeAddresses)
+	if err != nil {
+		return nil, err
+	}
+
+	return nodeAddresses, nil
+}
+
+// GetNodeAddressesAsync returns an instance of a type that can be used to get the
+// result of the RPC at some future time by invoking the Receive function on the
+// returned instance.
+//
+// See GetNodeAddresses for the blocking version and more details.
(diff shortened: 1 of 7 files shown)


rpcserver.go
```diff
@@ -2476,6 +2477,40 @@ func handleGetNetworkHashPS(s *rpcServer, cmd interface{}, closeChan <-chan stru
...
+				Code:    btcjson.ErrRPCInvalidParameter,
+				Message: "Address count out of range",
+			}
+		}
+	}
+
+	nodes := s.cfg.ConnMgr.NodeAddresses()
+	if int32(len(nodes)) < count {
+		count = int32(len(nodes))
+	}
```
onyb, line 2498: Will an extra assignment be better than traversing `nodes` twice?

```suggestion
	if numOfNodes := int32(len(nodes)); if  numOfNodes < count {
		count = numOfNodes
	}
```

rpcclient/net.go
```diff
@@ -281,6 +281,43 @@ func (c *Client) GetNetworkInfo() (*btcjson.GetNetworkInfoResult, error) {
...
+// returned instance.
+//
+// See GetNodeAddresses for the blocking version and more details.
+func (c *Client) GetNodeAddressesAsync(count *int32) FutureGetNodeAddressesResult {
+	cmd := btcjson.NewGetNodeAddressesCmd(count)
+	return c.sendCmd(cmd)
+}
+
+// GetNodeAddresses returns data about known node addresses.
+func (c *Client) GetNodeAddresses(count *int32) ([]btcjson.GetNodeAddressesResult, error) {
```
onyb, line 317: There are plenty of examples in rpcclient where each optional parameter has its own dedicated method. For example, instead of `GetNodeAddresses(count *int32)`, we can have the following:
- `GetNodeAddresses()`
- `GetNodeAddressesCount(count int32)`

What do you think?

rpcserver.go
lindlof: Implementations of `len` should be fast because slice knows its length. I do like the reduced repetition of your suggestion especially since `int32` conversion is duplicated here. I'm keen to name this variable something short like `nn` given its narrow scope.

rpcclient/net.go
lindlof: Thank you for bringing this up. After getting similar feedback in another PR I was looking around `rpcclient`. In RPC calls that take a single optional variable I can see many examples where a single function which takes a pointer is provided - similarly to `GetNodeAddresses(count *int32)`

Trying to be consistent is the best solution I think. I can change it to the non-pointer style that you're suggesting if you think that's better.

onyb: Fine with me as long as we are consistent. 👍

onyb:
@lindlof Could you please rebase this? Will have this merged tomorrow. 🙂

jcvernaleo:
Rebase/conflict fix and then this is good to go in.

lindlof:
@onyb @jcvernaleo Sweet, rebased

jcvernaleo (approved):
OK

Issue #1591, opened 2020-06-10
btcjson command parsing does not support null for optional int32


Rjected:
As discussed in #1571, it turns out `null` is not valid for optional `*int32` parameters. There are definitely RPC commands with optional ints such as `getchaintxstats` (implemented in #1571) and `getnodeaddresses` (implemented in #1590), and we should support entering `null` on the command line for these parameters, since they're optional.

cc @lindlof, add any other details if you'd like

Rjected:
Fixed by #1594, compatibility issues introduced were fixed by #1659.

PR #1594, opened 2020-06-14
Nullable optional JSON-RPC parameters

lindlof:
### Command marshalling null value
Fixes command marshalling dropping params that follow params with `nil` value. For instance the command:
```
btcjson.NewGetNetworkHashPSCmd(nil, btcjson.Int(2000))
```
Would previously marshall with empty params: `"params":[]`. Now the second optional parameter is included and padded with a `null` value: `"params":[null,2000]`.

The JSON-RPC server will substitute `null` values with the command's default values where possible.

### Command line null value
#1591 Allows providing `null` value from command line (similarly to Core command line utility):
```
btcctl getnetworkhashps null 10000
```

### Rationale

`getchaintxstats` is a good example where the ability to provide `null` parameter is especially critical. Default value of the first optional parameter is only known server-side. The fix in marshalling allows taking full advantage of default values of all commands in both `rpcclient` and command line.

In command line the ability to provide `null` not only unlocks all default values but also simplifies and allows for cleaner CLI usage.

This implementation does seem right however I haven't used Golang `reflect` all that much.

### DIFF
--- a/btcjson/cmdparse.go
+++ b/btcjson/cmdparse.go
@@ -16,19 +16,20 @@ import (
 func makeParams(rt reflect.Type, rv reflect.Value) []interface{} {
 	numFields := rt.NumField()
 	params := make([]interface{}, 0, numFields)
+	lastParam := -1
 	for i := 0; i < numFields; i++ {
 		rtf := rt.Field(i)
 		rvf := rv.Field(i)
+		params = append(params, rvf.Interface())
 		if rtf.Type.Kind() == reflect.Ptr {
 			if rvf.IsNil() {
-				break
+				// Omit optional null params unless a non-null param follows
+				continue
 			}
-			rvf.Elem()
 		}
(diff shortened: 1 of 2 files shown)


btcjson/cmdparse.go
```diff
@@ -16,19 +16,20 @@ import (
...
 		rtf := rt.Field(i)
 		rvf := rv.Field(i)
+		params = append(params, rvf.Interface())
 		if rtf.Type.Kind() == reflect.Ptr {
 			if rvf.IsNil() {
-				break
+				// Omit optional null params unless a non-null param follows
+				continue
 			}
-			rvf.Elem()
```
lindlof, line 26: Can't see this invocation having any effect so it was removed to clean it up. It doesn't modify `rvf` in any way and it shouldn't panic here because field type is pointer.

jcvernaleo (approved):
OK

PR #1595, opened 2020-06-18
Add getbalances RPC client command

federicobond:


btcjson/walletsvrresults.go
```diff
@@ -159,3 +159,17 @@ type GetBestBlockResult struct {
 	Hash   string `json:"hash"`
 	Height int32  `json:"height"`
 }
+
+// BalancesResult models the details data from the `getbalances` command.
+type BalancesResult struct {
```
onyb, line 164: Maybe call this `BalanceDetails` or something else, in order to differentiate it from `GetBalancesResult`. What do you think?

federicobond: Sounds good! I have renamed it to `BalanceDetailsResult` as many other structs seem to keep the *Result prefix.

PR #1596, opened 2020-06-18
btcjson,rpcclient: add support for PSBT commands to rpcclient

federicobond:


Rjected:
Hello, you'll need to fix the CI issues in order for this to get reviewed and merged. Thanks for the contribution!

federicobond:
@Rjected thanks for the heads up! Yeah, its weird though, I could not find out what exactly is it complaining about, the file appears formatted correctly.

Rjected:
Hey, sorry for getting back so late. Here are the changes you'd need to do to fix the formatting issues: https://github.com/Rjected/btcd/commit/23478ac91c0a5caa12bb6d4390c58160fe7f3881 . There are also now some conflicts, make sure those are fixed as well and then we can start reviewing!

federicobond:
Ah, that makes sense. Thanks for the heads up.

federicobond:
This is now ready for review.

rpcclient/wallet.go
```diff
@@ -2367,6 +2367,91 @@ func (c *Client) GetInfo() (*btcjson.InfoWalletResult, error) {
...
+	psbt string, sign *bool, sighashType *string, bip32Derivs *bool,
+) FutureWalletProcessPsbtResult {
+	cmd := btcjson.NewWalletProcessPsbtCmd(psbt, sign, sighashType, bip32Derivs)
+	return c.sendCmd(cmd)
+}
+
+// WalletProcessPsbt updates a PSBT with input information from our wallet and
+// then signs inputs.
+func (c *Client) WalletProcessPsbt(
+	psbt string, sign *bool, sighashType *string, bip32Derivs *bool,
```
onyb, line 2450: Why not have an enum type for `sighashType`? The possible values are quite limited and clearly documented in Bitcoin Core.

```
"ALL"
"NONE"
"SINGLE"
"ALL|ANYONECANPAY"
"NONE|ANYONECANPAY"
"SINGLE|ANYONECANPAY"
```

btcjson/walletsvrresults.go
```diff
@@ -173,3 +173,18 @@ type GetBalancesResult struct {
 	Mine      BalanceDetailsResult  `json:"mine"`
 	WatchOnly *BalanceDetailsResult `json:"watchonly"`
 }
+
+// WalletCreateFundedPsbtResult models the data returned from the
+// walletcreatefundedpsbtresult command.
+type WalletCreateFundedPsbtResult struct {
+	Psbt      string  `json:"psbt"`
```
onyb, line 180: Not so familiar with psbt transaction format, but can we use `*wire.MsgTx` here just like in [`FundRawTransactionResult`](https://github.com/btcsuite/btcd/blob/1db1b6f8217f5b489505c5c7e5235b53a7484779/btcjson/chainsvrresults.go#L695-L700). Will require special unmarshalling logic. See an example [here](https://github.com/btcsuite/btcd/blob/1db1b6f8217f5b489505c5c7e5235b53a7484779/btcjson/chainsvrresults.go#L702-L732).

btcjson/walletsvrresults.go
```diff
@@ -173,3 +173,18 @@ type GetBalancesResult struct {
...
+	Fee       float64 `json:"fee"`
```
onyb, line 181: This should be `btcutil.Amount`.

btcjson/walletsvrresults.go
```diff
@@ -173,3 +173,18 @@ type GetBalancesResult struct {
...
+	ChangePos int64   `json:"changepos"`
+}
+
+// WalletProcessPsbtResult models the data returned from the
+// walletprocesspsbtresult command.
+type WalletProcessPsbtResult struct {
+	Psbt     string `json:"psbt"`
```
onyb, line 188: Same remark as above, with regards to using `*wire.MsgTx`.

btcjson/btcwalletextcmds.go
```diff
@@ -93,6 +93,84 @@ func NewRenameAccountCmd(oldAccount, newAccount string) *RenameAccountCmd {
 	}
 }
 
+// PsbtInput represents an input to include in the PSBT created by the
+// WalletCreateFundedPsbtCmd command.
+type PsbtInput struct {
+	Txid     string `json:"txid"`
+	Vout     int64  `json:"vout"`
```
onyb, line 100: `Vout` is generally `uint32` types in `btcjson`.

btcjson/btcwalletextcmds.go
```diff
@@ -93,6 +93,84 @@ func NewRenameAccountCmd(oldAccount, newAccount string) *RenameAccountCmd {
...
+	Sequence int64  `json:"sequence"`
```
onyb, line 101: Same remark about `Sequence`. Generally considered to be of type `uint32`.

btcjson/btcwalletextcmds.go
```diff
@@ -93,6 +93,84 @@ func NewRenameAccountCmd(oldAccount, newAccount string) *RenameAccountCmd {
...
+}
+
+// PsbtOutput represents an output to include in the PSBT created by the
+// WalletCreateFundedPsbtCmd command.
+type PsbtOutput map[string]float64
+
+// NewPsbtOutput returns a new instance of a PSBT output to use with the
+// WalletCreateFundedPsbtCmd command.
+func NewPsbtOutput(address string, amount float64) PsbtOutput {
```
onyb, line 110: We should use `btcutil.Amount` instead of `float64` here.

btcjson/btcwalletextcmds.go
```diff
@@ -93,6 +93,84 @@ func NewRenameAccountCmd(oldAccount, newAccount string) *RenameAccountCmd {
...
+	out := make(map[string]float64)
+	out[address] = amount
+	return out
```
onyb, line 113: I think it's possible to do it in one-step:

```go
return map[string]btcutil.Amount{address: btcutil.Amount(5)}
```

btcjson/btcwalletextcmds.go
onyb, line 110: The [documentation](https://bitcoincore.org/en/doc/0.20.0/rpc/wallet/walletcreatefundedpsbt/) in Bitcoin Core talks about a `data` field in psbt outputs. Looks like you forgot to add it.

```
       {                              (json object)
         "data": "hex",               (string, required) A key-value pair. The key must be "data", the value is hex-encoded data
       },
```

btcjson/btcwalletextcmds.go
```diff
@@ -93,6 +93,84 @@ func NewRenameAccountCmd(oldAccount, newAccount string) *RenameAccountCmd {
...
+	Outputs     []PsbtOutput
+	Locktime    *int64
+	Options     *WalletCreateFundedPsbtOpts
+	Bip32Derivs *bool
+}
+
+// NewWalletCreateFundedPsbtCmd returns a new instance which can be used to issue a
+// walletcreatefundedpsbt JSON-RPC command.
+func NewWalletCreateFundedPsbtCmd(
+	inputs []PsbtInput, outputs []PsbtOutput, locktime *int64,
```
onyb, line 143: Let's use `uint32` for `locktime` for consistency.

btcjson/btcwalletextcmds.go
```diff
@@ -93,6 +93,84 @@ func NewRenameAccountCmd(oldAccount, newAccount string) *RenameAccountCmd {
...
+	out[address] = amount
+	return out
+}
+
+// WalletCreateFundedPsbtOpts represents the optional options struct provided
+// with a WalletCreateFundedPsbtCmd command.
+type WalletCreateFundedPsbtOpts struct {
+	ChangeAddress          *string  `json:"changeAddress,omitempty"`
+	ChangePosition         *int64   `json:"changePosition,omitempty"`
+	ChangeType             *string  `json:"change_type,omitempty"`
```
onyb, line 121: Should be an enum with values `legacy`, `p2sh-segwit`, and `bech32`. Let's try to update the `FundRawTransactionOpts` struct for this if possible.

btcjson/btcwalletextcmds.go
```diff
@@ -93,6 +93,84 @@ func NewRenameAccountCmd(oldAccount, newAccount string) *RenameAccountCmd {
...
+// WalletCreateFundedPsbtCmd command.
+func NewPsbtOutput(address string, amount float64) PsbtOutput {
+	out := make(map[string]float64)
+	out[address] = amount
+	return out
+}
+
+// WalletCreateFundedPsbtOpts represents the optional options struct provided
+// with a WalletCreateFundedPsbtCmd command.
+type WalletCreateFundedPsbtOpts struct {
```
onyb, line 118: This is exactly the same as `FundRawTransactionOpts`, so wondering if we should reuse it.

I would have recommended renaming `FundRawTransactionOpts` to something generic and using it here as well, but that would be a breaking change.

@Rjected What's your opinion on this?

onyb (changes requested):
Nice work on the PR!

Most of my remarks are about types. Please also feel free to pull in some descriptions from Bitcoin Core docs wherever relevant. It doesn't need to be an exact copy, but just enough to guide the users and show it beautifully on [pkg.go.dev](https://pkg.go.dev/).

btcjson/btcwalletextcmds.go
Rjected: It's possible this could just be an alias of `FundRawTransactionOpts`, and then we could make changes to `FundRawTransactionOpts` if necessary, so that's one option but I don't think it's the best way to do it. We _could_ rename the struct, and that would be breaking. On the other hand, we don't really reuse btcjson option structs (afaik) and the `fundrawtransaction` / `walletcreatefundedpsbt` apis are probably not going to change too much. I'm thinking it might be okay if we leave this as is, and update the structs independently if there are any future additions to their options. Overall I don't think it's too big of a deal if we duplicate some code or have some identical structs that represent different options in btcjson, especially if the alternative is a breaking change. WDYT?

onyb: So it's finally down to alias vs duplicating code. I'm fine with both - the former is less code to manage, and the latter has no code coupling.

If we duplicate the code, can we agree to share some common enum types (ex: a type for `ChangeType`)?

onyb:
I'm wondering if `btcwalletextcmds.go` is the right file for defining the types. This is what the header of the file says:
https://github.com/btcsuite/btcd/blob/d2b1a06d93c9b4acc4af5f47998ee2b464dcfa4c/btcjson/btcwalletextcmds.go#L5-L6

Compare that to `walletsvrcmds.go`, where we have this:
https://github.com/btcsuite/btcd/blob/d2b1a06d93c9b4acc4af5f47998ee2b464dcfa4c/btcjson/walletsvrcmds.go#L5-L6

Using `walletsvrcmds.go` would be more appropriate.

btcjson/walletsvrresults.go
federicobond: The PSBT transaction format is different from the wire tx format. It's described in [BIP 174](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki) as consisting of key-value maps.

btcjson/btcwalletextcmds.go
federicobond: Should I alias `PsbtOutput ` to an untyped `map[string]string` to support both cases? What's your advice on modelling heterogeneous lists like this?

btcjson/btcwalletextcmds.go
federicobond: Agree with @Rjected. Even though they are share the same structure today, I don't think these structs represent semantically the same type, and they may very well vary in the future as the inputs for either one of these methods are changed. There are also likely very few reasons to want to share the same input for calling both methods.

Sharing the type for `ChangeType` makes sense though.

btcjson/walletsvrresults.go
federicobond: Fee and Amount are represented as `float64` in every other result. Using `btcutil.Amount` here would introduce an inconsistency.

federicobond:
I moved the PSBT methods to `walletsvrcmds.go` as suggested.

btcjson/walletsvrresults.go
onyb: Oops, my bad. The result structs should indeed use `float64`. It's the constructors and `rpcclient` methods that should use `btcutil.Amount`.

btcjson/btcwalletextcmds.go
Rjected: Yep, this makes sense

btcjson/btcwalletextcmds.go
federicobond: I aliased `PbstOutput` to `map[string]interface{}` which makes sense for both encoding and decoding both types of outputs.

btcjson/chainsvrcmds.go
```diff
@@ -80,11 +80,21 @@ func NewCreateRawTransactionCmd(inputs []TransactionInput, amounts map[string]fl
...
 
+// ChangeType defines the different output types to use for the change address
+// of a transaction built by the node.
+type ChangeType string
+
+var (
+	ChangeTypeLegacy     ChangeType = "legacy"
+	ChangeTypeP2SHSegWit ChangeType = "p2sh-segwit"
+	ChangeTypeBech32     ChangeType = "bech32"
+)
```
onyb, line 91: Please add comments for each of these. For example:

* ChangeTypeLegacy indicates a P2PKH change address type.
* ChangeTypeP2SHSegWit indicates a P2WPKH-in-P2SH change address type.
* ChangeTypeBech32 indicates a P2WPKH change address type.

btcjson/walletsvrcmds.go
```diff
@@ -686,6 +692,105 @@ func NewWalletPassphraseChangeCmd(oldPassphrase, newPassphrase string) *WalletPa
...
+// PsbtOutput represents an output to include in the PSBT created by the
+// WalletCreateFundedPsbtCmd command.
+type PsbtOutput map[string]interface{}
+
+// NewPsbtOutput returns a new instance of a PSBT output to use with the
+// WalletCreateFundedPsbtCmd command.
+func NewPsbtOutput(address string, amount btcutil.Amount) PsbtOutput {
+	out := make(map[string]interface{})
+	out[address] = amount.ToBTC()
+	return out
```
onyb, line 712:
```suggestion
	return PsbtOutput{address: amount.ToBTC()}
```

btcjson/walletsvrcmds.go
```diff
@@ -686,6 +692,105 @@ func NewWalletPassphraseChangeCmd(oldPassphrase, newPassphrase string) *WalletPa
...
+type SighashType string
+
+var (
+	SighashAll                SighashType = "ALL"
+	SighashNone               SighashType = "NONE"
+	SighashSingle             SighashType = "SINGLE"
+	SighashAllAnyoneCanPay    SighashType = "ALL|ANYONECANPAY"
+	SighashNoneAnyoneCanPay   SighashType = "NONE|ANYONECANPAY"
+	SighashSingleAnyoneCanPay SighashType = "SINGLE|ANYONECANPAY"
+)
```
onyb, line 773: Would be nice to have comments here as well.

btcjson/walletsvrcmds.go
```diff
@@ -686,6 +692,105 @@ func NewWalletPassphraseChangeCmd(oldPassphrase, newPassphrase string) *WalletPa
...
+	out[address] = amount.ToBTC()
+	return out
+}
+
+// NewPsbtDataOutput returns a new instance of a PSBT data output to use with
+// the WalletCreateFundedPsbtCmd command.
+func NewPsbtDataOutput(data []byte) PsbtOutput {
+	out := make(map[string]interface{})
+	out["data"] = hex.EncodeToString(data)
+	return out
```
onyb, line 720:
```suggestion
	return PsbtOutput{"data": hex.EncodeToString(data)}

```

onyb (changes requested):
Added some more minor review comments.

btcjson/btcwalletextcmds.go
onyb: No strong opinion on this - `interface{}` is fine. There are other approaches, but that would be an overkill here. To me, it seems that the type is not concrete enough to deserve customized black magic.

@jakesyl Would you like to give your inputs here?

btcjson/walletsvrcmds.go
federicobond: Actually, I just noticed that rpcclient already defines a `SigHashType` type. I think it will be more clear to use that and then convert to `*string` for the btcjson constructors (to avoid a circular import). This is consistent with the behaviour of `SignRawTransaction`.

btcjson/btcwalletextcmds.go
jakesyl: missed this. I agree - `interface{}` should be fine here

federicobond:
Fixed the conflicts, should be ready to merge.

btcjson/walletsvrresults.go
onyb: Coming back to this topic, it's true that most commands in this file use the type `float64` for fees, but we should really compare it with `FundRawTransactionResult`:
https://github.com/btcsuite/btcd/blob/23d149cbfbfa13938dcb64ed62f22250eb6d239d/btcjson/chainsvrresults.go#L710-L715

We can keep it as-is, but should plan a refactoring to convert all amount-like fields to the `btcutil.Amount` type.

btcjson/walletsvrcmds.go
onyb: `SigHashType` should've really been in `btcjson` package. But there's nothing we can do about it now, without introducing a breaking change. Let's keep it for later.

rpcclient/wallet.go
```diff
@@ -2405,6 +2405,95 @@ func (c *Client) GetInfo() (*btcjson.InfoWalletResult, error) {
...
+
+// WalletProcessPsbtAsync returns an instance of a type that can be used
+// to get the result of the RPC at some future time by invoking the Receive
+// function on the returned instance.
+//
+// See WalletProcessPsbt for the blocking version and more details.
+func (c *Client) WalletProcessPsbtAsync(
+	psbt string, sign *bool, sighashType SigHashType, bip32Derivs *bool,
+) FutureWalletProcessPsbtResult {
+	cmd := btcjson.NewWalletProcessPsbtCmd(psbt, sign, btcjson.String(string(sighashType)), bip32Derivs)
```
onyb, line 2485: Small nit.

```suggestion
	cmd := btcjson.NewWalletProcessPsbtCmd(psbt, sign, btcjson.String(sighashType.String()), bip32Derivs)
```

onyb (approved):
All good on my side!

jcvernaleo (approved):
Looks good to me.

jakesyl (approved):

rpcclient/wallet.go
```diff
@@ -2405,6 +2405,95 @@ func (c *Client) GetInfo() (*btcjson.InfoWalletResult, error) {
...
+
+// WalletProcessPsbt updates a PSBT with input information from our wallet and
+// then signs inputs.
+func (c *Client) WalletProcessPsbt(
+	psbt string, sign *bool, sighashType SigHashType, bip32Derivs *bool,
+) (*btcjson.WalletProcessPsbtResult, error) {
+	return c.WalletProcessPsbtAsync(psbt, sign, sighashType, bip32Derivs).Receive()
+}
+
 // TODO(davec): Implement
```
jakesyl, line 2497: might make sense to remove these in favor of issues, but not germane to this review

PR #1601, opened 2020-07-08
GitHub Actions: Enable Go Race detector and code coverage

onyb:
This modifies the `goclean.sh` script to run tests with the race detector enabled. It also enables code coverage, and uploads the results to coveralls.io.

Running tests with `-race` and `-cover` flags was disabled in 6487ba1 and 6788df7 respectively, due to some limits on time/goroutines being hit on Travis CI. Since we have migrated to GitHub Actions, and don't have those same limitations, it is desirable to enable them back.

I have also simplified the way we run code coverage in the `goclean.sh` file, by launching it directly with `go test  ...`. I'm using `-covermode atomic` flag, which is the recommended setting for running with the race detector.

Here's the coverage report on my fork: https://coveralls.io/builds/31944135

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -25,3 +25,8 @@ jobs:
           GO111MODULE: "on"
         run: |
           sh ./goclean.sh
+
+      - name: Send coverage
+        uses: shogo82148/actions-goveralls@v1
+        with:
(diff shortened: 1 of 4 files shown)


jcvernaleo (approved):
OK

PR #1602, opened 2020-07-09
btcec: Avoid panic in fieldVal.SetByteSlice for large inputs

### LINKED ISSUE #1170: shall we restrict the size of input in btcec.FieldVal.SetByteSlice(b []byte)?

onyb:
## Summary

The implementation has been adapted from the dcrec module in dcrd. The bug was initially fixed in decred/dcrd@3d9cda1 while transitioning to a constant time algorithm. A large set of test vectors were subsequently added in decred/dcrd@8c6b52d.

The function signature has been preserved for backwards compatibility. This means that returning whether the value has overflowed, and the corresponding test vectors have not been backported.

This fixes #1170 and closes a previous attempt to fix the bug in #1178.

## Benchmarks

With changes from #1178:

```
benchmark                              iter     time/iter
---------                              ----     ---------
BenchmarkFieldVal_SetByteSlice-8   27170966   40.60 ns/op
```

With this PR:  (~2.8 times faster)

```
benchmark                              iter     time/iter
---------                              ----     ---------
BenchmarkFieldVal_SetByteSlice-8   72301308   14.60 ns/op
```

### DIFF
--- a/btcec/field.go
+++ b/btcec/field.go
@@ -226,20 +226,24 @@ func (f *fieldVal) SetBytes(b *[32]byte) *fieldVal {
 	return f
 }
 
-// SetByteSlice packs the passed big-endian value into the internal field value
-// representation.  Only the first 32-bytes are used.  As a result, it is up to
-// the caller to ensure numbers of the appropriate size are used or the value
-// will be truncated.
+// SetByteSlice interprets the provided slice as a 256-bit big-endian unsigned
+// integer (meaning it is truncated to the first 32 bytes), packs it into the
+// internal field value representation, and returns the updated field value.
+//
+// Note that since passing a slice with more than 32 bytes is truncated, it is
+// possible that the truncated value is less than the field prime.  It is up to
+// the caller to decide whether it needs to provide numbers of the appropriate
+// size or if it is acceptable to use this function with the described
+// truncation behavior.
 //
 // The field value is returned to support chaining.  This enables syntax like:
 // f := new(fieldVal).SetByteSlice(byteSlice)
 func (f *fieldVal) SetByteSlice(b []byte) *fieldVal {
 	var b32 [32]byte
-	for i := 0; i < len(b); i++ {
-		if i < 32 {
-			b32[i+(32-len(b))] = b[i]
-		}
+	if len(b) > 32 {
+		b = b[:32]
 	}
+	copy(b32[32-len(b):], b)
 	return f.SetBytes(&b32)
 }
 
--- a/btcec/field_test.go
+++ b/btcec/field_test.go
@@ -7,6 +7,7 @@ package btcec
 
 import (
 	"crypto/rand"
+	"encoding/hex"
 	"fmt"
 	"reflect"
 	"testing"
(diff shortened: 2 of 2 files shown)


btcec/field.go
```diff
@@ -226,20 +226,37 @@ func (f *fieldVal) SetBytes(b *[32]byte) *fieldVal {
...
+// Output Normalized: Yes if no overflow, no otherwise
+// Output Max Magnitude: 1
 func (f *fieldVal) SetByteSlice(b []byte) *fieldVal {
 	var b32 [32]byte
-	for i := 0; i < len(b); i++ {
-		if i < 32 {
-			b32[i+(32-len(b))] = b[i]
-		}
+
+	if size := len(b); size >= 32 {
```
davecgh, line 250: Since you aren't constant time here anyway due to the branch, this can be simplified to:

```Go
	var b32 [32]byte
	if len(b) > 32 {
		b = b[:32]
	}
	copy(b32[32-len(b):], b)
	return f.SetBytes(&b32)
}
```

onyb: There was no noticeable change in the benchmark (see the updated PR description). Here's the benchmarking code I used:

```go
func BenchmarkFieldVal_SetByteSlice(b *testing.B) {
	for n := 0; n < b.N; n++ {
		inBytes := hexToBytes("01fffffffffffffffffffffffffffffffffffffffffffffffffffffffdfffff85c")
		var f fieldVal
		f.SetByteSlice(inBytes)
	}
}
```

I do think your version is better due to its simplicity.

davecgh: That's likely because your benchmark is being completely dominated by the hex conversion.  If you do a benchmark that only measures setting the slice, which is what you want to measure, you should see the difference.

e.g.
```Go
func BenchmarkFieldValSetByteSlice(b *testing.B) {
	inBytes := hexToBytes("01fffffffffffffffffffffffffffffffffffffffffffffffffffffffdfffff85c")
	for n := 0; n < b.N; n++ {
		var f fieldVal
		f.SetByteSlice(inBytes)
	}
}
```

onyb: It was indeed the issue! The new benchmark result is down to `14.60 ns/op`, from `40.60 ns/op`! Thanks. 🙂

btcec/field.go
```diff
@@ -226,20 +226,30 @@ func (f *fieldVal) SetBytes(b *[32]byte) *fieldVal {
...
 
-// SetByteSlice packs the passed big-endian value into the internal field value
-// representation.  Only the first 32-bytes are used.  As a result, it is up to
-// the caller to ensure numbers of the appropriate size are used or the value
-// will be truncated.
+// SetByteSlice interprets the provided slice as a 256-bit big-endian unsigned
+// integer (meaning it is truncated to the first 32 bytes), packs it into the
+// internal field value representation, and returns whether or not the resulting
+// truncated 256-bit integer is greater than or equal to the field prime (aka it
+// overflowed) in constant time.
```
davecgh, line 233: Oops, I was a bit hasty in approving.  This isn't a constant time implementation nor does it return the overflow flag.

onyb: Nice catch. Updated the docs. 👍

PR #1603, opened 2020-07-12
server: handle notfound

tuxcanfly:
Also, increase ban score on notfound and return whether `addBanScore` disconnects the peer. Backport from https://github.com/decred/dcrd/pull/2253

### DIFF
--- a/netsync/manager.go
+++ b/netsync/manager.go
@@ -79,6 +79,13 @@ type headersMsg struct {
 	peer    *peerpkg.Peer
 }
 
+// notFoundMsg packages a bitcoin notfound message and the peer it came from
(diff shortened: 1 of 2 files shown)


netsync/manager.go
```diff
@@ -79,6 +79,13 @@ type headersMsg struct {
 	peer    *peerpkg.Peer
 }
 
+// notFoundMsg packages a Decred notfound message and the peer it came from
```
jcvernaleo, line 82: It isn't really a decred message at this point so the comment is a bit misleading.

tuxcanfly: Thanks. Fixed.

PR #1604, opened 2020-07-13
blockchain: Remove unnecessary tx hash

kcalvinalvin:
txHash is only used for printing error messages and doesn't have to be calculated beforehand. This gets rid of an expensive calculation and removes a lot of gc pressure.

### DIFF
--- a/blockchain/validate.go
+++ b/blockchain/validate.go
@@ -877,7 +877,6 @@ func CheckTransactionInputs(tx *btcutil.Tx, txHeight int32, utxoView *UtxoViewpo
 		return 0, nil
 	}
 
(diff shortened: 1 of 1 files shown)


jakesyl:
LGTM!

davecgh:
There really isn't any reason not to do this, but, the stated reason really isn't true.  The `tx` variable there is a `*btcutil.Tx` which you can clearly see [here](https://github.com/btcsuite/btcutil/blob/4649e4b73b34090724fff8a302751260c20551fd/tx.go#L41-L45) accesses a cached hash, and the hash has already been cached before it gets here.

jcvernaleo (approved):
OK

PR #1605, opened 2020-07-16
Update manager.go

lilianwen:
“we are not current” should be the first condition.


netsync/manager.go
```diff
@@ -1095,7 +1095,7 @@ func (sm *SyncManager) handleInvMsg(imsg *invMsg) {
 
 	// Ignore invs from peers that aren't the sync if we are not current.
 	// Helps prevent fetching a mass of orphans.
-	if peer != sm.syncPeer && !sm.current() {
+	if !sm.current() && peer != sm.syncPeer {
```
onyb, line 1098: Can you please provide some more context explaining why this change is necessary?

PR #1608, opened 2020-07-21
updated docs for getblock-verbosity fixes

JettScythe:
This adds documentation fixes for #1529 & #1560


docs/json_rpc_api.md
```diff
@@ -271,13 +271,13 @@ the method name for further details such as parameter and return information.
 |   |   |
 |---|---|
 |Method|getblock|
-|Parameters|1. block hash (string, required) - the hash of the block<br />2. verbose (boolean, optional, default=true) - specifies the block is returned as a JSON object instead of hex-encoded string<br />3. verbosetx (boolean, optional, default=false) - specifies that each transaction is returned as a JSON object and only applies if the `verbose` flag is true.<font color="orange">**This parameter is a btcd extension**</font>|
+|Parameters|1. block hash (string, required) - the hash of the block<br />2. verbosity (int, optional, default=1) - specifies the block is returned as a JSON object instead of hex-encoded string<font color="orange">**This parameter is a btcd extension**</font>|
```
onyb, line 274: Two remarks here:
* We could use the same description for the `verbosity` parameter, as in the RPC server help text:
  ```
  Specifies whether the block data should be returned as a hex-encoded string (0),
  as parsed data with a slice of TXIDs (1), or as parsed data with parsed
  transaction data (2).
  ```
* > This parameter is a btcd extension

  I think this is not relevant anymore, since we're following the same API as Bitcoin Core.

JettScythe: Pushed the suggested corrections :)

PR #1610, opened 2020-07-27
release: update release README.md and remove old scripts

Rjected:
This should hopefully help document the steps to publish a release on the maintainer side of things, if there are any details left out or additional steps that should be documented, feedback is welcome!

Would like input from:
@jcvernaleo 
@Roasbeef


release/README.md
```diff
@@ -6,11 +6,52 @@ binaries are now reproducible, allowing developers to build the binary on
 distinct machines, and end up with a byte-for-byte identical binary. However,
 this wasn't _fully_ solved in `go1.13`, as the build system still includes the
 directory the binary is built into the binary itself. As a result, our scripts
-utilize a work around needed until `go1.13.2`.  
+utilize a work around needed until `go1.13.2`.
```
Roasbeef, line 9: One other thing related to this, is that all the releases should note which Go version was used to build the release so people can properly reproduce the binaries if they wish.

release/README.md
```diff
@@ -19,13 +60,43 @@ the release binaries following these steps:
...
 and a manifest file containing the hash of each archive.
 
+### Pushing a release (for maintainers)
+
+Now that the directory `btcd-<TAG>` is created, the manifest file needs to be
+signed by a maintainer and the release files need to be published to GitHub.
+
+Sign the `manifest-<TAG>.txt` file like so:
+```sh
+gpg --output manifest-<TAG>.txt.sig --sign manifest-<TAG>.txt
```
Roasbeef, line 76: I usually sign with a _detached_ sig. This produces a smaller output (as the document itself isn't included in the signature). Based on your example above, signing with a detached sig would look something like: 
```
gpg --sign --detach-sig manifest-<TAG>.txt
```

We can drop the `--output`, as it's automatically create the sig in a file name that adds `.sig` to the referenced file to be signed.

Rjected: Changed to detached sig, should be good now

release/README.md
Rjected: Should have resolved this, placed notes in both the section for maintainers and the description up top

Rjected:
> Can you please add a small note about updating the [**`CHANGES`**](https://github.com/btcsuite/btcd/blob/master/CHANGES) file?

Should be addressed, let me know what you think!

PR #1615, opened 2020-08-15
Improve gc for isunspendable method

Rjected:
Through profiling, I've found that the `IsUnspendable` method is a large source of allocations during the sync process.

In an effort to reduce the allocations made by this method, and in the sync process overall, I've simply removed the part of `parseScriptTemplate` which allocates and appends to an array (which never ends up being used by `IsUnspendable`) and pointed `IsUnspendable` to that method, called `checkScriptTemplateParseable`.

To show that this actually did reduce the allocations (and runtime) of the `IsUnspendable` method, I've created a benchmark and run it on both versions of the code, once on the [`e761244`](https://github.com/btcsuite/btcd/commit/e7612446051c390a4372d7604029b1eaff8a7f1d) commit, and once on the [`0963732`](https://github.com/btcsuite/btcd/commit/096373242c26946aabaf87f554003675aed6dfcd) commit.
I used the command on the older commit:
```
go test -v -bench=BenchmarkIsUnspendable -run=^\$ -benchmem -benchtime 10s -count 30 >> ~/unspendableold.txt
```
I then used this command on the newer commit:
```
go test -v -bench=BenchmarkIsUnspendable -run=^\$ -benchmem -benchtime 10s -count 30 >> ~/unspendablenew.txt
```

After running these, `benchstat` yielded the following result:
```
name             old time/op    new time/op    delta
IsUnspendable-8     283ns ± 3%      47ns ± 2%  -83.34%  (p=0.000 n=28+27)

name             old alloc/op   new alloc/op   delta
IsUnspendable-8      768B ± 0%       32B ± 0%  -95.83%  (p=0.000 n=30+30)

name             old allocs/op  new allocs/op  delta
IsUnspendable-8      1.00 ± 0%      1.00 ± 0%     ~     (all equal)
```
The benchtime and count in my benchmarks are unnecessarily high, I get similar results with a lower count and deleted benchtime parameter:
```
go test -v -bench=BenchmarkIsUnspendable -run=^\$ -benchmem -count 10 >> ~/unspendablenew.txt
```

This shows that in at least my benchmarks, the allocations and runtime of this method were significantly reduced.
In the future we should definitely backport changes from [dcrd #1656](https://github.com/decred/dcrd/pull/1656) (I will make an issue for this), but that would be a massive PR to consensus critical code, would likely need to be adapted, and the improvements for this method seem good enough for a single, easily reviewable PR.
The point of this PR is to:
1. Improve allocations of `IsUnspendable`.
2. Show that there is significant work to be done on the `txscript` package, and it would go a very long way towards improving the sync time if we focus on performance using benchmarking and profiling tools.

This is a draft right now because a test isn't passing but this will be fixed soon.

Backporting the benchmark changes would be a great future addition in itself, and hopefully this PR highlights this.

### DIFF
--- a/txscript/script.go
+++ b/txscript/script.go
@@ -196,86 +196,58 @@ func IsPushOnlyScript(script []byte) bool {
 // the list of parsed opcodes up to the point of failure along with the error.
 func parseScriptTemplate(script []byte, opcodes *[256]opcode) ([]parsedOpcode, error) {
 	retScript := make([]parsedOpcode, 0, len(script))
+	var err error
 	for i := 0; i < len(script); {
 		instr := script[i]
 		op := &opcodes[instr]
 		pop := parsedOpcode{opcode: op}
-
-		// Parse data out of instruction.
-		switch {
-		// No additional data.  Note that some of the opcodes, notably
-		// OP_1NEGATE, OP_0, and OP_[1-16] represent the data
-		// themselves.
-		case op.length == 1:
-			i++
-
-		// Data pushes of specific lengths -- OP_DATA_[1-75].
-		case op.length > 1:
-			if len(script[i:]) < op.length {
-				str := fmt.Sprintf("opcode %s requires %d "+
-					"bytes, but script only has %d remaining",
-					op.name, op.length, len(script[i:]))
-				return retScript, scriptError(ErrMalformedPush,
-					str)
-			}
-
-			// Slice out the data.
-			pop.data = script[i+1 : i+op.length]
-			i += op.length
-
-		// Data pushes with parsed lengths -- OP_PUSHDATAP{1,2,4}.
-		case op.length < 0:
-			var l uint
-			off := i + 1
-
-			if len(script[off:]) < -op.length {
-				str := fmt.Sprintf("opcode %s requires %d "+
-					"bytes, but script only has %d remaining",
-					op.name, -op.length, len(script[off:]))
-				return retScript, scriptError(ErrMalformedPush,
-					str)
-			}
-
-			// Next -length bytes are little endian length of data.
-			switch op.length {
-			case -1:
-				l = uint(script[off])
-			case -2:
-				l = ((uint(script[off+1]) << 8) |
-					uint(script[off]))
-			case -4:
-				l = ((uint(script[off+3]) << 24) |
-					(uint(script[off+2]) << 16) |
-					(uint(script[off+1]) << 8) |
-					uint(script[off]))
-			default:
-				str := fmt.Sprintf("invalid opcode length %d",
-					op.length)
-				return retScript, scriptError(ErrMalformedPush,
-					str)
-			}
-
-			// Move offset to beginning of the data.
-			off += -op.length
-
-			// Disallow entries that do not fit script or were
-			// sign extended.
-			if int(l) > len(script[off:]) || int(l) < 0 {
-				str := fmt.Sprintf("opcode %s pushes %d bytes, "+
-					"but script only has %d remaining",
-					op.name, int(l), len(script[off:]))
-				return retScript, scriptError(ErrMalformedPush,
-					str)
-			}
-
-			pop.data = script[off : off+int(l)]
-			i += 1 - op.length + int(l)
+		i, err = pop.checkParseableInScript(script, i)
+		if err != nil {
+			return retScript, err
 		}
 
 		retScript = append(retScript, pop)
 	}
 
 	return retScript, nil
+}
+
+// checkScriptTemplateParseable is the same as parseScriptTemplate but does not
+// return the list of opcodes up until the point of failure so that this can be
+// used in functions which do not necessarily have a need for the failed list of
+// opcodes, such as IsUnspendable.
+//
+// This function returns a pointer to a byte. This byte is nil if the parsing
+// has an error, or if the script length is zero. If the script length is not
+// zero and parsing succeeds, then the first opcode parsed will be returned.
+//
+// Not returning the full opcode list up until failure also has the benefit of
+// reducing GC pressure, as the list would get immediately thrown away.
+func checkScriptTemplateParseable(script []byte, opcodes *[256]opcode) (*byte, error) {
+	var err error
+
+	// A script of length zero is an unspendable script but it is parseable.
+	var firstOpcode byte
+	var numParsedInstr uint = 0
+
+	for i := 0; i < len(script); {
+		instr := script[i]
+		op := &opcodes[instr]
+		pop := parsedOpcode{opcode: op}
+		i, err = pop.checkParseableInScript(script, i)
+		if err != nil {
+			return nil, err
+		}
+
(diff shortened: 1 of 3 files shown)


Rjected:
New benchmark results for commit [`b085b02`](https://github.com/btcsuite/btcd/commit/b085b029df29bf75917ac7de762b51ca0eacda18):
```
name             old time/op    new time/op    delta
IsUnspendable-8     283ns ± 3%       9ns ± 1%   -96.82%  (p=0.000 n=28+28)

name             old alloc/op   new alloc/op   delta
IsUnspendable-8      768B ± 0%        0B       -100.00%  (p=0.000 n=30+30)

name             old allocs/op  new allocs/op  delta
IsUnspendable-8      1.00 ± 0%      0.00       -100.00%  (p=0.000 n=30+30)
```
Since the tests also pass now, I'll make this a non draft. I'd also like to note that the benchmark is not an authoritative measure of performance. We should build stuff that helps build benchmarks which represent more of a real world load, take these results with a grain of salt!

txscript/script.go
```diff
@@ -278,6 +278,107 @@ func parseScriptTemplate(script []byte, opcodes *[256]opcode) ([]parsedOpcode, e
 	return retScript, nil
 }
 
+// checkScriptTemplateParseable is the same as parseScriptTemplate but does not
+// return the list of opcodes up until the point of failure so that this can be
+// used in functions which do not necessarily have a need for the failed list of
+// opcodes, such as IsUnspendable.
+// Not returning the full opcode list up until failure also has the benefit of
```
onyb, line 285: You probably want an additional new line here.

txscript/script.go
```diff
@@ -278,6 +278,107 @@ func parseScriptTemplate(script []byte, opcodes *[256]opcode) ([]parsedOpcode, e
...
+// reducing GC pressure, as the list would get immediately thrown away.
+func checkScriptTemplateParseable(script []byte, opcodes *[256]opcode) (uint, byte, error) {
+	// A script of length zero is an unspendable script but it is parseable.
+	if len(script) == 0 {
+		return 0, 0x00, nil
+	}
```
onyb, line 291: Isn't this already taken care of by the following two lines?

```go
var firstOpcode byte = OP_0
var numParsedInstr uint = 0
```

The comment still deserves a mention.

txscript/script.go
```diff
@@ -278,6 +278,107 @@ func parseScriptTemplate(script []byte, opcodes *[256]opcode) ([]parsedOpcode, e
...
+
+	var firstOpcode byte = OP_0
```
onyb, line 293: This feels like a misuse of `OP_0`. If I understand correctly, using any opcode will work here as long as it is not `OP_RETURN`.

txscript/script.go
```diff
@@ -851,10 +952,14 @@ func getWitnessSigOps(pkScript []byte, witness wire.TxWitness) int {
...
+	if len(pkScript) == 0 {
+		return false
+	}
+	numberOfPops, firstOpcode, err := checkScriptTemplateParseable(pkScript, &opcodeArray)
 	if err != nil {
 		return true
 	}
 
-	return len(pops) > 0 && pops[0].opcode.value == OP_RETURN
+	return numberOfPops > 0 && firstOpcode == OP_RETURN
```
onyb, line 964: How about using `(*byte, error)` as the return type of `checkScriptTemplateParseable`? The check would simply then be:

```go
firstOpcode, err := checkScriptTemplateParseable(pkScript, &opcodeArray)
if err != nil {
         return true
}

return firstOpcode != nil && *firstOpcode == OP_RETURN
```

This takes into account my other remark that if the script is empty, it doesn't make sense for the first opcode to be `OP_0`.

txscript/script.go
```diff
@@ -851,10 +952,14 @@ func getWitnessSigOps(pkScript []byte, witness wire.TxWitness) int {
 // guaranteed to fail at execution.  This allows inputs to be pruned instantly
 // when entering the UTXO set.
 func IsUnspendable(pkScript []byte) bool {
-	pops, err := parseScript(pkScript)
+	// Not provably unspendable
```
onyb, line 955: Is this necessary? It appears to me that it is already taken care of in `checkScriptTemplateParseable`.

txscript/script_test.go
```diff
@@ -4301,6 +4301,28 @@ func TestIsUnspendable(t *testing.T) {
...
+			expected: false,
+		},
+		{
+			// Spendable
+			pkScript: []byte{0x51},
+			expected: false,
+		},
+		{
+			// Unspendable
+			pkScript: []byte{0x6a},
```
onyb, line 4323:
```suggestion
			pkScript: []byte{OP_RETURN},
```

txscript/script_test.go
```diff
@@ -4301,6 +4301,28 @@ func TestIsUnspendable(t *testing.T) {
...
+			expected: false,
+		},
+		{
+			// Not Necessarily Unspendable
+			pkScript: []byte{},
+			expected: false,
+		},
+		{
+			// Spendable
+			pkScript: []byte{0x51},
```
onyb, line 4318:
```suggestion
			pkScript: []byte{OP_TRUE},
```

txscript/script_test.go
```diff
@@ -4301,6 +4301,28 @@ func TestIsUnspendable(t *testing.T) {
 				0xfa, 0x0b, 0x5c, 0x88, 0xac},
 			expected: false,
 		},
+		{
+			// Spendable
+			pkScript: []byte{0xa9, 0x14, 0x82, 0x1d, 0xba, 0x94, 0xbc, 0xfb,
```
onyb, line 4306: Whenever possible, it's nice to use human-friendly names for the opcodes. It's not necessary in this example, but I added some suggestions for the other test cases you added.

txscript/script.go
```diff
@@ -278,6 +278,107 @@ func parseScriptTemplate(script []byte, opcodes *[256]opcode) ([]parsedOpcode, e
...
+	}
+
+	var firstOpcode byte = OP_0
+	var numParsedInstr uint = 0
+	for i := 0; i < len(script); {
+		instr := script[i]
+		op := &opcodes[instr]
+		pop := parsedOpcode{opcode: op}
+
+		// Parse data out of instruction.
```
onyb, line 300: The entire `switch` block appears to be duplicated in `parseScriptTemplate`. Do you think it's worth doing a refactoring?

onyb (changes requested):
utACK. Good find and quite clever. 👍

jcvernaleo (commented):
Looks great but needs to address @onyb's comments then it can go in.

jcvernaleo:
I think this needs to be rebased for the proper checks too.

jcvernaleo:
@onyb could you take another look at this one to see if your comments were addressed?

onyb (approved):

jcvernaleo (approved):
Looks good to me too now.
OK

PR #1616, opened 2020-08-16
remove else after return

tylerchambers:
going through the code base, found a small nit.
could do to remove the else and outdent after the return in the if statement.

### DIFF
--- a/cmd/btcctl/config.go
+++ b/cmd/btcctl/config.go
@@ -135,9 +135,8 @@ func normalizeAddress(addr string, chain *chaincfg.Params, useWallet bool) (stri
 				// TODO: add port once regtest is supported in btcwallet
 				paramErr := fmt.Errorf("cannot use -wallet with -regtest, btcwallet not yet compatible with regtest")
 				return "", paramErr
-			} else {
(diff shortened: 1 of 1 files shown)


cmd/btcctl/config.go
```diff
@@ -135,9 +135,8 @@ func normalizeAddress(addr string, chain *chaincfg.Params, useWallet bool) (stri
 				// TODO: add port once regtest is supported in btcwallet
 				paramErr := fmt.Errorf("cannot use -wallet with -regtest, btcwallet not yet compatible with regtest")
 				return "", paramErr
-			} else {
-				defaultPort = "18334"
 			}
+			defaultPort = "18334"
```
ipriver, line 139: Good catch, "else" statement is unnecessary here as the previous "if" statement returns an error. 
But I think there is an important "TODO" note that gives information that in the future there might be a "defaultPort" value and the "return" will be also removed. This makes me think that if you just remove "else" statement another contributor will add it back again.
Also if I'm not wrong compiler makes optimizations for such statements and there is no difference in the byte-code between having "else" and removing it, so this case only affects code readability.

PR #1618, opened 2020-08-23
add clang support to goclean.sh

Rjected:
I recently switched to a setup without gcc, only clang, so I added some checks in the `goclean.sh` script to check whether or not clang or gcc are in the PATH.

### DIFF
--- a/goclean.sh
+++ b/goclean.sh
@@ -8,6 +8,32 @@
 
 set -ex
 
+# Check if we have gcc, if not then use clang, else exit. More compilers can be
+# added if needed.
+if command -v gcc &> /dev/null
+then
+
+    export CC=gcc
+
+elif command -v clang &> /dev/null
+then
+
+    export CC=clang
+
+else
+
+    echo "neither gcc nor clang are in the PATH (they may not be installed),
+    btcd will not compile due to cgo requiring a C compiler, exiting..."
+    exit
+
+fi
+
+if ! command -v golangci-lint &> /dev/null
+then
+    echo "golangci-lint is not in the PATH, it will not run, exiting..."
+    exit
+fi
+
 go test -tags="rpctest" ./...
 
 # Automatic checks


goclean.sh
```diff
@@ -8,6 +8,32 @@
...
+# added if needed.
+if command -v gcc &> /dev/null
+then
+
+    export CC=gcc
+
+elif command -v clang &> /dev/null
+then
+
+    export CC=clang
```
onyb, line 21: This is fine, but we should still allow the user to override the C compiler using the `CC` env var. What do you think?

Rjected: That makes sense, I'll change it so that if the user has the `CC` env variable set it will use that C compiler.

goclean.sh
```diff
@@ -8,6 +8,32 @@
...
+
+fi
+
+if ! command -v golangci-lint &> /dev/null
+then
+    echo "golangci-lint is not in the PATH, it will not run, exiting..."
+    exit
+fi
+
 go test -tags="rpctest" ./...
```
ipriver, line 37: Don't think it's necessary to skip running tests if "golangci-lint" is not installed.

goclean.sh
ipriver: Not sure if there could be situations when `go env` has an empty `CC` variable, it is set to `gcc` by default even if `gcc` is not installed.
Also, current behavior gonna overwrite `CC` variable to be `gcc` for me as I'm using `CC=clang` compiler while having `gcc` and `clang` both installed.

ipriver (changes requested):

goclean.sh
Rjected: Makes sense :+1:

goclean.sh
Rjected: You're right, I should change it so that it doesn't overwrite the CC variable after the script is run. However, there are situations (such as on my current machine) where the `CC` environment variable in `go env` is set to `gcc` when `gcc` is not installed, leading to tests crashing which is the point of the PR.

ipriver: Yeah, I'm understanding your idea why you've made this PR and that's why I'm asking you to be more careful with such changes as they can break things for other users. Just make sure it's gonna be error-safe after you make it right, thanks :)

Rjected:
Closing to focus on some more important things - if these changes would help fix a reported problem then I'm willing to reopen it.

PR #1619, opened 2020-08-27
btcd: fix conversion of int to string failing in Go 1.15

federicobond:
Without this, `btcd` fails to build in Go 1.15 with the following message:
````
./upnp.go:232:20: conversion from int to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
````

### DIFF
--- a/btcjson/cmdparse_test.go
+++ b/btcjson/cmdparse_test.go
@@ -201,7 +201,7 @@ func TestAssignFieldErrors(t *testing.T) {
 	}{
 		{
 			name: "general incompatible int -> string",
-			dest: string(0),
+			dest: "\x00",
 			src:  int(0),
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: btcjson/cmdparse_test.go [hunk 1]
```diff
@@ -201,7 +201,7 @@ func TestAssignFieldErrors(t *testing.T) {
 	}{
 		{
 			name: "general incompatible int -> string",
-			dest: string(0),
+			dest: "\x00",
```
onyb, line 204: Any reason to not use `string(rune(0))` here? By the way, `"\x00"` is fine too..

federicobond: No reason, it just felt more natural to write `"\x00"` for me.

onyb (approved):
LGTM. 🚀

PR #1621, opened 2020-08-31
ci: add go 1.15 to tests

jakesyl:
Note: this requires #1619 before it can be merged.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: [1.13, 1.14]
+        go: [1.14, 1.15]
     steps:
       - name: Set up Go
         uses: actions/setup-go@v2
--- a/go.mod
+++ b/go.mod
@@ -14,4 +14,4 @@ require (
 	golang.org/x/crypto v0.0.0-20200510223506-06a226fb4e37
 )
 
-go 1.12
+go 1.14
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ which are both under active development.
 
 ## Requirements
 
-[Go](http://golang.org) 1.12 or newer.
+[Go](http://golang.org) 1.14 or newer.
 
 ## Installation
 
--- a/release/README.md
+++ b/release/README.md
@@ -3,10 +3,7 @@
 This package contains the build script that the `btcd` project uses in order to
 build binaries for each new release. As of `go1.13`, with some new build flags,
 binaries are now reproducible, allowing developers to build the binary on
-distinct machines, and end up with a byte-for-byte identical binary. However,
-this wasn't _fully_ solved in `go1.13`, as the build system still includes the
-directory the binary is built into the binary itself. As a result, our scripts
-utilize a work around needed until `go1.13.2`.
+distinct machines, and end up with a byte-for-byte identical binary.
 Every release should note which Go version was used to build the release, so
 that version should be used for verifying the release.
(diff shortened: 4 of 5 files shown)


onyb (approved):

jakesyl:
One more thing on this. Someone with permissions (likely @Roasbeef) will have to add this as a required test once we get this merged

.github/workflows/go.yml
```diff
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: [1.13, 1.14]
+        go: [1.13, 1.14, 1.15]
```
jcvernaleo, line 9: I do not think we should be testing on three versions.  The policy has always been current and current-1.  So if 1.15 is current, please drop 1.13.

jakesyl: @jcvernaleo should be good to go here

jakesyl:
looks like we'll need to remove 1.13 as a required check (and switch it to 1.15

ipriver:
JFYI
How about to also update [README.md](https://github.com/btcsuite/btcd/blob/master/README.md#requirements) requirements version? If CI is testing only the 2 latest golang versions then there's no guarantee that the project works correctly with go version < 1.14.

jakesyl:
@ipriver good call. I missed this

jakesyl:
anyone know if this is still true? 
image

go.mod
```diff
@@ -14,4 +14,4 @@ require (
 	golang.org/x/crypto v0.0.0-20200510223506-06a226fb4e37
 )
 
-go 1.12
+go 1.14
```
onyb, line 17: Why not `1.15`?

onyb (changes requested):
Assuming the next release will be built with Go 1.15, we need to remove `darwin-386` from `release/release.sh`. [See why](https://golang.org/doc/go1.15#darwin).

onyb:
@jakesyl Regarding reproducible builds, the release script is still valid for Go 1.14 / 1.15 (i.e. we still need the `-trimpath` build flag):

```
➜  btcd git:(master) ✗ go build -ldflags="-s -w -buildid=" github.com/btcsuite/btcd          
➜  btcd git:(master) ✗ go tool objdump ./btcd | grep "/Users/" | wc -l                          
    2857
➜  btcd git:(master) ✗ go build -trimpath -ldflags="-s -w -buildid=" github.com/btcsuite/btcd 
➜  btcd git:(master) ✗ go tool objdump ./btcd | grep "/Users/" | wc -l                       
       0
```

Apart from the `-trimpath`, there is no ugly workaround in the release script, we should be able to remove these lines:

```
However, this wasn't _fully_ solved in `go1.13`, as the build system still includes the
directory the binary is built into the binary itself. As a result, our scripts
utilize a work around needed until `go1.13.2`.
```

go.mod
ipriver: I think 1.14 is fine here as it's the minimal supported version.

https://golang.org/ref/mod#go-mod-file-go
`A go directive sets the expected language version for the module.`

onyb: Makes sense. 👍

jakesyl:
@onyb all set

jakesyl:
ah, sorry. Need to fix `release/release.sh`

jakesyl:
_now_ I should be all set

jakesyl:
that's new

onyb (approved):

jcvernaleo (approved):
OK

PR #1626, opened 2020-09-06
btcjson: update ListTransactionsResult for Bitcoin 0.20.0

onyb:
This only adds new fields as optional, in order to make it backwards compatible with older versions of Bitcoin Core.


jcvernaleo (approved):
OK

PR #1627, opened 2020-09-07
Remove unnecessary GOMAXPROCS function calls

ipriver:
According to [go 1.5 release notes](https://golang.org/doc/go1.5):

> By default, Go programs run with GOMAXPROCS set to the number of cores available; in prior releases it defaulted to 1.

There's no need anymore to call it with `runtime.NumCPU()` argument, it is now used as a default. Removed legacy `GOMAXPROCS` calls from the project.

### DIFF
--- a/blockchain/scriptval_test.go
+++ b/blockchain/scriptval_test.go
@@ -6,7 +6,6 @@ package blockchain
 
 import (
 	"fmt"
-	"runtime"
 	"testing"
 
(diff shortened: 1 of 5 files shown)


onyb (approved):
Looks good!

jcvernaleo (approved):
Wow, that use of runtime was crazy old!

OK

PR #1505, opened 2019-12-15
rpc: add wrapper for listtransactions command with include_watchonly parameter

mycroft:

PR #1628, opened 2020-09-07
rpcclient: support listtransactions RPC with watchonly argument

onyb:
There are a total of 3 pull requests for the same thing - #1163, #1505, and #1368. This PR is just a rebase of #1163, which is more ready than the others.

Recommend doing a squash-and-merge, so that the PR title is used as the commit message.


jcvernaleo (approved):
OK

PR #1629, opened 2020-09-08
GetBlockTemplate RPC client implementation

elliottminns:
Basic implementation of GetBlockTemplate which has been marked with a `TODO`

### DIFF
--- a/btcjson/chainsvrresults.go
+++ b/btcjson/chainsvrresults.go
@@ -238,6 +238,7 @@ type GetBlockFilterResult struct {
 type GetBlockTemplateResultTx struct {
 	Data    string  `json:"data"`
 	Hash    string  `json:"hash"`
+	TxID    string  `json:"txid"`
 	Depends []int64 `json:"depends"`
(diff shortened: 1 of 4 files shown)


btcjson/chainsvrresults.go
```diff
@@ -238,6 +238,7 @@ type GetBlockFilterResult struct {
 type GetBlockTemplateResultTx struct {
 	Data    string  `json:"data"`
 	Hash    string  `json:"hash"`
+	TxID    string  `json:"txid"`
```
onyb, line 241: Prefer adding the `omitempty` tag here, otherwise, the response on the rpcserver side will include an empty string. Please also add a comment like this:

```
// TODO: remove omitempty once implemented in rpcserver
```

onyb (changes requested):
There are several fields missing in the results struct, but ok to not update them now since it'll require adding support in `rpcserver`.

Looks good overall.

onyb (approved):

jcvernaleo (approved):
OK

PR #1632, opened 2020-09-13
wire: add proper types for flag field and improve docs

onyb:
Summary of changes:

- Add a new const `TxFlagMarker` to indicate the flag prefix byte.
- Add a new `TxFlag` type to enumerate the flags supported by the
  tx parser.

  This allows us to avoid hardcoded magics, and will make it easier
  to support new flags in future.
- Improve code comments.

Closes #1598.


jcvernaleo (approved):
OK

PR #1633, opened 2020-09-15
rpcclient: implement getaddressinfo command

onyb:
| Docs | [getaddressinfo](https://bitcoincore.org/en/doc/0.20.0/rpc/wallet/getaddressinfo) |
-|-

Fields such as `label`, and `labels->purpose` are not included, since they are deprecated, and will be removed in Bitcoin Core 0.21.

Most of the fields in the results struct are repeated in the `embedded` field if (for example) the address itself is a script embedding another address. To have minimal code duplication, I have factored them out to a private struct.

Another noteworthy thing in this PR is the custom unmarshalling I did to support loading a `script` to a `txscript.ScriptClass`, without duplicating the entire struct with raw fields. I learnt about this technique in [this article](http://choly.ca/post/go-json-marshalling) if you want to read.

Tested locally with Bitcoin Core v0.20.1. See `rpcclient/example_test.go` for details.

### DIFF
--- a/btcjson/walletsvrcmds.go
+++ b/btcjson/walletsvrcmds.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2014 The btcsuite developers
+// Copyright (c) 2014-2020 The btcsuite developers
 // Use of this source code is governed by an ISC
 // license that can be found in the LICENSE file.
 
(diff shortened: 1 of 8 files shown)


jcvernaleo (approved):
OK

PR #1635, opened 2020-09-17
btcjson: add test for null params in searchrawtransactions

onyb:
This is about the bug originally reported in #1476. In fact, the fix has been made redundant thanks to support for nullable JSON-RPC params introduced in https://github.com/btcsuite/btcd/pull/1594. This PR simply adds a unit test as proof that the said bug is no longer valid.


ipriver (approved):
Test works. LGTM 👍

jcvernaleo (approved):
OK

Issue #1483, opened 2019-10-29
btcd insists on wanting to create a home directory despite being given --configfile --datadir --logdir cli arguments


lestephane:
I'm trying to run btcd as an unprivileged user who does not have a HOME directory (ie the home in /etc/passwd points to a /non-existing-location), and can only read and write to the necessary files btcd.conf, data/ directory and log/ directory. I have not turned on rpc in btcd.conf, so I don't care to define the RPC key and cert file, but I could down the line, making the use of defaultHomeDir completely redundant.

line 516 in `config.go` v0.20.0-beta creates the home unconditionally.

```Go
err = os.MkdirAll(defaultHomeDir, 0700)
```

Have you considered adding a --homedir directory cli argument, that I could point to an existing empty, non-writable directory for my use case?

PR #1636, opened 2020-09-18
--homedir cli arg

Tristyn:
A fix for #1483. btcd would create a directory under ~/ even if you specify a different data dir or the system doesn't have a home dir for the user.

### DIFF
--- a/config.go
+++ b/config.go
@@ -109,6 +109,7 @@ type config struct {
 	ConfigFile           string        `short:"C" long:"configfile" description:"Path to configuration file"`
 	ConnectPeers         []string      `long:"connect" description:"Connect only to the specified peers at startup"`
 	CPUProfile           string        `long:"cpuprofile" description:"Write CPU profile to the specified file"`
+	HomeDir              string        `long:"homedir" description:"Creates this directory at startup"`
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -514,7 +516,7 @@ func loadConfig() (*config, []string, error) {
 
 	// Create the home directory if it doesn't already exist.
 	funcName := "loadConfig"
-	err = os.MkdirAll(defaultHomeDir, 0700)
+	err = os.MkdirAll(cfg.HomeDir, 0700)
```
ipriver, line 519: What if you run `btcd` for the first time and you don't have `.btcd` (default directory for linux)?
https://github.com/btcsuite/btcd/blob/e5521de652133c1d59df6fc505fd3f6e5c23bea9/config.go#L70-L76
Everything else gonna fail, because `defaultHomeDir` directory wasn't created for these files.

Also, I'm not sure about the behaviour of other systems. Is it really safe to use `cfg.HomeDir` as an empty argument  string (default case if you don't provide this flag) for Windows?

ipriver (changes requested):

jcvernaleo:
This PR needs a rebase to get the correct github actions to run.

After that, @Tristyn do you have any response to the changes requeseted?

PR #1639, opened 2020-09-21
rpcserver: add txid to getblocktemplate response

onyb:
Follow up of #1629.

This PR populates both the `txid` and `hash` fields in the RPC server implementation.


jcvernaleo (approved):
OK

PR #1648, opened 2020-10-08
btcjson: add new JSON-RPC errors and document them

onyb:
Just some basic housekeeping - add new JSON-RPC errors and document all the exported error codes. See [source](https://github.com/bitcoin/bitcoin/blob/master/src/rpc/protocol.h).


jcvernaleo (approved):
OK

PR #1649, opened 2020-10-09
randomize port

lsharir:
This randomizes the port selection in rpctest.New (generateListeningAddresses), which allows our builds (during heavy server use) to retry a new sim if they hit a port collision.

### DIFF
--- a/integration/rpctest/rpc_harness.go
+++ b/integration/rpctest/rpc_harness.go
@@ -7,6 +7,7 @@ package rpctest
 import (
 	"fmt"
 	"io/ioutil"
+	"math/rand"
 	"net"
 	"os"
 	"path/filepath"
@@ -39,16 +40,6 @@ const (
 var (
 	// current number of active test nodes.
 	numTestInstances = 0
-
-	// processID is the process ID of the current running process.  It is
-	// used to calculate ports based upon it when launching an rpc
-	// harnesses.  The intent is to allow multiple process to run in
-	// parallel without port collisions.
-	//
-	// It should be noted however that there is still some small probability
-	// that there will be port collisions either due to other processes
-	// running or simply due to the stars aligning on the process IDs.
-	processID = os.Getpid()
 
 	// testInstances is a private package-level slice used to keep track of
 	// all active test harnesses. This global can be used to perform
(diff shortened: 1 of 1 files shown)


integration/rpctest/rpc_harness.go
```diff
@@ -482,8 +483,8 @@ func generateListeningAddresses() (string, string) {
 	localhost := "127.0.0.1"
 
 	portString := func(minPort, maxPort int) string {
-		port := minPort + numTestInstances + ((20 * processID) %
-			(maxPort - minPort))
+		rand.Seed(time.Now().UnixNano())
```
onyb, line 486: It's enough to set the seed only once, for example in line 484.

integration/rpctest/rpc_harness.go
```diff
@@ -482,8 +483,8 @@ func generateListeningAddresses() (string, string) {
...
+		port := minPort + numTestInstances + rand.Intn(maxPort-minPort)
```
onyb, line 487: Why not just `minPort + rand.Intn(maxPort-minPort)`?

onyb: You'll need to also update the comments in the lines 479 through 481, as the ports are not incremented anymore, but rather random.

It also seems `processID` doesn't serve any purpose anymore, so it can be removed too.

onyb (commented):
I added some review comments but would like to hear the opinion of others on whether we're ok doing away with deterministic ports in the integration tests.

integration/rpctest/rpc_harness.go
lsharir: yep, makes sense, thanks

integration/rpctest/rpc_harness.go
lsharir: 👍

onyb (approved):
This PR looks fine to me.

@lsharir A good next step would be to actually do some heavy stress-testing on the CI. Would you be interested in contributing such a feature? :)

jcvernaleo:
I'm all good with random ports for tests.  If it makes it easier to run tests then it is all good with me.

jcvernaleo (approved):
OK

jcvernaleo (changes requested):
Actually, could you squash this to one commit and give it a single, commit message to say what it does?  If not I can do it, but better for original author to do commit message if possible.

jcvernaleo:
Other than the squash/message, this one is good.

lsharir:
> Actually, could you squash this to one commit and give it a single, commit message to say what it does? If not I can do it, but better for original author to do commit message if possible.

this is done @jcvernaleo

jcvernaleo (approved):
OK

PR #1645, opened 2020-10-06
rpcclient: add more wallet commands

torkelrogstad:
Implement backupwallet, dumpwallet, loadwallet and unloadwallet.

Issue #1589, opened 2020-05-30
ListTransactions API is non-standard


onyb:
The [**`listtransactions`**](https://bitcoincore.org/en/doc/0.19.0/rpc/wallet/listtransactions/) RPC command accepts 4 optional arguments:
- `label`
- `count`
- `skip`
- `include_watchonly`

The way this is implemented in the rpcclient is via the following 4 methods (**note:** signature is out of date):
- [`ListTransactions(account string)`](https://github.com/btcsuite/btcd/blob/9f0179fd2c46caba343b6515602cf37172f14d5f/rpcclient/wallet.go#L103)
- [`ListTransactionsCount(account string, count int)`](https://github.com/btcsuite/btcd/blob/9f0179fd2c46caba343b6515602cf37172f14d5f/rpcclient/wallet.go#L122)
- [`ListTransactionsCountFrom(account string, count, from int)`](https://github.com/btcsuite/btcd/blob/9f0179fd2c46caba343b6515602cf37172f14d5f/rpcclient/wallet.go#L140)
- `ListTransactionsCountFromWatchOnly(account string, count, from int)`
  * Not implemented yet, but there is a pending PR: #1368 

I don't know the rationale behind this design, but it encourages bespoke signatures of publicly exposed methods in the rpcclient, which is a maintenance hazard. Also, imagine if tomorrow there is a new option added to the RPC, we'll need to implement yet another method as per our current approach.

I think we should really follow the standard here and have just one method with the following signature:
```go
ListTransactions(label *string, count *int, skip *int, includeWatchOnly *bool)
```

Rjected:
I do agree that we should have one exported function that has all parameters, including optional parameters, as the corresponding Core RPC command. I would note that `label` probably shouldn't be a pointer, because it is not optional.

I think it's okay for the methods `ListTransactionsCounts, ListTransactionsCountFrom, ListTransactionsCountFromWatchOnly` to exist though, and don't think that contradicts the need for a general method with all the optional parameters - the inclusion of `ListTransactionsCountFromWatchOnly` would mean that our rpcclient would support all of the same features as Core.
[...]

PR #1650, opened 2020-10-09
rpcclient: implement createwallet with functional options

onyb:
Rebased on #1645, to avoid having to resolve merge conflicts later. This PR only implements `createwallet` as a first-step towards functional options.

### Background

I find the current way of dealing with optional parameters in RPC commands very hard to maintain, and also clunky from a user's perspective. We implement one rpcclient method per optional parameter, with long-winding names, closely following the order of the parameters!

This has been previously discussed a bit in #1589.

#### Current style

We have multiple methods to choose from, depending upon the options you want to use. In the example below, it is impossible to specify just `watchOnly` without specifying `count`. 😞 

```go
ListTransactions(account string)
ListTransactionsCount(account string, count int)
ListTransactionsCountFrom(account string, count int, from int)
ListTransactionsCountFromWatchOnly(account string, count int, from int, watchOnly bool)
```

#### Proposed style

```go
ListTransactions(
        "myaccount",
        WithListTransactionsWatchOnly(),  // btcjson.Bool(true) if specified, default (nil) otherwise
        WithListTransactionsCount(100),   // Use options in any order you like!
)
```

The above is known as functional options pattern, which is widely used in the Go community. Dave Cheney has written about it [here](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis). Here's [another article](https://www.sohamkamani.com/golang/options-pattern) I really liked.

The API is made possible using the following variadic signature:

```
func (c *Client) ListTransactions(account string, opts ...ListTransactionsOpt) FutureListTransactionsResult
```

### DIFF
--- a/btcjson/walletsvrcmds.go
+++ b/btcjson/walletsvrcmds.go
@@ -63,6 +63,28 @@ func NewCreateMultisigCmd(nRequired int, keys []string) *CreateMultisigCmd {
 	}
 }
 
+// CreateWalletCmd defines the createwallet JSON-RPC command.
+type CreateWalletCmd struct {
+	WalletName         string
+	DisablePrivateKeys *bool   `jsonrpcdefault:"false"`
+	Blank              *bool   `jsonrpcdefault:"false"`
+	Passphrase         *string `jsonrpcdefault:"\"\""`
+	AvoidReuse         *bool   `jsonrpcdefault:"false"`
+}
+
+// NewCreateWalletCmd returns a new instance which can be used to issue a
+// createwallet JSON-RPC command.
+func NewCreateWalletCmd(walletName string, disablePrivateKeys *bool,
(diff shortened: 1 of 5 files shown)


jcvernaleo:
High level, I'm cool with this style.  Going to do a real review now though.

jcvernaleo (approved):
Looks good to me.  I like the new style.  Will give it a few days in case anyone else wants to look and then i'll merge.

jakesyl:
I like this. Must cleaner than what we have now and less tedious to implement than the builder strategy

jakesyl (approved):
Looks good to me! The only thing I might add is we should clarify some of our style guidelines here (https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md) at some point

onyb:
@jakesyl Indeed, we should document this. I'll do a separate PR to update the guidelines.

If you can, please feel free to merge the PR.

PR #1659, opened 2020-11-11
integration: optimize harness for better itest control, restore bitcoind compatibility

guggero:
Some improvements to the integration harness:
 - allow `btcd` binary to be specified instead of always building it from source
 - allow listener addresses to be overwritten to avoid port collisions
 - allow overwriting RPC connection retry parameters

Everything falls back to the previous behavior if not otherwise specified. Existing users will just need to specify one more parameter.

EDIT: The PR now also contains a commit that restores compatibility with `bitcoind > 0.19`! There was a side effect of #1594 that caused the `sendrawtransaction` command not to work anymore.

@Roasbeef @wpaulino

### DIFF
--- a/btcjson/chainsvrcmds.go
+++ b/btcjson/chainsvrcmds.go
@@ -11,6 +11,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"reflect"
 
 	"github.com/btcsuite/btcd/wire"
 )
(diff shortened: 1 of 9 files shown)


guggero:
I just tested this with our `lnd` itests and it seems because it's built against `btcd` master, we might also need to fix this issue before we can use it in `lnd`: https://github.com/lightningnetwork/lnd/pull/4589#issuecomment-691621200

jakesyl (approved):
This looks good to me. Function overwrites looks a little odd, but I don't think there's a better way to do it.

Would love to get someone else to look at it before merging

guggero:
> I just tested this with our `lnd` itests and it seems because it's built against `btcd` master, we might also need to fix this issue before we can use it in `lnd`: [lightningnetwork/lnd#4589 (comment)](https://github.com/lightningnetwork/lnd/pull/4589#issuecomment-691621200)

Added a commit to fix this.

btcjson/chainsvrcmds.go
```diff
@@ -819,11 +820,60 @@ func NewSearchRawTransactionsCmd(address string, verbose, skip, count *int, vinE
 	}
 }
 
+// AllowHighFeesOrMaxFeeRate defines a type that can either be the legacy
+// allowhighfees boolean field or the new maxfeerate int field.
+type AllowHighFeesOrMaxFeeRate struct {
+	Value interface{}
```
Roasbeef, line 826: Ahh, clever way to handle this!

Roasbeef (approved):
LGTM 🐲

Issue #1212, opened 2018-07-07
btcec: Implement Schnorr signatures


stevenroose:
According to the proposed BIP: https://github.com/sipa/bips/blob/bip-schnorr/bip-schnorr.mediawiki

I was thinking that it would make sense for btcec to be a separate repo. It could have a separate branch where we can support experimental crypto that is not accepted standard yet, like CT-related things, f.e. The upcoming versioning improvements in Go should make it less of an issue to have it as a separate repo as well.

Roasbeef:
What's the gain to extract `btcec` to a diff repo? In the past it was, but then we combined everything into a single repo as otherwise PR's touching diff packages was a major pain.

stevenroose:
> In the past it was, but then we combined everything into a single repo as otherwise PR's touching diff packages was a major pain. 

That issue should be resolved with a good dependency manager. dep or vgo will most probably make that a non-issue. I would think an argument in favor is that it allows for easier experimentation with not-yet-softforked crypto stuffs without bothering btcd development.

Also, cryptographic primitives that are related to Bitcoin, but not consensus-critical, could be added there, but btcd doesn't need them. F.e. tools to construct Schnorr multisigs or discrete log contracts. Basically all layer 2 crypto that btcd doesn't necessarily need.

hbakhtiyor:
https://github.com/hbakhtiyor/schnorr/ implemented, feedback appreciate

benjyz:
@stevenroose agreed, separate crypto package would be nice. any new thoughts, also with regards to schnorr?

gcsfred2:
PR's touching diff packages is a major pain. Consider adding Schnorr and/or EdDSA to btcsuite/btcd ( 
btcec).

Roasbeef:
The main blocker here that we were working on was to first switch over all field operations to used fixed sized limbs rather than big ints, as we'll gain quite a performance increase in the process. Once that's in place, the next step would be to implement proper constant time signing using the field elements (which #1667 doesn't use).

Roasbeef:
The related project `dcrd` has implemented a lot of what we'll need for this, so we can either use their modules directly or copy things over. They have a schnorr implementation as well, but it deviates from BIP 340.

PR #1667, opened 2020-11-19
Initial Schnorr signature support

philipglazman:
Adding a Schnorr BIP340 spec implementation to btcec. Hopefully this can be used as a building block as we get closer to Taproot activation as well as using it in signet/regtest.

Includes:
* Default signing
* Verification
* Public key conversion from compressed public keys or 32 byte schnorr public keys
* A new SchnorrPublicKey type that wraps the x(P) integer.
* SchnorrPubKey() method for PrivateKey type

Does not include batch verification.

Used the [BIP](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki) as of `3b1fb96` as reference.

Included test vectors from [BIP340](https://github.com/bitcoin/bips/blob/master/bip-0340/test-vectors.csv).

Haven't tested much on hardware yet.

Addresses https://github.com/btcsuite/btcd/issues/1212


jakesyl:
As per some of the concerns raised in IRC I've asked @ashwin111 to take a look at wether or not this is a constant-time implementation

ashwin111:
Generally speaking, *looks* ~constant time assuming my understanding of the implementation of the `curve.*` functions is accurate (that those are constant time relative to the fixed vector/scalar lengths).

ashwin111 (approved):

adiabat:
Hi- I tested out this code and I'm not sure if it's the signing side or the verification side, but I get "cannot create signature" failures (when it fails verification of the signature it just produced).

Here's some code you could add to the schnorr_test.go file to cause this failure:

```
func TestRandomSignatures(t *testing.T) {
	for i := 0; i < 1000; i++ {
		var secretKeyBytes, messageBytes [32]byte
		rand.Read(secretKeyBytes[:])
		rand.Read(messageBytes[:])
		schnorrSecretKey, _ := PrivKeyFromBytes(S256(), secretKeyBytes[:])
		sig, err := schnorrSecretKey.SchnorrSign(messageBytes[:], nil)
		if err != nil {
			t.Fatalf("Sign message %x with key %x signature %x fails: %s\n",
				messageBytes, secretKeyBytes, sig, err.Error())
		}
	}
}
```
I don't think there's anything wrong with using random messages / private keys as those should be valid.  If there are some restrictions they should be made clear and a different error message returned. (But I don't think that's the case here)

This failure seems to occur with the same probability (around 0.78%) whether the key is varied, the message is, or both are.  This suggests it fails 1/128th of the time, so it may be a padding problem.  I did not notice any pattern in the key or message that caused failure (there were some starting with 0 and some starting with f) but I didn't look that closely.

Please let me know if you're not able to reproduce this failure; I can show specific key / message pairs if that helps.

adiabat:
Looked a bit more, and `schnorrVerify` returns false from `Fail if x(R) != r` about 3/4ths the time, and `Fail if y(R) is not even` 1/4th the time.

So out of 512 random signatures, 1 will fail due to y(R) not even, and 3 fail due to x(R) !=r.

It might not actually map to powers of 2 this way, but it seems pretty close after running it for a while and trying millions of signatures.

philipglazman:
Thanks @adiabat. I was able to reproduce your issue.

It's was padding problem! Coordinates must be 32 bytes. 

Added commit to fix the padding issue and return better error messages from schnorrVerify.

btcec/schnorr.go
```diff
@@ -0,0 +1,333 @@
...
+
+	// (n - e)*P
+	ePx, ePy := curve.ScalarMult(Px, Py, new(big.Int).Sub(n, e).Bytes())
+
+	// R = s*G + (N-e)*P
+	Rx, Ry := curve.Add(sGx, sGy, ePx, ePy)
+
+	// Fail if R is at infinity.
+	if Rx.Cmp(zero) == 0 || Ry.Cmp(zero) == 0 {
+		return false, errors.New("point R is at infinity")
```
ysangkok, line 297: instead of constructing new errors on every failure (you would recognize them only by the string), why not give them top-level names and make them part of the interface? then the tests can assert that the correct error is returned.

gcsfred2:
Please consider prioritizing this PR.

gcsfred2:
@philipglazman Would you consider changing x to X in type SchnorrPublicKey or adding a constructor to SchnorrPublicKey ? I need to hold and represent public keys but not always I can initialize them from private keys.

Roasbeef:
Yep! This functionality has landed in master.

PR #1669, opened 2020-11-20
rpcclient: add ExtraHeaders in ConnConfig

10gic:
Add ExtraHeaders in ConnConfig struct. It's	useful when RPC provider need customized headers. For example, [blockdaemon](http://www.blockdaemon.com/) require the header X-Auth-Token.


jcvernaleo (approved):
I like this.  Looks good to me.

Will leave open a bit in case anyone else wants to chime in though.

onyb (approved):

PR #1672, opened 2020-12-07
Allow --notls when listening on all interfaces

abitrolly:
This is useful when wrapping `btcd` into docker container, which requires internal process to listen on external ports to forward them to 127.0.0.1:8334 on host.

### DIFF
--- a/config.go
+++ b/config.go
@@ -910,29 +910,14 @@ func loadConfig() (*config, []string, error) {
 	cfg.RPCListeners = normalizeAddresses(cfg.RPCListeners,
 		activeNetParams.rpcPort)
 
-	// Only allow TLS to be disabled if the RPC is bound to localhost
-	// addresses.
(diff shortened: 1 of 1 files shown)


jakesyl:
See discussion on #1229. I’m personally not opposed to this (I run in docker as well), but there are some good arguments for not adding as well

marcopeereboom:
🤦 🏅

abitrolly:
I've noticed the project got [`Dockerfile`](https://github.com/btcsuite/btcd/commits/master/Dockerfile) recently, and to my surprise `podman` was able to map external port to local `127.0.0.1:8888`. This might still be an issue for `docker`, because `podman` has some improvements (like an ability to run Kubernetes pod), but it works for me, so I guess the issue is not actual anymore. )
```
podman run -p 8888 991 -P admin -u admin --notls --rpclisten 127.0.0.1:8888
```

jakesyl:
Closing then - feel free to reopen

PR #1678, opened 2021-01-11
Fix error message returned by EstimateFee

Appelberg-s:
When you provide an argument to EstimateFee(numblocks uint32) that exceeds the estimateFeeDepth (which is set to 25), you get an error message that says "can only estimate fees for up to 100 blocks from now".  The variable used in the if condition and the variable used for creating the error message should be the same.


Appelberg-s:
Tried to reproduce the test failure but was unable to.

jcvernaleo (approved):
Nice catch.

OK

onyb (approved):

PR #1683, opened 2021-02-01
txscript: Refactor to implement tokenizer from drcd

kcalvinalvin:
parseScriptTemplate will allocate a copy of the script onto memory,
leading to excessive memory allocation. This commit brings the tokenizer
to btcd and modifies it to fit with Bitcoin.


Roasbeef:
@kcalvinalvin see #1684 for a version of that change with individual commits to make it easier to review.

PR #1685, opened 2021-02-02
txscript: Refactor to implement tokenizer from drcd

kcalvinalvin:
parseScriptTemplate will allocate a copy of the script onto memory,
leading to excessive memory allocation. This commit brings the tokenizer
to btcd and modifies it to fit with Bitcoin.


onyb:
Propose to close this in favour of #1684.

PR #1686, opened 2021-02-02
wire/common: Implement binarySerializer based on sync.Pool

kcalvinalvin:
sync.Pool implementation saves on the overhead of go channels and has the benefit of dynamically sizing the binaryFreeList.

sync.Pool is also used in the fmt package

### DIFF
--- a/wire/common.go
+++ b/wire/common.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"sync"
 	"time"
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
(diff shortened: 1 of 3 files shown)


cfromknecht:
@kcalvinalvin this looks like it would be a nice companion to #1426. there will be some rebase conflicts but it'd be cool to see how much it improves the numbers in that PR.

kcalvinalvin:
> @kcalvinalvin this looks like it would be a nice companion to #1426. there will be some rebase conflicts but it'd be cool to see how much it improves the numbers in that PR.

I can rebase this PR on top of #1426. Would it be better to just make a PR to cfromknecht:wire-opts?

BrannonKing:
A note on this: stack allocations would be faster than borrowing memory for all of the functions that return integers. There's no need to borrow anything on those.

PR #1687, opened 2021-02-02
blockchain: Add assumevalid

kcalvinalvin:
Assumevalid is taken from bitcoind and is different from checkpoints in
that it allows forking/still calls checkConnectBlock. The node just
won't check the scripts

### DIFF
--- a/blockchain/validate.go
+++ b/blockchain/validate.go
@@ -1152,6 +1152,24 @@ func (b *BlockChain) checkConnectBlock(node *blockNode, block *btcutil.Block, vi
 		runScripts = false
 	}
 
+	// If we're at a checkpoint, then ignore all assumeValid params
+	if checkpoint != nil {
+		// Check if we're at the assumeValidHash. If assumeValidHash isn't nil
+		// don't check signatures. This mimicks the behavior of the reference
+		// client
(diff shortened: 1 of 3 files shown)


blockchain/validate.go
```diff
@@ -1152,6 +1152,24 @@ func (b *BlockChain) checkConnectBlock(node *blockNode, block *btcutil.Block, vi
 		runScripts = false
 	}
 
+	// If we're at a checkpoint, then ignore all assumeValid params
+	if checkpoint != nil {
+		// Check if we're at the assumeValidHash. If assumeValidHash isn't nil
+		// don't check signatures. This mimicks the behavior of the reference
+		// client
+		if b.assumeValidHash != nil && node.hash.IsEqual(b.assumeValidHash) {
```
Roasbeef, line 1160: Shouldn't this instead be enforced during the headers first sync to ensure that the chain we're about to validate is actually a  _ancestor_  of the specified block hash?

Roasbeef (changes requested):

blockchain/validate.go
Roasbeef: Also IIRC, this needs to be coupled w/ a related constant that asserts the minimum about of chain work for the target branch.

Issue #1651, opened 2020-10-14
Panic in btcec.RecoverCompact


janos:
Calling `btcec.RecoverCompact` with `btcec.S256()` and certain signatures and hashes results in a panic.

I am providing a test [btcrecover.zip](https://github.com/btcsuite/btcd/files/5377231/btcrecover.zip) with a case as an example that can be reproduced. Panic is a result of big int mod inverse returning nil in `invr := new(big.Int).ModInverse(sig.R, curve.Params().N)` `github.com/btcsuite/btcd/btcec/signature.go:319` v0.21.0-beta.

Test execution result:

```
=== RUN   TestRecoverCompact_panic
=== RUN   TestRecoverCompact_panic/success
=== RUN   TestRecoverCompact_panic/failure
    /Users/janos/btcrecover/btcrecover_test.go:35: panic: runtime error: invalid memory address or nil pointer dereference
        goroutine 20 [running]:
        runtime/debug.Stack(0xc00009bab8, 0x115a7c0, 0x139a1d0)
        	/usr/local/go/src/runtime/debug/stack.go:24 +0x9f
        btcrecover.TestRecoverCompact_panic.func1.1(0xc000082c00)
        	/Users/janos/btcrecover/btcrecover_test.go:35 +0x57
        panic(0x115a7c0, 0x139a1d0)
        	/usr/local/go/src/runtime/panic.go:969 +0x175
        math/big.(*Int).Mul(0xc00009bda8, 0x0, 0xc00009bea8, 0x0)
        	/usr/local/go/src/math/big/int.go:168 +0xd0
        github.com/btcsuite/btcd/btcec.recoverKeyFromSignature(0x13a22e0, 0xc00009be98, 0xc0000b4100, 0x20, 0x20, 0x2, 0xc0000d6200, 0x1, 0x8, 0x0)
        	/Users/janos/go/pkg/mod/github.com/btcsuite/btcd@v0.21.0-beta/btcec/signature.go:322 +0x287
        github.com/btcsuite/btcd/btcec.RecoverCompact(0x13a22e0, 0xc0000ec050, 0x41, 0x41, 0xc0000b4100, 0x20, 0x20, 0x2e918628, 0x2e9186280000000f, 0x5f86c94b, ...)
        	/Users/janos/go/pkg/mod/github.com/btcsuite/btcd@v0.21.0-beta/btcec/signature.go:413 +0x251
        btcrecover.TestRecoverCompact_panic.func1(0xc000082c00)
[...]

PR #1691, opened 2021-02-03
btcec: validate R and S signature components in RecoverCompact

onyb:
|Issue|GH-1651|
-|-

Prevent panic in `btcec.RecoverCompact` by validating signature components. We check if `r` and `s` values are:
* Less than the elliptic curve order (N).
* Non-zero.

The `RecoverCompact` algorithm relies on a multiplicative inverse of the `r` value. The panic in #1651 was due to the fact that the `r` value (i.e., x-coordinate of random point `R`) was `0`. Such a point does not lie on the secp256k1 curve either, making the signature invalid.


jcvernaleo (approved):
Looks great!
OK

PR #1575, opened 2020-05-13
build: multiple cleanups

dajohi:
build: replace travis-ci with github actions.
build: update deps 
build: clean linter warnings

PR #1693, opened 2021-03-04
docs: update shields

jakesyl:
I was looking at the integration tests when I realized all of our shields are still linked to travis-ci despite us moving to github actions (#1575). While I was there, I switched the documentation links to `go.dev` since [godoc.org is now deprecated](https://blog.golang.org/godoc.org-redirect).

While we're at it probably makes sense to fix the link in the corner of the repo. @jcvernaleo I think you have permissions here, but if not can ping @Roasbeef 

btcsuite_btcd_ An alternative full node bitcoin implementation written in Go (golang)


jcvernaleo (approved):
Awesome.  Nice (and tedious) work!

OK

jcvernaleo:
@jakesyl I updated the doc link on the side bar.  If there is any other that I missed just let me know and I will fix them too.

PR #1695, opened 2021-03-09
rpcserver: Fix Error message returned by processRequest

qustavo:
When processRequest can't find a rpc command, `standardCmdResult` returns a `btcjson.ErrRPCMethodNotFound` but it gets ignored and a `btcjson.ErrRPCInvalidRequest` is returned instead.

For example when I run the `createwallet` command

```
btcctl --simnet --rpcuser=foo --rpcpass=bar createwallet testWallet
```

This is the error message  I get:
```
-32600: Invalid request: malformed
```

But instead I should get this (returned by `standardCmdResult`)
```
-32601: Method not found
```

This fixes `processRequest` behavior to return the right error message.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -4066,9 +4066,13 @@ func (s *rpcServer) processRequest(request *btcjson.Request, isAdmin bool, close
 			result, err = s.standardCmdResult(parsedCmd,
 				closeChan)
 			if err != nil {
-				jsonErr = &btcjson.RPCError{
-					Code:    btcjson.ErrRPCInvalidRequest.Code,
-					Message: "Invalid request: malformed",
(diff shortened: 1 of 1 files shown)


jcvernaleo (approved):
Good catch.
OK

jcvernaleo:
@gchaincl looks good but any chance you could squash these to one commit (with just the one commit message)?

jcvernaleo (changes requested):
Change is good but if possible, squash would be nice to keep history clean.

Thanks!

gchaincl:
Sure, will do

gchaincl:
@jcvernaleo commits squashed

PR #1700, opened 2021-03-20
chaincfg+blockchain: abstract/refactor BIP 9 version bits implementation to work w/ BIP 8 block heights 

Roasbeef:
In this PR, we further abstract our Version Bits implementation to be able to support _both_ BIP 9 time based timeouts, as well as BIP 8 block based timeouts. In order to bridge the two mechanisms, we introduces a series of new light interfaces that allow us to abstract away from when exactly a soft fork deployment "starts" and "ends". This PR doesn't yet add any new BIP 8 compatible deployments, as it's mean to be a pure refactoring that doesn't introduce any new behavior. It's also the case that BIP 8 has some extra Version Bits states that need to be added to our Version Bits implementation. 

The only thing missing from this PR is a few additional tests to confirm that no change in behavior has occurred. Reviewers should examine the change to how we define deployments that are "always started" and "never ended", as well as the usage of `time` methods to determine if a deployment has started or ended.

### DIFF
--- a/blockchain/thresholdstate.go
+++ b/blockchain/thresholdstate.go
@@ -156,6 +156,99 @@ func (b *BlockChain) PastMedianTime(blockHeader *wire.BlockHeader) (time.Time, e
 	return blockNode.CalcPastMedianTime(), nil
 }
 
+// thresholdStateTransition given a state, a previous node, and a toeholds
+// checker, this function transitions to the next state as defined by BIP 009.
+// This state transition function is also aware of the "speedy trial"
+// modifications made to BIP 0009 as part of the taproot softfork activation.
+func thresholdStateTransition(state ThresholdState, prevNode *blockNode,
+	checker thresholdConditionChecker,
+	confirmationWindow int32) (ThresholdState, error) {
+
+	switch state {
+	case ThresholdDefined:
+		// The deployment of the rule change fails if it
+		// expires before it is accepted and locked in. However
+		// speed deployments can only transition to failed
+		// after a confirmation window.
+		if !checker.IsSpeedy() && checker.HasEnded(prevNode) {
+			state = ThresholdFailed
+			break
+		}
+
+		// The state for the rule moves to the started state
+		// once its start time has been reached (and it hasn't
+		// already expired per the above).
+		if checker.HasStarted(prevNode) {
+			state = ThresholdStarted
+		}
+
+	case ThresholdStarted:
+		// The deployment of the rule change fails if it
+		// expires before it is accepted and locked in, but
+		// only if this deployment isn't speedy.
+		if !checker.IsSpeedy() && checker.HasEnded(prevNode) {
+			state = ThresholdFailed
+			break
+		}
+
+		// At this point, the rule change is still being voted
+		// on by the miners, so iterate backwards through the
+		// confirmation window to count all of the votes in it.
+		var count uint32
+		countNode := prevNode
+		for i := int32(0); i < confirmationWindow; i++ {
+			condition, err := checker.Condition(countNode)
+			if err != nil {
+				return ThresholdFailed, err
+			}
+			if condition {
+				count++
+			}
+
+			// Get the previous block node.
+			countNode = countNode.parent
+		}
+
+		switch {
+		// The state is locked in if the number of blocks in the
+		// period that voted for the rule change meets the
+		// activation threshold.
+		case count >= checker.RuleChangeActivationThreshold():
+			state = ThresholdLockedIn
+
+		// If this is a speedy deployment, we didn't meet the
+		// threshold above, and the deployment has expired, then
+		// we transition to failed.
+		case checker.IsSpeedy() && checker.HasEnded(prevNode):
+			state = ThresholdFailed
+		}
+
+	case ThresholdLockedIn:
+		// At this point, we'll consult the deployment see if a
+		// custom deployment has any other arbitrary conditions
+		// that need to pass before execution. This might be a
+		// minimum activation height or another policy.
+		//
+		// If we aren't eligible to active yet, then we'll just
+		// stay in the locked in position.
+		if !checker.EligibleToActivate(prevNode) {
+			state = ThresholdLockedIn
+		} else {
+			// The new rule becomes active when its
+			// previous state was locked in assuming it's
+			// now eligible to activate.
+			state = ThresholdActive
+		}
+
+	// Nothing to do if the previous state is active or failed since
+	// they are both terminal states.
+	case ThresholdActive:
+	case ThresholdFailed:
+	}
+
+	return state, nil
+}
+
 // thresholdState returns the current rule change threshold state for the block
 // AFTER the given node and deployment ID.  The cache is used to ensure the
 // threshold states for previous windows are only calculated once.
@@ -216,90 +309,17 @@ func (b *BlockChain) thresholdState(prevNode *blockNode, checker thresholdCondit
 
 	// Since each threshold state depends on the state of the previous
 	// window, iterate starting from the oldest unknown window.
+	var err error
 	for neededNum := len(neededStates) - 1; neededNum >= 0; neededNum-- {
 		prevNode := neededStates[neededNum]
 
-		switch state {
-		case ThresholdDefined:
-			// The deployment of the rule change fails if it
-			// expires before it is accepted and locked in. However
-			// speed deployments can only transition to failed
-			// after a confirmation window.
-			if !checker.IsSpeedy() && checker.HasEnded(prevNode) {
-				state = ThresholdFailed
-				break
-			}
-
-			// The state for the rule moves to the started state
-			// once its start time has been reached (and it hasn't
-			// already expired per the above).
-			if checker.HasStarted(prevNode) {
-				state = ThresholdStarted
-			}
-
-		case ThresholdStarted:
-			// The deployment of the rule change fails if it
-			// expires before it is accepted and locked in, but
-			// only if this deployment isn't speedy.
-			if !checker.IsSpeedy() && checker.HasEnded(prevNode) {
-				state = ThresholdFailed
-				break
-			}
-
-			// At this point, the rule change is still being voted
-			// on by the miners, so iterate backwards through the
-			// confirmation window to count all of the votes in it.
-			var count uint32
-			countNode := prevNode
-			for i := int32(0); i < confirmationWindow; i++ {
-				condition, err := checker.Condition(countNode)
-				if err != nil {
-					return ThresholdFailed, err
-				}
-				if condition {
-					count++
-				}
-
-				// Get the previous block node.
-				countNode = countNode.parent
-			}
-
-			switch {
-			// The state is locked in if the number of blocks in the
-			// period that voted for the rule change meets the
-			// activation threshold.
-			case count >= checker.RuleChangeActivationThreshold():
-				state = ThresholdLockedIn
-
-			// If this is a speedy deployment, we didn't meet the
-			// threshold above, and the deployment has expired, then
-			// we transition to failed.
-			case checker.IsSpeedy() && checker.HasEnded(prevNode):
-				state = ThresholdFailed
-			}
-
-		case ThresholdLockedIn:
-			// At this point, we'll consult the deployment see if a
-			// custom deployment has any other arbitrary conditions
-			// that need to pass before execution. This might be a
-			// minimum activation height or another policy.
-			//
-			// If we aren't eligible to active yet, then we'll just
-			// stay in the locked in position.
-			if !checker.EligibleToActivate(prevNode) {
-				state = ThresholdLockedIn
-
-			} else {
-				// The new rule becomes active when its
-				// previous state was locked in assuming it's
-				// now eligible to activate.
-				state = ThresholdActive
-			}
-
-		// Nothing to do if the previous state is active or failed since
-		// they are both terminal states.
-		case ThresholdActive:
-		case ThresholdFailed:
+		// Based on the current state, the previous node, and the
+		// condition checker, transition to the next threshold state.
+		state, err = thresholdStateTransition(
+			state, prevNode, checker, confirmationWindow,
+		)
+		if err != nil {
+			return state, err
 		}
 
 		// Update the cache to avoid recalculating the state in the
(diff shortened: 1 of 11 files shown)


Roasbeef:
The alternative to this PR is to just go with the old: "lower than this value is a block height" approach which is used for lock time related logic elsewhere. However, I figured it's better to just further abstract things and hide more of the implementation details from the call sits within the `blockchain` package.

blockchain/versionbits.go
```diff
@@ -123,27 +120,45 @@ type deploymentChecker struct {
...
-	return c.deployment.StartTime
+func (c deploymentChecker) HasStarted(blkNode *blockNode) bool {
+	deploymentStarter := c.deployment.DeploymentStarter
+	if clockStarter, ok := deploymentStarter.(chaincfg.ClockConsensusDeploymentStarter); ok {
+		clockStarter.SynchronizeClock(c.chain)
+	}
+
+	// Can't fail as we make sure to set the clock above.
+	header := blkNode.Header()
+	started, _ := deploymentStarter.HasStarted(&header)
```
gchaincl, line 138: Since `deploymentStarter` is an interface, and you don't know how the the actual implementation looks like, wouldn't if be better to check for the error?

Roasbeef:
So for this to potentially be "taproot ready", assuming this "speed trial" thing happens, then we'll need to also add a `minimum_activation_height` parameter. Basically this would be a new gate here: https://github.com/btcsuite/btcd/blob/master/blockchain/thresholdstate.go#L243 to only go to active if that height has passed.

chaincfg/params.go
```diff
@@ -69,13 +68,13 @@ type ConsensusDeployment struct {
...
+	// DeploymentStarter is used to determine if the given
+	// ConsensusDeployment has started or not.
+	DeploymentStarter ConsensusDeploymentStarter
 
-	// ExpireTime is the median block time after which the attempted
-	// deployment expires.
-	ExpireTime uint64
+	// DeploymentEnder is used to determine if the given
+	// ConsensusDeployment has ended or not.
+	DeploymentEnder ConsensusDeploymentEnder
```
wpaulino, line 77: This change also needs to be propagated to the RPC server (build errors below). We'll want to still expose `StartTime` and `EndTime` to populate the RPC response.

```
./rpcserver.go:1292:39: deploymentDetails.StartTime undefined (type chaincfg.ConsensusDeployment has no field or method StartTime)
./rpcserver.go:1293:39: deploymentDetails.ExpireTime undefined (type chaincfg.ConsensusDeployment has no field or method ExpireTime)
```

wpaulino (commented):
I really like this change, even though Taproot didn't activate through BIP 8, it paves the way forward for any future soft forks that do.

As you mentioned, this PR is still missing the state transition changes as part of Speedy Trial.

chaincfg/deployment_time_frame.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	medianTime, err := m.blockClock.PastMedianTime(blkHeader)
+	if err != nil {
+		return false, err
+	}
+
+	// We check both after and equal here as after will fail for equivalent
+	// times, and we want to be inclusive.
+	return medianTime.After(m.startTime) || medianTime.Equal(m.startTime), nil
```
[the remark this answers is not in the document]
Roasbeef: IIRC there was some edge case related to this we ran into in the past?

chaincfg/params.go
```diff
@@ -365,19 +376,31 @@ var RegressionNetParams = Params{
...
+				time.Time{}, // Never expires
+			),
 		},
 		DeploymentSegwit: {
-			BitNumber:  1,
-			StartTime:  0,             // Always available for vote
-			ExpireTime: math.MaxInt64, // Never expires.
+			BitNumber: 1,
+			DeploymentStarter: NewMedianTimeDeploymentStarter(
+				time.Unix(0, 0), // Always available for vote
```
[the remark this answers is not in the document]
Roasbeef: I think that was before i added the `IsZero` special case in the started/ended methods. Switched over to be more uniform.

chaincfg/deployment_time_frame.go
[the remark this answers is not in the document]
Roasbeef: Nevertheless, tho this is more verbose, I prefer it as it's more explicit IMO (explicitly cover the == case).

Roasbeef:
Initial comments addressed, next commit with add the possibility for a custom threshold per deployment and also the min activation height. Will then add a new dummy deployment to test against that has both of these new values specified.

blockchain/versionbits.go
```diff
@@ -123,27 +120,36 @@ type deploymentChecker struct {
...
 // checker is associated with.
 //
 // This is part of the thresholdConditionChecker interface implementation.
-func (c deploymentChecker) BeginTime() uint64 {
-	return c.deployment.StartTime
+func (c deploymentChecker) HasStarted(blkNode *blockNode) bool {
+	// Can't fail as we make sure to set the clock above when we
+	// instantiate *BlockChain.
+	header := blkNode.Header()
+	started, _ := c.deployment.DeploymentStarter.HasStarted(&header)
```
wpaulino, line 134: I realize it can't fail as is right now, but we should handle the error anyway in case this is not caught/updated whenever a new fallible deployment implementation is added.

Roasbeef:
Ok, this should be good to go now. Reviews should ensure they've internalized the commit body in this commit: https://github.com/btcsuite/btcd/pull/1700/commits/c1842e23c86b802d84edfc29d4c875982ce88f90

;
```diff
@@ -0,0 +1,547 @@
+// Copyright (c) 2013-2017 The btcsuite developers
```
kcalvinalvin, line 1: I think you added this file as a mistake.

Sjors:
From c1842e23c86b802d84edfc29d4c875982ce88f90:

> One thing worth noting here is that logic at the end of BIP 341 doesn't
match bitcoind as implemented. The bitcoind logic fixes an off-by-one
error in the BIP itself.

Just to make sure I understand what you're saying here. Are you referring to the transition from `LOCKED_IN` to `ACTIVE`?

BIP 341 [states](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#deployment):

```cpp
case LOCKED_IN:
        if (block.nHeight < min_activation_height) {
            return LOCKED_IN;
        }
        return ACTIVE;
```

The returned value (`LOCKED_IN` or `ACTIVE`) in this pseudo code refers to the state for the _current_ block, i.e. the block at `block.nHeight`. This follows the convention in BIP 9, where e.g. the pseudo code returns `DEFINED` for the genesis block.

So the current block is `ACTIVE` when its height is at or above the activation height.

Bitcoin Core does the [following](https://github.com/bitcoin/bitcoin/blob/v22.0/src/versionbits.cpp#L82-L85):

```cpp
case ThresholdState::LOCKED_IN: {
                // Progresses into ACTIVE provided activation height will have been reached.
                if (pindexPrev->nHeight + 1 >= min_activation_height) {
                    stateNext = ThresholdState::ACTIVE;
```

This is part of the function:
```cpp
    /** Returns the state for pindex A based on parent pindexPrev B. Applies any state transition if conditions are present.
     *  Caches state from first block of period. */
    ThresholdState GetStateFor(const CBlockIndex* pindexPrev,
```

When this function is called with the Tip block, it returns the state for the next (unmined!) block. So `pindexPrev->nHeight + 1` refers to the upcoming block, i.e. the rules that apply in the mempool. So the fork becomes `ACTIVE` for the mempool when `height of the tip block + 1` is `min_activation_height`.

So I don't think there's an off by one error in the BIP, but both the BIP and the Bitcoin Core implementation are confusing enough to be likely to cause off-by-one errors (and in fact they did with SegWit2x).

As an aside: afaik the deployment info was added to BIP 341 _after_ Taproot locked in with the speedy trial.

Roasbeef:
Decided ultimately to special case things a bit just so it matches 1:1, once this base PR is in, I want to follow up with some additional tests for the hybrid state machine.

Roasbeef:
Hey @Sjors, yeh I explained this on #btcd IRC. I was just initially tripped up by the fact that the BIP and the bitcoind implementation don't match up (one uses the current block (to be mined/accepted) while the other uses the previous. Our BIP 9 implementation always uses the previous block.

;
Roasbeef: Indeed...removed

blockchain/versionbits.go
Roasbeef: As is, starting can never fail. When/if a deployment that can fail here (which may come in a few years given the current rate of soft forks in the system), we can thread through the error. With the current function signature, not having to check the error makes the `thresholdStateTransition` more readable as there's less line noise.

Roasbeef:
Implemented the BIP 9 modifications, and added a new unit test here: https://github.com/btcsuite/btcd/pull/1700/commits/75d3489f30dfe9282ea1505920dcd9090796f429. 

This PR should be g2g now.

chaincfg/deployment_time_frame.go
```diff
@@ -0,0 +1,181 @@
...
+	// passed block header. The past median time is the median time of the
+	// 11 blocks prior to the passed block header.
+	PastMedianTime(*wire.BlockHeader) (time.Time, error)
+}
+
+// ConsensusDeploymentStarter determines if a given consensus deployment has
+// started. A deployment has started once according to the current "time", the
+// deployment is eligible for activation once a perquisite condition has
+// passed.
+type ConsensusDeploymentStarter interface {
```
guggero, line 32: Is there ever a deployment that has a different mechanism/implementation for starting and ending? Or could this just be a `ConsensusActivation` with two methods and the extended `ClockConsensusActivation` interface and `MedianTimeActivation` implementation... Would make code in later commits a bit simpler. But non-blocking.

chaincfg/params.go
```diff
@@ -94,6 +94,19 @@ type ConsensusDeployment struct {
...
+	// value being zero), modifies the traditional BIP 9 state machine by
+	// only transitioning from LockedIn to Active once the block height is
+	// greater than (or equal to) thus specified height.
+	MinActivationHeight uint32
+
+	// CustomActivationThreshold if set (non-zero), will _override_ the
+	// existing RuleChangeActivationThreshold value set at the
+	// network/chain level. This value divided by the active
+	// MinerConfirmationWindow denotes the threshold required for
+	// activation. A value of 1850 block denotes a 90% threshold.
```
guggero, line 107: Should this say `1815` instead?

chaincfg/params.go
```diff
@@ -415,6 +440,17 @@ var RegressionNetParams = Params{
 				time.Time{}, // Never expires
 			),
 		},
+		DeploymentTestDummyMinActivation: {
+			BitNumber:                 22,
+			CustomActivationThreshold: 72,  // Only needs 50% hash rate.
+			MinActivationHeight:       600, // Can only activate after height 600.
```
guggero, line 446: Not sure I understand this correctly, but does this mean Taproot would only activate after 600 blocks on regtest? That would require us to mine quite a few blocks during integration tests and similar tasks.

blockchain/thresholdstate.go
```diff
@@ -248,17 +264,37 @@ func (b *BlockChain) thresholdState(prevNode *blockNode, checker thresholdCondit
...
+			// At this point, we'll consult the deployment see if a
+			// custom deployment has any other arbitrary conditions
+			// that need to pass before execution. This might be a
+			// minimum activation height or another policy.
+			//
+			// If we aren't eligible to active yet, then we'll just
+			// stay in the locked in position.
+			if !checker.EligibleToActivate(prevNode) {
+				state = ThresholdLockedIn
+
```
guggero: nit: extra newline. Or add `break` to continue without indentation below?

integration/bip0009_test.go
```diff
@@ -268,7 +272,33 @@ func testBIP0009(t *testing.T, forkKey string, deploymentID uint32) {
...
+	}
+
+	// Otherwise, we'll need to mine additional blocks to pass the min
+	// activation height and ensure the rule set applies. For regtest the
+	// deployment can only activate after height 600, and at this point
+	// we've mined 4*144 blocks, so another confirmation window will put us
+	// over.
+	numBlocksLeft := confirmationWindow
+	for i := uint32(0); i < numBlocksLeft; i++ {
+		_, err := r.GenerateAndSubmitBlock(nil, signalForkVersion,
```
guggero, line 292: nit: formatting

integration/bip0009_test.go
```diff
@@ -268,7 +272,33 @@ func testBIP0009(t *testing.T, forkKey string, deploymentID uint32) {
...
+			time.Time{})
+		if err != nil {
+			t.Fatalf("failed to generated block %d: %v", i, err)
+		}
```
guggero, line 296: Before generating and submitting a block, should we assert that we're still in state `blockchain.ThresholdLockedIn`? This should be true until after we mine the very last block (after which we exit the loop).

guggero (commented):
I have limited understanding of activation mechanisms in general, but from what I see in this diff the changes make sense to me. Always nice to see how a well-defined interface can make things very straightforward to understand!

chaincfg/deployment_time_frame.go
Roasbeef: > Is there ever a deployment that has a different mechanism/implementation for starting and ending? 

Back when this was made, doing taproot activation with a combination of height/time was an open option. Since then time was used for start/end and height for the min time. The refactors here now aren't _totally_ necessary, but certainly made some of the added unit tests I wrote a bit easier.

chaincfg/params.go
Roasbeef: For tapproot, we'd set a height of zero so it can activate at anytime. This new deployment was added so I could test the "speedy trial" specific additions like the min activation height. 

This PR removed the old taproot deployment, as it can actually cause a panic in certain cases, as it needs to be defined for all network types (was only defined for segnet before).

integration/bip0009_test.go
Roasbeef: Good idea, added on.

guggero (approved):
LGTM :tada:

wpaulino:
Post-merge ACK 👍

Issue #1694, opened 2021-03-06
Change tx version to uint32


artjoma:
Bitcoin Testnet has transactions where transaction version = maxUint32
 
btcjson/chainsvrresults.go
`TxRawResult struct`

 Version int32 `json:"version"`
to 
Version uint32 `json:"version"`

MarcoFalke:
See also https://github.com/bitcoin/bitcoin/pull/16525

PR #1701, opened 2021-03-21
btcjson: Updated TxRawResult.Version from int32 to uint32

aureleoules:
Implements #1694.

### DIFF
--- a/btcjson/chainsvrresults.go
+++ b/btcjson/chainsvrresults.go
@@ -712,7 +712,7 @@ type TxRawResult struct {
 	Size          int32  `json:"size,omitempty"`
 	Vsize         int32  `json:"vsize,omitempty"`
 	Weight        int32  `json:"weight,omitempty"`
-	Version       int32  `json:"version"`
+	Version       uint32 `json:"version"`
 	LockTime      uint32 `json:"locktime"`
 	Vin           []Vin  `json:"vin"`
 	Vout          []Vout `json:"vout"`
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -764,7 +764,7 @@ func createTxRawResult(chainParams *chaincfg.Params, mtx *wire.MsgTx,
 		Weight:   int32(blockchain.GetTransactionWeight(btcutil.NewTx(mtx))),
 		Vin:      createVinList(mtx),
 		Vout:     createVoutList(mtx, chainParams, nil),
-		Version:  mtx.Version,
+		Version:  uint32(mtx.Version),
 		LockTime: mtx.LockTime,
 	}
 


onyb:
According to the [protocol documentation](https://en.bitcoin.it/wiki/Protocol_documentation#tx), it's explicitly mentioned that `version` field in raw transactions is a signed integer.

Screenshot 2021-04-12 at 22 57 43

onyb:
Just saw the linked issue #1694. It's weird that testnet3 has transactions where `version` won't fit in the type specified in the protocol docs. Maybe `int64` can be a middle ground here?

@jakesyl Do you know what a negative `version` field in a transaction might mean?

jakesyl:
@onyb after a little help from the [#bitcoin-core-dev](http://www.erisian.com.au/bitcoin-core-dev/log-2021-04-12.html) channel, I think this should be treated as a uint32_t.

As @luke-jr/@sipa pointed out, bitcoin-core's rpc interface treats it as a uint32 [here](https://github.com/bitcoin/bitcoin/blob/1c7be9ab90af14d24f4668b02d9f07cec6f88a78/src/core_write.cpp#L199) so that probably makes the most sense in our implementation. The bitcon-core code treats it as a `uint32_t` in (via `static_cast<uint32_t>(tx.nVersion))` w/ `tx.nVersion` stored as a [`int32_t`](https://github.com/bitcoin/bitcoin/blob/1c7be9ab90af14d24f4668b02d9f07cec6f88a78/src/primitives/transaction.h#L272:19)) in every case as well.

As for why, it appeared on testnet3 (only), there is a check [called](https://github.com/bitcoin/bitcoin/blob/41a8d2b96ff5cbc2efa9b0996f30228d0a625fdb/src/validation.cpp#L607) in [`MemPoolAccept::PreChecks`](https://github.com/bitcoin/bitcoin/blob/41a8d2b96ff5cbc2efa9b0996f30228d0a625fdb/src/validation.cpp#L575) ([`IsStandardTx`](https://github.com/bitcoin/bitcoin/blob/41a8d2b96ff5cbc2efa9b0996f30228d0a625fdb/src/policy/policy.cpp#L78) that checks if the version is negative (or encodes to a value greater than 2)):

`if (tx.nVersion > TX_MAX_STANDARD_VERSION || tx.nVersion < 1) `

this check [is not called](https://github.com/bitcoin/bitcoin/blob/41a8d2b96ff5cbc2efa9b0996f30228d0a625fdb/src/validation.cpp#L605) if `-testnet` or `-regtest` is called, which is why we're likely seeing it on testnet. 

(btw, since this is only a mempool check, this doesn't affect consensus).

onyb:
Thanks @jakesyl for following up on this. Let's go with a `uint32` then. In any case, `btcjson` structs are used only for the JSON-RPC comms, so the change should be safe. It is, however, a breaking change, but we've done similar changes in the past without bumping the major version number. Just need to be loud in the release notes. :)

@aureleoules Can you please rebase your PR to make the CI green?

jakesyl:
Honestly @onyb any issue releasing a new version soon? I feel like since we've got bulk-rpc in, it might make sense

onyb:
@jakesyl Indeed, it's long overdue. Going to prepare a draft for the release notes like I did last time just to get the ball rolling, and then we can coordinate with @jcvernaleo for the release.

aureleoules:
@onyb all green!

jcvernaleo (approved):
OK

onyb:
There seems to be a superfluous commit in the branch: 249fdcb2a3105ede1f4d906fc5cab93fcae248e3

jcvernaleo (dismissed):
Actually, take back my OK.  the actual change looks good, but needs a rebase first.

aureleoules:
I'm not sure why the CI failed, could you help me with the rebase?

onyb:
@aureleoules Your rebase was fine. You just hit a flaky test. If you rerun the Actions or rebase one more time, it should be green.

jakesyl:
@onyb just re-ran. We've gotta get your permissions fixed :)

jakesyl (approved):

onyb (approved):

PR #1704, opened 2021-03-30
build: update btcutil dependency

wpaulino:
This is currently an issue for projects that depend on the latest version of `btcutil` and use the btcd integration test harness.

### DIFF
--- a/integration/rpctest/memwallet.go
+++ b/integration/rpctest/memwallet.go
@@ -125,7 +125,7 @@ func newMemWallet(net *chaincfg.Params, harnessID uint32) (*memWallet, error) {
 
 	// The first child key from the hd root is reserved as the coinbase
 	// generation address.
-	coinbaseChild, err := hdRoot.Child(0)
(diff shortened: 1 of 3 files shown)


Roasbeef:
Hmm it looks like there's a test failure related to the conn/addr manager.

Roasbeef:
I wasn't able to reproduce locally, re-running the build.

integration/rpctest/memwallet.go
```diff
@@ -125,7 +125,7 @@ func newMemWallet(net *chaincfg.Params, harnessID uint32) (*memWallet, error) {
 
 	// The first child key from the hd root is reserved as the coinbase
 	// generation address.
-	coinbaseChild, err := hdRoot.Child(0)
+	coinbaseChild, err := hdRoot.Derive(0)
```
Roasbeef, line 128: Just wanted to note that this is fine since it's just an in-memory wallet, but otherwise we would've needed to use the "non standard" version.

Roasbeef (approved):
LGTM 🧪

PR #1716, opened 2021-04-26
rpctest: add witness commitment when calling CreateBlock

halseth:
If we tried to include transactions having witnesses, the block would be
invalid since the witness commitment was not added.


Roasbeef (approved):
LGTM 🍙

Issue #1706, opened 2021-04-13
Release planning for 0.22.0-beta


onyb:
Please find a draft of the release notes for the upcoming 0.22.0-beta version below.

Although there are a couple of issues remaining in the [milestone](https://github.com/btcsuite/btcd/milestone/3), I don't think it's necessary to wait for them. But do let me know if you'd like a particular PR to be included in the release.

cc: @jcvernaleo @jakesyl @Roasbeef

#### Protocol and network-related changes:

* Added support for witness tx and block in `notfound` msg.  (#1625)

* Added support for receiving `sendaddrv2` messages from a peer.  (#1670)

* Fixed bug in `peer` package causing last block height to go backwards.  (#1606)

* Added chain parameters for connecting to the public Signet network.  (#1692, #1718)

  Also added `--signet` command line flag to `btcctl` and `btcd` utilities.

#### Crypto changes:

* Fixed bug causing panic due to bad R and S signature components in `btcec.RecoverCompact`.  (#1691)
* Set the name (`secp256k1`) in the CurveParams of the S256 curve. (#1565)

#### Notable developer-related package changes:

* Improve gc for `txscript.IsUnspendable` method. (#1615)

  Significantly reduced allocations, which should bring a performance gain during sync.
* Used `RLock`/`RUnlock` when possible in the `addrmgr` package.  (#1697)
* Removed unknown block version warning in the `blockchain` package, due to false positives triggered by AsicBoost.  (#1463)
* Added proper types for flag field and improve docs, for `msgtx` in the `wire` package. (#1632)
* Added `chaincfg.RegisterHDKeyID` function to populate HD key ID pairs.  (#1617)
* Fixed flaky hash cache test due to resetting RNG.  (#1689)
* Added new method `mining.AddWitnessCommitment` to add the witness commitment as an `OP_RETURN` output within the coinbase transaction.  (#1716)

#### RPC changes:
* Implemented Batch JSON-RPC in `rpcclient` and server.  (#1583)

  Please refer to `rpcclient/examples/bitcoincorehttpbulk/README.md` for examples.

* Implemented `rpcclient` method to invoke `getdescriptorinfo` JSON-RPC command. (#1578)
* Updated the `rpcserver` handler for `validateaddress` JSON-RPC command to have parity with the bitcoind 0.20.0 interface.  (#1613)
* Implemented `rpcclient` method to invoke `getblockfilter` JSON-RPC command. (#1579)
[...]

PR #1719, opened 2021-05-13
mempool: add additional test case for inherited RBF replacement

Roasbeef:
In this commit, we add an additional test case for inherited RBF
replacement. This test case asserts that if a parent is marked as being
replaceable, but the child isn't, then the child can still be replaced
as according to BIP 125 it shoudl _inhreit_ the replaceability of its
parent.

The addition of this test case was prompted by the recently discovered
Bitcoin Core "CVE" [1]. It turns out that bitcoind doesn't properly
implement BIP 125. Namely it fails to allow a child to "inherit"
replaceability if its parent is also replaceable. Our implementation
makes this trait rather explicit due to its recursive implementation.
Kudos to the original implementer @wpaulino for getting this correct.

[1]: https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2021-May/018893.html

### DIFF
--- a/mempool/mempool_test.go
+++ b/mempool/mempool_test.go
@@ -1749,6 +1749,47 @@ func TestRBF(t *testing.T) {
 			},
 			err: "",
 		},
+		{
+			// A transaction that doesn't signal replacement, can
+			// be replaced if the parent signals replacement.
+			name: "inherited replacement",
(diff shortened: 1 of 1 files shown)


wpaulino (approved):

jcvernaleo (approved):
Looks great!
OK

onyb:
Doesn't need any addition to the release notes (#1706) I suppose.

PR #1733, opened 2021-07-15
rpcserverhelp: Remove extra period for gettxout--synopsis

kcalvinalvin:


jcvernaleo (approved):
Good catch.  Those one character typos are so hard to see.
OK

PR #1739, opened 2021-08-02
mempool: export isDust for use in other projects

Crypt-iQ:
This changes isDust to IsDust so other golang projects (btcwallet
or lnd) can use the precise dust calculation used by btcd.


onyb (approved):
Looks good!

jcvernaleo (approved):
Looks all good to me and makes sense to have this available.

OK

PR #1742, opened 2021-08-16
mempool: introduce GetDustThreshold to export dust limit calculation

Crypt-iQ:
This commit modifies no behavior and would allow other projects to retrieve the dust limit for a particular output type before the amount of the output is known (so `IsDust` wouldn't be possible). This is particularly useful in the Lightning Network for channel negotiation.  I did not include a unit test but I did locally check the following dust limits at the current 3000sat/kvb dust relay feerate:
- p2pkh: 546
- p2sh: 540
- p2wsh: 330
- p2wpkh: 294
- p2tr: 330

### DIFF
--- a/mempool/policy.go
+++ b/mempool/policy.go
@@ -172,17 +172,10 @@ func checkPkScriptStandard(pkScript []byte, scriptClass txscript.ScriptClass) er
 	return nil
 }
 
-// IsDust returns whether or not the passed transaction output amount is
-// considered dust or not based on the passed minimum transaction relay fee.
-// Dust is defined in terms of the minimum transaction relay fee.  In
-// particular, if the cost to the network to spend coins is more than 1/3 of the
-// minimum transaction relay fee, it is considered dust.
-func IsDust(txOut *wire.TxOut, minRelayTxFee btcutil.Amount) bool {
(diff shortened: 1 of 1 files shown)


jcvernaleo:
I just forced github to rerun the failing tests in case it was just a fluke.  If the tests fail again, could you take a look at them before we review?

Crypt-iQ:
@jcvernaleo the tests pass locally on my machine, looking at the CI the failures seem unrelated

mempool/policy.go
```diff
@@ -253,6 +246,20 @@ func IsDust(txOut *wire.TxOut, minRelayTxFee btcutil.Amount) bool {
 		totalSize += 107
 	}
 
+	return 3 * int64(totalSize)
```
Roasbeef, line 249: Why is the factor of `3` here retained? When in the original dust calculation, this value is used to ensure that the amount spent on fees to spend an output isn't _greater_ than 1/3 of the relay fee.

Roasbeef: Ah ok, we chatted offline and derived why this is in place.

Roasbeef (approved):
LGTM ⛳️

jcvernaleo (approved):
OK

PR #1750, opened 2021-09-13
btcec: check if recovered pk is at point of infinity

MariusVanDerWijden:
This adds a check to ensure that the recovered pk is not at the point of infinity


jcvernaleo (approved):
OK

PR #1752, opened 2021-09-16
peer+server: add new config option to optionally disable stall detection

Roasbeef:
In this commit, we add a new config options that allows one to start
`btcd` in an operating mode that disables the stall detection. This can
be useful in simnet/regtest integration tests settings where it's
important that `btcd` holds on to its possibly sole connection to the
only other node in the test harness.

A new config flag has been added to gate this behavior, which is off by
default.


kcalvinalvin:
For user friendliness, I think it'd be better if the `loadConfig()` function had a check to error out if the `DisableStallHandler` flag is given when the node is on mainnet (since you'd never want this on for mainnet).

Fairly sure you wouldn't want this for testnet either..

guggero (approved):
LGTM, pending @kcalvinalvin's suggestion :tada:

jcvernaleo:
I agree with the check to disable this on mainnet.  No opinion on testnet.

guggero:
We definitely need this to fix our itest flakes! I'm seeing this disconnect problem more often now.
cc @bhandras.

peer/peer.go
```diff
@@ -276,6 +276,13 @@ type Config struct {
 	// connection detecting and disconnect logic since they intentionally
 	// do so for testing purposes.
 	AllowSelfConns bool
+
+	// DisableStallHandler if true, then the stall handler that attempts to
+	// disconnect from peers that appear to be taking too long to respond
+	// to requests won't be activated. This can be useful in certain simnet
+	// scenarios where the stall behaver isn't important to the system
```
bhandras, line 283: nit: s/behaver/behavior

bhandras (approved):
LGTM! :tada:

jcvernaleo:
Could you rebase this PR so it runs the correct actions?

guggero:
Hmm, this seems to block `btcd` completely since the `stallControl` channel isn't being drained. I have a fixed version here: https://github.com/guggero/btcd/tree/no-stall

Roasbeef:
@jcvernaleo @kcalvinalvin PTAL

Applied that fix, thanks @guggero!

guggero (approved):
tACK e98a1a1.

Issue #1699, opened 2021-03-17
connmgr/tor.go: TorLookupIP


Crypt-iQ:
It seems that `buf` only uses `6+len(host)` bytes, but space is reserved for `7+len(host)`?
https://github.com/btcsuite/btcd/blob/f86ae60936d7134fad9bf0ab1b3a1c261abe79b7/connmgr/tor.go#L79-L86

PR #1759, opened 2021-10-15
reduce redundant memory allocation - resolves btcsuite/btcd#1699

ziollek:
Signed-off-by: Tomasz Ziolkowski <tomasz.ziolkowski@allegro.pl>


jcvernaleo (approved):
Good catch.

OK

PR #1765, opened 2021-10-26
Pin actions by SHA

naveensrinivasan:
Pin actions by SHA.

Pin actions to a full length commit SHA

Pinning an action to a full length commit SHA is currently the only way to use an action as
an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad
actor adding a backdoor to the action's repository,
as they would need to generate a SHA-1 collision for a valid Git object payload.

https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -12,11 +12,11 @@ jobs:
         go: [1.16.8, 1.17.1]
     steps:
       - name: Set up Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@331ce1d993939866bb63c32c6cbbfd48fa76fc57 #v2.0.0 and the same v2.1.4
         with:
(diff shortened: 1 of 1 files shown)


jcvernaleo:
I like the idea of this one, we just need them all to pass then I'll merge.

naveensrinivasan:
> I like the idea of this one, we just need them all to pass then I'll merge.

Would you be able to re-run the jobs? I don't think it has anything to do with the code changes. Thanks

jcvernaleo:
Sure, rerun started.

jcvernaleo:
Sure, rerun started.

jcvernaleo:
Looks like this still failed even after the rerun.  Any ideas?

PR #1768, opened 2021-11-01
txscript: add taproot script type

buck54321:
Add the `WitnessV1TaprootTy` script class and return it from `GetScriptClass` / `typeOfScript`.

Fix `ComputePkScript`, which was returning incorrect results for taproot (an probably other) scripts. `ComputePkScript` is now essentially useless for both v0 and v1 output witnesses.

Bump the **btcutil** dep to leverage new taproot address type.

### DIFF
--- a/txscript/pkscript.go
+++ b/txscript/pkscript.go
@@ -47,6 +47,9 @@ const (
 	// witnessV0ScriptHashLen is the length of a P2WSH script.
 	witnessV0ScriptHashLen = 34
 
+	// witnessV1TaprootLen is the length of a P2TR script.
+	witnessV1TaprootLen = 34
+
 	// maxLen is the maximum script length supported by ParsePkScript.
 	maxLen = witnessV0ScriptHashLen
 )
@@ -99,7 +102,7 @@ func ParsePkScript(pkScript []byte) (PkScript, error) {
 func isSupportedScriptType(class ScriptClass) bool {
 	switch class {
 	case PubKeyHashTy, WitnessV0PubKeyHashTy, ScriptHashTy,
-		WitnessV0ScriptHashTy:
+		WitnessV0ScriptHashTy, WitnessV1TaprootTy:
 		return true
 	default:
 		return false
@@ -131,6 +134,10 @@ func (s PkScript) Script() []byte {
 	case WitnessV0ScriptHashTy:
 		script = make([]byte, witnessV0ScriptHashLen)
 		copy(script, s.script[:witnessV0ScriptHashLen])
+
+	case WitnessV1TaprootTy:
+		script = make([]byte, witnessV1TaprootLen)
+		copy(script, s.script[:witnessV1TaprootLen])
 
 	default:
 		// Unsupported script type.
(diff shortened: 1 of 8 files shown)


txscript/pkscript.go
```diff
@@ -231,35 +238,24 @@ func computeNonWitnessPkScript(sigScript []byte) (PkScript, error) {
 
 // computeWitnessPkScript computes the script of an output by looking at the
 // spending input's witness.
+// IMPORTANT: With the addition of taproot, we can no longer say for certain
+// what kind of script the witness is in most cases. The only case in which we
+// can say for sure is when the witness data has an annex as the last push. In
+// that case, we can identify the script type, but we lack the ability to
+// reconstruct the script itself.
```
ellemouton, line 245: i would say that the only case we know for certain is when the number of witness elements is exactly 1. In that case we know it is p2tr keypath (for now at least. not future proof). 

I dont know if we can say that that it is taproot just cause there is an annex starting with 0x50 cause surely there can be other p2w scripts (pre-taproot) that just happen to have a final witness element starting with 0x50?

ellemouton: let me know what you think

philipglazman:
These new test vectors for BIP341 might be applicable. https://github.com/bitcoin/bips/pull/1225

txscript/pkscript.go
buck54321: > I dont know if we can say that that it is taproot just cause there is an annex starting with 0x50 cause surely there can be other p2w scripts (pre-taproot) that just happen to have a final witness element starting with 0x50?

I guess that's true. Mkay.

> i would say that the only case we know for certain is when the number of witness elements is exactly 1. In that case we know it is p2tr keypath (for now at least. not future proof).

I'm unable to verify this. Is it impossible to construct a p2wsh script spent by a length 1 witness?

Roasbeef:
Thanks for the PR! Want to hold off on a bit of this till we get https://github.com/btcsuite/btcd/pull/1684 in since it modifies a lot w.r.t how we parse scripts in general (main goal is GC optimization).

txscript/script.go
```diff
@@ -799,6 +799,9 @@ func getWitnessSigOps(pkScript []byte, witness wire.TxWitness) int {
 			witnessScript := witness[len(witness)-1]
 			return countSigOpsV0(witnessScript, true)
 		}
+	case 1:
+		// https://github.com/bitcoin/bitcoin/blob/368831371d97a642beb54b5c4eb6eb0fedaa16b4/src/script/interpreter.cpp#L2090
+		return 0
 	}
```
buck54321, line 805: I'm not certain about this.

txscript/standard.go
```diff
@@ -485,28 +520,31 @@ func isNullDataScript(scriptVersion uint16, script []byte) bool {
...
+			return MultiSigTy
+		case isNullDataScript(scriptVersion, script):
+			return NullDataTy
+		}
+	case 1:
+		switch {
+		case isWitnessTaprootScript(script):
+			return WitnessV1TaprootTy
+		}
 	}
```
buck54321, line 546: I guess this isn't really exercised since `typeOfScript` is only used in `CalcScriptInfo`, which is now clear defined only for v0 scripts. Should I leave?

benma:
> Thanks for the PR! Want to hold off on a bit of this till we get #1684 in since it modifies a lot w.r.t how we parse scripts in general (main goal is GC optimization).

The linked PR https://github.com/btcsuite/btcd/pull/1769 is merged. @buck54321 @Roasbeef would be great to get this merged soon, as it is blocking taproot support in the BitBoxApp :pray:

benma:
@buck54321 `PayToAddrScript` in standard.go needs to be extended by a `*btcutil.AddressTaproot` case.

txscript/pkscript.go
ellemouton: > I guess that's true. Mkay.

Ok, I was wrong about this. I was not super familiar with how scripts are parsed. So yeah, if it starts with `0x50` then it is the annex

chappjc: Sorry to interject as a third party, but while I found https://github.com/lightninglabs/neutrino/pull/234#issuecomment-983185360 very enlightening, I think it would all fall apart when witness version 2 rolls around, like what happened with version 1 showed up and looked like p2wsh.
So while it sounds like much can be surmised from the witness stack, a hard fail based on the result of `computeWitnessPkScript` seems risky.

ellemouton: > a hard fail based on the result of computeWitnessPkScript seems risky

yeah, especially when considering future witness versions like you said

gcsfred2:
script.go has `CalcWitnessSigHash` with `const scriptVersion = 0` hardcoded.

gcsfred2:
script.go has still
```
const (
	SigHashOld          SigHashType = 0x0

```
unchanged, but BIP341 redefines 0x00: "We define a new hashtype SIGHASH_DEFAULT (value 0x00)".

buck54321:
> script.go has `CalcWitnessSigHash` with `const scriptVersion = 0` hardcoded.

That gets trickier since it gets into the `ScriptTokenizer` and signature hash calculations. Can I call that work out of scope for this PR so that someone with more time or know-how can handle it in a separate effort? I don't want to slow you guys down.

gcsfred2:
Understood. I agree. No problem with me. 

> > script.go has `CalcWitnessSigHash` with `const scriptVersion = 0` hardcoded.
> 
> That gets trickier since it gets into the `ScriptTokenizer` and signature hash calculations. Can I call that work out of scope for this PR so that someone with more time or know-how can handle it in a separate effort? I don't want to slow you guys down.

gdsoumya:
Hi, just checking on the status of the PR. We are using btcd in our project and this PR is necessary for us to activate support for taproot. Any update would be appreciated, thanks!

Roasbeef:
> Can I call that work out of scope for this PR so that someone with more time or know-how can handle it in a separate effort? I don't want to slow you guys down.

np @buck54321, I'm building off this PR currently as a follow up to #1773 where I finish up the rest of the taproot validation logic.

Roasbeef:
Rolled these commits (with some modifications) into #1787, thanks for starting this @buck54321!

PR #1771, opened 2021-11-17
update btcutil to 9cdf59f60c51

benma:
It includes support for bech32m and makes dependency management easier for projects that use both btcd and btcutil and need the bech32m functions.


benma:
This change is also part of https://github.com/btcsuite/btcd/pull/1768/. Could merge this first.

Roasbeef:
`btcutil` now lives within this repo

PR #1773, opened 2021-11-19
btcec: create new btcec/v2 module that type aliases into the dcrec module

Roasbeef:
In this PR, we make the switch to remove most of the code from `btcec` in favor of creating a _new_ module that uses type aliases to take advantage of all the constant timeness fixes and optimizations in the `dcrec` repo. 

The changes were mostly mechanical, though there're a few things I need to fix like ensuring we still expose decoding signatures in the BER format. 

Taking a look at the benchmarks, most operations other than normalization (which IIRC is a bit slower now due to constant time fixes) enjoy some nice speeds up: 
```
benchcmp is deprecated in favor of benchstat: https://pkg.go.dev/golang.org/x/perf/cmd/benchstat
benchmark                            old ns/op     new ns/op     delta
BenchmarkAddJacobian-8               464           328           -29.20%
BenchmarkAddJacobianNotZOne-8        1138          372           -67.27%
BenchmarkScalarBaseMult-8            47336         31531         -33.39%
BenchmarkScalarBaseMultLarge-8       42465         32057         -24.51%
BenchmarkScalarMult-8                123355        117579        -4.68%
BenchmarkNAF-8                       582           168           -71.12%
BenchmarkSigVerify-8                 175414        120794        -31.14%
BenchmarkFieldNormalize-8            23.8          24.4          +2.39%
BenchmarkParseCompressedPubKey-8     24282         10907         -55.08%
```


Roasbeef:
The PR doesn't build as is due to some Go modules oddities. I think we may actually need to make a new _branch_ for the initial v2 switch over?

btcec/curve.go
```diff
@@ -0,0 +1,60 @@
...
+// MakeJacobianPoint returns a Jacobian point with the provided X, Y, and Z
+// coordinates.
+func MakeJacobianPoint(x, y, z *FieldVal) JacobianPoint {
+	return secp.MakeJacobianPoint(x, y, z)
+}
+
+// AddNonConst adds the passed Jacobian points together and stores the result
+// in the provided result param in *non-constant* time.
+func AddNonConst(p1, p2, result *JacobianPoint) {
+	secp.AddNonConst(p1, p1, result)
```
davecgh, line 20:
```suggestion
	secp.AddNonConst(p1, p2, result)
```

Roasbeef: 🤦🏾‍♂️  lolol fixed!

davecgh:
> signature_test.go:345: trailing crap. signature failed when shouldn't malformed signature: bad length: 68 != 69

We don't allow this condition in Decred, as we enforce proper DER signature lengths.  If memory serves, this must be allowed in BTC though and there are historical signatures on the chain that violate it.

I believe you'll have have to keep the `btcec` implementation slightly modified to accept the new types as a result.

Roasbeef:
Pushed up a new version with things re-worked slightly:
 * We create a new `v2` sub-folder where the new module lives. We need to do this due to the way `btcutil` is set up, since it uses `btcec`, and will not allow a new module creation otherwise. 
 * Due to the above, we need to convert between types of the new and old module at times. 

After going with this initial approach, thinking maybe things would be easier if we did it in the following order:
  1. Add the new v2 sub folder 
  2. Update btcutil to use the new v2 module 
  3. Update btcd to use the new v2 module

Roasbeef:
After the latest CI run, there's something off w/ sig parsing+verification still AFAICT. One peculiar thing is I had to remove the example (temporarily in an attempt to fix the module) as the sig validation was failing there. 

As is, we use our existing signature parsing, but _dcr's_ sig serialization, which afaict produces slightly stricter signatures.

Roasbeef:
Need to trace the source of this, which should then ideally resolve those other failing cases: 
```
=== RUN   Example_verifySignature
--- FAIL: Example_verifySignature (0.00s)
got:
Signature Verified? false
want:
Signature Verified? true
```

Roasbeef:
The build is now ✅! 

Will proceed with the second leg now where I make a new PR to make a new `ecdsa` sub-folder, and add a BIP-340 implementation along side it. 

I'm open to an alternative route vs the one laid out above, such as making this into a new `v2` branch. Also important to note that the duplication is only temporary, as we can nuke the _legacy_ version of `btcec` once a new `btcec/v2` module is tagged here.

davecgh:
Glad to hear it.  Please retain the copyright headers from the files that were copied over from DCR.

Roasbeef:
Also a note to reviewers: you'll want to check out the commits in order, so you can examine the diff due to the type alias port, and then the final API changes needed in the greater repo. In a few areas, I needed to convert types, since `btcutil` still uses the base v1 `btcec` package.

kcalvinalvin:
> After going with this initial approach, thinking maybe things would be easier if we did it in the following order:
> 
>     1. Add the new v2 sub folder
> 
>     2. Update btcutil to use the new v2 module
> 
>     3. Update btcd to use the new v2 module

What about just moving `btcutil` into the `btcd` repo? There's really no reason for it to be a separate thing since it's so interconnected with `btcd` anyways. Looks like decred also did this as they have the `dcrutil` package inside the `dcrd` repo.

txscript/sign.go
```diff
@@ -288,7 +281,16 @@ sigLoop:
 			// bad input.
 			pkaddr := addr.(*btcutil.AddressPubKey)
 
-			pubKey := pkaddr.PubKey()
+			pubKeyOld := pkaddr.PubKey()
+
+			// A bit of modules induced weirdness: we need to
+			// conver types here since btcutil is still using the
```
kcalvinalvin, line 287: nit: cover to convert. But I guess this will be gone in 1777 anyways

kcalvinalvin:
While reviewing, I saw that a lot of tests were removed as they were already being tested in dcrd anyways. This is an idea but if there isn't a strong objection against cgo, it'd be better to bind to Bitcoin core's secp256k1. Then there won't be any worries about bug incompatibility at the least.

davecgh:
I'm no longer maintaining btcd, so it's not up to me, but using cgo would be an absolute 100% no from me if I were.  If you care about maintaining quality software, binding crypto to cgo really is a bad idea in a project like this for several reasons. Among them:

* it causes all kinds of cross compilation issues
* it destroys the ability to create reproducible builds via the normal Go mechanisms
* it is a constant source of bugs and suffers frequent run-time issues (look through the various Go patch releases)
* it is difficult to review for correctness
* calling into it has quite a lot of overhead and all of this code is used in super hot paths

All of that aside, I'd also note that the crypto itself is standardized, so the worry about compatibility is not in the core crypto itself, rather the potential compatibility issues arise in serialization and parsing due to Bitcoin's allowance of non-standard signature signature serializations.  Pulling in and entire crypto library via cgo just to parse signatures seems like overkill, to say the least.

jcvernaleo:
I agree with @davecgh there.  I really think avoiding cgo is the way to go in nearly all cases.

btcec/v2/bench_test.go
```diff
@@ -0,0 +1,368 @@
...
+// half of the digits will be 0.
+//
+// The aforementioned properties are particularly beneficial for optimizing
+// elliptic curve point multiplication because they effectively minimize the
+// number of required point additions in exchange for needing to perform a mix
+// of fewer point additions and subtractions and possibly one additional point
+// doubling.  This is an excellent tradeoff because subtraction of points has
+// the same computational complexity as addition of points and point doubling is
+// faster than both.
+func naf(k []byte) nafScalar {
```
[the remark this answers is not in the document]
davecgh: Agreed.  There really isn't any reason to keep it for the sole purposes of benchmarking, particularly when it's rolled up into the internal implementation details of something that is exported and benchmarked itself.

btcec/v2/README.md
```diff
@@ -0,0 +1,68 @@
...
+under the liberal ISC license.
+
+Although this package was primarily written for btcd, it has intentionally been
+designed so it can be used as a standalone package for any projects needing to
+use secp256k1 elliptic curve cryptography.
+
+## Installation and Updating
+
+```bash
+$ go get -u github.com/btcsuite/btcd/btcec
```
guggero, line 23: Update to `v2`? And use `go install`?

integration/rpctest/memwallet.go
```diff
@@ -129,10 +129,15 @@ func newMemWallet(net *chaincfg.Params, harnessID uint32) (*memWallet, error) {
 	if err != nil {
 		return nil, err
 	}
-	coinbaseKey, err := coinbaseChild.ECPrivKey()
+	coinbaseKeyOld, err := coinbaseChild.ECPrivKey()
 	if err != nil {
 		return nil, err
 	}
+
+	// Some modules weirdness: we need to conver types here as btcutil
```
guggero, line 137: nit: s/conver/convert.
Also, add TODO to remove once we take in `btcutil`?

guggero (approved):
Took me a moment to understand the logic behind the code move and the commit structure but I think it's quite nice to see the diff between the `v1` and `v2` code in the first commit.
Nice work preserving that!

LGTM assuming we fix the TODOs in the tests with the attached patch: 
[tests.patch.txt](https://github.com/btcsuite/btcd/files/7685083/tests.patch.txt)

I'm also very much against using cgo as that would break reproducible builds in `lnd`.

integration/csv_fork_test.go
```diff
@@ -3,6 +3,7 @@
 // license that can be found in the LICENSE file.
 
 // This file is ignored during the regular tests due to the following build tag.
+//go:build rpctest
```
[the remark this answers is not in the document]
Roasbeef: Added by the Go 1.18 version of go vet, there's a new syntax for build tags: https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md

integration/rpcserver_test.go
```diff
@@ -3,6 +3,7 @@
 // license that can be found in the LICENSE file.
 
 // This file is ignored during the regular tests due to the following build tag.
+//go:build rpctest
```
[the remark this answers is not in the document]
Roasbeef: See comment above, applies to other instances as well.

txscript/sign.go
Roasbeef: With the new version of https://github.com/btcsuite/btcd/pull/1785 -- should be the case that I no longer need to do this.

btcec/v2/bench_test.go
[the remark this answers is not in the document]
Roasbeef: True, was nice to get the benchmark stats for the commit/PR-body, but not needed really. Will remove.

btcec/field_test.go
```diff
@@ -1080,10 +1117,12 @@ func TestFieldSetBytes(t *testing.T) {
 	}}
 
 	for _, test := range tests {
-		inBytes := hexToBytes(test.in)
+		_ = hexToBytes(test.in)
+
+		// TODO(roasbeef): can't access internal state
```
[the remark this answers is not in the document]
Roasbeef: Diff shows as outdated for me...same with the one above as well.

Roasbeef:
Now that https://github.com/btcsuite/btcd/pull/1788 is merged, I'll pick this up again to get it properly building. We also should be able to remove one of the intermediate commits that just copies things over. 

We'll also need to make sure we push out the proper git tags for the affected sub-modules afterwards.

btcec/field_test.go
```diff
@@ -313,15 +305,16 @@ func TestNormalize(t *testing.T) {
 	}
 
 	t.Logf("Running %d tests", len(tests))
-	for i, test := range tests {
-		f := new(fieldVal)
+	for range tests {
+		// TODO(roasbeef): can't access internal state
```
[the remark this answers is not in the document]
Roasbeef: Removed it. If we wish, we can start to run the tests of this particular dependancy.

Roasbeef:
Just pushed up the new version! This is ready for another round of review now.

btcutil/address.go
```diff
@@ -469,9 +463,6 @@ func (a *AddressPubKey) serialize() []byte {
 
 	case PKFCompressed:
 		return a.pubKey.SerializeCompressed()
-
-	case PKFHybrid:
-		return a.pubKey.SerializeHybrid()
```
Roasbeef, line 474: Ended up removing support for hybrid keys, as they don't exist in BIP 340 and have been non standard for some time elsewhere.

btcutil/address_test.go
```diff
@@ -338,64 +338,6 @@ func TestAddresses(t *testing.T) {
...
-					0x52, 0xc6, 0xb4, 0x0d, 0x45, 0x26, 0x48, 0x38, 0xc0, 0xbd,
-					0x96, 0x85, 0x26, 0x62, 0xce, 0x6a, 0x84, 0x7b, 0x19, 0x73,
-					0x76, 0x83, 0x01, 0x60, 0xc6, 0xd2, 0xeb, 0x5e, 0x6a, 0x4c,
-					0x44, 0xd3, 0x3f, 0x45, 0x3e}
-				return btcutil.NewAddressPubKey(serializedPubKey, &chaincfg.MainNetParams)
-			},
-			net: &chaincfg.MainNetParams,
-		},
-		{
-			name: "mainnet p2pk hybrid (0x07)",
```
Roasbeef, line 371: Removed this since we drop support for serializing in a hybrid format.

btcutil/hdkeychain/extendedkey.go
```diff
@@ -292,8 +291,8 @@ func (k *ExtendedKey) Derive(i uint32) (*ExtendedKey, error) {
 	// chance (< 1 in 2^127) this condition will not hold, and in that case,
 	// a child extended key can't be created for this index and the caller
 	// should simply increment to the next index.
-	ilNum := new(big.Int).SetBytes(il)
-	if ilNum.Cmp(btcec.S256().N) >= 0 || ilNum.Sign() == 0 {
+	var ilNum btcec.ModNScalar
+	if overflow := ilNum.SetByteSlice(il); overflow {
```
Roasbeef, line 295: Note that this is kinda optional, big.Ints work here, but wanted to make things more uniform so started to use the new `ModNScalar`.

btcutil/hdkeychain/extendedkey.go
```diff
@@ -313,35 +312,66 @@ func (k *ExtendedKey) Derive(i uint32) (*ExtendedKey, error) {
...
-		childKey = ilNum.Bytes()
+		var keyNum btcec.ModNScalar
+		if overflow := keyNum.SetByteSlice(k.key); overflow {
+			return nil, ErrInvalidChild
+		}
+
+		childKeyBytes := ilNum.Add(&keyNum).Bytes()
+		childKey = childKeyBytes[:]
+
+		// Strip leading zeroes from childKey, to match the expectation
```
Roasbeef, line 323: The only odd thing I encountered during this translation, ran into something similar when getting the tests to pass for this original PR. As is, `SetByteSlice` assumes that all leading zeroes are stripped off, which is why it's added in the modified `parseSig` operation. 

If this is removed, we fail `TestLeadingZero`.

btcec/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/btcsuite/btcd/btcec/v2
+
+go 1.17
+
+require (
+	github.com/btcsuite/btcd v0.22.0-beta
```
wpaulino: Seems odd to have this dependency here, it's only current use is to access `chainhash.DoubleHashB` in `btcec/example_test.go`.

btcec/ciphering.go
```diff
@@ -44,173 +38,5 @@ var (
...
-//		PublicKey [70]byte
-//		// Cipher text
-//		Data []byte
-//		// HMAC-SHA-256 Message Authentication Code
-//		HMAC [32]byte
-//	}
-//
-// The primary aim is to ensure byte compatibility with Pyelliptic.  Also, refer
-// to section 5.8.1 of ANSI X9.63 for rationale on this format.
-func Encrypt(pubkey *PublicKey, in []byte) ([]byte, error) {
```
wpaulino, line 70: Most of the errors declared at the top of this file and no longer used and can be removed as well.

btcec/README.md
```diff
@@ -20,47 +20,27 @@ use secp256k1 elliptic curve cryptography.
...
   Demonstrates signing a message with a secp256k1 private key that is first
   parsed form raw bytes and serializing the generated signature.
 
-* [Verify Signature](https://pkg.go.dev/github.com/btcsuite/btcd/btcec#example-package--VerifySignature)  
+* [Verify Signature](https://pkg.go.dev/github.com/btcsuite/btcd/btcec/v2#example-package--VerifySignature)  
   Demonstrates verifying a secp256k1 signature against a public key that is
   first parsed from raw bytes.  The signature is also parsed from raw bytes.
 
-* [Encryption](https://pkg.go.dev/github.com/btcsuite/btcd/btcec#example-package--EncryptMessage)
+* [Encryption](https://pkg.go.dev/github.com/btcsuite/btcd/btcec/v2#example-package--EncryptMessage)
```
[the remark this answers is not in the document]
Roasbeef: Opted to just remove the dated line from the README

btcec/go.mod
Roasbeef: /shrug 

Can ultimately be removed when we make `chaincfg` its own package.

Roasbeef:
Latest set of comments addressed. Should be g2g now, though I think the tool we use for coverages doesn't understand `v2` modules? 
```
/home/runner/work/_actions/shogo82148/actions-goveralls/v1/bin/goveralls_linux_amd64 -service=github -coverprofile=btcec/coverage.txt
Error: cannot read source of file "/home/runner/work/btcd/btcd/btcec/v2/btcec.go": open /home/runner/work/btcd/btcd/btcec/v2/btcec.go: no such file or directory
Error: Error: The process '/home/runner/work/_actions/shogo82148/actions-goveralls/v1/bin/goveralls_linux_amd64' failed with exit code 1
```

Makefile
```diff
@@ -110,7 +114,7 @@ lint: $(LINT_BIN)
 
 clean:
 	@$(call print, "Cleaning source.$(NC)")
-	$(RM) coverage.txt btcutil/coverage.txt btcutil/psbt/coverage.txt
+	$(RM) coverage.txt btcec/coverage.txt btcutil/coverage.txt btcutil/psbt/coverage.txt
```
wpaulino, line 117: Nit: add these to `gitignore`.

Roasbeef:
Pushed new rebased version after the BIP 9 PR was merged. Once we get this in, I'll do a new tag for the v2 module, then rebase the schnorr PR as that'll be the final PR that needs to be merged for the main taproot PR to start building.

wpaulino (approved):
> Should be g2g now, though I think the tool we use for coverages doesn't understand v2 modules?

It seems that the tool is using the go module path (`btcec/v2/`) instead of the directory path (`btcec/`). I wonder if moving all the contents of `btcec/` into a subdir `btcec/v2/` fixes it?

Roasbeef:
> It seems that the tool is using the go module path (btcec/v2/) instead of the directory path (btcec/). I wonder if moving all the contents of btcec/ into a subdir btcec/v2/ fixes it?

If this is the case, then it's the tool that's messed up IMO. Just having to update the module means we don't need to actually _move_ everything when/if we create a `v3` version of the module. 

I did a scan of the issue tracker for that repo and didn't see anything related to v2 modules....

Roasbeef:
Perhaps there's some extra YAML field we need to be setting somewhere to tell it _exactly_ where to look?

Roasbeef:
Pushed a commit to try something out....

Roasbeef:
Didn't work, but I think what's happening is that the coverage file itself has `v2` in the path so the tool attempts to grab the source to overlay/analyze the coverage (?) and that fails since there's no actual `v2` path.

Roasbeef:
Great success: https://github.com/btcsuite/btcd/pull/1773/commits/b3d263e902e14e3491a8513cc8b0784597a61700

An alternative path would be updating the `goveralls` repo itself to understand the v2 module semantics, but this is quick and it does the job for now.

Issue #1747, opened 2021-09-04
rpcclient: GetNetworkHashPS3 must be float64


mhb8898:
i try to use GetNetworkHashPS3 in my program and when it's try to deserialize the json output give me an error and tell the output must be float64 not int64

PR #1778, opened 2021-12-05
rpcclient+rpcserver+integration: GetNetworkHashPS3 must be float64

mattbajorek:
## Description
Fixes [rpcclient issue](https://github.com/btcsuite/btcd/issues/1747) caused by using the rpcclient with bitcoind and expecting an int64, but bitcoind is [returning a float64](https://github.com/bitcoin/bitcoin/pull/7480/files).

## Changes
- Change getnetworkhashps and handlers return type from an int64 to a float64 to be aligned with bitcoin core and to also avoid the overflow

## Testing
- Extended rpcserver integration tests to test client methods `GetNetworkHashPS`, `GetNetworkHashPS2`, and `GetNetworkHashPS3`

### DIFF
--- a/integration/rpcserver_test.go
+++ b/integration/rpcserver_test.go
@@ -13,7 +13,9 @@ import (
 	"os"
 	"runtime/debug"
 	"testing"
+	"time"
 
+	"github.com/btcsuite/btcd/blockchain"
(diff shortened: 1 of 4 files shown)


jcvernaleo (approved):
Looks good and thanks for adding tests!

PR #1782, opened 2021-12-10
btcd: don't override explicitly set GOGC

anupcshan:
If GOGC env var is explicitly set, use it. Otherwise, set GC to 10% (default).


jcvernaleo (approved):
OK

PR #1785, opened 2021-12-25
multi: move the btcutil repo into btcd as a sub-module

Roasbeef:
In this PR, we move the `btcutil` repo into btcd itself as a sub-module, as suggested in #1777. With this change we won't need to retain the copy of `btcec v1` and can skip to using the `v2 version instead directly. 

Once this PR is merged, we'll need to push a `btcutil/v1.0.0` tag. After that I can update the `go.mod` in the `btcutil` sub-module to use a replace directive as normal, since it needs to reference a version of the greater `btcd` module as is.

If we don't want to finally bite the bullet and move over to having more sub-modules in `btcd`, then I can modify the initial commit to just add the new package to the existing btcd super-module. 

The final commit borrows the `Makefile` that `btcutil` borrowed from the `lnd` repo.


Roasbeef:
```
--- FAIL: TestUseBlockHash (0.00s)
    builder_test.go:86: Address script build failed: unable to generate payment script for unsupported address type *btcutil.AddressScriptHash
```

So I think we might actually need to do things in two phases:
  1. PR that moves the package into the repo. 
  1a. Tag a new module after the above PR is merged. 
  2. PR that updates imports, and the top-level `go.mod`.

Roasbeef:
OK, so I might've found a way around things. Just pushed up a new version, builds for mew locally....

The change here is to not attempt to replace the main `btcd` module from within `btcutil`, and also stop using `PayToAddrScript` from within the `btcutl/gcs` tests. If this works, then we just need to merge this, then make a new PR after the new tags are added to put the replace directives in place.

Roasbeef:
Looks like we had a build y'all! 

After this, it'll remove the old CI builds and have the newer ones instead be shown as required.

positiveblue (approved):
LGTM :tada:

PR #1788, opened 2022-01-11
btcutil: update modules to replace to top-level btcd repo

Roasbeef:
Now that the new `btcutil/v1.0.0` tag has been pushed, we update the new
internal modules to point to the top-level `btcd` repo via replace
directives.


go.sum
```diff
@@ -1,9 +1,11 @@
 github.com/aead/siphash v1.0.1 h1:FwHfE/T45KPKYuuSAKyyvE+oPWcaQ+CUmFW0bPlM+kg=
 github.com/aead/siphash v1.0.1/go.mod h1:Nywa3cDsYNNK3gaciGTWPwHt0wlpNV15vwmswBAUSII=
 github.com/btcsuite/btcd v0.20.1-beta/go.mod h1:wVuoA8VJLEcwgqHBwHmzLRazpKxTv13Px/pDuV7OomQ=
+github.com/btcsuite/btcd v0.22.0-beta/go.mod h1:9n5ntfhhHQBIhUvlhDvD3Qg6fRUj4jkN0VB8L8svzOA=
 github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f h1:bAs4lUbRJpnnkd9VhRV3jjAVU7DJVjMaK+IsvSeZvFo=
 github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f/go.mod h1:TdznJufoqS23FtqVCzL0ZqgP5MqXbb4fg/WgDys70nA=
 github.com/btcsuite/btcutil v0.0.0-20190425235716-9e5f4b9a998d/go.mod h1:+5NJ2+qvTyV9exUAL/rxXi3DcLg2Ts+ymUAY5y4NvMg=
+github.com/btcsuite/btcutil v1.0.3-0.20201208143702-a53e38424cce/go.mod h1:0DVlHczLPewLcPGEIeUEzfOJhqGPQ0mJJRDBtD307+o=
```
Roasbeef, line 8: Not sure why this is still in the mod sum...

I guess because we need to make a _new_ tag after this set of changes?

btcutil/go.mod
```diff
@@ -4,8 +4,10 @@ go 1.16
 
 require (
 	github.com/aead/siphash v1.0.1
-	github.com/btcsuite/btcd v0.20.1-beta
+	github.com/btcsuite/btcd v0.22.0-beta
```
guggero, line 7: The `v0.22.0-beta` version still pulls in `btcsuite/btcutil`, that's why you end up with `btcsuite/btcutil` still being in the top level `go.sum`. If you change this to `v0.22.0-beta.0.20220111032746-97732e52810c`, it goes away.
This is because `replace` directives are ignored in transitive dependencies. So the `replace` directive in this file is ignored when go mod is looking at the top level `btcd` package. Perhaps we should try to get rid of this circular replace directive soon to avoid issues like this?

guggero (approved):
LGTM :tada:

jcvernaleo (approved):
OK

btcutil/go.mod
Roasbeef: IIRC, once we tag a _new_ version of `btcd` after that, the dep hash goes away?

Roasbeef: Ah, ok I see now, I can add that. 

> Perhaps we should try to get rid of this circular replace directive soon to avoid issues like this?

IMO the main matter is if we want to have `btcutil` be a proper sub-module or not. The `btcutil/psbt` module also exists. I think as long as we make more sub-modules we'll have this problem until each package is isolated as its own module.

Roasbeef: The main packages that `btcutil` uses are: `wire`, `chainhash`, `btcec`, and `txscript` (FWIW).

Roasbeef:
Pushed up @guggero's suggestion to clean up `go.sum` somewhat. We'll eventually need to make the decision to go all in on sub-modules or not as well...

PR #1799, opened 2022-01-31
btcutil/base58: fix bug with range

egonelbre:
The decoding was using a range over the string instead of iterating
over the bytes.

### DIFF
--- a/btcutil/base58/base58.go
+++ b/btcutil/base58/base58.go
@@ -54,8 +54,8 @@ func Decode(b string) []byte {
 		}
 
 		total := uint64(0)
-		for _, v := range t[:n] {
-			tmp := b58[v]
+		for k := 0; k < n; k++ {
+			tmp := b58[t[k]]
 			if tmp == 255 {
 				return []byte("")
 			}
--- a/btcutil/base58/base58_test.go
+++ b/btcutil/base58/base58_test.go
@@ -43,6 +43,7 @@ var invalidStringTests = []struct {
 	{"4kl8", ""},
 	{"0OIl", ""},
 	{"!@#$%^&*()-_=+~`", ""},
+	{"\xdc", ""},
 }
 
 var hexTests = []struct {


schjonhaug:
Is this PR fixing the crash that occurs if you try to decode something with an emoji? What is the allowed character set here?

For example, using [example-Decode](https://pkg.go.dev/github.com/btcsuite/btcd/btcutil/base58?utm_source=godoc#example-Decode), if you try to decode a string with 🤡, it’ll crash:

```
package main

import (
	"fmt"

	"github.com/btcsuite/btcd/btcutil/base58"
)

func main() {
	// Decode example modified base58 encoded data.
	encoded := "25JnwSn7XKfNQ🤡"
	decoded := base58.Decode(encoded)

	// Show the decoded data.
	fmt.Println("Decoded Data:", string(decoded))

}
```

Result:

```
Output:

panic: runtime error: index out of range [10095] with length 256

goroutine 1 [running]:
github.com/btcsuite/btcd/btcutil/base58.Decode({0x49fff5, 0x10})
	/tmp/gopath2191683068/pkg/mod/github.com/btcsuite/btcd/btcutil@v1.1.0/base58/base58.go:58 +0x395
main.main()
	/tmp/sandbox2173415571/prog.go:12 +0x29
```

egonelbre:
Yes, it's a fix for runes that are larger 0xFF, which does include emoji. The valid alphabet is `123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz` -- i.e. base58 alphabet used by bitcoin.

ssnickolay:
Why this is not merged yet? I'm using the lib to check BTC address is valid and the issue could crash the app ;[

chappjc:
Similar fix downstream: https://github.com/decred/base58/blob/5b84bad1ebfd283724c012ec3e0e02f6f80b0ee2/base58.go#L25
Worth merging this PR or a direct port of that solution.

PR #1804, opened 2022-02-04
Revert "reduce redundant memory allocatio - resolves btcsuite/btcd#1699"

Crypt-iQ:
This reverts commit 780cc0889fd2f326e4a2a9b513d8d30ff37e0cd2.

Oversight on my part, that extra byte is actually used. Without this, DNS resolution via tor proxy won't work

### DIFF
--- a/connmgr/tor.go
+++ b/connmgr/tor.go
@@ -76,7 +76,7 @@ func TorLookupIP(host, proxy string) ([]net.IP, error) {
 		return nil, ErrTorUnrecognizedAuthMethod
 	}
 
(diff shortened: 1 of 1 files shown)


Roasbeef (approved):
LGTM 🌳

Confirmed that this breaks Tor resolution as is, and this PR resolves things. The issue is that 2 bytes are used for the port (with only one explicitly set), so it's 7 bytes total. We should also update the method to be able to parse a response that's an IPv6 address. I'll make an issue to track that.

PR #1807, opened 2022-02-08
mining: return witness_script instead of raw witness_commitment in GBT

roylee17:
The `"default_witness_commit"` in the `getblocktemplate` response uses the witness_script instead of raw commitment.

### DIFF
--- a/mining/mining.go
+++ b/mining/mining.go
@@ -863,7 +863,7 @@ mempoolLoop:
 }
 
 // AddWitnessCommitment adds the witness commitment as an OP_RETURN outpout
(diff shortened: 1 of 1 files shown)


roylee17:
@Roasbeef

Roasbeef:
@roylee17 slipped by initial due to a lack of integration tests in this area. Would you mind adding a new test in the `integration` folder? It would use GBT to construct a new block and then assert that we also accept that same block.

roylee17:
[Accidentally deleted the branch, which generates the, PR.]

@Roasbeef I'll look into how to create that test.

PR #1812, opened 2022-02-11
multi: implement BIP-155 addrv2 support

Crypt-iQ:
**Commit overview:**

- `wire: netaddressv2 and tests` adds (de)serialization routines for BIP-155 addresses. It also allows `wire.NetAddress` to be converted into `wire.NetAddressV2`.
- `multi: update addrmgr, server to use NetAddressV2 instead of legacy` updates the address manager to use `wire.NetAddressV2` internally and does so without a database upgrade. The server is also updated to use `wire.NetAddressV2` where appropriate.
- `peer+wire: add addrv2 message, protocol negotiation` implements the p2p negotiation of BIP-155 (`sendaddr`, `addrv2`) and updates the peer to use `wire.NetAddressV2` where appropriate. The tests here were modified to not use `io.Pipe` in some cases due to the implementation of the BIP-155 "handshake" not working with a synchronous read/write pipe.

Closes https://github.com/btcsuite/btcd/pull/1671 as this pull ignores unknown messages both during version-verack (as does bitcoind) and after the version-verack handshake.

Fixes https://github.com/btcsuite/btcd/pull/1671


wire/netaddressv2.go
```diff
@@ -0,0 +1,576 @@
...
+// AddService adds a service to the Services bitfield.
+func (na *NetAddressV2) AddService(service ServiceFlag) {
+	na.Services |= service
+}
+
+// ToLegacy attempts to convert a NetAddressV2 to a legacy NetAddress. This
+// only works for ipv4, ipv6, or torv2 addresses as they can be encoded with
+// the OnionCat encoding. If this method is called on a torv3 address, nil will
+// be returned.
+func (na *NetAddressV2) ToLegacy() *NetAddress {
```
Roasbeef, line 112: 👍

wire/netaddressv2.go
```diff
@@ -0,0 +1,576 @@
...
+
+	switch a := na.Addr.(type) {
+	case *ipv4Addr:
+		legacyNa.IP = a.addr[:]
+	case *ipv6Addr:
+		legacyNa.IP = a.addr[:]
+	case *torv2Addr:
+		legacyNa.IP = a.onionCatEncoding()
+	case *torv3Addr:
+		return nil
```
Roasbeef, line 127: Return an error instead perhaps? Still getting through the commit, need to see how it's used downstream.

wire/netaddressv2.go
```diff
@@ -0,0 +1,576 @@
...
+	err := writeElement(w, uint32(na.Timestamp.Unix()))
+	if err != nil {
+		return err
+	}
+
+	if err := WriteVarInt(w, pver, uint64(na.Services)); err != nil {
+		return err
+	}
+
+	var netID networkID
```
Roasbeef, line 208: Style nit, can be under a single `var` statement: 
```go
var (
)
```

addrmgr/addrmanager.go
```diff
@@ -70,7 +70,7 @@ type serializedAddrManager struct {
 }
 
 type localAddress struct {
-	na    *wire.NetAddress
+	na    *wire.NetAddressV2
```
Roasbeef, line 73: Is this change backwards compatible w/ the _existing_ `peers.json` we keep on disk, or are we assuming we'll just do a one time translation (next time the new version reads the file)?

addrmgr/addrmanager.go
```diff
@@ -734,8 +740,25 @@ func (a *AddrManager) HostToNetAddress(host string, port uint16, services wire.S
 		if err != nil {
 			return nil, err
 		}
-		prefix := []byte{0xfd, 0x87, 0xd8, 0x7e, 0xeb, 0x43}
-		ip = net.IP(append(prefix, data...))
+
+		na = wire.NetAddressV2FromBytes(
+			time.Now(), services, data, port,
+		)
+	} else if len(host) == 62 && host[56:] == ".onion" {
```
Roasbeef, line 747: I think the new constants added in the `wire` package (when exported) can be used here instead of these raw numbers.

addrmgr/addrmanager.go
```diff
@@ -745,30 +768,23 @@ func (a *AddrManager) HostToNetAddress(host string, port uint16, services wire.S
...
-// ip is in the range used for Tor addresses then it will be transformed into
-// the relevant .onion address.
-func ipString(na *wire.NetAddress) string {
-	if IsOnionCatTor(na) {
-		// We know now that na.IP is long enough.
-		base32 := base32.StdEncoding.EncodeToString(na.IP[6:])
-		return strings.ToLower(base32) + ".onion"
+		na = wire.NetAddressV2FromBytes(time.Now(), services, ip, port)
+	} else {
+		// This is an non-nil IP address that was parsed in the else if
```
Roasbeef, line 774: Not following...only one of these clauses will be executed.

addrmgr/knownaddress.go
```diff
@@ -15,18 +15,18 @@ import (
 // to determine how viable an address is.
 type KnownAddress struct {
 	mtx         sync.RWMutex // na and lastattempt
-	na          *wire.NetAddress
-	srcAddr     *wire.NetAddress
+	na          *wire.NetAddressV2
```
Roasbeef, line 18: Similar question here re backwards compat.

server.go
```diff
@@ -342,21 +343,43 @@ func (sp *serverPeer) relayTxDisabled() bool {
 
 // pushAddrMsg sends an addr message to the connected peer using the provided
 // addresses.
-func (sp *serverPeer) pushAddrMsg(addresses []*wire.NetAddress) {
-	// Filter addresses already known to the peer.
+func (sp *serverPeer) pushAddrMsg(addresses []*wire.NetAddressV2) {
```
Roasbeef, line 346: I think we should update the `godoc` and/or the method name toe reflect that this pushes the _legacy_ message to the peer.

server.go
```diff
@@ -342,21 +343,43 @@ func (sp *serverPeer) relayTxDisabled() bool {
...
+	// the set of known addresses.
+	knownAddrs := make([]*wire.NetAddressV2, 0, len(known))
+	for _, knownAddr := range known {
+		currentKna := wire.NetAddressV2FromBytes(
+			knownAddr.Timestamp, knownAddr.Services,
+			knownAddr.IP, knownAddr.Port,
+		)
+		knownAddrs = append(knownAddrs, currentKna)
+	}
+	sp.addKnownAddresses(knownAddrs)
```
Roasbeef, line 410: 👍

peer/peer.go
```diff
@@ -102,6 +102,9 @@ type MessageListeners struct {
 	// OnAddr is invoked when a peer receives an addr bitcoin message.
 	OnAddr func(p *Peer, msg *wire.MsgAddr)
 
+	// OnAddrV2 is invoked when a peer receives an addrv2 bitcoin message.
+	OnAddrV2 func(p *Peer, msg *wire.MsgAddrV2)
```
Roasbeef, line 106: 👍

peer/peer.go
```diff
@@ -856,6 +873,40 @@ func (p *Peer) PushAddrMsg(addresses []*wire.NetAddress) ([]*wire.NetAddress, er
...
+		return nil, nil
+	}
+
+	m := wire.NewMsgAddrV2()
+	m.AddrList = make([]*wire.NetAddressV2, count)
+	copy(m.AddrList, addrs)
+
+	// Randomize the addresses sent if there are more than the maximum.
+	if count > wire.MaxV2AddrPerMsg {
+		// Shuffle the address list.
```
Roasbeef, line 895: Can use this std libb method in place: https://pkg.go.dev/math/rand#Shuffle

peer/peer.go
```diff
@@ -2092,14 +2151,71 @@ func (p *Peer) writeLocalVersionMsg() error {
...
+func (p *Peer) waitToFinishNegotiation(pver uint32) error {
+	// There are several possible messages that can be received here. We
+	// could immediately receive verack and be done with the handshake. We
+	// could receive sendaddrv2 and still have to wait for verack. Or we
+	// can receive unknown messages before and after sendaddrv2 and still
+	// have to wait for verack.
+	for {
+		remoteMsg, _, err := p.readMessage(wire.LatestEncoding)
+		if err == wire.ErrUnknownMessage {
+			continue
```
Roasbeef, line 2178: Log the message?

peer/peer.go
```diff
@@ -2289,7 +2434,12 @@ func NewOutboundPeer(cfg *Config, addr string) (*Peer, error) {
 		}
 		p.na = na
 	} else {
-		p.na = wire.NewNetAddressIPPort(net.ParseIP(host), uint16(port), 0)
+		// If host is an onion hidden service, it is likely that a
```
Roasbeef, line 2437: Can we prevent this at the API level? Otherwise, I think we should update the `godoc` to warn callers.

peer/peer.go
```diff
@@ -197,6 +200,9 @@ type MessageListeners struct {
 	// message.
 	OnSendHeaders func(p *Peer, msg *wire.MsgSendHeaders)
 
+	// OnSendAddrV2 is invoked when a peer receives a sendaddrv2 message.
+	OnSendAddrV2 func(p *Peer, msg *wire.MsgSendAddrV2)
```
Roasbeef, line 204: I don't see us ever specify this in the server peer?

Roasbeef (changes requested):
Great work! Completed an initial pass, and will start to run this using `btcd` and `neutrino` on testnet 

My main question is regarding backwards compatibility: are we still able to read the _old_ addr v1 JSOn messages? As a follow up: will we convert all the addrv1 to addrv2 on disk when we first start using this PR?

wire/netaddressv2.go
```diff
@@ -0,0 +1,576 @@
...
+)
+
+var (
+	// ErrInvalidAddressSize is an error that means an incorrect address
+	// size was decoded for a networkID or that the address exceeded the
+	// maximum size for an unknown networkID.
+	ErrInvalidAddressSize = fmt.Errorf("invalid address size")
+
+	// ErrSkippedNetworkID is returned when the cjdns, i2p, or unknown
+	// networks are encountered during decoding. btcd does not support i2p
```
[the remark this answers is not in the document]
Crypt-iQ: BIP suggested to not do this in case we gossip bad addresses and cause our peers to ban us, if the peer bans

peer/peer.go
Crypt-iQ: This is only for tests, there's not really a reason to use it otherwise but thought I'd add it for completeness.

addrmgr/addrmanager.go
Crypt-iQ: The `ip` field gets parsed and then it branches, it's not the cleanest way but it was pre-existing. I can definitely change it for readability

peer/peer.go
Crypt-iQ: It can be prevented with an error right there rather than continuing

Crypt-iQ: I updated the godoc, `wire.NetAddressV2FromBytes` would need to return an error to be able to fail gracefully here.

server.go
Crypt-iQ: Updated godoc

wire/netaddressv2.go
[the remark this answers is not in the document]
Roasbeef: Ah gotcha, makes sense.

Roasbeef:
Giving this a spin on mainnet+testnet now.

Roasbeef:
Attempted to do a diff of my `peers.json` file before and after, but then realized that won't really help here, since map iteration order and json encoding aren't deterministic...

Roasbeef:
cc @ellemouton

ellemouton (commented):
Looks good! Did a pass, left some nits & questions. Tested on regtest that nodes running this patch can successfully complete the handshake with older and newer btcd nodes and with new bitcoind nodes.

wire/netaddressv2.go
```diff
@@ -0,0 +1,588 @@
...
+	// addrv2-supporting btcd software to disconnect upon receiving the new
+	// addresses. This error can also be returned when an OnionCat-encoded
+	// torv2 address is received with the ipv6 networkID. This error
+	// signals to the caller to continue reading.
+	ErrSkippedNetworkID = fmt.Errorf("skipped networkID")
+)
+
+// maxNetAddressV2Payload returns the max payload size for an address used in
+// the addrv2 message.
+func maxNetAddressV2Payload() uint32 {
```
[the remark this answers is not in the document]
Crypt-iQ: nope, can change, just wanted to show the derivation

wire/netaddressv2.go
```diff
@@ -0,0 +1,588 @@
...
+
+// isKnownNetworkID returns true if the networkID is one listed above and false
+// otherwise.
+func isKnownNetworkID(netID uint8) bool {
+	return uint8(ipv4) <= netID && netID <= uint8(cjdns)
+}
+
+type ipv4Addr struct {
+	addr  [ipv4Size]byte
+	netID networkID
```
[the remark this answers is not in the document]
Crypt-iQ: `Network()` returns a string, and is only provided to make the structs conform to the net.Addr interface, so we'd have to convert back and forth to an int which I wanted to avoid. I think a `newIPV4Addr` and related constructors might work, but there may have been a reason I'm forgetting that I didn't do this. How would addr size fit in here?

addrmgr/addrmanager.go
```diff
@@ -745,30 +768,23 @@ func (a *AddrManager) HostToNetAddress(host string, port uint16, services wire.S
...
+		// above.
+		na = wire.NetAddressV2FromBytes(time.Now(), services, ip, port)
```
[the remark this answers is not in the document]
Crypt-iQ: If this branch is reached, then `ip` is either an IPv6 or IPv4 address. The parsing is a little funky because the above branch actually sets `ip` here. So `NetAddressV2FromBytes` should hit the ipv4/ipv6 case and return no problem

server.go
```diff
@@ -1328,6 +1356,45 @@ func (sp *serverPeer) OnAddr(_ *peer.Peer, msg *wire.MsgAddr) {
...
+// OnAddrV2 is invoked when a peer receives an addrv2 bitcoin message and is
+// used to notify the server about advertised addresses.
+func (sp *serverPeer) OnAddrV2(_ *peer.Peer, msg *wire.MsgAddrV2) {
+	// Ignore if simnet for the same reasons as the regular addr message.
+	if cfg.SimNet {
+		return
+	}
+
+	// An empty AddrV2 message is invalid.
+	if len(msg.AddrList) == 0 {
```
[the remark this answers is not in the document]
Crypt-iQ: Checked here (https://github.com/btcsuite/btcd/blob/cb6f21b598e90c8b1ea55e569b838797e9b37efe/wire/msgaddrv2.go#L30)

peer/peer.go
```diff
@@ -2109,12 +2223,22 @@ func (p *Peer) negotiateInboundProtocol() error {
...
 	}
 
+	var protoVersion uint32
+	p.flagsMtx.Lock()
+	protoVersion = p.protocolVersion
+	p.flagsMtx.Unlock()
+
+	if err := p.writeSendAddrV2Msg(protoVersion); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
Crypt-iQ: Since addrv2 did not come with a version bump (wtxidrelay also uses 70016 - WTXID_RELAY_VERSION in bitcoin core code), it wouldn't be possible to know if your peer supports addrv2 on the version alone (maybe they only support wtxidrelay). This approach is used by bitcoin core now to introduce new feature via messages (during handshake). I think version bumping could have been done, but maybe there's some other reason. In other words, I dont know

Roasbeef (approved):
LGTM 🐙

peer/peer.go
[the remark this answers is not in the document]
Roasbeef: Yeh agree that a version bump should have been done instead, as the current interactive negotiation stuff just serves to make the logic to figure out which features you can use more complicated, vs the way we do things on the LN side re looking at the feature bit intersection.

PR #1814, opened 2022-02-22
Fixes coveralls coverage report

vpereira01:
With PR #1785 multiple coverage reports are sent to coveralls. This makes coveralls only consider the last report sent. This fix treat all reports has parallel coverage reports so they can all be compiled later by coveralls.

Coverage and after before 
* Before https://coveralls.io/builds/44996210
* After https://coveralls.io/builds/45515236


jcvernaleo (approved):
Looks good.

OK

jcvernaleo:
Sorry for the slow response time on this PR.  Reviewed and merging it now.

PR #1815, opened 2022-02-22
harden daemon on OpenBSD

timkuijsten:
I've created a patch to harden the daemon on OpenBSD by using [pledge(2)](http://man.openbsd.org/pledge) and [unveil(2)](http://man.openbsd.org/unveil). This way the program gets killed if any syscall is called other than the basic networking or file system ones, preventing further exploitation of a system once a program is compromised. Furthermore, the view of the file system is restricted to only access the files and directories the daemon needs.

### DIFF
--- a/btcd.go
+++ b/btcd.go
@@ -18,6 +18,7 @@ import (
 	"github.com/btcsuite/btcd/blockchain/indexers"
 	"github.com/btcsuite/btcd/database"
 	"github.com/btcsuite/btcd/limits"
+	"github.com/btcsuite/btcd/ossec"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


jcvernaleo:
Love the idea of taking advantage of pledge!

Looks like the tests are failing in the GH actions though (which run on linux).

timkuijsten:
> Looks like the tests are failing in the GH actions though (which run on linux).

This is fixed now.

Whether or not we would like to pledge tools like btcctl is something we can decide later and IMHO not as important as seat belting the daemon.

jcvernaleo (approved):
Looks great.

OK

jcvernaleo:
@timkuijsten totally agree that daemon is the most important target for pledge.  Other tools are less vital but all else being equal, the more things that use pledge, the better.

PR #1820, opened 2022-03-02
btcec/schnorr/musig2: add new musig2 implementation based on musig2 draft BIP

Roasbeef:
In this commit, we add a fully featured `musig2` implementation capable of generating partial signatures, combining them, and finally ensuring that the combined signature is a valid BIP-340 schnorr signature. This implementation is based off of [this spec](https://github.com/ElementsProject/secp256k1-zkp/blob/master/doc/musig-spec.mediawiki) from the `secp256k1-zkp` repo. 

I consider much of the external API to still be a WIP, but everything is functional as is. The API will likely evolve as we start to use it upstream in `lnd`, `pool`, `loop`, etc, etc. 

Before this PR is ready for merging, I want to further optimize it (adding benchmarks along the way), as during signing and nonce generation a lot of values end up being computed multiple times, when instead they can be computed once and memoized from there. I think it would also be useful to export a nicer version of the bookkeeping struct created in the final test for multi party signing. 

TODOs
- [x] add benchmarks 
- [x] optimize logic by memoizing common routines (blinding factor, combined keys, etc)
- [x] create better multi-party signing API (context to hold all the keys, etc), similar to what's added in the final test
- [x] add support for BIP 341 taproot tweak output key derivation
- [x] add explicit support for BIP 86 output key derivation
- [x] extend context API to allow nonce generation before all keys are known (needed for certain contexts like LN funding)
- [x] properly cache intermediate computations to avoid redoing it all several times in the process

### DIFF
--- a/btcec/schnorr/musig2/context.go
+++ b/btcec/schnorr/musig2/context.go
@@ -10,6 +10,12 @@ import (
 )
 
 var (
+	// ErrSignersNotSpecified is returned when a caller attempts to create
+	// a context without specifying either the total number of signers, or
+	// the complete set of singers.
+	ErrSignersNotSpecified = fmt.Errorf("total number of signers or all " +
+		"signers must be known")
+
 	// ErrSignerNotInKeySet is returned when a the private key for a signer
 	// isn't included in the set of signing public keys.
 	ErrSignerNotInKeySet = fmt.Errorf("signing key is not found in key" +
@@ -18,6 +24,16 @@ var (
 	// ErrAlredyHaveAllNonces is called when RegisterPubNonce is called too
 	// many times for a given signing session.
 	ErrAlredyHaveAllNonces = fmt.Errorf("already have all nonces")
+
+	// ErrNotEnoughSigners is returned when a caller attempts to create a
+	// session from a context, but before all the required signers are
+	// known.
+	ErrNotEnoughSigners = fmt.Errorf("not enough signers")
+
+	// ErrAlredyHaveAllNonces is returned when a caller attempts to
+	// register a signer, once we already have the total set of known
+	// signers.
+	ErrAlreadyHaveAllSigners = fmt.Errorf("all signers registered")
 
 	// ErrAlredyHaveAllSigs is called when CombineSig is called too many
 	// times for a given signing session.
@@ -39,6 +55,10 @@ var (
 	// ErrTaprootInternalKeyUnavailable is returned when a user attempts to
 	// obtain the
 	ErrTaprootInternalKeyUnavailable = fmt.Errorf("taproot tweak not used")
+
+	// ErrNotEnoughSigners is returned if a caller attempts to obtain an
+	// early nonce when it wasn't specified
+	ErrNoEarlyNonce = fmt.Errorf("no early nonce available")
 )
 
 // Context is a managed signing context for musig2. It takes care of things
@@ -49,9 +69,6 @@ type Context struct {
 
 	// pubKey is our even-y coordinate public  key.
 	pubKey *btcec.PublicKey
-
-	// keySet is the set of all signers.
-	keySet []*btcec.PublicKey
 
 	// combinedKey is the aggregated public key.
 	combinedKey *AggregateKey
@@ -69,6 +86,10 @@ type Context struct {
 	// shouldSort keeps track of if the public keys should be sorted before
 	// any operations.
 	shouldSort bool
+
+	// sessionNonce will be populated if the earlyNonce option is true.
+	// After the first session is created, this nonce will be blanked out.
+	sessionNonce *Nonces
 }
 
 // ContextOption is a functional option argument that allows callers to modify
@@ -94,6 +115,17 @@ type contextOptions struct {
 	// bip86Tweak if true, then the weak will just be
 	// h_tapTweak(internalKey) as there is no true script root.
 	bip86Tweak bool
+
+	// keySet is the complete set of signers for this context.
+	keySet []*btcec.PublicKey
+
+	// numSigners is the total number of signers that will eventually be a
+	// part of the context.
+	numSigners int
+
+	// earlyNonce determines if a nonce should be generated during context
+	// creation, to be automatically passed to the created session.
+	earlyNonce bool
 }
 
 // defaultContextOptions returns the default context options.
@@ -129,12 +161,43 @@ func WithBip86TweakCtx() ContextOption {
 	}
 }
 
+// WithKnownSigners is an optional parameter that should be used if a session
+// can be created as soon as all the singers are known.
+func WithKnownSigners(signers []*btcec.PublicKey) ContextOption {
+	return func(o *contextOptions) {
+		o.keySet = signers
+		o.numSigners = len(signers)
+	}
+}
+
+// WithNumSigners is a functional option used to specify that a context should
+// be created without knowing all the signers. Instead the total number of
+// signers is specified to ensure that a session can only be created once all
+// the signers are known.
+//
+// NOTE: Either WithKnownSigners or WithNumSigners MUST be specified.
+func WithNumSigners(n int) ContextOption {
+	return func(o *contextOptions) {
+		o.numSigners = n
+	}
+}
+
+// WithEarlyNonceGen allow a caller to specify that a nonce should be generated
+// early, before the session is created. This should be used in protocols that
+// require some partial nonce exchange before all the signers are known.
+//
+// NOTE: This option must only be specified with the WithNumSigners option.
+func WithEarlyNonceGen() ContextOption {
+	return func(o *contextOptions) {
+		o.earlyNonce = true
+	}
+}
+
 // NewContext creates a new signing context with the passed singing key and set
 // of public keys for each of the other signers.
 //
 // NOTE: This struct should be used over the raw Sign API whenever possible.
-func NewContext(signingKey *btcec.PrivateKey,
-	signers []*btcec.PublicKey, shouldSort bool,
+func NewContext(signingKey *btcec.PrivateKey, shouldSort bool,
 	ctxOpts ...ContextOption) (*Context, error) {
 
 	// First, parse the set of optional context options.
@@ -143,10 +206,6 @@ func NewContext(signingKey *btcec.PrivateKey,
 		option(opts)
 	}
 
-	// As a sanity check, make sure the signing key is actually amongst the sit
-	// of signers.
-	//
-	// TODO(roasbeef): instead have pass all the _other_ signers?
 	pubKey, err := schnorr.ParsePubKey(
 		schnorr.SerializePubKey(signingKey.PubKey()),
 	)
@@ -154,64 +213,156 @@ func NewContext(signingKey *btcec.PrivateKey,
 		return nil, err
 	}
 
+	ctx := &Context{
+		signingKey: signingKey,
+		pubKey:     pubKey,
+		opts:       opts,
+		shouldSort: shouldSort,
+	}
+
+	switch {
+
+	// We know all the signers, so we can compute the aggregated key, along
+	// with all the other intermediate state we need to do signing and
+	// verification.
+	case opts.keySet != nil:
+		if err := ctx.combineSignerKeys(); err != nil {
+			return nil, err
+		}
+
+	// The total signers are known, so we add ourselves, and skip key
+	// aggregation.
+	case opts.numSigners != 0:
+		// Otherwise, we'll add ourselves as the only known signer, and
+		// await further calls to RegisterSigner before a session can
+		// be created.
+		opts.keySet = make([]*btcec.PublicKey, 0, opts.numSigners)
+		opts.keySet = append(opts.keySet, pubKey)
+
+		// If early nonce generation is specified, then we'll generate
+		// the nonce now to pass in to the session once all the callers
+		// are known.
+		if opts.earlyNonce {
+			ctx.sessionNonce, err = GenNonces()
+			if err != nil {
+				return nil, err
+			}
+		}
+
+	default:
+		return nil, ErrSignersNotSpecified
+	}
+
+	return ctx, nil
+}
+
+// combineSignerKeys is used to compute the aggregated signer key once all the
+// signers are known.
+func (c *Context) combineSignerKeys() error {
+	// As a sanity check, make sure the signing key is actually
+	// amongst the sit of signers.
 	var keyFound bool
-	for _, key := range signers {
-		if key.IsEqual(pubKey) {
+	for _, key := range c.opts.keySet {
+		if key.IsEqual(c.pubKey) {
 			keyFound = true
 			break
 		}
 	}
 	if !keyFound {
-		return nil, ErrSignerNotInKeySet
+		return ErrSignerNotInKeySet
 	}
 
-	// Now that we know that we're actually a signer, we'll generate the
-	// key hash finger print and second unique key index so we can speed up
-	// signing later.
-	keysHash := keyHashFingerprint(signers, shouldSort)
-	uniqueKeyIndex := secondUniqueKeyIndex(signers, shouldSort)
+	// Now that we know that we're actually a signer, we'll
+	// generate the key hash finger print and second unique key
+	// index so we can speed up signing later.
+	c.keysHash = keyHashFingerprint(c.opts.keySet, c.shouldSort)
+	c.uniqueKeyIndex = secondUniqueKeyIndex(
+		c.opts.keySet, c.shouldSort,
+	)
 
 	keyAggOpts := []KeyAggOption{
-		WithKeysHash(keysHash), WithUniqueKeyIndex(uniqueKeyIndex),
+		WithKeysHash(c.keysHash),
+		WithUniqueKeyIndex(c.uniqueKeyIndex),
 	}
 	switch {
-	case opts.bip86Tweak:
+	case c.opts.bip86Tweak:
 		keyAggOpts = append(
 			keyAggOpts, WithBIP86KeyTweak(),
 		)
-	case opts.taprootTweak != nil:
+	case c.opts.taprootTweak != nil:
 		keyAggOpts = append(
-			keyAggOpts, WithTaprootKeyTweak(opts.taprootTweak),
+			keyAggOpts, WithTaprootKeyTweak(c.opts.taprootTweak),
 		)
-	case len(opts.tweaks) != 0:
-		keyAggOpts = append(keyAggOpts, WithKeyTweaks(opts.tweaks...))
+	case len(c.opts.tweaks) != 0:
+		keyAggOpts = append(keyAggOpts, WithKeyTweaks(c.opts.tweaks...))
 	}
 
-	// Next, we'll use this information to compute the aggregated public
-	// key that'll be used for signing in practice.
-	combinedKey, _, _, err := AggregateKeys(
-		signers, shouldSort, keyAggOpts...,
+	// Next, we'll use this information to compute the aggregated
+	// public key that'll be used for signing in practice.
+	var err error
+	c.combinedKey, _, _, err = AggregateKeys(
+		c.opts.keySet, c.shouldSort, keyAggOpts...,
 	)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
-	return &Context{
-		signingKey:     signingKey,
-		pubKey:         pubKey,
-		keySet:         signers,
-		combinedKey:    combinedKey,
-		uniqueKeyIndex: uniqueKeyIndex,
-		keysHash:       keysHash,
-		opts:           opts,
-		shouldSort:     shouldSort,
-	}, nil
+	return nil
+}
+
+// EarlySessionNonce returns the early session nonce, if available.
+func (c *Context) EarlySessionNonce() (*Nonces, error) {
+	if c.sessionNonce == nil {
+		return nil, ErrNoEarlyNonce
+	}
+
+	return c.sessionNonce, nil
+}
+
+// RegisterSigner allows a caller to register a signer after the context has
+// been created. This will be used in scenarios where the total number of
+// signers is known, but nonce exchange needs to happen before all the signers
+// are known.
+//
+// A bool is returned which indicates if all the signers have been registered.
+//
+// NOTE: If the set of keys are not to be sorted during signing, then the
+// ordering each key is registered with MUST match the desired ordering.
+func (c *Context) RegisterSigner(pub *btcec.PublicKey) (bool, error) {
+	haveAllSigners := len(c.opts.keySet) == c.opts.numSigners
+	if haveAllSigners {
+		return false, ErrAlreadyHaveAllSigners
+	}
+
+	c.opts.keySet = append(c.opts.keySet, pub)
+
+	// If we have the expected number of signers at this point, then we can
+	// generate the aggregated key and other necessary information.
+	haveAllSigners = len(c.opts.keySet) == c.opts.numSigners
+	if haveAllSigners {
+		if err := c.combineSignerKeys(); err != nil {
+			return false, err
+		}
+	}
+
+	return haveAllSigners, nil
+}
+
+// NumRegisteredSigners returns the total number of registered signers.
+func (c *Context) NumRegisteredSigners() int {
+	return len(c.opts.keySet)
 }
 
 // CombinedKey returns the combined public key that will be used to generate
 // multi-signatures  against.
-func (c *Context) CombinedKey() *btcec.PublicKey {
-	return c.combinedKey.FinalKey
+func (c *Context) CombinedKey() (*btcec.PublicKey, error) {
+	// If the caller hasn't registered all the signers at this point, then
+	// the combined key won't be available.
+	if c.combinedKey == nil {
+		return nil, ErrNotEnoughSigners
+	}
+
+	return c.combinedKey.FinalKey, nil
 }
 
 // PubKey returns the public key of the signer of this session.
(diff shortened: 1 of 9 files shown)


btcec/schnorr/musig2/sign.go
```diff
@@ -0,0 +1,418 @@
...
+// nonce, and the series of partial signatures for each nonce.
+func CombineSigs(combinedNonce *btcec.PublicKey,
+	partialSigs []*PartialSignature) *schnorr.Signature {
+
+	var combinedSig btcec.ModNScalar
+	for _, partialSig := range partialSigs {
+		combinedSig.Add(partialSig.S)
+	}
+
+	// TODO(roasbeef): less verbose way to get the x coord...
```
Roasbeef, line 412:
```go
var xCoord btcec.FieldVal 
xCoord.SetByteSLice(combinedNonce.X().Bytes())
```

btcec/schnorr/musig2/keys.go
```diff
@@ -0,0 +1,145 @@
...
+	// key, can just store an index, otherwise this is O(n^2)
+	if isSecondKey(keySet, targetKey) {
+		return mu.SetInt(1)
+	}
+
+	// Otherwise, we'll compute the full finger print hash for this given
+	// key and then use that to compute the coefficient tagged hash:
+	//  * H(tag=KeyAgg coefficient, keyHashFingerprint(pks, pk) || pk)
+	var coefficientBytes bytes.Buffer
+	coefficientBytes.Write(keyHashFingerprint(keySet))
```
matheusd, line 100: Not sure if you intend this to follow the same API as some other impl (I'm not familiar with the existing musig impls), but this can be trivially moved to `AggregateKeys` to avoid repeating the hash.

For reference, here's my test impl for 3 round musig: https://github.com/matheusd/mrttree/blob/master/primitives.go#L80-L99

Roasbeef: Yep, as noted in the PR body I've done zero optimizations as is. My plan is to memoize/cache all this information in an internal signing/verification context.

Roasbeef: The other thing that can make things slower for really a really large number of signers is that the `isSecondKey` function ends up being O(N^2). Instead we can compute that once, then just record the key index and eliminate all the looping/serialization.

btcec/schnorr/musig2/nonces.go
```diff
@@ -0,0 +1,209 @@
...
+	// SecNonce holds the two 32-byte scalar values that are the private
+	// keys to the two public nonces.
+	SecNonce [SecNonceSize]byte
+}
+
+// secNonceToPubNonce takes our two secrete nonces, and produces their two
+// corresponding EC points, serialized in compressed format.
+func secNonceToPubNonce(secNonce [SecNonceSize]byte) [PubNonceSize]byte {
+	var k1Mod, k2Mod btcec.ModNScalar
+	k1Mod.SetByteSlice(secNonce[:])
```
guggero, line 42: This makes an assumption about the implementation of the underlying `SetByteSlice` truncating anything after 32 bytes. Should we make it more explicit though by using `secNonce[:btcec.PrivKeyBytesLen]`?

guggero, line 42: This makes an assumption about the implementation of the underlying `SetByteSlice` truncating anything after 32 bytes. Should we make it more explicit though by using `secNonce[:btcec.PrivKeyBytesLen]`?

btcec/schnorr/musig2/nonces.go
```diff
@@ -0,0 +1,209 @@
...
+			btcec.AddNonConst(
+				&aggregateNonce, pubNonceJ, &aggregateNonce,
+			)
+		}
+
+		// Now that we've aggregated all the points, we need to check
+		// if this point is the point at infinity, if so, then we'll
+		// just return the generator. At a later step, the malicious
+		// party will be detected.
+		var infinityPoint btcec.JacobianPoint
```
guggero, line 165: nit: this is later re-declared on the package level. Perhaps do that in this commit already and re-use it? Since it will be shadowed later.

guggero, line 165: nit: this is later re-declared on the package level. Perhaps do that in this commit already and re-use it? Since it will be shadowed later.

btcec/schnorr/musig2/keys.go
```diff
@@ -0,0 +1,145 @@
+// Copyright 2013-2016 The btcsuite developers
+// Copyright (c) 2015-2021 The Decred developers
```
guggero, line 2: Is there actual code being copied from Decred? Or could this just be `2013-2022 The btcsuite developers`? Same in all other files.

guggero, line 2: Is there actual code being copied from Decred? Or could this just be `2013-2022 The btcsuite developers`? Same in all other files.

btcec/schnorr/musig2/sign.go
```diff
@@ -0,0 +1,418 @@
...
+	// Next, we'll compute mu, our aggregation coefficient for the key that
+	// we're signing with.
+	mu := aggregationCoefficient(pubKeys, pubKey)
+
+	// With mu constructed, we can finally generate our partial signature
+	// as: s = (k1_1 + b*k_2 + e*mu*d) mod n.
+	s := new(btcec.ModNScalar)
+	s.Add(&k1).Add(k2.Mul(&nonceBlinder)).Add(e.Mul(mu).Mul(&privKeyScalar))
+
+	sig := NewPartialSignature(s, nonceKey)
```
guggero, line 225: I think we should overwrite the `secnonce` with all zeros here (might need to change it to a pointer type to have an effect) to avoid signing with the same nonce again. Could add a check at the beginning if it is all zeros.

guggero, line 225: I think we should overwrite the `secnonce` with all zeros here (might need to change it to a pointer type to have an effect) to avoid signing with the same nonce again. Could add a check at the beginning if it is all zeros.

guggero (commented):
Awesome PR! Looks very clean and follows the spec very closely, which both makes it easy to review. Did a first pass, will run some interoperability tests once I've updated my JS implementation to the latest spec.

btcec/schnorr/musig2/keys.go
```diff
@@ -0,0 +1,145 @@
...
+		keyBytes.Write(schnorr.SerializePubKey(key))
+	}
+
+	h := chainhash.TaggedHash(KeyHashTag, keyBytes.Bytes())
+	return h[:]
+}
+
+// isSecondKey returns true if the passed public key is the second key in the
+// (sorted) keySet passed.
+func isSecondKey(keySet []*btcec.PublicKey, targetKey *btcec.PublicKey) bool {
```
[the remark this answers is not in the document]
Roasbeef: Function is now gone after optimizations. We also don't assume that the list is sorted (see the test vectors for an example). We we need in the end here index of the second unique key, so we just compute that directly now, which can save us from an O(N^2) when the unique element is actually the very last one.

btcec/schnorr/musig2/nonces.go
Roasbeef: Done!

btcec/schnorr/musig2/sign.go
Roasbeef: Good idea! I've added this, along with a functional option to disable it for benchmarks, etc.

Roasbeef:
cc @Crypt-iQ

Roasbeef:
I've scrapped the prior approach and pushed up a new final commit that introduces an easier way to manage multi-party signing, along with a best-effort attempt at detecting nonce reuse. With that, the final checkbox on this PR is complete!

btcec/schnorr/musig2/sign.go
```diff
@@ -0,0 +1,418 @@
...
+		k2.Negate()
+	}
+
+	privKeyScalar := privKey.Key
+	if privKeyScalar.IsZero() {
+		return nil, ErrPrivKeyZero
+	}
+
+	// If the y coordinate of the public key is odd xor the y coordinate of
+	// the combined public key is odd, then we'll negate the private key.
```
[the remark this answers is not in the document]
Roasbeef: We regenerate the public key here from the private key. Even if the public key is serialized in x-only (we lose the information related to the y coordinate), the _true_ y coordinate for the key may have an odd y coordinate.

Doing this makes implementation a bit easier, as bip32 clients don't need to recycle keys until they end up with one that has an even y coord.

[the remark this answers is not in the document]
Roasbeef: BIP 340 has a similar step, at step 4 here: https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#default-signing

btcec/schnorr/musig2/sign.go
```diff
@@ -0,0 +1,418 @@
...
+	// If the combined nonce it eh point at infinity, then we'll bail out.
+	if nonce == infinityPoint {
+		return nil, ErrNoncePointAtInfinity
+	}
+
+	// Next we'll parse out our two secret nonces, which we'll be using in
+	// the core signing process below.
+	//
+	// TODO(roasbeef): no > n check actually exists here since mod n
+	// scalar, need to check overflow value instead?
```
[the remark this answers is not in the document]
Roasbeef: Nice catch, added!

Roasbeef:
Addressed latest comments, and fixed a possible issue w.r.t the way sorting was handled. Right now we check for a few places before sorting which makes things more efficient (n comparison ops vs nlogn). If the code is made a bit more clever, we can pass through context with the set of keys themselves (making a new internal wrapper struct) if things are sorted or not...

btcec/schnorr/musig2/sign.go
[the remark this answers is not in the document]
Crypt-iQ: ah ok, that confused me

btcec/schnorr/musig2/nonces.go
```diff
@@ -22,6 +19,10 @@ const (
 	SecNonceSize = 64
 )
 
+// zeroSecNonce is a secret nonce that's all zeroes. This is used to check that
+// we're not attempting to re-use a nonce, and also protect callers from it.
+var zeroSecNonce [SecNonceSize]byte
```
Crypt-iQ, line 24: unused?

Crypt-iQ (commented):
lgtm, still need to look at tests

btcec/schnorr/musig2/keys.go
```diff
@@ -0,0 +1,166 @@
...
+	h := chainhash.TaggedHash(KeyAggTagList, keyBytes.Bytes())
+	return h[:]
+}
+
+// isSecondKey returns true if the passed public key is the second key in the
+// (sorted) keySet passed.
+func isSecondKey(keySet []*btcec.PublicKey, targetKey *btcec.PublicKey) bool {
+	// For this comparison, we want to compare the raw serialized version
+	// instead of the full pubkey, as it's possible we're dealing with a
+	// pubkey that _actually_ has an odd y coordinate.
```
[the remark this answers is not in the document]
Roasbeef: Yeah, lets say we didn't make sure the signing key in this context had an even y coord, then it would show up here, but the sorting/uniqueness only considers the x coord.

btcec/schnorr/musig2/nonces.go
Roasbeef: Ah yeah, was left over from an attempt to zero out nonces when signing, but ended up rolling that context struct in stead.

Roasbeef:
Pushed up a commit to address the latest comments. The final question in my mind here is: should we make the existing `Sign` method _private_ so we _force_ users to always use the safer API? The main usage of the lower level sign method imo is to implement more advanced signing patterns like adaptor signatures, etc.

btcec/schnorr/musig2/context.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+// Session represents a musig2 signing session. A new instance should be
+// created each time a multi-signature is needed. The session struct handles
+// nonces management, incremental partial sig vitrifaction, as well as final
+// signature combination. Errors are returned when unsafe behavior such as
+// nonce re-use is attempted.
+//
+// NOTE: This struct should be used over the raw Sign API whenever possible.
+type Session struct {
```
guggero, line 133: I really like this API! It should make this quite easy to use and also safe.

One use case I was thinking about that wouldn't be possible with the current API is the channel funding 2-of-2 MuSig2 case. Because if we want to send our nonces over in the `open_channel` message, we wouldn't know the other party's public key yet, so couldn't create a full context yet. But we'd need a session to get access to "safe" nonces.

Maybe we can combine the two structs and keep some kind of progression state for the different steps?

guggero (commented):
Looks really good with the new Session+Context API!
I created a PR to end-to-end integrate this into an `lnd` itest and have found a few small issues.
And I wasn't able to produce a valid signature just yet, but that's probably a small thing in my code.

btcec/schnorr/musig2/sign.go
```diff
@@ -0,0 +1,423 @@
...
+	R *btcec.PublicKey
+}
+
+// NewPartialSignature returns a new instances of the partial sig struct.
+func NewPartialSignature(s *btcec.ModNScalar,
+	r *btcec.PublicKey) PartialSignature {
+
+	return PartialSignature{
+		S: s,
+		R: r,
```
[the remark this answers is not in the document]
Roasbeef: This is the value that's needed to verify the final combined signature, so I ended up stuffing it here (before the context struct was created).

btcec/schnorr/musig2/context.go
```diff
@@ -0,0 +1,280 @@
...
+	// any operations.
+	shouldSort bool
+}
+
+// NewContext creates a new signing context with the passed singing key and set
+// of public keys for each of the other signers.
+//
+// NOTE: This struct should be used over the raw Sign API whenever possible.
+func NewContext(signingKey *btcec.PrivateKey,
+	signers []*btcec.PublicKey, shouldSort bool) (*Context, error) {
```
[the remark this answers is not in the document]
Roasbeef: What do you mean it fails? If the key actually has an even y coord, then that'll effectively just lose that bit and present it as a key with an odd y coordinate.

btcec/schnorr/musig2/context.go
```diff
@@ -0,0 +1,280 @@
...
+	// shouldSort keeps track of if the public keys should be sorted before
+	// any operations.
+	shouldSort bool
+}
+
+// NewContext creates a new signing context with the passed singing key and set
+// of public keys for each of the other signers.
+//
+// NOTE: This struct should be used over the raw Sign API whenever possible.
+func NewContext(signingKey *btcec.PrivateKey,
```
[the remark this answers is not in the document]
Roasbeef: > Could we abstract the signing key behind an interface so we don't need to specify it here already? I'm trying to think of a way to support remote signing in a way where the watch-only node could keep track of the signing sessions and would only redirect the raw Sign() call to the signing node

Hmm, I think that's possible. But in the remote signing setting, isn't it the case that the caller already doesn't have access to the private key, and all the session state would actually live entirely in the remote signer? 

I can see how it can be nice to have the signing node be stateless, but given that it's the one that needs to manage the nonce state (which is where the danger is), I'm not sure how we can avoid it having that extra _in memory state_ (assuming all protocols are designed to drop the nonce state if things need to resume of connections be re-established or w/e). 

> Maybe a new SignSchnorr RPC on the signing service would help here?

Yeah this was my original idea, but then I saw the way you sort of made the decision implicitly when signing outputs to figure out if ECDSA vs schnorr should be used. One thing to keep in mind here though, is that the schnorr partial signature *is not* the same as a normal BIP-340 signature. So we'd need a `SignPartialSchnorr` that would only be used in the musig2 case, with something else being used for generic bip 340 signatures outside of the Bitcoin sighash scenario.

btcec/schnorr/musig2/context.go
Roasbeef: > One use case I was thinking about that wouldn't be possible with the current API is the channel funding 2-of-2 MuSig2 case. Because if we want to send our nonces over in the open_channel message, we wouldn't know the other party's public key yet,

Great point! I was coming at this from the PoV of initial commitment signing, where we already had created the joint public key at that point... 

One way to handle this would be to add a functional argument here that can be passed in which just uses the nonces specified by the caller. This would also be useful for deterministic tests. This way, you either initialize with the set of public keys (then the nonce is produced), or you do it with just your key (in order to gain access to the nonce) with you specifying the keys of the remote party in a future step. 

Would need to think about this a bit though, as it then exposes the caller (somewhat) to nonce generation themselves, but I think we can wrap that up at a high level.

Roasbeef: Or we permit passing in only a single key there, generate the nonces as normal, but return errors from certain calls until the other key is registered? So we'd allow a dynamic key set, with a similar `Register` method for the additional set of keys.

btcec/schnorr/musig2/context.go
```diff
@@ -0,0 +1,280 @@
...
+
+	s.pubNonces = append(s.pubNonces, localNonces.PubNonce)
+
+	return s, nil
+}
+
+// PublicNonce returns the public nonce for a signer. This should be sent to
+// other parties before signing begins, so they can compute the aggregated
+// public nonce.
+func (s *Session) PublicNonce() [PubNonceSize]byte {
```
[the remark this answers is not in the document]
Roasbeef: Hmm the idea was that you'd only store the context (since it's just the set of keys), and never store (assuming you mean write to disk?) the session. Are there any relevant fields you think should be exported in the context itself?

btcec/schnorr/musig2/context.go
```diff
@@ -0,0 +1,280 @@
...
+// like securely generating secret nonces, aggregating keys and nonces, etc.
+type Context struct {
+	// signingKey is the key we'll use for signing.
+	signingKey *btcec.PrivateKey
+
+	// pubKey is our even-y coordinate public  key.
+	pubKey *btcec.PublicKey
+
+	// keySet is the set of all signers.
+	keySet []*btcec.PublicKey
```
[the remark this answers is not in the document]
Roasbeef: Added a new `SigningKeys` method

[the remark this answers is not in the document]
Roasbeef: Or should this instead live in the session struct?

btcec/schnorr/musig2/context.go
```diff
@@ -0,0 +1,280 @@
...
+// signature combination. Errors are returned when unsafe behavior such as
+// nonce re-use is attempted.
+//
+// NOTE: This struct should be used over the raw Sign API whenever possible.
+type Session struct {
+	ctx *Context
+
+	localNonces *Nonces
+
+	pubNonces [][PubNonceSize]byte
```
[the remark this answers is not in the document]
Roasbeef: Added a `NumRegisteredNonces` method

btcec/schnorr/musig2/sign.go
```diff
@@ -0,0 +1,423 @@
...
+// isn't a valid schnorr signature by itself, as it needs to be aggregated
+// along with the other partial signatures to be completed.
+type PartialSignature struct {
+	S *btcec.ModNScalar
+
+	R *btcec.PublicKey
+}
+
+// NewPartialSignature returns a new instances of the partial sig struct.
+func NewPartialSignature(s *btcec.ModNScalar,
```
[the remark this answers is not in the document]
Roasbeef: Added an Encode/Decode, but only for the `S` value, as seems that's all you need to encode on the lnd side?

btcec/schnorr/musig2/context.go
Roasbeef: Also re context vs session, my initial idea was that both sides would maintain the some _context_, and then make new sessions on the fly each time they need to piggy back nonces into messages. By next week or so, I should have a more concrete sketch of the flow here both during funding as well as normal commitment updates.

btcec/schnorr/musig2/context.go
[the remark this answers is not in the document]
Roasbeef: Or you mean if you don't do this to the signer key amongst the passed set of `signers` then things fail? Or that we should just "re encode" everything here?

btcec/schnorr/musig2/context.go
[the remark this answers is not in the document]
guggero: I mean for keeping things in memory on the signing RPC side. We need to keep the session object in memory to be able to add nonces and partial signatures. But maybe we don't need the context after creating it and getting the combined key from it.

btcec/schnorr/musig2/context.go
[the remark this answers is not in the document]
guggero: Sorry, the comment was a bit imprecise. I got the `ErrSignerNotInKeySet` error when I didn't re-encode all the signer public keys. I was able to fix it by just doing the same `parse(serialize())` to all of them, including the local key. But just from the API/godoc it wasn't clear to me whether I should do that or not (since that will have an effect on the resulting signature for sure?).

Roasbeef:
Aight pushed another commit that supports taproot-native tweaking. A few new options have been added to make things easier in the context, and the may context also provides access to the aggregated key before tweaking, since that is needed to be used as the internal key.

Roasbeef:
Added another commit with explicit BIP 86 support (so signing with an output key that provably doesn't commit to a script path).

btcec/schnorr/musig2/keys.go
```diff
@@ -370,18 +391,34 @@ func AggregateKeys(keys []*btcec.PublicKey, sort bool,
...
 	// as the internal key for script path proofs.
+	finalKeyJ.ToAffine()
 	combinedKey := btcec.NewPublicKey(&finalKeyJ.X, &finalKeyJ.Y)
 
 	// At this point, if this is a taproot tweak, then we'll modify the
 	// base tweak value to use the BIP 341 tweak value.
 	if opts.taprootTweak {
+		// Emulate the same behavior as txscript.ComputeTaprootOutputKey
+		// which only operates on the x-only public key.
+		key, _ := schnorr.ParsePubKey(schnorr.SerializePubKey(
```
Roasbeef, line 402: I don't think this extra step is actually needed, since we commit to the x-only key below in the hash.

guggero (commented):
Amazing work! The API is very nice to use and feature complete for our first use cases in `lnd`.

We might want to make it possible to supply the private and public nonces to a session for certain use cases. But IMO that can also be done in a follow-up PR.

Found a few typos while working with the API, feel free to ignore :sweat_smile:

btcec/schnorr/musig2/context.go
```diff
@@ -0,0 +1,280 @@
...
+
+var (
+	// ErrSignerNotInKeySet is returned when a the private key for a signer
+	// isn't included in the set of signing public keys.
+	ErrSignerNotInKeySet = fmt.Errorf("Signing key is not found in key" +
+		" set")
+
+	// ErrAlredyHaveAllNonces is called when RegisterPubNonce is called too
+	// many times for a given signing session.
+	ErrAlredyHaveAllNonces = fmt.Errorf("Already have all nonces")
```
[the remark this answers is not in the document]
Roasbeef: Properly used now,

Roasbeef:
Ok, I've pushed a final (?) set of commits that adds the new nocne generation scheme in the updated spce (as a new commit), and also updates to add the remaining test vectors. 

I'll make a _new_ PR to allow the nonce generation to take place before the session itself has been created. 

One thing to note is that I used type params in one of the new commits, so that makes the module now require Go 1.18.

Roasbeef:
Alright, I just pushed the final (for real this time) commit which adds the ability to create a context without knowing all the signers ahead of time. A test has been added to exercise this mode and show callers how it can be done. I've also started to pass in more information into the aux input for the new nonce generation.

chappjc:
I don't know what the generics and Go 1.18 change means for btcec/v2 users.  There are a huge number of projects using btcec and a handful already using btcec/v2, including go-ethereum now, and if btcec/v2 can't be built with even Go 1.17, that's something to avoid if at all possible.  _Possibly_ it only makes a difference if someone tries to compile the musig2 package, but I'm not sure.

If generics become required for btcec/v2 _before_ resolving [that ambiguous import issue](https://github.com/btcsuite/btcd/issues/1839.), there's probably no point of the fix in https://github.com/btcsuite/btcd/pull/1848 any more.  go-ethereum (and my own projects) will likely have to revert to the btcec that shipped with v0.22.0-beta. If btcec/v2.1.4 could be cut with just that ambiguous import resolution, and/or this work would move btcec to v3, consumers could probably get by.

I wanna say it's gonna be fine with only the musig2 package using generics, but I'm a little anxious about that.
Does it look like this PR has to come first?  I wasn't completely clear about the tag changes that would make this first?

Roasbeef:
> I wanna say it's gonna be fine with only the musig2 package using generics, but I'm a little anxious about that.

Totally reasonable, the generics usage in that PR [is pretty trivial](https://github.com/Roasbeef/btcd/blob/afbf14a3a061b961c7fe0d21dcbbc6c941a33027/btcec/schnorr/musig2/nonces.go#L179). Removing it would just mean 10 or so extra more lines.

chappjc:
Cool, thanks for considering that.  However, I think @davecgh is on to a pretty good and relatively simple solution to a handful of issues, all with one v0.22.1 tag on a branch forking back off of the v0.22.0-beta tag's commit: https://github.com/btcsuite/btcd/issues/1839#issuecomment-1111642262
I mean, that would solve some issues and make the fix proposed by me in https://github.com/btcsuite/btcd/pull/1848 unnecessary (which is a good thing), but still it's probably prudent to be cautious with the btcec/v2 module's use of generics and go 1.18 requirement.

guggero (approved):
LGTM, very cool to get the early nonce generation and additional entropy features in as well!
I updated the `lnd` integration PR and made sure everything still works end-to-end.

Roasbeef:
Ended up "unrolling" the generics usage in the latest diff: https://github.com/btcsuite/btcd/compare/afbf14a3a061b961c7fe0d21dcbbc6c941a33027..93f45d9c94cf3aaf7850fff81335e68681e1275f#diff-483dff6c6a92e57a03ff4489e863f31d6b1e48f5b8ec036602cb3802471a8829R178-R200 

Pending CI, we should be g2g after this. Once this lands, I'll tag a new version of the `btcec` module.

Roasbeef:
Build looking good now, thanks for all the review! Excited for devs to experiment w/ this stuff in the wild

Issue #1822, opened 2022-03-07
break cyclic module require from btcec/v2 and btcd


chappjc:
https://github.com/btcsuite/btcd/blob/0c6dbfc3eac5ae817e61a1a3eaabdccffa003d2a/btcec/go.mod#L6

Please break the btcec/v2 module free of the main btcd module.  The two modules requiring each other is problematic for consumers.  If this requires making a chainhash module, I think that should be done asap.

When btcec/v2 is free of the btcd module require, There should be a new btcec/v2 tag, then *both* btcd and btcwallet should be updated as well.

With the btcd main module not following the rules of Go Semantic Import Versioning (SIV) it is very difficult for consumers to use these modules with other projects that also use them.  This is understandable as the main btcd is the CLI app, but it also includes many other packages.

PR #1780, opened 2021-12-08
Replace github.com/btcsuite/goleveldb imports with github.com/syndtr/goleveldb

anupcshan:
Generated using
```
$ find -name '*.go' | xargs sed -i "s#github.com/btcsuite/goleveldb#github.com/syndtr/goleveldb#g"
$ go mod tidy
$ go test -v ./...
```

Upstream goleveldb has 3+ years of bugfixes and improvements over the fork, which would be great to pull in.
I've been burned by a corruption which was fixed in https://github.com/syndtr/goleveldb/commit/eb135432c5aa4c841c91f3fdc871f07a94aa5a44.

Related to #1770

PR #1823, opened 2022-03-08
Remove circular dependency issue between `btcec/v2` and main package

guggero:
Fixes https://github.com/btcsuite/btcd/issues/1822.

After merging this PR we need to push two tags: `chaincfg/chainhash/v1.0.0` and `btcec/v2/v2.1.1`.
And I _think_ the two tags cannot be on the same commit (because go mod is just mean that way), so I'd suggest adding `chaincfg/chainhash/v1.0.0` to 7cc824e9 and `btcec/v2/v2.1.1` to 4ad74cd4.

cc @Roasbeef.

### DIFF
--- a/btcec/go.mod
+++ b/btcec/go.mod
@@ -9,3 +9,8 @@ require (
 )
 
 require github.com/decred/dcrd/crypto/blake256 v1.0.0 // indirect
+
+// We depend on chainhash as is, so we need to replace to use the version of
+// chainhash included in the version of btcd we're building in.
+// TODO(guggero): Remove this as soon as we have a tagged version of chainhash.
+replace github.com/btcsuite/btcd/chaincfg/chainhash => ../chaincfg/chainhash
--- a/btcec/go.mod
+++ b/btcec/go.mod
@@ -3,13 +3,9 @@ module github.com/btcsuite/btcd/btcec/v2
 go 1.17
 
 require (
-	github.com/btcsuite/btcd v0.22.0-beta.0.20220111032746-97732e52810c
+	github.com/btcsuite/btcd/chaincfg/chainhash v1.0.0
 	github.com/davecgh/go-spew v1.1.1
 	github.com/decred/dcrd/dcrec/secp256k1/v4 v4.0.1
 )
 
 require github.com/decred/dcrd/crypto/blake256 v1.0.0 // indirect
-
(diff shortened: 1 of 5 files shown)


guggero:
I added an updated version of https://github.com/btcsuite/btcd/pull/1780 as a commit to this PR as well (since the `btcutil/go.sum`) needed to be updated.

Roasbeef:
> After merging this PR we need to push two tags: chaincfg/chainhash/v1.0.0 and btcec/v2/v2.1.1.
> And I think the two tags cannot be on the same commit (because go mod is just mean that way

Weird, I think we've done this in the past? Though maybe there's some other sneaky interaction here.

btcec/go.mod
```diff
@@ -3,13 +3,9 @@ module github.com/btcsuite/btcd/btcec/v2
 go 1.17
 
 require (
-	github.com/btcsuite/btcd v0.22.0-beta.0.20220111032746-97732e52810c
+	github.com/btcsuite/btcd/chaincfg/chainhash v1.0.0
```
Roasbeef, line 6: 👍

Roasbeef: FWIW, we only have a spew dep due to these two lines in the tests: https://github.com/btcsuite/btcd/blob/master/btcec/pubkey_test.go#L244-L245

Roasbeef (approved):
LGTM 🦟

Changes look good to me, if we _really_ want to then we can eliminate the `spew` dep, but that only relies on the std-lib, so it doesn't bring any new deps into the fray.

Roasbeef:
Needs aa rebase, then we can merge this and push out the new tags, then update the mod files in another PR.

guggero:
Rebased!

guggero:
Again, I think it is very important to push the tags for the correct commits! Otherwise we might point to a version that has a replace directive, creating more pain points.

Here are the updated commit hashes to tag:
 - `chaincfg/chainhash/v1.0.0`: 7cc824e9b5768375d8d0770258f35bd67845698e
 -  `btcec/v2/v2.1.1` 4ad74cd4c261ab6cbb81b9ee441fc5b0e83be58b

btcec/go.mod
guggero: I think it's okay to have that `spew` dependency. Anything that's not in this repo isn't really a problem as a dependency.

PR #1824, opened 2022-03-10
build: update to btcec/v2.1.1 and chaincfg/chainhash/v1.00

Roasbeef:


Roasbeef:
Given the change in the module files, I think we need to tag _another_ version for the sub-modules with the more minimal `go.sum` files?

guggero (approved):
LGTM :taco: 

> Given the change in the module files, I think we need to tag another version for the sub-modules with the more minimal go.sum files?

Yes, I think that makes sense.

Roasbeef:
Tags pushed: https://github.com/btcsuite/btcd/tags

Issue #1623, opened 2020-08-31
Auto-build Docker images on CI


onyb:
[...] would be nice to include a docker image in future releases, [...]

_Originally posted by @jakesyl in https://github.com/btcsuite/btcd/pull/1465#issuecomment-683997565_

### Definition of done:
* ~When a PR is merged to `master` => build and publish to `latest` tag.~
* When a new release is cut, or a tag is created => build and publish to `v0.21.0` (say) tag.
* The images are published to the GitHub packages Docker repository.
  * This is to avoid creating an account on DockerHub.

jcvernaleo:
Is there really a good reason to provide images for the `latest` tag?  I love the idea of auto publishing releases but while we do not ever want to break `master`, I'm not convinced that should be run by someone who doesn't go and build it.

jcvernaleo:
And for the record, I might feel differently for a language that is harder to build than go but with go, the barrier to building is so low...

onyb:
> Is there really a good reason to provide images for the latest tag?

I just mentioned it because `latest` is the default tag in Docker. So it could definitely point to the most recent release - and I think you're right that `master` should not really be considered stable.

Regarding why have Docker at all, it's just a ubiquitous way to run stuff. For someone who's not familiar with Go (or doesn't want to install the compiler), Docker is a nice way to quickly test-drive btcd.

jcvernaleo:
The `latest` tag in docker has caused me a whole lot of trouble in the past.  I've basically moved to avoid using it completely and only use explicitly tagged things.

jakesyl:
@jcvernaleo has a good point here. At the very least I'd like to be able to validate we're not breaking docker builds as we go. I had a big issue with `libc6-compat` not being installed in alpine

jcvernaleo:
One of the reasons I never use alpine (linux, I use the alpine mail client happily) and do not suggest anyone else use it.

torkelrogstad:
I agree that `latest` is a foot-gun. You don't know what you're getting, but even worse, you're getting different images with different pulls! I use Docker to deploy BTC/LN infrastructure, and I like that my versions are stable. I use tagged images, so this won't affect me, but I can imagine a lot of users running into unforeseen issues with `latest`. Best not to publish it IMO, and only do tagged releases.

onyb:
Struck out the line about `latest` tag in the issue description. Thanks all for your inputs.

moesy:
[...]

PR #1830, opened 2022-03-15
Adds GH workflow to build docker images on version tags

vpereira01:
Adds a GitHub workflow, and custom docker file, using docker buildx which builds docker container images for different CPU architectures and publishes these images to GitHub packages. 

Workflow will automatically run when a version tag is created, pattern `v*`, example `v0.23.0-beta`. Also supports manual workflow trigger to allow the flexibility to build images without needing to create a new version tag.

Fixes: #1623 

Example of how packages are listed from my fork https://github.com/vpereira01/btcd/pkgs/container/btcd

Few notes:
* The step, `Extract metadata (tags, labels) for Docker`, will generate the tag **latest** when creating a tag `v0.99` on master. Example from my fork: https://github.com/vpereira01/btcd/runs/5561633723?check_suite_focus=true . Not sure if this should be kept given the comments on #1632  
* I think the packages must be made public. GitHub seems to default packages to be private so anonymous docker pulls will not be allowed.
* I have not used image digests on Dockerfile given the digest varies according to the TARGETARCH.
* The platforms `linux/amd64,linux/arm64,linux/arm,linux/386` seems good enough to cover must of uses cases. To support specific GOARM a lot rework is required.

References:
* https://www.docker.com/blog/docker-v2-github-action-is-now-ga/
* https://docs.github.com/en/actions/publishing-packages/publishing-docker-images#publishing-images-to-github-packages
* Possible optimizations https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/


.github/workflows/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# GitHub action dockerfile
+# Requires docker experimental features as buildx and BuildKit so not suitable for developers regular use.
+# https://docs.docker.com/develop/develop-images/build_enhancements/#to-enable-buildkit-builds
+
+###########################
+# Build binaries stage
+###########################
+FROM --platform=$BUILDPLATFORM golang:1.16.15-alpine3.15 AS build
```
jcvernaleo, line 8: I think go 1.16 is no longer supported.  Really should use 17 (or 18).

.github/workflows/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+###########################
+# Build docker image stage
+###########################
+FROM alpine:3.15
+COPY --from=build /go/bin /bin
+# 8333  Mainnet Bitcoin peer-to-peer port
+# 8334  Mainet RPC port
+EXPOSE 8333 8334
+ENTRYPOINT ["btcd"]
```
jcvernaleo, line 24: Files really should end in a newline.

.github/workflows/dimagespub.yml
```diff
@@ -0,0 +1,52 @@
...
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+       
+      - name: Build and push Docker images
+        uses: docker/build-push-action@ac9327eae2b366085ac7f6a2d02df8aa8ead720a
+        with:
+          file: .github/workflows/Dockerfile
+          labels: ${{ steps.meta.outputs.labels }}
+          platforms: ${{ env.TPLATFORMS }}
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
```
jcvernaleo, line 52: Really should end in a newline.

jcvernaleo (changes requested):
Three super minor comments.

.github/workflows/Dockerfile
vpereira01: Done.

jcvernaleo (approved):
Looks good now.

OK

PR #1831, opened 2022-03-16
build: update to btcec/v2.1.3 and chaincfg/chainhash v1.0.1

Roasbeef:
In this commit, we update the top level module to use the newly tagged
sub-modules. Once we remove the circular dependant in these sub-modules,
then we'll no longer have to do things like this.


guggero (approved):
LGTM :tada:

PR #1834, opened 2022-03-28
psbt: allow empty bip32 derivation path

guggero:
Fixes https://github.com/lightningnetwork/lnd/issues/6372.

BIP-0174 defines the derivation path being encoded as
  "<32-bit uint> <32-bit uint>*"
with the asterisk meaning 0 to n times. Which in turn means that an
empty path is valid, only the key fingerprint is mandatory.

cc @Roasbeef @ellemouton


Roasbeef:
I pushed the `btcutil/psbt/v1.1.2` tag.

Issue #1711, opened 2021-04-16
RPCS: Failed to marshal reply: rpcversion '' is invalid


regata:
Seeing this error when calling `getinfo` using https://github.com/petertodd/python-bitcoinlib:

```
RPCS: Failed to marshal reply: rpcversion '' is invalid
```

I'm using the master@2d7825c. I believe it happens because btcd expects `jsonrpc` field to be provided

https://github.com/btcsuite/btcd/blob/2d7825cf709fc6ac15921cecfd4d62ec78ccbba2/btcjson/jsonrpc.go#L96-L101

when processing RPC requests:

https://github.com/btcsuite/btcd/blob/2d7825cf709fc6ac15921cecfd4d62ec78ccbba2/rpcserver.go#L4152-L4155

However, bitcoinlib does not send it
https://github.com/petertodd/python-bitcoinlib/blob/0ea83c4fd499c63823754627c69ebdd79dc0d62c/bitcoin/rpc.py#L217-L220

What is the "ground truth" for the rpc spec? is the field expected to be set by clients?

onyb:
I think this is a regression due to the recently merged JSON-RPC Batch implementation. cc: @jakesyl

We should try to include a fix in 0.22.0 (#1706) - should be quite simple to implement from the look of it. @jcvernaleo

jcvernaleo:
I was under the impression that the rpcversion was supposed to be there.  @jakesyl care to look at this?

Sjors:
I'm seeing this error too with 0.22 beta...

nikicat:
I still getting this error, but it's easy to fix with this commit https://github.com/nikicat/btcd/commit/eaac8c96f8fb6bf2944a12015f967f0f483b617b

PR #1837, opened 2022-03-31
Default to JSONRPC 1.0 if the rpcVersion empty

jsfour:
This resolves https://github.com/btcsuite/btcd/issues/1711. The issue causes problems with downstream libraries if the `jsonrpc` parameter isn't included with an rpc request.

Oddly enough the comment on line 159 calls for defaulting to `RpcVersion1` but the code does not do it.

I just added in a quick check and default in the case of an empty version.


jcvernaleo (approved):
OK

PR #1838, opened 2022-04-02
chaincfg: use lower custom activation threshold for regtest+simnet

Roasbeef:
The existing values were copied over from the testnet deployment, which
uses a much larger miner confirmation window. As a result, the main
taproot deployment would require thousands of blocks to properly
activate in development environments.


guggero (approved):

jcvernaleo (approved):
OK

PR #1840, opened 2022-04-11
mempool: export checkStandardness func

ellemouton:
Export checkStandardness func so that it can be used elsewhere

### DIFF
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -15,12 +15,12 @@ import (
 	"github.com/btcsuite/btcd/blockchain"
 	"github.com/btcsuite/btcd/blockchain/indexers"
 	"github.com/btcsuite/btcd/btcjson"
+	"github.com/btcsuite/btcd/btcutil"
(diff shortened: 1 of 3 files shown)


Roasbeef (approved):
LGTM 🍁

It might also be useful to eventually export some more mempool components as well? So enough to be able to give it a mock UTXO set and apply the normal contextual validation. This might already be possible however. I think maybe we should defer merging this till the `lnd` PR using the new functionality gets further along so we can consider everything as a whole? No strong feelings in this direction fwiw.

ellemouton:
>  I think maybe we should defer merging this till the lnd PR using the new functionality gets further along so we can consider everything as a whole?

yeah agreed. Just wanted to open this one up so long so that the go.mod in the LND pr is pointing to something visable

Roasbeef:
We're seeing more instances of this in `lnd` when users end up creating mega transactions, so I think it's worth landing this, then circling back to see if we can add more contextual validation.

jcvernaleo (approved):
OK

Issue #1841, opened 2022-04-12
Cannot call FundPSBT with a psbt with only a zero value OP_RETURN output


benthecarman:
When creating trying to fund a psbt that creates an OP_RETURN with 0 value it gives the error `PSBT packet must contain at least one input or output`. If I change the output to be 1 satoshi, it will work.

example psbt: `cHNidP8BACABAAAAAAEAAAAAAAAAAA1qC2hlbGxvIHdvcmxkAAAAAAAA`

Seems to be related to btcd expecting witness serialization psbts

more info here:
https://github.com/lightningnetwork/lnd/issues/6386

PR #1842, opened 2022-04-12
psbt: always use non witness serialization format

guggero:
Fixes https://github.com/btcsuite/btcd/issues/1841.

BIP-0174 states that the transaction must be in the old serialization
format (without witnesses).

https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#specification (see first entry in table).

cc @Roasbeef, @positiveblue

Fixes https://github.com/lightningnetwork/lnd/issues/6386

### DIFF
--- a/btcutil/psbt/psbt.go
+++ b/btcutil/psbt/psbt.go
@@ -214,22 +214,12 @@ func NewFromRawBytes(r io.Reader, b64 bool) (*Packet, error) {
 		return nil, err
 	}
 	msgTx := wire.NewMsgTx(2)
-	err = msgTx.Deserialize(bytes.NewReader(value))
-	if err != nil {
-		// If there are no inputs in this yet incomplete transaction,
-		// the wire package still incorrectly assumes it's encoded in
-		// the witness format. We can fix this by just trying the non-
-		// witness encoding too. If that also fails, it's probably an
-		// invalid transaction.
-		msgTx = wire.NewMsgTx(2)
-		err2 := msgTx.DeserializeNoWitness(bytes.NewReader(value))
 
-		// If the second attempt also failed, something else is wrong
-		// and it probably makes more sense to return the original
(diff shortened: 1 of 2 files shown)


btcutil/psbt/psbt.go
```diff
@@ -214,22 +214,12 @@ func NewFromRawBytes(r io.Reader, b64 bool) (*Packet, error) {
 		return nil, err
 	}
 	msgTx := wire.NewMsgTx(2)
-	err = msgTx.Deserialize(bytes.NewReader(value))
+
+	// BIP-0174 states: "The transaction must be in the old serialization
+	// format (without witnesses)."
+	err = msgTx.DeserializeNoWitness(bytes.NewReader(value))
```
Roasbeef, line 220: Should we continue to try to decode the witness version and fall back to non-witness? Since it's possible there's old software out there encoding wit the witness version, we'd potentially break compatibility across lnd version for even btcutil versions with this change as is. 

It seems the most important thing here is for us to _encode_ using the new witness version, AFAICT. So if we just serialize w/o the witness, then things are compatible across the prior versions.

Roasbeef (changes requested):

btcutil/psbt/psbt.go
guggero: I can add the fallback. Though I don't think this change will create a compatibility issue. A PSBT packet with an unsigned TX that has a witness set cannot be serialized, the library will complain. And if you call `msgTx.Serialize()` with a transaction that has no witness, the old format will be used. Therefore the change to the serialization is purely to be more explicit about it. The actual important change is that we always _read_ the TX expecting it to be in the non-witness format.

Roasbeef (approved):
LGTM 🛰

Roasbeef:
The tag `btcutil/psbt/v1.1.3` has been pushed.

Issue #1843, opened 2022-04-13
Using rpcclient GetNewAddress shows -32600: Invalid request


leoyzx:
my version of btcd/btcwallet:
btcd version 0.22.0-beta
btcwallet version 0.7.0-alpha
After running`btcd --simnet --rpcuser=rpcuser --rpcpass=rpcpass --rpclisten=0.0.0.0:18556 --listen=0.0.0.0:18555`&`btcwallet --simnet --username=rpcuser --password=rpcpass` the btcwallet RPC server listening on 127.0.0.1:18554
after connecting to btcw rpcserver with the config
`connCfg := &rpcclient.ConnConfig{
		Host:         "localhost:18554",
		Endpoint:     "ws",
		User:         "rpcuser",
		Pass:         "rpcpass",
		Params: 	"simnet",
		Certificates: certs,
	}`
using the api `address,err:=client.GetNewAddress("default","legacy")` the err shows that`-32600: Invalid request`.
Does anyone know the possible solution of this situation?
PS: i have tried `ListAccounts`&`GetAddressesByAccount`,they can return the correct data.

chappjc:
You're using the `rpcclient` package from `master`, but the btcwallet from long ago.

In https://github.com/btcsuite/btcd/commit/061aef98af1321d2cad437afcb7dcbd562a5b3b6, the `getnewaddress` RPC was given the addess type argument, but corresponding btcwallet change is not merged yet https://github.com/btcsuite/btcwallet/pull/783

Just use the rpcclient package from the 0.22.0-beta tag.  You should probably also update your btcwallet from 0.7.0 to latest.

leoyzx:
i have update my btcwallet from 0.7.0 to 0.14.0(i didn't modify the code).
Using the new btcw server certificates&connecting to it.
In the code `address,err =client.GetNewAddress("test","legacy")`.
err still returns `-32600: Invalid request`.

leoyzx:
> You're using the `rpcclient` package from `master`, but the btcwallet from long ago.
> 
> In [061aef9](https://github.com/btcsuite/btcd/commit/061aef98af1321d2cad437afcb7dcbd562a5b3b6), the `getnewaddress` RPC was given the addess type argument, but corresponding btcwallet change is not merged yet [btcsuite/btcwallet#783](https://github.com/btcsuite/btcwallet/pull/783)
> 
[...]

PR #1844, opened 2022-04-14
rpcclient: add getnewaddresstype and revert breaking change

chappjc:
Since the address type was [added to the btcjson and rpcclient packages](https://github.com/btcsuite/btcd/commit/061aef98af1321d2cad437afcb7dcbd562a5b3b6), but the [corresponding btcwallet PR](https://github.com/btcsuite/btcwallet/pull/783) is not yet merged, it made it clear that the rpcclient would ideally be able to work with older btcwallet versions.  That is not straightforward with the recently modified `rpcclient.Client` methods.  See https://github.com/btcsuite/btcd/issues/1843.  This change seeks to make the rpcclient's `GetNewAddress` method compatible with both old and new btcwallet apps.

This reverts the previous breaking change to the `GetNewAddress` and `GetRawChangeAddress` `rpcclient.Client` methods, and adds the methods `GetNewAddressType` and `GetRawChangeAddressType` for requesting an address of a certain type.  This change allows the rpcclient package to continue to work with versions of the btcwallet app that do not recognize the address type parameter.

An alternate change that does not add new methods but lets the type param be omitted is here: https://github.com/btcsuite/btcd/issues/1843#issuecomment-1098044861 That keeps the current API with the breaking change however.

### DIFF
--- a/btcjson/walletsvrcmds_test.go
+++ b/btcjson/walletsvrcmds_test.go
@@ -388,7 +388,21 @@ func TestWalletSvrCmds(t *testing.T) {
 			},
 		},
 		{
-			name: "getnewaddress optional",
+			name: "getnewaddress optional acct",
+			newCmd: func() (interface{}, error) {
+				return btcjson.NewCmd("getnewaddress", "acct")
+			},
+			staticCmd: func() interface{} {
+				return btcjson.NewGetNewAddressCmd(btcjson.String("acct"), nil)
+			},
+			marshalled: `{"jsonrpc":"1.0","method":"getnewaddress","params":["acct"],"id":1}`,
+			unmarshalled: &btcjson.GetNewAddressCmd{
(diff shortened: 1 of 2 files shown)


chappjc:
To summarize, this is primarily to allow latest rpcclient package to work with older btcwallet apps.

Could also achieve that with 

```go
func (c *Client) GetNewAddressAsync(account, addrType string) FutureGetNewAddressResult {
	aType := &addrType
	if addrType == "" {
		aType = nil
	}
	cmd := btcjson.NewGetNewAddressCmd(&account, aType)
```

Entirely depends on design goals for the btcwallet packages.

Roasbeef (approved):
LGTM 🦴

chappjc:
Swinging back to https://github.com/btcsuite/btcwallet/pull/783 now

PR #1847, opened 2022-04-23
psbt: add support for new Taproot fields

guggero:
Adds new Taproot specific fields to PSBT as defined in https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki.

Test vectors taken from https://github.com/bitcoin/bitcoin/pull/22558.


btcutil/psbt/taproot.go
```diff
@@ -0,0 +1,219 @@
...
+// checkValid checks that both the pubkey and the signature are valid.
+func (s *TaprootScriptSpendSig) checkValid() bool {
+	return validateXOnlyPubkey(s.XOnlyPubKey) &&
+		validateSchnorrSignature(s.Signature)
+}
+
+// EqualKey returns true if this script spend signature's key data is the same
+// as the given script spend signature.
+func (s *TaprootScriptSpendSig) EqualKey(other *TaprootScriptSpendSig) bool {
+	return bytes.Equal(s.XOnlyPubKey, other.XOnlyPubKey) &&
```
Roasbeef, line 39: What about the sighash information?

btcutil/psbt/taproot.go
```diff
@@ -0,0 +1,219 @@
...
+	return validateControlBlock(s.ControlBlock)
+}
+
+// TaprootTapLeafScriptSorter implements sort.Interface for the
+// TaprootTapLeafScript struct.
+type TaprootTapLeafScriptSorter []*TaprootTapLeafScript
+
+func (s TaprootTapLeafScriptSorter) Len() int { return len(s) }
+
+func (s TaprootTapLeafScriptSorter) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
```
Roasbeef, line 75: FWIW if we want to cut down on the extra/duplicate definitions a bit, we can use the `sort.Slice` function instead: https://pkg.go.dev/sort#Slice. So the only thing we'd need specifically for each new slice type is the comparison function below.

btcutil/psbt/taproot.go
```diff
@@ -0,0 +1,219 @@
...
+		if err != nil || n != 32 {
+			return nil, ErrInvalidPsbtFormat
+		}
+	}
+
+	// Extract the remaining bytes from the reader (we don't actually know
+	// how many bytes we read due to the compact size integer at the
+	// beginning).
+	leftover := make([]byte, reader.Len())
+	_, err = reader.Read(leftover)
```
Roasbeef, line 154: I think we can do this in place: 
```go
io.Copy(io.Discard, reader)
```

btcutil/psbt/partial_input.go
```diff
@@ -13,16 +13,22 @@ import (
...
-	Bip32Derivation    []*Bip32Derivation
-	FinalScriptSig     []byte
-	FinalScriptWitness []byte
-	Unknowns           []*Unknown
+	NonWitnessUtxo         *wire.MsgTx
+	WitnessUtxo            *wire.TxOut
+	PartialSigs            []*PartialSig
+	SighashType            txscript.SigHashType
+	RedeemScript           []byte
+	WitnessScript          []byte
```
Roasbeef, line 21: Non-blocking, but we may want to consider using struct embedding to make the set of fields more preamble at a glance. Since the structs would be embed, there wouldn't be any additional layers of indirection required for callers.

btcutil/psbt/partial_input.go
```diff
@@ -328,6 +483,85 @@ func (pi *PInput) serialize(w io.Writer) error {
...
+			err := serializeKVPairWithType(
+				w, uint8(TaprootKeySpendSignatureType), nil,
+				pi.TaprootKeySpendSig,
+			)
+			if err != nil {
+				return err
+			}
+		}
+
+		sort.Sort(TaprootScriptSpendSigSorter(pi.TaprootScriptSpendSig))
```
Roasbeef, line 497: Where does the BIP actually mandate that these contents be sorted? Or is this just something we're adding to make testing easier since our serialization would be deterministic-ish?

Roasbeef (commented):
Nice work! Cool to be on the on the bleeding edge when it comes to stuff like this 🤓. Will circle back re testing between hardwarewallets/bitcoind on that `lnd` PR.

btcutil/psbt/taproot.go
guggero: The key only consists of the x-only-pubkey and the leaf hash. The signature and sighash is in the value.

btcutil/psbt/taproot.go
guggero: Yeah, makes sense. Went with an approach that adds a `SortBefore` method to the structs so the business logic can be implemented on the struct itself and not where we do the actual sort.

btcutil/psbt/taproot.go
guggero: I didn't know about `io.Discard`. But in this case we actually want to get the remaining bytes, we just don't know how many there are. I went with a `WriteTo` approach instead, hoping this makes it a bit more clear.

btcutil/psbt/partial_input.go
guggero: You mean to create a `TaprootPInput` that can be embedded to avoid having all those fields with the `Taproot...` prefix? We could do that, yes. Though we'd get a collision with the `Bip32Derivation` derivation name, not sure if that would lead to more mistakes or confusion?

guggero:
Rebased and addressed all comments.

guggero:
Unit test failure seems to be a flake... Can't reproduce locally.

Roasbeef:
Kicked things to re-run CI...

Roasbeef (approved):
LGTM 🎗

Issue #1839, opened 2022-04-05
ambiguous import: found package github.com/btcsuite/btcd/chaincfg/chainhash in multiple modules


shrimalmadhur:
Hello, 

I am using the beta version (`btcd@v0.22.0-beta`) and recently I got this error while building 

```
github.com/btcsuite/btcd/btcec tested by
github.com/btcsuite/btcd/btcec.test imports
github.com/btcsuite/btcd/chaincfg/chainhash: ambiguous import: found package github.com/btcsuite/btcd/chaincfg/chainhash in multiple modules:
github.com/btcsuite/btcd v0.22.0-beta (/Users/madhurshrimal/go/pkg/mod/github.com/btcsuite/btcd@v0.22.0-beta/chaincfg/chainhash)
github.com/btcsuite/btcd/chaincfg/chainhash v1.0.0 (/Users/madhurshrimal/go/pkg/mod/github.com/btcsuite/btcd/chaincfg/chainhash@v1.0.0)
```

It didn't use to come before and suddenly it popped up. Any idea how I can solve this?

sarvalabs-sai:
You may try deleting `go.mod` ,`go.sum` and do `go clean --modcache` before doing `go mod init`

chappjc:
@shrimalmadhur I believe the issue is that your `make format` target does `go run golang.org/x/tools/cmd/goimports` from the with no version suffix and from the module folder.  From the `go help run` output for both go 1.17 and 1.18 (the only two Go versions that are not EOL):

```
If the package argument has a version suffix (like @latest or @v1.0.0),
"go run" builds the program in module-aware mode, ignoring the go.mod file in
the current directory or any parent directory, if there is one. This is useful
for running programs without affecting the dependencies of the main module.

If the package argument doesn't have a version suffix, "go run" may run in
module-aware mode or GOPATH mode, depending on the GO111MODULE environment
variable and the presence of a go.mod file. See 'go help modules' for details.
If module-aware mode is enabled, "go run" runs in the context of the main
module.
```

You want `go run` to leave your module's go.mod alone.  So, you can try to append `@latest`:

```
go run golang.org/x/tools/cmd/goimports@latest
```

Or you can just install `goimports` like you do with golines and goveralls.

erwanor:
I was having the same issue building a downstream script that uses `ethclient` (four or five layers removed from `btcsuite/btcd/chaincfg/chainhash`).

```
github.com/myrepo/myrepo imports
[...]

PR #1825, opened 2022-03-10
mod: remove `btcutil` circular dependency with main module

guggero:
~Depends on https://github.com/btcsuite/btcd/pull/1824.~

Removes the circular dependency between the main module and the `btcutil` sub module by introducing a bunch of new sub modules.

PR #1848, opened 2022-04-25
btcec: remove import of chainhash

chappjc:
One option to quickly resolve https://github.com/btcsuite/btcd/issues/1839

This duplicates some code, but requires no import path changes, only a new btcec/v2 tag.  For a proper solution, a chainhash/v2 module is still required (e.g. https://github.com/btcsuite/btcd/commit/39b119350096dee8c34d3aaa426c17105795caa9 plus a [large cascade of deps](https://github.com/btcsuite/btcd/compare/master...chappjc:chainhashv2?expand=1) like btcutil/v2 and all the way up)  since it's not just btcec that will introduce this ambiguous import.

```
The chainhash import is challenging for consumers of the latest tagged
btcd module (v0.22.0-beta) because that module revision has its own
chainhash package.  This creates an ambiguous import when trying
to import both the older btcd module and btcec/v2 which requires the
standalone module version of chainhash.

This is the brute solution.  An alternative would be to create a
chaincfg/chainhash/v2 module and have btcec/v2 require that instead.
```

### DIFF
--- a/btcec/ecdsa/example_test.go
+++ b/btcec/ecdsa/example_test.go
@@ -5,13 +5,19 @@
 package ecdsa_test
 
 import (
+	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcec/v2/ecdsa"
-	"github.com/btcsuite/btcd/chaincfg/chainhash"
 )
+
+func doubleHash(b []byte) []byte {
+	first := sha256.Sum256(b)
+	second := sha256.Sum256(first[:])
+	return second[:]
+}
 
 // This example demonstrates signing a message with a secp256k1 private key that
 // is first parsed form raw bytes and serializing the generated signature.
@@ -27,7 +33,7 @@ func Example_signMessage() {
 
 	// Sign a message using the private key.
 	message := "test message"
-	messageHash := chainhash.DoubleHashB([]byte(message))
+	messageHash := doubleHash([]byte(message))
 	signature := ecdsa.Sign(privKey, messageHash)
 
 	// Serialize and display the signature.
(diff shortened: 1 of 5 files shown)


jcvernaleo:
I'm comfortable with this solution but would love to hear from @Roasbeef before approving/merging.

chappjc:
Yah, the v2 option seemed simpler but it has cascading effects. e.g. btcutil has to go to v2 since chainhash types are in it's exported API.

The more I think about this and see how it has been playing out for consumers, the more I think the "right" solution for the current btcd dev cycle (before your next product release) is to get to btcd/v2 (module, not product) right away, before all other module activities, because countless projects import the btcd(/v0) module.  Then get btcd/v2 on a bunch of new v2 submodules.  Otherwise, the legacy btcd module will become more and more problematic, especially if any new submodule imports it or anything ambiguously from it, such as chainhash.  https://github.com/btcsuite/btcd/pull/1825#issuecomment-1105911409

Whether we like it or not, with the invent of Go modules, the entire ecosystem is essentially *forced* to bump their semantic import version (the /vX on the import path) if there are any breaking changes.  While more modules => more burden for btcsuite, carving out sub-modules can at least reduce the frequency of such bumps.  It's a balance that's hard to get right.  (And it requires far to much thought to get right IMO, but that's the state of things)

Roasbeef:
I'm on board, but think we should wait for this to be merged first: https://github.com/btcsuite/btcd/pull/1820

It defines some new tagged hash combinations, and just needs another pass on my end up update the nonce derivation to the latest musig2 BIP. I don't really foresee many other standardized "tagged hash" prefixes in the near future, and needing to add to a new map element in two areas (optional since they can pass a custom tag in any case) isn't too bad.

Roasbeef:
If we're going along with https://github.com/btcsuite/btcd/pull/1850 then I think it makes sense to actually merge this _first_, so the new slimmer module retains the old Go version requirement.

chappjc:
Converting to draft as I'm actually fairly hopeful about the new resolution proposed in the comment https://github.com/btcsuite/btcd/issues/1839#issuecomment-1111642262. Happy to repoen this though if it seems desirable regardless.

chappjc:
This is fortunately no longer required. 🥳

PR #1850, opened 2022-04-27
build: switch to using Go 1.18 for CI in go.mod

Roasbeef:
In this commit, we switch to using Go 1.18 everywhere. This is a
deviation from our typical "last 2 versions" rule. Switching now gives
us access to the new type params feature immediately, versus near the
end of this year. With this change we'd also be able to start using the
new std lib fuzzing features as well.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -5,7 +5,7 @@ env:
   # go needs absolute directories, using the $HOME variable doesn't work here.
   GOCACHE: /home/runner/work/go/pkg/build
   GOPATH: /home/runner/work/go
-  GO_VERSION: 1.17.5
+  GO_VERSION: "1.18.x"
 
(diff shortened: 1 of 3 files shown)


guggero:
This needs a change in the Makefile as well, since `go get` is deprecated in `go 1.18` and needs to be replaced with `go install`.

Roasbeef:
So decided to hold off on this for a bit, influenced by comments from @chappjc. I updated that usage in the musig2 PR to be as compact at the call site as it would be w/ generics, but at the cost of some added function definitions.

Roasbeef:
Holding off on this for now, as we need to decide on a process for deciding when to update the Go version pinned for `btcd`, many downstream projects depend on it.

jcvernaleo:
This is super out of date so closing.

PR #1851, opened 2022-04-28
[release-v0.22.1] multi: Switch to chaincfg/chainhash module.

### LINKED ISSUE #1839: ambiguous import: found package github.com/btcsuite/btcd/chaincfg/chainhash in multiple modules

davecgh:
Currently, there are some issues for consumers importing the main `github.com/btcsuite/btcd@v0.22.0-beta` module:

* Anyone attempting to import anything `github.com/btcsuite/btcd` end up with an old and broken version `v0.18.0` since all of the later tags are pre-release versions and the go tooling prefers release tags over pre-release tags
* Any consumers that import `github.com/btcsuite/btcd@v0.22.0-beta` along with the new `github.com/btcsuite/btcd/btcec/v2` or `github.com/btcsuite/btcd/chaincfg/chainhash` modules will have conflicts due to the existence of both the package and the module at the same path
* The [documentation](https://pkg.go.dev/github.com/btcsuite/btcd) shows an ancient version and improperly claims it is bchd

These all boil down to a couple of things:

* As can be seen in the [module release workflow docs](https://go.dev/doc/modules/release-workflow), versions without pre-release components are preferred over those with them. In other words, `v0.22.0-beta` is seen as **before** `v0.18.1`, so currently anyone just importing or doing a plain `go get github.com/btcsuite/btcd` is going to attempt to get `v0.18.1` which is a broken version.
* The latest pre-release tag `v0.22.0-beta` has `chaincfg/chainhash` as a **package**, but there is also now a separate **module** with the same path.

This consists of the following changes which will resolve all of the aforementioned issues:

* Branches from the `v0.22.0-beta` tag so that no new code that is not ready to be released is introduced
* Removes the `chaincfg/chainhash` **package**
  * NOTE: The *package* is removed in this side branch only, not the master branch which still has the module code as expected
* Updates the `go.mod` to use the ` chaincfg/chainhash` **module** instead
* Bumps the version  of btcd and btcctl to v0.22.1
* Updates the CHANGES for the release

Once this is merged, all that will be needed is to create a `v0.22.1` tag at the `Update CHANGES file for 0.22.1 release.` commit.

Closes #1839.


Roasbeef:
+1 for branching off of 0.22, there're quite a lot of core changes in master due to the big taproot push we went through earlier this year. 

In the PR body:
> Bumps the version of dcrd and dcrctl to v0.22.1

Should be btcd + btcctl ;)

cmd/btcctl/version.go
```diff
@@ -18,11 +18,11 @@ const semanticAlphabet = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqr
 const (
 	appMajor uint = 0
 	appMinor uint = 22
-	appPatch uint = 0
+	appPatch uint = 1
 
 	// appPreRelease MUST only contain characters from semanticAlphabet
 	// per the semantic versioning spec.
-	appPreRelease = "beta"
+	appPreRelease = ""
```
Roasbeef, line 25: So `btcd` is now _officially_ out of beta? 🤩

davecgh:
> In the PR body:
> 
> > Bumps the version of dcrd and dcrctl to v0.22.1
> 
> Should be btcd + btcctl ;)

Oops.  Finger memory...  Updated.

Roasbeef (approved):
LGTM 🥨

chaincfg/chainhash/hash.go
```diff
@@ -1,128 +0,0 @@
-// Copyright (c) 2013-2016 The btcsuite developers
```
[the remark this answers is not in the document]
davecgh: Added a note.

cmd/btcctl/version.go
davecgh: Something I highly suggest is to separate the *product* version (aka the release versions of btcd) from the *go module* versions.  I know it might be a little annoying, but Go hijacked semver for its modules and, as things like this show, module versions really need to be independent.

The model I use for dcrd that has worked out really well is for each release to create a `release-vX.Y` branch and on that branch update the version info to remove the pre-release details, remove any overrides from `go.mod`, etc.  Then, bump the to the next pre-release version on the master branch where development continues as normal.

That way all releases have their own branch that you can back port to if it is necessary and it is decoupled from the go module version(s).

Roasbeef: > Something I highly suggest is to separate the product version (aka the release versions of btcd) from the go module versions. 

Yeah I'm slowly coming around to this realization myself. Takes a bit more work, but worth it not to cause headaches for consumers of the modules.

PR #1852, opened 2022-04-30
btcutil: add DecodeAddressForNet

### LINKED ISSUE #1846: btcutil: consider adding a new method that verifies address is for the correct net
In its current form, `btcutil.DecodeAddress` has some slightly surprising behavior: 

```go
addr, err := btcutil.DecodeAddress("bc1q7cyrfmck2ffu2ud3rn5l5a8yv6f0chkp0zpemf", &chaincfg.TestNet3Params)
if err != nil {
	log.Fatal(err)
}
log.Println("address: ", addr.String())
```

This doesn't error, but instead decoded just fine. It's explained in the godoc that the net is only used for legacy addresses. However, it looks like this is often overlooked. One example would be the instances of this in `lnd` found in this PR: https://github.com/lightningnetwork/lnd/pull/6448

Would it be an idea to add a new method that decodes the address, and then verifies it is for the given network? Names could for example be `DecodeAddressVerify`, or `DecodeAddressForNet`.

torkelrogstad:
DecodeAddressForNet verifies that the given address is actually for the
passed network.

Closes #1846


torkelrogstad:
Ping @guggero :)

btcutil/address.go
```diff
@@ -52,6 +52,10 @@ var (
 	// than assuming or defaulting to one or the other, this error is
 	// returned and the caller must decide how to decode the address.
 	ErrAddressCollision = errors.New("address collision")
+
+	// ErrIncorrect descirbes an error where an address could be decoded, but
```
guggero, line 56: nit: typo in "describes".

btcutil/address.go
```diff
@@ -223,6 +227,26 @@ func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
 	}
 }
 
+// DecodeAddressForNet decodes the string encoding of an address and returns
+// the Address if addr is a valid encoding for a known address type and the
+// given network.
+//
+// This method differs from DecodeAddress in that DecodeAddress tolerates
+// differing networks, in case the network parameter isn't needed to decode
+// the address
```
guggero, line 236: nit: add full stop.

btcutil/address_test.go
```diff
@@ -879,6 +879,68 @@ func TestAddresses(t *testing.T) {
...
+			}
+
+			// verify we can't decode for other nets
+			for _, net := range nets {
+				if net.Net == test.net.Net {
+					continue
+				}
+
+				decoded, err := btcutil.DecodeAddressForNet(test.addr, &net)
+				if err == nil {
```
guggero, line 907: nit: `if err != nil { ... continue ... }` to avoid one level of indentation?

btcutil/address_test.go
```diff
@@ -879,6 +879,68 @@ func TestAddresses(t *testing.T) {
 			continue
 		}
 
+		// Ensure addresses are only valid for their given network
+		// The cases where test.addr and test.encoded are different are
+		// pay-to-pubkey, and these are valid for all nets.
+		if test.valid && test.addr == test.encoded {
```
guggero, line 885: Nice test!

guggero (approved):
Thanks, LGTM, just a few minor nits :tada:

torkelrogstad:
Thanks for the review @guggero. Addressed your comments.

guggero:
@Roasbeef, PTAL.

btcutil/address_test.go
```diff
@@ -879,6 +879,67 @@ func TestAddresses(t *testing.T) {
 			continue
 		}
 
+		// Ensure addresses are only valid for their given network
```
Roasbeef, line 882: IMO rather than stuff this into the existing test (for the _other_ decoding function), we should make an isolated unit test to exercise the new function. Even just reviewing this fragment, it isn't very easy to follow what exactly is being tested here. A new TDD test would make things a lot clearer and more explicit.

Roasbeef (changes requested):

PR #1856, opened 2022-05-09
rpcclient: fix crash in http retry handler

bhandras:
This PR fixes a crash that happens every time when the post errors max retries times.

### DIFF
--- a/rpcclient/infrastructure.go
+++ b/rpcclient/infrastructure.go
@@ -768,11 +768,14 @@ func (c *Client) handleSendPostMessage(jReq *jsonRequest) {
 	}
 	url := protocol + "://" + c.config.Host
 
-	var err error
-	var backoff time.Duration
-	var httpResponse *http.Response
+	var (
+		err          error
+		backoff      time.Duration
+		httpResponse *http.Response
+	)
+
 	tries := 10
-	for i := 0; tries == 0 || i < tries; i++ {
+	for i := 0; i < tries; i++ {
 		bodyReader := bytes.NewReader(jReq.marshalledJSON)
 		httpReq, err := http.NewRequest("POST", url, bodyReader)
 		if err != nil {
@@ -799,7 +802,9 @@ func (c *Client) handleSendPostMessage(jReq *jsonRequest) {
 			if backoff > time.Minute {
 				backoff = time.Minute
 			}
-			log.Debugf("Failed command [%s] with id %d attempt %d. Retrying in %v... \n", jReq.method, jReq.id, i, backoff)
+			log.Debugf("Failed command [%s] with id %d attempt %d."+
+				" Retrying in %v... \n", jReq.method, jReq.id,
+				i, backoff)
 			time.Sleep(backoff)
 			continue
 		}
--- a/rpcclient/infrastructure.go
+++ b/rpcclient/infrastructure.go
@@ -761,7 +761,9 @@ out:
 // handleSendPostMessage handles performing the passed HTTP request, reading the
 // result, unmarshalling it, and delivering the unmarshalled result to the
 // provided response channel.
-func (c *Client) handleSendPostMessage(jReq *jsonRequest) {
+func (c *Client) handleSendPostMessage(jReq *jsonRequest,
+	shutdown chan struct{}) {
+
 	protocol := "http"
 	if !c.config.DisableTLS {
 		protocol = "https"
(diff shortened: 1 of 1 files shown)


guggero (approved):
LGTM, nice fix :tada:

rpcclient/infrastructure.go
```diff
@@ -768,11 +768,14 @@ func (c *Client) handleSendPostMessage(jReq *jsonRequest) {
...
-	var backoff time.Duration
-	var httpResponse *http.Response
+	var (
+		err          error
+		backoff      time.Duration
+		httpResponse *http.Response
+	)
+
 	tries := 10
-	for i := 0; tries == 0 || i < tries; i++ {
```
Roasbeef, line 775: Also seems to be inserting a bug fix within this formatting fix commit? 

Unclear what the purpose of that is, given that `tries` is never decremented...

rpcclient/infrastructure.go
```diff
@@ -776,8 +776,10 @@ func (c *Client) handleSendPostMessage(jReq *jsonRequest) {
 
 	tries := 10
 	for i := 0; i < tries; i++ {
+		var httpReq *http.Request
+
 		bodyReader := bytes.NewReader(jReq.marshalledJSON)
-		httpReq, err := http.NewRequest("POST", url, bodyReader)
+		httpReq, err = http.NewRequest("POST", url, bodyReader)
```
Roasbeef, line 784: When you say the error didn't leave the scope, you mean this `err` here was somehow dropped in the channel send below? Still trying to wrap my head around the fix here...

rpcclient/infrastructure.go
```diff
@@ -799,15 +801,22 @@ func (c *Client) handleSendPostMessage(jReq *jsonRequest) {
...
 			}
 			log.Debugf("Failed command [%s] with id %d attempt %d."+
 				" Retrying in %v... \n", jReq.method, jReq.id,
 				i, backoff)
-			time.Sleep(backoff)
+
+			select {
+			case <-time.After(backoff):
+
+			case <-shutdown:
```
Roasbeef, line 816: 👍

Roasbeef:
Tried this on my local machine running while running https://github.com/lightningnetwork/lnd/pull/6345. I still get the panic error, as the response is nil, and all the retries fail since the my `bitcoind` could actually never stand up (some openssl dynlib thing messed up my bitcoind install).

rpcclient/infrastructure.go
bhandras: I think this may have been left when someone was debugging retries and they set `tries` to zero to induce infinite retries.

rpcclient/infrastructure.go
bhandras: This `:=` overrides the `err` defined at line 772, making `err` outside this scope always `nil`.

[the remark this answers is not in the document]
bhandras: Yes agree, this piece of code is a bit hard to understand. Made this more explicit in the next commit, ptal.

bhandras:
> Tried this on my local machine running while running [lightningnetwork/lnd#6345](https://github.com/lightningnetwork/lnd/pull/6345). I still get the panic error, as the response is nil, and all the retries fail since the my `bitcoind` could actually never stand up (some openssl dynlib thing messed up my bitcoind install).

In my case the `bitcoind` wasn't borked but it still crashed after some recurring error that was masked in the retry loop. To be sure we never crash, added coverage for nil result too (will return an "invalid response" error).

guggero (approved):
Nice!

rpcclient/infrastructure.go
```diff
@@ -815,6 +817,14 @@ func (c *Client) handleSendPostMessage(jReq *jsonRequest) {
 		return
 	}
 
+	// We still want to return an error if for any reason the respone
+	// remains empty.
+	if httpResponse == nil {
```
Roasbeef, line 833: What about capturing that _final_ error to give the caller a more descriptive message? So in this case we'd make a new like` httpErr` error variable, which is used here (and also checked above?).

Roasbeef (commented):
Thanks for the clarifications! Latest changes are looking a lot better, one final comment re trying to pass back more descriptive error.

rpcclient/infrastructure.go
bhandras: IUUC what you mean, then the last commit I added should address it.

bhandras:
> Thanks for the clarifications! Latest changes are looking a lot better, one final comment re trying to pass back more descriptive error.

ptal

InvestorMoney (approved):

Roasbeef (approved):
LGTM 🚵🏼‍♂️

InvestorMoney (commented):
gh pr checkout 1856

InvestorMoney (commented):
rpcclient/infrastructure.go

PR #1858, opened 2022-05-19
Add ScriptPubKeyResult address field (rpc client & server)

lindlof:
Add `address` field to `ScriptPubKeyResult`. Deprecate `addresses` field.

This conforms to the corresponding change in Bitcoin Core:

- https://bitcoincore.org/en/doc/22.0.0/rpc/rawtransactions/getrawtransaction/
- https://bitcoincore.org/en/doc/22.0.0/rpc/rawtransactions/decoderawtransaction/
- https://bitcoincore.org/en/doc/22.0.0/rpc/blockchain/gettxout/

In Bitcoin Core a configuration flag is required to keep serving `addresses`. This PR serves both for the time being. The reason to require a configuration change is likely to incentivise adoption of the new field. I don't see that being as necessary for `btcd`.

### DIFF
--- a/btcjson/chainsvrresults.go
+++ b/btcjson/chainsvrresults.go
@@ -431,7 +431,8 @@ type ScriptPubKeyResult struct {
 	Hex       string   `json:"hex,omitempty"`
 	ReqSigs   int32    `json:"reqSigs,omitempty"`
 	Type      string   `json:"type"`
-	Addresses []string `json:"addresses,omitempty"`
+	Address   *string  `json:"address,omitempty"`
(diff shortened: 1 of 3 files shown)


gitteri:
Any update on this? would be great to merge this in

qhxcWallet (approved):
The changes here apply to the v22 version of Bitcoin core

qhxcWallet (approved):

kgpark1399 (approved):

lindlof:
Better to use https://github.com/btcsuite/btcd/pull/1969 because with this PR a multisig script might look like a normal address in RPC response

PR #1865, opened 2022-06-24
MuSig2: Catch up to 0.4.0

sputn1ck:
This PR fixes nonce generation to include XORing the passed secret key with the random bytes. (see: https://github.com/jonasnick/bips/blob/musig2/bip-musig2.mediawiki#nonce-generation-1)

Additionally I updated the nonce generation to 0.3.0 and added the testcases
(https://github.com/jonasnick/bips/blob/musig2/bip-musig2.mediawiki#Change_Log)
(https://github.com/jonasnick/bips/commit/f4216927a6daf796e36a53b6d2c8bc2782016d21)

I'm currently in the progress of catching up to 0.4.0
- [x] 0.3.0: Hash i - 1 instead of i in NonceGen
- [x] 0.3.1: Add NonceGen test vectors
- [x] 0.3.2: Add a lot of test vectors and improve handling of invalid contributions in reference code.
  - [x] Aggregate Keys testvectors
  - [x] Aggregate Nonces testvectores
  - [x] Sign testvectors
  - [x] Partial Sign testvectors
  - [x] Sig Agg testvectors
- [x] 0.4.0: Allow the output of NonceAgg to be infinity and add test vectors

### DIFF
--- a/btcec/btcec.go
+++ b/btcec/btcec.go
@@ -39,3 +39,18 @@ type CurveParams = secp.CurveParams
 func Params() *CurveParams {
 	return secp.Params()
 }
+
+// Generator returns the public key at the Generator Point.
+func Generator() *PublicKey {
+	var (
+		result JacobianPoint
+		k      secp.ModNScalar
+	)
+
+	k.SetInt(1)
+	ScalarBaseMultNonConst(&k, &result)
+
+	result.ToAffine()
+
+	return NewPublicKey(&result.X, &result.Y)
+}
--- a/btcec/curve.go
+++ b/btcec/curve.go
@@ -4,12 +4,17 @@
 package btcec
 
 import (
+	"fmt"
+
 	secp "github.com/decred/dcrd/dcrec/secp256k1/v4"
 )
 
 // JacobianPoint is an element of the group formed by the secp256k1 curve in
 // Jacobian projective coordinates and thus represents a point on the curve.
 type JacobianPoint = secp.JacobianPoint
+
+// infinityPoint is the jacobian representation of the point at infinity.
+var infinityPoint JacobianPoint
 
 // MakeJacobianPoint returns a Jacobian point with the provided X, Y, and Z
 // coordinates.
@@ -60,4 +65,51 @@ func ScalarBaseMultNonConst(k *ModNScalar, result *JacobianPoint) {
 // result. The resulting point will be normalized.
 func ScalarMultNonConst(k *ModNScalar, point, result *JacobianPoint) {
 	secp.ScalarMultNonConst(k, point, result)
+}
+
+// ParseJacobian parses a byte slice point as a secp.Publickey and returns the
+// pubkey as a JacobianPoint. If the nonce is a zero slice, the infinityPoint
+// is returned.
+func ParseJacobian(point []byte) (JacobianPoint, error) {
+	var result JacobianPoint
+
+	if len(point) != 33 {
+		str := fmt.Sprintf("invalid nonce: invalid length: %v",
+			len(point))
+		return JacobianPoint{}, makeError(secp.ErrPubKeyInvalidLen, str)
+	}
+
+	if point[0] == 0x00 {
+		return infinityPoint, nil
+	}
+
+	noncePk, err := secp.ParsePubKey(point)
+	if err != nil {
+		return JacobianPoint{}, err
+	}
+	noncePk.AsJacobian(&result)
+
+	return result, nil
+}
+
+// JacobianToByteSlice converts the passed JacobianPoint to a Pubkey
+// and serializes that to a byte slice. If the JacobianPoint is the infinity
+// point, a zero slice is returned.
+func JacobianToByteSlice(point JacobianPoint) []byte {
+	if point.X == infinityPoint.X && point.Y == infinityPoint.Y {
+		return make([]byte, 33)
+	}
+
+	point.ToAffine()
+
+	return NewPublicKey(
+		&point.X, &point.Y,
+	).SerializeCompressed()
+}
+
+// GeneratorJacobian sets the passed JacobianPoint to the Generator Point.
+func GeneratorJacobian(jacobian *JacobianPoint) {
+	var k ModNScalar
+	k.SetInt(1)
+	ScalarBaseMultNonConst(&k, jacobian)
 }
--- a/btcec/schnorr/musig2/keys.go
+++ b/btcec/schnorr/musig2/keys.go
@@ -26,6 +26,10 @@ var (
 	// ErrTweakedKeyIsInfinity is returned if while tweaking a key, we end
 	// up with the point at infinity.
 	ErrTweakedKeyIsInfinity = fmt.Errorf("tweaked key is infinity point")
+
+	// ErrTweakedKeyOverflows is returned if a tweaking key is larger than
+	// 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141.
+	ErrTweakedKeyOverflows = fmt.Errorf("tweaked key is to large")
 )
 
 // sortableKeys defines a type of slice of public keys that implements the sort
@@ -286,7 +290,10 @@ func tweakKey(keyJ btcec.JacobianPoint, parityAcc btcec.ModNScalar, tweak [32]by
 	// Next, map the tweak into a mod n integer so we can use it for
 	// manipulations below.
 	tweakInt := new(btcec.ModNScalar)
-	tweakInt.SetBytes(&tweak)
+	overflows := tweakInt.SetBytes(&tweak)
+	if overflows == 1 {
+		return keyJ, parityAcc, tweakAcc, ErrTweakedKeyOverflows
+	}
 
 	// Next, we'll compute: Q_i = g*Q + t*G, where g is our parityFactor and t
 	// is the tweakInt above. We'll space things out a bit to make it easier to
--- a/btcec/schnorr/musig2/musig2_test.go
+++ b/btcec/schnorr/musig2/musig2_test.go
@@ -1461,7 +1461,7 @@ func TestMusig2AggregateNoncesTestVectors(t *testing.T) {
 			},
 			expectedNonce: append(
 				append([]byte{}, expectedNonce[0:33]...),
-				getGBytes()...,
+				getInfinityBytes()...,
 			),
 		},
 	}
@@ -1827,6 +1827,10 @@ func getNegGBytes() []byte {
 	pk[0] = 0x3
 
 	return pk
+}
+
+func getInfinityBytes() []byte {
+	return make([]byte, 33)
 }
 
 func mustParseHex32(str string) [32]byte {
--- a/btcec/schnorr/musig2/musig2_test.go
+++ b/btcec/schnorr/musig2/musig2_test.go
@@ -74,6 +74,8 @@ func getInfinityTweak() KeyTweakDesc {
 const (
 	keyAggTestVectorName = "key_agg_vectors.json"
 
+	nonceAggTestVectorName = "nonce_agg_vectors.json"
+
 	signTestVectorName = "sign_vectors.json"
 )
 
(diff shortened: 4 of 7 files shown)


btcec/schnorr/musig2/musig2_test.go
```diff
@@ -82,34 +119,98 @@ func TestMuSig2KeyAggTestVectors(t *testing.T) {
...
+				strKeys   []string
+				strTweaks []jsonTweak
+				jsonError string
 			)
 			for _, keyIndex := range testCase.keyOrder {
 				keyBytes := testKeys[keyIndex]
 				pub, err := schnorr.ParsePubKey(keyBytes)
 				if err != nil {
+					if errors.Is(err, testCase.expectedError) {
+						return
```
Roasbeef, line 173: Why return vs continue here? IIUC, we want to continue to do all the cases after the first invalid one. 

Another way to write this would be:
```go
switch {
    case testCase.expectedError != nil && errors.Is(err, testCase.expectedError)
        continue:
    case err != nil:
        t.Fatalf()
}

File: btcec/schnorr/musig2/sign.go [hunk 2]
```diff
@@ -697,3 +697,12 @@ func CombineSigs(combinedNonce *btcec.PublicKey,
 
 	return schnorr.NewSignature(&nonceJ.X, &combinedSig)
 }
+
+// GetGenerator returns a public key at the Generator Point.
+func GetGenerator() *secp.PublicKey {
```
Roasbeef, line 717: How about we move this to just the top level `btcec` package? There it can be `Generator`, so the call path would be: `btcec.Generator()`.

btcec/schnorr/musig2/sign.go
```diff
@@ -697,3 +697,12 @@ func CombineSigs(combinedNonce *btcec.PublicKey,
...
+	var result btcec.JacobianPoint
```
Roasbeef, line 718: Can collapse this into a single var statement: 
```go 
var ( 
   result  
   k
)
```

btcec/schnorr/musig2/musig2_test.go
```diff
@@ -1017,3 +1143,19 @@ func to32ByteSlice(input []byte) [32]byte {
 	copy(output[:], input)
 	return output
 }
+
+func to66ByteSlice(input []byte) [66]byte {
```
Roasbeef, line 1711: Can use `PubNonceSize` here instead of `66`.

Roasbeef: It's also actually possible to just do a type cast here, but it ends up being pretty ugly, so we usually just make the copy explicit like here.

btcec/schnorr/musig2/musig2_test.go
```diff
@@ -446,17 +492,19 @@ func TestMuSig2SigningTestVectors(t *testing.T) {
...
-				keySet = append(keySet, signSetKeys[keyIndex])
-				strKeys = append(
-					strKeys, hex.EncodeToString(
-						schnorr.SerializePubKey(signSetKeys[keyIndex]),
-					),
-				)
+				keyBytes := signSetKeys[keyIndex]
+				pub, err := schnorr.ParsePubKey(keyBytes)
+
+				if err != nil {
```
Roasbeef, line 521: Similar comment here re the error handling.

btcec/schnorr/musig2/sign.go
```diff
@@ -85,7 +85,10 @@ func (p *PartialSignature) Decode(r io.Reader) error {
 		return nil
 	}
 
-	p.S.SetBytes(&sBytes)
+	overflows := p.S.SetBytes(&sBytes)
```
Roasbeef, line 88: 👍

btcec/schnorr/musig2/musig2_test.go
```diff
@@ -588,6 +598,238 @@ func TestMuSig2SigningTestVectors(t *testing.T) {
...
+				verifyPnonce2,
+				verifyPnonce3,
+			},
+			expectedError: secp256k1.ErrPubKeyNotOnCurve,
+		},
+	}
+
+	for _, testCase := range testCases {
+
+		// todo find name
```
Roasbeef, line 795: ?

btcec/schnorr/musig2/musig2_test.go
```diff
@@ -588,6 +598,238 @@ func TestMuSig2SigningTestVectors(t *testing.T) {
...
+		testName := fmt.Sprintf("%v/tweak=%v", testCase.pubnonceIndex, testCase.keyOrder)
+
+		t.Run(testName, func(t *testing.T) {
+
+			combinedNonce, err := AggregateNonces(testCase.nonces)
+			if err != nil {
+				if errors.Is(err, testCase.expectedError) {
```
Roasbeef, line 802: Similar comment re using a `switch` statement here to make the error handling more readable.

btcec/schnorr/musig2/musig2_test.go
```diff
@@ -1443,6 +1444,228 @@ func (mr *memsetRandReader) Read(buf []byte) (n int, err error) {
...
+		return nil, err
+	}
+	r2, err := btcec.ParsePubKey(
+		combinedNonce[btcec.PubKeyBytesLenCompressed:],
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	var nonce, r1J, r2J btcec.JacobianPoint
```
Roasbeef, line 1671: Should we use this as an intermediate routine for the normal signing/verification stuff?

btcec/schnorr/musig2/nonces.go
```diff
@@ -6,11 +6,13 @@ import (
 	"bytes"
 	"crypto/rand"
 	"encoding/binary"
+	"fmt"
```
Roasbeef, line 9: Commit message body could use a bit more detail.

btcec/schnorr/musig2/nonces.go
```diff
@@ -359,27 +359,8 @@ func AggregateNonces(pubNonces [][PubNonceSize]byte) ([PubNonceSize]byte, error)
 			)
 		}
 
-		// Now that we've aggregated all the points, we need to check
-		// if this point is the point at infinity, if so, then we'll
-		// just return the generator. At a later step, the malicious
-		// party will be detected.
-		if aggregateNonce == infinityPoint {
```
Roasbeef, line 342: You have a commit to point to for this change in the test vectors?

btcec/schnorr/musig2/nonces.go
```diff
@@ -392,18 +373,71 @@ func AggregateNonces(pubNonces [][PubNonceSize]byte) ([PubNonceSize]byte, error)
...
+	}
+
+	format := nonce[0]
+	switch format {
+	case 0x00:
+		result.X.SetInt(0)
+		result.Y.SetInt(0)
+		result.Z.SetInt(0)
+		return nil
+	case secp.PubKeyFormatCompressedEven, secp.PubKeyFormatCompressedOdd:
```
Roasbeef, line 410: `ParsePubkey` already contains this logic, so it's redundant here.

btcec/schnorr/musig2/nonces.go
```diff
@@ -392,18 +373,71 @@ func AggregateNonces(pubNonces [][PubNonceSize]byte) ([PubNonceSize]byte, error)
...
+		JacobianToByteSlice(combinedNonce2),
 	)
 
 	return finalNonce, nil
 }
+
+// NonceToJacobian parses the nonce as a secp.Pubkey and converts the pubkey
+// to the passed result. If the nonce is a zero slice, the infinityPoint is
+// returned.
+func NonceToJacobian(nonce []byte, result *btcec.JacobianPoint) error {
```
Roasbeef, line 396: How about we make this a bit more generic, and move it up to the `btcec` module level. 

So we'd end up w/ something like:
```go
func ParseJacobian(point []byte) (*btcec.JacobianPoint, error) {
    point, err := ParsePubKey(point)
    if err != nil {
        return err
    }
    
   var j btcec.JacobianPoint
    point.AsJacobian(j)
 
    return &j, nil
}
```

btcec/schnorr/musig2/nonces.go
```diff
@@ -392,18 +373,71 @@ func AggregateNonces(pubNonces [][PubNonceSize]byte) ([PubNonceSize]byte, error)
...
+	}
+	noncePk.AsJacobian(result)
+
+	return nil
+}
+
+// JacobianToByteSlice converts the passed secp.JacobianPoint to a secp.Pubkey
+// and serializes that to a byte slice. If the JacobianPoint is the infinity
+// point, a zero slice is returned.
+func JacobianToByteSlice(point secp.JacobianPoint) []byte {
```
Roasbeef, line 429: Similar comment here re elevating to the top level of the module.

btcec/schnorr/musig2/nonces.go
```diff
@@ -392,18 +373,71 @@ func AggregateNonces(pubNonces [][PubNonceSize]byte) ([PubNonceSize]byte, error)
...
+
+	if point.X == infinityPoint.X && point.Y == infinityPoint.Y {
+		return make([]byte, 33)
+	}
+
```
Roasbeef, line 434: You need to call `ToAffine()` before it's ok to just set the X and Y coords in the pubkey.

btcec/schnorr/musig2/nonces.go
Roasbeef, line 429: Also this should be `btcec.JacobianPoint` as it's a type alias.

btcec/schnorr/musig2/sign.go
```diff
@@ -709,3 +721,10 @@ func GetGenerator() *secp.PublicKey {
 	secp.ScalarBaseMultNonConst(&k, &result)
 	return secp.NewPublicKey(&result.X, &result.Y)
 }
+
+// SetJacobianToGenerator sets the passed JacobianPoint to the Generator Point.
+func SetJacobianToGenerator(jacobian *secp.JacobianPoint) {
```
Roasbeef, line 726: Similar q here, can be: `btcec.GeneratorJacobian`

btcec/schnorr/musig2/musig2_test.go
```diff
@@ -1665,6 +1685,13 @@ func aggNonceToPubkey(combinedNonce [66]byte, combinedKey *AggregateKey, msg [32
 	), nil
 
 }
+func mustNonceAgg(nonces [][66]byte) [66]byte {
```
Roasbeef, line 1688: Style nit: missing newline above here.

Roasbeef (changes requested):
Nice work! Happy to see there weren't any fundamental correctness issues discovered with the addition of the new test vectors. 

Completed an initial pass, mainly some style comments, and also some suggestions to elevate the utility functions added to the top-level module. This'll allow us to de-dup common routines like parsing a point into jacobian coords across a few of our projects.

btcec/schnorr/musig2/musig2_test.go
sputn1ck: Do you mean that we want to test all the remaining keys of the current testcase? This returns from ```t.Run(testname, func(t *testing){// returns from here})``` and moves to the next testcase.

btcec/schnorr/musig2/nonces.go
sputn1ck: Oh I must have forgot, this is actually the "breaking" 0.4.0 change https://github.com/jonasnick/bips/commit/20ba03106d245d42375e65f739c58dd9074daf6e

Roasbeef:
Bump.

btcec/schnorr/musig2/keys.go
```diff
@@ -286,8 +290,10 @@ func tweakKey(keyJ btcec.JacobianPoint, parityAcc btcec.ModNScalar, tweak [32]by
 	// Next, map the tweak into a mod n integer so we can use it for
 	// manipulations below.
 	tweakInt := new(btcec.ModNScalar)
-	tweakInt.SetBytes(&tweak)
-
+	overflows := tweakInt.SetBytes(&tweak)
+	if overflows == 1 {
+		return keyJ, parityAcc, tweakAcc, ErrTweakedKeyOverflows
+	}
```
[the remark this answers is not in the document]
sputn1ck: like this?
```go
if overflows == 1 {
		return keyJ, parityAcc, tweakAcc, ErrTweakedKeyOverflows

	}
```

sputn1ck:
Added the requested changes, however I'm unsure re `continue` on error vs `return`

btcec/schnorr/musig2/musig2_test.go
sputn1ck: I'm not sure I follow. Do you mean change the signature of `CombineSigs`?

btcec/btcec.go
```diff
@@ -39,3 +39,16 @@ type CurveParams = secp.CurveParams
 func Params() *CurveParams {
 	return secp.Params()
 }
+
+// Generator returns a public key at the Generator Point.
+func Generator() *PublicKey {
```
Roasbeef, line 44: 👍

btcec/schnorr/musig2/keys.go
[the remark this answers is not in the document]
Roasbeef: No I mean one _after_ this, so a new line between it and the comment that starts on the next line

btcec/schnorr/musig2/musig2_test.go
```diff
@@ -56,8 +94,11 @@ func TestMuSig2KeyAggTestVectors(t *testing.T) {
 	var jsonCases []jsonKeyAggTestCase
 
 	testCases := []struct {
-		keyOrder    []int
-		expectedKey []byte
+		keyOrder      []int
+		explicitKeys  []*secp256k1.PublicKey
```
Roasbeef, line 101: Should be `btcec`, this way we always go through our aliases.

btcec/schnorr/musig2/musig2_test.go
Roasbeef: I mean reducing duplication by moving some of the above into a new routine, so we can use it elsewhere. Don't consider it a blocker though, IIRC there's an existing TODO lingering around somewhere for it

btcec/curve.go
```diff
@@ -61,3 +66,51 @@ func ScalarBaseMultNonConst(k *ModNScalar, result *JacobianPoint) {
 func ScalarMultNonConst(k *ModNScalar, point, result *JacobianPoint) {
 	secp.ScalarMultNonConst(k, point, result)
 }
+
+// ParseJacobian parses the nonce as a secp.Pubkey and converts the pubkey
+// to the passed result. If the nonce is a zero slice, the infinityPoint is
+// returned.
+func ParseJacobian(nonce []byte) (JacobianPoint, error) {
```
Roasbeef, line 73: Not specific to a nonce, is really just a key, so perhaps rename to `point`?

btcec/curve.go
```diff
@@ -61,3 +66,51 @@ func ScalarBaseMultNonConst(k *ModNScalar, result *JacobianPoint) {
...
+	noncePk.AsJacobian(&result)
+
+	return result, nil
+}
+
+// JacobianToByteSlice converts the passed JacobianPoint to a Pubkey
+// and serializes that to a byte slice. If the JacobianPoint is the infinity
+// point, a zero slice is returned.
+func JacobianToByteSlice(point JacobianPoint) []byte {
+
```
Roasbeef, line 99: Stray new line

btcec/schnorr/musig2/nonces.go
```diff
@@ -331,22 +331,21 @@ func AggregateNonces(pubNonces [][PubNonceSize]byte) ([PubNonceSize]byte, error)
...
 		for i, pubNonceBytes := range pubNonces {
 			// Using the slicer, extract just the bytes we need to
 			// decode.
 			var nonceJ btcec.JacobianPoint
-			pubNonce, err := btcec.ParsePubKey(
-				slicer(pubNonceBytes),
-			)
+
+			nonceJ, err := btcec.ParseJacobian(
+				slicer(pubNonceBytes))
```
Roasbeef, line 344: I think it can fit all one the next line? If not, the closing paren here should be on a new line

Roasbeef (commented):
Latest changes are looking really good! Just a few minor comments

Roasbeef:
cc @guggero

btcec/schnorr/musig2/nonces.go
sputn1ck: It would be 81 chars

btcec/schnorr/musig2/nonces.go
```diff
@@ -270,6 +270,16 @@ func GenNonces(options ...NonceGenOption) (*Nonces, error) {
 		return nil, err
 	}
 
+	// If the options contain a secret key, we XOR it with with the tagged
+	// random bytes.
+	if len(opts.secretKey) != 0 {
```
guggero, line 289: Check the length being 32 exactly instead of just not 0? Otherwise we could panic below.

btcec/schnorr/musig2/nonces.go
```diff
@@ -270,6 +270,16 @@ func GenNonces(options ...NonceGenOption) (*Nonces, error) {
...
+		taggedHash := chainhash.TaggedHash(NonceAuxTag, randBytes[:])
+
+		for i, _ := range taggedHash {
```
guggero, line 292: To increase readability this could just be a classic for loop up to 32 (can be `chainhash.HashSize` to not use a magic number).

btcec/schnorr/musig2/musig2_test.go
```diff
@@ -825,3 +825,86 @@ func TestMuSigEarlyNonce(t *testing.T) {
 		t.Fatalf("final sig is invalid!")
 	}
 }
+
+func TestMusig2NonceGenTestVectors(t *testing.T) {
```
guggero, line 1331: nit: Godoc.

btcec/schnorr/musig2/musig2_test.go
```diff
@@ -825,3 +825,86 @@ func TestMuSigEarlyNonce(t *testing.T) {
...
+	t.Parallel()
+
+	msg := bytes.Repeat([]byte{0x01}, 32)
+	sk := bytes.Repeat([]byte{0x02}, 32)
+	aggpk := bytes.Repeat([]byte{0x07}, 32)
+	extra_in := bytes.Repeat([]byte{0x08}, 32)
```
guggero, line 1337: TIL about `bytes.Repeat()`, nice!

btcec/schnorr/musig2/nonces.go
```diff
@@ -175,6 +175,20 @@ func WithNonceAuxInput(aux []byte) NonceGenOption {
 	}
 }
 
+// withCustomOptions allows a caller to set pass a complete set of custom
```
guggero, line 178: nit: /to set pass/to set/ or "to pass"?

btcec/schnorr/musig2/nonces.go
```diff
@@ -175,6 +175,20 @@ func WithNonceAuxInput(aux []byte) NonceGenOption {
...
+// nonceGenOpts, without needing to create custom and checked structs such as
+// *btcec.PrivateKey. This is mainly used to match the testcases provided by
+// the MuSig BIP.
```
guggero, line 181: nit: MuSig2 BIP

btcec/btcec.go
```diff
@@ -39,3 +39,16 @@ type CurveParams = secp.CurveParams
 func Params() *CurveParams {
 	return secp.Params()
 }
+
+// Generator returns a public key at the Generator Point.
```
guggero, line 43: nit: returns **the** public key at the Generator Point?

btcec/btcec.go
```diff
@@ -39,3 +39,16 @@ type CurveParams = secp.CurveParams
...
+func Generator() *PublicKey {
+	var (
+		result JacobianPoint
+		k      secp.ModNScalar
+	)
+
+	k.SetInt(1)
+	ScalarBaseMultNonConst(&k, &result)
```
guggero, line 51: I think we need to call `result.ToAffine()` before taking its coordinates.

btcec/schnorr/musig2/musig2_test.go
```diff
@@ -26,14 +27,49 @@ var (
...
+	keySet := []*btcec.PublicKey{generator}
+
+	keysHash := keyHashFingerprint(keySet, true)
+	uniqueKeyIndex := secondUniqueKeyIndex(keySet, true)
+
+	n := &btcec.ModNScalar{}
+
+	n.SetByteSlice(invalidTweak)
+
+	coeff := aggregationCoefficient(keySet, generator, keysHash,
```
guggero, line 65: nit: formatting of multi-line function call.

btcec/curve.go
```diff
@@ -61,3 +66,50 @@ func ScalarBaseMultNonConst(k *ModNScalar, result *JacobianPoint) {
 func ScalarMultNonConst(k *ModNScalar, point, result *JacobianPoint) {
 	secp.ScalarMultNonConst(k, point, result)
 }
+
+// ParseJacobian parses a byte slice point as a secp.Pubkey and converts the
```
guggero, line 70: nit: s/secp.Pubkey/secp.PublicKey/

guggero (commented):
Very nice PR! Just a few minor comments, mostly style nits.

btcec/schnorr/musig2/nonces.go
sputn1ck: Should we throw an error if the secret key is not 0, but also not 32?

btcec/schnorr/musig2/nonces.go
sputn1ck: 🎶

btcec/schnorr/musig2/musig2_test.go
sputn1ck: like so?
```golang
coeff := aggregationCoefficient(
		keySet, generator, keysHash, uniqueKeyIndex,
	).Negate().Add(n)
```

sputn1ck:
Pushed the requested changes from @Roasbeef and @guggero

btcec/schnorr/musig2/musig2_test.go
guggero: Yes, that looks good to me :+1:

guggero (approved):

btcec/schnorr/musig2/nonces.go
Roasbeef: Looks like the latest diff changes the exact length 👍

Roasbeef (approved):
LGTM 🌺

btcec/schnorr/musig2/nonces.go
```diff
@@ -368,17 +371,18 @@ func AggregateNonces(pubNonces [][PubNonceSize]byte) ([PubNonceSize]byte, error)
...
 	if err != nil {
 		return finalNonce, err
 	}
 
-	copy(finalNonce[:], combinedNonce1.SerializeCompressed())
+	copy(finalNonce[:], btcec.JacobianToByteSlice(combinedNonce1))
 	copy(
 		finalNonce[btcec.PubKeyBytesLenCompressed:],
-		combinedNonce2.SerializeCompressed(),
+		btcec.JacobianToByteSlice(combinedNonce2),
```
Roasbeef, line 385: Looks like a test case expects an actual point, but then we serialize our "all zero" infinity point here (see CI).

Roasbeef (changes requested):

btcec/schnorr/musig2/nonces.go
sputn1ck: The test case should actually expect the infinity point (see https://github.com/jonasnick/bips/blob/8ddf79646ce9460896c9f5a74c3583ba6e405cb4/bip-musig2/reference.py#L504).

I broke that in the last force push. Fixed it in my last diff.

PR #1866, opened 2022-06-29
Bump btcd version in btcutil package

darioush:
Hi,

When I depend on `github.com/btcsuite/btcd/btcutil v1.1.1` (but not `github.com/btcsuite/btcd`), I end up with the 
`v0.22.0-beta.0.20220111032746-97732e52810c` version as a dependency in my go.sum.

This would normally be fine but it seems you have recently fixed an ambiguous import issue (https://github.com/btcsuite/btcd/issues/1839):
```	
github.com/btcsuite/btcd/chaincfg/chainhash: ambiguous import: found package github.com/btcsuite/btcd/chaincfg/chainhash in multiple modules:
	github.com/btcsuite/btcd v0.21.0-beta (/Users/darioush.jalali/go/pkg/mod/github.com/btcsuite/btcd@v0.21.0-beta/chaincfg/chainhash)
	github.com/btcsuite/btcd/chaincfg/chainhash v1.0.1 (/Users/darioush.jalali/go/pkg/mod/github.com/btcsuite/btcd/chaincfg/chainhash@v1.0.1)
```

So it would be beneficial to bump the version in the sub-module past your fix. 
Apologies in advance if you would have preferred an issue instead of a PR. Thanks :)


darioush:
Also noticed https://github.com/btcsuite/btcd/pull/1825 may fix this.

btcutil/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.16
 
 require (
 	github.com/aead/siphash v1.0.1
-	github.com/btcsuite/btcd v0.22.0-beta.0.20220111032746-97732e52810c
```
Roasbeef, line 7: This should be able to just jump to v0.23.0 now.

Roasbeef: Can you run a `go mod tidy` locally? If I do so, I get a small diff that isn't present in this PR. 

Also in the future, the best way to get turn around on a review is comment directly like you did here. Thanks!

darioush: Thanks. I ran `go mod tidy` with no diffs in the root directory but with some diffs when run from `btcutil/` and `btcetc/` so I added those changes.

(with go version = `go1.18.3`)

Roasbeef (approved):
LGTM 💵

PR #1868, opened 2022-07-04
txscript: add `SigHashTypeForScript` helper func

positiveblue:
Helper function that tries to identify the type of a script and returns the appropriate `SigHashType`.

### DIFF
--- a/txscript/sighash.go
+++ b/txscript/sighash.go
@@ -626,3 +626,20 @@ func CalcTapscriptSignaturehash(sigHashes *TxSigHashes, hType SigHashType,
 		sigHashes, hType, tx, idx, prevOutFetcher, opts...,
 	)
 }
+
+// SigHashTypeForScript returns the sighash flag for the given UTXO's pkScript.
+func SigHashTypeForScript(script []byte) (SigHashType, error) {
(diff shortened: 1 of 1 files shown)


guggero (approved):

txscript/sighash.go
```diff
@@ -626,3 +626,20 @@ func CalcTapscriptSignaturehash(sigHashes *TxSigHashes, hType SigHashType,
 		sigHashes, hType, tx, idx, prevOutFetcher, opts...,
 	)
 }
+
+// SigHashTypeForScript returns the sighash flag for the given UTXO's pkScript.
+func SigHashTypeForScript(script []byte) (SigHashType, error) {
```
Roasbeef, line 631: Hmm, I'm not sure how generally useful this is: it 100% assumes that a given script will use sighash all for taproot. Also there isn't necessarily a direct link between a script and the type of sighash it'll use. This would typically be dependent on the application context.

positiveblue: `This would typically be dependent on the application context.` <= my fault. I thought there as a direct (or at least in "almost all cases") link between sighash <=> scriptype

PR #1876, opened 2022-08-11
connmgr: IPv6 for TorLookupIP

Crypt-iQ:
This allows parsing a returned ipv6 address instead of returning an error when trying to resolve through the tor socks5 proxy

### DIFF
--- a/connmgr/tor.go
+++ b/connmgr/tor.go
@@ -106,24 +106,39 @@ func TorLookupIP(host, proxy string) ([]net.IP, error) {
 		}
 		return nil, ErrTorInvalidProxyResponse
 	}
-	if buf[3] != 1 {
+
+	addr := make([]net.IP, 1)
+
+	switch buf[3] {
+	case 1:
+		buf = make([]byte, 4)
+		bytes, err := conn.Read(buf)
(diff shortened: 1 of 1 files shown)


connmgr/tor.go
```diff
@@ -106,24 +106,39 @@ func TorLookupIP(host, proxy string) ([]net.IP, error) {
...
-	if err != nil {
-		return nil, err
-	}
-	if bytes != 4 {
-		return nil, ErrTorInvalidAddressResponse
-	}
+	addr := make([]net.IP, 1)
 
-	r := binary.BigEndian.Uint32(buf)
+	switch buf[3] {
```
Roasbeef, line 112: Can we extract some test cases from either the Tor docs, or that Python tool so we can verify the response parsing?

Roasbeef (changes requested):

connmgr/tor.go
Crypt-iQ: This isn't tor-specific -- this is the socks5 response format. We might be able to use a socks5 library, but I don't know if the library could let us specify the `0xF0` byte for "Tor resolve"

PR #1879, opened 2022-08-16
Add wire.Outpoint parser from string

GeorgeTsagk:
- Adds `NewOutPointFromString` function in `wire` package.
- Adds unit tests for `NewOutPointFromString`

### DIFF
--- a/wire/msgtx.go
+++ b/wire/msgtx.go
@@ -6,9 +6,11 @@ package wire
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"strconv"
+	"strings"
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 )
@@ -213,6 +215,29 @@ func NewOutPoint(hash *chainhash.Hash, index uint32) *OutPoint {
 		Hash:  *hash,
 		Index: index,
 	}
+}
+
+// NewOutPointFromString returns a new bitcoin transaction outpoint parsed from
+// the provided string, which should be in the format "hash:index".
+func NewOutPointFromString(outpoint string) (*OutPoint, error) {
+	parts := strings.Split(outpoint, ":")
+	if len(parts) != 2 {
+		return nil, errors.New("outpoint should be of the form txid:index")
+	}
+	hash, err := chainhash.NewHashFromStr(parts[0])
+	if err != nil {
+		return nil, err
+	}
+
+	outputIndex, err := strconv.ParseUint(parts[1], 10, 32)
+	if err != nil {
+		return nil, fmt.Errorf("invalid output index: %v", err)
+	}
+
+	return &OutPoint{
+		Hash:  *hash,
+		Index: uint32(outputIndex),
+	}, nil
 }
 
 // String returns the OutPoint in the human-readable form "hash:index".
--- a/wire/msgtx_test.go
+++ b/wire/msgtx_test.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/davecgh/go-spew/spew"
+	"github.com/stretchr/testify/require"
 )
 
 // TestTx tests the MsgTx API.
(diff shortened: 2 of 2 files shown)


wire/msgtx.go
```diff
@@ -215,6 +217,29 @@ func NewOutPoint(hash *chainhash.Hash, index uint32) *OutPoint {
...
+	parts := strings.Split(outpoint, ":")
+	if len(parts) != 2 {
+		return nil, errors.New("outpoint should be of the form txid:index")
+	}
+	hash, err := chainhash.NewHashFromStr(parts[0])
+	if err != nil {
+		return nil, err
+	}
+
+	outputIndex, err := strconv.Atoi(parts[1])
```
guggero, line 232: I think we should use `strconv.ParseInt(parts[1], 10, 64)` here, just because on a 32-bit system the plain `int` type would overflow with a full `uint32` value. It's quite unlikely for a TX to have that many outputs, I know. But I know of a specific use case where this could cause problems.

wire/msgtx_test.go
```diff
@@ -778,6 +778,84 @@ func TestTxWitnessSize(t *testing.T) {
...
+		},
+		{
+			name:   "empty string",
+			input:  "",
+			result: nil,
+			err:    true,
+		},
+	}
+
+	t.Logf("Running %d tests", len(tests))
```
guggero, line 827: Could use sub tests to get some logging for free.

wire/msgtx_test.go
```diff
@@ -778,6 +778,84 @@ func TestTxWitnessSize(t *testing.T) {
...
+	for i, test := range tests {
+		t.Logf("Test #%v: %v", i, test.name)
+		outpoint, err := NewOutPointFromString(test.input)
+
+		isErr := (err != nil)
+		if isErr != test.err {
+			t.Errorf("MsgTx.NewOutPointFromString: #%d got: %v,"+
+				" want: %v", i, isErr, test.err)
+			continue
```
guggero, line 836: Use `t.Fatalf` instead of `Errorf` and `continue`?

guggero (commented):
Nice feature! Definitely nice to have that in the `wire` package directly.

GeorgeTsagk:
@guggero 
Also added a new test case named `big index outpoint`, which includes a tx index which is a bit higher than the max `int32` value.

guggero (approved):
LGTM :tada:

wire/msgtx.go
```diff
@@ -215,6 +217,29 @@ func NewOutPoint(hash *chainhash.Hash, index uint32) *OutPoint {
...
+	parts := strings.Split(outpoint, ":")
+	if len(parts) != 2 {
+		return nil, errors.New("outpoint should be of the form txid:index")
+	}
+	hash, err := chainhash.NewHashFromStr(parts[0])
+	if err != nil {
+		return nil, err
+	}
+
+	outputIndex, err := strconv.ParseInt(parts[1], 10, 64)
```
bhandras, line 232: nit: probably better to use `ParseUint` and also to specify 32 for the `bitSize`.

https://pkg.go.dev/strconv#ParseUint

wire/msgtx_test.go
```diff
@@ -778,6 +778,88 @@ func TestTxWitnessSize(t *testing.T) {
...
+			err:    true,
+		},
+	}
+
+	for i, test := range tests {
+		t.Run(fmt.Sprintf("%v", test.name), func(t *testing.T) {
+			outpoint, err := NewOutPointFromString(test.input)
+
+			isErr := (err != nil)
+			if isErr != test.err {
```
bhandras, line 841: nit: why not use `require`?

wire/msgtx_test.go
```diff
@@ -778,6 +779,77 @@ func TestTxWitnessSize(t *testing.T) {
...
+		{
+			name:   "empty string",
+			input:  "",
+			result: nil,
+			err:    true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(fmt.Sprintf("%v", test.name), func(t *testing.T) {
```
bhandras, line 838: nit: could just pass `test.name` to `t.Run`

wire/msgtx_test.go
```diff
@@ -778,6 +779,77 @@ func TestTxWitnessSize(t *testing.T) {
...
+			outpoint, err := NewOutPointFromString(test.input)
+
+			isErr := (err != nil)
+			require.Equal(t, isErr, test.err)
+
+			if !isErr {
+				require.Equal(t, outpoint.Hash, test.result.Hash)
```
bhandras, line 845: nit: would `require.Equal(t, test.result, outpoint)` work? (Changed the order as the first should always be the expected result).

bhandras (approved):
Just two small nits remain, otherwise LGTM :tada:

Roasbeef (approved):
LGTM 🦖

PR #1882, opened 2022-09-03
build: bump golang base image version to 1.17

tochicool:
Image fails to build on the current 1.16 tag with the error:
```
> docker build . --build-arg ARCH=amd64
...
btcd.go:21:2: //go:build comment without // +build comment
The command '/bin/sh -c set -ex   && if [ "${ARCH}" = "amd64" ]; then export GOARCH=amd64; fi   && if [ "${ARCH}" = "arm32v7" ]; then export GOARCH=arm; fi   && if [ "${ARCH}" = "arm64v8" ]; then export GOARCH=arm64; fi   && echo "Compiling for $GOARCH"   && go install -v . ./cmd/...' returned a non-zero code: 1
```


jcvernaleo (approved):
OK

PR #1890, opened 2022-09-27
make: add install and release-install goals

guggero:
To simplify building the release-grade (stripped and
reproducible) binaries from source, we add the install and
release-install make goals. Running either of the commands will create
binaries in the $GOPATH/bin directories.
The main difference between the two goals is that the release-install
will not contain any local paths and no debug information.


sputn1ck (commented):
LTGM tACK! 🚀

sputn1ck (approved):

.gitignore
```diff
@@ -40,3 +40,10 @@ coverage.txt
 btcec/coverage.txt
 btcutil/coverage.txt
 btcutil/psbt/coverage.txt
+
+# Binaries produced by "make build"
+/addblock
```
jcvernaleo, line 45: Is the path here right?  Why `/addblock` and not just `addblock`?

jcvernaleo (changes requested):

Roasbeef (approved):
LGTM 🐊


Just needs a rebase!

jcvernaleo (approved):
Yep, rebase and good to go.

guggero:
Rebased.

PR #1893, opened 2022-10-03
update installation instruction and add git as build dependency

bokunodev:
update installation instruction and add git as build dependency

Changes to be committed:
	modified:   README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,8 +62,10 @@ recommended that `GOPATH` is set to a directory in your home directory such as
 - Run the following commands to obtain btcd, all dependencies, and install it:
 
 ```bash
-$ cd $GOPATH/src/github.com/btcsuite/btcd
-$ GO111MODULE=on go install -v . ./cmd/...
+$ cd $GOPATH
+$ git clone git@github.com:btcsuite/btcd.git
+$ cd btcd
+$ go install -v ./cmd/...
 ```
 
 - btcd (and utilities) will now be installed in ```$GOPATH/bin```.  If you did
@@ -77,9 +79,9 @@ $ GO111MODULE=on go install -v . ./cmd/...
 - Run the following commands to update btcd, all dependencies, and install it:
 
 ```bash
-$ cd $GOPATH/src/github.com/btcsuite/btcd
+$ cd $GOPATH/btcd
 $ git pull
-$ GO111MODULE=on go install -v . ./cmd/...
+$ go install -v ./cmd/...
 ```
 
 ## Getting Started


guggero (approved):

README.md
```diff
@@ -62,8 +62,10 @@ recommended that `GOPATH` is set to a directory in your home directory such as
 - Run the following commands to obtain btcd, all dependencies, and install it:
 
 ```bash
-$ cd $GOPATH/src/github.com/btcsuite/btcd
-$ GO111MODULE=on go install -v . ./cmd/...
+$ cd $GOPATH
+$ git clone git@github.com:btcsuite/btcd.git
```
chappjc, line 66: Users should not be filling up the top of their `GOPATH` like this.  The only things in `GOPATH` should be bin, pkg, src.
The user can checkout the repo whereever they want e.g. `mkdir -p ~/github/btcsuite`

chappjc: Regardless of how people decide to use their `GOPATH` folder, a more fundamental point is that you don't even need `GOPATH` defined in your shell these days (I no longer set it, and the [Go install instructions](https://go.dev/doc/install) no longer suggest to either).

From `go help gopath`:

> If the environment variable is unset, **GOPATH defaults**
> **to a subdirectory named "go" in the user's home directory**
> ($HOME/go on Unix, %USERPROFILE%\go on Windows),
> unless that directory holds a Go distribution.
> Run "go env GOPATH" to see the current GOPATH.
> 
> ...
> 
> **Each directory listed in GOPATH must have a prescribed structure**:
> 
> The **src directory** holds source code. The path below src
> determines the import path or executable name.
> 
> The **pkg directory** holds installed package objects.
> As in the Go tree, each target operating system and
> architecture pair has its own subdirectory of pkg
> (pkg/GOOS_GOARCH).
> 
> ...
> 
> The **bin directory** holds compiled commands.
> 
> ...
> 
> GOPATH and Modules
> 
> When using modules, **GOPATH is no longer used for resolving imports**.
> However, it is still used to store downloaded source code (in GOPATH/**pkg/mod**)
> and compiled commands (in GOPATH/**bin**).

So, the `GOPATH/src` path is almost completely obsolete, and I just don't think modern READMEs should advise using it to hold modern source (having a go.mod).  It's just my opinion, but any instructions should suggest a different place to checkout source _that contains a module_.

Regarding the problem of `replaces` in release tags, the resolution is fortunately simple (no need to put required modules in different repositories).  The required submodules just need to be tagged _first_ so that the product release tag can reference them in it's go.mod.  The release process for modules that require submodules from the same repo:
- tag any required submodules
- update the main module's go.mod (like the main btcd module) to use those tags, and remove all `replaces` statements
- tag the main module
If that is done going forward (as was for for some previous release), then `go install` can be used directly without bothering with git.  Until then, git is needed even to build releases from source, in addition to development code.

PR #1894, opened 2022-10-06
btcec/schnorr/musig2: update to musig 1.0.0

Roasbeef:
The two _concrete_ changes between version 0.4.0 (where we are before
this commit), and version 0.7.0 are:

  1. Variable length messages are now allowed, this comes with a new 8
     byte prefix for the messages.

     * Our implementation was already using a `[]byte` for the
       message/hash, so no extra API changes are needed here.

  2. The serialization for a blank message and a normal message (for
     nonce gen) is now distinct. A single byte is added (either 0 or 1)
     to indicate if a message was passed into nonce generation.

The major change in musig 1.0.0 is that plain public keys are used as
input to key aggregation.

Some correctness tests are failing, so leaving this in draft until I fix those
and also update to the latest set of test vectors.


Roasbeef:
Correctness tests work now, next step is to update all the test vectors.

Roasbeef:
Ok, this is ready for review now: all test vectors pass!

btcec/schnorr/musig2/sign.go
```diff
@@ -299,31 +299,22 @@ func Sign(secNonce [SecNonceSize]byte, privKey *btcec.PrivateKey,
...
-	}()
 	combinedKeyYIsOdd := func() bool {
 		combinedKeyBytes := combinedKey.FinalKey.SerializeCompressed()
 		return combinedKeyBytes[0] == secp.PubKeyFormatCompressedOdd
 	}()
 
-	// Next we'll compute our two parity factors for Q the combined public
-	// key, and P, the public key we're signing with. If the keys are odd,
-	// then we'll negate them.
+	// Next we'll compute the two parity factors for Q, the combined key.
```
guggero, line 327: nit: isn't it just one parity factor now?

btcec/schnorr/musig2/nonces.go
```diff
@@ -255,17 +255,19 @@ func genNonceAuxBytes(rand []byte, i int,
 		return nil, err
 	}
 
-	switch len(opts.msg) {
+	switch {
 	// If the message isn't present, then we'll just write out a single
 	// uint8 of a zero byte: m_prefixed = bytes(1, 0).
-	case 0:
+	case opts.msg == nil:
```
guggero, line 261: nit: squash this commit with previous one that updates the message encoding logic?

guggero (approved):
Very nice that we're now using the BIP's test vectors directly. LGTM :tada:

sputn1ck (approved):
LGTM!

We could think about adding the changes from https://github.com/jonasnick/bips/pull/74 (if that is a final and required change)

Roasbeef:
> I think there is no reason to not add the privkey to the other use of GenNonces here


Great point, added in a new commit tacked onto the end. 

> In order to mitigate https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2022-October/021000.html we could add in the public key to GenNonces, as in https://github.com/jonasnick/bips/pull/74.

So this is already available as a cutom option via `WithNonceCombinedKeyAux`.

PR #1896, opened 2022-10-10
wire: remove erroneous witness size check in wire parsing

Roasbeef:
In this commit, we fix a bug that would cause nodes to be unable to parse a given block from the wire. The block would be properly accepted if fed in via other mechanisms.

The issue here is that the old checks for the maximum witness size, circa segwit v0 where placed in the wire package _as well_ as the tx engine. This check should only be in the engine, since it's properly gated by other related scrip validation flags.

The fix itself is simple: limit witnesses only based on the maximum block size in bytes, or ~4MB.

### DIFF
--- a/wire/msgtx.go
+++ b/wire/msgtx.go
@@ -103,10 +103,9 @@ const (
 	maxWitnessItemsPerInput = 500000
 
 	// maxWitnessItemSize is the maximum allowed size for an item within
-	// an input's witness data. This number is derived from the fact that
-	// for script validation, each pushed item onto the stack must be less
-	// than 10k bytes.
-	maxWitnessItemSize = 11000
+	// an input's witness data. This value is bounded by the largest
+	// possible block size, post segwit v1 (taproot).
+	maxWitnessItemSize = 4_000_000
 )
 
 // TxFlagMarker is the first byte of the FLAG field in a bitcoin tx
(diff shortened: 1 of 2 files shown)


wire/msgtx.go
```diff
@@ -106,24 +106,26 @@ const (
 	// an input's witness data. This number is derived from the fact that
 	// for script validation, each pushed item onto the stack must be less
 	// than 10k bytes.
-	maxWitnessItemSize = 11000
+	maxWitnessItemSize = 4_000_000
```
otech47, line 109: why `4_000_000` instead of `4000000`?

hieblmi: Readability!?

Roasbeef: @otech47 thx for taking a look! It's the same thing, just using a literal to demarcate things in place of a comma (not valid syntax in Go).

Roasbeef: https://go.dev/doc/go1.13

otech47 (approved):

kaloudis:
tACK

wire/msgtx.go
```diff
@@ -106,24 +106,26 @@ const (
 	// an input's witness data. This number is derived from the fact that
 	// for script validation, each pushed item onto the stack must be less
 	// than 10k bytes.
```
vicariousdrama, line 108: Consider updating this comment in conjunction with the 4MB allowance

wire/msgtx.go
```diff
@@ -586,8 +587,9 @@ func (msg *MsgTx) BtcDecode(r io.Reader, pver uint32, enc MessageEncoding) error
 			// item itself.
 			txin.Witness = make([][]byte, witCount)
 			for j := uint64(0); j < witCount; j++ {
-				txin.Witness[j], err = readScript(r, pver,
-					maxWitnessItemSize, "script witness item")
+				txin.Witness[j], err = readScript(
```
ziggie1984, line 590: I am wondering why we do not enforce the different size limits for the different segwit version here ? For v_0 we stay with 11kb and for 4MB ?

wire/msgtx.go
bordalix: Why not 400_000?

That seems to be the maximum weight for transactions after Segwit support:
https://github.com/bitcoin/bitcoin/blob/master/src/policy/policy.h#L27

ziggie1984: The reason we don't use 400_000 is, bc thats only the policy rule of bitcoin-core nodes not the consensus of the bitcoin network. So transactions > 400_000 are considered non-standard but do comply with the consensus rules. As long as they are not bigger than the max_block size thats why we use 4_000_000

https://bitcoin.stackexchange.com/questions/100317/what-is-the-difference-between-policy-and-consensus-when-it-comes-to-a-bitcoin-c

bordalix: Thank for your answer.

wire/msgtx.go
Roasbeef: So that check is deeper in the consensus validation logic, this is a wire level DDoS defense limit/value.

PR #1899, opened 2022-10-10
GOGC and GOMAXPROCS configuration items added and defaults set

l0k18:
I think GOGC can be set via environment variable, the runtime checks for it, maybe, but for certain it is better done directly in the app.

Increasing GOMAXPROCS above the number of CPU threads can produce substantial improvements in initial block download and validation.

These both default to prior defaults if unset but provide a way for deployments to dramatically improve IBD and speed of updating the DB when new blocks arrive. Not important improvements but produce a ~50% time to initial sync at the cost of around double memory use.

IBD is a definite weak point for btcd versus bitcoind. There is minimal cost efficiency between a 4gb vs 2gb memory availability nowadays, so this brings btcd closer to parity of performance.

### DIFF
--- a/config.go
+++ b/config.go
@@ -16,12 +16,14 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"runtime/debug"
 	"sort"
 	"strconv"
 	"strings"
 	"time"
 
 	"github.com/btcsuite/btcd/blockchain"
+	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/connmgr"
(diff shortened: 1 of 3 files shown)


config.go
```diff
@@ -1136,6 +1143,28 @@ func loadConfig() (*config, []string, error) {
...
+	}
+
+	// Set the configured value on the runtime.
+	runtime.GOMAXPROCS(cfg.GoMaxProcs)
+
+	// Clamp GOGC between 1 and 1000
+	if cfg.GoGC < 1 {
+		cfg.GoGC = 1
+	}
+	if cfg.GoGC > 1000 {
```
Roasbeef, line 1161: FWIW this can already be set as an env variable: https://github.com/btcsuite/btcd/blob/718d268a8cf55af9b329bac9e171009352fbc10a/btcd.go#L331-L338

mleku:
fwiw, i extended this a little by embedding the sample configuration, integrating the RPC credentials rewrite, that to date has never been updated to 1.16 embed capability despite targeting 1.17.

https://github.com/mleku/btcd/commit/c08425ad7ca03e08ef7758e81c4df40c88ce21d4

PR #1900, opened 2022-10-10
Fixed ban bug that doesn't print numTxns

l0k18:
This is a typo fix. There is an underlying ban bug related to this that causes two btcd nodes to ban each other due to "transaction not found" errors.

### DIFF
--- a/server.go
+++ b/server.go
@@ -1440,7 +1440,7 @@ func (sp *serverPeer) OnNotFound(p *peer.Peer, msg *wire.MsgNotFound) {
 	}
 	if numTxns > 0 {
 		txStr := pickNoun(uint64(numTxns), "transaction", "transactions")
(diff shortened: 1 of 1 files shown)


jcvernaleo:
@l0k18 looks like a test fails.  Could you check that out?

l0k18:
https://github.com/Indra-Labs/btcd/commits/tx-not-found-error-bug 

Literally only changes one variable name specified in the error print. This test failure is not my work.

l0k18:
image

https://github.com/btcsuite/btcd/actions/runs/3219535761/jobs/5265307632

I don't see how this change would have affected this test but if so, the test is wrong.

jcvernaleo:
Well, looks like it was one of the random failures we see with GH Actions occasionally (still better than travis though).  Sorry for the extra trouble there.

jcvernaleo (approved):
OK

PR #1905, opened 2022-10-27
txscript: modify TweakTaprootPrivKey to operate on private key copy

Roasbeef:
In this commit, we fix an inadvertent mutation bug that would at times cause the private key passed into the tweak function to actually be *modified* in place.

We fix this by accepting the value instead of a pointer. The actual private key struct itself contains no pointer fields, so this is effectively a deep copy via dereference.

We also add a new unit test that fails w/o this change, to show that the private key was indeed being modified.


guggero (approved):
Nice, fix, LGTM :tada:

Crypt-iQ (approved):
very tricky

PR #1916, opened 2022-11-03
btcutil/psbt: add method Packet.GetTxFee

ffranr:
This commit adds a GetTxFee method to the psbt.Packet structure. The method returns the PSBT's fee.

### DIFF
--- a/btcutil/psbt/psbt.go
+++ b/btcutil/psbt/psbt.go
@@ -11,9 +11,9 @@ import (
 	"bytes"
 	"encoding/base64"
 	"errors"
-
 	"io"
 
+	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/wire"
 )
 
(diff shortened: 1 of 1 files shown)


btcutil/psbt/psbt.go
```diff
@@ -11,6 +11,7 @@ import (
 	"bytes"
 	"encoding/base64"
 	"errors"
+	"github.com/btcsuite/btcd/btcutil"
```
guggero, line 14: nit: import should be grouped with the other non-stdlib imports.

btcutil/psbt/psbt.go
```diff
@@ -395,3 +396,20 @@ func (p *Packet) SanityCheck() error {
 
 	return nil
 }
+
+// GetTxFee returns the transaction fee.  An error is returned if a transaction
+// input does not contain any UTXO information.
+func (p *Packet) GetTxFee() (btcutil.Amount, error) {
+	sumInputs, err := SumUtxoInputValues(p)
+	if err != nil {
+		return btcutil.Amount(0), err
```
guggero, line 405: You can just do `return 0, err`. Since 0 is then a constant, that's cast automatically at compile time. Makes it a bit easier to read.

guggero (approved):
Nice, LGTM :tada:

guggero:
cc @positiveblue.

Issue #1915, opened 2022-11-03
Explore possible refactors in btcd/wire/msgtx.go


otech47:
Given recent changes to this file in #1896 and #1907 were needed recently to patch and ship hotfixes in LND (ex: https://github.com/lightningnetwork/lnd/issues/7096 & https://github.com/lightningnetwork/lnd/issues/7002) it may be worth looking into whether some of these functions rely on, or are affected by, bitcoin consensus code so they can be moved/removed/refactored to reduce the chances of consensus conflicts and provide stronger assumptions for other software layers that may rely on btcd for consensus.

Some analysis of [the file](https://github.com/btcsuite/btcd/blob/1d767de1c739f3a5aca0d9c315c9e348ecae6cad/wire/msgtx.go#L845):

- 1 dependency on `chaincfg/chainhash`
- 20 constant values are defined, 2 of which involve maximum values for the number of witness items and their total size
- 7 `type`/`struct` definitions used throughout
- 36 functions
- 1 variable

The file is also covered by [10 test suites in `msgtx_test.go`](https://github.com/btcsuite/btcd/blob/1d767de1c739f3a5aca0d9c315c9e348ecae6cad/wire/msgtx_test.go#L1) (with multiple tests each) that does not seem to have been updated since taproot activation and perhaps could use further test cases to account for edge cases or non-standard transactions

![Screen Shot 2022-11-02 at 8 08 29 PM](https://user-images.githubusercontent.com/4914611/199625070-b1c68e6d-8b55-477f-bc61-f2ae025a07a7.png)

Does this seem like a worthwhile approach?

If the block weight limit were to be doubled as a consensus change tomorrow, would this file need to be changed to enforce consensus?

If so, can (should?) those changes be isolated and extracted into an explicitly marked consensus layer so that other software that depends on `/btcd/wire` can use the types/getters/other utility functions can be used independently of consensus assumptions?

My proficiency in go is not very high so I wanted to source feedback before diving deeper into these questions

guggero:
Refactor for the sake of refactor is often discouraged in critical code such as this. As refactor changes are hard to review and always bring the potential of introducing new bugs. So it's likely a big refactor PR won't be picked up for review very quickly.

But improving test coverage and adding specific test cases around some of the updated limits with Taproot consensus code would definitely be welcome IMO.

Roasbeef:
[...]

PR #1917, opened 2022-11-06
Add tests to trigger witness limit error paths

otech47:
To address #1915, I noticed the testing suite does not cover the possibility for very large transactions to trigger errors meant for denial-of-service protection.

Wire code is not consensus code per-say, but it's a grey area if you define consensus code as "anything that could halt execution before validation rules can be reached" so we should be wary of code that parses bytes straight from the wire since it could cause consensus failure

The `msgtx_test` suite also has not been updated since the taproot consensus upgrade so it's worth looking into what kinds of changes might make the wire parsing module more robust for P2TR spends. For example, all existing tests are only using `BaseEncoding` at the moment so adding redundant versions of those tests that use the `WitnessEncoding` may be useful?

Since refactoring such critical code is incredibly difficult and risky, this PR attempts to at least take the basic approach of adding more test cases to improve coverage.

I hope to add at least a few more cases before taking it out of draft state (and open to other suggestions)

- [x] Cover error path for `too many witness items to fit into max message size...` 
- [ ] Cover error path for `<scroptSize> is larger than the max allowed size...`
- [ ] Cover error path for `too many input transactions to fit into max message size...` (with `WitnessEncoding`)
- [ ] Cover error path for `too many output transactions to fit into max message size...` (with `WitnessEncoding`)


wire/msgtx_test.go
```diff
@@ -282,7 +282,8 @@ func TestTxWire(t *testing.T) {
 		// Latest protocol version with no transactions.
 		{
 			noTx,
-			noTx, noTxEncoded,
+			noTx,
+			noTxEncoded,
```
Roasbeef, line 286: 👍

wire/msgtx_test.go
```diff
@@ -779,6 +779,91 @@ func TestTxWitnessSize(t *testing.T) {
 	}
 }
 
+// TestTxWitnessLimits performs negative tests to ensure decoding
+// serialized transactions with witness data triggers proper error paths.
+func TestTxWitnessLimits(t *testing.T) {
+	// Use protocol version 70001 and transaction version 1 specifically
+	// here instead of the latest values because the test data is using
```
Roasbeef, line 786: Not quite wrapped to 80 chars.

wire/msgtx_test.go
```diff
@@ -779,6 +779,91 @@ func TestTxWitnessSize(t *testing.T) {
 	}
 }
 
+// TestTxWitnessLimits performs negative tests to ensure decoding
+// serialized transactions with witness data triggers proper error paths.
+func TestTxWitnessLimits(t *testing.T) {
```
Roasbeef, line 784: Can tack on `t.Parallel()`so it can run concurrently w/ all the other tests.

wire/msgtx_test.go
```diff
@@ -779,6 +779,91 @@ func TestTxWitnessSize(t *testing.T) {
...
+	tests := []struct {
+		buf     []byte          // Wire encoding
+		pver    uint32          // Protocol version for wire encoding
+		enc     MessageEncoding // Message encoding format
+		version uint32          // Transaction version
+		err     error           // Expected error
+	}{
+		// Transaction with 1 input and 4,000,001 items in the witness stack
+		{
+			encodedTxnWithManyWitnessItems,
```
Roasbeef, line 845: All the struct fields should be explicit, so:
```go
buf: encodedTxnWithManyWitnessItems, 
pver: pver,
```

etc

wire/msgtx_test.go
```diff
@@ -779,6 +779,91 @@ func TestTxWitnessSize(t *testing.T) {
...
+// TestTxWitnessLimits performs negative tests to ensure decoding
+// serialized transactions with witness data triggers proper error paths.
+func TestTxWitnessLimits(t *testing.T) {
+	// Use protocol version 70001 and transaction version 1 specifically
+	// here instead of the latest values because the test data is using
+	// bytes encoded with those versions.
+	pver := uint32(70001)
+	txVer := uint32(1)
+
+	var createTxnWithManyWitnessItems = func() []byte {
```
Roasbeef, line 791: Doesn't look to use many of the variables closed over, so I think we can just make this a regular function and then call it below.

wire/msgtx_test.go
```diff
@@ -779,6 +779,91 @@ func TestTxWitnessSize(t *testing.T) {
...
+		witnessItemCountEncoded := []byte{
+			0xfe, 0x01, 0x09, 0x3d, 0x00, // 4,000,001 items on the witness stack
+		}
+		encodedTxnBuffer = append(txnBeforeWitnessEncoded, witnessItemCountEncoded...)
+		// Add 4,000,001 witness items of 1-byte each
+		for i := uint(0); i < 4_000_001; i++ {
+			// First add the size of the witness item (1-byte)
+			encodedTxnBuffer = append(encodedTxnBuffer, "1"...)
+			// Then add the witness item itself (1-byte)
+			encodedTxnBuffer = append(encodedTxnBuffer, "0"...)
```
Roasbeef, line 824: Should add new lines between each logical stanza. Also the comments should be complete sentences.

wire/msgtx_test.go
```diff
@@ -779,6 +779,91 @@ func TestTxWitnessSize(t *testing.T) {
...
+		}
+		var locktimeEncoded = []byte{
+			0x0, 0x0, 0x0, 0x0, // Lock time
+		}
+		encodedTxnBuffer = append(encodedTxnBuffer, locktimeEncoded...)
```
Roasbeef, line 829: Another way to do this would be just to create the tx from scratch, but dynamically. 

So:
```go
megaTx := wire.NewMsgTx(2)
megaTx.AddTxIn(&wire.TxIn{})
megaTx.AddTxOut(&wire.TxOut{PkScript: script, amt: testAmt})

var megaWitness [][]byte
for i := uint(0); i < 4_000_001; i++ {
    megaWitness = append(megaWitness, []byte{0x00})
}
```

Roasbeef (changes requested):
Yo thanks for picking this up! A bit late w/ the review, but left some initial comments. Mainly that we can do this programmatically vs sort of hard coding the bytes of a raw tx.

wire/msgtx_test.go
otech47: hm i guess i was imagining that it'd be best to build the raw bytes to test as close to real input as possible but seems like it wouldn't make much of a difference in many cases so I may try to work those in

wire/msgtx_test.go
otech47: 👍

wire/msgtx_test.go
otech47: added a few more newlines where it seemed to make sense 👍

wire/msgtx_test.go
otech47: fixed

wire/msgtx_test.go
```diff
@@ -778,6 +779,149 @@ func TestTxWitnessSize(t *testing.T) {
 	}
 }
 
+// TestTxWitnessLimits performs negative tests to ensure decoding
+// serialized transactions with witness data triggers proper error paths.
+func TestTxWitnessLimits(t *testing.T) {
```
otech47, line 784: thinking more about it... if these limit checks are already happening deeper in consensus logic, I wonder if it doesn't make more sense to simply forget about witness limit checks and just do a <4MB size check on the entire txn? if these checks at a parsing level are really more about memory exhaustion safeguards, there shouldn't be a need to look for the witness specifically?

Roasbeef: > , I wonder if it doesn't make more sense to simply forget about witness limit checks and just do a <4MB size check on the entire txn?

That would certainly be the most conservative check. The issue with doing it on this layer though, is the possibility of future interactions that increase this value (say an ext block in the future or w/e). On the bitcoind side, they have such limits in place still, but it's around 32 MB or so (iirc the max p2p payload). So if we end up upgrading Bitcoin far in the future to 64 MB blocks or w/e, then that would cause a fork with these old clients.

Roasbeef:
Approved the CI run.

PR #1918, opened 2022-11-07
blockchain, btcjson: Implement getchaintips rpc call

### LINKED ISSUE #1912: Implement getchaintips RPC
The Bitcoin Core `getchaintips` RPC returns the list of known tip candidates, including which ones are considered invalid. This makes is easier for users (and automated systems) to figure out there's a potential consensus issue.

Specifically, ForkMonitor [uses this to send out an alert](https://github.com/BitMEXResearch/forkmonitor/blob/master/app/models/invalid_block.rb#L21) if any block is considered _valid_ by one implementation yet _invalid_ by another.

kcalvinalvin:
getchaintips rpc call is implemented and follows the algorithm that Bitcoin Core has in https://github.com/bitcoin/bitcoin/blob/50422b770a40f5fa964201d1e99fd6b5dc1653ca/src/rpc/blockchain.cpp#L1374.

Test coverage checks that the height, block hash, branchlen, and status are correct. I also tested the rpc on regtest with blocks: 

`78b945a390c561cf8b9ccf0598be15d7d85c67022bf71083c0b0bd8042fc30d7`
`1b97df2be6c634e40a5f410f4758ba707c4452cda6f3707b4935c9e7760df965`

which both build on top of the genesis block.

```
0000002006226e46111a0b59caaf126043eb5bbf28c34f3a5e332a1fc7b2b73cf188910f467c23519a04a309fb31f27572a5e8e8eadb9f1881a7c12b30a06438bdde486c5cca6463ffff7f200100000001020000000001010000000000000000000000000000000000000000000000000000000000000000ffffffff025100ffffffff0200f2052a01000000160014e96f87294541ef0574ba422198c73e702e22d7a60000000000000000266a24aa21a9ede2f61c3f71d1defd3fa999dfa36953755c690689799962b48bebd836974e8cf90120000000000000000000000000000000000000000000000000000000000000000000000000
```

raw blocks:
```
0000002006226e46111a0b59caaf126043eb5bbf28c34f3a5e332a1fc7b2b73cf188910f0cf276b0154586bd0755c963c2d2f4471504a8450575e30a7273d56b7e31f7cc7acf6463ffff7f200000000001020000000001010000000000000000000000000000000000000000000000000000000000000000ffffffff025100ffffffff0200f2052a010000001600149c349625c03446ec5b22f41d78d6089b82b909950000000000000000266a24aa21a9ede2f61c3f71d1defd3fa999dfa36953755c690689799962b48bebd836974e8cf90120000000000000000000000000000000000000000000000000000000000000000000000000
```

block: `716c71f0c4990732b51dc611825fdd11fc8dc471ffc3e5c1de86b5d6e4663ff2` builds on top of block `78b945a390c561cf8b9ccf0598be15d7d85c67022bf71083c0b0bd8042fc30d7` and it was used to test reorging and making sure getchaintips behavior matches that of bitcoin core.

raw block:
```
00000020d730fc4280bdb0c08310f72b02675cd8d715be9805cf9c8bcf61c590a345b9782b5d65a911e677d43c235c04fd504eb1c2c348b8853c0fcf10670c2c330949779fcf6463ffff7f200100000001020000000001010000000000000000000000000000000000000000000000000000000000000000ffffffff025200ffffffff0200f2052a01000000160014d0b47c9d09352c365bef96a0ed1fc67ac858e3140000000000000000266a24aa21a9ede2f61c3f71d1defd3fa999dfa36953755c690689799962b48bebd836974e8cf90120000000000000000000000000000000000000000000000000000000000000000000000000
```

Resolves #1912

### DIFF
--- a/blockchain/blockindex.go
+++ b/blockchain/blockindex.go
@@ -377,6 +377,44 @@ func (bi *blockIndex) UnsetStatusFlags(node *blockNode, flags blockStatus) {
 	bi.Unlock()
 }
 
+// InactiveTips returns all the block nodes that aren't in the best chain.
+//
+// This function is safe for concurrent access.
+func (bi *blockIndex) InactiveTips(bestChain *chainView) []*blockNode {
+	bi.RLock()
+	defer bi.RUnlock()
+
+	// Look through the entire blockindex and look for nodes that aren't in
+	// the best chain. We're gonna keep track of all the orphans and the parents
+	// of the orphans.
+	orphans := make(map[chainhash.Hash]*blockNode)
+	orphanParent := make(map[chainhash.Hash]*blockNode)
+	for hash, node := range bi.index {
+		found := bestChain.Contains(node)
+		if !found {
+			orphans[hash] = node
+			orphanParent[node.parent.hash] = node.parent
+		}
+	}
+
+	// If an orphan isn't pointed to by another orphan, it is a chain tip.
+	//
+	// We can check this by looking for the orphan in the orphan parent map.
+	// If the orphan exists in the orphan parent map, it means that another
+	// orphan is pointing to it.
+	tips := make([]*blockNode, 0, len(orphans))
+	for hash, orphan := range orphans {
+		_, found := orphanParent[hash]
+		if !found {
+			tips = append(tips, orphan)
+		}
+
+		delete(orphanParent, hash)
+	}
+
+	return tips
+}
+
 // flushToDB writes all dirty block nodes to the database. If all writes
 // succeed, this clears the dirty set.
 func (bi *blockIndex) flushToDB() error {
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -1256,6 +1256,119 @@ func (b *BlockChain) BestSnapshot() *BestState {
 	return snapshot
 }
 
+// TipStatus is the status of a chain tip.
+type TipStatus byte
+
+const (
+	// StatusUnknown indicates that the tip status isn't any of the defined
+	// statuses.
+	StatusUnknown TipStatus = iota
+
+	// StatusActive indicates that the tip is considered active and is in
+	// the best chain.
+	StatusActive
+
+	// StatusInvalid indicates that this tip or any of the ancestors of this
+	// tip are invalid.
+	StatusInvalid
+
+	// StatusValidFork is given if:
+	// 1: Not a part of the best chain.
+	// 2: Is not invalid.
+	// 3: Has the block data stored to disk.
+	StatusValidFork
+)
+
+// String returns the status flags as string.
+func (ts TipStatus) String() string {
+	switch ts {
+	case StatusActive:
+		return "active"
+	case StatusInvalid:
+		return "invalid"
+	case StatusValidFork:
+		return "valid-fork"
+	}
+	return fmt.Sprintf("unknown: %b", ts)
+}
+
+// ChainTip represents the last block in a branch of the block tree.
+type ChainTip struct {
+	// Height of the tip.
+	Height int32
+
+	// BlockHash hash of the tip.
+	BlockHash chainhash.Hash
+
+	// BranchLen is length of the fork point of this chain from the main chain.
+	// Returns 0 if the chain tip is a part of the best chain.
+	BranchLen int32
+
+	// Status is the validity status of the branch this tip is in.
+	Status TipStatus
+}
+
+// ChainTips returns all the chain tips the node itself is aware of.  Each tip is
+// represented by its height, block hash, branch length, and status.
+//
+// This function is safe for concurrent access.
+func (b *BlockChain) ChainTips() []ChainTip {
+	b.chainLock.RLock()
+	defer b.chainLock.RUnlock()
+
+	// Grab all the inactive tips.
+	tips := b.index.InactiveTips(b.bestChain)
+
+	// Add the current tip.
+	tips = append(tips, b.bestChain.Tip())
+
+	chainTips := make([]ChainTip, 0, len(tips))
+
+	// Go through all the tips and grab the height, hash, branch length, and the block
+	// status.
+	for _, tip := range tips {
+		var status TipStatus
+		switch {
+		// The tip is considered active if it's in the best chain.
+		case b.bestChain.Contains(tip):
+			status = StatusActive
+
+		// This block or any of the ancestors of this block are invalid.
+		case tip.status.KnownInvalid():
+			status = StatusInvalid
+
+		// If the tip meets the following criteria:
+		// 1: Not a part of the best chain.
+		// 2: Is not invalid.
+		// 3: Has the block data stored to disk.
+		//
+		// The tip is considered a valid fork.
+		//
+		// We can check if a tip is a valid-fork by checking that
+		// its data is available. Since the behavior is to give a
+		// block node the statusDataStored status once it passes
+		// the proof of work checks and basic chain validity checks.
+		//
+		// We can't use the KnownValid status since it's only given
+		// to blocks that passed the validation AND were a part of
+		// the bestChain.
+		case tip.status.HaveData():
+			status = StatusValidFork
+		}
+
+		chainTip := ChainTip{
+			Height:    tip.height,
+			BlockHash: tip.hash,
+			BranchLen: tip.height - b.bestChain.FindFork(tip).height,
+			Status:    status,
+		}
+
+		chainTips = append(chainTips, chainTip)
+	}
+
+	return chainTips
+}
+
 // HeaderByHash returns the block header identified by the given hash or an
 // error if it doesn't exist. Note that this will return headers from both the
 // main and side chains.
--- a/blockchain/chain_test.go
+++ b/blockchain/chain_test.go
@@ -5,6 +5,7 @@
 package blockchain
 
 import (
+	"fmt"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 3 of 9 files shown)


0xB10C:
I'm not familiar with the btcd process for this: How/when is [json_rpc_api.md](https://github.com/btcsuite/btcd/blob/master/docs/json_rpc_api.md) updated? Automatically when doing a release or manually in PRs? In case of the latter, adding RPC docs here would be good.

btcjson/chainsvrresults.go
```diff
@@ -117,6 +117,19 @@ type GetBlockVerboseTxResult struct {
...
+	Height    int32  `json:"height"`
+	Hash      string `json:"hash"`
+	BranchLen int32  `json:"branchlen"`
+	Status    string `json:"status"`
+}
+
+// GetChainTipsResult models the data from the getchaintips command.
+type GetChainTipsResult struct {
+	ChainTips []ChainTip `json:"chaintips"`
+}
```
0xB10C, line 131: I'm not sure how closely you want to follow the Bitcoin Core RPC response. From the [docs](https://github.com/btcsuite/btcd/blob/master/docs/json_rpc_api.md#1-overview) I got the impressions that some RPC calls are/were compatible with Bitcoin Core.

Bitcoin Core's `getchaintips` RPC returns a JSON list of chain tips as top level element:
```
[
  {
     "height": 115861,
     "hash":" 000000407476239a4870e510ce1dea262c9afa672eae3ab50a019812f83e0059",
     "branchlen": 0,
     "status": "active"
  }
]
```

This currently returns a JSON map with a single key called `chaintips` and the list of chain tips as value.


```
"chaintips": [
  {
     "height": 115861,
     "hash":" 000000407476239a4870e510ce1dea262c9afa672eae3ab50a019812f83e0059",
     "branchlen": 0,
     "status": "active"
  }
]
```

kcalvinalvin:
> I'm not familiar with the btcd process for this: How/when is [json_rpc_api.md](https://github.com/btcsuite/btcd/blob/master/docs/json_rpc_api.md) updated? Automatically when doing a release or manually in PRs? In case of the latter, adding RPC docs here would be good.

By the looks of it from this commit https://github.com/btcsuite/btcd/commit/a6c79c7a913b1f9f5e0a7c2aa9c6416ecd430c84, seems like you need to manually update the log when you implement the rpc call. I'll update the file.

btcjson/chainsvrresults.go
kcalvinalvin: Changed the output to match that of Bitcoin Core

kcalvinalvin:
Update `json_rpc_api.md` and change the RPC response to match that of Bitcoin Core

blockchain/blockindex.go
```diff
@@ -319,6 +319,44 @@ func (bi *blockIndex) UnsetStatusFlags(node *blockNode, flags blockStatus) {
...
+//
+// This function is safe for concurrent access.
+func (bi *blockIndex) InactiveTips(bestChain *chainView) []*blockNode {
+	bi.RLock()
+	defer bi.RUnlock()
+
+	// Look through the entire blockindex and look for nodes that aren't in
+	// the best chain. We're gonna keep track of all the orphans and the parents
+	// of the orphans.
+	orphans := map[chainhash.Hash]*blockNode{}
```
Roasbeef, line 332: Sort of non-traditional syntax here? Did a double take as was expecting:
```go
orphans := make(map[chainhash.Hash]*blockNode)
```

Usually this syntax is used to create a map literal where you already know some/all of the values.

blockchain/blockindex.go
```diff
@@ -319,6 +319,44 @@ func (bi *blockIndex) UnsetStatusFlags(node *blockNode, flags blockStatus) {
...
+			orphanParent[node.parent.hash] = node.parent
+		}
+	}
+
+	// If an orphan isn't pointed to by another orphan, it is a chain tip.
+	//
+	// We can check this by looking for the orphan in the orphan parent map.
+	// If the orphan exists in the orphan parent map, it means that another
+	// orphan is pointing to it.
+	var tips []*blockNode
```
Roasbeef, line 347: Can pre-size this based on either the orphans or the parent map.

blockchain/chain.go
```diff
@@ -1254,6 +1254,109 @@ func (b *BlockChain) BestSnapshot() *BestState {
...
+const (
+	// StatusUnknown indicates that the tip status isn't any of the defined
+	// statuses.
+	StatusUnknown TipStatus = iota
+	// StatusActive indicates that the tip is considered active and is in
+	// the best chain.
+	StatusActive
+	// StatusInvalid indicates that this tip or any of the ancestors of this
+	// tip are invalid.
+	StatusInvalid
```
Roasbeef, line 1269: Style nit: missing new line before the declarations.

blockchain/chain.go
```diff
@@ -1254,6 +1254,109 @@ func (b *BlockChain) BestSnapshot() *BestState {
...
+func (ts TipStatus) String() string {
+	switch ts {
+	case StatusActive:
+		return "active"
+	case StatusInvalid:
+		return "invalid"
+	case StatusValidFork:
+		return "valid-fork"
+	}
+	return "unknown"
```
Roasbeef, line 1287: Use `fmt.Sprintf` to print out the raw byte here which can be useful for logging to catch bugs?

blockchain/chain.go
```diff
@@ -1254,6 +1254,109 @@ func (b *BlockChain) BestSnapshot() *BestState {
...
+type ChainTip struct {
+	// Height of the tip.
+	Height int32
+	// BlockHash hash of the tip.
+	BlockHash chainhash.Hash
+	// Amount of blocks connecting this tip with the best chain. Returns
+	// 0 if the chain tip is a part of the best chain.
+	BranchLen int32
+	// Status is the validity status of the branch this tip is in.
+	Status TipStatus
```
Roasbeef, line 1300: Style nit: missing newline before the declarations.

blockchain/chain.go
```diff
@@ -1254,6 +1254,109 @@ func (b *BlockChain) BestSnapshot() *BestState {
...
+
+// ChainTip represents the last block in a branch of the block tree.
+type ChainTip struct {
+	// Height of the tip.
+	Height int32
+	// BlockHash hash of the tip.
+	BlockHash chainhash.Hash
+	// Amount of blocks connecting this tip with the best chain. Returns
+	// 0 if the chain tip is a part of the best chain.
+	BranchLen int32
```
Roasbeef, line 1298: `godoc` comment shoudl start with `BranchLen`

blockchain/chain.go
```diff
@@ -1254,6 +1254,109 @@ func (b *BlockChain) BestSnapshot() *BestState {
...
+	return "unknown"
+}
+
+// ChainTip represents the last block in a branch of the block tree.
+type ChainTip struct {
+	// Height of the tip.
+	Height int32
+	// BlockHash hash of the tip.
+	BlockHash chainhash.Hash
+	// Amount of blocks connecting this tip with the best chain. Returns
```
Roasbeef, line 1296: A bit confused re the wording here: this is meant to the the height gap between this non-main chain tip and the main chain itself?

blockchain/chain.go
```diff
@@ -1254,6 +1254,109 @@ func (b *BlockChain) BestSnapshot() *BestState {
...
+// represented by its height, block hash, branch length, and status.
+//
+// This function is safe for concurrent access.
+func (b *BlockChain) ChainTips() []ChainTip {
+	b.chainLock.RLock()
+	defer b.chainLock.RUnlock()
+
+	// Grab all the stale tips.
+	tips := b.index.InactiveTips(b.bestChain)
+	// Add the current tip.
```
Roasbeef, line 1313: Similar comment here re new line.

blockchain/chain.go
```diff
@@ -1254,6 +1254,109 @@ func (b *BlockChain) BestSnapshot() *BestState {
...
+	tips = append(tips, b.bestChain.Tip())
+
+	var chainTips []ChainTip
```
Roasbeef, line 1316: Can presize based on `len(tips)`.

blockchain/chain.go
```diff
@@ -1254,6 +1254,109 @@ func (b *BlockChain) BestSnapshot() *BestState {
...
+			// the proof of work checks and basic chain validity checks.
+			//
+			// We can't use the KnownValid status since it's only given
+			// to blocks that passed the validation AND were a part of
+			// the bestChain.
+			status = StatusValidFork
+		}
+
+		chainTip := ChainTip{
+			tip.height,
```
Roasbeef, line 1348: Should use literals for the name of each field.

blockchain/chain_test.go
```diff
@@ -964,3 +964,202 @@ func TestIntervalBlockHashes(t *testing.T) {
...
 	}
 }
+
+func TestChainTips(t *testing.T) {
+	type testcase struct {
+		name      string
+		chain     *BlockChain
+		chainTips map[chainhash.Hash]ChainTip
+	}
+	tests := []testcase{}
```
Roasbeef, line 974: Typically we'd do the initialization+declaration all in one statement. 

So one way to achieve this would be to have the `chainTips` be a closure that returns the chain tip itself:
```go
type testCase struct {
    chainTipGen func() map[chainhash.Hash]ChainTip
}
```

Then you'd have the current block as part of the slice definition itself. 

We typically don't use scoping blocks like this in the manner of C++ as it can make it harder to read/understand a code fragment.

Roasbeef (changes requested):
Thanks for picking this up! 

One thing I'd like to get back in the habit of is adding more [`rpctests`](https://github.com/btcsuite/btcd/tree/master/integration/rpctest). This would let us exercise the new RPC interface a bit more by doing things like mining invalid blocks, creating forks, etc -- then asserting that the final output is what we'd expect.

blockchain/chain.go
kcalvinalvin: It's meant as the amount of blocks the branch has from the oldest common ancestor with the active chain. Active chain always returns 0.

The below branch of "unknown" would have a `branchlen` of 2 since it forks off at block 10 and has 2 blocks built on top of block 10.
```
//      genesis -> 1 -> 2 -> 3 ... -> 10 -> 11  -> 12  -> 13 (active)
//                                      \-> 11a -> 12a (unknown)
```

kcalvinalvin:
Addressed all of the modifications requested. I'm not too familiar with how `rpctests` work. I'll look into it and add it as a new commit.

kcalvinalvin:
Added a new commit that adds a test for the `getchaintips` call in the integration tests.

blockchain/chain.go
Roasbeef: Makes a lot of sense, I think we can add some more of this context to the comment itself (can even add that diagram above).

Roasbeef (changes requested):
PR is shaping up pretty well, really dig the extra commit to add an integration test!

Roasbeef:
Interested in finishing this up for it to land?

integration/getchaintips_test.go
```diff
@@ -0,0 +1,367 @@
...
+		return fmt.Errorf("Expected status of %s but got %s", expected.Status, got.Status)
+	}
+
+	return nil
+}
+
+func TestGetChainTips(t *testing.T) {
+	// block1Hex is a block that builds on top of the regtest genesis block.
+	// Has blockhash of "36c056247e8c0589f6307995e4e13acf2b2b79cad9ecd5a4eeab2131ed0ecde5".
+	block1Hex := "0000002006226e46111a0b59caaf126043eb5bbf28c34f3a5e332a1fc7b2b73cf18891" +
```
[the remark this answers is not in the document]
kcalvinalvin: > Alternatively, we could also just generate these blocks.

We could and that'd be preferable. But at the moment there's no support for `invalidateblock` so we can't test for reorgs.

A timeline like below would work:
Get this PR in -> get `invalidateblock` in -> change tests to generate blocks

integration/getchaintips_test.go
```diff
@@ -0,0 +1,367 @@
...
+	if err != nil {
+		t.Fatalf("TestGetChainTips fail. Error: %v", err)
+	}
+
+	// Submit 4 blocks.
+	//
+	// Our chain view looks like so:
+	// (genesis block) -> 1 -> 2 -> 3 -> 4
+	blockStrings := []string{block1Hex, block2Hex, block3Hex, block4Hex}
+	for _, blockString := range blockStrings {
```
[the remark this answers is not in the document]
kcalvinalvin: Yup. But as mentioned in [here](https://github.com/btcsuite/btcd/pull/1918#discussion_r1264616210), we can't do that quite yet.

blockchain/chain.go
```diff
@@ -1254,6 +1254,116 @@ func (b *BlockChain) BestSnapshot() *BestState {
...
+
+	// Grab all the inactive tips.
+	tips := b.index.InactiveTips(b.bestChain)
+
+	// Add the current tip.
+	tips = append(tips, b.bestChain.Tip())
+
+	chainTips := make([]ChainTip, 0, len(tips))
+
+	// Go through all the tips and grab the height, hash, branch length, and the block
```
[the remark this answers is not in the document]
kcalvinalvin: All of the comments below are under 80 column though (with 8 space tabs). Could you point out which ones should be wrapped?

blockchain/chain_test.go
```diff
@@ -964,3 +965,193 @@ func TestIntervalBlockHashes(t *testing.T) {
...
+				chainTips[inactiveTip.BlockHash] = inactiveTip
+				chainTips[invalidTip.BlockHash] = invalidTip
+
+				return chain, chainTips
+			},
+		},
+	}
+
+	for _, test := range tests {
+		chain, expectedChainTips := test.chainTipGen()
```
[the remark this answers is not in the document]
kcalvinalvin: Is there a desirable column width limit? There's many parts of btcd code that's over 100.

(My personal taste is that anything under 110 is acceptable)

blockchain/blockindex.go
```diff
@@ -319,6 +319,44 @@ func (bi *blockIndex) UnsetStatusFlags(node *blockNode, flags blockStatus) {
...
+func (bi *blockIndex) InactiveTips(bestChain *chainView) []*blockNode {
+	bi.RLock()
+	defer bi.RUnlock()
+
+	// Look through the entire blockindex and look for nodes that aren't in
+	// the best chain. We're gonna keep track of all the orphans and the parents
+	// of the orphans.
+	orphans := make(map[chainhash.Hash]*blockNode)
+	orphanParent := make(map[chainhash.Hash]*blockNode)
+	for hash, node := range bi.index {
```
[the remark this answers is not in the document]
kcalvinalvin: Ah yeah I had a PR open for that a long time ago. Should clean that up too.

rpcserver.go
```diff
@@ -2192,6 +2193,28 @@ func handleGetBlockTemplate(s *rpcServer, cmd interface{}, closeChan <-chan stru
 	}
 }
 
+// handleGetChainTips implements the getchaintips command.
+func handleGetChainTips(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (interface{}, error) {
+	chainTips := s.cfg.Chain.ChainTips()
+
+	ret := []btcjson.GetChainTipsResult{}
+	for _, chainTip := range chainTips {
+		ret = append(ret, struct {
```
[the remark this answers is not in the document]
kcalvinalvin: Could you clarify what you mean here? I don't see any usage of `GetMempoolEntryResult` here and I don't quite see how `GetMempoolEntryResult` struct would help here

kcalvinalvin:
Addressed most things in the review and asked for clarification on the rest

saubyk:
Hi @Roasbeef this pr is also ready for your next round of review

Roasbeef (approved):
LGTM 🧩

PR #1919, opened 2022-11-09
btcec/schnorr: use private key copy for BIP-340 signatures

Roasbeef:
This is a fix similar to https://github.com/btcsuite/btcd/pull/1905. We'll always make a copy of the key in the local scope before passing it around elsewhere. Depending on the parity of the public key, the private key itself might need to be negated.

A similar test is added here that fails without the patch to the signature.go file.

### DIFF
--- a/btcec/schnorr/signature_test.go
+++ b/btcec/schnorr/signature_test.go
@@ -10,8 +10,10 @@ import (
 	"errors"
 	"strings"
 	"testing"
+	"testing/quick"
 
 	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/davecgh/go-spew/spew"
 	secp_ecdsa "github.com/decred/dcrd/dcrec/secp256k1/v4"
 	ecdsa_schnorr "github.com/decred/dcrd/dcrec/secp256k1/v4/schnorr"
 )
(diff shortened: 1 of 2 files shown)


btcec/schnorr/signature_test.go
```diff
@@ -254,3 +256,37 @@ func TestSchnorrVerify(t *testing.T) {
...
+
+// TestSchnorrSignNoMutate tests that generating a schnorr signature doesn't
+// modify/mutate the underlying private key.
+func TestSchnorrSignNoMutate(t *testing.T) {
+	t.Parallel()
+
+	// Assert that given a random private key and message, we can generate
+	// a signature from that w/o modifying the underlying private key.
+	f := func(privBytes, msg [32]byte) bool {
+		privKey, _ := btcec.PrivKeyFromBytes(privBytes[:])
```
kcalvinalvin, line 268: Since we're passing an array as a slice, `PrivKeyFromBytes()` can mutate `privBytes`.

I think doing a
```
copyPrivBytes:= privBytes
...
// We should be able to re-derive the private key from raw
// bytes and have that match up again.
privKeyCopy, _ := btcec.PrivKeyFromBytes(copyPrivBytes[:])
```

would be more prudent since we now guaranteed that the privKey we're using isn't mutated.

yyforyongyu: I think `copyPrivBytes:= privBytes` still copies the pointer? If we want to copy values we need to declare a new slice `copyPrivBytes := make([]byte, len(privBytes))`, then use `copy`. 

Or we can use array instead of slice.

kcalvinalvin: `privBytes` is an array that's being passed as a slice. It's a value not a reference so `:=` will suffice.

I made a runnable example here:
https://go.dev/play/p/B4W-TMo9Xmo

yyforyongyu: oh right, that's already an array. Was thinking maybe `btcec.PrivKeyFromBytes` can take an array, just to be safu.

Crypt-iQ: PrivKeyFromBytes doesn't appear to modify the byte slice since it has length 32

Crypt-iQ (approved):

guggero (approved):
LGTM :tada:

btcec/schnorr/signature_test.go
kcalvinalvin: This change costs almost nothing in terms of performance and guarantees correctness in case of future changes to PrivKeyFromBytes.

No brainer change in my view.

Roasbeef: Added the copy just to make it a bit more explicit what the test is attempting to assert.

Roasbeef:
Thx for the review y'all! Added the intermediate copy just to make things a bit more explicit, I think this is g2g now.

PR #1929, opened 2022-11-27
btcec/schnorr: use private key copy for BIP-340 signatures

Julius080:
This is a fix similar to, the private key itself might need to be negated.

A similar test is added here that fails without the patch to the signature.go file.


guggero:
This was already merged to master with https://github.com/btcsuite/btcd/pull/1919.

PR #1931, opened 2022-11-29
blockchain: refactor and export header validation checks

Crypt-iQ:
This PR allows an external library like `neutrino` to be able to perform contextual and context-free header validation. This is achieved by introducing two new interfaces `HeaderCtx` (which replaces `*blockNode` in the relevant functions) and `ChainCtx` (which replaces `*BlockChain` in the relevant functions). A library calling into these functions needs to only satisfy the interface to be able to validate any received headers.

### DIFF
--- a/blockchain/blockindex.go
+++ b/blockchain/blockindex.go
@@ -169,6 +169,60 @@ func (node *blockNode) Ancestor(height int32) *blockNode {
 	return n
 }
 
+// Height returns the blockNode's height in the chain.
+//
+// NOTE: Part of the HeaderCtx interface.
+func (node *blockNode) Height() int32 {
+	return node.height
+}
+
+// Bits returns the blockNode's nBits.
+//
+// NOTE: Part of the HeaderCtx interface.
+func (node *blockNode) Bits() uint32 {
+	return node.bits
+}
+
+// Timestamp returns the blockNode's timestamp.
+//
+// NOTE: Part of the HeaderCtx interface.
+func (node *blockNode) Timestamp() int64 {
+	return node.timestamp
+}
+
+// Parent returns the blockNode's parent.
+//
+// NOTE: Part of the HeaderCtx interface.
+func (node *blockNode) Parent() HeaderCtx {
+	if node.parent == nil {
+		// This is required since node.parent is a *blockNode and if we
+		// do not explicitly return nil here, the caller may fail when
+		// nil-checking this.
+		return nil
+	}
+
+	return node.parent
+}
+
+// RelativeAncestorCtx returns the blockNode's ancestor that is distance blocks
+// before it in the chain. This is equivalent to the RelativeAncestor function
+// below except that the return type is different.
+//
+// This function is safe for concurrent access.
+//
+// NOTE: Part of the HeaderCtx interface.
+func (node *blockNode) RelativeAncestorCtx(distance int32) HeaderCtx {
+	ancestor := node.RelativeAncestor(distance)
+	if ancestor == nil {
+		// This is required since RelativeAncestor returns a *blockNode
+		// and if we do not explicitly return nil here, the caller may
+		// fail when nil-checking this.
+		return nil
+	}
+
+	return ancestor
+}
+
 // RelativeAncestor returns the ancestor block node a relative 'distance' blocks
 // before this node.  This is equivalent to calling Ancestor with the node's
 // height minus provided distance.
@@ -182,17 +236,17 @@ func (node *blockNode) RelativeAncestor(distance int32) *blockNode {
 // prior to, and including, the block node.
 //
 // This function is safe for concurrent access.
-func (node *blockNode) CalcPastMedianTime() time.Time {
+func CalcPastMedianTime(node HeaderCtx) time.Time {
 	// Create a slice of the previous few block timestamps used to calculate
 	// the median per the number defined by the constant medianTimeBlocks.
 	timestamps := make([]int64, medianTimeBlocks)
 	numNodes := 0
 	iterNode := node
 	for i := 0; i < medianTimeBlocks && iterNode != nil; i++ {
-		timestamps[i] = iterNode.timestamp
+		timestamps[i] = iterNode.Timestamp()
 		numNodes++
 
-		iterNode = iterNode.parent
+		iterNode = iterNode.Parent()
 	}
 
 	// Prune the slice to the actual number of available timestamps which
@@ -216,6 +270,10 @@ func (node *blockNode) CalcPastMedianTime() time.Time {
 	medianTimestamp := timestamps[numNodes/2]
 	return time.Unix(medianTimestamp, 0)
 }
+
+// A compile-time assertion to ensure blockNode implements the HeaderCtx
+// interface.
+var _ HeaderCtx = (*blockNode)(nil)
 
 // blockIndex provides facilities for keeping track of an in-memory index of the
 // block chain.  Although the name block chain suggests a single chain of
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -437,7 +437,7 @@ func (b *BlockChain) calcSequenceLock(node *blockNode, tx *btcutil.Tx, utxoView 
 				prevInputHeight = 0
 			}
 			blockNode := node.Ancestor(prevInputHeight)
-			medianTime := blockNode.CalcPastMedianTime()
+			medianTime := CalcPastMedianTime(blockNode)
 
 			// Time based relative time-locks as defined by BIP 68
 			// have a time granularity of RelativeLockSeconds, so
@@ -595,7 +595,8 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
 	blockSize := uint64(block.MsgBlock().SerializeSize())
 	blockWeight := uint64(GetBlockWeight(block))
 	state := newBestState(node, blockSize, blockWeight, numTxns,
-		curTotalTxns+numTxns, node.CalcPastMedianTime())
+		curTotalTxns+numTxns, CalcPastMedianTime(node),
+	)
 
 	// Atomically insert info into the database.
 	err = b.db.Update(func(dbTx database.Tx) error {
@@ -708,7 +709,7 @@ func (b *BlockChain) disconnectBlock(node *blockNode, block *btcutil.Block, view
 	blockWeight := uint64(GetBlockWeight(prevBlock))
 	newTotalTxns := curTotalTxns - uint64(len(block.MsgBlock().Transactions))
 	state := newBestState(prevNode, blockSize, blockWeight, numTxns,
-		newTotalTxns, prevNode.CalcPastMedianTime())
+		newTotalTxns, CalcPastMedianTime(prevNode))
 
 	err = b.db.Update(func(dbTx database.Tx) error {
 		// Update best block state.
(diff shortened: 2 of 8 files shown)


blockchain/blockindex.go
```diff
@@ -182,17 +236,17 @@ func (node *blockNode) RelativeAncestor(distance int32) *blockNode {
...
 	timestamps := make([]int64, medianTimeBlocks)
 	numNodes := 0
 	iterNode := node
 	for i := 0; i < medianTimeBlocks && iterNode != nil; i++ {
-		timestamps[i] = iterNode.timestamp
+		timestamps[i] = iterNode.Timestamp()
 		numNodes++
 
-		iterNode = iterNode.parent
+		iterNode = iterNode.Parent()
```
guggero, line 249: The `iterNode` is of type `HeaderCtx` which is an interface. So the `!= nil` check in the loop might not work reliably anymore.

blockchain/chain.go
```diff
@@ -594,7 +594,7 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
 	blockSize := uint64(block.MsgBlock().SerializeSize())
 	blockWeight := uint64(GetBlockWeight(block))
 	state := newBestState(node, blockSize, blockWeight, numTxns,
-		curTotalTxns+numTxns, node.CalcPastMedianTime())
+		curTotalTxns+numTxns, CalcPastMedianTime(node))
```
guggero, line 597: nit: should we fix the formatting of this multi-line function call if we're touching the code anyway? Here and below.

blockchain/difficulty.go
```diff
@@ -192,109 +192,110 @@ func (b *BlockChain) calcEasiestDifficulty(bits uint32, duration time.Duration)
...
-//
-// This function MUST be called with the chain state lock held (for writes).
-func (b *BlockChain) findPrevTestNetDifficulty(startNode *blockNode) uint32 {
+func findPrevTestNetDifficulty(startNode HeaderCtx, c ChainCtx) uint32 {
 	// Search backwards through the chain for the last block without
 	// the special rule applied.
 	iterNode := startNode
-	for iterNode != nil && iterNode.height%b.blocksPerRetarget != 0 &&
-		iterNode.bits == b.chainParams.PowLimitBits {
+	for iterNode != nil && iterNode.Height()%c.BlocksPerRetarget() != 0 &&
```
guggero, line 199: Same issue here with the `nil` check of an interface.

blockchain/difficulty.go
```diff
@@ -192,109 +192,110 @@ func (b *BlockChain) calcEasiestDifficulty(bits uint32, duration time.Duration)
...
-// after the passed previous block node based on the difficulty retarget rules.
+// after the passed previous HeaderCtx based on the difficulty retarget rules.
 // This function differs from the exported CalcNextRequiredDifficulty in that
-// the exported version uses the current best chain as the previous block node
-// while this function accepts any block node.
-func (b *BlockChain) calcNextRequiredDifficulty(lastNode *blockNode, newBlockTime time.Time) (uint32, error) {
+// the exported version uses the current best chain as the previous HeaderCtx
+// while this function accepts any block node. This function accepts a ChainCtx
+// parameter that gives the necessary difficulty context variables.
+func calcNextRequiredDifficulty(lastNode HeaderCtx, newBlockTime time.Time,
```
guggero, line 220: nit: fix line length while we're touching this function definition?

blockchain/validate.go
```diff
@@ -685,30 +711,17 @@ func (b *BlockChain) checkBlockHeaderContext(header *wire.BlockHeader, prevNode
 	// chain before it.  This prevents storage of new, otherwise valid,
 	// blocks which build off of old blocks that are likely at a much easier
 	// difficulty and therefore could be used to waste cache and disk space.
-	checkpointNode, err := b.findPreviousCheckpoint()
+	checkpointNode, err := c.FindPreviousCheckpoint()
 	if err != nil {
 		return err
 	}
-	if checkpointNode != nil && blockHeight < checkpointNode.height {
+	if checkpointNode != nil && blockHeight < checkpointNode.Height() {
```
guggero, line 718: Nil check of an interface!

blockchain/validate.go
```diff
@@ -1288,3 +1301,113 @@ func (b *BlockChain) CheckConnectBlockTemplate(block *btcutil.Block) error {
...
+
+	return checkpoint, err
+}
+
+// A compile-time assertion to ensure BlockChain implements the ChainCtx
+// interface.
+var _ ChainCtx = (*BlockChain)(nil)
+
+// ChainCtx is an interface that abstracts away blockchain parameters.
+type ChainCtx interface {
```
guggero, line 1367: nit: should we move the interface definitions to the top of the file?

blockchain/validate.go
```diff
@@ -1288,3 +1301,113 @@ func (b *BlockChain) CheckConnectBlockTemplate(block *btcutil.Block) error {
...
+	// ChainParams returns the chain's configured chaincfg.Params.
+	ChainParams() *chaincfg.Params
+
+	// BlocksPerRetarget returns the number of blocks before retargeting
+	// occurs.
+	BlocksPerRetarget() int32
+
+	// MinRetargetTimespan returns the minimum amount of time to use in the
+	// difficulty calcuation.
```
guggero, line 1376: nit: typo in calculation.

guggero (commented):
Looks good! Only blocking comments are about the nil checks vs. interfaces (need to also read up on the specifics, when a normal nil check works and when not).

Crypt-iQ:
> Looks good! Only blocking comments are about the nil checks vs. interfaces (need to also read up on the specifics, when a normal nil check works and when not).

You can nil-check an interface, but it's a bit weird.

See here: https://go.dev/play/p/9NVSUoFo2oI
If `foo()` returns `innerFoo()` which returns `nil`, then the `nil` is converted to a `MyIface` with a `nil` value. This will fail the nil-check.

If `foo()` just calls `return nil`, then this will succeed the nil-check. I'll double-check the call-sites you mentioned just to be sure though

guggero:
> > Looks good! Only blocking comments are about the nil checks vs. interfaces (need to also read up on the specifics, when a normal nil check works and when not).
> 
> You can nil-check an interface, but it's a bit weird.
> 
> See here: https://go.dev/play/p/9NVSUoFo2oI If `foo()` returns `innerFoo()` which returns `nil`, then the `nil` is converted to a `MyIface` with a `nil` value. This will fail the nil-check.
> 
> If `foo()` just calls `return nil`, then this will succeed the nil-check. I'll double-check the call-sites you mentioned just to be sure though

Aaah, I see. Thanks for the example. I guess that makes sense. That's why you had that explicit `if x == nil { return nil` in there, gotcha.

Crypt-iQ:
I do wonder if there's not a way to get around nil-checking and to instead use some dummy value to compare against

Crypt-iQ (commented):
Left some comments, this pattern is kind of worrying since I feel like I've seen it before in our codebases, but can't exactly grep for it

blockchain/blockindex.go
```diff
@@ -169,6 +169,60 @@ func (node *blockNode) Ancestor(height int32) *blockNode {
...
+// This function is safe for concurrent access.
+//
+// NOTE: Part of the HeaderCtx interface.
+func (node *blockNode) RelativeAncestorCtx(distance int32) HeaderCtx {
+	ancestor := node.RelativeAncestor(distance)
+	if ancestor == nil {
+		// This is required since RelativeAncestor returns a *blockNode
+		// and if we do not explicitly return nil here, the caller may
+		// fail when nil-checking this.
+		return nil
```
Roasbeef, line 220: Ahh good ol nil interface headaches

blockchain/validate.go
```diff
@@ -633,22 +633,29 @@ func checkSerializedHeight(coinbaseTx *btcutil.Tx, wantHeight int32) error {
...
+	flags BehaviorFlags, skipCheckpoint bool, c ChainCtx) error {
+
 	fastAdd := flags&BFFastAdd == BFFastAdd
 	if !fastAdd {
 		// Ensure the difficulty specified in the block header matches
 		// the calculated difficulty based on the previous block and
 		// difficulty retarget rules.
-		expectedDifficulty, err := b.calcNextRequiredDifficulty(prevNode,
-			header.Timestamp)
+		expectedDifficulty, err := calcNextRequiredDifficulty(prevNode,
```
Roasbeef, line 657: Can fix line wrapping here.

blockchain/blockindex.go
Roasbeef: Yep, since we return `nil` explicitly.

blockchain/difficulty.go
Roasbeef: Yep ^

blockchain/validate.go
```diff
@@ -633,22 +633,29 @@ func checkSerializedHeight(coinbaseTx *btcutil.Tx, wantHeight int32) error {
...
 //
+// The skipCheckpoint boolean is used so that libraries can skip the checkpoint
+// sanity checks.
+//
 // This function MUST be called with the chain state lock held (for writes).
-func (b *BlockChain) checkBlockHeaderContext(header *wire.BlockHeader, prevNode *blockNode, flags BehaviorFlags) error {
+// NOTE: Ignore the above lock requirement if this function is not passed a
+// *Blockchain instance as the ChainCtx argument.
+func CheckBlockHeaderContext(header *wire.BlockHeader, prevNode HeaderCtx,
+	flags BehaviorFlags, skipCheckpoint bool, c ChainCtx) error {
```
Roasbeef, line 650: Minor nit re arg ordering here: context first? Similar to the sort of syntactic sugar we get for methods.

Crypt-iQ: fixed

blockchain/validate.go
Crypt-iQ: fixed

blockchain/validate.go
Crypt-iQ: fixed

blockchain/validate.go
Crypt-iQ: moved to new file

blockchain/difficulty.go
Crypt-iQ: think I fixed this

blockchain/chain.go
Crypt-iQ: wasn't sure what below is referencing?

Roasbeef (approved):
LGTM 🌪️

blockchain/validate.go
```diff
@@ -421,13 +421,13 @@ func CountP2SHSigOps(tx *btcutil.Tx, isCoinBaseTx bool, utxoView *UtxoViewpoint)
...
 
-// checkBlockHeaderSanity performs some preliminary checks on a block header to
+// CheckBlockHeaderSanity performs some preliminary checks on a block header to
 // ensure it is sane before continuing with processing.  These checks are
 // context free.
 //
 // The flags do not modify the behavior of this function directly, however they
 // are needed to pass along to checkProofOfWork.
-func checkBlockHeaderSanity(header *wire.BlockHeader, powLimit *big.Int, timeSource MedianTimeSource, flags BehaviorFlags) error {
+func CheckBlockHeaderSanity(header *wire.BlockHeader, powLimit *big.Int, timeSource MedianTimeSource, flags BehaviorFlags) error {
```
guggero, line 430: nit: wrap this too long function as well if we're changing this line?

guggero (approved):
LGTM :tada:

Crypt-iQ:
Changed `FindPreviousCheckpoint` to check and return an error from `findPreviousCheckpoint`

Roasbeef (approved):
LGTM 🧱

PR #1942, opened 2023-01-25
btcutil/psbt: export helper functions, fix/add encoding of unknown fields

guggero:
This PR fixes a bunch of small issues that allow the `btcutil/psbt` package to be used for adding custom fields (=`[]*Unknown`) to the transaction itself (global) as well as the outputs. We also export the BIP32 derivation info serialization/de-serialization related functions for easy re-use in other projects.

cc @ffranr.


btcutil/psbt/partial_output.go
```diff
@@ -28,6 +29,7 @@ func NewPsbtOutput(redeemScript []byte, witnessScript []byte,
 		RedeemScript:    redeemScript,
 		WitnessScript:   witnessScript,
 		Bip32Derivation: bip32Derivation,
+		Unknowns:        nil,
```
Roasbeef, line 32: No need to explicitly declare a nil pointer here.

btcutil/psbt/partial_output.go
```diff
@@ -144,8 +146,23 @@ func (po *POutput) deserialize(r io.Reader) error {
 			)
 
 		default:
-			// Unknown type is allowed for inputs but not outputs.
-			return ErrInvalidPsbtFormat
```
Roasbeef, line 146: Was this a BIP level thing? Or just a simplification that was made during the initial implementation as we didn't have a use case at that time?

btcutil/psbt/partial_output.go
```diff
@@ -144,8 +146,23 @@ func (po *POutput) deserialize(r io.Reader) error {
...
+			// A fall through case for any proprietary types.
+			keyintanddata := []byte{byte(keyint)}
```
Roasbeef, line 150: Should use camel case for `keyintanddata` and the other variables.

btcutil/psbt/psbt.go
```diff
@@ -129,7 +129,7 @@ type Packet struct {
 	Outputs []POutput
 
 	// Unknowns are the set of custom types (global only) within this PSBT.
-	Unknowns []Unknown
+	Unknowns []*Unknown
```
Roasbeef, line 132: Rationale behind the switch?

btcutil/psbt/partial_output.go
guggero: I think it was just forgotten or ignored during the initial implementation. I added a unit as a last commit to make sure encoding/decoding of the unknown fields works across all levels.

btcutil/psbt/partial_output.go
guggero: Oops, copy/paste. Fixed it (and some other formatting issues) in the existing code too.

btcutil/psbt/psbt.go
guggero: Mostly to unify the three levels where unknown fields are used. That way we can have more generic handler functions for those fields in the code that requires these changes. Added the rationale to the commit message as well.

Roasbeef (approved):
LGTM 👑

Roasbeef:
cc @sputn1ck

sputn1ck (approved):
LGTM

PR #1951, opened 2023-02-21
chainhash: JSON Unmarshal hash from string

RycCheen:
chainhash: JSON Unmarshal hash from appropriate string.
This commit(https://github.com/btcsuite/btcd/commit/1d6e578c2180a6ee45755270dd938894fb93d661) occurred a problem:
"error": "json: cannot unmarshal string into Go struct field OutPoint.TxIn.PreviousOutPoint.Hash of type chainhash.Hash"

fix it.

### DIFF
--- a/chaincfg/chainhash/hash.go
+++ b/chaincfg/chainhash/hash.go
@@ -116,6 +116,22 @@ func (hash Hash) MarshalJSON() ([]byte, error) {
 	return json.Marshal(hash.String())
 }
 
+// UnmarshalJSON parse the hash with JSON appropriate string value.
+func (hash *Hash) UnmarshalJSON(input []byte) error {
+    var sh string
+    err := json.Unmarshal(input, &sh)
+    if err != nil {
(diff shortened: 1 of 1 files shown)


chaincfg/chainhash/hash.go
```diff
@@ -116,6 +116,22 @@ func (hash Hash) MarshalJSON() ([]byte, error) {
 	return json.Marshal(hash.String())
 }
 
+// UnmarshalJSON parse the hash with JSON appropriate string value.
```
guggero, line 119: nit: s/parse/parses/

guggero (commented):
Thanks for the fix. Where exactly did you encounter the problem? When calling what method/function?

Can you please rebase to get rid of the merge commit?

RycCheen:
> 
In our business scenario, we need to serialize the wire. MsgTx structure.
This error occurred in the process of json.Unmarshal

PR #1953, opened 2023-02-23
btcd: Add memory profiling flag

kcalvinalvin:
Enables Go memory profiling.  If the cpuprofile shows a lot of time spent on gc, it's useful to then do a memory profile to see where the memory alloctions happen.

Unlike the --profile flag, this allows for easy generation of a memory profile for the entire duration of which btcd has been running for in various readble graphs.

### DIFF
--- a/btcd.go
+++ b/btcd.go
@@ -88,6 +88,18 @@ func btcdMain(serverChan chan<- *server) error {
 		defer pprof.StopCPUProfile()
 	}
 
+	// Write mem profile if requested.
+	if cfg.MemoryProfile != "" {
+		f, err := os.Create(cfg.MemoryProfile)
+		if err != nil {
+			btcdLog.Errorf("Unable to create memory profile: %v", err)
+			return err
+		}
+		defer f.Close()
+		defer pprof.WriteHeapProfile(f)
+		defer runtime.GC()
+	}
+
 	// Perform upgrades to btcd as new versions require it.
 	if err := doUpgrades(); err != nil {
 		btcdLog.Errorf("%v", err)
--- a/config.go
+++ b/config.go
@@ -111,6 +111,7 @@ type config struct {
 	ConfigFile           string        `short:"C" long:"configfile" description:"Path to configuration file"`
 	ConnectPeers         []string      `long:"connect" description:"Connect only to the specified peers at startup"`
 	CPUProfile           string        `long:"cpuprofile" description:"Write CPU profile to the specified file"`
+	MemoryProfile        string        `long:"memprofile" description:"Write memory profile to the specified file"`
 	DataDir              string        `short:"b" long:"datadir" description:"Directory to store data"`
 	DbType               string        `long:"dbtype" description:"Database backend to use for the Block Chain"`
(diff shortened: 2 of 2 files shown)


btcd.go
```diff
@@ -88,6 +88,17 @@ func btcdMain(serverChan chan<- *server) error {
...
 
+	// Write mem profile if requested.
+	if cfg.MemoryProfile != "" {
+		f, err := os.Create(cfg.MemoryProfile)
+		if err != nil {
+			btcdLog.Errorf("Unable to create memory profile: %v", err)
+			return err
+		}
+		defer f.Close()
+		defer pprof.WriteHeapProfile(f)
```
ellemouton, line 99: since the file is only ever written to on shutdown, can't this whole if statement be put in a `defer`? that way we also dont need to keep the file open for the lifetime of the binary

kcalvinalvin: I think it'd be nice if it errors out immediately if it's not able to create a file to write to.

I'm not sure on the costs of having a file descriptor for the duration of the program but if that's a concern, how about something like this:

```
        // Write mem profile if requested.
        if cfg.MemoryProfile != "" {
                f, err := os.Create(cfg.MemoryProfile)
                if err != nil {
                        btcdLog.Errorf("Unable to create memory profile: %v", err)
                        return err
                }
                f.Close()

                defer func() {
                        f, err := os.Create(cfg.MemoryProfile)
                        if err != nil {
                                btcdLog.Infof("Unable to create memory profile: %v", err)
                        }
                        err = pprof.WriteHeapProfile(f)
                        if err != nil {
                                btcdLog.Infof("Unable to create memory profile: %v", err)
                        }
                        f.Close()
                }()
        }
```

ellemouton: good point regarding failing fast on start up 👍  in that case, if the cost of the file descriptor is low then your original solution seems good enough!

Roasbeef (approved):
LGTM 🦕

Roasbeef:
cc @Crypt-iQ

btcd.go
Crypt-iQ: I think the cost is pretty low

btcd.go
```diff
@@ -88,6 +88,17 @@ func btcdMain(serverChan chan<- *server) error {
...
+	}
```
Crypt-iQ, line 100: could add a `defer runtime.GC()` call at the end so that the heap profile is accurate as far as what's actually lingering

Crypt-iQ (commented):
just a comment about running the gc

btcd.go
kcalvinalvin: Ok I'll keep the current implementation.

btcd.go
kcalvinalvin: @Roasbeef any opinions on this? It'll give us better data but it may slow things down during shutdown for normal users.

Roasbeef: IIUC, that would only be active for users already running with a flag that strictly reserved for profiling. So I think that's fine.

kcalvinalvin: Added a runtime.GC() so that garbage collection happens before writes

halseth (approved):
LGTM

PR #1954, opened 2023-02-27
txscript: allow script builder capacity to be specified

guggero:
The default script builder allocates 500 bytes of memory for each script. That is quite large for most applications and therefore wastes a lot of memory when only small scripts are created. To avoid breaking backward compatibility, we add a new functional option to the NewScriptBuilder function that allows the user to specify the initial allocation size.

cc @Roasbeef, @Crypt-iQ.

### DIFF
--- a/txscript/scriptbuilder_test.go
+++ b/txscript/scriptbuilder_test.go
@@ -7,7 +7,37 @@ package txscript
 import (
 	"bytes"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
+
+// TestScriptBuilderAlloc tests that the pre-allocation for a script via the
+// NewScriptBuilder function works as expected.
+func TestScriptBuilderAlloc(t *testing.T) {
+	// Using the default value, we should get a script with a capacity of
+	// 500 bytes, which is quite large for most scripts.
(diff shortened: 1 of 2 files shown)


txscript/scriptbuilder_test.go
```diff
@@ -7,8 +7,38 @@ package txscript
...
 	"bytes"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
 
+// TestScriptBuilderAlloc tests that the pre-allocation for a script via the
+// NewScriptBuilder function works as expected.
+func TestScriptBuilderAlloc(t *testing.T) {
+	// Using the default value, we should get a script with a capacity of
```
Roasbeef, line 17: Might be my monitor, but looks to slightly run over line length.

Roasbeef (approved):
LGTM 🦖

Crypt-iQ (approved):
nice optimization 🔬

txscript/scriptbuilder_test.go
guggero: No, it's actually 3 characters short of the full 80 :sweat_smile:

PR #1955, opened 2023-03-01
blockchain, cmd, netsync, main: Add utxocache

kcalvinalvin:
Updated the PR and is now very different from #1373.

Main differences are:

## No utxoviewpoint middleman during headers-first sync

In #1373, the view is still used as a middleman. The flow for doing block validation would be:
```
1. Make a new UtxoViewPoint.
2. Load the relevant inputs into the UtxoViewPoint from the cache.
3. Connect transactions in the UtxoViewPoint
4. Apply the connections to the cache from the UtxoViewPoint
5. Flush the cache to the database when needed
```
The problems with these approach were:
(2) is extremely expensive as there a lot of memory allocated by creating a whole new set of key/value pairs of the existing UTXOs.
(4) introduces a new code path from [view -> cache] and in turn introduces a whole new set of bugs.

The new code is different where the flow during headers-first (blocks up until the checkpoint) is:
```
1. Connect transactions in the cache
2. Flush the cache to the database when needed
```
This flow is much simpler and faster than the previous code.

For when the node is not in headers-first mode:
```
1. Make a new UtxoViewPoint
2. Load the relevant inputs into the UtxoViewPoint from the cache.
3. CheckConnectBlock with the UtxoViewPoint
4. Connect transactions in the cache
5. Flush the cache to the database when needed
```
This flow still has the overhead of (2) but no longer has the new bugs introduced by applying the transaction connections from [view -> cache]. Instead, the view is only used as a temporary mutable map **just for checkConnectBlock** and the txs are connected in cache just like in headers-first mode.

## No use of the cache during reorganizations

During reorganizations, a whole new possible bug was introduced where if the node is unexpectedly shutdown, the disconnected txs may only have been updated in the cache and not in the database. The new code forces a flush *before* the reorganization happens and keeps the same code as before, eliminating the new case of bugs introduced in #1373.

This allowed the removal of `TestUtxoCache_Reorg` and all the associated (and duplicated) code.

## No more utxoStateConsistency struct

It was enough to just save the block hash and not the single byte indicating if the cache is consistent or not. All the related code was removed.

## Map slice instead of a singular map

The singular map doubles in size when the map gets full. For bigger maps, this may result in us going over the promised memory limit. The mapslice allows us to allocate multiple maps with differing sizes, allowing acute control of the memory allocated memory.

Another solution to this would be to just force the user to only allow fixed amounts of memory for the utxo cache (1GB, 2GB, 4GB, 8GB, etc). This would get rid of the code in `sizehelper.go` and mapslice related code in `utxocache.go` but wouldn't allow the user flexibility. I'm ok with either route. This would get rid of 491 lines of code from being added.

## Known issue

During a flush, there's a large allocation of memory due to the fact that `database/ffldb` will hold onto every single bucket write to guarantee atomicity. It'll usually result in 2x the memory from what the cache is using. If the cache is using up 2GB, ffldb will allocate 2GB as well, resulting in 4GB of memory being taken up by btcd. I don't think there's any easy way around this other than making a different code path for utxo cache flushes.

### DIFF
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -642,6 +642,24 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
 				if err != nil {
 					return err
 				}
+
+				// We may need to flush if the prune will delete blocks that
+				// are past our last flush block.
+				//
+				// NOTE: the database will never be inconsistent here as the
+				// actual blocks are not deleted until the db.Update returns.
+				needsFlush, err := b.flushNeededAfterPrune(deletedHashes)
+				if err != nil {
+					return err
+				}
+				if needsFlush {
+					// Since the deleted hashes are past our last
+					// flush block, flush the utxo cache now.
+					err = b.utxoCache.flush(dbTx, FlushRequired, state)
+					if err != nil {
+						return err
+					}
+				}
 			}
 		}
 
@@ -719,7 +737,7 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
 	// Since we may have changed the UTXO cache, we make sure it didn't exceed its
 	// maximum size.  If we're pruned and have flushed already, this will be a no-op.
 	return b.db.Update(func(dbTx database.Tx) error {
-		return b.utxoCache.flush(dbTx, FlushIfNeeded, b.BestSnapshot())
+		return b.utxoCache.flush(dbTx, FlushIfNeeded, state)
 	})
 }
 
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -131,6 +131,10 @@ type BlockChain struct {
 	index     *blockIndex
 	bestChain *chainView
 
+	// The UTXO state holds a cached view of the UTXO state of the chain.
+	// It is protected by the chain lock.
+	utxoCache *utxoCache
+
 	// These fields are related to handling of orphan blocks.  They are
 	// protected by a combination of the chain lock and the orphan lock.
 	orphanLock   sync.RWMutex
@@ -551,9 +555,14 @@ func (b *BlockChain) getReorganizeNodes(node *blockNode) (*list.List, *list.List
 // connectBlock handles connecting the passed node/block to the end of the main
 // (best) chain.
 //
-// This passed utxo view must have all referenced txos the block spends marked
-// as spent and all of the new txos the block creates added to it.  In addition,
-// the passed stxos slice must be populated with all of the information for the
+// Passing in a utxo view is optional.  If the passed in utxo view is nil,
+// connectBlock will assume that the utxo cache has already connected all the
+// txs in the block being connected.
+// If a utxo view is passed in, this passed utxo view must have all referenced
+// txos the block spends marked as spent and all of the new txos the block creates
+// added to it.
+//
+// The passed stxos slice must be populated with all of the information for the
 // spent txos.  This approach is used because the connection validation that
 // must happen prior to calling this function requires the same details, so
 // it would be inefficient to repeat it.
@@ -602,6 +611,18 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
 		curTotalTxns+numTxns, CalcPastMedianTime(node),
 	)
 
+	// If a utxoviewpoint was passed in, we'll be writing that viewpoint
+	// directly to the database on disk.  In order for the database to be
+	// consistent, we must flush the cache before writing the viewpoint.
+	if view != nil {
+		err = b.db.Update(func(dbTx database.Tx) error {
+			return b.utxoCache.flush(dbTx, FlushRequired, state)
+		})
+		if err != nil {
+			return err
+		}
+	}
+
 	// Atomically insert info into the database.
 	err = b.db.Update(func(dbTx database.Tx) error {
 		// If the pruneTarget isn't 0, we should attempt to delete older blocks
@@ -640,6 +661,8 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
 		// Update the utxo set using the state of the utxo view.  This
 		// entails removing all of the utxos spent and adding the new
 		// ones created by the block.
+		//
+		// A nil viewpoint is a no-op.
 		err = dbPutUtxoView(dbTx, view)
 		if err != nil {
 			return err
@@ -670,7 +693,9 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
 
 	// Prune fully spent entries and mark all entries in the view unmodified
 	// now that the modifications have been committed to the database.
-	view.commit()
+	if view != nil {
+		view.commit()
+	}
 
 	// This node is now the end of the best chain.
 	b.bestChain.SetTip(node)
@@ -691,7 +716,11 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
 	b.sendNotification(NTBlockConnected, block)
 	b.chainLock.Lock()
 
-	return nil
+	// Since we may have changed the UTXO cache, we make sure it didn't exceed its
+	// maximum size.  If we're pruned and have flushed already, this will be a no-op.
+	return b.db.Update(func(dbTx database.Tx) error {
+		return b.utxoCache.flush(dbTx, FlushIfNeeded, b.BestSnapshot())
+	})
 }
 
 // disconnectBlock handles disconnecting the passed node/block from the end of
@@ -840,6 +869,15 @@ func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List) error 
 		return nil
 	}
 
+	// The rest of the reorg depends on all STXOs already being in the database
+	// so we flush before reorg.
+	err := b.db.Update(func(dbTx database.Tx) error {
+		return b.utxoCache.flush(dbTx, FlushRequired, b.BestSnapshot())
+	})
+	if err != nil {
+		return err
+	}
+
 	// Ensure the provided nodes match the current best chain.
 	tip := b.bestChain.Tip()
 	if detachNodes.Len() != 0 {
@@ -901,7 +939,7 @@ func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List) error 
 
 		// Load all of the utxos referenced by the block that aren't
 		// already in the view.
-		err = view.fetchInputUtxos(b.db, block)
+		err = view.fetchInputUtxos(b.db, nil, block)
 		if err != nil {
 			return err
 		}
@@ -968,7 +1006,7 @@ func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List) error 
 		// checkConnectBlock gets skipped, we still need to update the UTXO
 		// view.
 		if b.index.NodeStatus(n).KnownValid() {
-			err = view.fetchInputUtxos(b.db, block)
+			err = view.fetchInputUtxos(b.db, nil, block)
 			if err != nil {
 				return err
 			}
@@ -1020,7 +1058,7 @@ func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List) error 
 
 		// Load all of the utxos referenced by the block that aren't
 		// already in the view.
-		err := view.fetchInputUtxos(b.db, block)
+		err := view.fetchInputUtxos(b.db, nil, block)
 		if err != nil {
 			return err
 		}
@@ -1047,7 +1085,7 @@ func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List) error 
 
 		// Load all of the utxos referenced by the block that aren't
 		// already in the view.
-		err := view.fetchInputUtxos(b.db, block)
+		err := view.fetchInputUtxos(b.db, nil, block)
 		if err != nil {
 			return err
 		}
@@ -1067,6 +1105,15 @@ func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List) error 
 		if err != nil {
 			return err
 		}
+	}
+
+	// We call the flush at the end to update the last flush hash to the new
+	// best tip.
+	err = b.db.Update(func(dbTx database.Tx) error {
+		return b.utxoCache.flush(dbTx, FlushRequired, b.BestSnapshot())
+	})
+	if err != nil {
+		return err
 	}
 
 	// Log the point where the chain forked and old and new best chain
@@ -1121,11 +1168,21 @@ func (b *BlockChain) connectBestChain(node *blockNode, block *btcutil.Block, fla
 		// Perform several checks to verify the block can be connected
 		// to the main chain without violating any rules and without
 		// actually connecting the block.
-		view := NewUtxoViewpoint()
-		view.SetBestHash(parentHash)
-		stxos := make([]SpentTxOut, 0, countSpentOutputs(block))
 		if !fastAdd {
-			err := b.checkConnectBlock(node, block, view, &stxos)
+			// We create a viewpoint here to avoid spending or adding new
+			// coins to the utxo cache.
+			//
+			// checkConnectBlock spends and adds utxos before doing the
+			// signature validation and if the signature validation fails,
+			// we would be forced to undo the utxo cache.
+			//
+			// TODO (kcalvinalvin): Doing all of the validation before connecting
+			// the tx inside check connect block would allow us to pass the utxo
+			// cache directly to the check connect block.  This would save on the
+			// expensive memory allocation done by fetch input utxos.
+			view := NewUtxoViewpoint()
+			view.SetBestHash(parentHash)
+			err := b.checkConnectBlock(node, block, view, nil)
 			if err == nil {
 				b.index.SetStatusFlags(node, statusValid)
 			} else if _, ok := err.(RuleError); ok {
@@ -1141,23 +1198,16 @@ func (b *BlockChain) connectBestChain(node *blockNode, block *btcutil.Block, fla
 			}
 		}
 
-		// In the fast add case the code to check the block connection
-		// was skipped, so the utxo view needs to load the referenced
-		// utxos, spend them, and add the new utxos being created by
-		// this block.
-		if fastAdd {
-			err := view.fetchInputUtxos(b.db, block)
-			if err != nil {
-				return false, err
-			}
-			err = view.connectTransactions(block, &stxos)
-			if err != nil {
-				return false, err
-			}
+		// Connect the transactions to the cache.  All the txs are considered valid
+		// at this point as they have passed validation or was considered valid already.
+		stxos := make([]SpentTxOut, 0, countSpentOutputs(block))
+		err := b.utxoCache.connectTransactions(block, &stxos)
+		if err != nil {
+			return false, err
 		}
 
 		// Connect the block to the main chain.
-		err := b.connectBlock(node, block, view, stxos)
+		err = b.connectBlock(node, block, nil, stxos)
 		if err != nil {
 			// If we got hit with a rule error, then we'll mark
 			// that status of the block as invalid and flush the
@@ -1785,6 +1835,11 @@ type Config struct {
 	// This field is required.
 	DB database.DB
 
+	// The maximum size in bytes of the UTXO cache.
+	//
+	// This field is required.
+	UtxoCacheMaxSize uint64
+
 	// Interrupt specifies a channel the caller can close to signal that
 	// long running operations, such as catching up indexes or performing
 	// database migrations, should be interrupted.
@@ -1893,6 +1948,7 @@ func New(config *Config) (*BlockChain, error) {
 		maxRetargetTimespan: targetTimespan * adjustmentFactor,
 		blocksPerRetarget:   int32(targetTimespan / targetTimePerBlock),
 		index:               newBlockIndex(config.DB, params),
+		utxoCache:           newUtxoCache(config.DB, config.UtxoCacheMaxSize),
 		hashCache:           config.HashCache,
 		bestChain:           newChainView(nil),
 		orphans:             make(map[chainhash.Hash]*orphanBlock),
@@ -1942,10 +1998,23 @@ func New(config *Config) (*BlockChain, error) {
 		return nil, err
 	}
 
+	// Make sure the utxo state is catched up if it was left in an inconsistent
+	// state.
 	bestNode := b.bestChain.Tip()
+	if err := b.InitConsistentState(bestNode, config.Interrupt); err != nil {
+		return nil, err
+	}
 	log.Infof("Chain state (height %d, hash %v, totaltx %d, work %v)",
 		bestNode.height, bestNode.hash, b.stateSnapshot.TotalTxns,
 		bestNode.workSum)
 
 	return &b, nil
+}
+
+// CachedStateSize returns the total size of the cached state of the blockchain
+// in bytes.
+func (b *BlockChain) CachedStateSize() uint64 {
+	b.chainLock.Lock()
+	defer b.chainLock.Unlock()
+	return b.utxoCache.totalMemoryUsage()
 }
--- a/blockchain/sizehelper.go
+++ b/blockchain/sizehelper.go
@@ -1,0 +1,221 @@
+// Copyright (c) 2023 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+package blockchain
+
+import (
+	"math"
+)
+
+// These constants are related to bitcoin.
+const (
+	// outpointSize is the size of an outpoint.
+	//
+	// This value is calculated by running the following:
+	//	unsafe.Sizeof(wire.OutPoint{})
+	outpointSize = 36
+
+	// uint64Size is the size of an uint64 allocated in memory.
+	uint64Size = 8
+
+	// bucketSize is the size of the bucket in the cache map.  Exact
+	// calculation is (16 + keysize*8 + valuesize*8) where for the map of:
+	// map[wire.OutPoint]*UtxoEntry would have a keysize=36 and valuesize=8.
+	//
+	// https://github.com/golang/go/issues/34561#issuecomment-536115805
+	bucketSize = 16 + uint64Size*outpointSize + uint64Size*uint64Size
+
+	// This value is calculated by running the following on a 64-bit system:
+	//   unsafe.Sizeof(UtxoEntry{})
+	baseEntrySize = 40
+
+	// pubKeyHashLen is the length of a P2PKH script.
+	pubKeyHashLen = 25
+
+	// avgEntrySize is how much each entry we expect it to be.  Since most
+	// txs are p2pkh, we can assume the entry to be more or less the size
+	// of a p2pkh tx.  We add on 7 to make it 32 since 64 bit systems will
+	// align by 8 bytes.
+	avgEntrySize = baseEntrySize + (pubKeyHashLen + 7)
+)
+
+// The code here is shamelessely taken from the go runtime package.  All the relevant
+// code and variables are copied to here.  These values are only correct for a 64 bit
+// system.
+
+const (
+	_MaxSmallSize   = 32768
+	smallSizeDiv    = 8
+	smallSizeMax    = 1024
+	largeSizeDiv    = 128
+	_NumSizeClasses = 68
+	_PageShift      = 13
+	_PageSize       = 1 << _PageShift
+
+	MaxUintptr = ^uintptr(0)
+
+	// Maximum number of key/elem pairs a bucket can hold.
+	bucketCntBits = 3
+	bucketCnt     = 1 << bucketCntBits
+
+	// Maximum average load of a bucket that triggers growth is 6.5.
+	// Represent as loadFactorNum/loadFactorDen, to allow integer math.
+	loadFactorNum = 13
+	loadFactorDen = 2
+
+	// maxAlloc is the maximum size of an allocation. On 64-bit,
+	// it's theoretically possible to allocate 1<<heapAddrBits bytes. On
+	// 32-bit, however, this is one less than 1<<32 because the
+	// number of bytes in the address space doesn't actually fit
+	// in a uintptr.
+	//
+	// NOTE (kcalvinalvin): I just took the constant for a 64 bit system.
+	maxAlloc = 281474976710656
+)
+
+var class_to_size = [_NumSizeClasses]uint16{0, 8, 16, 24, 32, 48, 64, 80, 96, 112, 128, 144, 160, 176, 192, 208, 224, 240, 256, 288, 320, 352, 384, 416, 448, 480, 512, 576, 640, 704, 768, 896, 1024, 1152, 1280, 1408, 1536, 1792, 2048, 2304, 2688, 3072, 3200, 3456, 4096, 4864, 5376, 6144, 6528, 6784, 6912, 8192, 9472, 9728, 10240, 10880, 12288, 13568, 14336, 16384, 18432, 19072, 20480, 21760, 24576, 27264, 28672, 32768}
+var size_to_class8 = [smallSizeMax/smallSizeDiv + 1]uint8{0, 1, 2, 3, 4, 5, 5, 6, 6, 7, 7, 8, 8, 9, 9, 10, 10, 11, 11, 12, 12, 13, 13, 14, 14, 15, 15, 16, 16, 17, 17, 18, 18, 19, 19, 19, 19, 20, 20, 20, 20, 21, 21, 21, 21, 22, 22, 22, 22, 23, 23, 23, 23, 24, 24, 24, 24, 25, 25, 25, 25, 26, 26, 26, 26, 27, 27, 27, 27, 27, 27, 27, 27, 28, 28, 28, 28, 28, 28, 28, 28, 29, 29, 29, 29, 29, 29, 29, 29, 30, 30, 30, 30, 30, 30, 30, 30, 31, 31, 31, 31, 31, 31, 31, 31, 31, 31, 31, 31, 31, 31, 31, 31, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32, 32}
+var size_to_class128 = [(_MaxSmallSize-smallSizeMax)/largeSizeDiv + 1]uint8{32, 33, 34, 35, 36, 37, 37, 38, 38, 39, 39, 40, 40, 40, 41, 41, 41, 42, 43, 43, 44, 44, 44, 44, 44, 45, 45, 45, 45, 45, 45, 46, 46, 46, 46, 47, 47, 47, 47, 47, 47, 48, 48, 48, 49, 49, 50, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 52, 52, 52, 52, 52, 52, 52, 52, 52, 52, 53, 53, 54, 54, 54, 54, 55, 55, 55, 55, 55, 56, 56, 56, 56, 56, 56, 56, 56, 56, 56, 56, 57, 57, 57, 57, 57, 57, 57, 57, 57, 57, 58, 58, 58, 58, 58, 58, 59, 59, 59, 59, 59, 59, 59, 59, 59, 59, 59, 59, 59, 59, 59, 59, 60, 60, 60, 60, 60, 60, 60, 60, 60, 60, 60, 60, 60, 60, 60, 60, 61, 61, 61, 61, 61, 62, 62, 62, 62, 62, 62, 62, 62, 62, 62, 62, 63, 63, 63, 63, 63, 63, 63, 63, 63, 63, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 65, 65, 65, 65, 65, 65, 65, 65, 65, 65, 65, 65, 65, 65, 65, 65, 65, 65, 65, 65, 65, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67, 67}
+
+// calculateRoughMapSize returns a close enough estimate of the
+// total memory allocated by a map.
+// hint should be the same value as the number you give when
+// making a map with the following syntax: make(map[k]v, hint)
+//
+// bucketsize is (16 + keysize*8 + valuesize*8).  For a map of:
+// map[int64]int64, keysize=8 and valuesize=8.  There are edge cases
+// where the bucket size is different that I can't find the source code
+// for. https://github.com/golang/go/issues/34561#issuecomment-536115805
+//
+// I suspect it's because of alignment and how the compiler handles it but
+// when compared with how much the compiler allocates, it's a couple hundred
+// bytes off.
+func calculateRoughMapSize(hint int, bucketSize uintptr) int {
+	// This code is copied from makemap() in runtime/map.go.
+	//
+	// TODO check once in a while to see if this algorithm gets
+	// changed.
+	mem, overflow := mulUintptr(uintptr(hint), uintptr(bucketSize))
+	if overflow || mem > maxAlloc {
+		hint = 0
+	}
+
+	// Find the size parameter B which will hold the requested # of elements.
+	// For hint < 0 overLoadFactor returns false since hint < bucketCnt.
+	B := uint8(0)
+	for overLoadFactor(hint, B) {
+		B++
+	}
+
+	// This code is copied from makeBucketArray() in runtime/map.go.
+	//
+	// TODO check once in a while to see if this algorithm gets
+	// changed.
+	//
+	// For small b, overflow buckets are unlikely.
+	// Avoid the overhead of the calculation.
+	base := bucketShift(B)
+	numBuckets := base
+	if B >= 4 {
+		// Add on the estimated number of overflow buckets
+		// required to insert the median number of elements
+		// used with this value of b.
+		numBuckets += bucketShift(B - 4)
+		sz := bucketSize * numBuckets
+		up := roundupsize(sz)
+		if up != sz {
+			numBuckets = up / bucketSize
+		}
+	}
+	total, _ := mulUintptr(bucketSize, numBuckets)
+
+	if base != numBuckets {
+		// Add 24 for mapextra struct overhead. Refer to
+		// runtime/map.go in the std library for the struct.
+		total += 24
+	}
+
+	// 48 is the number of bytes needed for the map header in a
+	// 64 bit system. Refer to hmap in runtime/map.go in the go
+	// standard library.
+	total += 48
+	return int(total)
+}
+
+// calculateMinEntries returns the minimum number of entries that will make the
+// map allocate the given total bytes.  -1 on the returned entry count will
+// make the map allocate half as much total bytes (for returned entry count that's
+// greater than 0).
+func calculateMinEntries(totalBytes int, bucketSize int) int {
+	// 48 is the number of bytes needed for the map header in a
+	// 64 bit system. Refer to hmap in runtime/map.go in the go
+	// standard library.
+	totalBytes -= 48
+
+	numBuckets := totalBytes / bucketSize
+	B := uint8(math.Log2(float64(numBuckets)))
+	if B < 4 {
+		switch B {
+		case 0:
+			return 0
+		case 1:
+			return 9
+		case 2:
+			return 14
+		default:
+			return 27
+		}
+	}
+
+	B -= 1
+
+	return (int(loadFactorNum * (bucketShift(B) / loadFactorDen))) + 1
+}
+
+// mulUintptr returns a * b and whether the multiplication overflowed.
+// On supported platforms this is an intrinsic lowered by the compiler.
+func mulUintptr(a, b uintptr) (uintptr, bool) {
+	if a|b < 1<<(4*8) || a == 0 {
+		return a * b, false
+	}
+	overflow := b > MaxUintptr/a
+	return a * b, overflow
+}
+
+// divRoundUp returns ceil(n / a).
+func divRoundUp(n, a uintptr) uintptr {
+	// a is generally a power of two. This will get inlined and
+	// the compiler will optimize the division.
+	return (n + a - 1) / a
+}
+
+// alignUp rounds n up to a multiple of a. a must be a power of 2.
+func alignUp(n, a uintptr) uintptr {
+	return (n + a - 1) &^ (a - 1)
+}
+
+// Returns size of the memory block that mallocgc will allocate if you ask for the size.
+func roundupsize(size uintptr) uintptr {
+	if size < _MaxSmallSize {
+		if size <= smallSizeMax-8 {
+			return uintptr(class_to_size[size_to_class8[divRoundUp(size, smallSizeDiv)]])
+		} else {
+			return uintptr(class_to_size[size_to_class128[divRoundUp(size-smallSizeMax, largeSizeDiv)]])
+		}
+	}
+	if size+_PageSize < size {
+		return size
+	}
+	return alignUp(size, _PageSize)
+}
+
+// overLoadFactor reports whether count items placed in 1<<B buckets is over loadFactor.
+func overLoadFactor(count int, B uint8) bool {
+	return count > bucketCnt && uintptr(count) > loadFactorNum*(bucketShift(B)/loadFactorDen)
+}
+
+// bucketShift returns 1<<b, optimized for code generation.
+func bucketShift(b uint8) uintptr {
+	// Masking the shift amount allows overflow checks to be elided.
+	return uintptr(1) << (b & (8*8 - 1))
+}
--- a/blockchain/sizehelper_test.go
+++ b/blockchain/sizehelper_test.go
@@ -1,0 +1,71 @@
+// Copyright (c) 2023 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+package blockchain
+
+import (
+	"math"
+	"testing"
+)
+
+// calculateEntries returns a number of entries that will make the map allocate
+// the given total bytes.  The returned number is always the maximum number of
+// entries that will allocate inside the given parameters.
+func calculateEntries(totalBytes int, bucketSize int) int {
+	// 48 is the number of bytes needed for the map header in a
+	// 64 bit system. Refer to hmap in runtime/map.go in the go
+	// standard library.
+	totalBytes -= 48
+
+	numBuckets := totalBytes / bucketSize
+	B := uint8(math.Log2(float64(numBuckets)))
+	if B == 0 {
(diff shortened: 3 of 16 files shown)


Roasbeef:
Giving this a shot for a mainnet sync, will report back if it gets to a higher height than referenced above (I think the past I ran into an issue at that height too). 

Running with the following config:
```
 GOGC=50 btcd --txindex --nopeerbloomfilters --maxpeers=8 --utxocachemaxsize=10000
```

Roasbeef:
Had to restart my sync as had some memory pressure limits incorrectly configured (though surprised it was killed given the box has 32 GB of RAM and was running with a 10 GB cache size). 

I also think the current cache size estimate might be significantly undercounting the _true_ over head. Will do some profiles to dig into things.

Roasbeef:
Heap profile after around restart with 3 GB of cache used (note how `newUtxoCache` seems to report double that): 

Screenshot 2023-03-08 at 11 37 02 AM

Aside from cache, can see that memory building up for: cloning the cache entries, and also just decoding transactions themselves. Re tx decoding, we have some PRs up that attempt to reduce garbage during that process, and also cache the serialization to make hashing faster (but this is still in check pointed land), but ofc one task at a time. 

Logs of cache size+speed: 
```
2023-03-08 11:35:47.871 [INF] SYNC: Processed 335 blocks in the last 10.05s (264489 transactions, height 339095, 2015-01-15 11:34:26 -0800 PST, ~2751 MiB cache)
2023-03-08 11:35:57.873 [INF] SYNC: Processed 436 blocks in the last 10s (284726 transactions, height 339531, 2015-01-18 11:31:10 -0800 PST, ~2816 MiB cache)
2023-03-08 11:36:07.877 [INF] SYNC: Processed 428 blocks in the last 10s (304451 transactions, height 339959, 2015-01-21 13:41:44 -0800 PST, ~2868 MiB cache)
2023-03-08 11:36:17.894 [INF] SYNC: Processed 406 blocks in the last 10.01s (279593 transactions, height 340365, 2015-01-24 16:50:34 -0800 PST, ~2912 MiB cache)
```

Roasbeef:
Another profile taken at ~7GB cache size: 

Screenshot 2023-03-08 at 11 50 56 AM

Roasbeef:
One other thing I notice due to the cache design (purge everything once full, instead of just evicting on the fly) is that post purge, there's a significant bump in I/O to refill the cache we guarantee that the very next block triggers thousands of cache misses. The machine I'm running on has a pretty fast SSD  (Samsung 980 Pro M.2 Gen 4) but validation still sees a tangible drop off due to the increased I/O. 

An alternative strategy would be something like:
  * Keep the normal height/age based eviction in place 
  * Once persistently at capacity, start to evict very old UTXOs in the background to make space for new items
     * This can be done in a batched manner

Roasbeef:
CPU profile shows most of the time is spent on GC (this is the default `GOGC` setting)

Screenshot 2023-03-08 at 12 40 15 PM

Roasbeef:
I think the other issue with this design is the flushing. Rather than incrementally evict items from the cache to write back on disk, we end up flushing the entire cache back to disk all at once. For every large caches, I've seen this take up to 20 minutes to accomplish (still flushing as I type this), during which, the entire block download is stalled.

abitrolly:
Can Go release memory immediately after a variable is released?

It looks like there are two caches here. 1 cache TX lookup. 2 buffer for writing to disk.

The lookup cache can be a FIFO queue, where items are moved if they are being hit. The oldest part of FIFO is written to the disk periodically. There could be good static memory structure for such data that doesn't require GC. Is that what eviction means?

kcalvinalvin:
> Can Go release memory immediately after a variable is released?

With arenas I believe you could. But it wouldn't be for a single variable.

> The lookup cache can be a FIFO queue, where items are moved if they are being hit. The oldest part of FIFO is written to the disk periodically. There could be good static memory structure for such data that doesn't require GC. Is that what eviction means?

Yeah basically.

Roasbeef:
> The lookup cache can be a FIFO queue, where items are moved if they are being hit

A hit is basically a deletion: if you're looking up a UTXO, it's because it's been spent in the first place. The cache as is, lets us avoid unnecessary random writes: UTXOs that die young are never written to disk. However, "young" here is relative to the cache size. Once the cache overflows, we write it all to disk, then start from scratch, which guarantees a cold cache (I/O spikes after the purge, as all UTXOs need to be loaded in). 

Re GC optimization:
  * Today we allocate a flat map for each item in the cache: https://github.com/btcsuite/btcd/blob/6c278f4b04463648ccba3dfeede4ee7b4e4af426/blockchain/utxocache.go#L245
  * When we go to add an item, we end up allocating the entry on the heap, then adding it to the cache. 
  * This causes a lot of active pointers that need to be scanned by the GC, this increases as a function of the cache size. Eventually this dominates GC time as after a purge, the GC must remove all these, while we're also adding new items. 
 

One approach (basically a middle ground to the Arena, but also complementary to it potentially), would be to pre-allocate al the UTXO entries in a single swoop, in a single slice. This would allocate all the entries to contiguous (virtual) memory. We'd keep an index of the last unallocated (not set to anything) entry in this slice. We'd then use this to create new UTXO entries, as adding a new entry will just be adding a pointer to the map, instead of actually allocating. Assuming we keep the current purge behavior, then then entire slice (in theory) can be de-allocated all at once, as all the refs in the map have been deleted and also the main slice pointer to longer held onto. 

Going further, we could then also actually never de-allocate the slice. We'd just reset that pointer and know that anything "after" that is actually blank. This is more or less the "memory ballast" approach outlined here: https://blog.twitch.tv/en/2019/04/10/go-memory-ballast-how-i-learnt-to-stop-worrying-and-love-the-heap/. 

IIUC, this has basically been subsumed by the arena allocator, as it gives you a nicer API (that uses generics). 

Re this:
> Once persistently at capacity, start to evict very old UTXOs in the background to make space for new items

I think the reason the cache is designed the way it is rn (copied from bitcoind) was to enable recovery from unclean shutdowns (to avoid a cache flush each block). I need to revisit that design to properly analyze it, but maybe it turns out this isn't so critical after all. One other side effect of this is:
  * If you set a cache size of 10 GB, then do an unclean shutdown, then do 5 GB, on start up in order to recover the cache size swells back to 10 GB. 


Also this flag (` env GODEBUG=gctrace=1`) can be very helpful in optimization. As it shows what happens with each GC pause, and how long the runtime spends on that.

blockchain/utxocache.go
```diff
@@ -232,7 +232,7 @@ type utxoBatcher interface {
 	// getEntries attempts to fetch a serise of entries from the UTXO view.
 	// If a single entry is not found within the view, then an error is
 	// returned.
-	getEntries(outpoints map[wire.OutPoint]struct{}) (map[wire.OutPoint]*UtxoEntry, error)
```
Roasbeef: Ah I guess "conventional " wisdom is that `struct{}` or `interface{}` doesn't add over head, but that doesn't factor in the map bucket overhead 🤔

blockchain/utxocache.go
```diff
@@ -640,7 +632,7 @@ func (s *utxoCache) Commit(view *UtxoViewpoint, node *blockNode) error {
 		if node != nil && isBIP0030Node(node) {
 			// Use the entry from the view since we want to overwrite the
 			// cached entry.
-			cachedEntry = entry.Clone()
```
Roasbeef: The cloning also showed up a lot in my memory profiles, what makes it no longer required? Fact that these never get mutated anyway?

blockchain/utxocache.go
```diff
@@ -235,6 +235,125 @@ type utxoBatcher interface {
 	getEntries(outpoints []wire.OutPoint) ([]*UtxoEntry, error)
 }
 
+// mapSlice is a slice of maps for Utxo entries.  The slice of maps are needed to
+// guarantee that the map will only take up N amount of bytes.  As of v1.20, the
+// go runtime will allocate 2^N + few extra buckets, meaning that for large N, we'll
+// allocate a lot of extra memory if the amount of entries goes over the previously
+// allocated buckets.  A slice of maps allows us to have a better control of how much
+// total memory gets allocated by all the maps.
+type mapSlice struct {
```
Roasbeef, line 244: Interesting...in the commit body how was the gap of choosing a size of 18,000MiB leading to an allocation of 13,000MiB identified?

blockchain/utxocache.go
```diff
@@ -235,6 +235,125 @@ type utxoBatcher interface {
...
+	return size
+}
+
+// get looks for the outpoint in all the maps in the map slice and returns
+// the entry.  nil and false is returned if the outpoint is not found.
+func (ms *mapSlice) get(op wire.OutPoint) (*UtxoEntry, bool) {
+	var entry *UtxoEntry
+	var found bool
+
+	for _, m := range ms.maps {
```
Roasbeef, line 281: Kinda feels like we're just reinventing hashmap bucket operations (multi-look up, overflow, etc) but on a "user space" level vs letting the runtime just handle it. 

I wonder if the maps themselves can be ordered based on recency, so assumption that check the lowest level first as that's where the newest UTXOs are found. This would require actually doing eviction though.

blockchain/utxocache.go
```diff
@@ -235,6 +235,125 @@ type utxoBatcher interface {
...
+	numMaxElements -= 1
+	ms.maxEntries = append(ms.maxEntries, numMaxElements)
+	ms.maps = append(ms.maps, make(map[wire.OutPoint]*UtxoEntry, numMaxElements))
+}
+
+// deleteMaps deletes all maps except for the first one which should be the biggest.
+func (ms *mapSlice) deleteMaps() {
+	// The rest of the map is now available to be garbage collected.
+	m := ms.maps[0]
+	ms.maps = []map[wire.OutPoint]*UtxoEntry{m}
```
Roasbeef, line 351: Shouldn't this once again pre-allocate? Otherwise, it needs to go through the successive doubling as new items are added.

blockchain/sizehelper.go
```diff
@@ -0,0 +1,178 @@
...
+//
+// bucketsize is (16 + keysize*8 + valuesize*8).  For a map of:
+// map[int64]int64, keysize=8 and valuesize=8.  There are edge cases
+// where the bucket size is different that I can't find the source code
+// for. https://github.com/golang/go/issues/34561#issuecomment-536115805
+//
+// I suspect it's because of alignment and how the compiler handles it but
+// when compared with how much the compiler allocates, it's a couple hundred
+// bytes off.
+func CalculateRoughMapSize(hint int, bucketSize uintptr) int {
```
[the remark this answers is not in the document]
kcalvinalvin: > I wonder if there're functions in the `unsafe` or `runtime` package to make this calculation a bit more future proof. Still need to review the improvements, but possible that this behavior isn't consistent across Go versions, or a new version overhauls the map internals which adds estimate drift to our calculations here.

That could definitely happen. You could grab the raw value of B in struct `hmap` like how dgraph-io people did. https://github.com/dgraph-io/dgraph/blob/main/posting/size.go but if struct `hmap` ends up changing how their fields are ordered, it could break that code as well.

> I came upon this: https://stackoverflow.com/a/31855111
But maybe you've seen it already, using unsafe.Sizeof on the map entry itself may also allow us to get a tighter estimate.

I didn't see that stackoverflow answer but the selected answer is wrong. The answer states that `len(theMap)` should be taken to calculate how much the memory the map is taking up but since the map allocates in powers of 2 (+few overflow buckets), that answer will be a lot different than what the actual size of the map is.

`unsafe.Sizeof` isn't actually unsafe and it only returns a constant value of the fixed part of each data structure, like the pointer and length of a string, but not indirect parts like the contents of the string (pg. 354 of [The Go Programming Language by Alan Donovan and Brian Kernighan](http://www.cs.uniroma2.it/upload/2017/TSC/The%20Go%20Programming%20Language.pdf). So `unsafe.Sizeof()` will always return 48 bytes for `hmap` in a 64bit system as the entire struct takes up 6 words. It wouldn't help us better calculate the size of how much memory the buckets are taking up.

What I still don't have an exact answer for is the alignment of `wire.Outpoint` in the map. `wire.Outpoint` is 36bytes but since memory is allocated per word, we'll allocate 40 bytes for `wire.Outpoint`. However, when used in the map, I think the compiler does some optimizations and we actually end up taking up less than 40 bytes when pre-allocating for a lot of key/value pairs. I suspect this is at least one reason why the map size isn't accurate down to the byte.

blockchain/utxocache.go
```diff
@@ -505,12 +509,6 @@ func (s *utxoCache) spendEntry(outpoint wire.OutPoint, addIfNil *UtxoEntry) erro
 	// Mark the output as spent and modified.
 	entry.packedFlags |= tfSpent | tfModified
 
-	//TODO(stevenroose) check if it's ok to drop the pkScript
-	// Since we don't need it anymore, drop the pkScript value of the entry.
-	s.totalEntryMemory -= entry.memoryUsage()
-	entry.pkScript = nil
```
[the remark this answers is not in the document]
kcalvinalvin: I'll save the pprofs and post it to github but what I've noticed was `msgTx.Decode()` took up a significant size when the `utxomaxcachesize` was large enough. Profiling the master branch didn't have `msgTx.Decode()` taking up significant amount of heap.

Looking at the source code for `msgTx.Decode()` we can see that the deserialized pkscript is actually copied over to the large contiguous byte slice and the pkscript that was allocated separately is returned to the scriptPool. https://github.com/btcsuite/btcd/blob/a18c2cfbf8929b7950da34b09c1dec27646a61d6/wire/msgtx.go#L661-L676

These two bits of information gives me a high confidence that the script for a whole block is sticking around because we have a single entry that's still pointing to it. I know for a fact that go doesn't garbage collect parts of a slice as well.

> So setting to nil would actually let it be freed up?

Yeah I think that's right in that we should still be replacing with nil since once every last one of the entries stops pointing to it, that whole block's script can be gc-ed. Just really shouldn't be discounting the pkscript after we nil it out since it can still be taking up space.

blockchain/utxocache.go
kcalvinalvin: Yeah I've noticed the map bucket overhead messing with the size calculation.

No need for a map here anyways so I think a slice is a strict improvement.

blockchain/utxocache.go
kcalvinalvin: Ah yeah that was my finding. If the entry is used internally within the blockchain package, we can make sure it's not modified so no need to Clone(). Also, I wasn't sure how much the Clone() does to help with immutability since the pkscript is still mutable as the Clone() here is a shallow clone.

I saw that the mining code does fetch for entries from the tx and one of the tests fails if I don't call Clone() for the entries that are fetched by public functions. But not calling Clone() for functions that are private didn't have any effect and all the code that calls private fetch functions do not mutate the entries.

blockchain/utxocache.go
kcalvinalvin: After the calculation of the entries, the go map will always allocate 2^N buckets with a few extra overflow buckets if the given number of entries are slightly above what 2^N buckets can store. The estimate for the amount of entries that can fit is 218,103,808. Each bucket for us is 368bytes so N=25 is chosen. This ends up being ~11,776MiB with overflow buckets accounting for ~2,000MiB.

When the map is full, the number is doubled, resulting in ~27,000MiB.

You can verify with the go compiler with:

```
var x map[wire.OutPoint]*UtxoEntry
var entries int = CalculateEntries(18000*1024*1024, bucketSize+avgEntrySize)
func BenchmarkEmptyMap(b *testing.B) {
       for i := 0; i < b.N; i++ {
               x = make(map[wire.OutPoint]*UtxoEntry, entries)
       }
}
```

Looks like the old code of `(18_000*1024*1024)/avgEntrySize` will make the map allocate N=26 and go right over the limit from  the get go.

blockchain/utxocache.go
kcalvinalvin: > Kinda feels like we're just reinventing hashmap bucket operations

Yeah... true but if we let the runtime handle it it'll allocate something that's bigger than what the user specified with `--utxocachemaxsize`.

> I wonder if the maps themselves can be ordered based on recency, so assumption that check the lowest level first as that's where the newest UTXOs are found. This would require actually doing eviction though.

At the moment the maps are ordered from largest to smallest. And since the smaller maps are allocated later, the general order is oldest to youngest. I guess you could look up the smaller maps first. Maybe that'll have a positive impact?

blockchain/utxocache.go
kcalvinalvin: The oldest map/largest is kept and the smaller ones are left out so they can be gc-ed.

The current algorithm will take a look at the leftover memory and allocate a map that will fit inside of that. My hunch is that since the entry sizes can vary, it's better to let the smaller map be gc-ed so the second map size can be dynamic based on the needs at that moment.

kcalvinalvin:
Found a bug in `utxoCache.InitConsistentState()` where if you signal to quit while the utxo cache is rolling back, it'll get stuck in an infinite loop on the next start.

Work towards a fix.

kcalvinalvin:
Found a bug in `utxoCache.InitConsistentState()` where if you signal to quit while the utxo cache is rolling back, it'll reject the next block after the tip and cause all the blocks afterwards to be recognized as orphans.

Work towards a fix.

Roasbeef:
cc @halseth

kcalvinalvin:
The last 3 commits alleviate the `InconsistentState` bug that I've noticed. The PR is ready for review.

blockchain/utxocache.go
```diff
@@ -0,0 +1,1085 @@
...
+	utxoBatchSizeBlocks = 50
+
+	// utxoFlushPeriodicThreshold is the threshold percentage at which a flush is
+	// performed when the flush mode FlushPeriodic is used.
+	utxoFlushPeriodicThreshold = 90
+
+	// outpointSize is the size of an outpoint in a 64-bit system.
+	//
+	// This value is calculated by running the following on a 64-bit system:
+	//	unsafe.Sizeof(wire.OutPoint{})
```
halseth, line 31: nit: Index is explicitly int32, so 64-bit system or not doesn't matter IIUC.

blockchain/utxocache.go
```diff
@@ -0,0 +1,1085 @@
...
+// get looks for the outpoint in all the maps in the map slice and returns
+// the entry.  nil and false is returned if the outpoint is not found.
+func (ms *mapSlice) get(op wire.OutPoint) (*UtxoEntry, bool) {
+	var entry *UtxoEntry
+	var found bool
+
+	for _, m := range ms.maps {
+		entry, found = m[op]
+		if found {
+			break
```
halseth, line 277: style nit: return early with the found element?

blockchain/utxocache.go
```diff
@@ -0,0 +1,1085 @@
...
+			continue
+		}
+		if len(m) >= maxNum {
+			// Don't try to insert if the map already at max since
+			// that'll force the map to allocate double the memory it's
+			// currently taking up.
+			continue
+		}
+
+		m[op] = entry
```
halseth, line 306: looks like this should be checking success first, otherwise it will insert it in all of the maps? Alternatively just return when it is first found.

blockchain/utxocache.go
```diff
@@ -0,0 +1,1085 @@
...
+	}
+
+	// Add each of the entries to the UTXO cache and update their memory
+	// usage.
+	//
+	// NOTE: When the fetched entry is nil, it is still added to the cache
+	// as a miss; this prevents future lookups to perform the same database
+	// fetch.
+	for i, entry := range entries {
+		s.totalEntryMemory += entry.memoryUsage()
```
halseth, line 452: Won't this be wrong if the entry is already in the cache?

halseth: Could consider moving the updates to `totalEntryMemory` inside the `put/delete` methods.

blockchain/utxocache.go
```diff
@@ -0,0 +1,1085 @@
...
+	return entries, nil
+}
+
+// FetchEntry returns the UTXO entry for the given outpoint.  It returns nil if
+// there is no entry for the outpoint in the UTXO state.
+//
+// This method is safe for concurrent access.
+func (s *utxoCache) FetchEntry(outpoint wire.OutPoint) (*UtxoEntry, error) {
+	s.mtx.Lock()
+	entries, err := s.getEntries([]wire.OutPoint{outpoint})
```
halseth, line 520: suggestion: Use the `getEntry` method, or remove the `getEntry` logic and use `getEntries` everywhere.

halseth:
Posted some initial comments while familiarizing myself with the changeset 😄

blockchain/utxocache.go
kcalvinalvin: Ah good point. I'll change the code to return when the key is found and write a test for it.

I guess this wasn't being triggered since you'd never insert duplicate outpoints.

blockchain/utxocache.go
kcalvinalvin: Like `return entry, found` instead of `break`?

blockchain/utxocache.go
kcalvinalvin: > Won't this be wrong if the entry is already in the cache?

`fetchAndCacheEntries()` is only called by `getEntry()` and `getEntries()` after they have checked in the cache. So all the outpoints they're fetching are guaranteed to not be in the cache.

I guess this is confusing. Maybe move the function inside `getEntries()` and have `getEntry()` just call `genEntries()` or just get rid of `getEntries()` entirely.

blockchain/utxocache.go
```diff
@@ -0,0 +1,1085 @@
...
+
+// getEntryByHash attempts to find any available UTXO for the given hash by
+// searching the entire set of possible outputs for the given hash.
+//
+// This method is part of the utxoByHashSource interface.
+// This method should be called with the state lock held.
+func (s *utxoCache) getEntryByHash(hash *chainhash.Hash) (*UtxoEntry, error) {
+	// First attempt to find a utxo with the provided hash in the cache.
+	prevOut := wire.OutPoint{Hash: *hash}
+	for idx := uint32(0); idx < MaxOutputsPerBlock; idx++ {
```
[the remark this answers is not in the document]
kcalvinalvin: Seems like it's only called during reorgs. I'm not sure about how often it's called. But it shouldn't matter either way.

Could just get rid of the cache lookup to simplify the code.

blockchain/utxocache.go
kcalvinalvin: Ah read this comment after I posted https://github.com/btcsuite/btcd/pull/1955#discussion_r1165018164

Will do so.

blockchain/utxocache.go
halseth: yep

blockchain/utxocache.go
kcalvinalvin: Had to look up some docs to make sure.

I think you're right. [32]byte should be 32 on 32bit systems and uint32 should be 4. I'll remove the comments like these in other parts of the code as well.

kcalvinalvin:
Converted to draft because I see that the PR itself isn't trivial to review. I saw some things that could be refactored away (ex:`utxoView` interface isn't needed) and the git commits aren't doing any favors for the reviewers.

I'll push something more readable in the next couple of days.

halseth:
> Converted to draft because I see that the PR itself isn't trivial to review. I saw some things that could be refactored away (ex:`utxoView` interface isn't needed) and the git commits aren't doing any favors for the reviewers.
> 
> I'll push something more readable in the next couple of days.

Tag me here when done and I'll review it again 😀

kcalvinalvin:
Ended up doing a lot of refactor with git commits that are clearer and easier to review. There are performance gains from the previous implementation as well.

blockchain/utxocache.go
```diff
@@ -122,3 +129,460 @@ func (ms *mapSlice) deleteMaps() {
...
+
+	// Add each of the entries to the UTXO cache and update their memory
+	// usage.
+	//
+	// NOTE: When the fetched entry is nil, it is still added to the cache
+	// as a miss; this prevents future lookups to perform the same database
+	// fetch.
+	for i := range dbEntries {
+		s.totalEntryMemory += dbEntries[i].memoryUsage()
+		s.cachedEntries.put(missingOps[i], dbEntries[i], s.totalEntryMemory)
```
halseth, line 282: Maybe I'm misunderstanding, but why is `s.totalEntryMemory` used as parameter here? Shouldn't it be the size of the single entry?

blockchain/utxocache.go
```diff
@@ -122,3 +129,460 @@ func (ms *mapSlice) deleteMaps() {
...
+
+	return size
+}
+
+// fetchEntries returns the UTXO entries for the given outpoints.  The function always
+// returns as many entries as there are outpoints and the returns entries are in the
+// same order as the outpoints.  It returns nil if there is no entry for the outpoint
+// in the UTXO set.
+//
+// This method should be called with the state lock held.
```
halseth, line 225: Should document what state lock this refers to?

blockchain/chain.go
```diff
@@ -665,7 +688,11 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
 	b.sendNotification(NTBlockConnected, block)
 	b.chainLock.Lock()
 
-	return nil
+	// Since we may have changed the UTXO cache, we make sure it didn't exceed its
+	// maximum size.
+	b.stateLock.Lock()
```
halseth, line 693: why statelock here, and chainlock everywhere else?

halseth (changes requested):
Starting to look very good 👍

blockchain/sizehelper.go
```diff
@@ -0,0 +1,217 @@
...
+//
+// bucketsize is (16 + keysize*8 + valuesize*8).  For a map of:
+// map[int64]int64, keysize=8 and valuesize=8.  There are edge cases
+// where the bucket size is different that I can't find the source code
+// for. https://github.com/golang/go/issues/34561#issuecomment-536115805
+//
+// I suspect it's because of alignment and how the compiler handles it but
+// when compared with how much the compiler allocates, it's a couple hundred
+// bytes off.
+func calculateRoughMapSize(hint int, bucketSize uintptr) int {
```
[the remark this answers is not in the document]
Roasbeef: Providing a bit of context here: initially we had very loose estimation, basically `numKeys*outpointSize+numValues*utxoSize`, but then it turned out we were underestimating by nearly a _factor of 2_. This would lead to behavior like trying to allocate 5 GB for the cache, leading to over 10 GB in practice. IMO users will have an expectation that the allocated cache amount is more or less exact, as they're coming from  `bitcoind` which can enable more accurate accounting with its manual memory management. 

Over time we can start to abstract out the calculation a bit into a new sub-interface to allow other size accounting implementations. Poked around a bit and found this, which uses `runtime` reflection to attempt to gain accurate map size counting: https://github.com/streamonkey/size/blob/v0.0.1/size.go#L110-L134C1

saubyk:
Hi @halseth @Roasbeef this pr is ready for the next round of reviews.

halseth (changes requested):
Staring to looks mergable to me. Just a few questions otherwise I think this looks good now :)

Did you try running this compiled with the ´-race` flag to ensure concurrent access is safe?

blockchain/utxocache.go
```diff
@@ -122,3 +129,417 @@ func (ms *mapSlice) deleteMaps() {
...
+	copy(entry.pkScript, txOut.PkScript)
+
+	entry.blockHeight = blockHeight
+	entry.packedFlags = tfFresh | tfModified
+	if isCoinBase {
+		entry.packedFlags |= tfCoinBase
+	}
+
+	s.cachedEntries.put(outpoint, entry, s.totalEntryMemory)
+	s.totalEntryMemory += entry.memoryUsage()
```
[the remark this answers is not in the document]
kcalvinalvin: It should be incremented after `put`. I'll change this.

blockchain/utxocache.go
```diff
@@ -122,3 +129,417 @@ func (ms *mapSlice) deleteMaps() {
...
+// writeCache writes all the entries that are cached in memory to the database atomically.
+func (s *utxoCache) writeCache(bestState *BestState) error {
+	// Update commits and flushes the cache to the database.
+	// NOTE: The database has its own cache which gets atomically written
+	// to leveldb.
+	err := s.db.Update(func(dbTx database.Tx) error {
+		utxoBucket := dbTx.Metadata().Bucket(utxoSetBucketName)
+		for i := range s.cachedEntries.maps {
+			for outpoint, entry := range s.cachedEntries.maps[i] {
+				// No need to update the cache if the entry is nil.  This
```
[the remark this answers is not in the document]
kcalvinalvin: Yeah will change this so that it's:

```
if entry == nil || entry.IsSpent()
```

blockchain/utxocache.go
```diff
@@ -122,3 +129,417 @@ func (ms *mapSlice) deleteMaps() {
...
+				// No need to update the cache if the entry was not modified.
+				if !entry.isModified() {
+					delete(s.cachedEntries.maps[i], outpoint)
+					continue
+				}
+
+				// If the entry is spent, remove the entry from the database
+				// and the cache.
+				if entry.IsSpent() {
+					err := dbDeleteUtxoEntry(utxoBucket, outpoint)
```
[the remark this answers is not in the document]
kcalvinalvin: Yeah will change this so that it's:

```
if entry == nil || entry.IsSpent()
```

blockchain/sizehelper.go
```diff
@@ -0,0 +1,221 @@
...
```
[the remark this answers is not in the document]
kcalvinalvin: That's a good point. It's not a copied function and I'll put in a wrapper function where it just uses the constant.

I didn't wanna just call the constant in case some math with the bucketsize either changes or there's a more accurate number.

blockchain/utxocache.go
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest push

blockchain/utxocache.go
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest push

blockchain/utxocache.go
```diff
@@ -0,0 +1,124 @@
...
+func (ms *mapSlice) makeNewMap(totalEntryMemory uint64) {
+	// Get the size of the leftover memory.
+	memSize := ms.maxTotalMemoryUsage - totalEntryMemory
+	for _, maxNum := range ms.maxEntries {
+		memSize -= uint64(calculateRoughMapSize(maxNum, bucketSize))
+	}
+
+	// Get a new map that's sized to house inside the leftover memory.
+	numMaxElements := calculateMinEntries(int(memSize), bucketSize+avgEntrySize)
+	numMaxElements -= 1
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest push

blockchain/utxocache.go
```diff
@@ -0,0 +1,124 @@
...
+	var found bool
+
+	for _, m := range ms.maps {
+		entry, found = m[op]
+		if found {
+			return entry, found
+		}
+	}
+
+	return entry, found
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest push

blockchain/utxocache.go
```diff
@@ -0,0 +1,124 @@
...
+// mapSlice is a slice of maps for utxo entries.  The slice of maps are needed to
+// guarantee that the map will only take up N amount of bytes.  As of v1.20, the
+// go runtime will allocate 2^N + few extra buckets, meaning that for large N, we'll
+// allocate a lot of extra memory if the amount of entries goes over the previously
+// allocated buckets.  A slice of maps allows us to have a better control of how much
+// total memory gets allocated by all the maps.
+type mapSlice struct {
+	maps []map[wire.OutPoint]*UtxoEntry
+
+	// maxEntries is the maximum amount of elemnts that the map is allocated for.
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest push

blockchain/sizehelper_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	return int(loadFactorNum * (bucketShift(B) / loadFactorDen))
+}
+
+func TestCalculateEntries(t *testing.T) {
+	for i := 0; i < 10_000_000; i++ {
+		// It's not possible to calculate the exact amount of entries since
+		// the map will only allocate for 2^N where N is the amount of buckets.
+		//
+		// So to see if the calculate entries fucntion is working correctly,
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest push

kcalvinalvin:
> Did you try running this compiled with the ´-race` flag to ensure concurrent access is safe?

Am ibd-ing right now with a binary that was compile with the `-race` flag. It's ok so far. I'll let it catch up and let it run once it's caught up as well.

kcalvinalvin:
Additional things pushed:

1. Rebased on top of master
2. Added 3 new commits to allow for pruning to work with the utxo cache code.
3. Added an extra test in `TestUtxoCacheFlush` to check that the utxo cache flushes for `FlushMode.FlushPeriodic`

blockchain/utxocache.go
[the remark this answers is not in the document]
halseth: Still not addressed?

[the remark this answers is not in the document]
halseth: Not that it's a big deal, but maybe more changes are missing as well

blockchain/utxocache.go
```diff
@@ -122,3 +129,406 @@ func (ms *mapSlice) deleteMaps() {
...
+	}
+
+	return entries, nil
+}
+
+// addTxOut adds the specified output to the cache if it is not provably
+// unspendable.  When the cache already has an entry for the output, it will be
+// overwritten with the given output.  All fields will be updated for existing
+// entries since it's possible it has changed during a reorg.
+func (s *utxoCache) addTxOut(outpoint wire.OutPoint, txOut *wire.TxOut, isCoinBase bool, blockHeight int32) error {
```
halseth, line 291: nit: long line

blockchain/utxocache.go
```diff
@@ -122,3 +129,406 @@ func (ms *mapSlice) deleteMaps() {
...
+// will be marked as spent and if the utxo is fresh (meaning that the database on disk
+// never saw it), it will be removed from the cache.
+func (s *utxoCache) addTxIn(txIn *wire.TxIn, stxos *[]SpentTxOut) error {
+	// Ensure the referenced utxo exists in the view.  This should
+	// never happen unless there is a bug is introduced in the code.
+	entries, err := s.fetchEntries([]wire.OutPoint{txIn.PreviousOutPoint})
+	if err != nil {
+		return err
+	}
+	if len(entries) < 1 || entries[0] == nil {
```
halseth, line 356: check `len(entries) != 1` instead?

blockchain/utxocache.go
```diff
@@ -122,3 +129,406 @@ func (ms *mapSlice) deleteMaps() {
...
+
+	return nil
+}
+
+// connectTransaction updates the cache by adding all new utxos created by the
+// passed transaction and marking and/or removing all utxos that the transactions
+// spend as spent.  In addition, when the 'stxos' argument is not nil, it will
+// be updated to append an entry for each spent txout.  An error will be returned
+// if the cache and the database does not contain the required utxos.
+func (s *utxoCache) connectTransaction(tx *btcutil.Tx, blockHeight int32, stxos *[]SpentTxOut) error {
```
halseth, line 421: long line

halseth (approved):
This LGTM now  💯 

EDIT: Untested. Could you post some numbers of IBD using the latest changeset?

blockchain/utxocache.go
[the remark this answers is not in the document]
kcalvinalvin: Oh hmm, I swear I changed it... maybe I messed up the git rebase.

I'll fix that up.

blockchain/utxocache.go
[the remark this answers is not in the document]
kcalvinalvin: Ok I definitely screwed up the git rebase somewhere. This also is the same :/

I'll check the other review comments as well.

kcalvinalvin:
> EDIT: Untested. Could you post some numbers of IBD using the latest changeset?

Sure, I'll post some numbers+flamegraphs.

blockchain/utxocache.go
```diff
@@ -0,0 +1,124 @@
...
+
+		m[op] = entry
+		return // Return as we were successful in adding the entry.
+	}
+
+	// We only reach this code if we've failed to insert into the map above as
+	// all the current maps were full.  We thus make a new map and insert into
+	// it.
+	ms.makeNewMap(totalEntryMemory)
+	m := ms.maps[len(ms.maps)-1] // new maps are appended to the map slice.
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed so that `makeNewMap` returns a new map

blockchain/utxocache.go
kcalvinalvin: Addressed in the latest push

blockchain/utxocache.go
kcalvinalvin: Changed to `len(entries) != 1 || entries[0] == nil` in the latest push

blockchain/utxocache.go
kcalvinalvin: Changed in the latest push

kcalvinalvin:
While generating the numbers + flamegraphs, I found two bugs which are fixed and now tested by the latest push.

1: There would be a concurrent map access error with the `mapslice` after the node has caught up. This has been addressed with a lock on the utxocache struct. Added a test for this bug.

2: There was a bug where the flush wouldn't happen during a prune as they were both accessing the database. The code now has been changed so that you have to pass in a `database.Tx`. Added a test for this bug.

3: Added a test for `InitConsistentState` just for good measure as I saw it wasn't in the test coverage.

kcalvinalvin:
Synced up to block height 801867 with the following command: `./btcd --datadir=. --logdir=. --whitelist=127.0.0.1 --listen=:9333 --rpclisten=:9334 --rpcuser=calvin --rpcpass=calvin --addpeer=127.0.0.1 --prune=400000 --utxocachemaxsize=16000 --cpuprofile=cpuprof-1 --memprofile=memprof-1`

Time taken: `20 hours, 9 minutes, 33 seconds`
Time taken up til the last checkpoint at height 751,565: `11 hours, 10 minutes, 31 seconds`

My computer: CPU: Ryzen 3600, Memory: 48GB, 1tb sandisk ssd.

The local peer is a bitcoin core node and I think sometimes the btcd node is slowed down because the core node just isn't serving blocks quick enough. Not sure if this is rate limiting by the core node.

## Cpu profile

image

Most of the time taken was with GC and signature checking code.

image


## Memory profile

image

The 2 most costly allocations were for downloaded blocks and tx copying during sighash calculation.

image


Raw profiles are provided here:
[profiles.tar.gz](https://github.com/btcsuite/btcd/files/12988534/profiles.tar.gz)

halseth:
Nice! How does it compare to the non-UTXO case on the same setup?

Would connection to multiple (real) peers make a difference?

kcalvinalvin:
> Nice! How does it compare to the non-UTXO case on the same setup?

Oh I took the numbers a few months back but I can't find the numbers now. It was over 24 hours. I can get the numbers again if needed.

> Would connection to multiple (real) peers make a difference?

In my testing no since I pointed it at my local node. In a real world setting, it would matter but since there isn't code for parallel block downloads at the moment, there wouldn't be a different with this code.

Roasbeef:
Ok I ran a sync and got to chain tip in ~16 hours: 
```
2023-11-15 00:39:16.139 [INF] BTCD: Version 0.23.3-beta
2023-11-15 16:34:35.923 [INF] SYNC: Processed 18 blocks in the last 12.13s (70918 transactions, height 816940, 2023-11-15 15:07:42 -0800 PST, ~6272 MiB cache)
```

I ran with a 16 GB cache.

Spec are: 
  * Intel® Core™ i9-13900K (Cores 24, Threads 32)
  * 32 GB RAM 
  * 2TB NVMe PCIe Gen 4 M.2

That's pretty damn good! Similar to the above, the thing that takes the most time when it gets to modern day blocks is the serialization of the transaction over and over again. I have this [PR](https://github.com/btcsuite/btcd/pull/1376) that resolves that, it has a bug somewhere though where something is mutated but the value isn't updated (only fails in a unit test). 

Similar to @kcalvinalvin above, it's syncing directly against a local bitcoind. My internet isn't super faster, and we already know a single slow peer can stall the entire download as we don't have parallel block download implemented, so I think this can give us a more accurate picture. 

I'm running another sync test against master, so we can compare.

Roasbeef:
Also one other diff with that PR is that things are cached only after a call to `block.Bytes()`, while the earlier one caches at basically as soon as things are read off the wire.

kcalvinalvin:
> Also one other diff with that PR is that things are cached only after a call to `block.Bytes()`, while the earlier one caches at basically as soon as things are read off the wire.

Out of draft now!

blockchain/chain.go
```diff
@@ -1121,11 +1160,18 @@ func (b *BlockChain) connectBestChain(node *blockNode, block *btcutil.Block, fla
...
 		// to the main chain without violating any rules and without
 		// actually connecting the block.
-		view := NewUtxoViewpoint()
-		view.SetBestHash(parentHash)
-		stxos := make([]SpentTxOut, 0, countSpentOutputs(block))
 		if !fastAdd {
-			err := b.checkConnectBlock(node, block, view, &stxos)
+			// We create a viewpoint here to avoid mutating the utxo cache.
+			// The block is not considered valid until checkconnectblock
+			// returns and the mutation would force us to undo the cache.
```
Crypt-iQ, line 1193: I might be misunderstanding the comment or the code, but if this is an invalid block, I think it will still mutate the cache. `checkConnectBlock` calls `fetchInputUtxos` which will call `fetchUtxosFromCache`. Then the block can be invalid on one of the later checks.

kcalvinalvin: Pushed a fixup commit to re-word the comment.

I originally made that comment to mean that we can't have new coins added or coins spent until all the validation passes as that would require new code to undo the changes to the utxocache. This is because `checkConnectBlock` adds/spends utxos before the signature validation.

It's true that the utxo cache is being mutated by `fetchUtxosFromCache` but that mutation is ok in that it only loads up utxos from the database. It doesn't spend or add new utxos. This mutation is ok as the worst thing that can happen is extra usage of memory in the cache.

Crypt-iQ (commented):
Looks good, just going to test this out

blockchain/chain.go
```diff
@@ -602,6 +611,16 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
 		curTotalTxns+numTxns, CalcPastMedianTime(node),
 	)
 
+	// If a utxoviewpoint was passed in, we'll be writing that viewpoint
+	// directly to the database on disk.  In order for the database to be
+	// consistent, we must flush the cache before writing the viewpoint.
+	if view != nil {
```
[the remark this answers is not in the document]
kcalvinalvin: In the current code, when a reorg happens, the utxoviewpoint will be passed in. This allows for less code changes for the reorg code as we can keep the current code that uses the utxoviewpoint directly.

However, when this happens, the cache may have unwritten new utxos or spent coins that should be deleted. In order for the code to be correct, the utxoview should be modifying the utxocache. But this results in more code being needed.

To avoid interactions with the utxoviewpoint and the utxocache, we just flush the utxocache first whenever the utxoviewpoint is passed in.

blockchain/chain.go
```diff
@@ -640,6 +640,25 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
...
+				// NOTE: the database will never be inconsistent here as the
+				// actual blocks are not deleted until the db.Update returns.
+				needsFlush, err := b.flushNeededAfterPrune(deletedHashes)
+				if err != nil {
+					return err
+				}
+				if needsFlush {
+					// Flush the utxo cache. This is a different db.Update
+					// and it'll commit before this current db.Update is
+					// committed.
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in fixup commit `4fce9c01c50187a430b10aaeefdaf0eb1b6e06d5`

Crypt-iQ (approved):
LGTM 🦖

blockchain/utxocache.go
```diff
@@ -126,3 +133,410 @@ func (ms *mapSlice) deleteMaps() {
...
+	if len(entries) != 1 || entries[0] == nil {
+		return AssertError(fmt.Sprintf("missing input %v",
+			txIn.PreviousOutPoint))
+	}
+
+	// Only create the stxo details if requested.
+	entry := entries[0]
+	if stxos != nil {
+		// Populate the stxo details using the utxo entry.
+		var stxo = SpentTxOut{
```
Roasbeef, line 409: Why `var` vs the normal `:=`?

blockchain/utxocache.go
```diff
@@ -126,3 +133,410 @@ func (ms *mapSlice) deleteMaps() {
...
+	return nil
+}
+
+// connectTransaction updates the cache by adding all new utxos created by the
+// passed transaction and marking and/or removing all utxos that the transactions
+// spend as spent.  In addition, when the 'stxos' argument is not nil, it will
+// be updated to append an entry for each spent txout.  An error will be returned
+// if the cache and the database does not contain the required utxos.
+func (s *utxoCache) connectTransaction(
+	tx *btcutil.Tx, blockHeight int32, stxos *[]SpentTxOut) error {
```
Roasbeef, line 466: Why have a pointer to the slice passed in which is mutated each time, vs just returning the set of UTXOs spent due to the transaction? Is the goal to reduce memory churn my pre-allocating `stxos` ahead of time? Current usage made the control flow a bit harder to follow IMO.

blockchain/utxocache.go
```diff
@@ -126,3 +133,410 @@ func (ms *mapSlice) deleteMaps() {
...
+// writeCache writes all the entries that are cached in memory to the database atomically.
+func (s *utxoCache) writeCache(bestState *BestState) error {
+	// Update commits and flushes the cache to the database.
+	// NOTE: The database has its own cache which gets atomically written
+	// to leveldb.
+	err := s.db.Update(func(dbTx database.Tx) error {
+		utxoBucket := dbTx.Metadata().Bucket(utxoSetBucketName)
+		for i := range s.cachedEntries.maps {
+			for outpoint, entry := range s.cachedEntries.maps[i] {
+				// If the entry is nil or spent, remove the entry from the database
```
Roasbeef: Style nit: `delete(s.cachedEntries.maps[i], outpoint)` appears 3 times in the loop, we can reduce it to one: 
```go
switch {
    case  entry == nil || entry.IsSpent():
        	err := dbDeleteUtxoEntry(utxoBucket, outpoint)
		if err != nil {
			return err
		}

    case !entry.isModified():
   default:
		// Entry is fresh and needs to be put into the database.
		err := dbPutUtxoEntry(utxoBucket, outpoint, entry)
		if err != nil {
			return err
		}
}

delete(s.cachedEntries.maps[i], outpoint)
```

blockchain/utxocache.go
```diff
@@ -540,3 +541,130 @@ func (s *utxoCache) flush(mode FlushMode, bestState *BestState) error {
...
+}
+
+// InitConsistentState checks the consistency status of the utxo state and
+// replays blocks if it lags behind the best state of the blockchain.
+//
+// It needs to be ensured that the chainView passed to this method does not
+// get changed during the execution of this method.
+func (b *BlockChain) InitConsistentState(tip *blockNode, interrupt <-chan struct{}) error {
+	s := b.utxoCache
+	// Load the consistency status from the database.
```
Roasbeef, line 603: Style nit: missing space above.

Roasbeef (approved):
LGTM 🐊


I think this is ready to land after a rebase. There's a lot of code that would violate the lnd linter/style, but we haven't yet ported over the latest version of our linter to `btcd`. I'll make a new PR to port that over, and fix the violations once this lands.

blockchain/utxocache.go
kcalvinalvin: That's a good point. There wasn't much thought that went into this besides just re-using the function prototype for `connectTransaction` that was in the current codebase. I'm assuming that was the intention when previous authors wrote this.

blockchain/utxocache.go
kcalvinalvin: Addressed

blockchain/utxocache.go
kcalvinalvin: Addressed

blockchain/utxocache.go
kcalvinalvin: Addressed

blockchain/utxocache.go
```diff
@@ -126,3 +133,410 @@ func (ms *mapSlice) deleteMaps() {
...
+	outpoint wire.OutPoint, txOut *wire.TxOut, isCoinBase bool, blockHeight int32) error {
+
+	// Don't add provably unspendable outputs.
+	if txscript.IsUnspendable(txOut.PkScript) {
+		return nil
+	}
+
+	entry := new(UtxoEntry)
+	entry.amount = txOut.Value
+	// Deep copy the script when the script in the entry differs from the one in
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed

blockchain/utxocache.go
```diff
@@ -126,3 +133,410 @@ func (ms *mapSlice) deleteMaps() {
...
+	}
+
+	return entries, nil
+}
+
+// addTxOut adds the specified output to the cache if it is not provably
+// unspendable.  When the cache already has an entry for the output, it will be
+// overwritten with the given output.  All fields will be updated for existing
+// entries since it's possible it has changed during a reorg.
+func (s *utxoCache) addTxOut(
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed

blockchain/utxocache.go
```diff
@@ -126,3 +133,410 @@ func (ms *mapSlice) deleteMaps() {
...
+
+// fetchEntries returns the UTXO entries for the given outpoints.  The function always
+// returns as many entries as there are outpoints and the returns entries are in the
+// same order as the outpoints.  It returns nil if there is no entry for the outpoint
+// in the UTXO set.
+//
+// The returned entries are NOT safe for concurrent access.
+func (s *utxoCache) fetchEntries(outpoints []wire.OutPoint) ([]*UtxoEntry, error) {
+	entries := make([]*UtxoEntry, len(outpoints))
+	var missingOps []wire.OutPoint
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed

blockchain/utxocache.go
```diff
@@ -126,3 +133,410 @@ func (ms *mapSlice) deleteMaps() {
...
+	copy(entry.pkScript, txOut.PkScript)
+
+	entry.blockHeight = blockHeight
+	entry.packedFlags = tfFresh | tfModified
+	if isCoinBase {
+		entry.packedFlags |= tfCoinBase
+	}
+
+	s.cachedEntries.put(outpoint, entry, s.totalEntryMemory)
+	s.totalEntryMemory += entry.memoryUsage()
```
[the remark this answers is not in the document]
kcalvinalvin: I have a patch for this but I'm not confident that it's a small change. Is it ok if I do a follow up PR?

kcalvinalvin:
Rebased and addressed the comments from roasbeef

kcalvinalvin:
All fixup commits squashed now in the latest push.

blockchain/utxocache.go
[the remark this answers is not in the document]
Roasbeef: SGTM.

PR #1956, opened 2023-03-03
Export MakeScritpNum, AsSmallInt, and IsSmallInt

martonp:
I would like to import these functions to use in https://github.com/decred/dcrdex/pull/2196. The exact same functions are exported in dcrd, but I'd rather import from here on the off chance that something changes.

Thanks!


jcvernaleo:
The reasoning makes sense to me.  Just waiting for the GH actions to run then I'll approve.

jcvernaleo (approved):
Looks good.

OK

PR #1964, opened 2023-03-20
psbt: add verification method for utxo data

ziggie1984:
This adds a method to verify the utxo data of a psbt package especially the `NonWitnessUtxo | WitnessUtxo` data.

This will be used in other packages to make sure we have the correct psbt structure with all relevant input fields populated.

Relates to lightningnetwork/lnd#7527

Happy to discuss other approaches, other PRs which call this function in lnd and btcwallet would follow if we want this function in this package ?

### DIFF
--- a/btcutil/psbt/utils.go
+++ b/btcutil/psbt/utils.go
@@ -398,6 +398,30 @@ func VerifyInputOutputLen(packet *Packet, needInputs, needOutputs bool) error {
 	return nil
 }
 
+// InputsReadyToSign makes sure that all input data have the previous output
+// specified meaning that either nonwitness UTXO or the witness UTXO data is
+// specified in the psbt package. This check is necessary because of 2 reasons.
+// The sighash calculation is now different for witnessV0 and witnessV1 inputs
+// this means we need to check the previous output pkScript for the specific
+// type and the second reason is that the sighash calculation for taproot inputs
+// include the previous output pkscripts.
+func InputsReadyToSign(packet *Packet) error {
+	err := VerifyInputOutputLen(packet, true, true)
+	if err != nil {
+		return err
+	}
+
+	for i := range packet.UnsignedTx.TxIn {
+		input := packet.Inputs[i]
+		if input.NonWitnessUtxo == nil && input.WitnessUtxo == nil {
+			return fmt.Errorf("invalid PSBT, input with index %d "+
+				"missing utxo information", i)
+		}
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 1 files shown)


btcutil/psbt/utils.go
```diff
@@ -398,6 +398,35 @@ func VerifyInputOutputLen(packet *Packet, needInputs, needOutputs bool) error {
...
 }
 
+// VerifyInputs makes sure that all input data have the previous output
+// specified meaning that either nonwitness data or the witness data is
+// specified in the psbt package. This check is necessary because of 2 reasons.
+// The sighash calculation is now different for witnessV0 and witnessV1 inputs
+// this means we need to check the previousOutput pkScript for the specific
+// type and the second reason is that the sighash calculation for taproot inputs
+// include the previousOutput pkscript.
+func VerifyInputs(packet *Packet) error {
```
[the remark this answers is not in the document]
ziggie1984: I find `InputsReadyToSign` good, will go with this.

ziggie1984:
Thank you for taking a look so fast, really appreciate this :)

btcutil/psbt/utils.go
```diff
@@ -398,6 +398,27 @@ func VerifyInputOutputLen(packet *Packet, needInputs, needOutputs bool) error {
...
 
+// InputsReadyToSign makes sure that all input data have the previous output
+// specified meaning that either nonwitness UTXO or the witness UTXO data is
+// specified in the psbt package. This check is necessary because of 2 reasons.
+// The sighash calculation is now different for witnessV0 and witnessV1 inputs
+// this means we need to check the previous output pkScript for the specific
+// type and the second reason is that the sighash calculation for taproot inputs
+// include the previous output pkscripts.
+func InputsReadyToSign(packet *Packet) error {
+	VerifyInputOutputLen(packet, true, true)
```
positiveblue, line 409: I guess we want to return an error if this function "fails"?

btcutil/psbt/utils.go
```diff
@@ -398,6 +398,27 @@ func VerifyInputOutputLen(packet *Packet, needInputs, needOutputs bool) error {
...
+
+	for i := range packet.UnsignedTx.TxIn {
```
positiveblue, line 411: I guess than wherever this is called we have already checked that `UnsignedTx` is not `nil`?

btcutil/psbt/utils.go
ziggie1984: Totally thanks you

positiveblue (approved):
Thanks for this :tada:

PR #1969, opened 2023-04-24
Add address field (rpc client & server)

### LINKED ISSUE #1874: rpc getrawtransaction return vout.scriptPubKey.addresses is deprecated at bitcoind version 22.0
issue #1754 problem still exists, not fixed

Content: rpc getrawtransaction return vout.scriptPubKey.addresses is deprecated at bitcoind version 22.0

https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-22.0.md#new-and-updated-rpcs

lindlof:
Bitcoin Core has added `address` field and removed `addresses` and `reqSigs` rpc response fields. The fields were deprecated on version 22 and they've been removed from version 23. For instance, the following methods are affected:

- https://bitcoincore.org/en/doc/22.0.0/rpc/rawtransactions/getrawtransaction/
- https://bitcoincore.org/en/doc/22.0.0/rpc/rawtransactions/decoderawtransaction/
- https://bitcoincore.org/en/doc/22.0.0/rpc/blockchain/gettxout/

This PR adds the new `address` and deprecates the removed fields.

Closes #1874

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -738,6 +738,13 @@ func createVoutList(mtx *wire.MsgTx, chainParams *chaincfg.Params, filterAddrMap
 		vout.ScriptPubKey.Type = scriptClass.String()
 		vout.ScriptPubKey.ReqSigs = int32(reqSigs)
 
+		// Address is defined when there's a single well-defined
+		// receiver address. To spend the output a signature for this,
+		// and only this, address is required.
+		if len(encodedAddrs) == 1 && reqSigs <= 1 {
+			vout.ScriptPubKey.Address = encodedAddrs[0]
+		}
+
 		voutList = append(voutList, vout)
 	}
 
@@ -856,6 +863,13 @@ func handleDecodeScript(s *rpcServer, cmd interface{}, closeChan <-chan struct{}
 	}
 	if scriptClass != txscript.ScriptHashTy {
 		reply.P2sh = p2sh.EncodeAddress()
+	}
+
+	// Address is defined when there's a single well-defined
+	// receiver address. To spend the output a signature for this,
+	// and only this, address is required.
+	if len(addresses) == 1 && reqSigs <= 1 {
+		reply.Address = addresses[0]
 	}
 	return reply, nil
 }
@@ -2725,6 +2739,7 @@ func handleGetTxOut(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (i
 	var value int64
 	var pkScript []byte
 	var isCoinbase bool
+	var address string
 	includeMempool := true
 	if c.IncludeMempool != nil {
 		includeMempool = *c.IncludeMempool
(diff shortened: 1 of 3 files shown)


bog-h:
PR looks fine! I wanted to open a similar PR because we now receive the address field from the node and don't have it in the result.

pefish (approved):

lindlof:
Okay @guggero I've left the deprecated fields in. The commit to remove them exists in https://github.com/lindlof/btcd/commit/faae23164d635ea6145f2108a96d4673e2811fc3

bog-h (approved):

rpcserver.go
```diff
@@ -732,6 +732,9 @@ func createVoutList(mtx *wire.MsgTx, chainParams *chaincfg.Params, filterAddrMap
 		var vout btcjson.Vout
 		vout.N = uint32(i)
 		vout.Value = btcutil.Amount(v.Value).ToBTC()
+		if len(encodedAddrs) == 1 && reqSigs == 1 {
```
guggero, line 735: Are we sure the `reqSigs` is always set to 1 on old versions? Can't we just omit that condition? And I guess a comment above about why we do this (backward compatibility) would be useful too.

lindlof: Added the condition because if more than 1 signatures are required then we shouldn't consider the address to be the only address. This could result in terrible bugs, such as the consumer believing they have received the Bitcoin when it's actually in a multisig.

I'll double-check and can add a comment and make it also accept 0 as long as an address is available. The `reqSigs` should be correct but I agree it makes sense for this to be robust.

lindlof: Added a comment and changed the `reqSigs` check to also accept 0. `ExtractPkScriptAddrs` simply returns the `reqSigs` as 1 whenever an address is returned, except for multisig which has more logic for `reqSigs` and where checking the value of `reqSigs` is more interesting.

bog-h (approved):

guggero (approved):
utACK

ygaberman-px:
@guggero @bog-h is there anything preventing from merging this in? This would be really helpful for me

guggero:
I guess another review from someone who has more experience with the code base than myself would be great. Any volunteers, @Roasbeef, @kcalvinalvin, @chappjc?

rpcserver.go
kcalvinalvin: Also checked that `ExtractPkScriptAddrs` returns a single address and `reqSigs == 1` for P2PKH, P2WPKH, P2SH, P2WSH, and P2TR.

The change would only be meaningful to non standard multi-sig.

kcalvinalvin:
ACK fcf291947d51896664e848526482983a5cbb72db26fa1d78649c3d051cf602fc

Also tested by calling `getrawtransaction`, `decoderawtransaction`, and `gettxout` to a Bitcoin Core node. Works well.

Issue #1069, opened 2017-10-17
Pruning support?


anacrolix:
#533 mentions pruning will be possible when several PRs are merged, which they are. What is the status of this feature? Is there any other way I can reduce the disk usage of running btcd?

atamgp:
Hi, I am also interested to know what the planning is concerning pruning support and if it conceptually will be the same as the official node implementation with same limitations compared to non-pruning.

zaptrem:
Has this been implemented?

MrAlexWeber:
From a [comment on #533](https://github.com/btcsuite/btcd/issues/533#issuecomment-150871631):

> The work in PRs #380, #491, and #492 make infrastructure changes that will ultimately allow it to be supported.

shilch:
Bump

shilch:
Bump @davecgh

stevenroose:
I might be interested in working on this. Especially with segwit, witness data should be prunable.

HildisviniOttar:
The major limitation with bitcoind is `-prune` also disables `-txindex` for use of rpc `getrawtransaction`. If btcd could enable block pruning whilst still supporting fetching raw transactions, it would be very compelling.

kcalvinalvin:
@

stevenroose:
Yaay! How's utreexo going??

PR #1971, opened 2023-04-28
main, wire, blockchain, indexers, ffldb: Add pruning

kcalvinalvin:
Adds pruning to btcd.

### Change to ffldb:

1) Add `PruneBlocks()` to transaction interface. Prune blocks is a no-op when the total block file size is equal to or under the target size. If the total block files size is over the target size, oldest block files are deleted along with the block indexes that point to those block files. Returns the hashes of the deleted blocks.

### Change to wire:

1) Add SFNodeNetworkLimited.
2) Add SFNodeNetworkLimited to default services.

### Change to blockchain:

1) New `pruneTarget` param to indicate how much the block files should take up.l
2) Attempts to prune blocks, spend journals, and relevant indexes on every `connectBlock`

### Change to indexers:

1) Added PruneBlock to Manager and implemented PruneBlock to every indexer.

### Change to main:

1) Prune flag is defined. It limits the prune to 1536MiB as we need to guarantee 288 blocks to abide by the NODE_NETWORK_LIMITED rule set by BIP159.
2) --addrindex and --prune at the same time is not allowed.
3) --txindex and --prune at the same time is not allowed.

Fixes https://github.com/btcsuite/btcd/issues/1069

### DIFF
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -115,6 +115,10 @@ type BlockChain struct {
 	// fields in this struct below this point.
 	chainLock sync.RWMutex
 
+	// pruneTarget is the size in bytes the database targets for when the node
+	// is pruned.
+	pruneTarget uint64
+
 	// These fields are related to the memory block index.  They both have
 	// their own locks, however they are often also protected by the chain
 	// lock to help prevent logic races when blocks are being processed.
@@ -600,6 +604,26 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
 
 	// Atomically insert info into the database.
 	err = b.db.Update(func(dbTx database.Tx) error {
+		// If the pruneTarget isn't 0, we should attempt to delete older blocks
+		// from the database.
+		if b.pruneTarget != 0 {
+			// When the total block size is under the prune target, prune blocks is
+			// a no-op and the deleted hashes are nil.
+			deletedHashes, err := dbTx.PruneBlocks(b.pruneTarget)
+			if err != nil {
+				return err
+			}
+
+			// Only attempt to delete if we have any deleted blocks.
+			if len(deletedHashes) != 0 {
+				// Delete the spend journals of the pruned blocks.
+				err = dbPruneSpendJournalEntry(dbTx, deletedHashes)
+				if err != nil {
+					return err
+				}
+			}
+		}
+
 		// Update best block state.
 		err := dbPutBestState(dbTx, state, node.workSum)
 		if err != nil {
@@ -1702,6 +1726,11 @@ type Config struct {
 	// This field can be nil if the caller is not interested in using a
 	// signature cache.
 	HashCache *txscript.HashCache
+
+	// Prune specifies the target database usage (in bytes) the database
+	// will target for with block files.  Prune at 0 specifies that no
+	// blocks will be deleted.
+	Prune uint64
 }
 
 // New returns a BlockChain instance using the provided configuration details.
@@ -1757,6 +1786,7 @@ func New(config *Config) (*BlockChain, error) {
 		prevOrphans:         make(map[chainhash.Hash][]*orphanBlock),
 		warningCaches:       newThresholdCaches(vbNumBits),
 		deploymentCaches:    newThresholdCaches(chaincfg.DefinedDeployments),
+		pruneTarget:         config.Prune,
 	}
 
 	// Ensure all the deployments are synchronized with our clock if
--- a/blockchain/indexers/addrindex.go
+++ b/blockchain/indexers/addrindex.go
@@ -991,3 +991,15 @@ func NewAddrIndex(db database.DB, chainParams *chaincfg.Params) *AddrIndex {
 func DropAddrIndex(db database.DB, interrupt <-chan struct{}) error {
 	return dropIndex(db, addrIndexKey, addrIndexName, interrupt)
 }
+
+// AddrIndexInitialized returns true if the address index has been created previously.
+func AddrIndexInitialized(db database.DB) bool {
+	var exists bool
+	db.View(func(dbTx database.Tx) error {
+		bucket := dbTx.Metadata().Bucket(addrIndexKey)
+		exists = bucket != nil
+		return nil
+	})
+
+	return exists
+}
--- a/blockchain/indexers/cfindex.go
+++ b/blockchain/indexers/cfindex.go
@@ -355,3 +355,15 @@ func NewCfIndex(db database.DB, chainParams *chaincfg.Params) *CfIndex {
 func DropCfIndex(db database.DB, interrupt <-chan struct{}) error {
 	return dropIndex(db, cfIndexParentBucketKey, cfIndexName, interrupt)
 }
+
+// CfIndexInitialized returns true if the cfindex has been created previously.
+func CfIndexInitialized(db database.DB) bool {
+	var exists bool
+	db.View(func(dbTx database.Tx) error {
+		bucket := dbTx.Metadata().Bucket(cfIndexParentBucketKey)
+		exists = bucket != nil
+		return nil
+	})
+
+	return exists
+}
--- a/blockchain/indexers/txindex.go
+++ b/blockchain/indexers/txindex.go
@@ -481,3 +481,15 @@ func DropTxIndex(db database.DB, interrupt <-chan struct{}) error {
 
 	return dropIndex(db, txIndexKey, txIndexName, interrupt)
 }
+
+// TxIndexInitialized returns true if the tx index has been created previously.
+func TxIndexInitialized(db database.DB) bool {
+	var exists bool
+	db.View(func(dbTx database.Tx) error {
+		bucket := dbTx.Metadata().Bucket(txIndexKey)
+		exists = bucket != nil
+		return nil
+	})
+
+	return exists
+}
--- a/config.go
+++ b/config.go
@@ -1147,6 +1147,14 @@ func loadConfig() (*config, []string, error) {
 		return nil, nil, err
 	}
 
+	if cfg.Prune != 0 && cfg.TxIndex {
+		err := fmt.Errorf("%s: the --prune and --txindex options may "+
+			"not be activated at the same time", funcName)
+		fmt.Fprintln(os.Stderr, err)
+		fmt.Fprintln(os.Stderr, usageMessage)
+		return nil, nil, err
+	}
+
 	// Warn about missing config file only after all other configuration is
 	// done.  This prevents the warning on help messages and invalid
 	// options.  Note this should go directly before the return.
--- a/config.go
+++ b/config.go
@@ -66,6 +66,7 @@ const (
 	sampleConfigFilename         = "sample-btcd.conf"
 	defaultTxIndex               = false
 	defaultAddrIndex             = false
+	pruneMinSize                 = 1536
 )
 
 var (
@@ -146,6 +147,7 @@ type config struct {
 	Proxy                string        `long:"proxy" description:"Connect via SOCKS5 proxy (eg. 127.0.0.1:9050)"`
 	ProxyPass            string        `long:"proxypass" default-mask:"-" description:"Password for proxy server"`
 	ProxyUser            string        `long:"proxyuser" description:"Username for proxy server"`
+	Prune                uint64        `long:"prune" description:"Prune already validated blocks from the database. Must specify a target size in MiB (minimum value of 1536, default value of 0 will disable pruning)"`
 	RegressionTest       bool          `long:"regtest" description:"Use the regression test network"`
 	RejectNonStd         bool          `long:"rejectnonstd" description:"Reject non-standard transactions regardless of the default settings for the active network."`
 	RejectReplacement    bool          `long:"rejectreplacement" description:"Reject transactions that attempt to replace existing transactions within the mempool through the Replace-By-Fee (RBF) signaling policy."`
@@ -1135,6 +1137,14 @@ func loadConfig() (*config, []string, error) {
 		cfg.oniondial = func(a, b string, t time.Duration) (net.Conn, error) {
 			return nil, errors.New("tor has been disabled")
 		}
+	}
+
+	if cfg.Prune != 0 && cfg.Prune < pruneMinSize {
+		err := fmt.Errorf("%s: the minimum value for --prune is %d. Got %d",
+			funcName, pruneMinSize, cfg.Prune)
+		fmt.Fprintln(os.Stderr, err)
+		fmt.Fprintln(os.Stderr, usageMessage)
+		return nil, nil, err
 	}
 
 	// Warn about missing config file only after all other configuration is
--- a/config.go
+++ b/config.go
@@ -1155,6 +1155,14 @@ func loadConfig() (*config, []string, error) {
 		return nil, nil, err
 	}
 
+	if cfg.Prune != 0 && cfg.AddrIndex {
+		err := fmt.Errorf("%s: the --prune and --addrindex options may "+
+			"not be activated at the same time", funcName)
+		fmt.Fprintln(os.Stderr, err)
+		fmt.Fprintln(os.Stderr, usageMessage)
+		return nil, nil, err
+	}
+
 	// Warn about missing config file only after all other configuration is
 	// done.  This prevents the warning on help messages and invalid
 	// options.  Note this should go directly before the return.
--- a/database/ffldb/blockio.go
+++ b/database/ffldb/blockio.go
@@ -15,6 +15,9 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"sort"
+	"strconv"
+	"strings"
 	"sync"
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
@@ -23,6 +26,10 @@ import (
 )
 
 const (
+	// blockFileExtension is the extension that's used to store the block
+	// files on the disk.
+	blockFileExtension = ".fdb"
+
 	// The Bitcoin protocol encodes block height as int32, so max number of
 	// blocks is 2^31.  Max block size per the protocol is 32MiB per block.
 	// So the theoretical max at the time this comment was written is 64PiB
@@ -32,7 +39,7 @@ const (
 	// 512MiB each for a total of ~476.84PiB (roughly 7.4 times the current
 	// theoretical max), so there is room for the max block size to grow in
 	// the future.
-	blockFilenameTemplate = "%09d.fdb"
+	blockFilenameTemplate = "%09d" + blockFileExtension
 
 	// maxOpenFiles is the max number of open files to maintain in the
 	// open blocks cache.  Note that this does not include the current
@@ -713,36 +720,57 @@ func (s *blockStore) handleRollback(oldBlockFileNum, oldBlockOffset uint32) {
 }
 
 // scanBlockFiles searches the database directory for all flat block files to
-// find the end of the most recent file.  This position is considered the
-// current write cursor which is also stored in the metadata.  Thus, it is used
-// to detect unexpected shutdowns in the middle of writes so the block files
-// can be reconciled.
-func scanBlockFiles(dbPath string) (int, uint32) {
-	lastFile := -1
-	fileLen := uint32(0)
-	for i := 0; ; i++ {
-		filePath := blockFilePath(dbPath, uint32(i))
-		st, err := os.Stat(filePath)
-		if err != nil {
-			break
-		}
-		lastFile = i
+// find the first file, last file, and the end of the most recent file.  The
+// position at the last file is considered the current write cursor which is
+// also stored in the metadata.  Thus, it is used to detect unexpected shutdowns
+// in the middle of writes so the block files can be reconciled.
+func scanBlockFiles(dbPath string) (int, int, uint32, error) {
+	firstFile, lastFile, lastFileLen, err := int(-1), int(-1), uint32(0), error(nil)
 
-		fileLen = uint32(st.Size())
+	files, err := filepath.Glob(filepath.Join(dbPath, "*"+blockFileExtension))
+	if err != nil {
+		return 0, 0, 0, err
+	}
+	sort.Strings(files)
+
+	// Return early if there's no block files.
+	if len(files) == 0 {
+		return firstFile, lastFile, lastFileLen, nil
 	}
 
-	log.Tracef("Scan found latest block file #%d with length %d", lastFile,
-		fileLen)
-	return lastFile, fileLen
+	// Grab the first and last file's number.
+	firstFile, err = strconv.Atoi(strings.TrimSuffix(filepath.Base(files[0]), blockFileExtension))
+	if err != nil {
+		return 0, 0, 0, fmt.Errorf("scanBlockFiles error: %v", err)
+	}
+	lastFile, err = strconv.Atoi(strings.TrimSuffix(filepath.Base(files[len(files)-1]), blockFileExtension))
+	if err != nil {
+		return 0, 0, 0, fmt.Errorf("scanBlockFiles error: %v", err)
+	}
+
+	// Get the last file's length.
+	filePath := blockFilePath(dbPath, uint32(lastFile))
+	st, err := os.Stat(filePath)
+	if err != nil {
+		return 0, 0, 0, err
+	}
+	lastFileLen = uint32(st.Size())
+
+	log.Tracef("Scan found latest block file #%d with length %d", lastFile, lastFileLen)
+
+	return firstFile, lastFile, lastFileLen, err
 }
 
 // newBlockStore returns a new block store with the current block file number
 // and offset set and all fields initialized.
-func newBlockStore(basePath string, network wire.BitcoinNet) *blockStore {
+func newBlockStore(basePath string, network wire.BitcoinNet) (*blockStore, error) {
 	// Look for the end of the latest block to file to determine what the
 	// write cursor position is from the viewpoing of the block files on
 	// disk.
-	fileNum, fileOff := scanBlockFiles(basePath)
+	_, fileNum, fileOff, err := scanBlockFiles(basePath)
+	if err != nil {
+		return nil, err
+	}
 	if fileNum == -1 {
 		fileNum = 0
 		fileOff = 0
@@ -765,5 +793,5 @@ func newBlockStore(basePath string, network wire.BitcoinNet) *blockStore {
 	store.openFileFunc = store.openFile
 	store.openWriteFileFunc = store.openWriteFile
 	store.deleteFileFunc = store.deleteFile
-	return store
+	return store, nil
 }
--- a/database/ffldb/db.go
+++ b/database/ffldb/db.go
@@ -1764,6 +1764,21 @@ func (tx *transaction) PruneBlocks(targetSize uint64) ([]chainhash.Hash, error) 
 	return deletedBlockHashes, nil
 }
 
+// BeenPruned returns if the block storage has ever been pruned.
+//
+// This function is part of the database.Tx interface implementation.
+func (tx *transaction) BeenPruned() (bool, error) {
+	first, last, _, err := scanBlockFiles(tx.db.store.basePath)
+	if err != nil {
+		return false, err
+	}
+
+	// If the database is pruned, then the first .fdb will not be there.
+	// We also check that there isn't just 1 file on disk or if there are
+	// no files on disk by checking if first != last.
+	return first != 0 && (first != last), nil
+}
+
 // Commit commits all changes that have been made to the root metadata bucket
 // and all of its sub-buckets to the database cache which is periodically synced
 // to persistent storage.  In addition, it commits all new blocks directly to
--- a/database/ffldb/db.go
+++ b/database/ffldb/db.go
@@ -1669,6 +1669,101 @@ func (tx *transaction) writePendingAndCommit() error {
 	return tx.db.cache.commitTx(tx)
 }
 
+// PruneBlocks deletes the block files until it reaches the target size
+// (specified in bytes).  Throws an error if the target size is below
+// the maximum size of a single block file.
+//
+// This function is part of the database.Tx interface implementation.
+func (tx *transaction) PruneBlocks(targetSize uint64) ([]chainhash.Hash, error) {
+	// Ensure transaction state is valid.
+	if err := tx.checkClosed(); err != nil {
+		return nil, err
+	}
+
+	// Ensure the transaction is writable.
+	if !tx.writable {
+		str := "prune blocks requires a writable database transaction"
+		return nil, makeDbErr(database.ErrTxNotWritable, str, nil)
+	}
+
+	// Make a local alias for the maxBlockFileSize.
+	maxSize := uint64(tx.db.store.maxBlockFileSize)
+	if targetSize < maxSize {
+		return nil, fmt.Errorf("got target size of %d but it must be greater "+
+			"than %d, the max size of a single block file",
+			targetSize, maxSize)
+	}
+
+	first, last, lastFileSize, err := scanBlockFiles(tx.db.store.basePath)
+	if err != nil {
+		return nil, err
+	}
+
+	// If we have no files on disk or just a single file on disk, return early.
+	if first == last {
+		return nil, nil
+	}
+
+	// Last file number minus the first file number gives us the count of files
+	// on disk minus 1.  We don't want to count the last file since we can't assume
+	// that it is of max size.
+	maxSizeFileCount := last - first
+
+	// If the total size of block files are under the target, return early and
+	// don't prune.
+	totalSize := uint64(lastFileSize) + (maxSize * uint64(maxSizeFileCount))
+	if totalSize <= targetSize {
+		return nil, nil
+	}
+
+	log.Tracef("Using %d more bytes than the target of %d MiB. Pruning files...",
+		totalSize-targetSize,
+		targetSize/(1024*1024))
+
+	deletedFiles := make(map[uint32]struct{})
+
+	// We use < not <= so that the last file is never deleted.  There are other checks in place
+	// but setting it to < here doesn't hurt.
+	for i := uint32(first); i < uint32(last); i++ {
+		err = tx.db.store.deleteFileFunc(i)
+		if err != nil {
+			return nil, fmt.Errorf("PruneBlocks: Failed to delete block file "+
+				"number %d: %v", i, err)
+		}
+
+		// Add the file index to the deleted files map so that we can later
+		// delete the block location index.
+		deletedFiles[i] = struct{}{}
+
+		// If we're already at or below the target usage, break and don't
+		// try to delete more files.
+		totalSize -= maxSize
+		if totalSize <= targetSize {
+			break
+		}
+	}
+
+	// Delete the indexed block locations for the files that we've just deleted.
+	var deletedBlockHashes []chainhash.Hash
+	cursor := tx.blockIdxBucket.Cursor()
+	for ok := cursor.First(); ok; ok = cursor.Next() {
+		loc := deserializeBlockLoc(cursor.Value())
+
+		_, found := deletedFiles[loc.blockFileNum]
+		if found {
+			deletedBlockHashes = append(deletedBlockHashes, *(*chainhash.Hash)(cursor.Key()))
+			err := cursor.Delete()
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	log.Tracef("Finished pruning. Database now at %d bytes", totalSize)
+
+	return deletedBlockHashes, nil
+}
+
 // Commit commits all changes that have been made to the root metadata bucket
 // and all of its sub-buckets to the database cache which is periodically synced
 // to persistent storage.  In addition, it commits all new blocks directly to
--- a/database/ffldb/db.go
+++ b/database/ffldb/db.go
@@ -2016,7 +2016,10 @@ func openDB(dbPath string, network wire.BitcoinNet, create bool) (database.DB, e
 	// according to the data that is actually on disk.  Also create the
 	// database cache which wraps the underlying leveldb database to provide
 	// write caching.
-	store := newBlockStore(dbPath, network)
+	store, err := newBlockStore(dbPath, network)
+	if err != nil {
+		return nil, convertErr(err.Error(), err)
+	}
 	cache := newDbCache(ldb, store, defaultCacheSize, defaultFlushSecs)
 	pdb := &db{store: store, cache: cache}
 
--- a/database/ffldb/driver_test.go
+++ b/database/ffldb/driver_test.go
@@ -319,6 +319,21 @@ func TestPrune(t *testing.T) {
 		if err != nil {
 			t.Fatal(err)
 		}
+		err = db.View(func(tx database.Tx) error {
+			pruned, err := tx.BeenPruned()
+			if err != nil {
+				return err
+			}
+
+			if pruned {
+				err = fmt.Errorf("The database hasn't been pruned but " +
+					"BeenPruned returned true")
+			}
+			return err
+		})
+		if err != nil {
+			t.Fatal(err)
+		}
 
 		var deletedBlocks []chainhash.Hash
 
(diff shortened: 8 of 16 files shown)


saubyk:
cc @halseth in case you're interested in reviewing this pr. Thanks

database/ffldb/db.go
```diff
@@ -1669,6 +1669,94 @@ func (tx *transaction) writePendingAndCommit() error {
...
+		return nil, err
+	}
+
+	// Ensure the transaction is writable.
+	if !tx.writable {
+		str := "prune blocks requires a writable database transaction"
+		return nil, makeDbErr(database.ErrTxNotWritable, str, nil)
+	}
+
+	if targetSize < tx.db.store.maxBlockFileSize {
```
halseth, line 1689: not 288*blocksize?

database/interface.go
```diff
@@ -389,6 +389,15 @@ type Tx interface {
...
 	FetchBlockRegions(regions []BlockRegion) ([][]byte, error)
 
+	// PruneBlocks deletes the block files until it reaches the target size
+	// (specificed in bytes).
+	//
+	// The interface contract guarantees at least the following errors will
+	// be returned (other implementation-specific errors are possible):
+	//   - ErrTxNotWritable if attempted against a read-only transaction
+	//   - ErrTxClosed if the transaction has already been closed
+	PruneBlocks(targetSize uint32) ([][]byte, error)
```
halseth, line 399: document return value. Chould be `[]chainhash.Hash`?

halseth (changes requested):
Looks pretty good to me :)

database/ffldb/db.go
kcalvinalvin: No because I wanted the pruning behavior to be similar to bitcoin core's where the user can set a prune size. For example, the user could want a prune size of 70GB instead of the smallest possible size.

database/interface.go
kcalvinalvin: Yes it could be `[]chainhash.Hash`. I'll update the code and document the return value

database/ffldb/blockio.go
```diff
@@ -713,36 +719,57 @@ func (s *blockStore) handleRollback(oldBlockFileNum, oldBlockOffset uint32) {
...
+
+	lastFile, firstFile = -1, -1
+	for _, file := range files {
+		// Extract the file number.
+		fileNum := file
+		fileNum = strings.TrimPrefix(fileNum, dbPath+"/")
+		fileNum = strings.TrimSuffix(fileNum, blockFileExtension)
+
+		// Turn the string into a number.
+		lastFile, err = strconv.Atoi(fileNum)
```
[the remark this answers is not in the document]
kcalvinalvin: Good point... Seems like `filepath.Glob` will spit out things in the order the files appear on your filesystem (which was always sorted).

I'll put in a sort so it's guaranteed the fileNames will be sorted.

database/ffldb/driver_test.go
```diff
@@ -253,6 +255,129 @@ func TestPersistence(t *testing.T) {
...
+
+		// Check that all the blocks that say were deleted are deleted from the
+		// block index bucket as well.
+		err = db.View(func(tx database.Tx) error {
+			for _, deletedBlock := range deletedBlocks {
+				_, err := tx.FetchBlock(&deletedBlock)
+				if dbErr, ok := err.(database.Error); !ok ||
+					dbErr.ErrorCode != database.ErrBlockNotFound {
+
+					return fmt.Errorf("Expected ErrBlockNotFound but got nil")
```
guggero, line 345: Might not be `nil` but an unexpected error, so include it in the message?

database/interface.go
```diff
@@ -389,6 +389,21 @@ type Tx interface {
...
 
+	// PruneBlocks deletes the block files until it reaches the target size
+	// (specificed in bytes).
+	//
+	// The interface contract guarantees at least the following errors will
+	// be returned (other implementation-specific errors are possible):
+	//   - ErrTxNotWritable if attempted against a read-only transaction
+	//   - ErrTxClosed if the transaction has already been closed
+	//
+	// NOTE: The data returned by this function is only valid during a
```
guggero, line 400: The actual comment on the implementation should be identical to this one IMO.

blockchain/chain.go
```diff
@@ -665,6 +670,37 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
...
+			// When the total block size is under the prune target, prune blocks is
+			// a no-op and the deleted hashes are nil.
+			deletedHashes, err := dbTx.PruneBlocks(b.pruneTarget)
+			if err != nil {
+				return err
+			}
+
+			// No need to try to delete the spend journal entries or the indexes
+			// if we haven't deleted any blocks.
+			if deletedHashes == nil {
```
guggero, line 686: Check `len()` instead of `== nil` to ignore underlying implementation assumptions?

blockchain/indexers/cfindex.go
```diff
@@ -255,6 +255,37 @@ func (idx *CfIndex) DisconnectBlock(dbTx database.Tx, block *btcutil.Block,
 	return nil
 }
 
+// PruneBlock is invoked when an older block is deleted after it's been
+// processed.  For cfindx, PruneBlock is exactly the same as disconnecting a block.
+// TODO (kcalvinalvin): Consider keeping the filters at a later date to help with
```
guggero, line 260: I guess we could also support fast rescan using compact filters in the future when pruning? Maybe it would be a great idea to have an option like `prune-cfilters=true/false` that is `false` by default (since filters for mainnet are currently only around 8 GB).

wire/protocol.go
```diff
@@ -93,18 +93,23 @@ const (
 	// SFNode2X is a flag used to indicate a peer is running the Segwit2X
 	// software.
 	SFNode2X
+
+	// SFNodeNetWorkLimited is a flag used to indicate a peer supports serving
```
guggero, line 97: Ah, if this is an actual p2p requirement that we need to be able to serve at least the last 288 blocks, I think we should update our minimum prune value accordingly (see previous comment of not using the max block file size as the minimum).

config.go
```diff
@@ -66,6 +66,7 @@ const (
 	sampleConfigFilename         = "sample-btcd.conf"
 	defaultTxIndex               = false
 	defaultAddrIndex             = false
+	pruneMinSize                 = 1536
```
guggero, line 69: Ah, I see there's another min size here. Shouldn't we sync this with what's being checked in `database/ffldb/db.go`?

config.go
```diff
@@ -145,6 +146,7 @@ type config struct {
 	Proxy                string        `long:"proxy" description:"Connect via SOCKS5 proxy (eg. 127.0.0.1:9050)"`
 	ProxyPass            string        `long:"proxypass" default-mask:"-" description:"Password for proxy server"`
 	ProxyUser            string        `long:"proxyuser" description:"Username for proxy server"`
+	Prune                uint32        `long:"prune" description:"Prune already validated blocks from the database. Must specify a target size in MiB (minimum value of 1024)"`
```
guggero, line 149: Description doesn't match chosen min size constant value.

guggero (commented):
Great work, very cool to see such an impactful feature being introduced in only 8 commits! Did a first pass review, will do some manual testing in the next round.

database/ffldb/db.go
```diff
@@ -1669,6 +1669,94 @@ func (tx *transaction) writePendingAndCommit() error {
...
+	}
+
+	var deletedBlockHashes []chainhash.Hash
+	cursor := tx.blockIdxBucket.Cursor()
+	for ok := cursor.First(); ok; ok = cursor.Next() {
+		loc := deserializeBlockLoc(cursor.Value())
+
+		_, found := deletedFiles[loc.blockFileNum]
+		if found {
+			deletedBlockHashes = append(deletedBlockHashes, *(*chainhash.Hash)(cursor.Key()))
```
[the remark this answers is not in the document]
kcalvinalvin: Not quite sure what you mean here. Mine is set to 8 spaces already and I couldn't find examples of what you mentioned here.

Also, afaik, tabspace shouldn't matter with tabs anyways since the editor you're using should render it as whatever you set right?

database/ffldb/db.go
kcalvinalvin: That's true that we have to keep at least 2 days worth of blocks but that's not the check being done here. The check here is if the database can actually reach the `targetSize` that the caller asked for. Since each of the `.blk` files are 512MiB each, it's impossible to reach anything less than that. For example, if the caller asks for a target size of  256MiB, it'll error out here.

That's the only guarantee that the database makes. The caller should be careful to not ask for anything lower than what it'd need. That's enforced in `config.go`

config.go
kcalvinalvin: So the 1536 number came from the fact that we have these constants.

```
blocksNeeded = 288
ConsensusMaxBlockSize = 4,000,000
onDiskBlkFileSizeInBytes = 536,870,912
```

We need to keep 288 blocks to adhere to `NODE_NETWORK_LIMITED` BIP. Each block can be a maximum of 4MB. Each block file (XXXXXX.fdb) is 512MiB.

```
(blocksNeeded*ConsensusMaxBlockSize)/onDiskBlkFileSizeInBytes = 2.145
```

We do the math and we need to keep 2.14 on disk .fdb files to adhere to the `NODE_NETWORK_LIMITED` rule. Can't keep .14 files so we round up to 3. `512MiB * 3` is 1536MiB which is out `pruneMinSize`.

As mentioned [here](https://github.com/btcsuite/btcd/pull/1971#discussion_r1211140556), the check in `database/ffldb/db.go` is a different check so it's ok.

wire/protocol.go
kcalvinalvin: Refer to the reply here: https://github.com/btcsuite/btcd/pull/1971#discussion_r1200443717

database/ffldb/db.go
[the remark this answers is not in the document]
guggero: I mean that the code in this PR often flows over the targeted 80 character line length limit we use in `lnd`. Though I see that is not specifically mentioned in this project's code contribution guidelines. So feel free to ignore as much of the existing code also doesn't seem to use that limit.

database/ffldb/blockio.go
```diff
@@ -713,36 +720,58 @@ func (s *blockStore) handleRollback(oldBlockFileNum, oldBlockOffset uint32) {
...
-func scanBlockFiles(dbPath string) (int, uint32) {
-	lastFile := -1
-	fileLen := uint32(0)
-	for i := 0; ; i++ {
-		filePath := blockFilePath(dbPath, uint32(i))
+// find the first file, last file, and the end of the most recent file.  The
+// position at the last file is considered the current write cursor which is
+// also stored in the metadata.  Thus, it is used to detect unexpected shutdowns
+// in the middle of writes so the block files can be reconciled.
+func scanBlockFiles(dbPath string) (firstFile int, lastFile int, lastFileLen uint32, err error) {
```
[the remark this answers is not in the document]
kcalvinalvin: Changed to `return firstFile, lastFile, lastFileLen, nil` in the latest push

database/ffldb/blockio.go
```diff
@@ -713,36 +720,58 @@ func (s *blockStore) handleRollback(oldBlockFileNum, oldBlockOffset uint32) {
...
+	// The only error we can get is a bad pattern error.  Since we're hardcoding
+	// the pattern, we should not have an error at runtime.
+	files, _ := (filepath.Glob(filepath.Join(dbPath, "*"+blockFileExtension)))
+	sort.Strings(files)
+
+	lastFile, firstFile = -1, -1
+	for _, file := range files {
+		// Extract the file number.
+		fileNum := file
+		fileNum = strings.TrimPrefix(fileNum, dbPath+"/")
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest push

database/ffldb/blockio.go
```diff
@@ -713,36 +720,58 @@ func (s *blockStore) handleRollback(oldBlockFileNum, oldBlockOffset uint32) {
...
-	for i := 0; ; i++ {
-		filePath := blockFilePath(dbPath, uint32(i))
+// find the first file, last file, and the end of the most recent file.  The
+// position at the last file is considered the current write cursor which is
+// also stored in the metadata.  Thus, it is used to detect unexpected shutdowns
+// in the middle of writes so the block files can be reconciled.
+func scanBlockFiles(dbPath string) (firstFile int, lastFile int, lastFileLen uint32, err error) {
+	// The only error we can get is a bad pattern error.  Since we're hardcoding
+	// the pattern, we should not have an error at runtime.
+	files, _ := (filepath.Glob(filepath.Join(dbPath, "*"+blockFileExtension)))
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest push

database/ffldb/db.go
```diff
@@ -1669,6 +1669,94 @@ func (tx *transaction) writePendingAndCommit() error {
 	return tx.db.cache.commitTx(tx)
 }
 
+// PruneBlocks deletes the block files until it reaches the target size
+// (specificed in bytes).  Throws an error if the target size is below
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest push

database/ffldb/db.go
```diff
@@ -1669,6 +1669,94 @@ func (tx *transaction) writePendingAndCommit() error {
...
+	}
+
+	// Last file number minus the first file number gives us the count of files
+	// on disk minus 1.  We don't want to count the last file since we can't assume
+	// that it is of max size.
+	maxSizeFileCount := last - first
+
+	// If the total size of block files are under the target, return early and
+	// don't prune.
+	totalSize := lastFileSize + (tx.db.store.maxBlockFileSize * uint32(maxSizeFileCount))
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest push

database/ffldb/db.go
```diff
@@ -1669,6 +1669,94 @@ func (tx *transaction) writePendingAndCommit() error {
...
+	if totalSize <= targetSize {
+		return nil, err
+	}
+
+	log.Tracef("Using %d more bytes than the target of %d MiB. Pruning files...",
+		totalSize-targetSize,
+		targetSize/(1024*1024))
+
+	deletedFiles := make(map[uint32]struct{})
+	for i := uint32(first); i <= uint32(last); i++ {
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest push. There were other checks to stop it from happening but change it anyways since it doesn't hurt

config.go
kcalvinalvin: Addressed in the latest push

blockchain/chain.go
kcalvinalvin: Addressed in the latest push

database/ffldb/driver_test.go
kcalvinalvin: Addressed in the latest push

database/ffldb/driver_test.go
```diff
@@ -253,6 +255,129 @@ func TestPersistence(t *testing.T) {
...
+		}
+		err = db.Update(func(tx database.Tx) error {
+			for i, block := range blocks {
+				err := tx.StoreBlock(block)
+				if err != nil {
+					return fmt.Errorf("StoreBlock #%d: unexpected error: "+
+						"%v", i, err)
+				}
+			}
+			return err
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest push

database/ffldb/db.go
```diff
@@ -1669,6 +1669,94 @@ func (tx *transaction) writePendingAndCommit() error {
...
+	// Last file number minus the first file number gives us the count of files
+	// on disk minus 1.  We don't want to count the last file since we can't assume
+	// that it is of max size.
+	maxSizeFileCount := last - first
+
+	// If the total size of block files are under the target, return early and
+	// don't prune.
+	totalSize := lastFileSize + (tx.db.store.maxBlockFileSize * uint32(maxSizeFileCount))
+	if totalSize <= targetSize {
+		return nil, err
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest push

database/ffldb/db.go
```diff
@@ -1669,6 +1669,94 @@ func (tx *transaction) writePendingAndCommit() error {
...
+	}
+
+	first, last, lastFileSize, err := scanBlockFiles(tx.db.store.basePath)
+	if err != nil {
+		return nil, err
+	}
+
+	// If we have no files on disk or just a single file on disk, return early.
+	if first == last {
+		return nil, err
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest push

database/ffldb/driver_test.go
```diff
@@ -253,6 +255,129 @@ func TestPersistence(t *testing.T) {
...
 
+// TestPrune tests that the older .fdb files are deleted with a call to prune.
+func TestPrune(t *testing.T) {
+	t.Parallel()
+
+	// Create a new database to run tests against.
+	dbPath := filepath.Join(os.TempDir(), "ffldb-prunetest")
+	_ = os.RemoveAll(dbPath)
+	db, err := database.Create(dbType, dbPath, blockDataNet)
+	if err != nil {
```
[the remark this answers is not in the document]
kcalvinalvin: This one? https://pkg.go.dev/github.com/stretchr/testify/require

I don't see it in the go.mod though

blockchain/indexers/cfindex.go
kcalvinalvin: I just realized that the fast rescan for bitcoin core is for the wallet. Since btcd doesn't have any wallet functionality with itself, there isn't a need for the compact filters for this reason.

Does btcwallet ask for the filters? Might be worth keeping the filters around for it?

database/interface.go
kcalvinalvin: I think it's better to have the comments separate since the interface and the implementation are different. It's also how other comments are done as well in the existing code.

database/ffldb/driver_test.go
```diff
@@ -253,6 +255,129 @@ func TestPersistence(t *testing.T) {
 	}
 }
 
+// TestPrune tests that the older .fdb files are deleted with a call to prune.
+func TestPrune(t *testing.T) {
+	t.Parallel()
+
+	// Create a new database to run tests against.
+	dbPath := filepath.Join(os.TempDir(), "ffldb-prunetest")
+	_ = os.RemoveAll(dbPath)
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest push

database/ffldb/blockio.go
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest push

wire/protocol.go
losh11: I believe `SFNodeNetworkLimited` is untyped here

blockchain/indexers/txindex.go
```diff
@@ -430,6 +430,16 @@ func (idx *TxIndex) DisconnectBlock(dbTx database.Tx, block *btcutil.Block,
...
 }
 
+// PruneBlock is invoked when an older block is deleted after it's been
+// processed.
+// NOTE: For TxIndex, it's a no-op as TxIndex isn't allowed to be enabled
+// with pruning.
+//
+// This is part of the Indexer interface.
+func (idx *TxIndex) PruneBlock(dbTx database.Tx, blockHash chainhash.Hash) error {
+	return nil
```
ProofOfKeags, line 440: seems like this ought to be an error given the rest of this PR

blockchain/indexers/addrindex.go
```diff
@@ -809,6 +809,16 @@ func (idx *AddrIndex) DisconnectBlock(dbTx database.Tx, block *btcutil.Block,
...
 }
 
+// PruneBlock is invoked when an older block is deleted after it's been
+// processed.
+// NOTE: For AddrIndex, it's a no-op as AddrIndex isn't allowed to be enabled
+// with pruning.
+//
+// This is part of the Indexer interface.
+func (idx *AddrIndex) PruneBlock(dbTx database.Tx, blockHash chainhash.Hash) error {
+	return nil
```
ProofOfKeags, line 819: seems like this ought to be an error given the rest of this PR

database/ffldb/blockio.go
```diff
@@ -713,36 +720,61 @@ func (s *blockStore) handleRollback(oldBlockFileNum, oldBlockOffset uint32) {
...
+		return 0, 0, 0, err
+	}
+	sort.Strings(files)
+
+	lastFile, firstFile = -1, -1
+	for _, file := range files {
+		// Extract the file number.
+		fileNum := file
+		fileNum = filepath.Base(fileNum)
+		fileNum = strings.TrimSuffix(fileNum, blockFileExtension)
```
ProofOfKeags, line 741: Inline this.

```suggestion
		fileNum := strings.TrimSuffix(filepath.Base(file), blockFileExtension)
```

If you don't want to inline it I strongly suggest not mutating `fileNum` like this and just create different names for the fragments. If you're worried about overallocation consider allocating stack space outside the loop and only setting each of these values once inside the loop.

database/ffldb/blockio.go
```diff
@@ -713,36 +720,61 @@ func (s *blockStore) handleRollback(oldBlockFileNum, oldBlockOffset uint32) {
...
-			break
+			return 0, 0, 0, err
+		}
+		if firstFile == -1 {
+			firstFile = lastFile
 		}
-		lastFile = i
 
-		fileLen = uint32(st.Size())
+		lastFileLen = uint32(st.Size())
```
ProofOfKeags, line 759: Move this outside the loop. This is making a bunch of syscalls where we immediately throw away the result.

ProofOfKeags (changes requested):
There are a couple of things I think are necessary to change like the Prune Cache uint width. There are a number of things that seem fishy to me architecturally but nothing I can point to that's broken about it. I think even with the architectural smells that the benefit of a prunable btcd is worth merging these changes.

blockchain/chain.go
```diff
@@ -114,6 +115,10 @@ type BlockChain struct {
 	// fields in this struct below this point.
 	chainLock sync.RWMutex
 
+	// pruneTarget is the size in bytes the database targets for when the node
+	// is pruned.
+	pruneTarget uint32
```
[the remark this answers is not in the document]
kcalvinalvin: Good catch! Addressed in the latest push

config.go
```diff
@@ -1136,6 +1138,30 @@ func loadConfig() (*config, []string, error) {
 		}
 	}
 
+	if cfg.Prune != 0 && cfg.Prune < pruneMinSize {
+		err := fmt.Errorf("%s: the minimum value for --prune is %d. Got %d",
+			funcName, pruneMinSize, cfg.Prune)
+		fmt.Fprintln(os.Stderr, err)
+		fmt.Fprintln(os.Stderr, usageMessage)
+		return nil, nil, err
+	}
```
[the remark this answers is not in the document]
ProofOfKeags: Seems reasonable

config.go
ProofOfKeags: Yeah I don't mind the intervals at all. It's hardly an issue now that the overall prune cache size isn't maxed at ~4G

ProofOfKeags (approved):

database/ffldb/driver_test.go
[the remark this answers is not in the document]
guggero: It's here: https://github.com/btcsuite/btcd/blob/3faf68c8c57d925baea2bdcbcbc8a7efbbf99080/go.mod#L16, we mainly started using it in the `txscript` and `btcec` packages. It reduces most assertions to just a single line which I think is nice. But non blocking.

config.go
guggero: Ah okay, the two different limits now make sense to me, thanks for the explanation.

guggero (commented):
Amazing work, I think the code looks good to go.

I tested several things on testnet and found no bugs:
 - Sync from empty state to chain tip without pruning
 - Reduce pruning to large value slightly below the actual size (30GB), expect only a few block files to be pruned
 - Reduce pruning to 20GB, expect more block files to be pruned
 - Reduce pruning to min value, expect all but three block files to be pruned
 - Remove pruning again, make sure the node continues working
 - Start new sync from empty state to chain tip with pruning set to minimal value, expect there never to be more than 3 block files to be present


I found one small UX thing that we might want to fix:
When running `btcctl getblockchaininfo` there's a boolean `pruned` that is always set to `false`, even when pruning is turned on. Should be fairly easy to add, so I'm waiting for that before fully approving the PR.

database/ffldb/blockio.go
```diff
@@ -713,36 +720,61 @@ func (s *blockStore) handleRollback(oldBlockFileNum, oldBlockOffset uint32) {
...
-	lastFile := -1
-	fileLen := uint32(0)
-	for i := 0; ; i++ {
-		filePath := blockFilePath(dbPath, uint32(i))
+// find the first file, last file, and the end of the most recent file.  The
+// position at the last file is considered the current write cursor which is
+// also stored in the metadata.  Thus, it is used to detect unexpected shutdowns
+// in the middle of writes so the block files can be reconciled.
+func scanBlockFiles(dbPath string) (
+	firstFile int, lastFile int, lastFileLen uint32, err error) {
```
Roasbeef, line 728: So we typically never used named returns, as for the most part they make code more difficult to follow in line. So in this case, we'd just declare them upfront, or even return a new struct with the field unfolded.

database/ffldb/blockio.go
```diff
@@ -713,36 +720,61 @@ func (s *blockStore) handleRollback(oldBlockFileNum, oldBlockOffset uint32) {
...
+
+	files, err := filepath.Glob(filepath.Join(dbPath, "*"+blockFileExtension))
+	if err != nil {
+		return 0, 0, 0, err
+	}
+	sort.Strings(files)
+
+	lastFile, firstFile = -1, -1
+	for i, file := range files {
```
Roasbeef, line 737: Why do we need to actually _scan_ all the files here? With this new usage of `filepath.Glob` we get all the file names up front, then sort them. So we can just get the first, and the last file directly using indexes. Then this way, we don't need to continue scanning and stat'ing them all after we've found the first and last file above.

database/ffldb/db.go
```diff
@@ -1669,6 +1669,99 @@ func (tx *transaction) writePendingAndCommit() error {
...
+
+	// If the total size of block files are under the target, return early and
+	// don't prune.
+	totalSize := uint64(lastFileSize) + (maxSize * uint64(maxSizeFileCount))
+	if totalSize <= targetSize {
+		return nil, nil
+	}
+
+	log.Tracef("Using %d more bytes than the target of %d MiB. Pruning files...",
+		totalSize-targetSize,
```
Roasbeef, line 1720: As is, doesn't this fail to account for the size of the last file on disk which may not be full? Or is the intended feature contract here that: we support pruning, but you can have `N GB + maxBlockFile` bytes on disk?

database/ffldb/db.go
```diff
@@ -1669,6 +1669,99 @@ func (tx *transaction) writePendingAndCommit() error {
...
+		targetSize/(1024*1024))
+
+	deletedFiles := make(map[uint32]struct{})
+
+	// < not <= so that the last file is never deleted.  There are other checks in place
```
Roasbeef, line 1725: Style nit: can prefix with "We use" so it becomes a complete sentence.

database/ffldb/db.go
```diff
@@ -1669,6 +1669,99 @@ func (tx *transaction) writePendingAndCommit() error {
...
+		}
+	}
+
+	// Delete the indexed block locations for the files that we've just deleted.
+	var deletedBlockHashes []chainhash.Hash
+	cursor := tx.blockIdxBucket.Cursor()
+	for ok := cursor.First(); ok; ok = cursor.Next() {
+		loc := deserializeBlockLoc(cursor.Value())
+
+		_, found := deletedFiles[loc.blockFileNum]
```
Roasbeef, line 1752: Why would there ever be a gap here? Shouldn't we always be deleting contiguous block files?

database/ffldb/driver_test.go
```diff
@@ -253,6 +255,127 @@ func TestPersistence(t *testing.T) {
...
+
+		// Check that all the blocks that say were deleted are deleted from the
+		// block index bucket as well.
+		err = db.View(func(tx database.Tx) error {
+			for _, deletedBlock := range deletedBlocks {
+				_, err := tx.FetchBlock(&deletedBlock)
+				if dbErr, ok := err.(database.Error); !ok ||
+					dbErr.ErrorCode != database.ErrBlockNotFound {
+
+					return fmt.Errorf("Expected ErrBlockNotFound but got %v", dbErr)
```
Roasbeef, line 343: Overflows 80 char columns, a few other instances of this in the diff.

database/ffldb/driver_test.go
```diff
@@ -253,6 +255,127 @@ func TestPersistence(t *testing.T) {
...
+			}
+
+			return nil
+		})
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		var deletedBlocks []chainhash.Hash
+		// This should leave 3 files on disk.
```
Roasbeef, line 318: Style nit: should have a newline above the comment.

database/ffldb/driver_test.go
```diff
@@ -253,6 +255,127 @@ func TestPersistence(t *testing.T) {
...
+		for _, block := range blocks {
+			bytes, err := block.Bytes()
+			if err != nil {
+				t.Fatal(err)
+			}
+			blockHashMap[*block.Hash()] = bytes
+		}
+
+		err = db.Update(func(tx database.Tx) error {
+			_, err := tx.PruneBlocks(1024)
```
Roasbeef, line 305: What about the boundary cases when: try to prune more than allowed, no files on disk, already pruned to target, etc.

blockchain/chain.go
```diff
@@ -667,6 +671,37 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
 	b.sendNotification(NTBlockConnected, block)
 	b.chainLock.Lock()
 
+	// If the pruneTarget isn't 0, we should attempt to delete older blocks
+	// from the database.
+	if b.pruneTarget != 0 {
+		err = b.db.Update(func(dbTx database.Tx) error {
+			// When the total block size is under the prune target, prune blocks is
+			// a no-op and the deleted hashes are nil.
+			deletedHashes, err := dbTx.PruneBlocks(b.pruneTarget)
```
Roasbeef, line 680: Why not run this in the above transaction where all the other indexes are updated?

blockchain/indexers/addrindex.go
```diff
@@ -809,6 +809,16 @@ func (idx *AddrIndex) DisconnectBlock(dbTx database.Tx, block *btcutil.Block,
 	return nil
 }
 
+// PruneBlock is invoked when an older block is deleted after it's been
+// processed.
+// NOTE: For AddrIndex, it's a no-op as AddrIndex isn't allowed to be enabled
```
Roasbeef, line 814: Wouldn't we want to _reclaim_ the space on disk due to the addr index? Or is the idea that you need to first drop the entire thing, _then_ enable pruning?

blockchain/indexers/cfindex.go
Roasbeef: > Does btcwallet ask for the filters(not familiar with the codebase)? Might be worth keeping the filters around for it?

We don't do that _yet_ but it's something we've been meaning on adding.

blockchain/indexers/cfindex.go
```diff
@@ -255,6 +255,37 @@ func (idx *CfIndex) DisconnectBlock(dbTx database.Tx, block *btcutil.Block,
...
+// reindexing as a pruned node.
+//
+// This is part of the Indexer interface.
+func (idx *CfIndex) PruneBlock(dbTx database.Tx, blockHash chainhash.Hash) error {
+	for _, key := range cfIndexKeys {
```
Roasbeef, line 265: We can do a small refactor here to allow this inner function to take the block hash directly, then we can call it for `DisconnectBlock` as well.

server.go
```diff
@@ -44,8 +44,8 @@ import (
 const (
 	// defaultServices describes the default services that are supported by
 	// the server.
-	defaultServices = wire.SFNodeNetwork | wire.SFNodeBloom |
-		wire.SFNodeWitness | wire.SFNodeCF
+	defaultServices = wire.SFNodeNetwork | wire.SFNodeNetworkLimited |
+		wire.SFNodeBloom | wire.SFNodeWitness | wire.SFNodeCF
```
Roasbeef, line 48: If pruning isn't active, then this bit should be removed in `newServer`: 
https://github.com/btcsuite/btcd/blob/85687d455d8cca1077dae5fbf20accd59cb00da9/server.go#L2726-L2733

Roasbeef (changes requested):
Nice work! I love how small the final diff actually is. I had prior versions that attempted to do more fine grained accounting, which weren't anywhere near as clean as this implementation is. 

My main question in the diff is how we should handle the existing indexes already preseent:
  * Should we still allow `txindex` since it allows you to fetch blocks beyond the prune horizon? 
  * Should we attempt to delete the old information for existing indexes, making them partially present, or just wipe the entire thing? 
  * Should we continue to keep the compact filters as well, since they can be used to still serve light clients?

database/ffldb/driver_test.go
kcalvinalvin: My personal preference is that 80 char limit doesn't really make sense in the current world. Even the linux kernel now does a default of 100 chars.

Since the rest of the codebase isn't doing 80 char either, does it make sense to enforce this here? I can wrap it here but I do think that the rule is outdated.

server.go
kcalvinalvin: In Core, `NODE_NETWORK_LIMITED` is always on and `NODE_NETWORK` bit is added if the node isn't pruned.

https://github.com/bitcoin/bitcoin/blob/a4ca4975880c4f870c6047065c70610af2529e74/src/init.cpp#L786

https://github.com/bitcoin/bitcoin/blob/a4ca4975880c4f870c6047065c70610af2529e74/src/init.cpp#L1598C13-L1601

This is reflected in bitnodes.io for archival nodes https://bitnodes.io/nodes/18.27.79.17-8333/

I think it's better to follow the behavior of Core nodes since it's what everyone expects.

blockchain/indexers/addrindex.go
kcalvinalvin: Currently if you have had the addr index enabled, it won't prune the ones you've already indexed unless you do it manually.

I guess we could do either one of these:

1: Don't do anything with the already present addr index data (current)
2: Drop the addr index if pruning is enabled
3: Force the user to drop the addr index before allowing them to enable pruning

I'd favor either 1 or 3. Explicit user action or doing nothing seems better.

database/ffldb/db.go
kcalvinalvin: `N GB + maxBlockFile` is the current implementation.

I didn't give that much thought into it since you'll at max be `512MiB` off. Which wouldn't really change the user experience as the utxo set is much bigger than that.

Since we're already fetching the `lastFileSize`, I could change it so that the last block file length is accounted for in the size limit.

database/ffldb/blockio.go
kcalvinalvin: One thing I had in mind was that you could technically have a `511MiB` `.fdb` file if the last block being written onto the disk doesn't fit.

But yeah you'd save a lot of i/o by just skipping. I'll implement that since that seems better overall.

database/ffldb/db.go
kcalvinalvin: Is the order of blocks being saved onto the disk guaranteed?

Here we're looping over the blockLocations that are mapped by the blockhash, not the actual blocks in the `*.fdb` filess (that happens in the above loop). You can't assume which blocklocation was in which `*.fdb` file so I loop through them and delete them individually.

database/ffldb/driver_test.go
kcalvinalvin: The current test will test for `try to prune more than allowed`.

I'm not sure if `no files on disk` and `already pruned to target` should end in an error? For both cases, it's meeting the requirement of: `sizeOnDisk < targetPruneSize`.

database/ffldb/blockio.go
kcalvinalvin: Addressed in the latest push.

database/ffldb/driver_test.go
kcalvinalvin: Addressed in the latest push

kcalvinalvin:
> * Should we still allow `txindex` since it allows you to fetch blocks beyond the prune horizon?

I'm not sure. We could allow for it and just return empty strings when you attempt to fetch an already pruned block.

> * Should we attempt to delete the old information for existing indexes, making them partially present, or just wipe the entire thing?

Refer to https://github.com/btcsuite/btcd/pull/1971#discussion_r1284037346

> * Should we continue to keep the compact filters as well, since they can be used to still serve light clients?

I'm indifferent either way. The user does have the option to drop the entire thing if they want. Though it may be nice to be able to only keep a portion of it?

blockchain/indexers/addrindex.go
Roasbeef: > 3: Force the user to drop the addr index before allowing them to enable pruning

I prefer this path myself. As otherwise, the queries of indexes like the addr index wouldn't be possible as it wants to return the block regions where the transaction exist, but they may not be accessible after pruning. 

The only index that can still serve its purpose with pruned data is the neutrino filter index, as querying it doesn't require any block data at all (just slinging out the premade filters for each block).

halseth (approved):
LGTM 🦩

guggero (commented):
> When running btcctl getblockchaininfo there's a boolean pruned that is always set to false, even when pruning is turned on. Should be fairly easy to add, so I'm waiting for that before fully approving the PR.

I think this isn't addressed yet. Other than that (and the other remaining comments), this is good to go!

database/ffldb/driver_test.go
Roasbeef: Ah yeah you're right this this didn't make it into the initial version of our [code contribution guidelines](https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md) but it _has_ been adhered to as a standard nevertheless. I'll seek to have it updated so we can also use the linter to flag instances automatically in the future.

database/ffldb/db.go
```diff
@@ -1669,6 +1669,99 @@ func (tx *transaction) writePendingAndCommit() error {
...
+		if found {
+			deletedBlockHashes = append(deletedBlockHashes, *(*chainhash.Hash)(cursor.Key()))
+			err := cursor.Delete()
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+
```
Roasbeef, line 1761: I think let's also log here the new size once the pruning has been completed? That way the logs can be used approximate how long the pruning step took: 
```
2023-08-15 17:24:25.031 [TRC] BCDB: Scan found latest block file #54 with length 215346498
2023-08-15 17:24:25.031 [TRC] BCDB: Using 23963495746 more bytes than the target of 5000 MiB. Pruning files...
2023-08-15 17:25:08.715 [DBG] CHAN: Accepted block 0000000000000008a838becbe29d3a29fe89e291b8ad18fdf9021f8024585b2d
```

guggero (approved):

database/ffldb/driver_test.go
kcalvinalvin: Addressed in the latest push

kcalvinalvin:
Made some additional changes:

1: Correctly return the prune status for `getblockchaininfo`
2: Got rid of pruning code for indexers. Compact filters are no longer pruned when the blocks are deleted.
3: Force user to drop already existing addr and/or tx index when trying to enable pruning.
4: Error if user enables addr and/or tx index while pruned.
5: Error if user tries to enable cfindex if they started the ibd while pruned and cfindex disabled. 
5: Force user to drop already indexed cfindex data when trying to disable cfindex while pruned.

kcalvinalvin:
The last push just adds a more descriptive error message if the user tries to disable `cfindex` while the node is already pruned and has cfindex enabled.

kcalvinalvin:
Sorry for yet another additional push. 3 things here.

1) The biggest thing: I realized that a lot of the code depends on `cfg.Prune` to tell if the node is pruned or not. But the code allowed the user to omit the flag even if the node had been previously pruned. This would've resulted in broadcasting to peers that the node isn't pruned and also would've given the wrong prune status for `getblockchaininfo`.

Instead of trying to see every case where I'd also have to call the database for the prune status, I added a commit that forces the user to prune if the database had been pruned previously like bitcoind. I don't think this really affects the UX and negates future bugs as well that can arise from not checking db for the prune status and just relying on `cfg.Prune`.

2) `getblockchaininfo` now depends on `cfg.Prune != 0`
3) Added additional message for the `--prune` flag on `-h` that tells the user about the default value of `cfg.Prune == 0` and how that disables pruning.

Roasbeef:
>  This would've resulted in broadcasting to peers that the node isn't pruned

Can you elaborate on this? Broadcasting as in the set of services bits we advertise, or something else? IIUC, those service bits will always give away if a node has been pruned or not.

Roasbeef (approved):
LGTM 🎏

+1 to the additional safeguarding around user configuration/downgrades

PR #1972, opened 2023-04-28
blockchain: Use slices when fetching utxos

kcalvinalvin:
Maps have a higher overhead than slices.  As slices can be used instead of maps, we avoid the overhead of making a map.

This was one of the overheads I've noticed when profiling ibd for PR #1955

### DIFF
--- a/blockchain/utxoviewpoint.go
+++ b/blockchain/utxoviewpoint.go
@@ -503,7 +503,7 @@ func (view *UtxoViewpoint) commit() {
 // Upon completion of this function, the view will contain an entry for each
 // requested outpoint.  Spent outputs, or those which otherwise don't exist,
 // will result in a nil entry in the view.
-func (view *UtxoViewpoint) fetchUtxosMain(db database.DB, outpoints map[wire.OutPoint]struct{}) error {
+func (view *UtxoViewpoint) fetchUtxosMain(db database.DB, outpoints []wire.OutPoint) error {
 	// Nothing to do if there are no requested outputs.
 	if len(outpoints) == 0 {
 		return nil
@@ -517,13 +517,13 @@ func (view *UtxoViewpoint) fetchUtxosMain(db database.DB, outpoints map[wire.Out
 	// so other code can use the presence of an entry in the store as a way
 	// to unnecessarily avoid attempting to reload it from the database.
 	return db.View(func(dbTx database.Tx) error {
-		for outpoint := range outpoints {
-			entry, err := dbFetchUtxoEntry(dbTx, outpoint)
+		for i := range outpoints {
+			entry, err := dbFetchUtxoEntry(dbTx, outpoints[i])
 			if err != nil {
 				return err
 			}
 
-			view.entries[outpoint] = entry
+			view.entries[outpoints[i]] = entry
 		}
 
 		return nil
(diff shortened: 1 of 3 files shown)


blockchain/utxoviewpoint.go
```diff
@@ -517,7 +517,7 @@ func (view *UtxoViewpoint) fetchUtxosMain(db database.DB, outpoints map[wire.Out
 	// so other code can use the presence of an entry in the store as a way
 	// to unnecessarily avoid attempting to reload it from the database.
 	return db.View(func(dbTx database.Tx) error {
-		for outpoint := range outpoints {
+		for _, outpoint := range outpoints {
```
guggero, line 520: Should we use index based access to the slice to avoid any Golang loop variable issues?

blockchain/utxoviewpoint.go
```diff
@@ -626,23 +626,23 @@ func (b *BlockChain) FetchUtxoView(tx *btcutil.Tx) (*UtxoViewpoint, error) {
 	// Create a set of needed outputs based on those referenced by the
 	// inputs of the passed transaction and the outputs of the transaction
 	// itself.
-	neededSet := make(map[wire.OutPoint]struct{})
+	needed := make([]wire.OutPoint, 0, len(tx.MsgTx().TxIn))
```
guggero, line 629: We're pre-allocating for the length of `TxIn` but then loop over `TxOut`.

blockchain/utxoviewpoint.go
kcalvinalvin: I can change it to index based access.

Could you let me know what the Golang loop variable issues are? I'm not aware of these issues so it'd be good to know.

kcalvinalvin:
Addressed the reviews in the latest push

blockchain/utxoviewpoint.go
guggero: The variable is defined per-loop, not per-iteration. So depending on how it is used, it can lead to unexpected side effects. See https://mohamed-abdel-mohaimen.medium.com/using-loop-iterator-variable-issues-in-golang-b76c17fb71b6 or similar articles.

kcalvinalvin:
Pushed an extra commit with benchmarks.

I'm seeing a ~25% performance improvement.
```
[N] calvin@bitcoin ~/b/g/u/g/s/g/b/b/blockchain (utxoview-use-slices)> go test -run=XXX -bench=BenchmarkUtxo
goos: linux
goarch: amd64
BenchmarkUtxoFetchMap-10        17066674                71.06 ns/op
BenchmarkUtxoFetchSlices-10     19618462                52.60 ns/op
PASS
ok      github.com/btcsuite/btcd/blockchain     2.384s
```

blockchain/utxoviewpoint.go
```diff
@@ -533,25 +533,25 @@ func (view *UtxoViewpoint) fetchUtxosMain(db database.DB, outpoints map[wire.Out
...
-	for outpoint := range outpoints {
+	needed := make([]wire.OutPoint, 0, len(outpoints))
+	for _, outpoint := range outpoints {
 		// Already loaded into the current view.
 		if _, ok := view.entries[outpoint]; ok {
 			continue
 		}
 
-		neededSet[outpoint] = struct{}{}
+		needed = append(needed, outpoint)
```
guggero, line 550: I think this is the classic example of where things go wrong with Golang loop variables. Should use index based access to the `outpoint`.

guggero (changes requested):

blockchain/utxoviewpoint.go
kcalvinalvin: Addressed with the latest push.

guggero (approved):
Nice, LGTM :tada:

blockchain/utxoviewpoint.go
```diff
@@ -533,25 +533,25 @@ func (view *UtxoViewpoint) fetchUtxosMain(db database.DB, outpoints map[wire.Out
...
 	// Nothing to do if there are no requested outputs.
 	if len(outpoints) == 0 {
 		return nil
 	}
 
 	// Filter entries that are already in the view.
-	neededSet := make(map[wire.OutPoint]struct{})
-	for outpoint := range outpoints {
+	needed := make([]wire.OutPoint, 0, len(outpoints))
+	for i := range outpoints {
```
halseth, line 544: nit: could do `for _, outpoint := range outpoints` to minimize diff

halseth (approved):
Great change, LGTM 💯

blockchain/utxoviewpoint.go
guggero: No, I explicitly asked to use the index based access, because the `append()` would otherwise not work as expected because of the Golang loop variable issue (see my resolved comments above).

halseth: I don't think that is an issue in this case, since it's a non-ref struct being used?

That being said, better safe than sorry 😀

guggero: Hmm, you're right, in this case it's not problematic. But we've been burned by this a couple of times in the past, so I basically only use the key/index based access. Also makes it easier to review if you know the loop bug can't bite you.

PR #1979, opened 2023-05-09
blockchain, integration, mining, main: Rolling merkle root calculation

kcalvinalvin:
I saw #1421 and noticed that it was slower than the current function for calculating merkle roots. Just looking to see if the direction I took here is ok.

I replaced everything in `rolling_merkle.go` with Utreexo structs/functions. The comments and struct names should probably change. FWIW, the main function `add` is using the same algorithm as the one found in blake3 hash.

Here's the benchstat results. It's about the same/slightly faster than the current merkle root function and allocates even less memory than the function in #1421. The added code is a lot smaller as well.

```
[I] calvin@bitcoin ~/b/g/u/g/s/g/b/b/blockchain (merkle-calc-fast)> benchstat merkle.txt rolling_merkle.txt
goos: linux
goarch: amd64
                │  merkle.txt  │         rolling_merkle.txt          │
                │    sec/op    │    sec/op     vs base               │
Merkle/1000-10    539.0µ ±  2%   542.6µ ±  6%       ~ (p=0.165 n=10)
Merkle/2000-10    1.081m ±  1%   1.078m ±  1%       ~ (p=0.436 n=10)
Merkle/4000-10    2.189m ± 29%   2.180m ±  2%       ~ (p=0.315 n=10)
Merkle/8000-10    4.410m ±  6%   4.352m ±  1%       ~ (p=0.105 n=10)
Merkle/16000-10   8.764m ±  2%   8.859m ± 34%       ~ (p=0.247 n=10)
Merkle/32000-10   17.60m ±  5%   17.29m ±  2%  -1.78% (p=0.007 n=10)
geomean           3.088m         3.078m        -0.34%

                │   merkle.txt   │         rolling_merkle.txt         │
                │      B/op      │    B/op     vs base                │
Merkle/1000-10      48416.0 ± 0%   320.0 ± 0%  -99.34% (p=0.000 n=10)
Merkle/2000-10      96800.0 ± 0%   352.0 ± 0%  -99.64% (p=0.000 n=10)
Merkle/4000-10     193568.0 ± 0%   384.0 ± 0%  -99.80% (p=0.000 n=10)
Merkle/8000-10     387104.0 ± 0%   416.0 ± 0%  -99.89% (p=0.000 n=10)
Merkle/16000-10    774176.0 ± 0%   448.0 ± 0%  -99.94% (p=0.000 n=10)
Merkle/32000-10   1548320.0 ± 0%   480.0 ± 0%  -99.97% (p=0.000 n=10)
geomean             267.3Ki        396.2       -99.86%

                │   merkle.txt   │         rolling_merkle.txt          │
                │   allocs/op    │ allocs/op   vs base                 │
Merkle/1000-10     1002.000 ± 0%   1.000 ± 0%   -99.90% (p=0.000 n=10)
Merkle/2000-10     2002.000 ± 0%   1.000 ± 0%   -99.95% (p=0.000 n=10)
Merkle/4000-10     4002.000 ± 0%   1.000 ± 0%   -99.98% (p=0.000 n=10)
Merkle/8000-10     8002.000 ± 0%   1.000 ± 0%   -99.99% (p=0.000 n=10)
Merkle/16000-10   16002.000 ± 0%   1.000 ± 0%   -99.99% (p=0.000 n=10)
Merkle/32000-10   32002.000 ± 0%   1.000 ± 0%  -100.00% (p=0.000 n=10)
geomean              5.661k        1.000        -99.98%
```


kcalvinalvin:
There was still some Utreexo specific stuff that made it calculate wrong roots. I fixed it in the latest force push.

Still the same results from the benchmarks so the benchstat results are valid.

Roasbeef:
Concept ACK, great work building on the other optimization with some of the utreexo derived fine tuning!

kcalvinalvin:
Ready for reviews now!

@Roasbeef

blockchain/rolling_merkle.go
```diff
@@ -0,0 +1,149 @@
...
+import (
+	"crypto/sha256"
+	"math/bits"
+
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+)
+
+// parentHash returns the hash of the left and right hashes passed in.
+func parentHash(l, r chainhash.Hash) chainhash.Hash {
```
yyforyongyu, line 12: So this is the same as `HashMerkleBranches`?

blockchain/rolling_merkle.go
```diff
@@ -0,0 +1,149 @@
...
+	// numLeaves is the total leaves the store has processed.  numLeaves
+	// is required for the root calculation algorithm to work.
+	numLeaves uint64
+}
+
+// newRollingMerkleTreeStore returns a rollingMerkleTreeStore with the roots
+// allocated based on the passed in size.
+func newRollingMerkleTreeStore(size int) rollingMerkleTreeStore {
+	var alloc int
+	if size != 0 {
```
yyforyongyu, line 40: Can the size be 0?

blockchain/rolling_merkle.go
```diff
@@ -0,0 +1,149 @@
...
+		alloc = bits.Len64(uint64(size) - 1)
+	}
+	return rollingMerkleTreeStore{roots: make([]chainhash.Hash, 0, alloc)}
+}
+
+// add adds a single hash to the merkle tree store.  Refer to algorithm 1 "AddOne" in
+// the utreexo paper (https://eprint.iacr.org/2019/611.pdf) for the exact algorithm.
+func (s *rollingMerkleTreeStore) add(add chainhash.Hash) {
```
yyforyongyu, line 48: Missing unit tests, which is helpful for understanding this method.

blockchain/rolling_merkle.go
```diff
@@ -0,0 +1,149 @@
...
+	// We can tell where the roots are by looking at the binary representation
+	// of the numLeaves. Wherever there's a 1, there's a root.
+	//
+	// numLeaves of 8 will be '1000' in binary, so there will be one root at
+	// row 3. numLeaves of 3 will be '11' in binary, so there's two roots. One at
+	// row 0 and one at row 1.
+	//
+	// In this loop below, we're looking for these roots by checking if there's
```
yyforyongyu, line 56: We should mention we are looking for the `1` from **right to left**, and stops when seeing a `0`.

blockchain/rolling_merkle.go
```diff
@@ -0,0 +1,149 @@
...
+		case witness && i == 0:
+			var zeroHash chainhash.Hash
+			s.add(zeroHash)
+		case witness:
+			s.add(*adds[i].WitnessHash())
+		default:
+			s.add(*adds[i].Hash())
+		}
+	}
+
```
yyforyongyu, line 91: Strickly speaking, after this line, we should no longer check `adds` but rely on `s.numLeaves` since the tree state is already updated by `adds`. Later on we may have to use empty leaves, while it's not related to `adds` anymore.

blockchain/rolling_merkle.go
```diff
@@ -0,0 +1,149 @@
...
+	roots []chainhash.Hash
+
+	// numLeaves is the total leaves the store has processed.  numLeaves
+	// is required for the root calculation algorithm to work.
+	numLeaves uint64
+}
+
+// newRollingMerkleTreeStore returns a rollingMerkleTreeStore with the roots
+// allocated based on the passed in size.
+func newRollingMerkleTreeStore(size int) rollingMerkleTreeStore {
```
yyforyongyu, line 38: Use `uint64` here?

blockchain/rolling_merkle.go
```diff
@@ -0,0 +1,149 @@
...
+
+		// Calculate the hash of the new root and append it.
+		newRoot = parentHash(root, newRoot)
+	}
+	s.roots = append(s.roots, newRoot)
+	s.numLeaves++
+}
+
+// calcMerkleRoot returns the merkle root for the passed in transactions.
+func (s *rollingMerkleTreeStore) calcMerkleRoot(adds []*btcutil.Tx, witness bool) chainhash.Hash {
```
yyforyongyu, line 75: Also need unit tests.

blockchain/merkle_test.go
```diff
@@ -21,3 +24,89 @@ func TestMerkle(t *testing.T) {
...
+		32000,
+	}
+
+	txs := make([][]*btcutil.Tx, len(sizes))
+	for i, size := range sizes {
+		txs[i] = makeTxs(size)
+	}
+
+	for i, size := range sizes {
+		name := fmt.Sprintf("%d", size)
```
yyforyongyu, line 72: nit: could do `txs = makeTxs(size)` here instead, and same below.

blockchain/merkle_test.go
```diff
@@ -16,12 +16,20 @@ import (
 // TestMerkle tests the BuildMerkleTreeStore API.
 func TestMerkle(t *testing.T) {
 	block := btcutil.NewBlock(&Block100000)
-	merkles := BuildMerkleTreeStore(block.Transactions(), false)
-	calculatedMerkleRoot := merkles[len(merkles)-1]
+	calcMerkleRoot := CalcMerkleRoot(block.Transactions(), false)
+	merkleStoreTree := BuildMerkleTreeStore(block.Transactions(), false)
+	merkleStoreRoot := merkleStoreTree[len(merkleStoreTree)-1]
+
+	if calcMerkleRoot != *merkleStoreRoot {
```
yyforyongyu, line 23: use `require.Equal` instead?

yyforyongyu (changes requested):
Dope🤙 Excited to see this performance gain! Left some comments, mostly about the new algo.

blockchain/rolling_merkle.go
kcalvinalvin: Oh yeah you're right, it serves the same functionality.

Though the `parentHash` function is faster because there's no allocation being done vs `HashMerkleBranches` which does a single allocation. I've tested and verified that this is because `HashMerkleBranches` is returning a pointer.

Would it be ok if I were to change the `HashMerkleBranches` function to return a `chainhash.Hash` instead of a pointer? Then `parentHash` and `HashMerkleBranches` would be equal.

```
BenchmarkHashMerkleBranches-10           2279146               526.5 ns/op            32 B/op          1 allocs/op
BenchmarkParentHash-10                   2392340               497.1 ns/op             0 B/op          0 allocs/op
```

I'll modify the 

Code used for the above benchmark:

```
func BenchmarkHashMerkleBranches(b *testing.B) {
       var aHash, bHash chainhash.Hash
       for i := 0; i < b.N; i++ {
               HashMerkleBranches(&aHash, &bHash)
       }
}

func BenchmarkParentHash(b *testing.B) {       
       var aHash, bHash chainhash.Hash
       for i := 0; i < b.N; i++ {
               parentHash(aHash, bHash)
       }
}
```

blockchain/rolling_merkle.go
kcalvinalvin: It can't be 0 since this is only called in `CalcMerkleRoot` and the passed in size is `len(transactions)` in `CalcMerkleRoot`.

I put this check here since if the code were to be called elsewhere with 0, there'd be an overflow. It's not too bad since you'd be allocating `64` for `[]chainhash.Hash` but I thought it'd be better not to allow for that.

I can remove the branch if need be.

blockchain/rolling_merkle.go
```diff
@@ -0,0 +1,149 @@
...
+		}
+	}
+
+	// If we only have one tx, then the hash of that tx is the merkle root.
+	if len(adds) == 1 {
+		return s.roots[0]
+	}
+
+	// Add on the last tx again if there's an odd number of txs.
+	if len(adds) > 0 && len(adds)&1 == 1 {
```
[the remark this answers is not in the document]
kcalvinalvin: Seems like bitcompare is slightly faster. It's almost the same. Is the modulo more readable? I could change it here if so.

Benchstat results:

```
goos: linux
goarch: amd64
       │ modulo.txt  │           bitcompare.txt           │
Odd-10   2.478µ ± 3%   2.464µ ± 2%  -0.59% (p=0.041 n=10)
```

Code used:

```
+var varbool bool
+
+func BenchmarkOdd(b *testing.B) {
+       for i := 0; i < b.N; i++ {
+               benchmarkOdd(1_000)
+       }
+}
+
+func benchmarkOdd(n int) {
+       for i := 0; i < n; i++ {
+               // varbool = 2&1 == 1
+               varbool = 2%2 != 0
+       }
+}
```

blockchain/merkle_test.go
kcalvinalvin: addressed in the latest push

blockchain/merkle_test.go
kcalvinalvin: addressed in the latest push

blockchain/rolling_merkle.go
kcalvinalvin: addressed in the latest push

blockchain/rolling_merkle.go
kcalvinalvin: addressed in the latest push.

Though could you reiterate what you meant by

> Later on we may have to use empty leaves, while it's not related to adds anymore.

Didn't quite catch what you meant by empty leaves

blockchain/rolling_merkle.go
kcalvinalvin: Addressed in the latest push

blockchain/rolling_merkle.go
kcalvinalvin: Addressed in the latest push

blockchain/rolling_merkle.go
kcalvinalvin: `CalcMerkleRoot` test is essentially providing unit tests for this as well since `CalcMerkleRoot` is just:

```
s := newRollingMerkleTreeStore(uint64(len(transactions)))                                                                                                                                                                                                                                        
return s.calcMerkleRoot(transactions, witness) 
```

I could add more unit tests to `CalcMerkleRoot`

ProofOfKeags: This API kind of makes me uncomfortable. You can use any rollingMerkleTreeStore as a call target but this only really works if that store was the unit store value or if it has only been seeded with Tx's thus far. I'd strongly recommend constructing the store in this function, or changing the adds vector to be the same as the type of `add` for the store.

blockchain/rolling_merkle.go
```diff
@@ -0,0 +1,149 @@
...
+		}
+	}
+
+	// If we only have one leaf, then the hash of that tx is the merkle root.
+	if s.numLeaves == 1 {
+		return s.roots[0]
+	}
+
+	// Add on the last tx again if there's an odd number of txs.
+	if len(adds) > 0 && len(adds)&1 == 1 {
```
ProofOfKeags, line 98: There doesn't seem to be anything conceptually different about these steps than the loop that follows below other than the fact that it is compensating for the APIs differing between Tx and Hash. I believe this function would be better served to be totally unaware of any Bitcoin mechanics and just operate on hashes. Instead make the transformation from []Tx to []Hash prior to extracting the merkle root from one of these stores.

blockchain/merkle.go
```diff
@@ -154,6 +154,36 @@ func BuildMerkleTreeStore(transactions []*btcutil.Tx, witness bool) []*chainhash
...
+//	  h12 = h(h1 + h2)            h34 = h(h3 + h4)
+//	   /            \              /            \
+//	h1 = h(tx1)  h2 = h(tx2)    h3 = h(tx3)  h4 = h(tx4)
+//
+// The additional bool parameter indicates if we are generating the merkle tree
+// using witness transaction id's rather than regular transaction id's. This
+// also presents an additional case wherein the wtxid of the coinbase transaction
+// is the zeroHash.
+func CalcMerkleRoot(transactions []*btcutil.Tx, witness bool) chainhash.Hash {
+	s := newRollingMerkleTreeStore(uint64(len(transactions)))
```
ProofOfKeags, line 183: I think you should also do the transformation from []Tx to []Hash here to preserve the API. If you are concerned about allocations you can make the `calcMerkleRoot` function accept []Hashable or something to that effect.

ProofOfKeags (changes requested):
Overall I really like this PR. I do think there are certain API changes that are necessary for the RMTS to be broadly useful, rather that solely tailored to the block tree verification process. As I understand it, this PR is ultimately in service of [this](https://github.com/btcsuite/btcd/pull/1955) which is for optimizing sync times, so I am well aware that API ergonomics may not have been an original design goal. I do think that we can take this opportunity to have a broadly useful piece of machinery, though.

The specific changes I am requesting are to clean up the `calcMerkleRoot` function so that it doesn't have special provisioning for Bitcoin specific structures and correspondingly allows you to clean up the internal implementation logic and simplifying redundant checks. Everything else I've left comments wise is non-blocking or adds to the documentation to help future readers of this code.

kcalvinalvin:
> I do think there are certain API changes that are necessary for the RMTS to be broadly useful, rather that solely tailored to the block tree verification process.

I strongly disagree making this function more broadly useful as the merkle tree function used for calculating the tx commitment is flawed and has a vulnerability (CVE-2012-2459). It's possible to create a collision using duplicate leaves and it's described in detail at: https://github.com/bitcoin/bitcoin/blob/79e8247ddb166f9b980f40249b7372a502402a4d/src/consensus/merkle.cpp#L8-L41

If anyone wants to use the following code for merkle tree stuff, they should just use https://github.com/utreexo/utreexo as it's essentially the same.

blockchain/rolling_merkle.go
```diff
@@ -0,0 +1,149 @@
...
+	// In this loop below, we're looking for these roots by checking if there's
+	// a '1', starting from the LSB.  If there is a '1', we'll hash the root being
+	// added with that root until we hit a '0'.
+	newRoot := add
+	for h := uint8(0); (s.numLeaves>>h)&1 == 1; h++ {
+		// Grab the last root.
+		root := s.roots[len(s.roots)-1]
+
+		// Pop off the last root.
+		s.roots = s.roots[:len(s.roots)-1]
```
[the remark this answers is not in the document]
kcalvinalvin: Sure. Addressed in the latest push

blockchain/rolling_merkle.go
```diff
@@ -0,0 +1,149 @@
...
+	// of the numLeaves.  Wherever there's a 1, there's a root.
+	//
+	// numLeaves of 8 will be '1000' in binary, so there will be one root at
+	// row 3. numLeaves of 3 will be '11' in binary, so there's two roots.  One at
+	// row 0 and one at row 1.
+	//
+	// In this loop below, we're looking for these roots by checking if there's
+	// a '1', starting from the LSB.  If there is a '1', we'll hash the root being
+	// added with that root until we hit a '0'.
+	newRoot := add
```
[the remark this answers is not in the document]
kcalvinalvin: > to be the root of the tallest subtree we are capable of fully computing

This doesn't sound quite right to me because it implies that there can be multiple roots that satisfy this condition. The algorithm used only results in 1 specific root that can be calculated when adding a leaf at any given leaf count.

How's this:

```
newRoot is the Hash we are adding.  If there's a sibling root we can hash with in
any given row, it'll be hashed together to form a new root a row above.
```

kcalvinalvin (commented):
Changed a few things. Left comments on comment changes.

blockchain/rolling_merkle.go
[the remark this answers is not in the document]
ProofOfKeags: Perhaps? I'm not too opinionated about what gets added here. Nothing is strictly necessary but the way I imagined it is that at any moment we have a collection of fully computable subtrees that we are holding onto that can be combined with future data. The first time we encounter a 0 is when we can no longer compute the roots of a taller subtree and so we must pause this partial computation and await more data. The other reason I put this comment in is that mutating values (which seems to be especially common in go idiom) harms readability if the name of the variable we use differs substantially from its role over its entire lifetime, this comment was also there to try and fill that gap.

blockchain/rolling_merkle.go
```diff
@@ -0,0 +1,149 @@
...
+	roots []chainhash.Hash
+
+	// numLeaves is the total leaves the store has processed.  numLeaves
+	// is required for the root calculation algorithm to work.
+	numLeaves uint64
+}
+
+// newRollingMerkleTreeStore returns a rollingMerkleTreeStore with the roots
+// allocated based on the passed in size.
+func newRollingMerkleTreeStore(size uint64) rollingMerkleTreeStore {
```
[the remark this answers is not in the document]
ProofOfKeags: oops I made a mistake and confused the default behavior of go slices with rust's vectors. It looks like this comment is entirely wrong and you should ignore it completely. On another node, I think it would be better to make the merkle store resilient to overgrowing its initial size and have the cost be performance rather than a runtime error. Thoughts?

blockchain/rolling_merkle.go
```diff
@@ -0,0 +1,149 @@
...
+
+// rollingMerkleTreeStore calculates the merkle root by only allocating O(logN)
+// memory where N is the total amount of leaves being included in the tree.
+type rollingMerkleTreeStore struct {
+	// roots are where the temporary merkle roots get stored while the
+	// merkle root is being calculated.
+	roots []chainhash.Hash
+
+	// numLeaves is the total leaves the store has processed.  numLeaves
+	// is required for the root calculation algorithm to work.
```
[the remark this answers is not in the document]
ProofOfKeags: Right but you never clear the values in the roots slice afaict so even though there is data there, any root value that corresponds to an *internal* zero is conceptually garbage with respect to the algorithm here. My comment about truncating the trailing zeroes is indeed wrong though because 1000 still has 3 garbage values in the roots slice. I went to verify go's slice popping behavior, and changing the length the way you do in the loop doesn't zero the memory of what was there before. Maybe you think that including this detail in the comments is unnecessary, but I am fairly confident that the claims I make in this comment are correct.

blockchain/rolling_merkle.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// newRollingMerkleTreeStore returns a rollingMerkleTreeStore with the roots
+// allocated based on the passed in size.
+func newRollingMerkleTreeStore(size uint64) rollingMerkleTreeStore {
+	var alloc int
+	if size != 0 {
+		alloc = bits.Len64(size - 1)
+	}
+	return rollingMerkleTreeStore{roots: make([]chainhash.Hash, 0, alloc)}
```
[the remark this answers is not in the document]
ProofOfKeags: Sorry I am new to the specifics of go, I was unclear on the fact that

```go
var alloc int
```
is implicitly
```go
var alloc int = 0
```

blockchain/rolling_merkle.go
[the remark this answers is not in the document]
ProofOfKeags: Actually upon further inspection it appears I was right. The `append` function in slices will do a reallocation/copy if you overrun the capacity. So the reservedSize argument is merely a performance optimization. My original comment here is correct.

blockchain/rolling_merkle.go
```diff
@@ -0,0 +1,149 @@
...
+	h.Reset()
+	h.Write(first)
+	return *((*chainhash.Hash)(h.Sum(buf)))
+}
+
+// rollingMerkleTreeStore calculates the merkle root by only allocating O(logN)
+// memory where N is the total amount of leaves being included in the tree.
+type rollingMerkleTreeStore struct {
+	// roots are where the temporary merkle roots get stored while the
+	// merkle root is being calculated.
```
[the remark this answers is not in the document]
ProofOfKeags: Reading the code, assuming indices closest to zero means "left" and indices of the largest value are "right", It is the short trees that appear furthest to the *right*

blockchain/merkle.go
ProofOfKeags: I do not believe the CVE you cite is relevant here. There is nothing that ensures the uniqueness of the Tx values in the slice being inputted into this function so nothing is actually prevented by this choice of API. Conflating security checks and data structure computations is not something I would recommend. If you need to do security checks, then do them, but don't make it part of the data structure API.

blockchain/rolling_merkle.go
ProofOfKeags: Again, I think the only reason this isn't a singular loop is because you are combining the concerns of Bitcoin specific digest calculations and the actual extraction of a full merkle root. The hash duplication operation is the same at all heights of the calculation, irrespective of whether or not the hash is a leaf hash or not. The fact that this function needs a leaf case doesn't make a lot of sense if you treat it as a full hash tree. At that point you only have your base case which is if there's only one root.

blockchain/rolling_merkle.go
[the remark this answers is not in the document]
kcalvinalvin: > Actually upon further inspection it appears I was right. The `append` function in slices will do a reallocation/copy if you overrun the capacity. So the reservedSize argument is merely a performance optimization. My original comment here is correct.

No the function will never allocate more than `log2(size)`, which is what the function is doing in newRollingMerkleTreeStore. You can try tweaking the benchmarking code, you'll only ever get 1 allocation per merkle root calculation.

blockchain/rolling_merkle.go
kcalvinalvin: > Again, I think the only reason this isn't a singular loop is because you are combining the concerns of Bitcoin specific digest calculations and the actual extraction of a full merkle root. The hash duplication operation is the same at all heights of the calculation, irrespective of whether or not the hash is a leaf hash or not. The fact that this function needs a leaf case doesn't make a lot of sense if you treat it as a full hash tree. At that point you only have your base case which is if there's only one root.

Could you write the code change you suggest here? Asking as simply getting rid of lines 95-103 won't work (as in that it won't calculate the correct roots) and am unsure what you mean.

blockchain/rolling_merkle.go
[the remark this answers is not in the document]
ProofOfKeags: I'm not sure what you're saying at this point. Either it does allocate when it overruns or it doesn't. I'm asserting here that it does. It seems you agree with me. I'm not challenging your benchmarks. I'm not saying your code is slow. I'm merely documenting the behavior I am inferring from the code, which is that if you underreserve the capacity you end up using when creating a new RMTS, it will gracefully allocate more memory to accommodate that. My goal with this entire review is to make sure that the code is accessible to others who are trying to read it. This was a non-obvious property of the code that I figured I'd describe more thoroughly in the comments.

blockchain/rolling_merkle.go
ProofOfKeags: Yeah I'll PR into your branch the change that I think would simplify this.

blockchain/merkle.go
ProofOfKeags: > CalcMerkleRoot is only relevant for merkle root that's committed to the block header and should not be used for anything else because of the vulnerability.

My point here is that your choice to use a Tx slice doesn't actually defend against the vuln. You can simply duplicate the last Tx at the end of the slice and trigger the issue. I looked over this yesterday and didn't see any operation in this implementation where you assert the uniqueness of every transaction in the slice, which is what would be required to sidestep the vulnerability you cite. That tells me that this check has to be done by the caller, and in that case the Tx slice doesn't buy you anything other than an intentionally more constrained API.

> Is there any use case that you have in mind that would require a slice of hashes instead of slice of txs?

in the context of btcd? no. But my point here has always been that this is a broadly useful piece of machinery and Merkle Trees are very commonly used throughout the ecosystem.

If the *only* purpose of this PR is to make btcd block sync fast, then it's probably fine as is. I just think we should take a little bit longer of a view here.

kcalvinalvin: > other than an intentionally more constrained API.

Yes. That's desirable.

> But my point here has always been that this is a broadly useful piece of machinery and Merkle Trees are very commonly used throughout the ecosystem.

This specific implementation is not. No one should be using this. Satoshi coded this in so we're stuck with it.

blockchain/rolling_merkle.go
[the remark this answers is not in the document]
kcalvinalvin: > I'm not sure what you're saying at this point

I'm frustrated because it seems like you've not even run the code and still stating incorrect statements.

> I'm merely documenting the behavior I am inferring from the code

You've incorrectly inferred it.

> which is that if you underreserve the capacity you end up using when creating a new RMTS, it will gracefully allocate more memory to accommodate that. 

That never happens in this implementation. `newRollingMerkleTreeStore` will *always* allocate enough.

> My goal with this entire review is to make sure that the code is accessible to others who are trying to read it.

That's good! I'm stating that your change here is gonna confuse readers more because it's incorrect.

blockchain/merkle.go
ProofOfKeags: > Yes. That's desirable.

I don't understand how a more constrained API, where you don't perform the security checks required to actually dodge the CVE you cite, is desirable.

> This specific implementation is not. No one should be using this. Satoshi coded this in so we're stuck with it.

My understanding of the purpose of this implementation is to compute merkle roots in O(log(n)) memory instead of O(n). What am I missing here? I don't understand what you mean by "Satoshi coded *this* in", what is "this"?

ProofOfKeags: FYI my comment here is non-blocking

blockchain/rolling_merkle.go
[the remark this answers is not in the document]
ProofOfKeags: FYI sorting out the left vs right thing is blocking as I think it is critical for people to understand the organization here.

blockchain/rolling_merkle.go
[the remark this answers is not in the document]
ProofOfKeags: FYI this thread is nonblocking

blockchain/rolling_merkle.go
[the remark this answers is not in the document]
ProofOfKeags: > I'm frustrated because it seems like you've not even run the code and still stating incorrect statements.

I have run the benchmarks you cite. However, regardless of whether or not I have, the statements I am making in this thread are *independent* of the benchmark. My claims are about clarifying the role of the size reservation parameter in this private function. It so happens that you sidestep allocations by making the allocation based off of the size of the Tx vector passed into the public API. This comment, the one I am suggesting editing, is documenting the behavior of the private function, which will, in fact, reallocate if you call `.add` enough times in excess of the initial reserved size.

> That never happens in this implementation. newRollingMerkleTreeStore will always allocate enough.

Only when called by `CalcMerkleRoot`. This comment lives on the constructor/allocator of your data structure though. Even though its API is private and therefore not the concern of the public API users, anyone servicing this code may interact with the private API as they touch this code. The property I state is true of the function that the comment is attached to.

I began this investigation in response to @yyforyongyu 's question about whether or not the size can be zero. I asked myself the question "what would happen if you did call this with a size of 0?". Turns out, all the machinery will still work just fine, it'll just reallocate log(n) times. Not nearly as bad as crashing, so I actually think this is a really good design! I just was trying to specify that this function's API is capable of handling a mis-estimation of this value and the worst that will happen is an extra log(n) allocations.

> I'm stating that your change here is gonna confuse readers more because it's incorrect.

I have once again reviewed the code, because I'm new to go and it's possible I had missed something, but after a thorough investigation I once again am concluding my initial analysis to be correct. Someone else can help adjudicate this dispute if need be but in my estimation, the suggested comment for this function is accurate.

blockchain/rolling_merkle.go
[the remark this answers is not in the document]
ProofOfKeags: This is resolved

blockchain/rolling_merkle.go
ProofOfKeags: FYI this is nonblocking but I still have it on my list to help with this.

blockchain/merkle.go
kcalvinalvin: Won't be addressing this in this PR.

blockchain/rolling_merkle.go
kcalvinalvin: Won't be addressed in this PR

blockchain/rolling_merkle.go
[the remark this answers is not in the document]
kcalvinalvin: Won't be addressing this in this PR

blockchain/rolling_merkle.go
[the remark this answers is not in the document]
kcalvinalvin: Oops, got my right vs left wrong. I'll add the following in.

```
// roots are where the temporary merkle roots get stored while the
// merkle root is being calculated.  Every root has 2^n leaves and the
// tallest tree is furthest to the left and the shortest tree is furthest to
// the right.
```

blockchain/rolling_merkle.go
kcalvinalvin: I won't be addressing this in this PR. There's different ways to do a merkle root calculation and the one implemented here should only be used for calculating merkle root hashes for bitcoin header commitment.

It's a mistake for this to be widely available.

blockchain/rolling_merkle.go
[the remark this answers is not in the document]
kcalvinalvin: added in

> > to be the root of the tallest subtree we are capable of fully computing
> 
> This doesn't sound quite right to me because it implies that there can be multiple roots that satisfy this condition. The algorithm used only results in 1 specific root that can be calculated when adding a leaf at any given leaf count.
> 
> How's this:
> 
> ```
> newRoot is the Hash we are adding.  If there's a sibling root we can hash with in
> any given row, it'll be hashed together to form a new root a row above.
> ```

Will add in the above.

blockchain/rolling_merkle.go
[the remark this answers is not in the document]
kcalvinalvin: Won't be addressing this in this PR.

blockchain/rolling_merkle.go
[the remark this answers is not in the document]
ProofOfKeags: Seems great

blockchain/rolling_merkle.go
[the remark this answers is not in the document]
ProofOfKeags: Excellent

ProofOfKeags (approved):

yyforyongyu (changes requested):
Latest round looks good🎉 Left some comments re style nits and tests. Think once addressed we are good to go🏖

blockchain/rolling_merkle_test.go
```diff
@@ -0,0 +1,128 @@
...
+				{0x02},
+			},
+			expectedRoots: []chainhash.Hash{
+				func() chainhash.Hash {
+					hash, err := chainhash.NewHashFromStr(
+						"c2bf026e62af95cd" +
+							"7b785e2cd5a5f1ec" +
+							"01fafda85886a8eb" +
+							"d34482c0b05dc2c2")
+					if err != nil {
```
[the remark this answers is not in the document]
kcalvinalvin: Which method is desired in `require` for testing that errors are nil?

Something like this?:
`require.Equal(err , nil)`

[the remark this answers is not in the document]
guggero: For errors, there's `require.NoError(t, err)`.

blockchain/rolling_merkle.go
kcalvinalvin: I added a comment on the function in the latest push.

I think it'd be sorta weird to tell the end user why an unexported function is throwing a warning or an info in the logs.

blockchain/rolling_merkle.go
kcalvinalvin: Added a commit to change the `HashMerkleBranches` to return a `chainhash.Hash`. Removed `parentHash`.

blockchain/rolling_merkle.go
```diff
@@ -0,0 +1,152 @@
+package blockchain
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest push

blockchain/rolling_merkle.go
kcalvinalvin: Ok, can I mark this as resolved?

blockchain/rolling_merkle.go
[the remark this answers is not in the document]
kcalvinalvin: Changed to `%2 == 2` in the latest push

blockchain/rolling_merkle_test.go
```diff
@@ -0,0 +1,128 @@
...
+		// 00  (00 is also a root)
+		{
+			leaves: []chainhash.Hash{
+				{0x00},
+			},
+			expectedRoots: []chainhash.Hash{
+				{0x00},
+			},
+		},
+
```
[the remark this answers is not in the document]
kcalvinalvin: Added the test cases in the latest push

blockchain/rolling_merkle_test.go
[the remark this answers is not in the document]
kcalvinalvin: Changed to `require.NoError(t, err)` in the latest push

blockchain/rolling_merkle_test.go
```diff
@@ -0,0 +1,128 @@
...
+		},
+	}
+
+	for _, test := range tests {
+		s := newRollingMerkleTreeStore(uint64(len(test.leaves)))
+		for _, leaf := range test.leaves {
+			s.add(leaf)
+		}
+
+		require.Equal(t, s.roots, test.expectedRoots)
```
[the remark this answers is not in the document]
kcalvinalvin: Added check for `s.numLeaves` in the latest push

yyforyongyu (approved):
LGTM⛵️

Roasbeef (approved):
LGTM 🌭

Roasbeef:
Completed sync tests on mainnet+testnet as a final sanity check!

PR #1980, opened 2023-05-10
txscript/engine: add execution StepCallback

halseth:
We add a new StepCallback as optional function closure on the Engine that will be called every time a step has been performed during script execution.

This is only meant to be used in debugging.

More specifically, this is used for Tapsim: https://github.com/halseth/tapsim

### DIFF
--- a/txscript/engine.go
+++ b/txscript/engine.go
@@ -285,6 +285,33 @@ type Engine struct {
 	witnessProgram  []byte
 	inputAmount     int64
 	taprootCtx      *taprootExecutionCtx
+
+	// stepCallback is an optional function that will be called every time
+	// a step has been performed during script execution.
+	//
+	// NOTE: This is only meant to be used in debugging, and SHOULD NOT BE
+	// USED during regular operation.
+	stepCallback func(*StepInfo) error
+}
+
+// StepInfo houses the current VM state information that is passed back to the
+// stepCallback during script execution.
+type StepInfo struct {
+	// ScriptIndex is the index of the script currently being executed by
+	// the Engine.
+	ScriptIndex int
+
+	// OpcodeIndex is the index of the next opcode that will be executed.
+	// In case the execution has completed, the opcode index will be
+	// incrementet beyond the number of the current script's opcodes. This
+	// indicates no new script is being executed, and execution is done.
+	OpcodeIndex int
+
+	// Stack is the Engine's current content on the stack:
+	Stack [][]byte
+
+	// AltStack is the Engine's current content on the alt stack.
+	AltStack [][]byte
 }
 
 // hasFlag returns whether the script engine instance has the passed flag set.
@@ -1023,6 +1050,17 @@ func (vm *Engine) Step() (done bool, err error) {
 	return false, nil
 }
 
+// copyStack makes a deep copy of the provided slice.
+func copyStack(stk [][]byte) [][]byte {
+	c := make([][]byte, len(stk))
+	for i := range stk {
+		c[i] = make([]byte, len(stk[i]))
+		copy(c[i][:], stk[i][:])
+	}
+
+	return c
+}
+
 // Execute will execute all scripts in the script engine and return either nil
 // for successful validation or an error if one occurred.
 func (vm *Engine) Execute() (err error) {
@@ -1031,6 +1069,22 @@ func (vm *Engine) Execute() (err error) {
 	// will allow for the addition of new scripting languages.
 	if vm.version != 0 {
 		return nil
+	}
+
+	// If the stepCallback is set, we start by making a call back with the
+	// initial engine state.
+	var stepInfo *StepInfo
+	if vm.stepCallback != nil {
+		stepInfo = &StepInfo{
+			ScriptIndex: vm.scriptIdx,
+			OpcodeIndex: vm.opcodeIdx,
+			Stack:       copyStack(vm.dstack.stk),
+			AltStack:    copyStack(vm.astack.stk),
+		}
+		err := vm.stepCallback(stepInfo)
+		if err != nil {
+			return err
+		}
 	}
 
 	done := false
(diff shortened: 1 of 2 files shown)


txscript/engine.go
```diff
@@ -285,6 +285,36 @@ type Engine struct {
...
+	ScriptIndex int
+
+	// OpcodeIndex is the index of the next opcode that will be executed.
+	OpcodeIndex int
+
+	// Witness is the witness of the current input.
+	//
+	// NOTE: this will be only be set before the witness program has been
+	// verified.
+	Witness [][]byte
```
Roasbeef, line 311: Is the witness actually needed? Given to create the VM in the first place, you had to have the witness. 

The other thing is that this field can actually change over time w/ the way execution works: ps2h, nested p2sh, segwit v0, taproot, etc.

Roasbeef (commented):
Super useful! I do something like this at times locally to be able to print out debug information to introspect into a stack when tests fail.

txscript/engine.go
halseth: Yea, you're right. It's not really needed: https://github.com/btcsuite/btcd/pull/1980/commits/d1cb5debf4068853b282ad803e09602b63484301

txscript/engine.go
```diff
@@ -1060,6 +1107,40 @@ func (vm *Engine) Execute() (err error) {
...
+
+		if vm.StepCallback != nil {
+			scriptIdx := vm.scriptIdx
+			opcodeIdx := vm.opcodeIdx
+
+			// In case the execution has completed, we keep the
+			// current script index while increasing the opcode
+			// index. This is to indicate that no new script is
+			// being executed.
+			if done {
```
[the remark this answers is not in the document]
halseth: Not entirely sure what you mean? Move this whole callback step AFTER `CheckErrorCondition` has been called?

txscript/engine.go
```diff
@@ -1060,6 +1100,31 @@ func (vm *Engine) Execute() (err error) {
 
 			return dstr + astr
 		}))
+
+		if vm.StepCallback != nil {
+			scriptIdx := vm.scriptIdx
+			opcodeIdx := vm.opcodeIdx
+
+			// In case the execution has completed, we keep the
```
kcalvinalvin, line 1108: May be good to have this documented in `StepInfo` as well.

txscript/engine.go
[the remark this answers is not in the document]
kcalvinalvin: I think what he means is that we could avoid the `done` branch in this loop if you have the callback inside `CheckErrorCondition` because you know the Engine is done. 

imo how it's implemented here is fine, I think it's better to just have every change in this one function (if what I said above is what roasbeef meant).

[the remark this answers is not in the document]
halseth: Calling `ScriptCallback` from within `CheckErrorCondition` means an additional place where we have to repeat this code. I think doing it like this is simpler tbh.

txscript/engine.go
halseth: Good call: 3033ba39

halseth:
Rebased and addressed comments. PTAL @Roasbeef @kcalvinalvin

kcalvinalvin (commented):
ACK 3033ba391f1a67c171d976624aa2af0c8409fbf2

halseth:
Rebased. PTAL @Roasbeef

txscript/engine.go
```diff
@@ -285,6 +285,33 @@ type Engine struct {
 	witnessProgram  []byte
 	inputAmount     int64
 	taprootCtx      *taprootExecutionCtx
+
+	// StepCallback is an optional function that will be called every time
+	// a step has been performed during script execution.
+	//
+	// NOTE: This is only meant to be used in debugging, and SHOULD NOT BE
+	// USED during regular operation.
+	StepCallback func(*StepInfo) error
```
Roasbeef, line 294: Can be a private variable? In that maybe we introduce a new specific constructor? Like `txscript.NewDebugEngine`.

txscript/engine.go
```diff
@@ -1033,6 +1060,22 @@ func (vm *Engine) Execute() (err error) {
...
 	}
 
+	// If the StepCallback is set, we start by making a call back with the
+	// initial engine state.
+	var stepInfo *StepInfo
+	if vm.StepCallback != nil {
+		stepInfo = &StepInfo{
+			ScriptIndex: vm.scriptIdx,
+			OpcodeIndex: vm.opcodeIdx,
+			Stack:       vm.dstack.stk,
```
Roasbeef, line 1070: Perhaps this should actually be a _copy_ so the callback doesn't inadvertently affect the script execution (mutation, etc). Main application here seems to be debugging or introspection, so doesn't seem to be as sensitive to extra allocations.

Roasbeef: Also for your use case, would it be easier to obtain the string version (like called below), or just leave it with the stack?

Roasbeef (changes requested):
Looking pretty good, left some final comments. I think we should also add a basic test for `txscript.NewDebugEngine` as protection against regressions here. Can be a simple P2PKH spend or something like that.

txscript/engine.go
halseth: Fixed: https://github.com/btcsuite/btcd/pull/1980/commits/55a376e6ea591ba5a6e3ed2302787c95c8135d06

txscript/engine.go
halseth: Fixed: https://github.com/btcsuite/btcd/pull/1980/commits/55a376e6ea591ba5a6e3ed2302787c95c8135d06

It is useful to be able to inspect the raw stack instead of the string version.

halseth:
Rabased.

txscript/engine_debug_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/stretchr/testify/require"
+)
+
+// TestDebugEngine checks that the StepCallbck called during debug script
+// execution contains the expected data.
+func TestDebugEngine(t *testing.T) {
```
Roasbeef, line 18: 👍

Roasbeef (approved):
LGTM 🎿

PR #1985, opened 2023-05-23
blockchain+chaincfg: disable retargeting for regtest

guggero:
This commit emulates the behavior of Bitcoin Core introduced in https://github.com/bitcoin/bitcoin/pull/6853 that disables retargeting of the required proof of work for regtest.

Attempts to fix https://github.com/lightninglabs/neutrino/pull/256 in a less hacky way (needs a PR in that repo as well after this one has been merged).

### DIFF
--- a/blockchain/difficulty.go
+++ b/blockchain/difficulty.go
@@ -219,7 +219,15 @@ func (b *BlockChain) findPrevTestNetDifficulty(startNode *blockNode) uint32 {
 // This function differs from the exported CalcNextRequiredDifficulty in that
 // the exported version uses the current best chain as the previous block node
 // while this function accepts any block node.
-func (b *BlockChain) calcNextRequiredDifficulty(lastNode *blockNode, newBlockTime time.Time) (uint32, error) {
+func (b *BlockChain) calcNextRequiredDifficulty(lastNode *blockNode,
+	newBlockTime time.Time) (uint32, error) {
+
+	// Emulate the same behavior as Bitcoin Core that for regtest there is
+	// no difficulty retargeting.
+	if b.chainParams.PoWNoRetargeting {
+		return b.chainParams.PowLimitBits, nil
+	}
+
 	// Genesis block.
 	if lastNode == nil {
 		return b.chainParams.PowLimitBits, nil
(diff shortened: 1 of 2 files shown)


yyforyongyu (approved):
Thanks for picking this up! Something I also wanted to do as it can save us quite some blocks in `lnd`'s itest!

blockchain/difficulty.go
```diff
@@ -218,7 +221,15 @@ func (b *BlockChain) findPrevTestNetDifficulty(startNode *blockNode) uint32 {
 // This function differs from the exported CalcNextRequiredDifficulty in that
 // the exported version uses the current best chain as the previous block node
 // while this function accepts any block node.
-func (b *BlockChain) calcNextRequiredDifficulty(lastNode *blockNode, newBlockTime time.Time) (uint32, error) {
+func (b *BlockChain) calcNextRequiredDifficulty(lastNode *blockNode,
+	newBlockTime time.Time) (uint32, error) {
+
+	// Emulate the same behavior as Bitcoin Core that for regtest there is
+	// no difficulty retargeting.
+	if b.chainParams.PoWNoRetargeting {
```
kcalvinalvin, line 229: Since this is only desired for Regtest (maybe simnet in btcd as well?), how about something like:

```
       if b.chainParams.Name == chaincfg.RegressionNetParams.Name {
```

instead of introducing a new field in `Params`?.

kcalvinalvin (commented):
Looks ok otherwise

blockchain/difficulty.go
guggero: IMO this makes it just way more explicit what's going on. See https://github.com/lightninglabs/neutrino/pull/256#discussion_r1162444890 :sweat_smile:

chappjc (approved):
LGTM.  Just a note about the [initial objections from @davecgh](https://github.com/bitcoin/bitcoin/pull/6853#issuecomment-149219109), but I think the ship has sailed on this since it's what bitcoind does and apps like neutrino need to match.

Roasbeef:
> but I think the ship has sailed on this since it's what bitcoind does and apps like neutrino need to match.

FWIW simnet _does_ do retargeting, so applications that depend on being able to replicate such behavior can use that in place.

Roasbeef (approved):
LGTM 💫

Needs rebase.

PR #1991, opened 2023-06-05
Convert `witnessToHex` into a method `ToHexStrings` on `TxWitness`

ffranr:
Add a method `ToHexStrings` to `TxWitness`. Remove stand alone function `witnessToHex(witness wire.TxWitness) []string`.

Motivation: `witnessToHex` is hard to access and to find.

### DIFF
--- a/wire/msgtx.go
+++ b/wire/msgtx.go
@@ -5,6 +5,7 @@
 package wire
 
 import (
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
@@ -300,6 +301,22 @@ func (t TxWitness) SerializeSize() int {
 	}
 
 	return n
+}
+
+// ToHexStrings formats the witness stack as a slice of hex-encoded strings.
+func (t TxWitness) ToHexStrings() []string {
+	// Ensure nil is returned when there are no entries versus an empty
+	// slice so it can properly be omitted as necessary.
+	if len(t) == 0 {
+		return nil
+	}
+
+	result := make([]string, len(t))
+	for idx, wit := range t {
+		result[idx] = hex.EncodeToString(wit)
+	}
+
+	return result
 }
 
 // TxOut defines a bitcoin transaction output.
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -646,23 +646,6 @@ func handleDebugLevel(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) 
 	return "Done.", nil
 }
 
-// witnessToHex formats the passed witness stack as a slice of hex-encoded
-// strings to be used in a JSON response.
-func witnessToHex(witness wire.TxWitness) []string {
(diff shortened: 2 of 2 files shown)


halseth (commented):
ACK

wire/msgtx.go
```diff
@@ -309,6 +309,23 @@ func (t TxWitness) SerializeSize() int {
...
+func (t TxWitness) ToHexStrings() []string {
+	// Ensure nil is returned when there are no entries versus an empty
+	// slice so it can properly be omitted as necessary.
+	if len(t) == 0 {
+		return nil
+	}
+
+	result := make([]string, 0, len(t))
+	for _, wit := range t {
+		result = append(result, hex.EncodeToString(wit))
```
guggero, line 325: Whenever we have an index based loop (e.g. not iterating over a map), I think we should use the index based access:
```
	result := make([]string, len(t))
	for idx, wit := range t {
		result[idx] = hex.EncodeToString(wit)
	}
```

IMO the `append()` based version always needs a second look when reading the code, because the initial allocation can cause unwanted behavior if the middle parameter is missing (`make([]T, 0, n)` vs. `make([]T, n)`).
The index based access just makes it more clear in my eyes.

wire/msgtx.go
```diff
@@ -326,6 +328,33 @@ func (t TxWitness) ToHexStrings() []string {
 	return result
 }
 
+// MarshalJSON serialises the structure as a JSON appropriate value.
+//
+// NOTE: The 2D byte array is encoded as an array of hex strings.
+func (t TxWitness) MarshalJSON() ([]byte, error) {
```
guggero, line 334: Do we really want those `MarshalJSON`/`UnmarshalJSON` methods? Those represent a breaking change for users that might have relied on the "old" serialization format. Didn't we have the same problem with the chain hash where adding the (un)marshal methods broke some user's code?
I don't really like Go's approach to JSON serialization because depending on a method being present or not makes the serialization behavior quite non-deterministic when updating dependencies.
That's why I think we shouldn't rely on it (which is probably also the reason those methods don't exist on most of our structs).

guggero (commented):
Thanks for the PR. I think we should just have the helper method (`ToHexStrings()` and perhaps a `ParseWitness()` function) but skip the JSON (un)marshal support (see inline comment for the reason).

wire/msgtx.go
ffranr: Will remove from PR.

ffranr:
I've reduced this PR to just converting `witnessToHex` into a method `ToHexStrings` on TxWitness`.

guggero (approved):
Very nice, LGTM :tada: 

Something that's also on my wish list is a simple `Serialize() ([]byte, error)` method for `TxWitness`, if you feel like adding that as well (new PR or this one, your call).

PR #2003, opened 2023-07-10
musig2: fix early nonce gen option

sputn1ck:
Previously the early nonce generation option was not being respected when creating the context, with the `WithKnownSigners` option being used.

### DIFF
--- a/btcec/schnorr/musig2/context.go
+++ b/btcec/schnorr/musig2/context.go
@@ -234,22 +234,21 @@ func NewContext(signingKey *btcec.PrivateKey, shouldSort bool,
 		opts.keySet = make([]*btcec.PublicKey, 0, opts.numSigners)
 		opts.keySet = append(opts.keySet, pubKey)
 
-		// If early nonce generation is specified, then we'll generate
-		// the nonce now to pass in to the session once all the callers
-		// are known.
-		if opts.earlyNonce {
-			var err error
-			ctx.sessionNonce, err = GenNonces(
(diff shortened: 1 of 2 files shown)


btcec/schnorr/musig2/context.go
```diff
@@ -234,24 +234,24 @@ func NewContext(signingKey *btcec.PrivateKey, shouldSort bool,
...
-			if err != nil {
-				return nil, err
-			}
-		}
-
 	default:
 		return nil, ErrSignersNotSpecified
 	}
 
+	// If early nonce generation is specified, then we'll generate
```
Roasbeef, line 241: Ah I guess my previous use case always had the _number_ of signers specified (case of LN channel funding).

Roasbeef (commented):
Motivation makes sense, can we tack on a unit test to ensure the new code path works as expected and as a guard against regressions?

btcec/schnorr/musig2/context.go
```diff
@@ -234,24 +234,24 @@ func NewContext(signingKey *btcec.PrivateKey, shouldSort bool,
...
+	// the nonce now to pass in to the session once all the callers
+	// are known.
```
guggero, line 243: nit: we could re-flow the comment to only take up 2 lines now.

guggero (approved):
Nice fix, LGTM :tada:

Roasbeef (approved):
LGTM 🎎

Issue #2004, opened 2023-07-13
Problem about making modular transaction with psbt


0xEclair:
I want to create  2 of 2 multisig tx with psbt. But got error: `non-mandatory-script-verify-flag (Invalid Schnorr signature)`

Here are steps, psbt1 and psbt2 are different instances:

1. create input0 and output0 in psbt1
2. call `txscript.TaprootWitnessSignature` with `txscript.SigHashSingle|txscript.SigHashAnyOneCanPay` and use private_key1
3. `updater.Upsbt.Inputs[0].TaprootKeySpendSig = witness from 2`
4. create input0 and output0 in psbt2
5. append psbt1.input0 as psbt2.input1, append psbt1.output0 as psbt2.output1, also append into `UnsignedTx`
6. call `txscript.TaprootWitnessSignature` with `txscript.SigHashDefault` and use another private_key
7. `updater.Upsbt.Inputs[0].TaprootKeySpendSig = witness from 6`
8. psbt2.Finalize(0) and psbt2.Finalize(1)
9. Extract

Then got error.

Here is codes of 5:
```
psbt2.Inputs = append(psbt2.Inputs, psbt1.Inputs[0])
psbt2.UnsignedTx.TxIn = append(psbt2.UnsignedTx.TxIn, psbt1.UnsignedTx.TxIn[0])
psbt2.UnsignedTx.TxIn[1].SignatureScript = nil
psbt2.Outputs = append(psbt2.Outputs, psbt1.Outputs[0])
psbt2.UnsignedTx.TxOut = append(psbt2.UnsignedTx.TxOut, psbt1.UnsignedTx.TxOut[0])
```

Roasbeef:
Look at the sighashes you're signing (you can insert print statements to see why they differ before the sig msg is hashed). If they're different, then that's why you get an invalid final signature. Also look at the ordering of the keys in the script as well: tapscript doesn't have a normal multi-sig op code, so you either need to use checksigadd or a normal checksig.

0xEclair:
> Look at the sighashes you're signing (you can insert print statements to see why they differ before the sig msg is hashed). If they're different, then that's why you get an invalid final signature. Also look at the ordering of the keys in the script as well: tapscript doesn't have a normal multi-sig op code, so you either need to use checksigadd or a normal checksig.

Print sighashes, they differ. But they're different UnsignedTx and OutFetcher, they should be same?

sighashes1:
```
[...]

PR #2006, opened 2023-07-14
fix: missing sighashtype when finalizing taproot keyspend signature

### LINKED ISSUE #2005: btcutil/psbt: finalizer ignores sighash flag when finalizing taproot keyspend signature
The finalizer should check the sighash flag in https://github.com/btcsuite/btcd/blob/2dbc98bdf3dc3f72a749c246cd0171bdd7abafd2/btcutil/psbt/finalizer.go#L522.
If it is not equal to `txscript.SigHashDefault` (and the signature is exactly 64 bytes, so it doesn't yet include the sighash flag), it should be appended to the signature.

Related: https://github.com/btcsuite/btcd/issues/2004

0xEclair:
Add sighash flag at end of taproot keyspend signature(witness).

Related: #2004 

Fixes #2005


btcutil/psbt/finalizer.go
```diff
@@ -519,6 +519,9 @@ func finalizeTaprootInput(p *Packet, inIndex int) error {
 	switch {
 	// Key spend path.
 	case len(pInput.TaprootKeySpendSig) > 0:
+		if pInput.SighashType != txscript.SigHashDefault && len(pInput.TaprootKeySpendSig) == 64 {
```
Roasbeef, line 522: Can you wrap this to 80 characters? 

You can also use the `schnorr.SignatureSize` constant instead of `64 here.

Roasbeef: Finally, I think we also want a test here to ensure that the finalizer always adds the proper sighash flags if needed.

Roasbeef (changes requested):

0xEclair:
Thanks for the review @Roasbeef. Addressed your comments.

btcutil/psbt/finalizer.go
```diff
@@ -519,6 +520,15 @@ func finalizeTaprootInput(p *Packet, inIndex int) error {
 	switch {
 	// Key spend path.
 	case len(pInput.TaprootKeySpendSig) > 0:
+		// Make sure TaprootKeySpendSig is equal to size of signature, if not,
+		// we assume that sighash flag was appended to the signature.
+		if len(pInput.TaprootKeySpendSig) == schnorr.SignatureSize {
+			// Append to the signature if flag is not equal to Default.
+			if pInput.SighashType != txscript.SigHashDefault {
+				sigHashType := byte(pInput.SighashType)
+				pInput.TaprootKeySpendSig = append(pInput.TaprootKeySpendSig, sigHashType)
```
guggero, line 529: I think we shouldn't modify the packet itself.
How about this?
```
		sig := pInput.TaprootKeySpendSig

		// Make sure TaprootKeySpendSig is equal to size of signature,
		// if not, we assume that sighash flag was appended to the
		// signature.
		if len(pInput.TaprootKeySpendSig) == schnorr.SignatureSize {
			// Append to the signature if flag is not equal to the
			// default sighash (that can be omitted).
			if pInput.SighashType != txscript.SigHashDefault {
				sigHashType := byte(pInput.SighashType)
				sig = append(sig, sigHashType)
			}
		}
		serializedWitness, err = writeWitness(sig)
```

btcutil/psbt/psbt_test.go
```diff
@@ -749,6 +749,66 @@ func TestPsbtExtractor(t *testing.T) {
...
+		"AllAnyOneCanPay":    "0200000000010130ac25ec34af987b9e0518ff05cd491bd2d339660a4bfeea49a580c9233fbd9d0000000000ffffffff010b0c0000000000002251209c1f4b7970d790c99b7265b53adec03551708fd7d67db78359f9c472fe642ad10141e86b7ea8d6fc2cbd99b1091c25a2a37b333b5d82ea559579553cf7ba08c0fe3bead26c458f4917a6e069a3712c15f0999adb243603c783133676c1a09cc574b28100000000",
+		"NoneAnyOneCanPay":   "020000000001015499da1d93851a8add52fcab05acab60eaaf16571e0015f678b68775937d11200000000000ffffffff01430b0000000000002251209c1f4b7970d790c99b7265b53adec03551708fd7d67db78359f9c472fe642ad101412d42b46429b739786020e52b69b969468aa69ca40af390ba13441c8e6dc9e53f679c2bd2ff0ef912f48922cd64f4a7bfe7e492e5ecc8603b63e0ea772385faab8200000000",
+		"SingleAnyOneCanPay": "020000000001011bbe693ee5b3d75a5c8ad190e151c81e5b1ff1090982ea712c375e7d4a6069ce0100000000ffffffff012b0f0000000000002251209c1f4b7970d790c99b7265b53adec03551708fd7d67db78359f9c472fe642ad101418e018d0ae9cd730f7eae428a456e920b4ded67c9a7500a82ba25dd23f98418c1f060680daa4352b262fdffab691a4a67fc603352c1d21ace7cc6d83490facde78300000000",
+	}
+
+	for key, signedPsbtStr := range signedPsbtData {
+		signedPsbtBytes, err := hex.DecodeString(signedPsbtStr)
+		if err != nil {
```
guggero, line 775: We started using the `require` library for new code which would make this quite a bit more compact. See https://github.com/btcsuite/btcd/blob/a2854c8fc3b135b6e658cfb827563b19a8ed1f25/btcutil/psbt/psbt_test.go#L1574 for an example.

btcutil/psbt/psbt_test.go
```diff
@@ -749,6 +749,66 @@ func TestPsbtExtractor(t *testing.T) {
...
+			t.Fatalf("Failed to decode hex: %v", err)
+		}
+
+		signedPsbt, err := NewFromRawBytes(bytes.NewReader(signedPsbtBytes), false)
+		if err != nil {
+			t.Fatalf("Failed to parse PSBT: %v", err)
+		}
+
+		// only one input for each psbt
```
guggero, line 784: nit: Comments should always be full sentences with capitalization and punctuation.

guggero (commented):
Thanks for the fix!
Left a few comments. Also, can you please squash the commits into one or two.

btcutil/psbt/finalizer.go
0xEclair: Of course, it's very important for readability.

0xEclair:
Thanks for the review @guggero. Addressed your comments.

guggero (approved):
LGTM :tada:

0xEclair:
PTAL @Roasbeef :)

Roasbeef (approved):
LGTM 🌸

Issue #1974, opened 2023-05-04
rpcclient: listaccounts is deprecated and replaced with listlabels


brandonpille:
A couple of commands have been deprecated and replaced by others. For example listaccounts is now listlabels. Most accounts related rpcs are now renamed to label. See https://chainquery.com/bitcoin-cli

roy9495:
Hi, I want to contribute. But I don't know what code changes I have to make. Can you please elaborate on this issue?

roy9495:
So these are the commands that got deprecated https://chainquery.com/bitcoin-cli#deprecated. Do you want all occurrences of these commands replaced by the alternatives, in this repo?
Thanks.

brandonpille:
Yes, it doesn't make sense to use deprecated calls by default as they will disappear some day. Or at least do the same as in the bitcoin nodes. In the new bitcoin node you can pass an argument telling the node to use the old (deprecated) calls instead. I think it makes sense to do that. We should, by default, use new calls and if wanted use the deprecated calls.

brandonpille:
Hi, do we have an estimate for this?

brandonpille:
If someone wants to contribute, here are the required changes (got that from bitcoin repository: https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-0.17.0.md#label-and-account-apis-for-wallet):

| Deprecated Method       | New Method            | Notes       |
| :---------------------- | :-------------------- | :-----------|
| `getaccount`            | `getaddressinfo`      | `getaddressinfo` returns a json object with address information instead of just the name of the account as a string. |
| `getaccountaddress`     | n/a                   | There is no replacement for `getaccountaddress` since labels do not have an associated receive address. |
| `getaddressesbyaccount` | `getaddressesbylabel` | `getaddressesbylabel` returns a json object with the addresses as keys, instead of a list of strings. |
| `getreceivedbyaccount`  | `getreceivedbylabel`  | _no change in behavior_ |
| `listaccounts`          | `listlabels`          | `listlabels` does not return a balance or accept `minconf` and `watchonly` arguments. |
| `listreceivedbyaccount` | `listreceivedbylabel` | Both methods return new `label` fields, along with `account` fields for backward compatibility. |
| `move`                  | n/a                   | _no replacement_ |
| `sendfrom`              | n/a                   | _no replacement_ |
[...]

Issue #1934, opened 2023-01-06
btcjson: use type params to revamp package to be more extensible and eliminate duplication 


Roasbeef:
In the beginning, we would keep up with the Bitcoin Core RPC interface mostly in lock step. This meant for each change, we'd also add the message (or modify) it and try to keep up with that behavior. As the years have passed, the RPC interfaces have drifted more and more over time, and at this point we are no longer explicity putting in the effort to add every new RPC call, sometimes due to architectural differences. 

However, since we don't implement all the calls they have, this means that the `rpcclient` library start to lack functionality that developers need to build on top of `btcd`. If a new call `Foo` is added, then we need to update both the `btcjson` and the `rpcclient` packages for this to be available to developers. 

With Go 1.18, generics or type params were used. We can use this feature to:
  * Eliminate all the duplicated code in the package 
  * Create generic abstractions that allow developers to simply _define_ the new message type they want to use seamlessly 

It's worth noting that devs can already use the [`RawRequest`](https://pkg.go.dev/github.com/btcsuite/btcd/rpcclient#Client.RawRequest) method to issue any arbitrary JSON RPC requests. For this workflow, they'd define the new json struct, then pass it in as params, and then decode the opaque json blob on the other side. 

---

Here's a sketch of what that would look like. Today we end up with a ton of duplicated code, since we implement a custom future type for each message, then also a series of methods to issue an async vs a sync code. Instead, we can rely on interfaces that look something like:
```go
type Response struct {
	Resp json.RawMessage
	Err  error
}

type Command interface {
	Name() string
	Marshall() json.RawMessage
}

type Client interface {
	SendCmd(cmd Command) chan *Response
}

type MessageFuture[M any] interface {
	Receive() (M, error)
}

type RpcCall[Args any, Req Command, Resp any] interface {
	MapReq(arg Args) Req

	MapResp(resp *Response) MessageFuture[Resp]
}

func AsyncRequest[Arg any, Req Command, Resp any, CmdReq RpcCall[Arg, Req, Resp]](
	client Client, cmdReq RpcCall[Arg, Req, Resp], args Arg) MessageFuture[Resp] {

	jsonCmd := cmdReq.MapReq(args)

	resp := client.SendCmd(jsonCmd)

	return cmdReq.MapResp(<-resp)
}
```

Along the way we can do away with most instances of reflection in the package as well. 

Then a new call can be added purely by the developer that needs it by implementing the `RpcCall` for the new command.
```go

[...]

PR #2014, opened 2023-07-29
btcjson, rpcclient: revamp package and eliminate duplication

0xEclair:
Implemented `New Method`, used proper Deprecated comment for `Deprecated Method`, renamed account named parameter to label for `Change Method`.

For more APIs: https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-0.17.0.md#label-and-account-apis-for-wallet

Related: #1974

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -62,4 +62,4 @@ retract (
 	v0.13.0-beta
 )
 
-go 1.17
+go 1.18
--- a/rpcclientv2/blochchaincmd.go
+++ b/rpcclientv2/blochchaincmd.go
@@ -1,0 +1,47 @@
+package rpcclientv2
+
+import (
+	"encoding/json"
+)
+
+type GetDifficultyRpcCmd struct{}
+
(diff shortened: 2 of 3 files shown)


Roasbeef:
See https://github.com/btcsuite/btcd/issues/1934 

The changes here would introduce breaking changes for any existing clients. The current `btcjson` package allows one to define custom messages for encoding/decoding, which means an independent codebase can always keep up with the latest set of breaking changes that `bitcoind` implements for their JSON-RPC interface. The goal of the linked issue above is to give callers more freedom to define w/e arbitrary JSON-RPC methods. `btcd` no longer makes it an explicit goal to keep up with `bitcoind`'s RPC interface, but the package as is allows a user to track all changes if they wish.

0xEclair:
Yes, i agree.  Current style of btcjson and rpcclient has a lot of duplicated code. I'll refactor it.

0xEclair:
Added an example based on the sketch.

PR #2020, opened 2023-08-15
txscript:Add supported script type WitnessV1PubKeyHash

loganever:
# support script type P2WPKH V1. 

I noticed some pk script couldn't be parsed

Example:
Tx:ba5fd9fe19d4b4b859779a98602e279c13072d17367aa6fb0bc81c2bb2e1e006

This tx's first output pk script is 5114c5e507cae1475ff181ce9ab56d74aca5e4508cf9, which means the first opcode is OP_1, and the second opcode is OP_DATA_20. That is a version 1 P2WPKH, and not be supported, so I add support script type P2WPKH V1 code. I hope this is helpful.

[BIP350](https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki)

### DIFF
--- a/btcutil/address.go
+++ b/btcutil/address.go
@@ -172,6 +172,9 @@ func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
 
 			switch len(witnessProg) {
 			case 20:
+				if witnessVer == 1 {
+					return newAddressWitnessPubKeyHashV1(hrp, witnessProg)
+				}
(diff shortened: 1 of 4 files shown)


loganever:
> A v1 P2WPKH type does not exist. Someone burned money, that's probably why it's only a couple of sats. That will never be spendable again (without a soft fork). The only valid v1 output is a push of 32 bytes.

ok,I got it

PR #2023, opened 2023-08-21
btcutil: reuse serialized tx during TxHash

kcalvinalvin:
btcutil.Block caches the serialized raw bytes of the block during ibd. This serialized block bytes includes the serialized tx. The current tx hash generation will re-serialized the de-serialized tx to create the raw bytes and it'll only then hash that.

This commit changes the code so that the re-serialization never happens, saving tons of cpu and memory overhead.

### DIFF
--- a/btcutil/block.go
+++ b/btcutil/block.go
@@ -154,12 +154,26 @@ func (b *Block) Transactions() []*Tx {
 		b.transactions = make([]*Tx, len(b.msgBlock.Transactions))
 	}
 
+	// Offset of each tx.  80 accounts for the block header size.
+	offset := 80 + wire.VarIntSerializeSize(uint64(len(b.msgBlock.Transactions)))
+
 	// Generate and cache the wrapped transactions for all that haven't
 	// already been done.
 	for i, tx := range b.transactions {
 		if tx == nil {
 			newTx := NewTx(b.msgBlock.Transactions[i])
 			newTx.SetIndex(i)
+
+			size := b.msgBlock.Transactions[i].SerializeSize()
+
+			// The block may not always have the serializedBlock.
+			if len(b.serializedBlock) > 0 {
+				// This allows for the reuse of the already serialized tx.
+				newTx.setBytes(b.serializedBlock[offset : offset+size])
+
+				// Increment offset for this block.
+				offset += size
+			}
 			b.transactions[i] = newTx
 		}
 	}
@@ -234,6 +248,9 @@ func NewBlockFromBytes(serializedBlock []byte) (*Block, error) {
 		return nil, err
 	}
 	b.serializedBlock = serializedBlock
+	// This initializes []btcutil.Tx to have the serialized raw transactions cached.
+	// Helps speed up things like generating the txhash.
+	b.Transactions()
 	return b, nil
 }
 
@@ -257,9 +274,13 @@ func NewBlockFromReader(r io.Reader) (*Block, error) {
 // NewBlockFromBlockAndBytes returns a new instance of a bitcoin block given
 // an underlying wire.MsgBlock and the serialized bytes for it.  See Block.
 func NewBlockFromBlockAndBytes(msgBlock *wire.MsgBlock, serializedBlock []byte) *Block {
-	return &Block{
+	b := &Block{
 		msgBlock:        msgBlock,
 		serializedBlock: serializedBlock,
 		blockHeight:     BlockHeightUnknown,
 	}
+	// This initializes []btcutil.Tx to have the serialized raw transactions cached.
+	// Helps speed up things like generating the txhash.
+	b.Transactions()
+	return b
 }
--- a/btcutil/tx.go
+++ b/btcutil/tx.go
@@ -27,6 +27,7 @@ type Tx struct {
 	txHashWitness *chainhash.Hash // Cached transaction witness hash
 	txHasWitness  *bool           // If the transaction has witness data
 	txIndex       int             // Position within a block or TxIndexUnknown
+	rawBytes      []byte          // Raw bytes for the tx in the raw block.
 }
 
 // MsgTx returns the underlying wire.MsgTx for the transaction.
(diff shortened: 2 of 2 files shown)


btcutil/tx.go
```diff
@@ -37,30 +38,69 @@ func (t *Tx) MsgTx() *wire.MsgTx {
...
-	hash := t.msgTx.TxHash()
+	var hash chainhash.Hash
+
+	// If we have the raw bytes, then don't call msgTx.TxHash as that has the
+	// overhead of serialization.
+	if len(t.rawBytes) > 0 {
+		// If the raw bytes contain the witness, we must strip it out before
+		// calculating the hash.
+		if t.HasWitness() {
+			baseSize := t.msgTx.SerializeSizeStripped()
```
halseth, line 58: Hm, is this same code used when parsing the block in the first place? Feels like one maybe should cache the bytes at that point, to avoid duplicating this parsing logic.

halseth (changes requested):
What is the speedup?

kcalvinalvin:
> What is the speedup?

The speedup is resulted from the fact that the tx is never serialized. Since we receive it serialized in the block, we just hash those bytes, saving a lot of overhead during txhash calculation.

It's a different optimization from #1978 and this speedup is specific to ibd.

btcutil/block.go
```diff
@@ -154,12 +154,26 @@ func (b *Block) Transactions() []*Tx {
...
+	offset := 80 + wire.VarIntSerializeSize(uint64(len(b.msgBlock.Transactions)))
+
 	// Generate and cache the wrapped transactions for all that haven't
 	// already been done.
 	for i, tx := range b.transactions {
 		if tx == nil {
 			newTx := NewTx(b.msgBlock.Transactions[i])
 			newTx.SetIndex(i)
+
+			size := b.msgBlock.Transactions[i].SerializeSize()
```
Roasbeef, line 167: IIUC, with this change, we only get the benefit when `block.Transactions()` is called, while https://github.com/btcsuite/btcd/pull/1376 will ensure the serialization is cached as soon as things are read off the wire. The one thing this has that the other PR doesn't is storing the serialization for both segwit and non-segwit transactions. Doing this actually simplifies the logic a bit, as then you can just tee the entire thing and skip the segwit specific parts for the non-segwit encoding.

btcutil/tx.go
Roasbeef: https://github.com/btcsuite/btcd/pull/1376 does the caching in-line and at the very lowest level. The diff is also a bit more minimal as well. I think the test case there is just due to a lack of clone before working on a transaction. I've synced with it and haven't encountered that issue in particular.

Roasbeef (changes requested):

btcutil/tx.go
Roasbeef: Also the validation logic uses a direct `wire.MsgTx` rather than a `btcutil.Tx`, which is why the memoization above doesn't help us in this case.

btcutil/block.go
Roasbeef: Also this isn't addressing the same issue as #1376 is, see this profile: https://github.com/btcsuite/btcd/pull/1376#issuecomment-1467126518 

When we go to calculate the sighash we want to hash the entire serialization, which isn't the same thing as a normal dsha256 to compute the `txid`: https://github.com/btcsuite/btcd/blob/7520523a9a54aff213f5455b048a55f818118a84/txscript/sighash.go#L173

btcutil/tx.go
```diff
@@ -37,30 +38,69 @@ func (t *Tx) MsgTx() *wire.MsgTx {
...
+			bytes := make([]byte, 0, baseSize)
```
guggero, line 59: nit: `result` or `nonWitnessBytes` to not collide with the imported package name?

guggero (commented):
Did a first round to load some context. Can't really add much to the discussion whether to use this PR or https://github.com/btcsuite/btcd/pull/1376, other than I think this code is probably harder to reason about.

Crypt-iQ (commented):
Code looks good, I did a once over to see if `btcutil.Tx` had any cases where it was modified after using `Hash` but couldn't find any. I think somebody else should also double-check just to be safe

Roasbeef:
> Code looks good, I did a once over to see if btcutil.Tx had any cases where it was modified after using Hash but couldn't find any.

Yeah I think that's one advtange: for block validation, things come off the wire, then go into the `btcutil` wrapper, after which they're more or less immutable.

Roasbeef:
Running a combined branch for a sync test (utxo cache, this PR, and #1978), so far on my machine it's on schedule for a 6-ish hour sync (connected to `bitcoind` node over LAN, same machine). 

One thing that still shows up is this hotspot later into the chain with full validation: 
Screenshot 2023-12-15 at 3 53 53 PM

Currently of the opinion that my attempt in https://github.com/btcsuite/btcd/pull/1376 may not be the best way to go about it, due to unintended consequences (PSBT interactions signing the long sighash, etc). Spending more time with routines like `calcSignatureHash`, it's also clear that it could actually lead to incorrect behavior when computing the sighash. When computing the non-segwit sighash, we blank out the sigScripts of inputs not currently being spent. But the cached version of `SerializeNoWitness` would emit the serialized version we read on the wire, which would include the sigScripts that were blanked our earlier.

Roasbeef:
Zooming in a bit more, we can see that the channel based binary free list logic is actually what is making things so slow here when writing inputs/outputs:
Screenshot 2023-12-15 at 4 07 51 PM

Roasbeef:
Closing in favor of #1376, see this comment for some details: https://github.com/btcsuite/btcd/pull/2023#issuecomment-1858625556

TL;DR: I think we need to remove/optimize the binary free list instead.

Roasbeef:
Oops, closed the wrong version...

btcutil/tx.go
```diff
@@ -37,30 +38,69 @@ func (t *Tx) MsgTx() *wire.MsgTx {
...
 	// Return the cached hash if it has already been generated.
 	if t.txHashWitness != nil {
 		return t.txHashWitness
 	}
 
 	// Cache the hash and return it.
-	hash := t.msgTx.WitnessHash()
+	var hash chainhash.Hash
+	if len(t.rawBytes) > 0 {
+		hash = chainhash.DoubleHashH(t.rawBytes)
```
Roasbeef, line 99: This can now use `DoubleHashRaw`.

btcutil/tx.go
```diff
@@ -37,30 +38,69 @@ func (t *Tx) MsgTx() *wire.MsgTx {
...
+
+			// Append the last 4 bytes which are the locktime bytes.
+			bytes = append(bytes, t.rawBytes[len(t.rawBytes)-4:]...)
+
+			hash = chainhash.DoubleHashH(bytes)
+		} else {
+			hash = chainhash.DoubleHashH(t.rawBytes)
+		}
+	} else {
+		hash = t.msgTx.TxHash()
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest push

btcutil/tx.go
```diff
@@ -90,6 +130,16 @@ func (t *Tx) SetIndex(index int) {
 	t.txIndex = index
 }
 
+// SetBytes sets the raw bytes of the tx.
+func (t *Tx) SetBytes(bytes []byte) {
+	t.rawBytes = bytes
```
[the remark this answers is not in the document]
kcalvinalvin: Privated the function in the latest push

btcutil/tx.go
```diff
@@ -90,6 +130,16 @@ func (t *Tx) SetIndex(index int) {
...
+}
+
+// RawBytes returns the cached raw bytes of the tx.
+func (t *Tx) RawBytes() []byte {
```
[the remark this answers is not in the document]
kcalvinalvin: Removed in the latest push

btcutil/tx.go
kcalvinalvin: Addressed in the latest push

kcalvinalvin:
Rebased and addressed the review comments.

Roasbeef:
Did the final nits in https://github.com/btcsuite/btcd/pull/2081

PR #2025, opened 2023-08-23
chainhash: add support to legacy-marshaled hashes

Beerosagos:
Recent commits 1d6e578 and 72ea23e introduced a change in the way Hashes are serialized and deserialized. This change could cause errors in downstream applications that persisted hashes serialized using the previous methods.

This introduces support for legacy-serialized hashes unmarshaling and restores the compatibility with previous versions.

### DIFF
--- a/chaincfg/chainhash/hash.go
+++ b/chaincfg/chainhash/hash.go
@@ -118,6 +118,12 @@ func (hash Hash) MarshalJSON() ([]byte, error) {
 
 // UnmarshalJSON parses the hash with JSON appropriate string value.
 func (hash *Hash) UnmarshalJSON(input []byte) error {
+	// If the first byte indicates an array, the hash could have been marshalled
+	// using the legacy method and e.g. persisted.
+	if len(input) > 0 && input[0] == '[' {
+		return decodeLegacy(hash, input)
+	}
+
 	var sh string
 	err := json.Unmarshal(input, &sh)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


benma:
cc @Roasbeef - this would be helpful as the BitBoxApp stores json-marshalled transactions, which contain the `Hash` type in the prevout info. The change in serialization format broke the BitBoxApp as it can't load persisted transactions anymore. This PR would fix it in an ergonomic way without needing a lot of custom boilerplate code.

chaincfg/chainhash/hash.go
```diff
@@ -118,6 +118,12 @@ func (hash Hash) MarshalJSON() ([]byte, error) {
 
 // UnmarshalJSON parses the hash with JSON appropriate string value.
 func (hash *Hash) UnmarshalJSON(input []byte) error {
+	// If the fist byte indicates an array, the hash could have been marshalled
```
benma, line 121: typo `first`

chaincfg/chainhash/hash.go
```diff
@@ -217,3 +223,17 @@ func Decode(dst *Hash, src string) error {
 
 	return nil
 }
+
+// decodeLegacy decodes an Hash that has been encoded with the legacy method
+// (i.e. represented as a bytes array) to a destination.
+func decodeLegacy(dst *Hash, src []byte) error {
```
guggero, line 229: Can we add a test case to the `TestHashJsonMarshal` function please?

guggero (commented):
Makes sense, thanks for the fix. Should add a unit test case though.

Beerosagos:
@guggero @benma Thanks for the review! Pushed an update to fix the typo and add a test case PTAL :pray:

guggero (approved):
LGTM :tada:

Roasbeef (approved):
LGTM 👑

PR #2027, opened 2023-08-30
rpcclient: handle RPC requests based on their priority

yyforyongyu:
This PR introduces a new flag `lowPriority` on RPC requests to differentiate their priorities so more urgent RPC requests can be processed earlier. This is needed as `bitcoind` will block when too many RPCs requests are made at the same time, yet our recent addition of mempool notifier puts a lot of pressure on it because it's calling `getrawtransaction` for every new transaction, causing other RPC requests to halt. To solve it, we introduce a new method `GetRawTransactionLazy`, which will make the RPC calls more friendly and leave room for other more urgent RPC calls to go through.

### DIFF
--- a/rpcclient/infrastructure.go
+++ b/rpcclient/infrastructure.go
@@ -29,6 +29,7 @@ import (
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/go-socks/socks"
 	"github.com/btcsuite/websocket"
+	"golang.org/x/sync/errgroup"
 )
 
 var (
@@ -90,6 +91,12 @@ const (
 	// requestRetryInterval is the initial amount of time to wait in between
 	// retries when sending HTTP POST requests.
 	requestRetryInterval = time.Millisecond * 500
+
+	// rpcRequestInterval is the duration to wait before making more RPC
+	// requests to bitcoind. This is only used by low priority messages.
+	// For every 4 messages, we would sleep this duration before making
+	// more post requests.
+	rpcRequestInterval = time.Millisecond * 10
 )
 
 // jsonRequest holds information about a json request that is used to properly
@@ -928,6 +935,39 @@ func (c *Client) nextPostRequest() chan *jsonRequest {
 // while allowing the sender to continue running asynchronously.  It must be run
 // as a goroutine.
 func (c *Client) sendPostHandler() {
+	// Add an errgroup to handle the post messages in goroutines.
+	var eg errgroup.Group
+
+	// Limit the number of concurrent RPC calls.
+	//
+	// TODO(yy): Ideally we should pass the `-rpcthreads` used in
+	// `bitcoind` here to better increase the performance.
+	const numThreads = 4
+	eg.SetLimit(numThreads)
+
+	// lowReqCounter counts the number of low priority requests.
+	var lowReqCounter uint64
+
+	// handleLowPriorityRequest handles a low priority RPC request. It will
+	// sleep 10ms for every 4 messages to avoid slow response from
+	// bitcoind.
+	handleLowPriorityRequest := func(req *jsonRequest) {
+		lowReqCounter++
+		if lowReqCounter%numThreads == 0 {
+			log.Tracef("Sleeping %v for low priority command [%s] "+
+				"with id %d.", rpcRequestInterval,
+				req.method, req.id)
+
+			select {
+			case <-time.After(rpcRequestInterval):
+			case <-c.shutdown:
+				return
+			}
+		}
+
+		c.handleSendPostMessage(req, c.shutdown)
+	}
+
 out:
 	for {
 		// Send any messages ready for send until the shutdown channel
@@ -939,8 +979,27 @@ out:
 				break out
 			}
 
-			c.handleSendPostMessage(jReq, c.shutdown)
+			// For low priority messages we will process them
+			// sequentially.
+			if jReq.lowPriority {
+				handleLowPriorityRequest(jReq)
+			} else {
+				// For high priority messages we'll process
+				// them in goroutines to priorize them.
+				eg.Go(func() error {
+					c.handleSendPostMessage(
+						jReq, c.shutdown,
+					)
+
+					return nil
+				})
+			}
 		}
+	}
+
+	// Wait for all goroutines to report back.
+	if err := eg.Wait(); err != nil {
+		log.Errorf("Error sending post message: %v", err)
 	}
 
 	// Drain any wait channels before exiting so nothing is left waiting
--- a/rpcclient/infrastructure.go
+++ b/rpcclient/infrastructure.go
@@ -1116,6 +1116,22 @@ func (c *Client) sendRequest(jReq *jsonRequest) {
 // future.  It handles both websocket and HTTP POST mode depending on the
 // configuration of the client.
 func (c *Client) SendCmd(cmd interface{}) chan *Response {
+	return c.sendCmdWithPriority(cmd, false)
+}
+
+// SendCmdSlow flags the the passed command as low priority and sends it to the
+// associated server and returns a response channel on which the reply will be
+// delivered at some point in the future. It handles both websocket and HTTP
+// POST mode depending on the configuration of the client.
+func (c *Client) SendCmdSlow(cmd interface{}) chan *Response {
+	return c.sendCmdWithPriority(cmd, true)
+}
+
+// sendCmdWithPriority is the internal implementation of SendCmd/SendCmdLazy
+// and takes a bool to indicate whether this cmd has a low priority.
+func (c *Client) sendCmdWithPriority(cmd interface{},
+	lowPriority bool) chan *Response {
+
 	rpcVersion := btcjson.RpcVersion1
 	if c.batch {
 		rpcVersion = btcjson.RpcVersion2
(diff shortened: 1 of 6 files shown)


rpcclient/infrastructure.go
```diff
@@ -892,7 +929,7 @@ out:
 		// Send any messages ready for send until the shutdown channel
 		// is closed.
 		select {
-		case jReq := <-c.highPriorityPostQueue:
+		case jReq := <-c.nextPostRequest():
```
guggero, line 970: This now no longer considers the `c.shutdown` channel. Because we'll block on the method call `c.nextPostRequest()` and then only actually `select` once that returns. So I think we need to consider `c.shutdown` within `nextPostRequest()` and then we can simply return the next request (instead of a channel) or a boolean that indicates we're shutting down.

rpcclient/infrastructure.go
```diff
@@ -905,7 +942,7 @@ out:
 cleanup:
 	for {
 		select {
-		case jReq := <-c.highPriorityPostQueue:
+		case jReq := <-c.nextPostRequest():
```
guggero, line 995: Same here, this breaks assumptions and will block forever on shutdown.

rpcclient/infrastructure.go
```diff
@@ -924,13 +931,60 @@ func (c *Client) nextPostRequest() chan *jsonRequest {
...
 func (c *Client) sendPostHandler() {
+	// Add an errgroup to handle the post messages in goroutines.
+	var eg errgroup.Group
+
+	// Limit the number of concurrent RPC calls.
+	//
+	// TODO(yy): Ideally we should pass the `-rpcthreads` used in
+	// `bitcoind` here to better increase the performance.
+	const numThreads = 4
+	eg.SetLimit(numThreads)
```
guggero, line 940: Don't we ever need to wait for this error group? Or we're basically using it as a simple "thread pool"?

rpcclient/infrastructure.go
```diff
@@ -924,13 +931,60 @@ func (c *Client) nextPostRequest() chan *jsonRequest {
...
+
+	// lowReqCounter counts the number of low priority requests.
+	lowReqCounter := 0
```
guggero, line 943: Should we use an explicit `uint64` type here to avoid rolling over into negative? Wouldn't really matter too much as we're only using the modulus, but would give us some more explicit behavior.

rpcclient/infrastructure.go
```diff
@@ -90,6 +91,12 @@ const (
 	// requestRetryInterval is the initial amount of time to wait in between
 	// retries when sending HTTP POST requests.
 	requestRetryInterval = time.Millisecond * 500
+
+	// rpcRequestInterval is the duration to wait before making more RPC
+	// requests to bitcoind. This is only used by low priority messages.
+	// For every 4 messages, we would sleep this duration before making
+	// more post requests.
+	rpcRequestInterval = time.Millisecond * 10
```
guggero, line 99: Could we maybe make this configurable when instantiating the client?

rpcclient/infrastructure.go
```diff
@@ -1097,6 +1097,22 @@ func (c *Client) sendRequest(jReq *jsonRequest) {
...
 // configuration of the client.
 func (c *Client) SendCmd(cmd interface{}) chan *Response {
+	return c.sendCmdWithPriority(cmd, false)
+}
+
+// SendCmdLazy flags the the passed command as low priority and sends it to the
+// associated server and returns a response channel on which the reply will be
+// delivered at some point in the future. It handles both websocket and HTTP
+// POST mode depending on the configuration of the client.
+func (c *Client) SendCmdLazy(cmd interface{}) chan *Response {
```
guggero, line 1102: Low prio: To me, the word "lazy" in programming indicates "don't do it immediately, only when we actually need the result". Which kind of fits, but also not fully. Wasn't able to come up with a better suffix though, so I asked ChatGPT. Not sure if any of those fit better:
```
SendCmdLight - Using "Light" suggests a less intensive or lower priority operation.
SendCmdLazy - "Lazy" often implies that the action isn't urgently prioritized.
SendCmdIdle - "Idle" might indicate that the task is of low urgency.
SendCmdSlow - Directly indicates that it's a less priority-sensitive function.
SendCmdBack - As in "backburner", where it's not the immediate focus.
SendCmdMinor - Signifies lesser importance without being too long.
SendCmdRelax - Suggests that the function can "relax" and doesn't need to hurry.
SendCmdSoft - A softer version, possibly less critical.
SendCmdCasual - Implying the task is non-urgent.
SendCmdChill - A colloquial way to denote low urgency.
```

guggero (changes requested):
Great idea and optimiziations! Though I think some assumptions around reading from channels have changed and might introduce new blocking behavior.

rpcclient/infrastructure.go
yyforyongyu: good catch, fixed!

rpcclient/infrastructure.go
yyforyongyu: here is a bit different as we wanna drain the channel, so changed to read the channels directly.

rpcclient/infrastructure.go
yyforyongyu: yeah it's more like limiting concurrent calls, but think we should also wait in the end so now I added it.

rpcclient/infrastructure.go
yyforyongyu: updated

rpcclient/infrastructure.go
yyforyongyu: Done!

rpcclient/infrastructure.go
yyforyongyu: Maybe `SendCmdWithLowPriority`? kinda long...

guggero: Yeah, it's a bit long. My favorite from the list is actually `SendCmdSlow`.

yyforyongyu: switched to `slow`

Roasbeef:
What about taking a more generic approach here? The request type is the paramter, and we can have a generic method that'll handle the priotization. 

I also think we should step back a bit before proceeding to make sure we're actually fixing something here:
  * For `bitcoind`, the client runs in `POST` mode. 
  * In `POST` mode, a single goroutine handles all the incoming requests (it'll block an retry up to 10 times before giving up): https://github.com/btcsuite/btcd/blob/ec401d00a149d817e227b6afbc35b2399aca489d/rpcclient/infrastructure.go#L885-L921 
  * Similarly, `bitcoind` also has a set of RPC threads that consume the requests, and block, eventually returning an error if they can't be processed. 

In short, adding another channel to consume requests off of, doesn't help to actually address the issue of [head of line blocking](https://en.wikipedia.org/wiki/Head-of-line_blocking). If we're sending 10 `getblock` requests, and one of them blocks because of `bitcoind` hitting a queue limit, then the entire sequence will block. I think what we're after here is either utilizing JSON-RPC batching properly, or doing a rechitcture to make the `POST` mode as concurrent as the normal `ws` mode. 

The `ws` mode has proper _pipelining_: a message request adds an entry to match the response, then sends out the request, not blocking on the response: https://github.com/btcsuite/btcd/blob/ec401d00a149d817e227b6afbc35b2399aca489d/rpcclient/infrastructure.go#L986-L996. This can further be improved by having multiple worker goroutines read off the queue to handle the responses.

yyforyongyu:
The initial idea was to implement [a special case](https://github.com/btcsuite/btcwallet/pull/886) that can be used by our mempool poller as it consumes a lot of RPC resources. The original observation was, a lot of `getrawtransaction` requests were queued, causing the call to `getblockchaininfo` to be blocked by calling `getinfo` in `lnd`.

However, after digging into `bitcoind`'s debug log, the cause of the slowness seemed to be on the `bitcoind` side - whenever a new block tip is connected, or a new peer is connected, the daemon would require a lock, blocking the RPC responses until the aforementioned tasks are finished, as shown in the logs,
```
2023-08-31T09:17:35Z ThreadRPCServer method=getrawtransaction user=yy
2023-08-31T09:17:35Z ThreadRPCServer method=getrawtransaction user=yy
2023-08-31T09:17:39Z New outbound peer connected: version: 70016, blocks=805566, peer=24 (outbound-full-relay)
2023-08-31T09:17:42Z New outbound peer connected: version: 70016, blocks=805566, peer=22 (outbound-full-relay)
2023-08-31T09:17:42Z ThreadRPCServer method=getrawtransaction user=yy
2023-08-31T09:17:42Z ThreadRPCServer method=getrawtransaction user=yy
2023-08-31T09:17:48Z ThreadRPCServer method=getrawtransaction user=yy
```

```
2023-08-31T09:05:35Z ThreadRPCServer method=getrawtransaction user=yy
2023-08-31T09:05:39Z UpdateTip: new best=00000000000000000003134988bcace49938ffe175379a14c66529580c202473 height=805559 version=0x20400000 log2_work=94.390304 tx=886852495 date='2023-08-31T08:29:35Z' progress=0.999993 cache=53.6MiB(319643txo)
2023-08-31T09:05:48Z ThreadRPCServer method=getrawtransaction user=yy
```

Thus this PR is no longer the right fix. If we are looking for async RPC calls, we should instead use `rpcclient.BatchNew` as suggested by @Roasbeef.

PR #2035, opened 2023-09-25
wire, netsync: change isSyncCandidate behavior

kcalvinalvin:
isSyncCandidate is now changed to return true even if the peer is a
pruned node if and only if our chaintip is within 288 blocks of the
peer.

Rationale being that pruned nodes that signal NODE_NETWORK_LIMITED
MUST serve 288 blocks from their chaintip.  If our chaintip is within that
range, this peer can be a sync candidate even if they aren't an archival node.

### DIFF
--- a/netsync/manager.go
+++ b/netsync/manager.go
@@ -397,20 +397,55 @@ func (sm *SyncManager) isSyncCandidate(peer *peerpkg.Peer) bool {
 		if host != "127.0.0.1" && host != "localhost" {
 			return false
 		}
-	} else {
-		// The peer is not a candidate for sync if it's not a full
-		// node. Additionally, if the segwit soft-fork package has
-		// activated, then the peer must also be upgraded.
-		segwitActive, err := sm.chain.IsDeploymentActive(chaincfg.DeploymentSegwit)
-		if err != nil {
-			log.Errorf("Unable to query for segwit "+
-				"soft-fork state: %v", err)
-		}
-		nodeServices := peer.Services()
-		if nodeServices&wire.SFNodeNetwork != wire.SFNodeNetwork ||
-			(segwitActive && !peer.IsWitnessEnabled()) {
+
+		// Candidate if all checks passed.
+		return true
+	}
+
+	// If the segwit soft-fork package has activated, then the peer must
+	// also be upgraded.
+	segwitActive, err := sm.chain.IsDeploymentActive(
+		chaincfg.DeploymentSegwit,
+	)
+	if err != nil {
+		log.Errorf("Unable to query for segwit soft-fork state: %v",
+			err)
+	}
+
+	if segwitActive && !peer.IsWitnessEnabled() {
+		return false
+	}
+
+	var (
+		nodeServices = peer.Services()
+		fullNode     = nodeServices.HasFlag(wire.SFNodeNetwork)
+		prunedNode   = nodeServices.HasFlag(wire.SFNodeNetworkLimited)
+	)
+
+	switch {
+	case fullNode:
+		// Node is a sync candidate if it has all the blocks.
+
+	case prunedNode:
+		// Even if the peer is pruned, if they have the node network
+		// limited flag, they are able to serve 2 days worth of blocks
+		// from the current tip. Therefore, check if our chaintip is
+		// within that range.
+		bestHeight := sm.chain.BestSnapshot().Height
+		peerLastBlock := peer.LastBlock()
+
+		// bestHeight+1 as we need the peer to serve us the next block,
+		// not the one we already have.
+		if bestHeight+1 <=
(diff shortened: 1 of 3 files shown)


Roasbeef:
cc @ellemouton

netsync/manager.go
```diff
@@ -406,9 +406,30 @@ func (sm *SyncManager) isSyncCandidate(peer *peerpkg.Peer) bool {
...
 				"soft-fork state: %v", err)
 		}
+		if segwitActive && !peer.IsWitnessEnabled() {
+			return false
+		}
+
 		nodeServices := peer.Services()
-		if nodeServices&wire.SFNodeNetwork != wire.SFNodeNetwork ||
-			(segwitActive && !peer.IsWitnessEnabled()) {
+
```
ellemouton, line 414: just some style comments:

- could be nice for readability to add a `HasFlags` method to `ServiceFlags`:
   ```
   func (f ServiceFlag) HasFlag(s ServiceFlag) bool {
	return f&s == s
   }
   ```
- im not sure how strongly the btcd repo follows the same style guide as LND, but we mostly wrap at 80 characters.
- while we are here, I think we can remove one indentation from the func by returning early for the regtest case.

With the above 3 comments integrated, then the result will be:

```
func (sm *SyncManager) isSyncCandidate(peer *peerpkg.Peer) bool {
	// Typically a peer is not a candidate for sync if it's not a full node,
	// however regression test is special in that the regression tool is
	// not a full node and still needs to be considered a sync candidate.
	if sm.chainParams == &chaincfg.RegressionNetParams {
		// The peer is not a candidate if it's not coming from localhost
		// or the hostname can't be determined for some reason.
		host, _, err := net.SplitHostPort(peer.Addr())
		if err != nil {
			return false
		}

		if host != "127.0.0.1" && host != "localhost" {
			return false
		}

		// Candidate if all checks passed.
		return true
	}

	// The peer is not a candidate for sync if it's not a full node.
	// Additionally, if the segwit soft-fork package has activated, then the
	// peer must also be upgraded.
	segwitActive, err := sm.chain.IsDeploymentActive(
		chaincfg.DeploymentSegwit,
	)
	if err != nil {
		log.Errorf("Unable to query for segwit soft-fork state: %v",
			err)
	}

	if segwitActive && !peer.IsWitnessEnabled() {
		return false
	}

	var (
		nodeServices = peer.Services()
		fullNode     = nodeServices.HasFlag(wire.SFNodeNetwork)
		prunedNode   = nodeServices.HasFlag(wire.SFNodeNetworkLimited)
	)

	switch {
	case fullNode:
		// Node is a sync candidate if it has all the blocks.

	case prunedNode:
		// Even if the peer is pruned, if they have the node network
		// limited flag, they are able to serve 2 days worth of blocks
		// from the current tip. Therefore, check if our chaintip is
		// within that range.
		bestHeight := sm.chain.BestSnapshot().Height
		peerLastBlock := peer.LastBlock()

		// bestHeight+1 as we need the peer to serve us the next block,
		// not the one we already have.
		if bestHeight+1 <
			peerLastBlock-wire.NodeNetworkLimitedBlockThreshold {

			return false
		}

	default:
		// If the peer isn't an archival node, and it's not signaling
		// NODE_NETWORK_LIMITED, we can't sync off of this node.
		return false
	}

	// Candidate if all checks passed.
	return true
}
```

wire/protocol.go
```diff
@@ -151,6 +151,12 @@ func (f ServiceFlag) String() string {
 	return s
 }
 
+const (
```
halseth, line 154: nit: declare up with the service flags? This location seems a bit arbitrary

netsync/manager.go
```diff
@@ -406,9 +406,30 @@ func (sm *SyncManager) isSyncCandidate(peer *peerpkg.Peer) bool {
 			log.Errorf("Unable to query for segwit "+
 				"soft-fork state: %v", err)
 		}
+		if segwitActive && !peer.IsWitnessEnabled() {
```
halseth, line 409: review nit: this is a nice refactor, would be even better if the change (breaking up the if conditions) was done in its own commit (easier to review) :)

halseth (approved):
LGTM 💯

netsync/manager.go
Roasbeef: @kcalvinalvin do you plan to adopt the fragment above?

netsync/manager.go
kcalvinalvin: Would this still be needed with all the above changes?

wire/protocol.go
kcalvinalvin: Addressed

netsync/manager.go
kcalvinalvin: Adopted the fragment above

Roasbeef (approved):
LGTM 🌽

PR #2043, opened 2023-10-13
integration: add test to check prune status

kcalvinalvin:
Adds a check to make sure that prune status on getblockchaininfo returns true for pruned nodes.

### DIFF
--- a/integration/prune_test.go
+++ b/integration/prune_test.go
@@ -1,0 +1,41 @@
+// Copyright (c) 2023 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+// This file is ignored during the regular tests due to the following build tag.
+//go:build rpctest
+// +build rpctest
+
+package integration
+
+import (
+	"testing"
+
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/integration/rpctest"
+)
+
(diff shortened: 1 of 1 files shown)


integration/prune_test.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func TestPrune(t *testing.T) {
+	t.Parallel()
+
+	// Boilerplate code to make a pruned node.
+	btcdCfg := []string{"--prune=1536"}
+	r, err := rpctest.New(&chaincfg.SimNetParams, nil, btcdCfg, "")
+	if err != nil {
+		t.Fatal("unable to create primary harness: ", err)
```
guggero, line 25: Should we use the `require` library for new code? Makes things more compact IMO.

integration/prune_test.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+	if err := r.SetUp(false, 0); err != nil {
+		t.Fatalf("unable to setup chain: %v", err)
+	}
+	defer r.TearDown()
```
guggero, line 30: Could use `t.Cleanup()`.

guggero (approved):
Nice test, LGTM :tada:

integration/prune_test.go
kcalvinalvin: Addressed in a fixup

halseth (approved):
LGTM ✅

PR #2051, opened 2023-10-30
blockchain: export CheckSerializedHeight

Crypt-iQ:
Some callers only want to check the coinbase height rather than be given the height. Also reduce allocations for cb height parsing


Roasbeef (approved):
LGTM 🚵🏼‍♂️

PR #2055, opened 2023-11-06
wire: add method TxID to MsgTx

ffranr:
This commit adds a method to MsgTx called TxID. This method returns the transaction ID (txid) of the subject transaction. This commit also adds a function for deriving the transaction ID from the double hash of the transaction.

### DIFF
--- a/wire/msgtx.go
+++ b/wire/msgtx.go
@@ -353,6 +353,11 @@ func (msg *MsgTx) TxHash() chainhash.Hash {
 	return chainhash.DoubleHashRaw(msg.SerializeNoWitness)
 }
 
+// TxID generates the transaction ID of the transaction.
+func (msg *MsgTx) TxID() string {
+	return msg.TxHash().String()
+}
+
 // WitnessHash generates the hash of the transaction serialized according to
 // the new witness serialization defined in BIP0141 and BIP0144. The final
 // output is used within the Segregated Witness commitment of all the witnesses
--- a/wire/msgtx_test.go
+++ b/wire/msgtx_test.go
@@ -756,6 +756,34 @@ func TestTxSerializeSizeStripped(t *testing.T) {
 	}
 }
 
+// TestTxID performs tests to ensure the serialize size for various transactions
+// is accurate.
+func TestTxID(t *testing.T) {
+	// Empty tx message.
+	noTx := NewMsgTx(1)
+	noTx.Version = 1
+
+	tests := []struct {
+		in   *MsgTx // Tx to encode.
+		txid string // Expected transaction ID.
+	}{
+		// No inputs or outputs.
+		{noTx, "d21633ba23f70118185227be58a63527675641ad37967e2aa461559f577aec43"},
+
+		// Transaction with an input and an output.
+		{multiTx, "0100d15a522ff38de05c164ca0a56379a1b77dd1e4805a6534dc9b3d88290e9d"},
+
+		// Transaction with an input which includes witness data, and
+		// one output.
+		{multiWitnessTx, "0f167d1385a84d1518cfee208b653fc9163b605ccf1b75347e2850b3e2eb19f3"},
+	}
+
+	for i, test := range tests {
+		txid := test.in.TxID()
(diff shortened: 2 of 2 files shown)


ffranr:
I am aware of `chainhash.Hash.String()`, but I still think this PR is a good idea.

This PR makes it clear what `txid` is in relation to `MsgTx`. It isn't clear that `chainhash.Hash.String()` is `txid`. And it isn't necessarily always `txid` because `chainhash.Hash` doesn't have to be a double sha256.

wire/msgtx.go
```diff
@@ -366,6 +367,29 @@ func (msg *MsgTx) TxHash() chainhash.Hash {
...
+		revBytes[i], revBytes[size-1-i] = revBytes[size-1-i], revBytes[i]
+	}
+
+	// Hex encode the reversed bytes.
+	return hex.EncodeToString(revBytes[:])
+}
+
+// TxID generates the transaction ID of the transaction.
+func (msg *MsgTx) TxID() string {
+	// Encode the transaction and calculate the double sha256 hash of the
```
kcalvinalvin, line 386: Why not just:

```
// TxID is a wrapper around TxHash and returns a string of the transaction hash.
func (msg *MsgTx) TxID() string {
        return msg.TxHash().String()
}
```

wire/msgtx.go
```diff
@@ -366,6 +367,29 @@ func (msg *MsgTx) TxHash() chainhash.Hash {
 	return chainhash.DoubleHashH(buf.Bytes())
 }
 
+// TxIdFromMsgTxHash generates the transaction ID (txid) of the transaction from
+// the double hash of the transaction.
+func TxIdFromMsgTxHash(txHash chainhash.Hash) string {
```
kcalvinalvin, line 372: Unnecessary code duplication.

kcalvinalvin (commented):
I think it's ok to have a method that returns a string directly.

But there's no need for all this code to do so.

wire/msgtx.go
ffranr: Because sometimes we may already have the `TxHash` available. If we have a function for computing the txid from the txhash then we can skip recomputing the hash.

However, I've gone with your suggestion of `msg.TxHash().String()` because all I really care about right now is adding `MsgTx.TxID`.

wire/msgtx.go
ffranr: I duplicated the code here because I don't think `chainhash.Hash.String` is correct. From my POV, it shouldn't reverse the byte order before creating a string from the hash. From what I understand, that reversal routine should be unique to txid.

In other words, I don't think `chainhash.Hash.String` should be the transaction ID, so I wanted to get away from that method. IMO, duplicating the code in this PR made it clear the relationship between a txhash and a txid. 

However, I only really care about adding `MsgTx.TxID` in this PR. So I've removed this duplicate code.

saubyk:
@yyforyongyu for your review

wire/msgtx_test.go
```diff
@@ -756,6 +756,38 @@ func TestTxSerializeSizeStripped(t *testing.T) {
...
+
+		// Transaction with an input and an output.
+		{multiTx, "0100d15a522ff38de05c164ca0a56379a1b77dd1e4805a6534dc9b3d88290e9d"},
+
+		// Transaction with an input which includes witness data, and
+		// one output.
+		{multiWitnessTx, "0f167d1385a84d1518cfee208b653fc9163b605ccf1b75347e2850b3e2eb19f3"},
+	}
+
+	t.Logf("Running %d tests", len(tests))
```
yyforyongyu, line 781: nit: seems unnecessary as go will already print the tests for you?

wire/msgtx_test.go
```diff
@@ -756,6 +756,38 @@ func TestTxSerializeSizeStripped(t *testing.T) {
...
+	for i, test := range tests {
+		txid := test.in.TxID()
+		if txid != test.txid {
```
yyforyongyu, line 784: think we usually do `require.Equal` which gives you a nice info about a test failure.

yyforyongyu (approved):

PR #2057, opened 2023-11-06
added tests for GetBestBlockHashAsync

ClaytonNorthey92:
now testing that GetBestBlockHashAsync sends the getbestblockhash command via websocket connection and that the channel returned can be used to send the response when it is received


jcvernaleo (approved):
OK

PR #2058, opened 2023-11-09
musig2: export SignOptions members

ellemouton:
Export the members of the SignOptions struct so that they can be inspected elsewhere.

### DIFF
--- a/btcec/schnorr/musig2/sign.go
+++ b/btcec/schnorr/musig2/sign.go
@@ -104,65 +104,65 @@ func (p *PartialSignature) Decode(r io.Reader) error {
 
 // SignOption is a functional option argument that allows callers to modify the
 // way we generate musig2 schnorr signatures.
-type SignOption func(*signOptions)
+type SignOption func(*SignOptions)
 
-// signOptions houses the set of functional options that can be used to modify
+// SignOptions houses the set of functional options that can be used to modify
 // the method used to generate the musig2 partial signature.
-type signOptions struct {
-	// fastSign determines if we'll skip the check at the end of the
+type SignOptions struct {
+	// FastSign determines if we'll skip the check at the end of the
 	// routine where we attempt to verify the produced signature.
-	fastSign bool
+	FastSign bool
 
-	// sortKeys determines if the set of keys should be sorted before doing
+	// SortKeys determines if the set of keys should be sorted before doing
 	// key aggregation.
-	sortKeys bool
+	SortKeys bool
 
-	// tweaks specifies a series of tweaks to be applied to the aggregated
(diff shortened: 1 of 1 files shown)


btcec/schnorr/musig2/sign.go
```diff
@@ -104,65 +104,65 @@ func (p *PartialSignature) Decode(r io.Reader) error {
...
 
-	// tweaks specifies a series of tweaks to be applied to the aggregated
+	// Tweaks specifies a series of tweaks to be applied to the aggregated
 	// public key, which also partially carries over into the signing
 	// process.
-	tweaks []KeyTweakDesc
+	Tweaks []KeyTweakDesc
 
-	// taprootTweak specifies a taproot specific tweak.  of the tweaks
+	// TaprootTweak specifies a taproot specific tweak of the tweaks
```
kcalvinalvin, line 125: Good eye!

kcalvinalvin:
ACK 35e0f57c10963673d3712a5a43dc903c484089a1

Looked at the git diffs, built & tested locally.

btcec/schnorr/musig2/sign.go
```diff
@@ -104,65 +104,65 @@ func (p *PartialSignature) Decode(r io.Reader) error {
...
 // way we generate musig2 schnorr signatures.
-type SignOption func(*signOptions)
+type SignOption func(*SignOptions)
 
-// signOptions houses the set of functional options that can be used to modify
+// SignOptions houses the set of functional options that can be used to modify
 // the method used to generate the musig2 partial signature.
-type signOptions struct {
-	// fastSign determines if we'll skip the check at the end of the
+type SignOptions struct {
```
Roasbeef, line 111: Why does any of this need to be exported here? Having these fields be private enables a type of encapsulation and information hiding here. 

API design here is that new options are added to the package, as they're a bit more "curated", and exported via the functions like `WithFastSign`.

Roasbeef (changes requested):

btcec/schnorr/musig2/sign.go
ellemouton: check out the link in the PR description. It is used in the `RPCKeyRing.SignMuSig2` method [here](https://github.com/lightningnetwork/lnd/pull/8163/commits/b21177a36812b4d77d08fc1e9eeb33a30b63aadb). 

We cant pass the options as is over gRPC, so if we want to be able to do MuSig2 signing over the signer client, then we need to be able to inspect the signing options

Roasbeef: Ah, in the past, what I ended up doing there was to just sorta replicate the option types in an unrolled manner, then you map then to `SignOption` at the latest stage. We had to do this for the `musig2` options as well.

PR #2061, opened 2023-11-16
refactor: rm redundant check as every value of type byte is >= 0

Halimao:
remove redundant check for byte type field `op.value`, as every value of type byte is >= 0


Roasbeef:
FWIW, unless changes are absolutely necessary, we generally prefer not to touch code areas that deal with low level script execution like this (though this does appear to be benign as all bytes are unsigned).

Halimao:
> FWIW, unless changes are absolutely necessary, we generally prefer not to touch code areas that deal with low level script execution like this (though this does appear to be benign as all bytes are unsigned). 

Got it👌

PR #2065, opened 2023-11-28
Fix typos on signature.go

EvanTedesco:
Fixes typos in the documentation.

### DIFF
--- a/btcec/schnorr/signature.go
+++ b/btcec/schnorr/signature.go
@@ -124,8 +124,7 @@ func schnorrVerify(sig *Signature, hash []byte, pubKeyBytes []byte) error {
 	// 7. Fail if is_infinite(R)
 	// 8. Fail if not hash_even_y(R)
 	// 9. Fail is x(R) != r.
-	// 10. Return success iff not failure occured before reachign this
-	// point.
+	// 10. Return success if failure did not occur before reaching this point.
 
 	// Step 1.
 	//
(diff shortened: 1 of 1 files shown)


jcvernaleo (approved):
Good catches.

OK

btcec/schnorr/signature.go
```diff
@@ -124,8 +124,7 @@ func schnorrVerify(sig *Signature, hash []byte, pubKeyBytes []byte) error {
 	// 7. Fail if is_infinite(R)
 	// 8. Fail if not hash_even_y(R)
 	// 9. Fail is x(R) != r.
-	// 10. Return success iff not failure occured before reachign this
-	// point.
+	// 10. Return success if failure did not occur before reaching this point.
```
guggero, line 127: The term `iff` is a scientific notation for "if and only if". So this is now wrong and not in line with the actual BIP: https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#verification

jcvernaleo: Damn, I missed that because the original line was not at all correct English.

jcvernaleo: An `iff` mixed in with the rest of the typos was indistinguishable from another type :)

Anyway, if you PR putting the `iff` back I'll merge.  If not, I'll PR tomorrow.

PR #2071, opened 2023-12-15
rpctest: integration test harness fixes

guggero:
This fixes two issues with the current RPC integration test framework:
 - Generated test temp directories were created using a non-mutex-protected harness ID and a deprecated way to create temp directories, both is now fixed
 - Because Go creates a sub process for each package that is tested, a package-global variable to track used TCP ports doesn't work reliably, especially if multiple packages are run in parallel. This PR adds a new way to issue unique TCP ports per process ID, which needs to be the parent process ID of each sub-process to work as expected.

### DIFF
--- a/blockchain/sizehelper.go
+++ b/blockchain/sizehelper.go
@@ -63,14 +63,32 @@ const (
 	loadFactorNum = 13
 	loadFactorDen = 2
 
-	// maxAlloc is the maximum size of an allocation. On 64-bit,
-	// it's theoretically possible to allocate 1<<heapAddrBits bytes. On
-	// 32-bit, however, this is one less than 1<<32 because the
-	// number of bytes in the address space doesn't actually fit
-	// in a uintptr.
+	// _64bit = 1 on 64-bit systems, 0 on 32-bit systems
+	_64bit = 1 << (^uintptr(0) >> 63) / 2
+
+	// PtrSize is the size of a pointer in bytes - unsafe.Sizeof(uintptr(0))
+	// but as an ideal constant. It is also the size of the machine's native
+	// word size (that is, 4 on 32-bit systems, 8 on 64-bit).
+	PtrSize = 4 << (^uintptr(0) >> 63)
+
+	// heapAddrBits is the number of bits in a heap address that's actually
+	// available for memory allocation.
 	//
-	// NOTE (kcalvinalvin): I just took the constant for a 64 bit system.
-	maxAlloc = 281474976710656
+	// NOTE (guggero): For 64-bit systems, we just assume 40 bits of address
+	// space available, as that seems to be the lowest common denominator.
+	// See heapAddrBits in runtime/malloc.go of the standard library for
+	// more details
+	heapAddrBits = 32 + (_64bit * 8)
+
+	// maxAlloc is the maximum size of an allocation on the heap.
+	//
+	// NOTE(guggero): With the somewhat simplified heapAddrBits calculation
+	// above, this will currently limit the maximum allocation size of the
+	// UTXO cache to around 300GiB on 64-bit systems. This should be more
+	// than enough for the foreseeable future, but if we ever need to
+	// increase it, we should probably use the same calculation as the
+	// standard library.
+	maxAlloc = (1 << heapAddrBits) - (1-_64bit)*1
 )
 
 var class_to_size = [_NumSizeClasses]uint16{0, 8, 16, 24, 32, 48, 64, 80, 96, 112, 128, 144, 160, 176, 192, 208, 224, 240, 256, 288, 320, 352, 384, 416, 448, 480, 512, 576, 640, 704, 768, 896, 1024, 1152, 1280, 1408, 1536, 1792, 2048, 2304, 2688, 3072, 3200, 3456, 4096, 4864, 5376, 6144, 6528, 6784, 6912, 8192, 9472, 9728, 10240, 10880, 12288, 13568, 14336, 16384, 18432, 19072, 20480, 21760, 24576, 27264, 28672, 32768}
@@ -175,7 +193,7 @@ func calculateMinEntries(totalBytes int, bucketSize int) int {
 // mulUintptr returns a * b and whether the multiplication overflowed.
 // On supported platforms this is an intrinsic lowered by the compiler.
 func mulUintptr(a, b uintptr) (uintptr, bool) {
-	if a|b < 1<<(4*8) || a == 0 {
+	if a|b < 1<<(4*PtrSize) || a == 0 {
 		return a * b, false
 	}
 	overflow := b > MaxUintptr/a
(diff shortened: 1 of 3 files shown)


integration/rpctest/rpc_harness.go
```diff
@@ -563,6 +563,99 @@ func NextAvailablePort() int {
...
+	// No ports available? Must be a mistake.
+	panic("no ports available for listening")
+}
+
+// GenerateProcessUniqueListenerAddresses is a function that returns two
+// listener addresses with unique ports per the given process id and should be
+// used to overwrite rpctest's default generator which is prone to use colliding
+// ports.
+func GenerateProcessUniqueListenerAddresses(pid int) (string, string) {
+	port1 := NextAvailablePortForProcess(pid)
```
Roasbeef, line 653: Why two vs 1 here? RPC listen and p2p listen?

integration/rpctest/rpc_harness.go
```diff
@@ -563,6 +563,99 @@ func NextAvailablePort() int {
 	panic("no ports available for listening")
 }
 
+// NextAvailablePortForProcess returns the first port that is available for
```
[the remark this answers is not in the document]
guggero: No, unfortunately not. If you spin up two tests at the same time (but not in the same process, so package-level variables are both initialized to the same value), they both request ports and the starting port (e.g. 8334) are unoccupied at the moment. So they get the same port allocated and when later actually spinning up the nodes, they get errors.

integration/rpctest/rpc_harness.go
guggero: Yeah, it's just the signature of the `ListenAddressGenerator` package-level variable that we overwrite with this if needed. The two ports are p2p and rpc listen ports, yes.

integration/rpctest/rpc_harness.go
```diff
@@ -248,10 +248,10 @@ func (h *Harness) SetUp(createTestChain bool, numMatureOutputs uint32) error {
 	// Start the btcd node itself. This spawns a new process which will be
 	// managed
 	if err := h.node.start(); err != nil {
-		return err
+		return fmt.Errorf("error starting node: %w", err)
```
yyforyongyu, line 251: 👍

integration/rpctest/rpc_harness.go
```diff
@@ -563,6 +563,99 @@ func NextAvailablePort() int {
 	panic("no ports available for listening")
 }
 
+// NextAvailablePortForProcess returns the first port that is available for
+// listening by a new node, using a lock file to make sure concurrent access for
+// parallel tasks within the same process don't re-use the same port. It panics
+// if no port is found and the maximum available TCP port is reached.
+func NextAvailablePortForProcess(pid int) int {
+	lockFile := fmt.Sprintf(
+		"%s/rpctest-port-pid-%d.lock", os.TempDir(), pid,
```
yyforyongyu, line 572: curious about whether this would behave the same on windows?

integration/rpctest/rpc_harness.go
```diff
@@ -563,6 +563,99 @@ func NextAvailablePort() int {
...
+	}
+
+	lastPort, err := strconv.Atoi(string(port))
+	if err != nil {
+		panic(fmt.Errorf("error parsing port: %w", err))
+	}
+
+	// We take the next one.
+	lastPort++
+	for lastPort < 65535 {
```
yyforyongyu, line 616: I think if we just listen to port 0 it will give us a random free port? Something like this
https://github.com/phayes/freeport/blob/master/freeport.go#L8

or this,
https://gist.github.com/sevkin/96bdae9274465b2d09191384f86ef39d

integration/rpctest/rpc_harness.go
guggero: Oh, good point. I think we need to explicitly close a file before we can delete it. And we should probably use `filepath.Join()` instead of `fmt.Sprintf()` to create the path. Will fix.

Roasbeef (approved):
LGTM ⚙️

Merge pending that windows fix.

integration/rpctest/rpc_harness.go
guggero: Updated things to be more Windows friendly.

guggero:
I updated things a bit with the node's `dataDir` and `logDir`, now it should be more clear where things are stored and who's in charge of cleaning up.

guggero:
It looks like #1955 broke the compilation on 32-bit systems. I added a commit to fix that, @kcalvinalvin could you take a look at the last commit in this PR please?

blockchain/sizehelper.go
```diff
@@ -63,14 +63,25 @@ const (
...
+	// heapAddrBits is the number of bits in a heap address that's actually
+	// available for memory allocation.
 	//
-	// NOTE (kcalvinalvin): I just took the constant for a 64 bit system.
-	maxAlloc = 281474976710656
+	// NOTE (guggero): For 64-bit systems, we just assume 40 bits of address
+	// space available, as that seems to be the lowest common denominator.
+	// See heapAddrBits in runtime/malloc.go of the standard library for
+	// more details
+	heapAddrBits = 32 + (_64bit * 8)
```
guggero, line 81: @kcalvinalvin: If we want to restore the original functionality, we can just set this to `32 + (_64bit * 32)`. But I'm not sure what exactly how `calculateRoughMapSize()` is used in the first place, whether this needs to represent the absolute maximum or just a reasonable maximum we don't want to go over in any case?

yyforyongyu (commented):
LGTM🏖

yyforyongyu (approved):

blockchain/sizehelper.go
kcalvinalvin: Hmm the problem is that `sizehelper.go` would be completely broken for a 32 bit system.

`calculateRoughMapSize()` is used to get how much the go map is using for the utxocache. The problem would be that it's specifically coded for 64 bit systems and wouldn't return the correct values for a 32 bit system. This would mean that it's inaccurate in calculating how much memory the map is using and the cache may be using more than what the user designated.

I don't think there's an easy fix here other than disabling the utxo cache for 32 bit systems...

guggero: Okay, I guess that's something to look into. But to fix the _compilation_ on 32-bit systems, what would you propose? I tried to make the code more similar to `runtime/malloc.go` with my changes. But I don't think I understand what the impact of `maxAlloc` is on the overall correctness of `calculateRoughMapSize()`.

kcalvinalvin: Ah `maxAlloc` is just to check if whatever the user requested was over that size. If it was over, then the runtime won't try to allocate that much space and will just try default to 0 `make([]uint64, 0)` <- this 0 here.

```
func Test1(t *testing.T) {
        mib := 300_000
        hint := calculateMinEntries(mib*1024*1204, bucketSize+avgEntrySize) - 1
        size := calculateRoughMapSize(hint, bucketSize)
        fmt.Println(size, hint)
}
```

Played around with that code above and seems like the new `maxAlloc` will cause issues when the user allocates ~400GiB as the utxocache when the old `maxAlloc` value will allow for that. The new one is ok at ~300GiB.

Seems ok since we're not expecting anyone to do that soon? Maybe just a comment on it explaining this would be good.

guggero: Okay, cool. Thanks for verifying. Added the comment in the latest push, what do you think?

kcalvinalvin: Looks great!

PR #2072, opened 2023-12-15
btcutil: update to chaincfg/chainhash/v1.1.0

Roasbeef:
This is a follow up to https://github.com/btcsuite/btcd/pull/1978. 

We bump to the new `chainhash` version, then start to use the functionality where applicable. 

Next in the sequence is:
  1. Merge this. 
  2. Update `btcutil/psbt`. 
  3. Merge that 
  4. Update the top-level pacakge (use the two new version above), then merge that. 

We'll also want to revisit this PR, as it aims to resolve some of the current module issues that cause PR chains like this each time we update a sub-module: https://github.com/btcsuite/btcd/pull/1825


guggero (approved):
LGTM :tada:

PR #2073, opened 2023-12-16
wire: only borrow/return binaryFreeList buffers at the message level

Roasbeef:
**NOTE**: This is a rebased version of https://github.com/btcsuite/btcd/pull/1426

The original PR body follows:
--- 
This PR optimizes the `wire` packages serialization of small buffers by minimizing the number of borrow/return round trips during message serialization. Currently the `wire` package uses a `binaryFreeList` from which 8-byte buffers are borrowed and returned for the purpose of serializing small integers and varints.

## Problem
To understand the problem, consider calling [`WriteVarInt`](https://github.com/btcsuite/btcd/blob/master/wire/common.go#L538) on a number greater than `0xfc` (which requires writing the discriminant and a 2, 4, or 8 byte value following).

For instance, writing 20,000 will invoke [`PutUint8`](https://github.com/btcsuite/btcd/blob/master/wire/common.go#L538) and then [`PutUint16`](https://github.com/btcsuite/btcd/blob/master/wire/common.go#L548). Expanding this out to examine the message passing, we see:
```
buffer <- freelist

PutUInt8(buffer, discriminant)
w.Write(buffer)

freelist <- buffer

buffer <- freelist

PutUint16(buffer, value)
w.Write(buffer)

freelist <- buffer
```

Each `<-` requires a channel `select`, which more-or-less bears the performance implication of a mutex. This cost, in addition to need to wake up other goroutines and switch executions, imparts a significant performance penalty. In the context of block serialization, several hundred thousand of these operations may be performed.

## Solution

In our example above, we can improve this by only using two `<-`, one to borrow and one to return, as so:

```
buffer <- freelist

PutUInt8(buffer, discriminant)
w.Write(buffer)

PutUint16(buffer, value)
w.Write(buffer)

freelist <- buffer
```

As expected, cutting the number channels sends in half cuts also cuts the latency in half, which can be seen in the benchmarks below for larger `VarInt`s, 

The remainder of this PR is to propagate this pattern all the way up to the top level of messages in the wire package, such that deserializing a message only incurs one borrow and one return. Any subroutines are made to conditionally borrow from the `binarySerializer` if the invoker has not provided them with a buffer, and conditionally return if they indeed were required to borrow.

A good example of how these channel sends/receives can add up is in MsgTx serialization, which is now upwards of 80% faster as a result of these optimizations:
```
BenchmarkSerializeTx-8            683           142           -79.21%
BenchmarkSerializeTxSmall-8       724           143           -80.25%
BenchmarkSerializeTxLarge-8       1476002       182111        -87.66%
```

## Preliminary Benchmarks
```
BenchmarkWriteVarInt1-8           71.9          74.6          +3.76%
BenchmarkWriteVarInt3-8           148           70.9          -52.09%
BenchmarkWriteVarInt5-8           149           71.0          -52.35%
BenchmarkWriteVarInt9-8           147           72.8          -50.48%
BenchmarkReadVarInt1-8            78.8          77.9          -1.14%
BenchmarkReadVarInt3-8            159           87.7          -44.84%
BenchmarkReadVarInt5-8            155           88.3          -43.03%
BenchmarkReadVarInt9-8            158           86.9          -45.00%
BenchmarkReadVarStr4-8            120           119           -0.83%
BenchmarkReadVarStr10-8           138           130           -5.80%
BenchmarkWriteVarStr4-8           101           105           +3.96%
BenchmarkWriteVarStr10-8          103           105           +1.94%
BenchmarkReadOutPoint-8           91.3          28.7          -68.57%
BenchmarkWriteOutPoint-8          78.9          10.2          -87.07%
BenchmarkReadTxOut-8              245           118           -51.84%
BenchmarkWriteTxOut-8             151           93.5          -38.08%
BenchmarkReadTxIn-8               338           139           -58.88%
BenchmarkWriteTxIn-8              238           31.0          -86.97%
BenchmarkDeserializeTxSmall-8     1119          586           -47.63%
BenchmarkDeserializeTxLarge-8     2476063       1275815       -48.47%
BenchmarkSerializeTx-8            683           142           -79.21%
BenchmarkSerializeTxSmall-8       724           143           -80.25%
BenchmarkSerializeTxLarge-8       1476002       182111        -87.66%
BenchmarkReadBlockHeader-8        406           69.1          -82.98%
BenchmarkWriteBlockHeader-8       431           21.1          -95.10%
BenchmarkDecodeGetHeaders-8       13537         12238         -9.60%
BenchmarkDecodeHeaders-8          1025275       236709        -76.91%
BenchmarkDecodeGetBlocks-8        13206         11684         -11.53%
BenchmarkDecodeAddr-8             337977        157519        -53.39%
BenchmarkDecodeInv-8              5990935       1898169       -68.32%
BenchmarkDecodeNotFound-8         6285831       1864701       -70.33%
BenchmarkDecodeMerkleBlock-8      4357          2606          -40.19%
BenchmarkTxHash-8                 1928          1222          -36.62%
BenchmarkDoubleHashB-8            993           969           -2.42%
BenchmarkDoubleHashH-8            932           1029          +10.41%

BenchmarkWriteBlockHeader-8       4              0              -100.00%
// all others remain 0 alloc

BenchmarkWriteBlockHeader-8       16            0             -100.00%
// all others remain unchanged
```

## Notes

I'm still in the process of going through and adding benchmarks to top-level messages in order to guage the overall performance benefit, expect more to be added at a later point.

There are a few remaining messages which have not yet been optimized, e.g. `MsgAlert`, `MsgVesion`, etc. I plan to add those as well but decided to start with the ones that were more performance critical.


Roasbeef:
cc @ellemouton

Roasbeef:
With this PR, combined with the rest on the v0.24 milestone, I was able to achieve a 12 hour sync over LAN!

wire/msgtx.go
```diff
@@ -791,7 +809,8 @@ func (msg *MsgTx) BtcEncode(w io.Writer, pver uint32, enc MessageEncoding) error
 		}
 	}
 
-	return binarySerializer.PutUint32(w, littleEndian, msg.LockTime)
+	err = binarySerializer.PutUint32(w, littleEndian, msg.LockTime)
```
guggero: nit: why this change?

wire/msgtx.go
```diff
@@ -809,7 +808,8 @@ func (msg *MsgTx) btcEncode(w io.Writer, pver uint32, enc MessageEncoding,
 		}
 	}
 
-	err = binarySerializer.PutUint32(w, littleEndian, msg.LockTime)
+	littleEndian.PutUint32(buf[:4], msg.LockTime)
+	_, err = w.Write(buf[:4])
```
guggero, line 765: Ah, I see, probably just a rebase artifact.

wire/msgtx.go
```diff
@@ -968,25 +968,66 @@ func NewMsgTx(version int32) *MsgTx {
...
+	buf := binarySerializer.Borrow()
+	err := readOutPointBuf(r, pver, version, op, buf)
+	binarySerializer.Return(buf)
+	return err
+}
+
+// readOutPointBuf reads the next sequence of bytes from r as an OutPoint.
+//
+// If b is non-nil, the provided buffer will be used for serializing small
+// values.  Otherwise a buffer will be drawn from the binarySerializer's pool
```
guggero, line 926: nit: this part of the comment should be on `readOutPoint` I think, since we don't draw a buffer from the binarySerializer here. Same below for `writeOutPointBuf`.

wire/msgtx.go
```diff
@@ -1037,8 +1037,30 @@ func writeOutPointBuf(w io.Writer, pver uint32, version int32, op *OutPoint,
...
+// memory exhaustion attacks and forced panics through malformed messages.  The
+// fieldName parameter is only used for the error message so it provides more
+// context in the error.
+//
+// If b is non-nil, the provided buffer will be used for serializing small
+// values.  Otherwise a buffer will be drawn from the binarySerializer's pool
+// and return when the method finishes.
+//
+// NOTE: b MUST either be nil or at least an 8-byte slice.
+func readScriptBuf(r io.Reader, pver uint32, buf []byte, maxAllowed uint32, fieldName string) ([]byte, error) {
```
guggero: nit: line length.

wire/msggetblocks.go
```diff
@@ -51,16 +51,21 @@ func (msg *MsgGetBlocks) AddBlockLocatorHash(hash *chainhash.Hash) error {
 // BtcDecode decodes r using the bitcoin protocol encoding into the receiver.
 // This is part of the Message interface implementation.
 func (msg *MsgGetBlocks) BtcDecode(r io.Reader, pver uint32, enc MessageEncoding) error {
-	err := readElement(r, &msg.ProtocolVersion)
-	if err != nil {
+	buf := binarySerializer.Borrow()
```
guggero, line 54: same comment here re `defer binarySerializer.Return()`.

wire/msgblock.go
```diff
@@ -137,23 +137,28 @@ func (msg *MsgBlock) DeserializeNoWitness(r io.Reader) error {
 func (msg *MsgBlock) DeserializeTxLoc(r *bytes.Buffer) ([]TxLoc, error) {
 	fullLen := r.Len()
 
+	buf := binarySerializer.Borrow()
```
guggero, line 142: Not sure if it's helpful to mention every instance where we can use `defer binarySerializer.Return()` to mark all the commits/files that need to be updated. I'm guessing not, so I'm not going to mention it any further.
It does sound like a tedious update in a lot of commits, but IMO it's worth it to avoid the buffer channel to fill up and effectively become useless because of a missing `Return()` somewhere.

guggero (commented):
Very nice changes and nicely split up into multiple commits.
I did as thorough a review as concentration allowed and didn't find anything wrong with the code.

The main thing I would change is to use a consistent pattern of:
```
buf := binarySerializer.Borrow()
defer binarySerializer.Return(buf)
```
everywhere to avoid a single mistake somewhere causing the buffers to all be occupied and becoming useless (which probably wouldn't show up in benchmark tests).

wire/msgtx.go
Roasbeef: Lost in the rebase sands...

wire/msgblock.go
Roasbeef: Good point, i think when this was originally written there might have been some concerns w.r.t the performance of `defer`. However since then, things have been optimized significantly.

PR #2074, opened 2023-12-16
fix: enhance code comments

threewebcode:
Enhance and tidy the code comments

### DIFF
--- a/mining/mining.go
+++ b/mining/mining.go
@@ -602,7 +602,7 @@ mempoolLoop:
 	blockSigOpCost := coinbaseSigOpCost
 	totalFees := int64(0)
 
-	// Query the version bits state to see whether segwit has been activated, if
+	// Query the version bits state to see if segwit has been activated, if
 	// so then this means that we'll include any transactions with witness
 	// data in the mempool, and also add the witness commitment as an
 	// OP_RETURN output in the coinbase transaction.
--- a/mining/mining.go
+++ b/mining/mining.go
@@ -520,7 +520,7 @@ mempoolLoop:
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


threewebcode:
@davecgh, Please help review

threewebcode:
@davecgh, Please help review

mining/mining.go
```diff
@@ -520,7 +520,7 @@ mempoolLoop:
 			continue
 		}
 
-		// Fetch all of the utxos referenced by the this transaction.
+		// Fetch all of the utxos referenced by this transaction.
```
jcvernaleo, line 523: Good catch here.

mining/mining.go
```diff
@@ -602,7 +602,7 @@ mempoolLoop:
 	blockSigOpCost := coinbaseSigOpCost
 	totalFees := int64(0)
 
-	// Query the version bits state to see if segwit has been activated, if
+	// Query the version bits state to see whether segwit has been activated, if
```
jcvernaleo, line 605: I don't think there is any need for this change.  The simpler original version is fine.

jcvernaleo (changes requested):
One change is good, I'd rather you revert the other one.

mining/mining.go
threewebcode: Thanks for your feedback.

mining/mining.go
threewebcode: Thanks for your opinion and i will discard this change.

threewebcode:
@jcvernaleo The change set is updated with your review. Please take a look again.

jcvernaleo (approved):
OK

PR #2075, opened 2023-12-19
multi: update main package and btcutil to chainhash/v1.1.0, use optimized dsha256 

Roasbeef:
In this commit, we update the top-level btcd package to use the latest
version of btcutil and also the chainhash package. With this version
bump, we can now use the new optimized dsha256 routine where applicable.

With this commit, I've covered most of the areas we'll hash an entire
transaction/block/header, but we may want to optimize some other areas
further, in particular, the witness sighash calc.


Roasbeef:
Here's a benchmark run vs master: 
```
BenchmarkCalcSigHash-8            1081499       1244944       +15.11%
BenchmarkCalcWitnessSigHash-8     34499         32224         -6.59%

BenchmarkCalcSigHash-8            712            623            -12.50%
BenchmarkCalcWitnessSigHash-8     445            801            +80.00%

BenchmarkCalcSigHash-8            1290508       933085        -27.70%
BenchmarkCalcWitnessSigHash-8     46992         19936         -57.58%
```

Roasbeef:
Here's another comparison, old is w/o the last commit, new is w/ it: 
```
benchmark                          old ns/op     new ns/op     delta
BenchmarkCalcSigHash-32            1866235       1975372       +5.85%
BenchmarkCalcWitnessSigHash-32     65135         45449         -30.22%

benchmark                          old allocs     new allocs     delta
BenchmarkCalcSigHash-32            623            623            +0.00%
BenchmarkCalcWitnessSigHash-32     534            801            +50.00%

benchmark                          old bytes     new bytes     delta
BenchmarkCalcSigHash-32            933080        933081        +0.00%
BenchmarkCalcWitnessSigHash-32     58384         19936         -65.85%
```

Roasbeef:
Here's a run with the same computer as the first run above, but this time with `GOAMD64=v3`: 
```
benchmark                          old ns/op     new ns/op     delta
BenchmarkCalcSigHash-32            1989442       1837011       -7.66%
BenchmarkCalcWitnessSigHash-32     59216         43734         -26.14%

benchmark                          old allocs     new allocs     delta
BenchmarkCalcSigHash-32            712            623            -12.50%
BenchmarkCalcWitnessSigHash-32     445            801            +80.00%

benchmark                          old bytes     new bytes     delta
BenchmarkCalcSigHash-32            1290506       933081        -27.70%
BenchmarkCalcWitnessSigHash-32     46992         19936         -57.58%
```

txscript/sighash.go
```diff
@@ -289,7 +292,14 @@ func calcWitnessSignatureHashRaw(subScript []byte, sigHashes *TxSigHashes,
 	binary.LittleEndian.PutUint32(bHashType[:], uint32(hashType))
 	sigHash.Write(bHashType[:])
 
-	return chainhash.DoubleHashB(sigHash.Bytes()), nil
+	sigHashBytes := chainhash.DoubleHashRaw(func(w io.Writer) error {
+		// TODO(rosabeef): put entire calc func into this? then no
```
guggero: I think we should do this, yes. Also, there's another instance of `DoubleHashB` just a few lines above.

txscript/sighash.go
```diff
@@ -200,103 +200,105 @@ func calcWitnessSignatureHashRaw(subScript []byte, sigHashes *TxSigHashes,
...
-	// hashPrevOuts, otherwise we just write zeroes for the prev outs.
-	if hashType&SigHashAnyOneCanPay == 0 {
-		sigHash.Write(sigHashes.HashPrevOutsV0[:])
-	} else {
-		sigHash.Write(zeroHash[:])
-	}
+	sigHashBytes := chainhash.DoubleHashRaw(func(w io.Writer) error {
+		// First write out, then encode the transaction's version
+		// number.
+		var bVersion [4]byte
```
guggero, line 206: We use a couple of these 4 byte arrays. Since `PutUint32` overwrites all bytes, I think it would be safe to re-use a single one and just call it `scratch`? It's likely all on the stack so probably doesn't make a huge difference though.

txscript/sighash.go
```diff
@@ -200,103 +200,105 @@ func calcWitnessSignatureHashRaw(subScript []byte, sigHashes *TxSigHashes,
...
+		binary.LittleEndian.PutUint32(
+			bIndex[:], txIn.PreviousOutPoint.Index,
+		)
+		w.Write(bIndex[:])
+
+		if isWitnessPubKeyHashScript(subScript) {
+			// The script code for a p2wkh is a length prefix
+			// varint for the next 25 bytes, followed by a
+			// re-creation of the original p2pkh pk script.
+			w.Write([]byte{0x19})
```
guggero, line 250: Not sure if this would make a big difference, but to avoid multiple allocations, we could do:
```
w.Write([]byte{0x19, OP_DUP, OP_HASH160, OP_DATA_20})
```
instead. But non-blocking of course.

txscript/sighash.go
```diff
@@ -200,103 +200,105 @@ func calcWitnessSignatureHashRaw(subScript []byte, sigHashes *TxSigHashes,
...
+		w.Write(bAmount[:])
+		var bSequence [4]byte
+		binary.LittleEndian.PutUint32(bSequence[:], txIn.Sequence)
+		w.Write(bSequence[:])
+
+		// If the current signature mode isn't single, or none, then we
+		// can re-use the pre-generated hashoutputs sighash fragment.
+		// Otherwise, we'll serialize and add only the target output
+		// index to the signature pre-image.
+		if hashType&sigHashMask != SigHashSingle &&
```
guggero, line 278: I think this would benefit from using a `switch` statement perhaps?

guggero (commented):
Very nice optimization! One go module issue, the rest are just non-blocking ideas for further optimizations since we touch an area that is used very often.

txscript/sighash.go
```diff
@@ -169,10 +169,13 @@ func calcSignatureHash(sigScript []byte, hashType SigHashType, tx *wire.MsgTx, i
 	// The final hash is the double sha256 of both the serialized modified
 	// transaction and the hash type (encoded as a 4-byte little-endian
 	// value) appended.
-	wbuf := bytes.NewBuffer(make([]byte, 0, txCopy.SerializeSizeStripped()+4))
-	txCopy.SerializeNoWitness(wbuf)
-	binary.Write(wbuf, binary.LittleEndian, hashType)
-	return chainhash.DoubleHashB(wbuf.Bytes())
+	sigHashBytes := chainhash.DoubleHashRaw(func(w io.Writer) error {
+		txCopy.SerializeNoWitness(w)
+		binary.Write(w, binary.LittleEndian, hashType)
```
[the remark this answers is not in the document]
Roasbeef: FWIW, `DoubleHashRaw` will swallow the error from serialization, but emit it if the hash writing fails.

txscript/sighash.go
Roasbeef: Can give it a shot to compare w/ a benchmark.

txscript/sighash.go
Roasbeef: Perhaps, but also prefer not to change this code much if not absolutely necessary.

btcutil/psbt/go.mod
```diff
@@ -3,19 +3,29 @@ module github.com/btcsuite/btcd/btcutil/psbt
...
 	github.com/btcsuite/btcd/btcec/v2 v2.1.3
-	github.com/btcsuite/btcd/btcutil v1.1.3
+	github.com/btcsuite/btcd/btcutil v1.1.4
 	github.com/btcsuite/btcd/chaincfg/chainhash v1.1.0
 	github.com/davecgh/go-spew v1.1.1
 	github.com/stretchr/testify v1.7.0
 )
 
 require (
+	github.com/aead/siphash v1.0.1 // indirect
```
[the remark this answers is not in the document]
guggero: Yeah, could be. I bumped everything to 1.19 in my module cleanup PR: https://github.com/btcsuite/btcd/pull/1825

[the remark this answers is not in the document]
Roasbeef: Actually, no you're right, `go mod tidy` eliminated a lot of stuff.

txscript/sighash.go
Roasbeef: Slight decrease here in time/op: 
```
⛰   benchstat old.txt new.txt
name                  old time/op    new time/op    delta
CalcWitnessSigHash-8    31.9µs ± 2%    31.5µs ± 1%  -1.31%  (p=0.000 n=10+10)

name                  old alloc/op   new alloc/op   delta
CalcWitnessSigHash-8    19.9kB ± 0%    19.9kB ± 0%    ~     (all equal)

name                  old allocs/op  new allocs/op  delta
CalcWitnessSigHash-8       801 ± 0%       801 ± 0%    ~     (all equal)
```

txscript/sighash.go
Roasbeef: I tried to use `binary.Write` first, but that made things slower, so then I switched to just using a single `[8]byte` and slicing into that. 

Ended up with a nice speed up on all marks: 
```
name                  old time/op    new time/op    delta
CalcWitnessSigHash-8    31.5µs ± 0%    29.2µs ± 0%   -7.05%  (p=0.000 n=10+10)

name                  old alloc/op   new alloc/op   delta
CalcWitnessSigHash-8    19.9kB ± 0%    18.5kB ± 0%   -7.14%  (p=0.000 n=10+10)

name                  old allocs/op  new allocs/op  delta
CalcWitnessSigHash-8       801 ± 0%       445 ± 0%  -44.44%  (p=0.000 n=10+10)
```

txscript/sighash.go
Roasbeef: Didn't make much off a diff seemingly: 
```
name                  old time/op    new time/op    delta
CalcWitnessSigHash-8    29.1µs ± 0%    29.6µs ± 2%  +1.70%  (p=0.000 n=9+9)

name                  old alloc/op   new alloc/op   delta
CalcWitnessSigHash-8    18.5kB ± 0%    18.5kB ± 0%    ~     (all equal)

name                  old allocs/op  new allocs/op  delta
CalcWitnessSigHash-8       445 ± 0%       445 ± 0%    ~     (all equal)
```

I guess the compiler can optimize those instances.

Roasbeef:
The final comparison: 
```
goos: linux
goarch: amd64
pkg: github.com/btcsuite/btcd/txscript
cpu: 13th Gen Intel(R) Core(TM) i9-13900K
                      │   old.txt    │               new.txt               │
                      │    sec/op    │   sec/op     vs base                │
CalcSigHash-32          2.151m ±  9%   1.910m ± 4%  -11.23% (p=0.009 n=10)
CalcWitnessSigHash-32   63.55µ ± 15%   42.25µ ± 6%  -33.53% (p=0.000 n=10)
geomean                 369.8µ         284.0µ       -23.18%

                      │    old.txt    │               new.txt                │
                      │     B/op      │     B/op      vs base                │
CalcSigHash-32          1260.3Ki ± 0%   911.2Ki ± 0%  -27.70% (p=0.000 n=10)
CalcWitnessSigHash-32    45.89Ki ± 0%   18.08Ki ± 0%  -60.61% (p=0.000 n=10)
geomean                  240.5Ki        128.3Ki       -46.63%

                      │  old.txt   │               new.txt                │
                      │ allocs/op  │ allocs/op   vs base                  │
CalcSigHash-32          712.0 ± 0%   623.0 ± 0%  -12.50% (p=0.000 n=10)
CalcWitnessSigHash-32   445.0 ± 0%   445.0 ± 0%        ~ (p=1.000 n=10) ¹
geomean                 562.9        526.5        -6.46%
```

PR #2078, opened 2023-12-24
Add OP_TXHASH and OP_CHECKTXHASHVERIFY opcodes

sputn1ck:
Implementation of OP_TXHASH and OP_CHECKTXHASHVERIFY, as per https://github.com/bitcoin/bips/pull/1500.

I mainly did this to learn about the BIP, maybe others might find this interesting and useful.

The code passes all testvectors specified in the BIP, I also added an example of restricting a transaction to be only spent to specified outputs.

I also added a rudimentary cache with a benchmark.

### DIFF
--- a/txscript/txhash_test.go
+++ b/txscript/txhash_test.go
@@ -317,6 +317,144 @@ func TestExampleTx(t *testing.T) {
 	}
 }
 
+// This test benchmarks the execution of the engine with and without the
+// txfs cache.
+func BenchmarkTxfsCache(t *testing.B) {
+	var (
+		inputCount            = 100
+		outputCount           = 100
+		outputValue           = int64(5000)
+		chekTxHashVerifyCount = 200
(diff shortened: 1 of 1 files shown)


txscript/txhash.go
```diff
@@ -0,0 +1,1291 @@
...
+
+	// TXFSOutputs is the bit that indicates that the outputs should be
+	// hashed into the TxHash.
+	TXFSOutputs uint8 = 1 << 6
+
+	// TXFSControl is the bit that indicates that the TxFieldSelector
+	// should be hashed into the TxHash.
+	TXFSControl uint8 = 1 << 7
+
+	// TXFSAll is the bit that indicates that all fields should be hashed
```
halseth, line 47: s/bit/byte

halseth (commented):
thanks for the implementation :)

Issue #2087, opened 2024-01-02
blockchain: accounting error (off-by-one?) in utxo set cache size 


Roasbeef:
Was running some sync tests with various configurations and spotted this upon restart: 
```
2024-01-01 16:48:11.662 [INF] CHAN: Flushing UTXO cache of 3127 MiB with 1 entries to disk. For large sizes, this can take up to several minutes...
```

1 entry, yet 3 GB cache size, smells like an accounting error....

Roasbeef:
cc @kcalvinalvin

kcalvinalvin:
Will take a look today but my initial assumption is that the 3GB part is how much the preallocated map is taking up.

The actual allocation happens like so:

1: User passes in the acceptable mem usage.
2: Pre-allocate a map based on this. It's usually like half of the mem usage (ex: 8gb of utxocache will alloc ~4gb initially)
3: utxo cache memory usage increases as each utxoentry is cached. The variable sized pkscript places a big part of this memory usage.

Did that happen at the very start?  If we're not in headers first, it'll try to "flush" the utxocache. If it was "flushing 1 entry" at the very first start that may be an off by one on the log message

kcalvinalvin:
Fixed with #2089

PR #2089, opened 2024-01-03
blockchain: set the lastflushtime when setting the lastflushhash

kcalvinalvin:
On startup when the headers-first mode is off, when receiving the first block, the periodic flush will trigger.  The lastflushtime wasn't set which resulted in the flush being triggered on the first block on restart.

Addresses #2087


guggero (approved):

halseth (approved):
LGTM

PR #2093, opened 2024-01-04
fix: correct the typos [skip ci]

threewebcode:
Use the change to rectify some typos.

### DIFF
--- a/txscript/opcode.go
+++ b/txscript/opcode.go
@@ -1278,7 +1278,7 @@ func opcodeIfDup(op *opcode, data []byte, vm *Engine) error {
 		return err
 	}
 
-	// Push copy of data iff it isn't zero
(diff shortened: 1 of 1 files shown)


txscript/opcode.go
```diff
@@ -1278,7 +1278,7 @@ func opcodeIfDup(op *opcode, data []byte, vm *Engine) error {
 		return err
 	}
 
-	// Push copy of data iff it isn't zero
+	// Push copy of data if it isn't zero
```
guggero, line 1281: https://en.wikipedia.org/wiki/If_and_only_if

threewebcode: Restore this change and please take a look again.

guggero (commented):
Looks good after squashing the two commits.

PR #2095, opened 2024-01-05
[NO MERGE] tapscript: enable OP_CAT

halseth:
[SHOULD NOT BE MERGED]

Enabling OP_CAT as specified in [BIP-CAT](https://github.com/bitcoin/bips/pull/1525)

Note this does currently lack any form of activation or flags for disabling it, hence many existing tests will fail.

Should not be merged before BIP is accepted and activation is decided on.

### DIFF
--- a/txscript/reference_test.go
+++ b/txscript/reference_test.go
@@ -63,15 +63,18 @@ func parseHex(tok string) ([]byte, error) {
 
 var testPrivKey, _ = btcec.NewPrivateKey()
 
+// Tags that are used to know how to parse the various elements of the
+// reference test json.
+const (
+	tagScript        = "#SCRIPT#"
+	tagControlblock  = "#CONTROLBLOCK#"
+	tagTaprootOutput = "#TAPROOTOUTPUT#"
+)
+
 // parseWitnessStack parses a json array of witness items encoded as hex into a
 // slice of witness elements.
 func parseWitnessStack(elements []interface{}) ([][]byte,
 	*IndexedTapScriptTree, error) {
-
-	const (
-		autogenScript       = "<AUTOGEN:SERIALIZED_SCRIPT>"
-		autogenControlblock = "<AUTOGEN:CONTROLBLOCK>"
-	)
 
 	var (
 		witness  = make([][]byte, len(elements))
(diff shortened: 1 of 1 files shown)


txscript/engine.go
```diff
@@ -335,8 +335,6 @@ func (vm *Engine) isBranchExecuting() bool {
 // conditional).
 func isOpcodeDisabled(opcode byte) bool {
 	switch opcode {
-	case OP_CAT:
-		return true
```
0xBEEFCAF3, line 339: Just curious. Does removing this case still throw some error when CAT (0x7e) is being used in a segwitv0 or legacy context?

halseth: Yes, if opcode is being executed without a taproot context, it will fall back to old behavior: https://github.com/btcsuite/btcd/pull/2095/commits/9541657983b8a54e3225e450bcd961e2c29eb29b#diff-27954f8a5853761d7c36b3084d7eb53d7c9a6788eac76b4b20dc573f723a8296R1954

txscript/opcode.go
```diff
@@ -1944,6 +1943,44 @@ func opcodeHash256(op *opcode, data []byte, vm *Engine) error {
...
+	if err != nil {
+		return err
+	}
+
+	x1, err := vm.dstack.PopByteArray()
+	if err != nil {
+		return err
+	}
+
+	var buf bytes.Buffer
```
halseth, line 1967: Note to self: could preallocate known number of bytes before copying.

txscript/engine.go
```diff
@@ -335,8 +335,6 @@ func (vm *Engine) isBranchExecuting() bool {
 // conditional).
 func isOpcodeDisabled(opcode byte) bool {
 	switch opcode {
-	case OP_CAT:
```
halseth, line 338: Similar to the [discussion in Inquistion](https://github.com/bitcoin-inquisition/bitcoin/pull/39#discussion_r1464005864), this check must still be triggered for legacy script, since the opcode won't be executed if inside a conditional.

txscript/opcode.go
halseth: Done: https://github.com/btcsuite/btcd/pull/2095/commits/5e4aee20946bb74aa42b10f80e4d6cd24d177680

txscript/engine.go
halseth: Done: https://github.com/btcsuite/btcd/pull/2095/commits/30ed9eccc127faccbb47f3e03a92825366c9d71f

halseth:
Updated in line with reference implementation.

PR #2108, opened 2024-01-22
netsync: don't update mempool/fee estimator unless we're synced up

kcalvinalvin:
Noticed during ibd that there's a slight overhead for handleBlockchainNotification on mempool/fee estimator updates. Since there's no reason to be looking at this while we're not caught up, return early and avoid the calls.


guggero (approved):

Roasbeef (approved):
LGTM 🧚🏽

PR #2123, opened 2024-02-16
blockchain: fix inconsistent utxocache and database on reorg

kcalvinalvin:
During reorgs, the assumption being made was that the utxocache is ignored and we're directly accessing the database for utxo set modifications.

However, after a block disconnect, the `BlockChain.chainLock`, the lock that prevents modifications to the utxo set, was being freed for the callback functions for the blockchain notifications.
https://github.com/btcsuite/btcd/blob/13152b35e191385a874294a9dbc902e48b1d71b0/blockchain/chain.go#L853-L858

If a caller calls either of the two methods below, the utxocache could be loaded with utxos that are fetched from the database:

https://github.com/btcsuite/btcd/blob/13152b35e191385a874294a9dbc902e48b1d71b0/blockchain/utxoviewpoint.go#L687-L693

https://github.com/btcsuite/btcd/blob/13152b35e191385a874294a9dbc902e48b1d71b0/blockchain/utxoviewpoint.go#L722-L732

Since the UTXOs fetched here are considered unmodified and therefore wouldn't be saved in the database when the `utxoCache` is flushed at the end of a reorganization, if a caller makes a call to the above two functions during a reorg, they could be mistakenly think that a utxo exists when it was removed from the utxo set.

Example:

There's a chain like so where a utxo created in `b1001` is spent in `b1002`. Blocks `b1002` and `b1001` are being disconnected out by a longer chain.
```
b1000() -> b1001(b1000.tx[1].out[0]) -> b1002(b1001.tx[1].out[0])
```

Mempool tries to fetch `b1001.tx[1].out[0]` three times in total. `2`, `8` and `11`.
At `6` the utxo `b1001.tx[1].out[0]` get loaded up to the cache. The cache never gets reset and that utxo gets fetched at `12`. When the reorg finishes, the mempool can no longer fetch `b1001.tx[1].out[0]`

```
         database                 UTXO cache                 BlockChain                                     Mempool

1.     {b1002.tx[0].out[0]}       {}                         chainLock.Lock() (for reorg)
2.     {b1002.tx[0].out[0]}       {}                                                                        Try to acquire lock (fetch output b1001.tx[1].out[0])
4.     {b1001.tx[1].out[0]}       {}                         disconnect b1002                               Try to acquire lock (fetch output b1001.tx[1].out[0])
5.     {}                         {}                         chainLock.Unlock() (subscriber callback)       Acquired lock (fetch output b1001.tx[1].out[0])
6.     {}                         {b1001.tx[1].out[0]}                                                      Fetched b1001.tx[1].out[0] (should happen)
7.     {}                         {b1001.tx[1].out[0]}       chainLock.Lock()   (finished callback)
8.     {}                         {b1001.tx[1].out[0]}                                                      Try to acquire lock (fetch output b1001.tx[1].out[0])
9.     {}                         {b1001.tx[1].out[0]}                                                      Try to acquire lock (fetch output b1001.tx[1].out[0])
10.    {b1000.tx[1].out[0]}       {b1001.tx[1].out[0]}       disconnect b1001                               Try to acquire lock (fetch output b1001.tx[1].out[0])
11.    {b1000.tx[1].out[0]}       {b1001.tx[1].out[0]}       chainLock.Unlock() (subscriber callback)       Acquired lock (fetch output b1001.tx[1].out[0])
12.    {b1000.tx[1].out[0]}       {b1001.tx[1].out[0]}                                                      Fetched b1001.tx[1].out[0] (shouldn't happen)
13.    {b1000.tx[1].out[0]}       {b1001.tx[1].out[0]}       chainLock.Lock()   (finished callback)
14.    {b1000.tx[1].out[0]}       {}                         Flush UTXO set
15.    {b1000.tx[1].out[0]}       {}                                                                        Try to acquire lock (fetch output b1001.tx[1].out[0])
16.    {b1000.tx[1].out[0]}       {}                                                                        Acquired lock (fetch output b1001.tx[1].out[0])
17.    {b1000.tx[1].out[0]}       {}                                                                        Did not find b1001.tx[1].out[0] (should happen)
```

### DIFF
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -569,7 +569,7 @@ func (b *BlockChain) getReorganizeNodes(node *blockNode) (*list.List, *list.List
 //
 // This function MUST be called with the chain state lock held (for writes).
 func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
-	view *UtxoViewpoint, stxos []SpentTxOut) error {
+	stxos []SpentTxOut) error {
 
 	// Make sure it's extending the end of the best chain.
 	prevHash := &block.MsgBlock().Header.PrevBlock
@@ -610,18 +610,6 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
 	state := newBestState(node, blockSize, blockWeight, numTxns,
 		curTotalTxns+numTxns, CalcPastMedianTime(node),
 	)
-
-	// If a utxoviewpoint was passed in, we'll be writing that viewpoint
-	// directly to the database on disk.  In order for the database to be
-	// consistent, we must flush the cache before writing the viewpoint.
-	if view != nil {
-		err = b.db.Update(func(dbTx database.Tx) error {
-			return b.utxoCache.flush(dbTx, FlushRequired, state)
-		})
-		if err != nil {
-			return err
-		}
-	}
 
 	// Atomically insert info into the database.
 	err = b.db.Update(func(dbTx database.Tx) error {
@@ -676,16 +664,6 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
 			return err
 		}
 
-		// Update the utxo set using the state of the utxo view.  This
-		// entails removing all of the utxos spent and adding the new
-		// ones created by the block.
-		//
-		// A nil viewpoint is a no-op.
-		err = dbPutUtxoView(dbTx, view)
-		if err != nil {
-			return err
-		}
-
 		// Update the transaction spend journal by adding a record for
 		// the block that contains all txos spent by it.
 		err = dbPutSpendJournalEntry(dbTx, block.Hash(), stxos)
@@ -707,12 +685,6 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
 	})
 	if err != nil {
 		return err
-	}
-
-	// Prune fully spent entries and mark all entries in the view unmodified
-	// now that the modifications have been committed to the database.
-	if view != nil {
-		view.commit()
 	}
 
 	// This node is now the end of the best chain.
@@ -1118,18 +1090,18 @@ func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List) error 
 		n := e.Value.(*blockNode)
 		block := attachBlocks[i]
 
-		// Update the view to mark all utxos referenced by the block
+		// Update the cache to mark all utxos referenced by the block
 		// as spent and add all transactions being created by this block
 		// to it.  Also, provide an stxo slice so the spent txout
 		// details are generated.
 		stxos := make([]SpentTxOut, 0, countSpentOutputs(block))
-		err := b.utxoCache.connectTransactions(block, &stxos)
+		err = b.utxoCache.connectTransactions(block, &stxos)
 		if err != nil {
 			return err
 		}
 
 		// Update the database and chain state.
-		err = b.connectBlock(n, block, nil, stxos)
+		err = b.connectBlock(n, block, stxos)
 		if err != nil {
 			return err
 		}
@@ -1226,7 +1198,7 @@ func (b *BlockChain) connectBestChain(node *blockNode, block *btcutil.Block, fla
 		}
 
 		// Connect the block to the main chain.
-		err = b.connectBlock(node, block, nil, stxos)
+		err = b.connectBlock(node, block, stxos)
 		if err != nil {
 			// If we got hit with a rule error, then we'll mark
 			// that status of the block as invalid and flush the
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -796,6 +796,15 @@ func (b *BlockChain) disconnectBlock(node *blockNode, block *btcutil.Block, view
 			return err
 		}
 
+		// Flush the cache on every disconnect.  Since the code for
+		// reorganization modifies the database directly, the cache
+		// will be left in an inconsistent state if we don't flush it
+		// prior to the dbPutUtxoView that happends below.
+		err = b.utxoCache.flush(dbTx, FlushRequired, state)
+		if err != nil {
+			return err
+		}
+
 		// Update the utxo set using the state of the utxo view.  This
 		// entails restoring all of the utxos spent and removing the new
 		// ones created by the block.
@@ -880,20 +889,14 @@ func countSpentOutputs(block *btcutil.Block) int {
 //
 // This function may modify node statuses in the block index without flushing.
 //
+// This function never leaves the utxo set in an inconsistent state for block
+// disconnects.
+//
 // This function MUST be called with the chain state lock held (for writes).
 func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List) error {
 	// Nothing to do if no reorganize nodes were provided.
 	if detachNodes.Len() == 0 && attachNodes.Len() == 0 {
 		return nil
-	}
-
-	// The rest of the reorg depends on all STXOs already being in the database
-	// so we flush before reorg.
-	err := b.db.Update(func(dbTx database.Tx) error {
-		return b.utxoCache.flush(dbTx, FlushRequired, b.BestSnapshot())
-	})
-	if err != nil {
-		return err
 	}
 
 	// Ensure the provided nodes match the current best chain.
@@ -957,7 +960,7 @@ func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List) error 
 
 		// Load all of the utxos referenced by the block that aren't
 		// already in the view.
-		err = view.fetchInputUtxos(b.db, nil, block)
+		err = view.fetchInputUtxos(nil, b.utxoCache, block)
 		if err != nil {
 			return err
 		}
@@ -1024,7 +1027,7 @@ func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List) error 
 		// checkConnectBlock gets skipped, we still need to update the UTXO
 		// view.
 		if b.index.NodeStatus(n).KnownValid() {
-			err = view.fetchInputUtxos(b.db, nil, block)
+			err = view.fetchInputUtxos(nil, b.utxoCache, block)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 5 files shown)


blockchain/fullblocks_test.go
```diff
@@ -146,6 +146,62 @@ func TestFullBlocks(t *testing.T) {
...
+		expectedCallBack := func(notification *blockchain.Notification) {
+			switch notification.Type {
+
+			case blockchain.NTBlockDisconnected:
+				block, ok := notification.Data.(*btcutil.Block)
+				if !ok {
+					t.Fatalf("expected a block")
+				}
+
+				if block.Hash().IsEqual(&item.BlockHash) {
```
guggero, line 159: nit: invert condition and `return` to continue non-indented below?

guggero (approved):
Thanks a lot for the fix! LGTM :tada:

blockchain/fullblocks_test.go
kcalvinalvin: Applied in the latest push

blockchain/fullblocktests/generate.go
```diff
@@ -2044,6 +2068,55 @@ func Generate(includeLargeReorg bool) (tests [][]TestInstance, err error) {
 	}
 	accepted()
 
+	// Create a chain where the utxo created in b82a is spent in b83a.
```
Roasbeef, line 2071: If I check out only this commit, then run the tests (expecting a failing test case), I get this panic: 
```
    fullblocks_test.go:220: Testing block b82 (hash 42246d9ea2cef3bac886bf25591884535d4206da1e5a3ced9317c4015a4ebec9, height 129)
    fullblocks_test.go:339: Testing tip for block b83a (hash 119f8757e953251625d83e585718acc4f43fa0fd12a657e885e08b1294158e5f, height 130)
    fullblocks_test.go:220: Testing block b83 (hash 4f95ef6bb94d5bcec3700a619caedb103bf88c8a37fe0c36da9c8a0ef5cc6103, height 130)
    fullblocks_test.go:339: Testing tip for block b83a (hash 119f8757e953251625d83e585718acc4f43fa0fd12a657e885e08b1294158e5f, height 130)
    fullblocks_test.go:220: Testing block b84 (hash 0599c9a4b9f468b3d30a2734e7da26830cbc67faeab1b65d40adedf5f3dde3fd, height 131)
    fullblocks_test.go:198: unexpected utxo 91f9eaef0b25c7905841d855c70753f51cc71d714f9ffd8f0a0d88a248653f9a:0 to exist but it's not nil and not spent
    fullblocks_test.go:200: unexpected utxo 91f9eaef0b25c7905841d855c70753f51cc71d714f9ffd8f0a0d88a248653f9a:0 exists but it's not nil and not spent


fatal error: sync: Unlock of unlocked RWMutex

goroutine 1628 [running]:
sync.fatal({0x1002b6e4c?, 0x1400009a028?})
	/Users/roasbeef/go/src/runtime/panic.go:1061 +0x20
sync.(*RWMutex).Unlock(0x14005dea07c)
	/Users/roasbeef/go/src/sync/rwmutex.go:209 +0x74
runtime.Goexit()
	/Users/roasbeef/go/src/runtime/panic.go:523 +0x14c
testing.(*common).FailNow(0x140013a0d00)
	/Users/roasbeef/go/src/testing/testing.go:999 +0x48
testing.(*common).Fatalf(0x140013a0d00, {0x1002bfb90?, 0x8390e4f3e79edc26?}, {0x14002db14b8?, 0x14000000000?, 0x0?})
	/Users/roasbeef/go/src/testing/testing.go:1083 +0x64
github.com/btcsuite/btcd/blockchain_test.TestFullBlocks.func1.2(0x100434f20?)
	/Users/roasbeef/gocode/src/github.com/btcsuite/btcd/blockchain/fullblocks_test.go:200 +0x1dc
github.com/btcsuite/btcd/blockchain.(*BlockChain).sendNotification(0x14005dea000, 0x2, {0x100467800?, 0x14006b97ab0})
	/Users/roasbeef/gocode/src/github.com/btcsuite/btcd/blockchain/notifications.go:78 +0xf8
github.com/btcsuite/btcd/blockchain.(*BlockChain).disconnectBlock(0x14005dea000, 0x14001213d80, 0x14006b97ab0, 0x14005138b10)
	/Users/roasbeef/gocode/src/github.com/btcsuite/btcd/blockchain/chain.go:857 +0x4a0
github.com/btcsuite/btcd/blockchain.(*BlockChain).reorganizeChain(0x14005dea000, 0x14003d7bf80, 0x14003d7bf50)
	/Users/roasbeef/gocode/src/github.com/btcsuite/btcd/blockchain/chain.go:1093 +0xc40
github.com/btcsuite/btcd/blockchain.(*BlockChain).connectBestChain(0x14005dea000, 0x14001213f80, 0x14006b97880, 0x0)
	/Users/roasbeef/gocode/src/github.com/btcsuite/btcd/blockchain/chain.go:1288 +0x670
github.com/btcsuite/btcd/blockchain.(*BlockChain).maybeAcceptBlock(0x14005dea000, 0x14006b97880, 0x489868?)
	/Users/roasbeef/gocode/src/github.com/btcsuite/btcd/blockchain/accept.go:79 +0x160
github.com/btcsuite/btcd/blockchain.(*BlockChain).ProcessBlock(0x14005dea000, 0x14006b97880, 0x0)
	/Users/roasbeef/gocode/src/github.com/btcsuite/btcd/blockchain/process.go:228 +0x258
github.com/btcsuite/btcd/blockchain_test.TestFullBlocks.func2({{0x1002aaa45, 0x3}, 0x1400116c180, 0x83, 0x1, 0x0})
	/Users/roasbeef/gocode/src/github.com/btcsuite/btcd/blockchain/fullblocks_test.go:223 +0x138
github.com/btcsuite/btcd/blockchain_test.TestFullBlocks(0x140013a0d00)
	/Users/roasbeef/gocode/src/github.com/btcsuite/btcd/blockchain/fullblocks_test.go:358 +0x3d0
testing.tRunner(0x140013a0d00, 0x100486208)
	/Users/roasbeef/go/src/testing/testing.go:1595 +0xe8
created by testing.(*T).Run in goroutine 1
	/Users/roasbeef/go/src/testing/testing.go:1648 +0x33c
```

blockchain/chain.go
```diff
@@ -796,6 +796,15 @@ func (b *BlockChain) disconnectBlock(node *blockNode, block *btcutil.Block, view
 			return err
 		}
 
+		// Flush the cache on every disconnect.  Since the code for
+		// reorganization modifies the database directly, the cache
+		// will be left in an inconsistent state if we don't flush it
+		// prior to the dbPutUtxoView that happends below.
+		err = b.utxoCache.flush(dbTx, FlushRequired, state)
```
Roasbeef, line 803: 👍

blockchain/chain.go
```diff
@@ -880,22 +889,16 @@ func countSpentOutputs(block *btcutil.Block) int {
 //
 // This function may modify node statuses in the block index without flushing.
 //
+// This function never leaves the utxo set in an inconsistent state for block
+// disconnects.
+//
```
Roasbeef, line 894: Stray empty comment line here.

blockchain/chain.go
```diff
@@ -1024,7 +1027,7 @@ func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List) error
 		// checkConnectBlock gets skipped, we still need to update the UTXO
 		// view.
 		if b.index.NodeStatus(n).KnownValid() {
-			err = view.fetchInputUtxos(b.db, nil, block)
+			err = view.fetchInputUtxos(nil, b.utxoCache, block)
```
Roasbeef, line 1030: With this change, is it correct that now `fetchInputUtxos` will never be called with a `nil` second argument, so the extra logic to account for that case can be removed?

blockchain/chain.go
```diff
@@ -1089,51 +1102,52 @@ func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List) error
...
 			return err
 		}
+
+		// Flush the cache again after the disconnect block because the
+		// utxo cache briefly becomes accessible due to the chainLock
+		// being unlocked for the notification callback.
+		//
+		// This is needed as we cannot leave the utxo set in an inconsistent
+		// state during block disconnects.
+		err = b.db.Update(func(dbTx database.Tx) error {
```
Roasbeef, line 1118: Is this needed given that in `disconnectBlock`, we'll now (as of this commit) flush the cache in the same db txns that we update the rest of the indexes in?

Roasbeef (commented):
Nice work tracking down this fix! Have some questions around some lingering logic that looks redundant after this change, and a panic I get in the unit tests w/ the first commit (new tests), but w/o the second (the fix).

blockchain/fullblocktests/generate.go
kcalvinalvin: Ah yeah that's supposed to happen for any errors that happen during `b.sendNotification()` because the `t.Fatal(err)` happens first and then the `b.chainLock.Lock()` never gets called.

https://github.com/btcsuite/btcd/blob/13152b35e191385a874294a9dbc902e48b1d71b0/blockchain/chain.go#L853-L858

The `defer` in `ProcessBlock` will still get triggered so the runtime complains that you've unlocked an unlocked lock.

https://github.com/btcsuite/btcd/blob/13152b35e191385a874294a9dbc902e48b1d71b0/blockchain/process.go#L142-L144

I guess we should do the below. I'll make a separate PR for this as well.

```diff
diff --git a/blockchain/chain.go b/blockchain/chain.go
index 60420022..416ed929 100644
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -730,9 +730,11 @@ func (b *BlockChain) connectBlock(node *blockNode, block *btcutil.Block,
        // Notify the caller that the block was connected to the main chain.
        // The caller would typically want to react with actions such as
        // updating wallets.
-       b.chainLock.Unlock()
-       b.sendNotification(NTBlockConnected, block)
-       b.chainLock.Lock()
+       func() {
+               b.chainLock.Unlock()
+               defer b.chainLock.Lock()
+               b.sendNotification(NTBlockConnected, block)
+       }()
 
        // Since we may have changed the UTXO cache, we make sure it didn't exceed its
        // maximum size.  If we're pruned and have flushed already, this will be a no-op.
@@ -853,9 +855,11 @@ func (b *BlockChain) disconnectBlock(node *blockNode, block *btcutil.Block, view
        // Notify the caller that the block was disconnected from the main
        // chain.  The caller would typically want to react with actions such as
        // updating wallets.
-       b.chainLock.Unlock()
-       b.sendNotification(NTBlockDisconnected, block)
-       b.chainLock.Lock()
+       func() {
+               b.chainLock.Unlock()
+               defer b.chainLock.Lock()
+               b.sendNotification(NTBlockDisconnected, block)
+       }()
 
        return nil
}
```

kcalvinalvin: https://github.com/btcsuite/btcd/pull/2128

blockchain/chain.go
kcalvinalvin: It's kinda nice to separate this from

`// This function MUST be called with the chain state lock held (for writes).`

right? Since that's the format used in other parts of the code as well.

Though I can get rid of it if you think it's better

blockchain/chain.go
kcalvinalvin: Yeah this isn't needed if we get rid of the option to fetch from the database in `fetchInputUtxos`

blockchain/chain.go
kcalvinalvin: The current code has it so that it still fetches from the database directly on disconnects but this isn't needed and causes more problems.

We can get rid of the option for calling `fetchInputUtxos` with the database.

blockchain/chain.go
kcalvinalvin: Addressed in the latest push

blockchain/chain.go
kcalvinalvin: Addressed in the latest push

blockchain/chain.go
Roasbeef: Ah nvm, when I looked at the diff originally, didn't see the comment below.

Roasbeef (approved):
LGTM 🍖

PR #2125, opened 2024-02-23
Add `gettxspendingprevout` for `btcd` and fix version check

yyforyongyu:
This PR adds the RPC `gettxspendingprevout` for `btcd` so we can do mempool lookups. This RPC is also ported to the `rpcclient` to replace the raw request made on the btcwallet side.

In addition, we now start to track `btcd` versions in `rpcclient` to make sure RPC updates are handled properly.


rpcserver.go
```diff
@@ -3906,6 +3907,49 @@ func handleTestMempoolAccept(s *rpcServer, cmd interface{},
 	return results, nil
 }
 
+// handleGetTxSpendingPrevOut implements the gettxspendingprevout command.
+func handleGetTxSpendingPrevOut(s *rpcServer, cmd interface{},
```
Roasbeef, line 3911: Short and sweet 😎

rpcserver_test.go
```diff
@@ -411,3 +412,86 @@ func TestHandleTestMempoolAcceptFees(t *testing.T) {
...
+// TestGetTxSpendingPrevOut checks that handleGetTxSpendingPrevOut handles the
+// cmd as expected.
+func TestGetTxSpendingPrevOut(t *testing.T) {
+	t.Parallel()
+
+	require := require.New(t)
+
+	// Create a mock mempool.
+	mm := &mempool.MockTxMempool{}
+	defer mm.AssertExpectations(t)
```
Roasbeef, line 425: 👍 

started to use this defer pattern elsewhere to ensure that the expectations are actually validated

rpcclient/infrastructure.go
```diff
@@ -1648,6 +1658,9 @@ func parseBitcoindVersion(version string) BackendVersion {
 	case version < bitcoind22Str:
 		return BitcoindPre22
 
+	case version < bitcoind24Str:
+		return BitcoindPre24
```
Roasbeef: IIUC, we need to start to check this in `btcwallet`? FWIW, didn't see it in the latest PR of yours in that repo.

integration/chain_test.go
```diff
@@ -0,0 +1,144 @@
...
+			}
+		})
+	}
+}
+
+// createTxInMempool creates a tx and puts it in the mempool.
+func createTxInMempool(t *testing.T, r *rpctest.Harness) *wire.MsgTx {
+	// Create a fresh output for usage within the test below.
+	const outputValue = btcutil.SatoshiPerBitcoin
+	outputKey, testOutput, testPkScript, err := makeTestOutput(r, t,
```
Roasbeef, line 110: Funky line folding, can be: 
```go
makeTestOutput(
        r, t, outputValue,
)
```

integration/chain_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+	addrScript, err := txscript.PayToAddrScript(addr)
+	require.NoError(t, err)
+
+	tx.AddTxOut(&wire.TxOut{
+		PkScript: addrScript,
+		Value:    outputValue - 1000,
+	})
+
+	sigScript, err := txscript.SignatureScript(tx, 0, testPkScript,
```
Roasbeef, line 134: Same line folding here as above

rpcclient/chain.go
```diff
@@ -441,7 +441,7 @@ func unmarshalGetBlockChainInfoResultSoftForks(chainInfo *btcjson.GetBlockChainI
 	version BackendVersion, res []byte) error {
 
 	// Versions of bitcoind on or after v0.19.0 use the unified format.
-	if version > BitcoindPre19 {
+	if version.SupportUnifiedSoftForks() {
```
Roasbeef, line 444: 👍

Roasbeef (approved):
LGTM 🌂

integration/chain_test.go
yyforyongyu: hmm copy-fu, now fixed!

rpcclient/infrastructure.go
yyforyongyu: I think we'll need to check this in `lnd` at the callsite. If we check it in `btcwallet`, then it's another error to be returned to the caller. On the other hand, I think the `rpcclient` package should be put in `btcwallet` instead to keep `btcd` "pure" - came up with a draft PR for this when I was dealing with updating errors returned from `TestMempoolAccept`, will put more details there.

Roasbeef: >  I think the rpcclient package should be put in btcwallet instead to keep btcd "pure" - came up with a draft PR for this when I was dealing with updating errors returned from TestMempoolAccept, will put more details there.

Don't think I agree with this, `rpcclient` as is does generic JSON-RPC for conformant Bitcoin nodes. We even use the client to test directly against (in itests) the server-side implementation of the various RPC calls. 

I don't see why the check can't just be in the wallet though. We already have that check for `bitcoind` when we wan to use the newer set of RPC calls.

Roasbeef (approved):
LGTM 🎒

saubyk:
cc: @Crypt-iQ for review

Chinwendu20:
Nice pr @yyforyongyu I like the progression. 

question/suggestion: Do you think we should separate the implementation of the Backend version interface into separate files? like bitcoin.go and btcd.go in the same folder. In the future if we want to track more backends we just create a new file and that way it is easier to follow?

rpcserver.go
```diff
@@ -3906,6 +3907,49 @@ func handleTestMempoolAccept(s *rpcServer, cmd interface{},
...
+	closeChan <-chan struct{}) (interface{}, error) {
```
guggero, line 3912: nit: since `closeChan` is unused, should we just use `_` as the argument name?

rpcserver_test.go
```diff
@@ -411,3 +412,86 @@ func TestHandleTestMempoolAcceptFees(t *testing.T) {
...
+	// Create a request that will cause an error.
+	cmd := &btcjson.GetTxSpendingPrevOutCmd{
+		Outputs: []*btcjson.GetTxSpendingPrevOutCmdOutput{
+			{Txid: "invalid"},
+		},
+	}
+
+	// Call the method handler and assert the error is returned.
+	closeChan := make(chan struct{})
+	results, err := handleGetTxSpendingPrevOut(s, cmd, closeChan)
```
guggero, line 443: nit: could just pass in `nil` since `closeChan` is not used anyway.

rpcclient/backend_version.go
```diff
@@ -2,13 +2,32 @@ package rpcclient
 
 import "strings"
 
-// BackendVersion represents the version of the backend the client is currently
-// connected to.
-type BackendVersion uint8
+// BackendVersion defines an interface to handle the version of the backend
+// used by the client.
+type BackendVersion interface {
```
guggero, line 7: Nice :star_struck:

rpcclient/infrastructure.go
```diff
@@ -1589,36 +1589,37 @@ func (c *Client) BackendVersion() (BackendVersion, error) {
 	// Parse the btcd version and cache it.
 	case nil:
 		log.Debugf("Detected btcd version: %v", info.Version)
-		version := Btcd
-		c.backendVersion = &version
-		return *c.backendVersion, nil
+		version := parseBtcdVersion(info.Version)
+		c.backendVersion = version
```
guggero, line 1593: nit: why not `c.backendVersion = parseBtcdVersion(info.Version)` directly?

guggero (approved):
Very nice, LGTM :tada:

PR #2128, opened 2024-02-27
blockchain: always relock chainLock for subscription callbacks

kcalvinalvin:
For various b.sendNotifcation() callbacks, if a runtime panic happens, we don't get any useful debugging information since the error that happens first is the "unlock of unlocked mutex" error.

This is because we temporarily unlock the chainLock for callbacks and then relock them.  However, since the relocking code is executed after the completion of the callback, if an error happens during that callback, we never relock the chainLock.

Switching to an anonymous function and having the unlock code as a defer will ensure that the lock always relocks.

### DIFF
--- a/blockchain/accept.go
+++ b/blockchain/accept.go
@@ -84,9 +84,11 @@ func (b *BlockChain) maybeAcceptBlock(block *btcutil.Block, flags BehaviorFlags)
 	// Notify the caller that the new block was accepted into the block
 	// chain.  The caller would typically want to react by relaying the
 	// inventory to other peers.
-	b.chainLock.Unlock()
-	b.sendNotification(NTBlockAccepted, block)
-	b.chainLock.Lock()
+	func() {
+		b.chainLock.Unlock()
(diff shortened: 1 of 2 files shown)


blockchain/accept.go
```diff
@@ -84,9 +84,11 @@ func (b *BlockChain) maybeAcceptBlock(block *btcutil.Block, flags BehaviorFlags)
 	// Notify the caller that the new block was accepted into the block
 	// chain.  The caller would typically want to react by relaying the
 	// inventory to other peers.
-	b.chainLock.Unlock()
-	b.sendNotification(NTBlockAccepted, block)
-	b.chainLock.Lock()
+	func() {
```
Roasbeef, line 87: What exactly does the function closure achieve here? Can't we just do the plain defer?

Roasbeef: Is it attempting to rely on defer order execution after a panic?

Roasbeef (approved):
LGTM ✈️

PR #2131, opened 2024-03-05
addrmgr: fix intermittent addrmanager_internal_test bug

kcalvinalvin:
Sometimes the tests in the addrmanager_internal_test will fail with: "addrmanager_internal_test.go: expected to find 5 addresses, found 4"

This is because the generated address with randAddr() may not always generate an address that's routable. If the address is not routable, that address will not be added to the manager when calling AddAddress().

This bug is fixed by creating a new function routableRandAddr() that always creates a routable random address.


guggero (approved):
Nice fix, LGTM :tada:

Roasbeef (approved):
LGTM 🧩

PR #2134, opened 2024-03-07
blockchain: fix a bug where a duplicate entry is possible in the mapslice

kcalvinalvin:
Adds a test and fixes the bug to fix a case where a duplicate entry in the mapslice is possible.

Not sure if this is related to the other cases we've been seeing since the outpoints should never have duplicates, fixing this is good since the code is assuming that the mapslice will not have duplicates.


Roasbeef (approved):
LGTM 🛺

Issue #2059, opened 2023-11-11
ecdsa.ParseDERSignature() succeeds for an invalid signature (append arbitrary data)


benma:
Appending garbage data to a valid DER signature does not fail `ParseDERSignature()`.

See for example this modification of the unit tests, which does not fail the unit tests:

https://github.com/btcsuite/btcd/compare/master...benma:btcd:der-sigs

The documentation of ParseDerSignature() does not mention anything like this. I would expect signatures longer than 72 bytes to be invalid and return an error.

Is this a bug?

benma:
Wdyt @guggero ?

Roasbeef:
Hi @benma, 

In the normal execution path, any calls to `ParseDERSignature` are [preceded by sanity checks](https://github.com/btcsuite/btcd/blob/d15dd7108765d4ae810dae6c5446b81ee76f3fbf/txscript/sigvalidate.go#L70-L78) implemented by [`checkSignatureEncoding`](https://github.com/btcsuite/btcd/blob/d15dd7108765d4ae810dae6c5446b81ee76f3fbf/txscript/engine.go#L1217). `checkSignatureEncoding` prevents such signatures from being accepted as it has a [raw max size check](https://github.com/btcsuite/btcd/blob/d15dd7108765d4ae810dae6c5446b81ee76f3fbf/txscript/engine.go#L1288-L1292). Other sanity checks including the BIP 66 low-s constraint is also verified in that routine. 

As far as `ParseDerSignature` w.r.t your example above: the encoded length is unchanged (second byte in the signature), so the signature can still be parsed as valid. The encoded [length is used as a guide when parsing, causing the appended bytes to be ignored](https://github.com/btcsuite/btcd/blob/d15dd7108765d4ae810dae6c5446b81ee76f3fbf/btcec/ecdsa/signature.go#L89-L90).

In summary, such a signature would never be accepted as valid via normal execution paths. Precondition checks in the VM will cause such signatures to be rejected before we attempt to fully parse them.

benma:
@Roasbeef thanks for the response!

I can see that internally for btcd, there is no problem. However, `ParseDERSignature()` is an exported function and part of the `btcec` library that users other than btcd use. I think it would make sense if the function by itself only accepts valid DER-encoded signatures, or otherwise document in the docstring that it parses it only from the beginning of a potentially larger byte slice so that there are no surprises. Is there any downside into having `ParseDERSignature()` be more strict and fail for invalid sigs?

[...]

PR #2135, opened 2024-03-07
Add check for maximum signature length in ecdsa.ParseDERSignature

SulaimanAminuBarkindo:
* Updated the `ecdsa.ParseDERSignature` to include the MaxSigLen constant, which defines the maximum length of a DER encoded signature.
* The MaxSigLen constant is set to 72 bytes, as per the explanation in the comment.
* A new test case has been added to test the functionality with a long signature.
* The test case checks if signatures longer than the specified maximum length are properly handled.
* The test ensures that signatures exceeding the maximum length are correctly identified as invalid.

Fixes https://github.com/btcsuite/btcd/issues/2059


guggero (approved):
Thanks for the PR, LGTM :tada:

Roasbeef (approved):
LGTM 🪡

PR #2142, opened 2024-03-21
btcclient+btcjson: defaultMaxFeeRate to BTC/kvB

YusukeShimizu:
`defaultMaxFeeRate` was set to 1e8 / 10(sat/kb) as a parameter.  
But BTC/kvB is the expected value, so the units was wrong.  
This updates `defaultMaxFeeRate` to BTC/kvB and sets it to 0.1, which is the default value of Bitcoin Core.  
* https://bitcoincore.org/en/doc/26.0.0/rpc/rawtransactions/sendrawtransaction/

Because `maxFeeRate` sanity check has been added in bitcoin core v27.0 or later, sendrawtransaction cannot be executed without this change.
* https://github.com/bitcoin/bitcoin/pull/29434

Fixes https://github.com/lightningnetwork/lnd/issues/8571

### DIFF
--- a/btcjson/chainsvrcmds.go
+++ b/btcjson/chainsvrcmds.go
@@ -862,7 +862,7 @@ func (a *AllowHighFeesOrMaxFeeRate) UnmarshalJSON(data []byte) error {
 	case bool:
 		a.Value = Bool(v)
 	case float64:
-		a.Value = Int32(int32(v))
+		a.Value = Float64(v)
 	default:
 		return fmt.Errorf("invalid allowhighfees or maxfeerate value: "+
 			"%v", unmarshalled)
@@ -893,9 +893,10 @@ func NewSendRawTransactionCmd(hexTx string, allowHighFees *bool) *SendRawTransac
 
 // NewSendRawTransactionCmd returns a new instance which can be used to issue a
 // sendrawtransaction JSON-RPC command to a bitcoind node.
+// maxFeeRate is the maximum fee rate for the transaction in BTC/kvB.
 //
 // A 0 maxFeeRate indicates that a maximum fee rate won't be enforced.
-func NewBitcoindSendRawTransactionCmd(hexTx string, maxFeeRate int32) *SendRawTransactionCmd {
+func NewBitcoindSendRawTransactionCmd(hexTx string, maxFeeRate float64) *SendRawTransactionCmd {
 	return &SendRawTransactionCmd{
 		HexTx: hexTx,
 		FeeSetting: &AllowHighFeesOrMaxFeeRate{
--- a/btcjson/chainsvrcmds.go
+++ b/btcjson/chainsvrcmds.go
@@ -16,6 +16,10 @@ import (
 	"github.com/btcsuite/btcd/wire"
 )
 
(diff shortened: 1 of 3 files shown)


btcjson/chainsvrcmds.go
```diff
@@ -895,7 +895,7 @@ func NewSendRawTransactionCmd(hexTx string, allowHighFees *bool) *SendRawTransac
 // sendrawtransaction JSON-RPC command to a bitcoind node.
 //
 // A 0 maxFeeRate indicates that a maximum fee rate won't be enforced.
-func NewBitcoindSendRawTransactionCmd(hexTx string, maxFeeRate int32) *SendRawTransactionCmd {
+func NewBitcoindSendRawTransactionCmd(hexTx string, maxFeeRate float64) *SendRawTransactionCmd {
```
guggero, line 898: Should we add a note in the Godoc comment above what the unit for the `maxFeeRate` is?

guggero (approved):
Thanks a lot for the report and the fix!
I confirmed locally that this fixes the `Fee rates larger than or equal to 1BTC/kvB are not accepted` error in `lnd`.
I'll create a PR to run full `lnd` unit and integration tests on `bitcoind v27.0` (rc1) soon.

btcjson/chainsvrcmds.go
YusukeShimizu: I added a comment about `maxFeeRate` unit.
https://github.com/btcsuite/btcd/pull/2142/commits/714e994d72681988a1866e79abe136e1aebbb05e  
I will rebase after review.

> We have some of this in lnd

Would it be possible to tell me where in lnd the above mentioned file is located?

I assume that adding a type alias for BTC/kvB will be handled in a separate PR once it is decided, as it requires an update to btcutil.

guggero: The fee rate types are here: https://github.com/lightningnetwork/lnd/blob/60bc30dd08bb6abb39d7f8481ebd990d587e5469/lnwallet/chainfee/rates.go.

Yes, because of the Go modules, there first needs to be a PR in `btcutil`, then after merging that a tag is pushed and then the code can be used in this file here.

But maybe, because `bitcoind` uses its own fee rate type (`BTC/kvB` instead of `sat/kWU` we use in `lnd`), it would be okay to define the type in this package directly, as AFAIK it's the only place where we actually use `BTC/kvB`.
It looks like the fee rate is used for `fundrawtransaction` and `sendrawtransaction` currently.

YusukeShimizu: Thank you for sharing your location.

Yes, I agree with adding the type alias here.
It is used by `TestMempoolAccept`, `fundrawtransaction` and `SendRawTransaction`, so I am replacing their definitions with this type alias

https://github.com/btcsuite/btcd/pull/2142/commits/bd5af2b56fc36c1be1729e459b95d7f575e0134a

guggero (commented):
Can you please squash the commits? Otherwise looks good to me, thank you!

YusukeShimizu:
I have squashed it.

Roasbeef (approved):
LGTM 💸

PR #2153, opened 2024-04-02
blockchain: Add IsAncestor method to blockNode

kcalvinalvin:
IsAncestor() provides functionality for testing if a block node is an ancestor of anther block node.

Cherry-picked from [0d6a93e687e3470608cccfc58cc4a2ddb5ea8262](https://github.com/utreexo/utreexod/commit/0d6a93e687e3470608cccfc58cc4a2ddb5ea8262).

It's a part of #2154 as this method is used in utreexod code for testing invalidateblock and reconsiderblock methods on blockchain.

### DIFF
--- a/blockchain/blockindex.go
+++ b/blockchain/blockindex.go
@@ -274,6 +274,28 @@ func (node *blockNode) RelativeAncestorCtx(distance int32) HeaderCtx {
 	return ancestor
 }
 
+// IsAncestor returns if the other node is an ancestor of this block node.
+func (node *blockNode) IsAncestor(otherNode *blockNode) bool {
+	// Return early as false if the otherNode is nil.
+	if otherNode == nil {
+		return false
+	}
+
+	ancestor := node.Ancestor(otherNode.height)
+	if ancestor == nil {
+		return false
+	}
+
+	// If the otherNode has the same height as me, then the returned
+	// ancestor will be me.  Return false since I'm not an ancestor of me.
+	if node.height == ancestor.height {
+		return false
+	}
+
+	// Return true if the fetched ancestor is other node.
+	return ancestor.Equals(otherNode)
+}
+
 // RelativeAncestor returns the ancestor block node a relative 'distance' blocks
 // before this node.  This is equivalent to calling Ancestor with the node's
 // height minus provided distance.
--- a/blockchain/blockindex.go
+++ b/blockchain/blockindex.go
@@ -135,6 +135,20 @@ func newBlockNode(blockHeader *wire.BlockHeader, parent *blockNode) *blockNode {
 	return &node
 }
 
+// Equals compares all the fields of the block node except for the parent and
+// ancestor and returns true if they're equal.
+func (node *blockNode) Equals(other *blockNode) bool {
+	return node.hash == other.hash &&
+		node.workSum.Cmp(other.workSum) == 0 &&
+		node.height == other.height &&
+		node.version == other.version &&
+		node.bits == other.bits &&
+		node.nonce == other.nonce &&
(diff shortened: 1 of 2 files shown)


blockchain/blockindex.go
```diff
@@ -260,6 +260,16 @@ func (node *blockNode) RelativeAncestorCtx(distance int32) HeaderCtx {
 	return ancestor
 }
 
+// IsAncestor returns if the other node is an ancestor of this block node.
+func (node *blockNode) IsAncestor(otherNode *blockNode) bool {
+	ancestor := node.Ancestor(otherNode.height)
```
Roasbeef, line 265: This can return `nil`, so we should do a nil check here.

blockchain/blockindex.go
```diff
@@ -260,6 +260,16 @@ func (node *blockNode) RelativeAncestorCtx(distance int32) HeaderCtx {
...
+	if ancestor == node {
```
Roasbeef, line 266: Both the arg, the method, and the method target are pointers, so this will only succeed if all the pointers across the structure are consistent. Instead, I think we should refer here so we do an actual equality check. We may actually want to consider making an `Equals` method for `blockNode`.

Roasbeef, line 266: When would this case ever be triggered (`ancestor == node`)? 

From revisiting the implementation of `Ancestor`, only if `n` is `nil` would the same node be returned, or if they have an equal height. Perhaps we can make the latter case more explicit by just doing a height comparison?

Roasbeef (changes requested):
Nice addition! Mostly straight forward to follow, though I'm concerned that the equality check is done using pointers, rather than the value itself. Eg: if we make a new instance or copy of `otherNode` (that's already in the index, and is an ancestor of `node`), then from my reading we'll return `false`, when we should return `true`.

blockchain/blockindex.go
kcalvinalvin: Yeah the only case this would be triggered is if the equal height is passed to `Ancestor`. I'll change it to a height comparison

kcalvinalvin: Yeah I'll add an Equals here

blockchain/chain_test.go
```diff
@@ -1155,3 +1155,159 @@ func TestChainTips(t *testing.T) {
...
+
+	// Is 1 an ancestor of 2?
+	//
+	// 	genesis -> 1  -> 2  -> 3 (active)
+	//            \ -> 1a (valid-fork)
+	//            \ -> 1b (invalid)
+	shouldBeTrue = branch0Nodes[1].IsAncestor(branch0Nodes[0])
+	if !shouldBeTrue {
+		t.Errorf("TestIsAncestor fail. Node %s is an ancestor of node %s but got false",
+			branch0Nodes[1].hash.String(), branch0Nodes[2].hash.String())
```
Crypt-iQ, line 1208: wrong indices in `t.Errorf`

Crypt-iQ (commented):
Looks good once laolu's comments addressed. One nit in the test

blockchain/blockindex.go
kcalvinalvin: Addressed in the latest push

blockchain/blockindex.go
kcalvinalvin: Addressed in the latest push

kcalvinalvin: Addressed in the latest push

blockchain/chain_test.go
kcalvinalvin: Addressed in the latest push

Roasbeef (approved):
LGTM 📣

PR #2155, opened 2024-04-02
blockchain, fullblocktests, workmath, testhelper: add InvalidateBlock() method to BlockChain

kcalvinalvin:
Depends on #2153

All of the code except for the last two commits are for refactoring code. The vast majority is just moving code from `fullblocktests` to `testhelper` so that the testing code for `InvalidateBlock()` can call the functions instead of having to copy over and duplicate functions.

The `InvalidateBlock()` method is operational and is tested but is not revealed through rpc in this PR.

### DIFF
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -1798,6 +1798,144 @@ func (b *BlockChain) LocateHeaders(locator BlockLocator, hashStop *chainhash.Has
 	return headers
 }
 
+// InvalidateBlock invalidates the requested block and all its descedents.  If a block
+// in the best chain is invalidated, the active chain tip will be the parent of the
+// invalidated block.
+//
+// This function is safe for concurrent access.
+func (b *BlockChain) InvalidateBlock(hash *chainhash.Hash) error {
+	b.chainLock.Lock()
+	defer b.chainLock.Unlock()
+
+	node := b.index.LookupNode(hash)
+	if node == nil {
+		// Return an error if the block doesn't exist.
+		return fmt.Errorf("Requested block hash of %s is not found "+
+			"and thus cannot be invalidated.", hash)
+	}
+	if node.height == 0 {
+		return fmt.Errorf("Requested block hash of %s is a at height 0 "+
+			"and is thus a genesis block and cannot be invalidated.",
+			node.hash)
+	}
+
+	// Nothing to do if the given block is already invalid.
+	if node.status.KnownInvalid() {
+		return nil
+	}
+
+	// Set the status of the block being invalidated.
+	b.index.SetStatusFlags(node, statusValidateFailed)
+	b.index.UnsetStatusFlags(node, statusValid)
+
+	// If the block we're invalidating is not on the best chain, we simply
+	// mark the block and all its descendants as invalid and return.
+	if !b.bestChain.Contains(node) {
+		// Grab all the tips excluding the active tip.
+		tips := b.index.InactiveTips(b.bestChain)
+		for _, tip := range tips {
+			// Continue if the given inactive tip is not a descendant of the block
+			// being invalidated.
+			if !tip.IsAncestor(node) {
+				continue
+			}
+
+			// Keep going back until we get to the block being invalidated.
+			// For each of the parent, we'll unset valid status and set invalid
+			// ancestor status.
+			for n := tip; n != nil && n != node; n = n.parent {
+				// Continue if it's already invalid.
+				if n.status.KnownInvalid() {
+					continue
+				}
+				b.index.SetStatusFlags(n, statusInvalidAncestor)
+				b.index.UnsetStatusFlags(n, statusValid)
+			}
+		}
+
+		if writeErr := b.index.flushToDB(); writeErr != nil {
+			return fmt.Errorf("Error flushing block index "+
+				"changes to disk: %v", writeErr)
+		}
+
+		// Return since the block being invalidated is on a side branch.
+		// Nothing else left to do.
+		return nil
+	}
+
+	// If we're here, it means a block from the active chain tip is getting
+	// invalidated.
+	//
+	// Grab all the nodes to detach from the active chain.
+	detachNodes := list.New()
+	for n := b.bestChain.Tip(); n != nil && n != node; n = n.parent {
+		// Continue if it's already invalid.
+		if n.status.KnownInvalid() {
+			continue
+		}
+
+		// Change the status of the block node.
+		b.index.SetStatusFlags(n, statusInvalidAncestor)
+		b.index.UnsetStatusFlags(n, statusValid)
+		detachNodes.PushBack(n)
+	}
+
+	// Push back the block node being invalidated.
+	detachNodes.PushBack(node)
+
+	// Reorg back to the parent of the block being invalidated.
+	// Nothing to attach so just pass an empty list.
+	err := b.reorganizeChain(detachNodes, list.New())
+	if err != nil {
+		return err
+	}
+
+	if writeErr := b.index.flushToDB(); writeErr != nil {
+		log.Warnf("Error flushing block index changes to disk: %v", writeErr)
+	}
+
+	// Grab all the tips.
+	tips := b.index.InactiveTips(b.bestChain)
+	tips = append(tips, b.bestChain.Tip())
+
+	// Here we'll check if the invalidation of the block in the active tip
+	// changes the status of the chain tips.  If a side branch now has more
+	// worksum, it becomes the active chain tip.
+	var bestTip *blockNode
+	for _, tip := range tips {
+		// Skip invalid tips as they cannot become the active tip.
+		if tip.status.KnownInvalid() {
+			continue
+		}
+
+		// If we have no best tips, then set this tip as the best tip.
+		if bestTip == nil {
+			bestTip = tip
+		} else {
+			// If there is an existing best tip, then compare it
+			// against the current tip.
+			if tip.workSum.Cmp(bestTip.workSum) == 1 {
+				bestTip = tip
+			}
+		}
+	}
+
+	// Return if the best tip is the current tip.
+	if bestTip == b.bestChain.Tip() {
+		return nil
+	}
+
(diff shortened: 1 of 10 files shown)


blockchain/testhelper/README.md
```diff
@@ -0,0 +1,16 @@
+testhelper
```
Roasbeef, line 1: We could also make this an `internal` package if we don't envision any external usage (hidden from default `godoc`, can't be imported outside the project).

blockchain/common_test.go
```diff
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/btcsuite/btcd/blockchain/testhelper"
```
Roasbeef, line 17: Commit references `invalidateblock` and `reconsiderblock`, but only the latter was added in this PR.

blockchain/chain.go
```diff
@@ -1798,6 +1798,142 @@ func (b *BlockChain) LocateHeaders(locator BlockLocator, hashStop *chainhash.Has
...
+				if n.status.KnownInvalid() {
+					continue
+				}
+				b.index.SetStatusFlags(n, statusInvalidAncestor)
+				b.index.UnsetStatusFlags(n, statusValid)
+			}
+		}
+
+		if writeErr := b.index.flushToDB(); writeErr != nil {
+			log.Warnf("Error flushing block index changes to disk: %v", writeErr)
```
Roasbeef, line 1857: Hmm, perhaps we should just hard fail here instead? As possible that we have unclean state in the db at this point, and accepting further blocks may cause cascading issues.

blockchain/chain.go
```diff
@@ -1798,6 +1798,142 @@ func (b *BlockChain) LocateHeaders(locator BlockLocator, hashStop *chainhash.Has
...
+		if n.status.KnownInvalid() {
+			continue
+		}
+
+		// Change the status of the block node.
+		b.index.SetStatusFlags(n, statusInvalidAncestor)
+		b.index.UnsetStatusFlags(n, statusValid)
+		detachNodes.PushBack(n)
+	}
+	// Push back the block node being invalidated.
```
Roasbeef, line 1881: Style nit: missing space above.

blockchain/chain_test.go
```diff
@@ -1311,3 +1313,381 @@ func TestIsAncestor(t *testing.T) {
...
 	}
 }
+
+func TestInvalidateBlock(t *testing.T) {
+	tests := []struct {
+		name     string
+		chainGen func() (*BlockChain, []*chainhash.Hash, func())
+	}{
+		{
+			name: "One branch, invalidate once",
```
Roasbeef, line 1323: Should start with a lower case character for uniformity.

blockchain/chain_test.go
```diff
@@ -1311,3 +1313,381 @@ func TestIsAncestor(t *testing.T) {
...
+			chainGen: func() (*BlockChain, []*chainhash.Hash, func()) {
+				source := rand.NewSource(time.Now().UnixNano())
+				rand := rand.New(source)
+
+				chain, params, tearDown := utxoCacheTestChain("TestInvalidateBlock-one-branch-invalidate-once")
```
Roasbeef, line 1328: I know we don't have the full linter active in this project, but can we wrap all this to 80 char columns to match most of the rest of the codebase?

blockchain/chain_test.go
```diff
@@ -1311,3 +1313,381 @@ func TestIsAncestor(t *testing.T) {
...
+					}
+
+					allSpends = append(allSpends, newSpendableOuts...)
+
+					var nextSpendsTmp []*testhelper.SpendableOut
+					for i := 0; i < len(allSpends); i++ {
+						randIdx := rand.Intn(len(allSpends))
+
+						spend := allSpends[randIdx]                                       // get
+						allSpends = append(allSpends[:randIdx], allSpends[randIdx+1:]...) // delete
```
Roasbeef, line 1364: https://pkg.go.dev/slices#Delete

blockchain/chain_test.go
```diff
@@ -1311,3 +1313,381 @@ func TestIsAncestor(t *testing.T) {
...
+				}
+
+				var allSpends []*testhelper.SpendableOut
+				nextBlock := b1
+				nextSpends := spendableOuts1
+
+				var invalidateHash *chainhash.Hash
+
+				// Create a chain with 11 blocks.
+				for b := 0; b < 10; b++ {
```
Roasbeef, line 1403: Can consolidate into a helper func based on the other instance above where we make the same 11 block chain.

blockchain/chain_test.go
```diff
@@ -1311,3 +1313,381 @@ func TestIsAncestor(t *testing.T) {
...
+					if b%10 == 0 {
+						// Commit the two base blocks to DB
+						if err := chain.FlushUtxoCache(FlushRequired); err != nil {
+							t.Fatalf("unexpected error while flushing cache: %v", err)
+						}
+					}
+				}
+
+				// Create a side chain with 7 blocks that builds on block 1.
+				var altSpends []*testhelper.SpendableOut
```
Roasbeef, line 1525: Same here re helper func for the 7 block chain.

blockchain/chain_test.go
```diff
@@ -1311,3 +1313,381 @@ func TestIsAncestor(t *testing.T) {
...
+				}
+
+				return chain, []*chainhash.Hash{invalidateHash}, tearDown
+			},
+		},
+	}
+
+	for _, test := range tests {
+		chain, invalidateHashes, tearDown := test.chainGen()
+		func() {
```
Roasbeef, line 1568: Why the closure func that gets immediately executed?

Roasbeef: We can also use `t.Run` here, then that way the test name gets printed during execution, and we can run just one of the sub-tests by name.

Roasbeef (changes requested):

blockchain/common_test.go
Roasbeef: I think for cases like the issue with the pruning bug w/ the utxo set cache, we actually want `reconsiderblock`, as `invalidateblock` marks all the dependents as invalid, so it won't attempt to _reprocess_ anything after the target block.

blockchain/chain_test.go
```diff
@@ -1311,3 +1313,381 @@ func TestIsAncestor(t *testing.T) {
...
+					nextBlock = newBlock
+
+					if newBlock.Height() == 5 {
+						invalidateHash = newBlock.Hash()
+					}
+
+					allSpends = append(allSpends, newSpendableOuts...)
+
+					var nextSpendsTmp []*testhelper.SpendableOut
+					for i := 0; i < len(allSpends); i++ {
```
Crypt-iQ, line 1360: A comment explaining what this is doing would be very helpful

blockchain/chain_test.go
```diff
@@ -1311,3 +1313,381 @@ func TestIsAncestor(t *testing.T) {
...
 }
+
+func TestInvalidateBlock(t *testing.T) {
+	tests := []struct {
+		name     string
+		chainGen func() (*BlockChain, []*chainhash.Hash, func())
+	}{
+		{
+			name: "One branch, invalidate once",
+			chainGen: func() (*BlockChain, []*chainhash.Hash, func()) {
```
Crypt-iQ, line 1324: think this should be functions below rather than explicitly in full form here

Crypt-iQ (commented):
looks good, think the tests could be cleaned up a little and have more comments

Roasbeef:
Dependent PR merged, needs a rebase!

blockchain/chain_test.go
kcalvinalvin: Could you clarify what you mean here?

blockchain/chain_test.go
kcalvinalvin: Addressed in the latest pushes.

blockchain/chain_test.go
kcalvinalvin: Addressed in the latest pushes.

blockchain/chain_test.go
kcalvinalvin: Addressed in the latest pushes.

blockchain/chain.go
kcalvinalvin: Addressed in the latest pushes.

blockchain/chain.go
kcalvinalvin: Addressed in the latest pushes.

blockchain/testhelper/README.md
kcalvinalvin: Addressed in the latest pushes. Made an internal package inside the package blockchain

blockchain/chain_test.go
kcalvinalvin: Will finish this promptly

blockchain/chain_test.go
Crypt-iQ: I just mean moving the `chainGen` function into a function outside of the test so you just call `chainGen:funcHere` for readability. Just my preference, but it's non blocking

Crypt-iQ (commented):
changes look good to me, only thing is whether `reconsiderblock` should be implemented instead

blockchain/common_test.go
kcalvinalvin: I'll add it in as another PR so that it doesn't add on to the needed reviews on this PR.

Was holding it off because the test for `reconsiderblock` utilizes `invalidateblock`

blockchain/chain_test.go
kcalvinalvin: Addressed in the latest push

blockchain/chain_test.go
kcalvinalvin: Was doing this so that even when the test fails, we still run the teardown

Crypt-iQ (approved):
LGTM 🏑

Roasbeef (approved):
LGTM 👑

PR #2167, opened 2024-04-14
blockchain: Test & Check Memory Usage

web3santa:
Use the newUtxoCache function to create a 1MB simulated newUtxoCache instance.

Call the totalMemoryUsage method to compare the returned result to the expected size of 1 MB.

If the result is different from the expected size, the test fails and outputs an error message.

This test code is responsible for ensuring that the memory usage of newUtxoCache matches the expected value.

### DIFF
--- a/blockchain/utxocache_test.go
+++ b/blockchain/utxocache_test.go
@@ -960,3 +960,20 @@ func TestInitConsistentState(t *testing.T) {
 			blocks[len(blocks)-1].Height())
 	}
 }
+
+func TestTotalMemoryUsage(t *testing.T) {
+	// Mocked utxoCache instance
+	s := newUtxoCache(nil, 1*1024*1024)
+
+	// Call the totalMemoryUsage method
+	result := s.totalMemoryUsage()
+
+	expectedSize := uint64(1 * 1024 * 1024)
+
+	resultShoulBeTrue := result == expectedSize
+
+	// Compare the expected result with the actual result
+	if resultShoulBeTrue {
+		t.Errorf("Expected memory usage %d, got %d", expectedSize, result)
+	}
+}


blockchain/utxocache_test.go
```diff
@@ -960,3 +960,20 @@ func TestInitConsistentState(t *testing.T) {
...
+
+	// Call the totalMemoryUsage method
+	result := s.totalMemoryUsage()
+
+	expectedSize := uint64(1 * 1024 * 1024)
+
+	resultShoulBeTrue := result == expectedSize
+
+	// Compare the expected result with the actual result
+	if resultShoulBeTrue {
```
SulaimanAminuBarkindo, line 976: It appears that there is a logical error in the condition `if resultShoulBeTrue`. This condition implies that the test should fail when the memory usage matches the expected size, which is the opposite of the intended behavior.

I recommend changing the condition to `if !resultShoulBeTrue` or ` if result != expectedSize`, indicating that the test should fail when the memory usage does not match the expected size

SulaimanAminuBarkindo: Additionally, it's important to note that [totalMemoryUsage](https://github.com/btcsuite/btcd/blob/master/blockchain/utxocache.go#L253) and [maxTotalMemoryUsage](https://github.com/btcsuite/btcd/blob/master/blockchain/utxocache.go#L215) represent different concepts. [maxTotalMemoryUsage](https://github.com/btcsuite/btcd/blob/master/blockchain/utxocache.go#L215) is the maximum memory usage allowed for the cache in normal circumstances, while [totalMemoryUsage](https://github.com/btcsuite/btcd/blob/master/blockchain/utxocache.go#L253) is the map size + the size that the utxo entries are taking up.

blockchain/utxocache_test.go
SulaimanAminuBarkindo, line 1: Thank you for adding a test to verify the memory usage of `newUtxoCache` using the `totalMemoryUsage` method.

SulaimanAminuBarkindo (changes requested):

kcalvinalvin:
This is essentially the same test as https://github.com/btcsuite/btcd/blob/d881c686e61db35e332fb0309178152dac589b03/blockchain/sizehelper_test.go#L30
since there aren't any additional entries being added.

Could you change the tests to include utxo entries so that the test isn't a duplicate?

Issue #2149, opened 2024-03-29
Support Sending RPC Requests to Bitcoin Core Over Unix Socket


robertmin1:
It would be great to support sending RPC requests to Bitcoin Core over a Unix socket. Bitcoin Core only supports RPC over HTTP. To work around this limitation, we can send HTTP requests over a Unix socket and then redirect them to the Bitcoin Core server using a program such as socat. This approach has been tested and works without any issues.

I wanted to get your input before working on a PR. This will only involve adding a `config field` for the `socket path` and adding a `Dial Context` to the `HTTP client`  + Explanation on how to use it 

Here is an example of http over unix socket: https://gist.github.com/teknoraver/5ffacb8757330715bcbcc90e6d46ac74#file-unixhttpc-go-L27-L31
Benefits: Security

robertmin1:
Closed by #2168

PR #2168, opened 2024-04-15
Enable the RpcClient to send requests to Bitcoin Core over a Unix socket.

robertmin1:
Enable the RpcClient to send  requests to Bitcoin Core over a Unix socket.
Issue #2149

### DIFF
--- a/rpcclient/examples/bitcoincoreunixsocket/README.md
+++ b/rpcclient/examples/bitcoincoreunixsocket/README.md
@@ -1,0 +1,41 @@
+Bitcoin Core HTTP POST Over Unix Socket Example
+==============================
+
+This example shows how to use the rpcclient package to connect to a Bitcoin
+Core RPC server using HTTP POST mode over a Unix Socket with TLS disabled 
+and gets the current block count.
+
+## Running the Example
+
+The first step is to use `go get` to download and install the rpcclient package:
+
+```bash
+$ go get github.com/btcsuite/btcd/rpcclient
+```
+
+Next, modify the `main.go` source to specify the correct RPC username and
+password for the RPC server:
+
+```Go
+	User: "yourrpcuser",
+	Pass: "yourrpcpass",
+```
+
+As Bitcoin Core supports only TCP/IP, we'll redirect RPC requests from the 
+Unix Socket to Bitcoin Core. For this example, we'll use the `socat` command:
+
+```bash
+$ socat -d UNIX-LISTEN:"my-unix-socket-path",fork TCP:"host-address"
+$ socat -d UNIX-LISTEN:/tmp/test.XXXX,fork TCP:localhost:8332
+```
+
+Finally, navigate to the example's directory and run it with:
+
+```bash
+$ cd $GOPATH/src/github.com/btcsuite/btcd/rpcclient/examples/bitcoincorehttp
+$ go run *.go
+```
+
+## License
+
+This example is licensed under the [copyfree](http://copyfree.org) ISC License.
--- a/rpcclient/examples/bitcoincoreunixsocket/main.go
+++ b/rpcclient/examples/bitcoincoreunixsocket/main.go
@@ -1,0 +1,39 @@
+// Copyright (c) 2014-2017 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package main
+
+import (
+	"log"
+
+	"github.com/btcsuite/btcd/rpcclient"
+)
+
+func main() {
+	// Connect to local bitcoin core RPC server using HTTP POST mode over a
+	// Unix Socket.
+	connCfg := &rpcclient.ConnConfig{
+		// For unix sockets, use unix:// + "your unix socket path".
+		Host:		"unix:///tmp/test.XXXX",		    
+		User:		"yourrpcuser",
+		Pass:		"yourrpcpass",
+		HTTPPostMode:	true, // Bitcoin core only supports HTTP POST mode.
+		DisableTLS:	true, // Bitcoin core does not provide TLS by default.
+	}
+
+	// Notice the notification parameter is nil since notifications are
+	// not supported in HTTP POST mode.
+	client, err := rpcclient.New(connCfg, nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer client.Shutdown()
+
+	// Get the current block count.
+	blockCount, err := client.GetBlockCount()
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.Printf("Block count: %d", blockCount)
+}
--- a/rpcclient/infrastructure.go
+++ b/rpcclient/infrastructure.go
@@ -7,6 +7,7 @@ package rpcclient
 import (
 	"bytes"
 	"container/list"
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/base64"
@@ -20,6 +21,8 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"strconv"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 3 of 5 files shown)


SulaimanAminuBarkindo (commented):
Thanks for this pull request! 
Overall, this pull request looks good and seems like a valuable addition to btcd. To further improve its usability and clarity, I have a couple of suggestions:

**Example Usage:** It would be helpful to add some example usage demonstrations in the [examples](https://github.com/btcsuite/btcd/tree/master/rpcclient/examples) folder for the rpcclient with this new functionality. This would make it easier for users to understand how to use the new feature in practice.

**Testing:** To further improve the clarity of the pull request, consider adding some steps to the PR description outlining how to test the newly implemented functionality with Unix sockets. This would be beneficial for both reviewers and future users.

robertmin1:
Great! I've included an example that should clarify how to test it further. The only tricky part is that a host needs to be passed to avoid raising errors. The value doesn't matter as long as it's valid. If the client has a default value, this shouldn't be a problem. Perhaps when setting the DialContext, we can hard code a value like `localhostunix.`

SulaimanAminuBarkindo (approved):
Thank you for providing the example files. 

LGTM

rpcclient/infrastructure.go
```diff
@@ -1338,6 +1343,13 @@ func newHTTPClient(config *ConnConfig) (*http.Client, error) {
 		},
 	}
 
+	// Configure the client to use a Unix socket if one is configured.
+	if config.UnixSocketPath != "" {
```
guggero, line 1347: Can't we do this in a bit of a cleaner way? I think we don't need another field and can just parse the host into a `net.Address` (example: https://github.com/lightningnetwork/lnd/blob/cd566eb0973240667d37823aaee2b5a61a012173/lncfg/address.go#L209). Then we can always set a dialer on the client:

```
	Transport: &http.Transport{
		DialContext: func(_ context.Context, _, _ string) (net.Conn, error) {
			return net.Dial(parsedAddr.Network(), parsedAddr.String())
		},
	},
```

robertmin1: Thanks, this is a much cleaner implementation, I'll work on it

robertmin1: Would it be more advantageous to duplicate a portion of that code instead of importing the function?
 It appears that the function performs additional actions and requires a defaultPort and tcpResolver.

robertmin1: Would it be preferable to duplicate the code instead of adding a new import for `"github.com/lightningnetwork/lnd/lncfg"`

guggero: Yes, definitely. Otherwise it's easy to get circular package dependencies. Also, I don't think we need Tor/onion support, so that part isn't needed.

robertmin1:
Updated the code. I didn't set a default port when calling `ParseAddressString` since the host doesn't seem to have one. 
Additionally, I used `ParseAddressString` to set a placeholder for Unix sockets in `handleSendPostMessage` in a cleaner way

rpcclient/infrastructure.go
```diff
@@ -1698,3 +1725,73 @@ func (c *Client) Send() error {
 
 	return nil
 }
+
+func ParseAddressString(strAddress string, defaultPort string) (net.Addr, error) {
```
SulaimanAminuBarkindo, line 1729: I believe this looks cleaner now, just a few nitpicks to address. please ensure to add comments for the functions added, as specified in the contribution guide linked [here](https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md?plain=1#L107)

robertmin1: Thanks, added the comments

SulaimanAminuBarkindo (approved):

rpcclient/examples/bitcoincoreunixsocket/README.md
```diff
@@ -0,0 +1,41 @@
...
+```Go
+	User: "yourrpcuser",
+	Pass: "yourrpcpass",
+```
+
+As Bitcoin Core supports only TCP/IP, we'll redirect RPC requests from the 
+Unix Socket to Bitcoin Core. For this example, we'll use the `socat` command:
+
+```
+	socat -d UNIX-LISTEN:"my-unix-socket-path",fork TCP:"host-address"
```
guggero, line 28: nit: use the `bash` syntax hint for the code block as well? And prefix the commands with a `$`?

rpcclient/examples/bitcoincoreunixsocket/main.go
```diff
@@ -0,0 +1,37 @@
...
+package main
+
+import (
+	"log"
+
+	"github.com/btcsuite/btcd/rpcclient"
+)
+
+func main() {
+	// Connect to local bitcoin core RPC server using HTTP POST mode over a Unix Socket.
```
guggero, line 14: nit: all code should be wrapped at 80 characters (while configuring your IDE/editor to count tabs as 8 spaces).

rpcclient/examples/bitcoincoreunixsocket/main.go
```diff
@@ -0,0 +1,37 @@
...
+	connCfg := &rpcclient.ConnConfig{
+		// For unix sockets, unix:// + "your unix socket path"
```
guggero, line 16: nit: all comments should be full sentences with punctuation, here and the inline comments below.

rpcclient/examples/bitcoincoreunixsocket/main.go
```diff
@@ -0,0 +1,37 @@
...
+		Host:         	"unix:///tmp/test.XXXX", 			    
+		User:         	"yourrpcuser",
+		Pass:         	"yourrpcpass",
+		HTTPPostMode: 	true, // Bitcoin core only supports HTTP POST mode
+		DisableTLS:   	true, // Bitcoin core does not provide TLS by default
+	}
+	// Notice the notification parameter is nil since notifications are
```
guggero, line 23: nit: add newline before comment block.

rpcclient/infrastructure.go
```diff
@@ -756,14 +759,31 @@ func (c *Client) handleSendPostMessage(jReq *jsonRequest) {
...
+		}
+		return
+	}
+
+	var url string
+	switch parsedAddr.Network(){
+	case "unix", "unixpacket":
+		// Using a placeholder URL because a non-empty URL is required
+		// the Unix domain socket is specified in the DialContext.
+		url = protocol + "://" + "unix"
```
guggero, line 782: nit: can just write `url = protocol + "://unix"`

rpcclient/infrastructure.go
```diff
@@ -756,14 +759,31 @@ func (c *Client) handleSendPostMessage(jReq *jsonRequest) {
...
+		jReq.responseChan <- &Response{
+			err: fmt.Errorf("failed to parse address %v", err),
+		}
+		return
+	}
+
+	var url string
+	switch parsedAddr.Network(){
+	case "unix", "unixpacket":
+		// Using a placeholder URL because a non-empty URL is required
```
guggero, line 780: nit: this sounds like they are two separate sentences, so should add punctuation/capitalization accordingly.

rpcclient/infrastructure.go
```diff
@@ -756,14 +759,31 @@ func (c *Client) handleSendPostMessage(jReq *jsonRequest) {
...
 	}
-	url := protocol + "://" + c.config.Host
 
 	var (
 		err, lastErr error
 		backoff      time.Duration
 		httpResponse *http.Response
 	)
 
+	parsedAddr, err := ParseAddressString(c.config.Host, "")
```
guggero, line 769: Since we're not using the default port (it's always empty), we should probably remove that argument and update the two new functions accordingly.

guggero (commented):
Thanks for the updates. Looks better now, but still got a couple of asks.

robertmin1:
Great! Thank you for reviewing. I've implemented the suggested changes.

robertmin1:
Any update on this? @guggero

jcvernaleo (approved):
OK

rpcclient/infrastructure.go
```diff
@@ -1698,3 +1725,80 @@ func (c *Client) Send() error {
...
+// ParseAddressString converts an address in string format to a net.Addr that is
+// compatible with btcd. UDP is not supported because btcd needs reliable
+// connections. We accept a custom function to resolve any TCP addresses so
+// that caller is able control exactly how resolution is performed.
+func ParseAddressString(strAddress string) (net.Addr, error) {
+    var parsedNetwork, parsedAddr string
+
+    // Addresses can either be in network://address:port format,
+    // network:address:port, address:port, or just port. We want to support
+    // all possible types.
```
lcmmhcc, line 1738: i guess u don't support the address like "https://getblock.io", which is supported in v0.24.2,  @robertmin1

PR #2173, opened 2024-04-22
release: add darwin-arm64 to target binaries

kcalvinalvin:
All currently released Apple computers have arm architecture and it's good to release something native for these computers.


jcvernaleo (approved):
100% agree.

PR #2174, opened 2024-04-22
chaincfg: update dnsseeds

kcalvinalvin:
To summarize:

Removed two seeds: `testnet-seed.bitcoin.schildbach.de`(testnet), `seed.bitcoinstats.com`(mainnet).

One from schildbach.de is completely unreachable and is also removed in Core. bitcoinstats.com seed is reachable but returns bad results. Documented at https://github.com/bitcoin/bitcoin/issues/29911

Added seeds from Sjors on all three networks.
Added seed from Wiz on mainnet.

I can separate the commits out into different PRs if desired

### DIFF
--- a/chaincfg/params.go
+++ b/chaincfg/params.go
@@ -59,6 +59,7 @@ var (
 	// DefaultSignetDNSSeeds is the list of seed nodes for the default
 	// (public, Taproot enabled) signet network.
 	DefaultSignetDNSSeeds = []DNSSeed{
+		{"seed.signet.bitcoin.sprovoost.nl", true},
 		{"178.128.221.177", false},
 		{"2a01:7c8:d005:390::5", false},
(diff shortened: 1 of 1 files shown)


guggero (approved):
tACK

`seed.bitcoin.sprovoost.nl` currently doesn't respond (or gives me a `SERVFAIL`), but I see that the same domain name is indeed present in Bitcoin Core (and is likely a temporary server failure, cc @sjors)

Sjors:
@guggero does my seed fail if you test it with this script? https://github.com/bitcoin-core/bitcoin-maintainer-tools/blob/main/check-dnsseeds.py

If not, which dig command did you use? The seed can return IPv4 and IPv6 records and can be reached via IPv4 and IPv6. Is one of there permutations not working, or all of them?

(I'm seeing some issues as well from my end, so looking into it)

guggero:
I just did a `dig A seed.bitcoin.sprovoost.nl`. That seems to work now, so thanks for the fix!

jcvernaleo (approved):
OK

PR #2178, opened 2024-04-26
Update standardness rules congruent to Bitcoin Core

ProofOfKeags:
This PR updates a couple of relay policy rules to bring them in line with what Bitcoin Core does. We do two things here,

1. Make CODESEPARATOR non-standard in non-segwit scripts
2. Make transactions <65 bytes (excluding the witness) non-standard

References:
- https://github.com/bitcoin/bitcoin/pull/26265
- https://github.com/bitcoin/bitcoin/pull/14039
- https://github.com/bitcoin/bitcoin/pull/11423


wire/msgtx.go
```diff
@@ -589,6 +596,12 @@ func (msg *MsgTx) btcDecode(r io.Reader, pver uint32, enc MessageEncoding,
 				sbuf = sbuf[len(txin.Witness[j]):]
 			}
 		}
+
+		// Check that if the witness flag is set that we actually have
```
Roasbeef, line 600: Nice, this is something that _is_ actually spelled out in the old BIP (144): 
> If the witness is empty, the old serialization format must be used.

https://github.com/bitcoin/bips/blob/master/bip-0144.mediawiki

wire/msgtx_test.go
```diff
@@ -849,6 +849,17 @@ func TestTxOutPointFromString(t *testing.T) {
 	}
 }
 
+// TestTxSuperfluousWitnessRecord ensures that btcd fails to parse a tx with
+// the witness marker flag set but without any actual witnesses.
+func TestTxSuperfluousWitnessRecord(t *testing.T) {
+	m := &MsgTx{}
+	rbuf := bytes.NewReader(multiWitnessFlagNoWitness)
+	err := m.BtcDecode(rbuf, ProtocolVersion, WitnessEncoding)
+	if err != errSuperfluousWitnessRecord {
```
Roasbeef, line 858: Can use `errors.Is` here.

txscript/engine.go
```diff
@@ -114,6 +114,10 @@ const (
 	// ScriptVerifyDiscourageUpgradeablePubkeyType defines if unknown
 	// public key versions (during tapscript execution) is non-standard.
 	ScriptVerifyDiscourageUpgradeablePubkeyType
+
+	// ScriptVerifyConstScriptCode fails non-segwit scripts if a signature
+	// match is found in the script code or if OP_CODESEPARATOR is used.
+	ScriptVerifyConstScriptCode
```
Roasbeef, line 120: +1 for aligning here, this wasn't in the original set of BIPs, but was added afterwards: https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2017-November/015292.html

 Being closer to `bitcoind` here policy wise may make it easier to keep up with w/e the outcome of the [Great Consensus Cleanup Revival](https://delvingbitcoin.org/t/great-consensus-cleanup-revival/710) is.

txscript/sigvalidate.go
```diff
@@ -20,9 +20,14 @@ import (
...
 type signatureVerifier interface {
-	// Verify returns true if the signature verifier context deems the
+	// Verify returns whether or not the signature verifier context deems the
 	// signature to be valid for the given context.
-	Verify() bool
+	Verify() verifyResult
+}
+
+type verifyResult struct {
+	valid bool
```
Roasbeef, line 29: `sigValid`?

txscript/sigvalidate.go
```diff
@@ -20,9 +20,14 @@ import (
...
+	match bool
```
Roasbeef, line 30: `sigMatch`?

txscript/sigvalidate.go
```diff
@@ -20,9 +20,14 @@ import (
 // pre-segwit, segwit v0, segwit v1 (taproot key spend validation), and the
 // base tapscript verification.
 type signatureVerifier interface {
-	// Verify returns true if the signature verifier context deems the
+	// Verify returns whether or not the signature verifier context deems the
 	// signature to be valid for the given context.
-	Verify() bool
+	Verify() verifyResult
```
Roasbeef, line 25: FWIW this only exists pre-segwit (doesn't apply to segwit v0 or v1). I wonder if we should implement this as a new super set interface? Downside is that it would end up bubbling up into the VM code (forced type assertion). 

Will give it a think over.

Roasbeef (changes requested):
Nice PR, there're a few lingering policy rules bitcoind applies that we don't, so nice to be able to periodically bridge the gap. Hopefully one day this'll all be in a nice BIP to make it easier to track relevant policy that full node implementations should consider. 

Left a few comments in line

txscript/opcode.go
```diff
@@ -1953,6 +1953,12 @@ func opcodeCodeSeparator(op *opcode, data []byte, vm *Engine) error {
 
 	if vm.taprootCtx != nil {
 		vm.taprootCtx.codeSepPos = uint32(vm.tokenizer.OpcodePosition())
+	} else if !vm.isWitnessVersionActive(0) &&
```
Roasbeef, line 1956: I think we can tighten this policy check a bit. The `bitcoind` version disables for all segwit versions, while this would only disable for segwit v0. Instead we can check that `vm.witnessProgram == nil`.

Roasbeef (changes requested):

Roasbeef (approved):
LGTM 🧆

Zoiediez:
> Nice PR, there're a few lingering policy rules bitcoind applies that we don't, so nice to be able to periodically bridge the gap. Hopefully one day this'll all be in a nice BIP to make it easier to track relevant policy that full node implementations should consider. 
> 
> 
> 
> Left a few comments in line

I would love transfer every thing to you guys once its settle thanks for your help

PR #2182, opened 2024-05-01
blockchain, main: add and fix logs

kcalvinalvin:
The trailing `: ` is removed from the utxocache log. It just looks better.

If the node is pruned, we now log what MiB it's pruned to. Logging alerts the user that the node is pruned and it helps for debugging purposes as well since we can now see from the logging whether or not the node is pruned.


saubyk (approved):
LGTM

Roasbeef (approved):
LGTM ⛺️

PR #2181, opened 2024-04-30
blockchain: Add ReconsiderBlock()

kcalvinalvin:
Adds `ReconsiderBlock()` on blockchain that will clear up any block validation statuses and re-validate the block. If the block is part of a longer chain, then the chain will reorganize to that chain.

Depends on #2155

PR #2196, opened 2024-06-07
blockchain: Add ReconsiderBlock()

Crypt-iQ:
Adds ReconsiderBlock() on blockchain that will clear up any block validation statuses and re-validate the block. If the block is part of a longer chain, then the chain will reorganize to that chain.

Replaces #2181

### DIFF
--- a/blockchain/chain.go
+++ b/blockchain/chain.go
@@ -1950,6 +1950,96 @@ func (b *BlockChain) InvalidateBlock(hash *chainhash.Hash) error {
 	return err
 }
 
+// ReconsiderBlock reconsiders the validity of the block with the given hash.
+//
+// This function is safe for concurrent access.
+func (b *BlockChain) ReconsiderBlock(hash *chainhash.Hash) error {
+	b.chainLock.Lock()
+	defer b.chainLock.Unlock()
+
+	log.Infof("Reconsidering block_hash=%v", hash[:])
+
+	reconsiderNode := b.index.LookupNode(hash)
+	if reconsiderNode == nil {
+		// Return an error if the block doesn't exist.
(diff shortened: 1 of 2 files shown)


blockchain/chain.go
```diff
@@ -1033,96 +1131,12 @@ func (b *BlockChain) reorganizeChain(detachNodes, attachNodes *list.List) error
...
 		b.index.SetStatusFlags(n, statusValid)
-
-		newBest = n
-	}
-
-	// Flush the utxo cache for the block disconnect below.  The disconnect
-	// code assumes that it's directly modifying the database so the cache
-	// will be left in an inconsistent state.  It needs to be flushed beforehand
-	// in order for that to not happen.
-	err := b.db.Update(func(dbTx database.Tx) error {
```
Crypt-iQ, line 1047: Outstanding question is whether we need to do this still?

Crypt-iQ: @kcalvinalvin

kcalvinalvin: It wasn't needed [here](https://github.com/btcsuite/btcd/commit/a254998bc5e348a227cdb454ffca0e9e3857c4a1) (the PR for when it was added) since the flush on `disconnectBlock` would always ensure that the right state would be saved to disk.

No need for it to be here and it's ok to remove it.

kcalvinalvin:
Verified that this PR is just a cleanup for the commits:

`c14546b369dde37d868985f1112468b70d52967f`
`7c04bd86ac54fbe404e1085e569a35d83ea370e2`

from #2181

Looks fine for me

Roasbeef (approved):
LGTM 🌴

PR #2198, opened 2024-06-14
psbt: add new fields for MuSig2

guggero:
Implements the new input and output fields to the PSBT library according to [BIP-0373](https://github.com/bitcoin/bips/pull/1540).


guggero:
The [BIP-0373](https://github.com/bitcoin/bips/pull/1540) PR was merged a while ago. I also rebased on latest master that has the Golang module fixes and addressed all TODOs. This is ready for review IMO.

psbt/musig2.go
```diff
@@ -0,0 +1,286 @@
...
+	)
+}
+
+// MuSig2PartialSig represents a partial signature provided by a participant in
+// a MuSig2 signing session.
+type MuSig2PartialSig struct {
+	// PubKey is the public key of the participant providing this nonce.
+	PubKey *btcec.PublicKey
+
+	// AggregateKey is the plain (non-tweaked) aggregate public key the
```
TechLateef, line 270: nit: These comments still refer to a nonce, but this struct represents a MuSig2 partial signature. Could we update the wording accordingly?

psbt/musig2.go
```diff
@@ -0,0 +1,286 @@
...
+func ReadMuSig2PartialSig(keyData, value []byte) (*MuSig2PartialSig, error) {
+	const pubKeyLen = btcec.PubKeyBytesLenCompressed
+	const minLength = 2 * pubKeyLen
+	const maxLength = minLength + sha256.Size
+
+	if len(keyData) != minLength && len(keyData) != maxLength {
+		return nil, ErrInvalidKeyData
+	}
+
+	if len(value) != 32 {
```
TechLateef: nit: Would it make sense to add a `musig2.PartialSigSize` constant alongside `PubNonceSize` and `SecNonceSize` instead of using the hardcoded 32?

psbt/musig2.go
```diff
@@ -0,0 +1,286 @@
...
+		copy(
+			value[idx*btcec.PubKeyBytesLenCompressed:],
+			key.SerializeCompressed(),
+		)
+	}
+
+	return serializeKVPairWithType(w, typ, participants.KeyData(), value)
+}
+
+// MuSig2PubNonce represents a public nonce provided by a participant in a
```
TechLateef, line 143: suggestion: `ReadMuSig2PubNonce` and `ReadMuSig2PartialSig` validate that the optional `TapLeafHash` is either omitted or `sha256.Size` bytes, but `KeyData()` currently serializes whatever length is present. Since I couldn't find any earlier validation, would it make sense to validate `TapLeafHash` during serialization (or in `KeyData()`) as well? Otherwise an invalid in-memory struct (for example, a 10-byte `TapLeafHash`) could be serialized into an unparseable PSBT key-value pair.

psbt/utils.go
```diff
@@ -478,3 +478,36 @@ func FindLeafScript(pInput *PInput,
...
+	for idx, txIn := range packet.UnsignedTx.TxIn {
+		in := packet.Inputs[idx]
+
+		// Skip any input that has no UTXO.
+		if in.WitnessUtxo == nil && in.NonWitnessUtxo == nil {
+			continue
+		}
+
+		if in.NonWitnessUtxo != nil {
+			prevIndex := txIn.PreviousOutPoint.Index
```
TechLateef, line 590: suggestion: Should we validate that `prevIndex < uint32(len(in.NonWitnessUtxo.TxOut))` before indexing `TxOut`? 

`SumUtxoValues` already performs a similar bounds check, and without one here a malformed PSBT could trigger an index-out-of-range panic.

psbt/finalizer.go
```diff
@@ -595,3 +633,570 @@ func finalizeTaprootInput(p *Packet, inIndex int) error {
...
+	sig := appendSighashType(schnorrSig.Serialize(), pInput.SighashType)
+	return writeWitness(sig, leaf.Script, leaf.ControlBlock)
+}
+
+// musig2InputReady reports whether a taproot input has a complete and
+// internally consistent set of MuSig2 fields ready for finalization: the
+// number of nonces matches the number of partial signatures, and every
+// nonce and partial signature references the same (AggregateKey,
+// TapLeafHash) pair.
+func musig2InputReady(pInput *PInput) bool {
```
TechLateef, line 1307: nit: `musig2InputReady` assumes `MuSig2PartialSigs` is non-empty and will panic when indexing the first element if it's ever called directly with both MuSig2 slices empty. Since the helper already validates the relationship between the slices, would it make sense to also guard against the empty case so it's self-contained?

psbt/finalizer.go
```diff
@@ -595,3 +633,570 @@ func finalizeTaprootInput(p *Packet, inIndex int) error {
...
+//     supported; return an explicit error rather than silently producing a
+//     wrong signature.
+func selectMuSig2KeyAggTweaks(p *Packet, pInput *PInput,
+	partialSigAggregate *btcec.PublicKey, keys []*btcec.PublicKey,
+	sigHashMsg [32]byte) ([]musig2.KeyAggOption, []musig2.CombineOption,
+	error) {
+
+	var bareAggregate *btcec.PublicKey
+	if len(pInput.MuSig2Participants) > 0 {
+		bareAggregate = pInput.MuSig2Participants[0].AggregateKey
```
TechLateef: Suggestion: The BIP allows `MuSig2` fields to be associated with different aggregate public keys, and `PSBT_IN_MUSIG2_PARTICIPANT_PUBKEYS` is keyed by the aggregate pubkey.
 Would it make sense to locate the participant record matching `partialSigAggregate` instead of assuming `MuSig2Participants[0]` corresponds to the partial signatures? Otherwise, if multiple participant records are present, we could select the wrong aggregate key.

psbt/musig2.go
guggero: Good catch, fixed!

psbt/musig2.go
guggero: Hmm, `PartialSigSize` would be the wrong choice. Because in the context of the `musig2` package, a partial sig is both `S` and the combined `R`. But in the `psbt` context, it's just the scalar `S`.
But since I had to export something else from the package, I also exported `schnorr.ScalarSize` and used that here instead.

psbt/musig2.go
guggero: Very good point! Added a `Verify` and covering unit tests.

psbt/utils.go
guggero: Yes, makes sense, added.

psbt/finalizer.go
guggero: Added a check, thanks.

psbt/finalizer.go
guggero: Yes, you're correct. Fixed and added tests for.

guggero:
@TechLateef thank you very much for the thorough review! I've addressed all your comments.

psbt/musig2.go
TechLateef: okay great that make sense

psbt/finalizer.go
```diff
@@ -606,3 +644,612 @@ func finalizeTaprootInput(p *Packet, inIndex int) error {
...
+// combineMuSig2Sig aggregates the keys and nonces of the given signing set and
+// combines its partial signatures into a single BIP-340 Schnorr signature. The
+// keyAggOpts and combineOpts must describe the same tweak chain: tweaks applied
+// during key aggregation must match the tweaks accumulated by the combine
+// option.
+func combineMuSig2Sig(sigHashMsg [32]byte, set *muSig2SigningSet,
+	keyAggOpts []musig2.KeyAggOption,
+	combineOpts []musig2.CombineOption) (*schnorr.Signature, error) {
+
+	aggKey, _, _, err := musig2.AggregateKeys(set.keys, true, keyAggOpts...)
```
sputn1ck: **[P1] Preserve participant aggregation order**

The signing set is reconstructed from nonce-map order, then `AggregateKeys` is always called with sorting enabled. BIP 373 stores participants in the required `KeyAgg` order and BIP 327 makes sorting optional. I reproduced this with a valid explicitly unsorted session; `MaybeFinalizeAll` rejects it. Please select the matching participant record, map nonces and signatures into its `Keys` order, and aggregate with `sort=false`.

sputn1ck (commented):
BIP-373 implementation review. Inline findings below.

psbt/finalizer.go
```diff
@@ -606,3 +644,612 @@ func finalizeTaprootInput(p *Packet, inIndex int) error {
...
+func musig2BIP32DerivedTweaks(p *Packet, pInput *PInput,
+	bareAggregate *btcec.PublicKey, keys []*btcec.PublicKey,
+	sigHashMsg [32]byte) ([]musig2.KeyAggOption, []musig2.CombineOption,
+	error) {
+
+	xpub, err := findAggregateXpub(p, bareAggregate)
+	if err != nil {
+		return nil, nil, err
+	}
+	if xpub == nil {
```
sputn1ck: **[P1] Implement the BIP 328 fallback**

Derived aggregate finalization requires `PSBT_GLOBAL_XPUB` and explicitly rejects its absence, but BIP 373 says to assume BIP 328's fixed synthetic chain code when no matching global xpub exists. The official case 4c partial signatures therefore cannot be aggregated after removing the preaggregated key signature.

psbt/finalizer.go
```diff
@@ -606,3 +644,612 @@ func finalizeTaprootInput(p *Packet, inIndex int) error {
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error calculating tapscript signature "+
+			"hash: %w", err)
+	}
+
+	var sigHashMsg [32]byte
+	copy(sigHashMsg[:], sigHash)
+
+	// No taproot tweak: the aggregate key is the key in the script and is
```
sputn1ck: **[P1] Apply derivation tweaks to script spends**

The tapscript finalizer always combines with no tweaks. BIP 373 permits the key inside a leaf to be derived from a parent MuSig2 aggregate, in which case the BIP32 derivation tweaks must also be applied here. Otherwise this method installs an invalid witness.

psbt/finalizer.go
```diff
@@ -606,3 +644,612 @@ func finalizeTaprootInput(p *Packet, inIndex int) error {
...
+// the parallel slices are aligned regardless of the order the fields appear on
+// the input. The aggregate key and tap leaf hash are taken from the partial
+// signatures and verified to agree with the nonces.
+func extractMuSig2SigningSet(pInput *PInput) (*muSig2SigningSet, error) {
+	numSigs := len(pInput.MuSig2PartialSigs)
+	if numSigs == 0 {
+		return nil, fmt.Errorf("no MuSig2 partial signatures on input")
+	}
+
+	if len(pInput.MuSig2PubNonces) != numSigs {
```
sputn1ck: **[P1] Group independent signing contexts**

This requires the total nonce and signature counts to match and later requires every record to use one aggregate and leaf. BIP 373 keys these records so several aggregates or spend paths can coexist. A complete key-path session plus one nonce for another leaf becomes `ErrNotFinalizable`. Please group by aggregate key and tapleaf hash, then select a complete context.

psbt/utils.go
```diff
@@ -531,3 +531,44 @@ func FindLeafScript(pInput *PInput,
...
+
+// PrevOutputFetcher returns a txscript.PrevOutFetcher built from the UTXO
+// information in a PSBT packet.
+func PrevOutputFetcher(packet *Packet) *txscript.MultiPrevOutFetcher {
+	fetcher := txscript.NewMultiPrevOutFetcher(nil)
+	for idx, txIn := range packet.UnsignedTx.TxIn {
+		in := packet.Inputs[idx]
+
+		// Skip any input that has no UTXO.
+		if in.WitnessUtxo == nil && in.NonWitnessUtxo == nil {
```
sputn1ck, line 585: **[P1] Return an error for missing prevouts**

Missing or invalid prevouts are silently skipped. `NewTxSigHashes` subsequently dereferences every fetched prevout for Taproot. I reproduced a panic by finalizing input 0 of a two-input packet whose second input lacked UTXO data. Please validate completeness before constructing the sighash cache.

psbt/finalizer.go
```diff
@@ -606,3 +644,612 @@ func finalizeTaprootInput(p *Packet, inIndex int) error {
...
+	// to re-derive the final nonce from the participants' public nonces.
+	nonceJ, _, err := musig2.ComputeSigningNonce(
+		aggregateNonce, aggKey.FinalKey, sigHashMsg,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error computing signing nonce: %w", err)
+	}
+	nonceJ.ToAffine()
+
+	return musig2.CombineSigs(
```
sputn1ck: **[P2] Verify the combined signature**

`CombineSigs` is returned without verifying the BIP340 result against the actual output or leaf key, after which finalization discards all MuSig2 data. Mutating one partial scalar makes `MaybeFinalizeAll` report success while `txscript` rejects the witness. Please verify against the real spend key and sighash before destructive finalization.

psbt/updater.go
```diff
@@ -375,3 +375,115 @@ func (u *Updater) AddOutWitnessScript(witnessScript []byte,
...
+// AddInMuSig2Participants adds a PSBT_IN_MUSIG2_PARTICIPANT_PUBKEYS field
+// to the input at index inIndex. Returns ErrDuplicateKey if a participants
+// record with the same aggregate key is already present.
+func (u *Updater) AddInMuSig2Participants(inIndex int,
+	participants *MuSig2Participants) error {
+
+	if inIndex < 0 || inIndex >= len(u.Upsbt.Inputs) {
+		return ErrInvalidPsbtFormat
+	}
+	if participants == nil || participants.AggregateKey == nil {
```
sputn1ck: **[P2] Validate before updater mutation**

The four new updater methods append without calling the records' existing `Validate` methods, while `SanityCheck` does not inspect them. Empty participant lists, malformed leaf hashes, and a missing partial-signature scalar all return success; `SignMuSig2` can therefore report `SignSuccesful` for data that serialization rejects.

psbt/partial_input.go
```diff
@@ -590,6 +648,52 @@ func (pi *PInput) serialize(w io.Writer) error {
 				return err
 			}
 		}
+
+		slices.SortFunc(
```
sputn1ck, line 654: **[P2] Preflight serialization records**

Sorting dereferences exported slice elements before checking for nil, so a nil MuSig2 record panics. Duplicate records serialize successfully even though this library rejects the resulting PSBT when reparsing it. Please validate nils and duplicate `KeyData` before sorting or writing; output participants need the same treatment.

psbt/finalizer.go
guggero: Fixed and added tests for!

psbt/finalizer.go
guggero: Addressed and unit tested.

psbt/finalizer.go
guggero: Fixed.

psbt/finalizer.go
guggero: Makes sense, fixed.

TechLateef (approved):
LGTM!

PR #2205, opened 2024-06-25
rpcclient: update error str to match both versions

yyforyongyu:
This commit updates the error str to match the same error returned from `btcd` for both pre-0.24.2 and post-0.24.2.

### DIFF
--- a/rpcclient/errors.go
+++ b/rpcclient/errors.go
@@ -411,7 +411,10 @@ var BtcdErrMap = map[string]error{
 	"transaction already exists in blockchain":    ErrTxAlreadyConfirmed,
 
 	// A transaction in the mempool.
-	"already have transaction in mempool": ErrTxAlreadyInMempool,
+	//
+	// NOTE: For btcd v0.24.2 and beyond, the error message is "already
+	// have transaction in mempool".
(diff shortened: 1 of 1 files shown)


guggero (approved):

jcvernaleo (approved):
OK

rpcclient/errors.go
```diff
@@ -411,7 +411,10 @@ var BtcdErrMap = map[string]error{
 	"transaction already exists in blockchain":    ErrTxAlreadyConfirmed,
 
 	// A transaction in the mempool.
-	"already have transaction in mempool": ErrTxAlreadyInMempool,
+	//
+	// NOTE: For btcd v0.24.2 and beyond, the error message is "already
```
hieblmi, line 415: shouldn't this be `For btcd versions prior to v0.24.2, the error message is "already have transaction in mempool"`?

yyforyongyu: In `v0.24` the error msg is `already have transaction`, in `0.24.2` the error msg is changed to `already have transaction in mempool`, otherwise the match won't fail since `already have transaction` is a subset of `already have transaction in mempool`.

hieblmi: Thanks now it makes sense.

PR #2206, opened 2024-06-25
psbt: decode keytype as compact size

### LINKED ISSUE #2199: Issue/Bug with PSBT deserialization
Hello, 

BTCD _does not fail as it should_ when trying to deserialize the following bytes:

`70736274ff01001c000000000002000000000000000000000000736210ff01000001010010ff70ff01001c00000000000000000000000000000000000000000000`

For the above input, Bitcoin-core _fails_ while trying to deserialize and throws the error `ReadCompactSize(): size too large`.

This issue partly seems to be due to incorrect implementation of [bip-174](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki). Bip-174 says that `keytype` is a `compact size int` and from what I see btcd simply treats `keytype` [as a single byte](https://github.com/btcsuite/btcd/blob/master/btcutil/psbt/utils.go#L253-L258).

The test case given below can be used to reproduce this issue:

```golang

func TestKeyType(t *testing.T) {
	k, _ := hex.DecodeString("70736274ff01001c000000000002000000000000000000000000736210ff01000001010010ff70ff01001c00000000000000000000000000000000000000000000")
	if _, err := NewFromRawBytes(bytes.NewReader(k), false); err == nil {
		require.Error(t, err, "Err should not be nil")
	}
}
```

cc: @brunoerg

guggero:
Fixes #2199.

Previous to this fix the keytype was only interpreted as a single byte, even though BIP-0174 states it is to be parsed as a CompactSize/VarInt.


btcutil/psbt/psbt.go
```diff
@@ -37,6 +37,9 @@ const MaxPsbtValueLength = 4000000
 // deserialize from the wire. Anything more will return ErrInvalidKeyData.
 const MaxPsbtKeyLength = 10000
 
+// MaxPsbtKeyValue is the maximum value of a key type in a PSBT.
+const MaxPsbtKeyValue = 0x02000000
```
brunoerg, line 41: Just to clarify, I think the BIP doesn't specify a maximum value of a key type in a PSBT. Bitcoin Core uses this maximum value for every CompactSize occurrence.

guggero: Correct. Without this check the value mentioned in the issue wouldn't have caused the test to fail. So I assumed we want to mimic `bitcoind`'s behavior here.
Do you think that warrants a sentence in the constant's comment?

brunoerg: Yes, it would be great.

guggero: Done.

Roasbeef (approved):
LGTM ⚱️

brunoerg (approved):
ACK 4bff77856409dafc4d8caca6f6e82b7084cdb7da

PR #2208, opened 2024-07-01
ffldb: close block files

kcalvinalvin:
On methods readBlock and readBlockRegion, the opened files were never closed which resulted in errors when attempting to delete these files on pruning on windows.  Properly closing the files avoids this error.

Doing a full block sync on windows to make sure. Definitely seems to be the fix as I'm not having any issues so far (few hours into ibd).

### DIFF
--- a/database/ffldb/blockio.go
+++ b/database/ffldb/blockio.go
@@ -281,17 +281,7 @@ func (s *blockStore) openFile(fileNum uint32) (*lockableFile, error) {
 	lruList := s.openBlocksLRU
 	if lruList.Len() >= maxOpenFiles {
 		lruFileNum := lruList.Remove(lruList.Back()).(uint32)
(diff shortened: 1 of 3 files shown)


Roasbeef:
> he opened files were never closed which resulted in errors when attempting to delete these files on pruning on windows.

I've had similar issues with Windows, basically that you always need to close files before doing operations such as rename or delete.

kcalvinalvin:
> > he opened files were never closed which resulted in errors when attempting to delete these files on pruning on windows.
> 
> I've had similar issues with Windows, basically that you always need to close files before doing operations such as rename or delete.

Yeah that seems to have been the issue. No troubles syncing with pruning turned on

saubyk:
cc: @guggero for review

Roasbeef (approved):
LGTM 🎎

PR #2209, opened 2024-07-02
rpcclient: update `ErrTxAlreadyConfirmed` to match both versions

yyforyongyu:
This commit updates the error str to match the same error `ErrTxAlreadyConfirmed` returned from `btcd` for both pre-0.24.2 and post-0.24.2.

### DIFF
--- a/rpcclient/errors.go
+++ b/rpcclient/errors.go
@@ -406,9 +406,14 @@ var BtcdErrMap = map[string]error{
 	// A transaction with too many sigops.
 	"sigop cost is too hight": ErrTooManySigOps,
 
-	// A transaction already in the blockchain.
+	// A transaction already in the database.
 	"database contains entry for spent tx output": ErrTxAlreadyKnown,
-	"transaction already exists in blockchain":    ErrTxAlreadyConfirmed,
+
+	// A transaction already in the blockchain.
+	//
+	// NOTE: For btcd v0.24.2 and beyond, the error message is "transaction
+	// already exists in blockchain".
+	"transaction already exists": ErrTxAlreadyConfirmed,
 
 	// A transaction in the mempool.
 	//


leafl3d (approved):
looks good

Roasbeef (changes requested):
I think the past few patches, regressions, and bugs in this area have shown that this approach is too fragile. The initial culprit [was this commit](https://github.com/btcsuite/btcd/commit/ef54c49df443815d50765e8c4f31a87944d950a6#diff-37e0f6f3ccf82f3fcc49edc58fdace00678d70956ff412c13c64b764dc58a6f4L1353) that "improved" some of the errors to give additional detail. This in turn broke the string based error matching logic. 

Today `bitcoind` will send 3 primary JSON-RPC errors due to error cases encountered during the processing of `sendrawtransaction`: https://github.com/bitcoin/bitcoin/blob/173ab0ccf2164dd8cd3d486559dffda0d4a1a813/src/rpc/protocol.h#L47-L49

`btcd` on the other hand uses the existing error codes from the `reject` message (which `bitcoind` removed quite sometime ago unfortunately): https://github.com/btcsuite/btcd/blob/master/wire/msgreject.go#L21-L26. 

To make our handling of `sendrawtransaction` RPC errors more robust, what if we just switched nearly entirely to: 
  1. Detecting which backend is being used (we know that `btcd` operates over `ws` as an example). 
  2. Depending on the backend, check the set of distinct errors. 

IIUC, this forces our error handling to be much coarser. A side effect of this may be that we're able able to properly detect subtle cases like rejecting due to being invalid vs not following the RBF rules. To fill this gap, elsewhere we already use `testmempoolaccept`, which has a [more structured error output](https://developer.bitcoin.org/reference/rpc/testmempoolaccept.html) it still isn't great though, as if any of these error strings change, then we enter unexpected behavior territory.

ziggie1984:
maybe we funnel this feedback in regards of the error strings vs concrete error types back to bitcoind so that they can adopt or at least think about other ways which might make the error handling easier for third party applications using their API ?

PR #2211, opened 2024-07-10
remove `error` return value for `SignCompact`

zhiqiangxu:
There's no need for `error` since it's always `nil`.


guggero (approved):

jcvernaleo (approved):
OK

torkelrogstad:
It's unfortunate that this was included in a patch dependency bump, as it is in fact a breaking change. It causes the latest version of the btcec and lnd Go libraries to not compile together.

PR #2212, opened 2024-07-13
[WiP] bip324 v2 p2p protocol

lnliz:
#### Add support for bip324 v2 p2p protocol

I worked on bip324 code for the [bitcoin-bip324-proxy project](https://github.com/lnliz/bitcoin-bip324-proxy) and created the v2 p2p transport code so it can be used as a library for other bitcoin projects written in Go. This is POC showing that btcd can use the bip324_transport package to establish v2 connections.
Please disregard the noise in go.mod/go.sum file - I think this needs refactor of the bip324 crypto and transport package in a new package or repo so that proxy dependencies are not pulled in.

This is not a final PR, only to start a conversation and to demo functionality. There are many open questions:
- is btcd project interested in adding bip324 v2 protocol support?
- if yes, what's a good way forward?
- how are external dependencies normally handled? 
- I added the v2 code in the code where I thought it would make sense but could use help in fixing the approach. 
- the bip324_transport code is not very well tested yet, some help with reviewing and improvement suggestions are welcome.
- still missing: when trying to connect to a v1 node this will fail. bitcoin core then marks the node for reconnect with v1 - this needs something similar.

tested with:
```
go run . --blocksonly --noonion --maxpeers=10 --nodnsseed --debuglevel=debug --connect=aa.bb.cc.dd:8333
```
where "aa.bb.cc.dd:8333" is a bitcoin core node that supports v2 (-> bitcoin core v27.0 and up)
or run without "connect" parameter and btcd will try the list of peers it already has


saubyk:
Hi @lnliz thanks for the pr. This is an important area of work for sure

Just to let you know, this project is already being worked on by @Crypt-iQ (should have the draft pr up soon)
It would be good to reconcile and also get your reviews on it as well

@Crypt-iQ can also post some initial review comments on this pr, but I would urge you to hold-off on this for a bit

saubyk:
Hi @lnliz I believe https://github.com/btcsuite/btcd/pull/2219 is a precursor pr for BIP324

PR #2215, opened 2024-07-17
blockchain: increase periodic flush interval

kcalvinalvin:
The periodic flush intervals were set to 5 minutes previously which on average means that the node will flush every block when the node is caught up to the tip.

This put an overhead on the node after the node is caught up beyond the last checkpointed block as it'll start calling the periodic flush function.  So while the node is still performing ibd, it'll be flushing every 5 minutes, hurting performance.

Changing this value to 48 hours doesn't hurt the node too much as the average worst case would be 288 blocks to reindex which is quick. For a decent computer, 48 hours should be enough time to sync all the blocks beyond the last checkpointed block as well.  If not, 1-2 flushes don't hurt as much as flushing every 5 minutes.


Roasbeef:
IIUC, with this most syncs wouldn't ever sync the utxo set cache until the very end (at least on a fast computer, possibly syncing over LAN). Let's say it crashes right before sync finishes, well before the 48 hour mark, how long would it take to reconstruct/reconcile the UTXO set cache on start up?

PR #2216, opened 2024-07-18
txscript: add new ScriptTemplate DSL for writing Scripts

Roasbeef:
In this commit, we add a new function, `ScriptTemplate` to make the process of making custom Bitcoin scripts a bit less verbose.

ScriptTemplate processes a script template with parameters and returns the corresponding script bytes. This functions allows Bitcoin scripts to be created using a DSL-like syntax, based on Go's templating system.

An example of a simple p2pkh template would be:

   `OP_DUP OP_HASH160 0x14e8948c7afa71b6e6fad621256474b5959e0305 OP_EQUALVERIFY OP_CHECKSIG`

Strings that have the `0x` prefix are assumed to byte strings to be pushed ontop of the stack. Integers can be passed as normal. If a value can't be parsed as an integer, then it's assume that it's a byte slice without the 0x prefix.

Normal go template operations can be used as well. The params argument houses paramters to pass into the script, for example a local variable storing a computed public key.

### DIFF
--- a/txscript/template.go
+++ b/txscript/template.go
@@ -1,0 +1,236 @@
+package txscript
+
+import (
+	"bufio"
+	"bytes"
+	"encoding/hex"
+	"fmt"
+	"html/template"
+	"strconv"
+	"strings"
+)
+
+// ScriptTemplateOpt is a function type for configuring the script template.
+type ScriptTemplateOption func(*templateConfig)
+
+// templateConfig holds the configuration for the script template.
+type templateConfig struct {
+	params map[string]interface{}
+
+	customFuncs template.FuncMap
+}
+
+// WithScriptTemplateParams adds parameters to the script template.
+func WithScriptTemplateParams(params map[string]interface{}) ScriptTemplateOption {
+	return func(cfg *templateConfig) {
+		for k, v := range params {
+			cfg.params[k] = v
+		}
+	}
+}
+
+// WithCustomTemplateFunc adds a custom function to the template.
+func WithCustomTemplateFunc(name string, fn interface{}) ScriptTemplateOption {
+	return func(cfg *templateConfig) {
+		cfg.customFuncs[name] = fn
+	}
+}
+
+// ScriptTemplate processes a script template with parameters and returns the
+// corresponding script bytes. This functions allows Bitcoin scripts to be
+// created using a DSL-like syntax, based on Go's templating system.
+//
+// An example of a simple p2pkh template would be:
+//
+//	`OP_DUP OP_HASH160 0x14e8948c7afa71b6e6fad621256474b5959e0305 OP_EQUALVERIFY OP_CHECKSIG`
+//
+// Strings that have the `0x` prefix are assumed to byte strings to be pushed
+// ontop of the stack. Integers can be passed as normal. If a value can't be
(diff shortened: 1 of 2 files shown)


txscript/template.go
```diff
@@ -0,0 +1,226 @@
...
+		}
+	}
+
+	return len(s) > 0
+}
+
+// processScript converts the template output to actual script bytes. We scan
+// each line, then go through each element one by one, deciding to either add a
+// normal op code, a push data, or an integer value.
+func processScript(script string) ([]byte, error) {
```
yyforyongyu, line 108: let's add a unit test?

txscript/template.go
```diff
@@ -0,0 +1,226 @@
...
+		return nil, fmt.Errorf("failed to execute template: %w", err)
+	}
+
+	return processScript(buf.String())
+}
+
+// looksLikeInt checks if a string looks like an integer.
+func looksLikeInt(s string) bool {
+	// Check if the string starts with an optional sign
+	if len(s) > 0 && (s[0] == '+' || s[0] == '-') {
```
yyforyongyu, line 91: when will `+` or `-` be used?

txscript/template.go
```diff
@@ -0,0 +1,236 @@
...
+				return nil, fmt.Errorf("invalid "+
+					"integer: %s", token)
+			}
+
+			builder.AddInt64(val)
+
+		// Otherwise, we assume it's a byte string without the 0x
+		// prefix.
+		default:
+			data, err := hex.DecodeString(token)
```
guggero, line 161: Why do we allow both? Wouldn't it be "safer" or more explicit if we only allowed data pushes to be declared with the `0x` prefix and error out here? Then we also wouldn't need both the `hex` and `hex_str` functions, which I would need to look up the difference for if I was using this library.
Or is this for backward compatibility with existing scripts that don't use the prefix?

txscript/template_test.go
```diff
@@ -0,0 +1,358 @@
...
+}
+
+// TestScriptTemplateHelperFunctions tests the helper functions used in
+// templates.
+func TestScriptTemplateHelperFunctions(t *testing.T) {
+	t.Run("rangeIter", func(t *testing.T) {
+		result := rangeIter(2, 5)
+		expected := []int{2, 3, 4}
+
+		if len(result) != len(expected) {
```
guggero: nit: use `require` library everywhere?

guggero (approved):
Very nice!

txscript/template_test.go
Roasbeef: Done!

txscript/template.go
Roasbeef: > Then we also wouldn't need both the hex and hex_str functions, which I would need to look up the difference for if I was using this library.

The `hex` function is nice, as it makes the templating a bit easier (eg: you can just specify the bytes without going through that manual step of adding the prefix). I have a PR in `lnd` (many stacked on top of this one) that uses this to change how we specify Scripts in the `input` package: https://github.com/lightningnetwork/lnd/pull/9639/commits/748d6e109f7ccea227872fa75e956dd5f1678643#diff-5417cb1a81f342c8a87e557eb6d82431a0f451123c73cb638c361bf55ee41201R926.

PR #2219, opened 2024-07-25
ellswift: introduce ElligatorSwift encoding and decoding funcs

Crypt-iQ:
The BIP324 ElligatorSwift test vectors are also included. They can be found [here](https://github.com/bitcoin/bips/blob/master/bip-0324/ellswift_decode_test_vectors.csv) and [here](https://github.com/bitcoin/bips/blob/master/bip-0324/xswiftec_inv_test_vectors.csv). This code could be more optimized, possibly by not using so many `new(FieldVal)` invocations among some other things.

### DIFF
--- a/btcec/ellswift.go
+++ b/btcec/ellswift.go
@@ -1,0 +1,392 @@
+package btcec
+
+import (
+	"crypto/rand"
+	"fmt"
+
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+)
+
+var (
+	// c is sqrt(-3) (mod p)
+	c FieldVal
+
+	cBytes = [32]byte{
+		0x0a, 0x2d, 0x2b, 0xa9, 0x35, 0x07, 0xf1, 0xdf,
+		0x23, 0x37, 0x70, 0xc2, 0xa7, 0x97, 0x96, 0x2c,
+		0xc6, 0x1f, 0x6d, 0x15, 0xda, 0x14, 0xec, 0xd4,
+		0x7d, 0x8d, 0x27, 0xae, 0x1c, 0xd5, 0xf8, 0x52,
+	}
+
+	ellswiftTag = []byte("bip324_ellswift_xonly_ecdh")
+
+	// ErrPointNotOnCurve is returned when we're unable to find a point on the
+	// curve.
+	ErrPointNotOnCurve = fmt.Errorf("point does not exist on secp256k1 curve")
+)
+
+func init() {
+	c.SetByteSlice(cBytes[:])
+}
+
+// XSwiftEC() takes two field elements (u, t) and gives us an x-coordinate that
+// is on the secp256k1 curve. This is used to take an ElligatorSwift-encoded
+// public key (u, t) and return the point on the curve it maps to.
+// TODO: Rewrite these so to avoid new(FieldVal).Add(...) usage?
+// NOTE: u, t MUST be normalized. The result x is normalized.
+func XSwiftEC(u, t *FieldVal) *FieldVal {
+	// 1. Let u' = u if u != 0, else = 1
+	if u.IsZero() {
+		u.SetInt(1)
+	}
+
+	// 2. Let t' = t if t != 0, else 1
+	if t.IsZero() {
+		t.SetInt(1)
+	}
+
+	// 3. Let t'' = t' if g(u') != -(t'^2); t'' = 2t' otherwise
+	// g(x) = x^3 + ax + b, a = 0, b = 7
+
+	// Calculate g(u').
+	gu := new(FieldVal).SquareVal(u).Mul(u).AddInt(7).Normalize()
+
+	// Calculate the right-hand side of the equation (-t'^2)
+	rhs := new(FieldVal).SquareVal(t).Negate(1).Normalize()
+
+	if gu.Equals(rhs) {
+		// t'' = 2t'
+		t = t.Add(t)
+	}
+
+	// 4. X = (u'^3 + b - t''^2) / (2t'')
+	tSquared := new(FieldVal).SquareVal(t).Negate(1)
+	xNum := new(FieldVal).SquareVal(u).Mul(u).AddInt(7).Add(tSquared)
+	xDenom := new(FieldVal).Add2(t, t).Inverse()
+	x := xNum.Mul(xDenom)
+
+	// 5. Y = (X+t'') / (u' * c)
+	yNum := new(FieldVal).Add2(x, t)
+	yDenom := new(FieldVal).Mul2(u, &c).Inverse()
+	y := yNum.Mul(yDenom)
+
+	// 6. Return the first x in (u'+4Y^2, -X/2Y - u'/2, X/2Y - u'/2) for which
+	//    x^3 + b is square.
+
+	//    6a. Calculate u' +4Y^2 and determine if x^3+7 is square.
+	ySqr := new(FieldVal).Add(y).Mul(y)
+	quadYSqr := new(FieldVal).Add(ySqr).MulInt(4)
+	firstX := new(FieldVal).Add(u).Add(quadYSqr)
+
+	firstXCurve := new(FieldVal).Add(firstX).Square().Mul(firstX).AddInt(7)
+
+	// Now determine if firstXCurve is square (on the curve).
+	if new(FieldVal).SquareRootVal(firstXCurve) {
+		return firstX.Normalize()
+	}
+
+	//    6b. Calculate -X/2Y - u'/2 and determine if x^3 + 7 is square
+	doubleYInv := new(FieldVal).Add(y).Add(y).Inverse()
+	xDivDoubleYInv := new(FieldVal).Add(x).Mul(doubleYInv)
+	negXDivDoubleYInv := new(FieldVal).Add(xDivDoubleYInv).Negate(1)
+	invTwo := new(FieldVal).AddInt(2).Inverse()
+	negUDivTwo := new(FieldVal).Add(u).Mul(invTwo).Negate(1)
+	secondX := new(FieldVal).Add(negXDivDoubleYInv).Add(negUDivTwo)
+
+	secondXCurve := new(FieldVal).Add(secondX).Square().Mul(secondX).AddInt(7)
+
+	// Now determine if secondXCurve is square.
+	if new(FieldVal).SquareRootVal(secondXCurve) {
+		return secondX.Normalize()
+	}
+
+	//    6c. Calculate X/2Y -u'/2 and determine if x^3 + 7 is square
+	thirdX := new(FieldVal).Add(xDivDoubleYInv).Add(negUDivTwo)
+
+	thirdXCurve := new(FieldVal).Add(thirdX).Square().Mul(thirdX).AddInt(7)
+
+	// Now determine if thirdXCurve is square.
+	if new(FieldVal).SquareRootVal(thirdXCurve) {
+		return thirdX.Normalize()
+	}
+
+	// Should have found a square above.
+	panic("unreachable - no calculated x-values were square")
+}
+
+// XSwiftECInv takes two field elements (u, x) (where x is on the curve) and
+// returns a field element t. This is used to take a random field element u and
+// a point on the curve and return a field element t where (u, t) forms the
+// ElligatorSwift encoding.
+// TODO: Rewrite these so to avoid new(FieldVal).Add(...) usage?
+// NOTE: u, x MUST be normalized. The result `t` is normalized.
+func XSwiftECInv(u, x *FieldVal, caseNum int) *FieldVal {
+	v := new(FieldVal)
+	s := new(FieldVal)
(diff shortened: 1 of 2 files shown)


saubyk:
cc: @ellemouton @ProofOfKeags for review

btcec/ellswift.go
```diff
@@ -0,0 +1,398 @@
...
+	panic("unreachable - no calculated x-values were square")
+}
+
+// XSwiftECInv takes two field elements (u, x) (where x is on the curve) and
+// returns a field element t. This is used to take a random field element u and
+// a point on the curve and return a field element t where (u, t) forms the
+// ElligatorSwift encoding.
+// TODO: Rewrite these so to avoid new(FieldVal).Add(...) usage?
+// NOTE: u, x MUST be normalized. The result `t` is normalized.
+func XSwiftECInv(u, x *FieldVal, caseNum int) *FieldVal {
```
ProofOfKeags, line 125: 🫡 also good

btcec/ellswift.go
```diff
@@ -0,0 +1,398 @@
...
+		return t
+	}
+
+	panic("should not reach here")
+}
+
+// XElligatorSwift takes the x-coordinate of a point on secp256k1 and generates
+// ElligatorSwift encoding of that point composed of two field elements (u, t).
+// NOTE: x MUST be normalized. The return values u, t are normalized.
+func XElligatorSwift(x *FieldVal) (*FieldVal, *FieldVal, error) {
```
ProofOfKeags, line 229: 🫡

btcec/ellswift.go
```diff
@@ -0,0 +1,398 @@
...
+	// We'll choose a random `u` value and a random case so that we can
+	// generate a `t` value.
+	// TODO: How does this loop need to be bounded, see secp256k1 repo's impl.
+	for i := 0; i < 15000; i++ {
```
ProofOfKeags, line 233: Yeah this seems like it's not part of the spec. I'd imagine you did this as a "just in case". Given that the loop starts with rand generation there is exponential dropoff in the likelihood of failing every loop?

btcec/ellswift.go
```diff
@@ -0,0 +1,398 @@
...
+			return u, t, nil
+		}
+	}
+
+	return nil, nil, ErrNoEllSwiftEncoding
+}
+
+// EllswiftCreate generates a random private key and returns that along with
+// the ElligatorSwift encoding of its corresponding public key.
+func EllswiftCreate() (*PrivateKey, [64]byte, error) {
```
ProofOfKeags, line 268: 🫡

btcec/ellswift.go
```diff
@@ -0,0 +1,398 @@
...
+	copy(ellswiftPub[0:32], (*uBytes)[:])
+	copy(ellswiftPub[32:64], (*tBytes)[:])
+
+	// Return (priv, ellswift_pub)
+	return privKey, ellswiftPub, nil
+}
+
+// EllswiftECDHXOnly takes the ElligatorSwift-encoded public key of a
+// counter-party and performs ECDH with our private key.
+func EllswiftECDHXOnly(ellswiftTheirs [64]byte, privKey *PrivateKey) ([32]byte,
```
ProofOfKeags, line 302: 🫡

btcec/ellswift.go
```diff
@@ -0,0 +1,398 @@
...
+		return nil, err
+	}
+
+	if initiating {
+		// Initiating, place our public key encoding first.
+		var msg []byte
+		msg = append(msg, ellswiftOurs[:]...)
+		msg = append(msg, ellswiftTheirs[:]...)
+		msg = append(msg, ecdhPoint[:]...)
+		return chainhash.TaggedHash([]byte("bip324_ellswift_xonly_ecdh"), msg),
```
ProofOfKeags, line 389: extract to single const?

ProofOfKeags (approved):
Looks to be a correct implementation of the BIP324 gear.

btcec/ellswift_test.go
```diff
@@ -0,0 +1,819 @@
...
+		// that the correct t-value is calculated.
+		for caseNum, expTString := range test.cases {
+			tVal := XSwiftECInv(uVal, xVal, caseNum)
+
+			if tVal == nil {
+				if expTString != "" {
+					t.Fatalf("t value different than expected")
+				}
+
+				return
```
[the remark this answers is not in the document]
Crypt-iQ: good catch

btcec/ellswift.go
Crypt-iQ: yeah I misread their code, in libsecp it's just a while loop

Crypt-iQ:
> This is great! fwiw, I had great success using the csv files from the BIP324 directly for the tests and it gave me great test coverage too, see esp the big test running through all the "packet_encoding_test_vectors.csv" rows [here](https://github.com/lnliz/bitcoin-bip324-proxy/blob/29236f977475e72b636ddb000f85c09b5b0e14bb/crypto/bip324_test.go#L76)

I don't have a strong opinion on whether we should add the csv files directly. In other parts of the codebase (i.e. other BIP test vectors), we just have the test vectors in the format that I have in this PR

ProofOfKeags (approved):

saubyk:
cc @ellemouton for review comments on this pr. Thanks

btcec/ellswift.go
```diff
@@ -0,0 +1,392 @@
...
+
+	// 6. Return the first x in (u'+4Y^2, -X/2Y - u'/2, X/2Y - u'/2) for which
+	//    x^3 + b is square.
+
+	//    6a. Calculate u' +4Y^2 and determine if x^3+7 is square.
+	ySqr := new(FieldVal).Add(y).Mul(y)
+	quadYSqr := new(FieldVal).Add(ySqr).MulInt(4)
+	firstX := new(FieldVal).Add(u).Add(quadYSqr)
+
+	firstXCurve := new(FieldVal).Add(firstX).Square().Mul(firstX).AddInt(7)
```
kcalvinalvin, line 81: Maybe refactor the check for if x is on the curve into a separate function since it's used 3 times?

btcec/ellswift.go
```diff
@@ -0,0 +1,392 @@
...
+
+	thirdXCurve := new(FieldVal).Add(thirdX).Square().Mul(thirdX).AddInt(7)
+
+	// Now determine if thirdXCurve is square.
+	if new(FieldVal).SquareRootVal(thirdXCurve) {
+		return thirdX.Normalize()
+	}
+
+	// Should have found a square above.
+	panic("unreachable - no calculated x-values were square")
```
kcalvinalvin, line 114: Not sure about the panic here. If there's a bug somewhere in the library and we reach this part of the code, it'll crash the nodes. Returning an error will allow us to gracefully shut the node down in case we reach here.

btcec/ellswift.go
```diff
@@ -0,0 +1,392 @@
...
+			return nil
+		}
+
+		// Let r be the square root of -s(4(u^3 + 7) + 3u^2s)
+		uSqr := new(FieldVal).Add(u).Square()
+		lhs := new(FieldVal).Add(uSqr).Mul(u).AddInt(7).MulInt(4)
+		rhs := new(FieldVal).Add(uSqr).MulInt(3).Mul(s)
+
+		// Add the two terms together and multiply by -s.
+		lhs.Add(rhs).Normalize().Mul(s).Negate(1)
```
kcalvinalvin, line 160: Does it not matter if we call `Normalize()` during or after? Asking since

```
lhs.Add(rhs).Mul(s).Negate(1).Normalize()
```
passes the tests and the BIP specifies that the modulo be done afterwards

btcec/ellswift.go
```diff
@@ -0,0 +1,392 @@
...
+	}
+
+	w := new(FieldVal)
+
+	if !w.SquareRootVal(s) {
+		// If no square root was found, return None.
+		return nil
+	}
+
+	switch caseNum & 5 {
```
kcalvinalvin, line 187: Feels like we can reduce code duplication here

kcalvinalvin (commented):
Went through bip324 and compared the functions specified in the bip and in the reference code with the code here

litecomb:
WriteV2MessageN should allow specify the garbage []byte, if nil default garbage is used
ReadV2MessageN should allow reading the garbage []byte

Rationale: this will permit custom nonstandard packets communication in the protocol, while still being compatible with bitcoin.

litecomb:
// transportVersion is the transport version we are currently using.
transportVersion = []byte{}



packet likely needed here to make inbound connections work with respect to bitcoin 0.27.1 acting as client

litecomb:
when bitcoin 27.1 is connecting as a client, we don't find his garbage terminator inside CompleteHandshake() and it gets stuck.
Also I don't get why // Send version packet. is generated packet but not sent the packet.
But at least when btcd is the client it seems to work. But not when btcd is the server.

btcec/ellswift.go
```diff
@@ -0,0 +1,392 @@
...
+	ScalarBaseMultNonConst(&privKey.Key, &result)
+	result.ToAffine()
+	return &result.X
+}
+
+// liftX returns the point P with x-coordinate `x` and even y-coordinate. If a
+// point exists on the curve, it returns true and false otherwise.
+// TODO: Use quadratic residue formula instead (see: BIP340)?
+func liftX(x *FieldVal) (*PublicKey, bool) {
+	ySqr := new(FieldVal).Add(x).Square().Mul(x).AddInt(7)
```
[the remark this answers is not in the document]
Crypt-iQ: Yes I think that works, I have to double check that negating the field element flips the parity.

btcec/ellswift.go
```diff
@@ -0,0 +1,392 @@
...
+	}
+}
+
+// EllswiftCreate generates a random private key and returns that along with
+// the ElligatorSwift encoding of its corresponding public key.
+func EllswiftCreate() (*PrivateKey, [64]byte, error) {
+	var randPrivKeyBytes [64]byte
+
+	// Generate a random private key
+	_, err := rand.Read(randPrivKeyBytes[:])
```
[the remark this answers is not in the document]
Crypt-iQ: Yup it does reduce it by the group order, but this should be 64 bytes not 32

btcec/ellswift.go
Crypt-iQ: I think we can only reduce code duplication in the case below if we calculate the shared values for all cases up-front which makes things less efficient since not all of these calculated values will be used.

btcec/ellswift.go
Crypt-iQ: This has to do with the way the `btcec` library deals with magnitude. We only need to normalize the output in certain scenarios (see godoc comments of each `FieldVal` method). The output needs to be normalized in this specific scenario because it's possible that the magnitude is greater than allowed for the `Mul` call.

btcec/ellswift.go
```diff
@@ -0,0 +1,392 @@
...
+// the x-only ecdh point.
+func V2Ecdh(priv *PrivateKey, ellswiftTheirs, ellswiftOurs [64]byte,
+	initiating bool) (*chainhash.Hash, error) {
+
+	ecdhPoint, err := EllswiftECDHXOnly(ellswiftTheirs, priv)
+	if err != nil {
+		return nil, err
+	}
+
+	if initiating {
```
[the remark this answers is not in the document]
Crypt-iQ: This is a linter thing, it doesn't want two cases

btcec/ellswift.go
```diff
@@ -0,0 +1,392 @@
...
+
+	ellswiftTag = []byte("bip324_ellswift_xonly_ecdh")
+
+	// ErrPointNotOnCurve is returned when we're unable to find a point on the
+	// curve.
+	ErrPointNotOnCurve = fmt.Errorf("point does not exist on secp256k1 curve")
+)
+
+func init() {
+	c.SetByteSlice(cBytes[:])
```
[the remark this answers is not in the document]
Crypt-iQ: I like using `init` to set constants, it's easier for me to reason about and more idiomatic imo.

btcec/ellswift.go
kcalvinalvin: Ah I see. Seems like `Negate` also has a max magnitude limit.

Also, seems like where the mod takes place doesn't really matter in this case according to this SE answer https://math.stackexchange.com/questions/849713/where-does-modulus-take-place

kcalvinalvin:
ACK on d0706cdac41f848a85ed8286ebc605637008e4d0

btcec/ellswift.go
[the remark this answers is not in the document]
Crypt-iQ: This works, I'll address this the next time I push up.

saubyk:
@Crypt-iQ are you planning to push more changes to this pr or is it good to go from your end?

Crypt-iQ:
Pretty sure this PR is g2g unless the `DecompressY` change wants to be brought in. IMO probably better to just review the other PR and close this one instead.

PR #2223, opened 2024-08-06
main: add flag to write execution traces

kcalvinalvin:
Execution traces are part of the go runtime tooling and is useful to check what is slowing down the ibd.  For example, a slow ibd because of slow block downloads from peers won't show up on cpu profiling but will on a trace.

Also execution tracing has [gotten a lot better](https://go.dev/blog/execution-traces-2024) with go 1.22

### DIFF
--- a/config.go
+++ b/config.go
@@ -114,6 +114,7 @@ type config struct {
 	ConnectPeers         []string      `long:"connect" description:"Connect only to the specified peers at startup"`
 	CPUProfile           string        `long:"cpuprofile" description:"Write CPU profile to the specified file"`
 	MemoryProfile        string        `long:"memprofile" description:"Write memory profile to the specified file"`
+	TraceProfile         string        `long:"traceprofile" description:"Write execution trace to the specified file"`
 	DataDir              string        `short:"b" long:"datadir" description:"Directory to store data"`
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -114,6 +114,7 @@ type config struct {
 	ConnectPeers         []string      `long:"connect" description:"Connect only to the specified peers at startup"`
 	CPUProfile           string        `long:"cpuprofile" description:"Write CPU profile to the specified file"`
 	MemoryProfile        string        `long:"memprofile" description:"Write memory profile to the specified file"`
+	TraceProfile         string        `long:"trace" description:"Write execution trace to the specified file"`
```
guggero, line 117: nit: call the flag `traceprofile` to not confuse it with a log level and be more in line with the other profile flags?

guggero (approved):
tACK, LGTM :tada:

config.go
Roasbeef: Makes sense ^

Roasbeef: Bump on this, can get merged once addressed.

kcalvinalvin: Addressed in the latest push

Roasbeef (approved):
LGTM 🐝

dstadulis:
Exciting new functionality 🤓 💪

PR #2225, opened 2024-08-06
blockchain: copy utxo status bytes to avoid UB

Crypt-iQ:
It is undefined behavior if we directly use the value from a Get call after the transaction has completed.


Roasbeef (approved):
Confirmed that we currently directly use the return value after the transaction has ended: https://github.com/btcsuite/btcd/blob/b161cd6a199b4e35acec66afc5aad221f05fe1e3/blockchain/utxocache.go#L608-L613

PR #2226, opened 2024-08-07
peer, main, netsync, blockchain: parallel block downloads

kcalvinalvin:
~~This PR modifies `netsync.Manager` so that all the header-first blocks downloaded before the last checkpoint is done out of order by utilizing `query.WorkManager` from neutrino.~~

~~Gonna put it in draft for now as testing is sorta difficult and I'm not convinced it's downloading blocks fasuter for mainnet. By my testing it works just fine in testnet but mainnet seems to be slow when downloading blocks. Still identifying where the bottleneck is and will make adjustments accordingly.~~

~~If anyone else would like to give this a try please let me know if you see speed ups or slow downs from this PR.~~

Parallel block download architecture
------------------------------------

## Block Download Window

We mainly follow the "block download window" model that bitcoin core is using. This model
allows the blocks to be downloaded out of order inside the given window.

In the given blocks below from block 01 to 15, if the window size is 4, the download sequence
might look like so:

```
     window
       |
 -------------
 |           |
b01 b02 b03 b04 b05 b06 b07 b08 b09 b10 b11 b12 b13 b14 b15
```

Any one of these blocks can be downloaded in parallel and these blocks are then queued to multiple
peers. Once b01 is downloaded, the block window moves forward, resulting in something that looks
like this:

```
         window
           |
     -------------
     |           |
b01 b02 b03 b04 b05 b06 b07 b08 b09 b10 b11 b12 b13 b14 b15
```

If b02 was already downloaded before we've downloaded b01, then the block window will shift by 2,
resulting in something that looks like so:

```
             window
               |
         -------------
         |           |
b01 b02 b03 b04 b05 b06 b07 b08 b09 b10 b11 b12 b13 b14 b15
```

The parameters used in the code:

  maxBlockDownloadWindow = 1024

## GetData size per peer

When requesting blocks from peers, we'll use a getdata message. This can have a maximum of 50,000 block hashes.
If we ask for more blocks from a single peer, we can reduce the latency since we only send a single getdata message
for many blocks. However, this may also result in a single peer slowing us down as that peer may be a slow peer.

By playing around with the paramter, I came up with 32 max blocks per peer. This results in insignificant slowdown
for a single fast peer while still being small enough to not be slowed down too much by a single peer.

The parameters used are:

  maxInFlightBlocksPerPeer = 32

Peer selection
---------------

Currently there's not an explicit peer selection logic in btcd itself. There's a PR on the neutrino side that allows
btcd to disconnect off of peers when they are not able to deliver a block before the timeout of 2 seconds.
https://github.com/lightninglabs/neutrino/pull/308

I've tested different types of peer selection methods such as:

1: disconnecting peers when they're timing out.
2: disconnecting peers when they're slower than a certain speed.
3: a comparative system where the peer that delievers the block slowest 2 consecutives times in a given window is disconnected.

and tried these with various tweaks to them as well but the underlying problem seems to be that we're just not connecting to
fast enough peers.

The peer selection logic only eliminates us of slow peers. If we're connected to 8 (the default amount of peers) slow peers,
then the IBD is slow no matter what. I thought this was a problem with btcd itself but testing on the latest version of Bitcoin 
core showed that it's just as slow as well.

My conclusion
-------------

I believe the current PR that I have is the simplest way of achieving ibd with parallel block downloads.
Since it's not any slower than the latest version of Bitcoin Core, I think it's worthwhile to merge what 
we have now and in follow up PRs make changes to the peer selection logic from the connmanager side.


kcalvinalvin:
Seems like the slowdowns are coming from a single peer that's slowing down the block processing. `peerWorkManager` having capabilities to completely disconnect a peer would help in this.

Roasbeef:
> Seems like the slowdowns are coming from a single peer that's slowing down the block processing. peerWorkManager having capabilities to completely disconnect a peer would help in this.

Interesting, is the issue that the single peer is assigned blocks uniformly and is always the last one to send (haven't looked in PR at detail yet). I have this tracking issue in the `neutrino` repo for adding stuff like dynamic tuning, better work assignment, and also work stealing. With work stealing, the faster peers would steal the block request from the work queue of the slow peer, with faster peers helping us to be as slow as the fastest peer.

saubyk:
cc: @Crypt-iQ @ProofOfKeags for review

kcalvinalvin:
Pushed all the code that I have and updated the original comment with extra information

saubyk:
cc: @starius for review

netsync/manager.go
```diff
@@ -543,6 +544,31 @@ func (sm *SyncManager) isSyncCandidate(peer *peerpkg.Peer) bool {
 	return true
 }
 
+// notifyPeerSubscribers notifies all the current peer subscribers of the peer
+// that was passed in.
+func (sm *SyncManager) notifyPeerSubscribers(peer *peerpkg.Peer) {
+	// Loop for alerting subscribers to the new peer that was connected to.
+	n := 0
+	for i, sub := range sm.peerSubscribers {
```
yyforyongyu, line 563: i think we can use a map instead for `peerSubscribers`

netsync/manager.go
```diff
@@ -1126,30 +1136,76 @@ func (sm *SyncManager) fetchHeaderBlocks() {
 		return
 	}
 
+	queueFetch := func(msg *wire.MsgGetData, quit chan struct{}) {
+		if msg == nil || len(msg.InvList) == 0 {
+			return
+		}
+
+		// Keep fetching until we don't have an error.
```
yyforyongyu, line 1145: Can we really do this? Is it safe?

netsync/manager.go
```diff
@@ -1926,6 +1927,29 @@ type peerSubscription struct {
 	cancel <-chan struct{}
 }
 
+// ConnectedPeers returns all the currently connected peers to the channel
+// and then any additional new peers on connect.
+func (sm *SyncManager) ConnectedPeers() (<-chan query.Peer, func(), error) {
```
yyforyongyu, line 2025: I think we should name it to sth else, it's confusing to see three `ConnectedPeers` defined in the repo, maybe `SubscribePeers`.

netsync/manager.go
```diff
@@ -847,22 +823,136 @@ func (sm *SyncManager) handleBlockMsg(bmsg *blockMsg) {
...
+
+	// If we didn't ask for this block then the peer is misbehaving.
+	blockHash := bmsg.block.Hash()
+	if _, exists := state.requestedBlocks[*blockHash]; !exists {
+		// The regression test intentionally sends some blocks twice
+		// to test duplicate block insertion fails.  Don't disconnect
+		// the peer or ignore the block when we're in regression test
+		// mode in this case so the chain code is actually fed the
+		// duplicate blocks.
+		if sm.chainParams != &chaincfg.RegressionNetParams {
```
yyforyongyu, line 981: In this case I think it's the tests should be updated, not the code, otherwise we cannot cover the branch here in our tests.

yyforyongyu (commented):
Fantastic feature! I need a second round to load all the context to dive deeper. As for now I think for new code we should have unit tests. Meanwhile is the dependent PR ready for review? I saw it’s draft stage. https://github.com/lightninglabs/neutrino/pull/308

netsync/manager.go
```diff
@@ -461,10 +461,17 @@ func (sm *SyncManager) handleNewPeerMsg(peer *peerpkg.Peer) {
 		return
 	}
 
-	log.Infof("New valid peer %s (%s)", peer, peer.UserAgent())
-
 	// Initialize the peer state.
 	isSyncCandidate := sm.isSyncCandidate(peer)
+	if !sm.current() && !isSyncCandidate {
```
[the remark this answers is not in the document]
kcalvinalvin: Oh hmm good point. In the case of when we're IBD-ing this doesn't matter but this will still disconnect peers that may try to sync off of us.

I'll address this later with other changes.

kcalvinalvin:
There were some git hell that I had to figure out of... Lost some of my code.

The latest push is good and depends on https://github.com/lightninglabs/neutrino/pull/308

saubyk:

mohamedawnallah:
Thanks @kcalvinalvin for the PR, I will review it by the end of this week 🙏

netsync/manager.go
```diff
@@ -1070,9 +1070,6 @@ func (sm *SyncManager) handleBlockMsgHeadersFirst(bmsg *blockMsg) {
 	// Add the block to the queue.
 	sm.queuedBlocks[*blockHash] = bmsg
 
-	// Remove block from the request map as we've added it to queuedBlocks.
-	delete(state.requestedBlocks, *blockHash)
```
yyforyongyu: Removing this check creates a DoS vector as the peer can now keep sending us orphan blocks to fill up the memory used by `b.orphans`.

netsync/manager.go
```diff
@@ -840,23 +987,128 @@ func (sm *SyncManager) handleBlockMsg(bmsg *blockMsg) {
...
+		}
+	}
+
+	// If we have any blocks to process, process them now.
+	isCheckpointBlock := false
+	for _, processBlock := range processBlocks {
+		// Process the block to include validation, best chain selection, orphan
+		// handling, etc.
+		_, _, err := sm.chain.ProcessBlock(processBlock.block, blockchain.BFFastAdd)
+		if err != nil {
```
yyforyongyu, line 1062: I think there exists a DoS vector that the peer can send us the same orphan block twice and shut down our node.

yyforyongyu (changes requested):
Comments from the previous round are not addressed, maybe due to this?
> There were some git hell that I had to figure out of... Lost some of my code.

My main comments are,
- would be nice to split the PR into at least two parts, so it's easier to review and we can have faster merges,
  - the first part deals with the split of processing block msgs based on `headersFirstMode`
  - the second part deals with the connection handling and parallel block queries (maybe they should be their own PRs)
- please add unit tests, given IBD is extremely critical.

I'm still in the process of digesting the IBD process, tho this PR is mainly about speeding up the checkpointed blocks (also just learnt that we don't have `assumevalid` in btcd yet...). My main question is whether it's possible for the remote peer to trick us into using `BFFastAdd` flag for blocks beyond checkpoints, will dig.

netsync/manager.go
kcalvinalvin: So `query.WorkManager` is designed around downloading the entire GetData request. This works well for block headers or cfilters but not really for blocks since we verify as they come in.

I wanted to have a guarantee that when you call `queueFetch`, you'll get all the blocks in that request. `queueFetch` will call a new query to `query.WorkManager` which will try to peers to download from.

This seemed to work well in my testing though now I see that you can design to not have this if you make use of `SyncManager.requestedBlocks` map. Though since you'll now need to have a lock around the map to prevent async access.

netsync/manager.go
kcalvinalvin: I named it this because we use this function to meet the interface for `query.WorkManager`. Kinda weird to pass in `SubscribePeers` to meet the interface for `ConnectedPeers` imo.

How about naming the other `ConnectedPeers` in main to `SubscribePeers`?

peer/peer.go
```diff
@@ -1961,6 +2016,12 @@ func (p *Peer) Disconnect() {
 	close(p.quit)
 }
 
+// OnDisconnect returns a channel that will be closed when this peer is
+// disconnected.
+func (p *Peer) OnDisconnect() <-chan struct{} {
+	return p.quit
```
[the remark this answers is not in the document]
kcalvinalvin: You can't close a receive-only channel so it should be fine right? I checked and I'm now allowed to compile the file.

image

netsync/manager.go
kcalvinalvin: Ok so this code doesn't exist anymore but it's something that's already existed in the code so I think it's out of the scope for this PR.

netsync/manager.go
kcalvinalvin: I don't think they can since we ignore blocks that we never requested during headers-first but I'll add a test to make sure it does not fail.

netsync/manager.go
kcalvinalvin: Since we specifically process blocks that are only in the headers chain that we've verified, I don't think this is a risk.

Also, even if we didn't have this check the orphan blocks will get removed after orphan expiration so we're safe from oom attacks.

netsync/manager.go
```diff
@@ -334,16 +335,48 @@ func (sm *SyncManager) startSync() {
...
+		if sm.headerList.Len() > 0 && sm.nextCheckpoint != nil {
+			e := sm.headerList.Back()
+			node := e.Value.(*headerNode)
+
+			// If the final hash equals next checkpoint, that
+			// means we've verified the downloaded headers and
+			// can start fetching blocks.
+			if node.hash.IsEqual(sm.nextCheckpoint.Hash) {
+				sm.startHeader = sm.headerList.Front()
+				sm.fetchHeaderBlocks()
```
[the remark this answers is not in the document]
kcalvinalvin: Not quite.

This branch specifically checks for block headers that we may already have if the sync peer disconnects off of us while downloading headers or blocks.

This lets us avoid downloading extra headers.

netsync/manager.go
```diff
@@ -1204,11 +1209,6 @@ func (sm *SyncManager) handleInvMsg(imsg *invMsg) {
 		// for the peer.
 		peer.AddKnownInventory(iv)
```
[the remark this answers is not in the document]
kcalvinalvin: Hmm I think this was part of an older code that didn't return from `handleInvMsg()` when headersFirstMode was active.

netsync/manager.go
```diff
@@ -840,23 +987,128 @@ func (sm *SyncManager) handleBlockMsg(bmsg *blockMsg) {
...
+			if dbErr, ok := err.(database.Error); ok && dbErr.ErrorCode ==
+				database.ErrCorruption {
+				panic(dbErr)
+			}
+		}
+
+		// Look for checkpointed blocks.
+		if processBlock.block.Hash().IsEqual(sm.nextCheckpoint.Hash) {
+			isCheckpointBlock = true
+		}
```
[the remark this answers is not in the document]
kcalvinalvin: > Currently, this section of code panics whenever ANY block (not just checkpoint blocks) fails validation in headers-first mode.

Yes this is the correct behavior since headers-first is based off of checkpoints in btcd.

netsync/manager.go
```diff
@@ -28,6 +30,16 @@ const (
 	// more.
 	minInFlightBlocks = 10
 
+	// maxInFlightBlocks is the maximum number of blocks per peer that
+	// should be in the request queue for headers-first mode before
+	// requesting more.
+	maxInFlightBlocksPerPeer = 32
```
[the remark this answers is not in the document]
kcalvinalvin: It could but:

1: I'm not sure having more options is good for the end user. Especially since most don't even know about the utxo caching.
2: I'll leave further improvements for later PRs.

netsync/manager.go
```diff
@@ -28,6 +30,16 @@ const (
...
+
+	// blockDownloadWindow is the maximum number of blocks that are allowed
+	// to be fetch from the current best tip.  This number limits the memory
+	// usage a node will have.
+	blockDownloadWindow = 1024
```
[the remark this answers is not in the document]
kcalvinalvin: So I did shrink this down to 256 blocks which would mean 1GB in the absolute worst case (probably won't ever happen since 4MB blocks need to be specially crafted) and 512MB in the average case where blocks are 2MB.

I think this does address the concerns of btcd taking up too much memory and I'll leave futher improvements for later PRs.

netsync/manager.go
```diff
@@ -840,23 +987,128 @@ func (sm *SyncManager) handleBlockMsg(bmsg *blockMsg) {
...
+	peer := bmsg.peer
+	_, exists := sm.peerStates[peer]
+	if !exists {
+		log.Warnf("Received block message from unknown peer %s", peer)
 		return
 	}
+	blockHash := bmsg.block.Hash()
+
+	// Add the block to the queue.
+	sm.queuedBlocks[*blockHash] = bmsg
```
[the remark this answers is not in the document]
kcalvinalvin: https://github.com/btcsuite/btcd/pull/2226#discussion_r2253079483

netsync/manager.go
```diff
@@ -863,43 +863,15 @@ func (sm *SyncManager) handleBlockMsg(bmsg *blockMsg) {
 		}
 	}
 
-	// When in headers-first mode, if the block matches the hash of the
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest commit

netsync/manager.go
```diff
@@ -935,6 +937,170 @@ func (sm *SyncManager) handleBlockMsg(bmsg *blockMsg) {
 	}
 }
 
+// handleBlockMsgInHeadersFirst handles block messages from all peers when the
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest commit

peer/peer.go
```diff
@@ -1446,6 +1480,27 @@ out:
...
+			default:
+			}
+
+			// Spin up a goroutine so that we don't block here.
+			go func(subscription chan wire.Message,
+				quit chan struct{}) {
+
+				select {
+				case subscription <- rmsg:
+				case <-p.quit:
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in the latest commit

netsync/manager.go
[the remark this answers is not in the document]
kcalvinalvin: Been addressed in previous commits

peer/peer.go
```diff
@@ -1098,6 +1103,35 @@ func (p *Peer) readMessage(encoding wire.MessageEncoding) (wire.Message, []byte,
...
 }
 
+// recvMsgsubscription is two channels for a subscriber of recevied messages.
+// msgChan for sending the messages and quit for cancelling the subscription.
+type recvMsgsubscription struct {
+	msgChan chan wire.Message
+	quit    chan struct{}
+}
+
+// SubscribeRecvMsg adds a OnRead subscription to the peer.  All bitcoin
```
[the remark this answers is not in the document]
kcalvinalvin: Been addressed in previous commits

peer/peer.go
```diff
@@ -1098,6 +1103,35 @@ func (p *Peer) readMessage(encoding wire.MessageEncoding) (wire.Message, []byte,
 	return msg, buf, nil
 }
 
+// recvMsgsubscription is two channels for a subscriber of recevied messages.
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in previous commits

netsync/manager.go
```diff
@@ -1534,6 +1851,13 @@ func (sm *SyncManager) NewPeer(peer *peerpkg.Peer) {
 	sm.msgChan <- &newPeerMsg{peer: peer}
 }
 
+// queuePeerToBeDisconnected takes in a string of a peer and passes it to the
+// message channel in order for the peer to be disconnected.  It's done in this
+// fashion to avoid hv
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed as this no longer exists. Replaced with disconnectPeerMsg

netsync/manager.go
```diff
@@ -173,6 +192,86 @@ func limitAdd(m map[chainhash.Hash]struct{}, hash chainhash.Hash, limit int) {
...
+
+	// If we have blocks we're expecting, we've progressed but not finished.
+	if len(c.blocks) > 0 {
+		return query.Progress{
+			Finished:   false,
+			Progressed: true,
+		}
+	}
+
+	// We have no more blocks we're expecting from heres so we're finished.
```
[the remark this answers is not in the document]
kcalvinalvin: Already addressed in previous commits

netsync/manager.go
```diff
@@ -173,6 +192,86 @@ func limitAdd(m map[chainhash.Hash]struct{}, hash chainhash.Hash, limit int) {
...
+	sm *SyncManager) checkpointedBlocksQuery {
+
+	m := make(map[chainhash.Hash]struct{}, len(msg.InvList))
+	for _, inv := range msg.InvList {
+		m[inv.Hash] = struct{}{}
+	}
+	return checkpointedBlocksQuery{msg, sm, m}
+}
+
+// handleResponse returns that the progress is progressed and finished if the
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in previous commits.

netsync/manager.go
```diff
@@ -28,6 +29,16 @@ const (
 	// more.
 	minInFlightBlocks = 10
 
+	// maxInFlightBlocks is the maximum number of blocks per peer that
+	// should be in the request queue for headers-first mode before
+	// requesting more.
+	maxInFlightBlocksPerPeer = 32
+
+	// blockDownloadWindow is the maximum number of blocks that are allowed
+	// to be fetch from the current best tip.  This number limits the memory
```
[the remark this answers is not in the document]
kcalvinalvin: Addressed in previous commits

netsync/manager.go
kcalvinalvin: Addressed in previous commits

netsync/manager.go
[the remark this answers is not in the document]
kcalvinalvin: No longer the case and we return early if we're in headers-first mode.

netsync/manager.go
```diff
@@ -327,7 +327,12 @@ func (sm *SyncManager) startSync() {
 		// Clear the requestedBlocks if the sync peer changes, otherwise
 		// we may ignore blocks we need that the last sync peer failed
 		// to send.
-		sm.requestedBlocks = make(map[chainhash.Hash]struct{})
+		//
+		// We don't reset it during headersFirstMode since it's not used
+		// during headersFirstMode.
+		if !sm.headersFirstMode {
+			sm.requestedBlocks = make(map[chainhash.Hash]struct{})
```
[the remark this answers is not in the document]
kcalvinalvin: No longer the case as this code has been removed

netsync/manager.go
kcalvinalvin: So I wasn't thinking ahead when I said I'll write a test for this because it's really hard to write a test for this.

But I'm confident that this is safe since:
1: We ignore blocks from peers we aren't connected to.
2: We do not process blocks that aren't based off the headers that we've already downloaded and verified.

netsync/manager.go
```diff
@@ -328,18 +420,55 @@ func (sm *SyncManager) startSync() {
...
 		// we may ignore blocks we need that the last sync peer failed
 		// to send.
-		sm.requestedBlocks = make(map[chainhash.Hash]struct{})
-
-		locator, err := sm.chain.LatestBlockLocator()
-		if err != nil {
-			log.Errorf("Failed to get block locator for the "+
-				"latest block: %v", err)
-			return
+		//
```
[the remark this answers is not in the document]
kcalvinalvin: The logic for `choose a sync candidate` is now handled by `returnBestPeer`.
The logic for `in headers-first mode, request headers from the candidate` is now handled by `startHeadersFirstSync`.

The remaining logic for `in normal mode, request blocks from the candidate` remains in startSync.

netsync/manager.go
```diff
@@ -328,18 +420,55 @@ func (sm *SyncManager) startSync() {
...
 
+		sm.syncPeer = bestPeer
+
+		// Reset the last progress time now that we have a non-nil
+		// syncPeer to avoid instantly detecting it as stalled in the
+		// event the progress time hasn't been updated recently.
+		sm.lastProgressTime = time.Now()
+
+		// Check if we have some headers already downloaded.
+		var locator blockchain.BlockLocator
```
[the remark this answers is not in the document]
kcalvinalvin: No longer an issue to due refactoring out the logic into different functions.

https://github.com/btcsuite/btcd/pull/2226#discussion_r2253513936

netsync/manager.go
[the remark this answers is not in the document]
moawnallah: > So I did shrink this down to 256 blocks which would mean 1GB in the absolute worst case (probably won't ever happen since 4MB blocks need to be specially crafted) and 512MB in the average case where blocks are 2MB.
> 
> I think this does address the concerns of btcd taking up too much memory and I'll leave futher improvements for later PRs.

Reducing the scope of development on first iteration SGTM

netsync/manager.go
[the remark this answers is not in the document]
moawnallah: > It could but:
> 
> 1: I'm not sure having more options is good for the end user. Especially since most don't even know about the utxo caching. 2: I'll leave further improvements for later PRs.

Sounds good! We can iterate on this later 👍

kcalvinalvin:
Updated the neutrino library to commit `9bf820ee5929b7547ec98c2bb7303bb129777563`. I'll update it again if the preferred way to update it is to release a new verison and point to that neutrino lib.

netsync/manager.go
```diff
@@ -28,6 +30,16 @@ const (
 	// more.
 	minInFlightBlocks = 10
 
+	// maxInFlightBlocks is the maximum number of blocks per peer that
```
yyforyongyu, line 33:
```suggestion
	// maxInFlightBlocksPerPeer is the maximum number of blocks per peer that
```

netsync/manager.go
```diff
@@ -28,6 +30,16 @@ const (
...
+	// should be in the request queue for headers-first mode before
+	// requesting more.
+	maxInFlightBlocksPerPeer = 32
+
+	// blockDownloadWindow is the maximum number of blocks that are allowed
+	// to be fetched from the current best tip.  This number limits the memory
```
yyforyongyu, line 39:
```suggestion
	// to be fetched from the current best tip. This number limits the memory
```

netsync/manager.go
```diff
@@ -28,6 +30,16 @@ const (
...
+	// usage a node will have.
+	blockDownloadWindow = 256
```
yyforyongyu, line 41: Given that this will use 1 GB of RAM we def should make these configurable in a follow up PR.

netsync/manager.go
```diff
@@ -173,6 +190,89 @@ func limitAdd(m map[chainhash.Hash]struct{}, hash chainhash.Hash, limit int) {
...
+// received. If a block in the map of blocks is received, it returns that the
+// query.Progress has progressed. If there are no more blocks in the map, it
+// returns that the query.Progress is finished. If it receives a non-block message
+// or a block that is not in the map, it returns that query.Progress has not
+// progressed.
+func (c *checkpointedBlocksQuery) handleResponse(req, resp wire.Message,
+	peerAddr string) query.Progress {
+
+	block, ok := resp.(*wire.MsgBlock)
+	if !ok {
```
yyforyongyu, line 223: Can you explain why it's ok to drop other msgs here?

netsync/manager.go
```diff
@@ -328,61 +415,139 @@ func (sm *SyncManager) startSync() {
...
+func (sm *SyncManager) startSync() {
+	// Return now if we're already syncing.
+	if sm.syncPeer != nil {
+		return
+	}
+
+	// Once the segwit soft-fork package has activated, we only
+	// want to sync from peers which are witness enabled to ensure
+	// that we fully validate all blockchain data.
+	segwitActive, err := sm.chain.IsDeploymentActive(chaincfg.DeploymentSegwit)
```
yyforyongyu, line 480: we should add a field `segwitActive` to `SyncManager` and perform this check inside `New`, checking it here and silence the error is bad - if `startSync` fails, the `SyncManager` enters a state where it has no sync peer, and the stall detector is useless due to `sm.syncPeer` being nil. The only way to recover is for a new peer to connect or for an existing peer to disconnect.

netsync/manager.go
```diff
@@ -457,6 +622,34 @@ func (sm *SyncManager) isSyncCandidate(peer *peerpkg.Peer) bool {
...
+// notifyPeerSubscribers notifies all the current peer subscribers of the peer
+// that was passed in.
+func (sm *SyncManager) notifyPeerSubscribers(peer *peerpkg.Peer) {
+	sm.peerSubscriptionLock.Lock()
+	defer sm.peerSubscriptionLock.Unlock()
+
+	for sub := range sm.peerSubscribers {
+		select {
+		case <-sub.cancel:
+			delete(sm.peerSubscribers, sub)
```
yyforyongyu, line 634: nit: can we move the cleanup to other places and keep the method sticking to [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single-responsibility_principle)?

netsync/manager.go
```diff
@@ -560,6 +760,29 @@ func (sm *SyncManager) handleDonePeerMsg(peer *peerpkg.Peer) {
 	}
 }
 
+// handleDisconnectPeerMsg disconnects the passed in peer address.  If the peer
+// is a sync peer, we set another candidate as the sync peer.
+func (sm *SyncManager) handleDisconnectPeerMsg(peerAddr string) {
+	for peer, state := range sm.peerStates {
+		if peer.Addr() == peerAddr {
```
yyforyongyu, line 767: nit: check `if peer.Addr() != peerAddr` instead

netsync/manager.go
```diff
@@ -560,6 +760,29 @@ func (sm *SyncManager) handleDonePeerMsg(peer *peerpkg.Peer) {
...
+			peer.Disconnect()
+
+			delete(sm.peerStates, peer)
+			sm.clearRequestedState(state)
+
+			// Set a new peer as the sync peer if we're
+			// disconnecting it.
+			if peer == sm.syncPeer {
+				// Segwit is always active.
+				sm.syncPeer, _ = sm.returnBestPeer(true)
```
yyforyongyu, line 778: why are we silencing the error here? And the `sm.syncPeer` could be set to nil, and the sync process will be stalled?

netsync/manager.go
yyforyongyu: I checked the `query.WorkManager` and didn't find an interface method named `ConnectedPeers()` tho

yyforyongyu: also no need to return `error` given it's always nil.

netsync/manager.go
```diff
@@ -1671,6 +2109,66 @@ func (sm *SyncManager) Pause() chan<- struct{} {
...
+// queueFetchManager queues the given getdata to the fetch manager and waits for
+// the resulting error from the channel and returns the value.
+func (sm *SyncManager) queueFetchManager(msg *wire.MsgGetData) (
+	map[chainhash.Hash]bool, error) {
+
+	r := newCheckpointedBlocksQuery(msg, sm)
+	errChan := sm.fetchManager.Query(
+		r.requests(),
+		query.Cancel(sm.quit),
+		query.NumRetries(0),
```
yyforyongyu, line 2159: So retrying is disabled? why is that?

netsync/manager.go
```diff
@@ -845,23 +1047,146 @@ func (sm *SyncManager) handleBlockMsg(bmsg *blockMsg) {
...
+	haveInv, err := sm.haveInventory(&wire.InvVect{
+		Type: wire.InvTypeBlock,
+		Hash: *bmsg.block.Hash(),
+	})
+	if err != nil {
+		// This error should never happen since it's can only mean a
+		// database error.  We'll accept the block and if it's a
+		// duplicate, we'll just have an extra block in memory until
+		// we finish the headers-first download or we shutdown.
+		log.Warnf("failed to check if we already have block %v. %v",
```
yyforyongyu, line 1080: This should be `Errorf`

netsync/manager.go
```diff
@@ -845,23 +1047,146 @@ func (sm *SyncManager) handleBlockMsg(bmsg *blockMsg) {
...
+			bmsg.block.Hash(), err)
+	}
+	if haveInv {
+		log.Debugf("Ignoring duplicate block of %v during "+
+			"headers-first mode", bmsg.block.Hash())
```
yyforyongyu, line 1085:
```suggestion
			"headers-first mode", blockHash)
```

netsync/manager.go
```diff
@@ -845,23 +1047,146 @@ func (sm *SyncManager) handleBlockMsg(bmsg *blockMsg) {
...
+	_, exists := sm.peerStates[peer]
+	if !exists {
+		log.Warnf("Received block message from unknown peer %s", peer)
+		return
+	}
+	blockHash := bmsg.block.Hash()
+
+	haveInv, err := sm.haveInventory(&wire.InvVect{
+		Type: wire.InvTypeBlock,
+		Hash: *bmsg.block.Hash(),
```
yyforyongyu, line 1073:
```suggestion
		Hash: blockHash,
```

netsync/manager.go
```diff
@@ -845,23 +1047,146 @@ func (sm *SyncManager) handleBlockMsg(bmsg *blockMsg) {
...
+		return
+	}
+
+	// Add the block to the queue.
+	sm.queuedBlocks[*blockHash] = bmsg
+
+	// Log the progress as we have received the block.
+	sm.lastProgressTime = time.Now()
+
+	firstNodeEl := sm.headerList.Front()
```
yyforyongyu, line 1095: I think this `for` loop is a bit complex and inefficient. Current flow,
  1. A block arrives.
  2. It is unconditionally added to the `sm.queuedBlocks` map.
  3. A complex for loop is then initiated to iterate from the start of the headerList to see if the block we just added (or any other block in the queue) can now be processed.

A better flow,
  1. A block arrives.
  2. Check if this block is the one specific block we are currently waiting for (i.e., does its hash match the hash at the front of the headerList?).
  3. If it is not the block we are waiting for, add it to the `sm.queuedBlocks` map for later.
  4. If it is the block we are waiting for, process it immediately. Then, start a loop to check if the next block in the sequence is now waiting in the queue, and continue processing as long as possible.

netsync/manager.go
yyforyongyu: > We do not process blocks that aren't based off the headers that we've already downloaded and verified.

Yeah I think that's correct.

netsync/manager.go
```diff
@@ -845,23 +1047,146 @@ func (sm *SyncManager) handleBlockMsg(bmsg *blockMsg) {
...
+
+	// If we have any blocks to process, process them now.
+	isCheckpointBlock := false
+	for _, processBlock := range processBlocks {
+		// Process the block to include validation, best chain selection, orphan
+		// handling, etc.
+		_, _, err := sm.chain.ProcessBlock(processBlock.block, blockchain.BFFastAdd)
+		if err != nil {
+			// This is a checkpointed block and therefore should never fail
+			// validation.  If it did, then the binary is likely corrupted.
```
yyforyongyu, line 1142: > If it did, then the binary is likely corrupted.

I don't think it's valid. I mean, if the binary is corrupted, why doesn't the attacker just modify the panic too? I think most likely there is a bug in `ProcessBlock`, in which case it makes sense to shut down.

We will need more explicit docs explaining why it's ok to panic, or better, use critical logs here to gracefully shut down in case there are db operations. It will be confusing at first glance as we are processing external data sent from an untrusted peer, yet it seems like the peer can cause us to crash - we should mention that because we have checkpoints, the block hash is already checked and it's what we requested, the only failure here is a bug in `ProcessBlock`, in which case we should fail fast.

netsync/manager.go
```diff
@@ -892,12 +1217,22 @@ func (sm *SyncManager) handleBlockMsg(bmsg *blockMsg) {
...
+		// start requesting blocks from it.
+		sm.updateSyncPeer(false)
+		return
+	}
+
+	err = sm.syncPeer.PushGetBlocksMsg(locator, &zeroHash)
 	if err != nil {
 		log.Warnf("Failed to send getblocks message to peer %s: %v",
-			peer.Addr(), err)
+			sm.syncPeer.Addr(), err)
```
yyforyongyu, line 1235: This method is a bit long and complicated, I would favor breaking it down so it's easier for the reviewers, sth like,
```go
// handleBlockMsgHeadersFirst is the main entry point for a received block
// during headers-first sync. It coordinates queuing the block and then
// attempting to process any available sequential blocks.
func (sm *SyncManager) handleBlockMsgHeadersFirst(bmsg *blockMsg) error {
	blockHash := bmsg.block.Hash()

	// Step 1: Add the new block to our queue of out-of-order blocks.
	// A peer that sends a block we already have is misbehaving.
	// NOTE: This check is not perfect due to race conditions but is a
	// reasonable first-pass filter.
	if _, exists := sm.queuedBlocks[*blockHash]; exists {
		log.Warnf("Peer %s sent duplicate block %v in headers-first mode. "+
			"Disconnecting.", bmsg.peer, blockHash)
		bmsg.peer.Disconnect()
		return fmt.Errorf("peer %s sent duplicate block %v", bmsg.peer, blockHash)
	}
	sm.queuedBlocks[*blockHash] = bmsg
	sm.lastProgressTime = time.Now()

	// Step 2: Attempt to process any sequential blocks that are now available.
	// The peer that delivered the block which unlocks processing is passed
	// for attribution in case of errors.
	if err := sm.processQueuedBlocks(bmsg.peer); err != nil {
		// A validation or processing error occurred. The responsible peer has
		// already been disconnected by the helper function. We just need to
		// propagate the error so the sync state can be reset.
		return fmt.Errorf("failed to process queued blocks: %w", err)
	}

	// Step 3: If the pipeline is running low, request more blocks.
	// This is now clean and separate from the processing logic.
	if sm.shouldFetchMoreBlocks() {
		sm.fetchHeaderBlocks()
	}

	return nil
}
```

peer/peer.go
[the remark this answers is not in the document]
yyforyongyu: yeah correct - TIL

peer/peer.go
```diff
@@ -1493,6 +1528,28 @@ out:
...
+		for sub := range p.subscribers {
+			select {
+			case <-sub.quit:
+				delete(p.subscribers, sub)
+				continue
+			default:
+			}
+
+			// Spin up a goroutine so that we don't block here.
+			go func(subscription recvMsgsubscription,
```
yyforyongyu, line 1542: Just wanna note that this will create many short-lived goroutines given it spawns a new goroutine for every single message delivered to every single subscriber - we may consider buffering the subscriber's channel, tho non-blocking.

kcalvinalvin:
Ok so maybe I should've said it earlier but I don't really think this code is good.

The query package manager isn't really fit for the use in btcd because of a few reasons:

1: You can't tell which peer returned which block from the netsync package, making it hard to tell which peers to punish. The PR on the query package was meant to address this but it really is a hack and very ugly.

2: netsync package now has two different routes in where the block is received. The current path from the peer package in which the blocks are queued with `QueueBlock`. The extra path added by the query package. This makes it hard to reason about the code since you're receiving the same block in two different parts of the code.

3: The aforementioned (2) results in a lot of exceptions in the code since it's receiving the block in two different code paths. This results in a code bloat and I don't really see this as maintainable.

For the above reasons, I can't really see how this code would be manageable going forward. I myself don't really see this code as something good to have in the btcd codebase. I can't stand behind this code.

Sorry for all the time wasted on this. I tried to make it work but I don't think this is architecturally sound. I'll push a different PR soon.

PR #2233, opened 2024-08-14
wire: check TXID length before creating outpoint

jharveyb:
`chainhash.NewHashFromStr` will zero-pad the given string if it's smaller than 32 bytes / 64 hex chars.

That's fine, but when creating an OutPoint from a string, the TXID part of the string must be exactly 64 hex chars to (possibly) be a valid TXID. Right now, `NewOutPointFromString` accepts a TXID of less than 64 chars, and will end up padding the end with 0's.

This PR rejects incorrectly-sized TXID strings in `NewOutPointFromString`.


guggero (approved):

Roasbeef (approved):
LGTM 💿

PR #2237, opened 2024-08-15
log: pull in logrotator, switch to ZSTD compressor

jharveyb:
The log rotator we currently use is from a now-unmaintained repo: https://github.com/jrick/logrotate/tree/master

Which is fine, it works. But this issue https://github.com/lightninglabs/taproot-assets/issues/1084 prompted me to look at how we could switch from gzip to zstd for log compression (better compression at the default level: https://pkg.go.dev/github.com/klauspost/compress/zstd#readme-performance).

Since the log rotator is one short file, it seems easiest to just pull it into this repo directly vs. fork the `logrotate` package and release this updated version with ZSTD support, and then pull that in as a dependency.

If this gets merged to `btcd`, we could do the same `lnd` and then drop the `logrotate` dependency there as well. Once it's in the lnd `build` package, it could be used from the `taproot-assets` project.

The diff from upstream `logrotate` is quite small, and can be checked easily:

```bash
wget https://github.com/jrick/logrotate/raw/master/rotator/rotator.go
wget https://github.com/jharveyb/btcd/raw/logrotator_vendor_zstd/logrotator.go
diff -y --suppress-common-lines rotator.go logrotator.go
```

This is a more minimal diff from difftastic (https://difftastic.wilfred.me.uk/) with upstream on the left, and my version on the right:

<details>

```go
logrotator.go --- 1/6 --- Go
14 //                                                      14 //
15 // THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS A 15 // THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS A
.. ND CONTRIBUTORS "AS IS"                                 .. ND CONTRIBUTORS "AS IS"
16 // AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BU 16 // AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BU
.. T NOT LIMITED TO, THE                                   .. T NOT LIMITED TO, THE
17 // IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FO 17 // IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FO
.. R A PARTICULAR PURPOSE ARE                              .. R A PARTICULAR PURPOSE
18 // DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER O 18 // ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLD
[...]

PR #2238, opened 2024-08-16
log: bump logrotate dep, switch to zstd compressor

jharveyb:
Replacement for #2237 since `logrotate` was updated.

This issue https://github.com/lightninglabs/taproot-assets/issues/1084 prompted me to look at how we could switch from gzip to zstd for log compression (better compression at the default level: https://pkg.go.dev/github.com/klauspost/compress/zstd#readme-performance).

The `klauspost` ZSTD implementation seems like the best pure-Go choice, at least until it gets into the stdlib.

### DIFF
--- a/config.go
+++ b/config.go
@@ -41,6 +41,7 @@ const (
 	defaultLogLevel              = "info"
 	defaultLogDirname            = "logs"
 	defaultLogFilename           = "btcd.log"
+	defaultLogCompressor         = Gzip
 	defaultMaxPeers              = 125
 	defaultBanDuration           = time.Hour * 24
 	defaultBanThreshold          = 100
@@ -125,6 +126,7 @@ type config struct {
 	FreeTxRelayLimit     float64       `long:"limitfreerelay" description:"Limit relay of transactions with no transaction fee to the given amount in thousands of bytes per minute"`
 	Listeners            []string      `long:"listen" description:"Add an interface/port to listen for connections (default all interfaces port: 8333, testnet: 18333)"`
 	LogDir               string        `long:"logdir" description:"Directory to log output."`
+	LogCompressor        string        `long:"logcompressor" description:"Compression algorithm to use when rotating logs." choice:"gzip" choice:"zstd"`
 	MaxOrphanTxs         int           `long:"maxorphantx" description:"Max number of orphan transactions to keep in memory"`
 	MaxPeers             int           `long:"maxpeers" description:"Max number of inbound and outbound peers"`
 	MiningAddrs          []string      `long:"miningaddr" description:"Add the specified payment address to the list of addresses to use for generated blocks -- At least one address is required if the generate option is set"`
@@ -428,6 +430,7 @@ func loadConfig() (*config, []string, error) {
 		RPCMaxConcurrentReqs: defaultMaxRPCConcurrentReqs,
 		DataDir:              defaultDataDir,
 		LogDir:               defaultLogDir,
+		LogCompressor:        defaultLogCompressor,
 		DbType:               defaultDbType,
 		RPCKey:               defaultRPCKeyFile,
 		RPCCert:              defaultRPCCertFile,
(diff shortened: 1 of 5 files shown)


jharveyb:
Needed to pick an older version of `compress` as they stopped testing builds with `go1.17` at version `v1.15.15`.

https://github.com/klauspost/compress/commit/0793ca190d9604e0254fab30905bc6ffbeb30fce

jrick:
v1.1.1 fixes an (albeit unlikely) data race.

guggero (approved):
LTM after a version bump to `v1.1.` as mentioned above.

jharveyb:
Will update this to be a config option that defaults to the (existing) gzip before merge.

jharveyb:
Added a config option that allows `gzip` and `zstd` as options. Tested as follows, generating logs by starting testnet IBD with `debuglevel=trace`:

- An invalid compressor string, 'nocompress' is rejected both as CLI flag and config file option
- The help message shows the new flag
- Setting `logcompressor=zstd` via config file and CLI worked, and produced compressed logs with suffix `zst` that zgrep handled successfully
- Setting `logcompressor=gzip` via config file worked, and produced compressed logs with suffix `gz` that zgrep handled successfully
- Omitting the `logcompressor` option from the config file led to compressed logfiles with suffiz `gz`, maintaining the existing behavior

config.go
```diff
@@ -125,6 +126,7 @@ type config struct {
 	FreeTxRelayLimit     float64       `long:"limitfreerelay" description:"Limit relay of transactions with no transaction fee to the given amount in thousands of bytes per minute"`
 	Listeners            []string      `long:"listen" description:"Add an interface/port to listen for connections (default all interfaces port: 8333, testnet: 18333)"`
 	LogDir               string        `long:"logdir" description:"Directory to log output."`
+	LogCompressor        string        `long:"logcompressor" description:"Compression algorithm to use when rotating logs.  Valid algorithms are: gzip and zstd."`
```
guggero, line 129: Could add `choice:"gzip" choice:"zstd"` to give some validation responsibility to the config parsing library.

guggero (approved):
utACK, LGTM :tada:

config.go
jharveyb: nice, didn't realize those were being useful anywhere 😓

jharveyb:
Updated to address feedback.

doc.go
```diff
@@ -76,6 +76,8 @@ Application Options:
 	                            (default all interfaces port: 8333, testnet:
 	                            18333, signet: 38333)
 	    --logdir=               Directory to log output
+	    --logcompressor=        Compression algorithm to use when rotating logs.
```
guggero, line 79: nit: this now shows `--logcompressor=[gzip,zstd]` with the `choice` added to the CLI flag.

log.go
```diff
@@ -121,6 +142,30 @@ func initLogRotator(logFile string) {
 		os.Exit(1)
 	}
 
+	// Reject unknown compressors.
+	if !supportedCompressor(cfg.LogCompressor) {
+		fmt.Printf("specified log compressor [%v] is invalid",
```
guggero, line 147: nit: need return here?

doc.go
guggero: Ah, sorry, I just meant that the text in the doc should probably be updated to show the actual output you would now see on the command line. So just add `[gzip,zstd]` to the `doc.go`.

yyforyongyu (approved):
LGTM!

jharveyb:
Rebased.

yyforyongyu:
hmmm weird CI failures...

kcalvinalvin:
Oh sorry I think this is me forgetting to put in build tags for `getchaintips_test.go` and `invalidate_reconsider_block_test.go`. CI passes for me on my branch https://github.com/kcalvinalvin/btcd/actions/runs/11233768822/job/31228158339

```diff
diff --git a/integration/getchaintips_test.go b/integration/getchaintips_test.go
index 1570ba74..5a4d39b5 100644
--- a/integration/getchaintips_test.go
+++ b/integration/getchaintips_test.go
@@ -1,3 +1,7 @@
+// This file is ignored during the regular tests due to the following build tag.
+//go:build rpctest
+// +build rpctest
+
 package integration
 
 import (
diff --git a/integration/invalidate_reconsider_block_test.go b/integration/invalidate_reconsider_block_test.go
index 4fe6ff00..cfc59840 100644
--- a/integration/invalidate_reconsider_block_test.go
+++ b/integration/invalidate_reconsider_block_test.go
@@ -1,3 +1,7 @@
+// This file is ignored during the regular tests due to the following build tag.
+//go:build rpctest
+// +build rpctest
+
 package integration
```

Roasbeef (changes requested):
FWIW, this ends up adding an entirely new dependency, just for slightly better log compression, how necessary is this? Seems like a nice to have, since a user can just watch the logs on disk, then rewrite them with their desired compression format.

PR #2239, opened 2024-08-20
btcec: update gomod and go version

yyforyongyu:
TODO:
- [ ] tag `btcec/v2.3.5`
- [ ] tag `btcutil/psbt/v1.1.10`
- [ ] tag `btcutil/v1.1.7`


btcec/go.mod
```diff
@@ -1,16 +1,19 @@
 module github.com/btcsuite/btcd/btcec/v2
 
-go 1.17
+go 1.22.6
```
guggero, line 3: I think we might need to keep this at an older version? Or at least have the old 2-part format, since this seems to break the unit tests (or the linter).

yyforyongyu: i think there's something wrong with our CI here, still digging but the tests are failing locally but not on github actions hmmm

yyforyongyu: turns out that locally all the sub-modules are pointed to the local versions as indicated by the `go.work`, which is now fixed once the versions are updated here.

guggero (approved):

btcec/go.mod
Roasbeef: Hmm, I do know there's this hardware wallet (?) that relies on GopherJS (Go transpiler) which doesn't support the latest version. In the past, we wanted to introduce some code using generics, but backed off at their request. Though ultimately we can't hold it all back forever...

yyforyongyu: hmm yeah didn't know that before. It looks like [they only support up to 0.19.0](https://github.com/gopherjs/gopherjs?tab=readme-ov-file#whats-new), should we use `0.19.0` instead?

davecgh (commented):
I no longer maintain this, so it's not my call, but I would **highly** suggest avoiding an entire dependency (and all of its dependencies, transitively) that everyone who uses the btcec module then has to also pull all for a single test.

I'm talking about testify here.  Adding an entire dependency for a single test is kind of ridiculous.

btcec/go.mod
```diff
@@ -1,16 +1,19 @@
 module github.com/btcsuite/btcd/btcec/v2
 
-go 1.17
+go 1.22
```
davecgh, line 3: Also, requiring Go 1.22 when it doesn't actually require it is bad for consumers of the dependency too.

PR #2242, opened 2024-08-22
bugfix: check if both begin and end are numbers in `UnmarshalJSON`

brunoerg:
In the current code, if we provide a non-number value (e.g. `"method":"deriveaddresses","params":["A",["A","B"]]`), it might cause a panic. This PR fixes it.


guggero (approved):
Thanks, LGTM :tada:

Roasbeef (approved):
LGTM 📈

PR #2244, opened 2024-09-01
btcutil: Add library functions for BIP-0352 silent payment _send_ support

guggero:
Adds library/API functionality for Silent Payment **send** support.

Implements the following BIPs (or pull requests to BIPs):
 - [BIP-0352 - Silent Payments](https://github.com/bitcoin/bips/blob/master/bip-0352.mediawiki) (added in https://github.com/bitcoin/bips/pull/1458)
 - [BIP-0374 - Discrete Log Equality Proofs](https://github.com/bitcoin/bips/blob/master/bip-0374.mediawiki) (added in https://github.com/bitcoin/bips/pull/1689)
 - [BIP-0375 - Sending Silent Payments with PSBTs](https://github.com/bitcoin/bips/blob/master/bip-0375.mediawiki) (added in https://github.com/bitcoin/bips/pull/1687)

Goal is to address **send** support first as that's rather straightforward. Receive support is much more involved as it requires scanning the chain in a specific manner.


benma:
This is great work and very welcome. However, BIP-375 specifies exclusion for PSBTv0, so this will not be compatible with other PSBT libraries/wallets. Imho it would be best to focus on adding support for PSBTv2 first (https://github.com/btcsuite/btcd/issues/2328).

myxmaster:
Hey, just wanted to drop a quick note as an LND user who's been following this PR and the overall silent payments effort.

I understand that PSBTv2 support (https://github.com/btcsuite/btcd/issues/2328) would be needed for full BIP375 compliance and interoperability with external signers/hardware wallets. But even without that, having SP send support land in btcutil would already be a huge step forward, since it would unblock LND PR https://github.com/lightningnetwork/lnd/pull/9398 and allow LND users to send to silent payment addresses using LND's internal wallet and signer.

Would love to see this land. Thanks for all the great work on this @guggero!

TechLateef:
Hey @guggero  and @benma , just a heads up that I'm starting work on the PSBTv2 support requested in #2328. Hopefully, that will help unblock full BIP375 compliance for this PR soon!

silentpayments/dleq.go
```diff
@@ -0,0 +1,206 @@
...
+	// Generate a random 32-byte value.
+	var r [32]byte
+	if _, err := crand.Read(r[:]); err != nil {
+		return nil, [64]byte{}, err
+	}
+
+	B := ScalarMult(privateKey.Key, scanKey)
+	G := Generator()
+
+	proof, err := DLEQProof(privateKey, B, G, r, nil)
```
yyforyongyu: `CreateShare` passes the computed ECDH share (`a·scanKey`) as the `B` argument to `DLEQProof`. `DLEQProof` then computes `C = a·B`, so this proof is for `a²·scanKey`, not for the returned share. A verifier using `(A=privateKey.PubKey(), B=scanKey, C=share)` will reject it. This should pass `scanKey` to `DLEQProof` and return the separately computed share.

silentpayments/dleq.go
```diff
@@ -0,0 +1,206 @@
...
+
+	// Let t be the byte-wise xor of bytes(32, a) and hash_BIP0374/aux(r).
+	aBytes := a.Key.Bytes()
+	t := chainhash.TaggedHash(TagBIP0374Aux, r[:])
+	for i := 0; i < len(t); i++ {
+		t[i] ^= aBytes[i]
+	}
+
+	// Let rand = hash_BIP0374/nonce(t || cbytes(A) || cbytes(C)).
+	rand := chainhash.TaggedHash(
```
yyforyongyu, line 63: Current BIP374 includes the optional message in the nonce derivation: `hash_BIP0374/nonce(t || cbytes(A) || cbytes(C) || m')`. This implementation omits `m`, which can leak `a` if the same aux randomness is reused with different messages.

silentpayments/input.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	switch script.Class() {
+	case txscript.PubKeyHashTy, txscript.WitnessV0PubKeyHashTy,
+		txscript.WitnessV1TaprootTy:
+
+		// These types are supported in any case.
+		return true, script, nil
+
+	case txscript.ScriptHashTy:
```
yyforyongyu, line 181: This accepts every P2SH output as silent-payment compatible. BIP352 only includes P2SH-P2WPKH in shared-secret derivation; arbitrary P2SH/multisig inputs should be excluded or otherwise identified precisely, otherwise sender and receiver can derive different secrets.

silentpayments/address.go
```diff
@@ -0,0 +1,337 @@
...
+	hrp, data, bechVersion, err := bech32.DecodeNoLimitWithVersion(addr)
+	if err != nil {
+		return nil, err
+	}
+
+	if bechVersion != bech32.VersionM {
+		return nil, errors.New("invalid bech32 version")
+	}
+
+	regrouped, err := bech32.ConvertBits(data[1:], 5, 8, false)
```
yyforyongyu, line 259: `DecodeAddress` slices `data[1:]` and later reads `data[0]` before checking that `data` is non-empty. A checksum-valid Bech32m string with an empty data part can panic here instead of returning an error.

psbt/partial_output.go
```diff
@@ -144,6 +147,31 @@ func (po *POutput) deserialize(r io.Reader) error {
 				po.TaprootBip32Derivation, taprootDerivation,
 			)
 
+		case SilentPaymentV0InfoOutputType:
```
yyforyongyu, line 150: BIP375 defines `PSBT_OUT_SP_V0_INFO` with no key data, but this case does not reject non-empty `keyData`. Malformed PSBTs with extra key data are accepted and normalized instead of returning `ErrInvalidKeyData`.

psbt/partial_output.go
```diff
@@ -144,6 +147,31 @@ func (po *POutput) deserialize(r io.Reader) error {
...
+			}
+
+			info, err := ReadSilentPaymentInfo(value)
+			if err != nil {
+				return err
+			}
+
+			po.SilentPaymentInfo = info
+
+		case SilentPaymentV0LabelOutputType:
```
yyforyongyu, line 166: Same issue for `PSBT_OUT_SP_V0_LABEL`: the field has no key data per BIP375, so this should reject non-empty `keyData` before parsing the value.

silentpayments/dleq.go
```diff
@@ -0,0 +1,206 @@
...
+	// Return the proof.
+	return proof, nil
+}
+
+// DLEQVerify verifies a DLEQ proof according to BIP-0374 for the given public
+// keys A, B, C, G, proof, and optional message m.
+func DLEQVerify(A, B, C, G *btcec.PublicKey, proof [64]byte, m *[32]byte) bool {
+	// Fail if any of is_infinite(A), is_infinite(B), is_infinite(C),
+	// is_infinite(G).
+	if !A.IsOnCurve() || !B.IsOnCurve() || !C.IsOnCurve() {
```
yyforyongyu: The comment quotes BIP374's requirement to fail if any of `A`, `B`, `C`, or `G` is infinite/off-curve, but the code only checks `A`, `B`, and `C`. Since `G` is caller-supplied, it should be validated too.

silentpayments/address.go
```diff
@@ -0,0 +1,337 @@
...
+	for _, recipient := range recipients {
+		var key [33]byte
+		copy(key[:], recipient.ScanKey.SerializeCompressed())
+
+		groups[key] = append(groups[key], recipient)
+	}
+
+	idx := 0
+	grouped := make([][]Address, len(groups))
+	for _, group := range groups {
```
yyforyongyu: `GroupByScanKey` builds the returned slice by ranging over a map, so group order is randomized. Since `AddressOutputKeys` documents that recipients are ordered by output index and returns a flat result slice, this can make the returned output order nondeterministic for multiple scan-key groups.

yyforyongyu (commented):
Fired some agents to review it while catching up with the bip🤓

ping-localhost:
I currently use `btcutil.DecodeAddress` to validate if a provided Bitcoin address is valid. Seeing how it hasn't been changed, I'll have to make a fallback for Silent Payments once this is merged?

Is it maybe a good idea to add a more generic `ValidateAddress`-function that can deal with all types? Just brain dumping, so feel free to ignore me.

guggero:
> I currently use `btcutil.DecodeAddress` to validate if a provided Bitcoin address is valid. Seeing how it hasn't been changed, I'll have to make a fallback for Silent Payments once this is merged?
> 
> Is it maybe a good idea to add a more generic `ValidateAddress`-function that can deal with all types? Just brain dumping, so feel free to ignore me.

Hmm, that's a valid point. But I'm not sure what would be best here. Adding anything to the existing `DecodeAddress` in the now-extracted `address/v2` package can't be done, as that would introduce a circular package dependency.
And adding a new `DecodeAnyAddress` would need to return an `address.Address` interface. But that interface requires the `ScriptAddress() []byte` method to be defined that can be used to produce the `pkScript` of a receiver output.
But that's exactly the complication with Silent Payment addresses: They are dependent on the inputs being spent.
So for safety reasons, the `ScriptAddress` method would need to panic for Silent Payment addresses and any caller would need to know how to deal with them.

So perhaps a simple `silentpayments.IsValidAddress()` would be enough to make the detection easier?

silentpayments/dleq.go
guggero: Great catch! Fixed and added unit test.

silentpayments/dleq.go
guggero: Yes, this was updated later on in the BIP. Fixed and did a full review on the latest version of the BIP.

silentpayments/input.go
guggero: Correct. Passing in the public key now so we can at least assert the address wasn't _made_ from an incorrect (e.g. uncompressed key).

silentpayments/address.go
guggero: Yes, good catch. Added a fuzz test as well and ran it for a few hours on a beefy machine.

psbt/partial_output.go
guggero: Yes, fixed.

silentpayments/dleq.go
guggero: Makes sense, fixed.

silentpayments/address.go
guggero: Very good point, addressed and unit tested.

guggero:
Thanks a lot for the review, @yyforyongyu! I addressed all your points and did a full (agentic) review against the three updated BIPs and made sure all updated test vectors also pass. Should be quite solid now.

ping-localhost:
> So perhaps a simple silentpayments.IsValidAddress() would be enough to make the detection easier?

Yeah, that would help already! Anything to make the process easier is very much appreciated.

guggero:
> > So perhaps a simple silentpayments.IsValidAddress() would be enough to make the detection easier?
> 
> Yeah, that would help already! Anything to make the process easier is very much appreciated.

Okay, I've added `IsValidAddress(string) bool`.

PR #2245, opened 2024-09-02
btcjson: turn warnings into StringOrArray type

### LINKED ISSUE #2224: getnetworkinfo core RPC introduced new field in json object.
The current master branch of bitcoin-core introduced a new field into the `getnetworkinfo` rpc json-response. We should adapt the btcd code to account for this however I am not sure how we should do this in a backwards compatible manner. Any suggestions from prior RPC updates of the core software ?

```
bitcoin-cli --regtest getnetworkinfo
{
  ...
  "warnings": [
    "This is a pre-release test build - use at your own risk - do not use for mining or merchant applications"
  ]
}
```

```
diff --git a/btcjson/chainsvrresults.go b/btcjson/chainsvrresults.go
index 11c0483d..cfb75485 100644
--- a/btcjson/chainsvrresults.go
+++ b/btcjson/chainsvrresults.go
@@ -380,7 +380,7 @@ type GetNetworkInfoResult struct {
        RelayFee        float64                `json:"relayfee"`
        IncrementalFee  float64                `json:"incrementalfee"`
        LocalAddresses  []LocalAddressesResult `json:"localaddresses"`
-       Warnings        string                 `json:"warnings"`
+       Warnings        []string               `json:"warnings"`
 }

```

Thanks @michael1011 for pointing this out.

guggero:
Fixes #2224 and lightningnetwork/lnd#9053.

Depending on the version of Bitcoin Core, the "warnings" field in the response to getnetworkinfo is either a single string value or an array of strings.
We can easily parse those two variants with a custom type that implements an UnmarshalJSON method.

### DIFF
--- a/btcjson/chainsvrresults.go
+++ b/btcjson/chainsvrresults.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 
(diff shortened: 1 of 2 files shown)


guggero:
cc @ziggie1984.

btcjson/chainsvrresults.go
```diff
@@ -363,6 +364,49 @@ type LocalAddressesResult struct {
 	Score   int32  `json:"score"`
 }
 
+// StringOrArray defines a type that can be used as type that is either a single
+// string value or a string array in JSON-RPC commands, depending on the version
+// of the chain backend.
+type StringOrArray []string
+
+// MarshalJSON implements the json.Marshaler interface
```
ziggie1984, line 372: Nit: missing point at the end.

btcjson/chainsvrresults.go
```diff
@@ -363,6 +364,49 @@ type LocalAddressesResult struct {
...
+func (h StringOrArray) MarshalJSON() ([]byte, error) {
+	return json.Marshal(h)
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface
```
ziggie1984, line 377: Nit: same here.

ziggie1984 (approved):
Cool Idea, LGTM.

Wondering when we update to make generics available as a default here, still using golang 1.17 ? Ready to bump the version ?

guggero:
cc @Roasbeef would be nice to get this in. `bitcoind v28.0 RC2` just dropped, so can't be too much longer until the final version drops and people will update, causing their `lnd` to not start anymore because of this.

Roasbeef (approved):
LGTM 🏦

PR #2256, opened 2024-09-30
rpcclient: add timeout to http request.

ziggie1984:
Not sure why we did not have a timeout in the first place, but I think it makes sense. Did not choose to add it as a config value but can be changed.

### DIFF
--- a/rpcclient/infrastructure.go
+++ b/rpcclient/infrastructure.go
@@ -92,6 +92,10 @@ const (
 	// requestRetryInterval is the initial amount of time to wait in between
 	// retries when sending HTTP POST requests.
 	requestRetryInterval = time.Millisecond * 500
+
+	// defaultHTTPTimeout is the default timeout for an http request, so the
+	// request does not block indefinitely.
+	defaultHTTPTimeout = time.Minute
 )
 
 // jsonRequest holds information about a json request that is used to properly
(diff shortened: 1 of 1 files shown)


rpcclient/infrastructure.go
```diff
@@ -92,6 +92,10 @@ const (
 	// requestRetryInterval is the initial amount of time to wait in between
 	// retries when sending HTTP POST requests.
 	requestRetryInterval = time.Millisecond * 500
+
+	// defaultHttpTimeout is the default timeout for an http request, so the
+	// request does not block indefintily.
+	defaultHttpTimeout = time.Second * 2
```
Roasbeef, line 98: I think when the node is running on a Pi, certain I/O heavy calls can take a while, especially while processing blocks. Maybe 10 seconds instead? We do have retry logic already.

ziggie1984: good idea set to 10 sec.

rpcclient/infrastructure.go
```diff
@@ -92,6 +92,10 @@ const (
 	// requestRetryInterval is the initial amount of time to wait in between
 	// retries when sending HTTP POST requests.
 	requestRetryInterval = time.Millisecond * 500
+
+	// defaultHTTPTimeout is the default timeout for an http request, so the
+	// request does not block indefinitely.
+	defaultHTTPTimeout = time.Second * 10
```
yyforyongyu, line 98: I think this is usually 60s in "traditional" business...

yyforyongyu (approved):

PR #2260, opened 2024-10-02
multi: introduce v2transport and implement BIP324

Crypt-iQ:
Depends on https://github.com/btcsuite/btcd/pull/2219

This PR implements the [BIP324 encrypted version 2 handshake](https://github.com/bitcoin/bips/blob/master/bip-0324.mediawiki). The service flag is added to the list of supported services so that btcd will optimistically try to make v2 connections.


saubyk:
cc: @kcalvinalvin for review

kcalvinalvin:
Haven’t yet taken a look at other parts in details but seems like the PR is missing `NODE_P2P_V2` signaling support. Not sure if this part of the code was out of scope for the PR but

Should add

```

SFNodeP2PV2 = 1 << 11

```

here: https://github.com/btcsuite/btcd/blob/24eb815168f49dea84767817717a11bd7928eb23/wire/protocol.go#L72-L105

And other necessary code at:

https://github.com/btcsuite/btcd/blob/24eb815168f49dea84767817717a11bd7928eb23/server.go#L2723-L2743

kcalvinalvin (commented):
Initial round of review. Went through BIP324 and compared the flows and the pseudocode provided in the BIP with the code in `transport.go` and `chacha.go`

litecomb:
WriteV2MessageN should allow specify the garbage []byte, if nil default garbage is used
ReadV2MessageN should allow reading the garbage []byte

Rationale: this will permit custom nonstandard packets communication in the protocol, while still being compatible with bitcoin.

litecomb:
It would be great if v2transport wasn't promoted to go module to ease reviewing

v2transport/transport.go
```diff
@@ -0,0 +1,570 @@
...
+	// Check and see if the received bytes match the v1 protocol's message
+	// prefix. If it does, we'll revert to the v1 protocol. If it doesn't,
+	// we'll treat this as a v2 peer.
+	for len(p.receivedPrefix) < len(v1Prefix) {
+		var receiveBytes []byte
+		receiveBytes, err = p.Receive(1)
+		if err != nil {
+			return err
+		}
+
```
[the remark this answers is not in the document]
Crypt-iQ: I believe this is implemented on the receiving side

Crypt-iQ:
Pushed up the latest set of changes, will work on addressing the comments. Inbound v2 connections now work. The one thing that is not implemented that bitcoind implements and the BIP text [recommends](https://github.com/bitcoin/bips/blob/master/bip-0324.mediawiki#signaling-specification) we implement is optimistically reconnecting via v1 if an outbound v2 connection failed.

v2transport/transport.go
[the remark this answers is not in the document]
Crypt-iQ: I think this is implemented properly because we'll continue to parse the 16 bytes and eventually call `CompleteHandshake` which will compare the already-parsed 16 bytes and error out with `errWrongNetV1Peer`. Ideally it would have a test though.

v2transport/transport.go
```diff
@@ -0,0 +1,570 @@
...
+}
+
+// CompleteHandshake finishes the v2 protocol negotiation and optionally sends
+// decoy packets after sending the garbage terminator.
+func (p *Peer) CompleteHandshake(initiating bool, decoyContentLens []int,
+	net wire.BitcoinNet) error {
+
+	var receivedPrefix []byte
+	if initiating {
+		receivedPrefix = make([]byte, 0)
```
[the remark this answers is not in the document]
Crypt-iQ: I've pre-allocated now, but I think this case should actually be removed?

v2transport/transport.go
```diff
@@ -0,0 +1,570 @@
...
+	v1Prefix = append(v1Prefix, magic[:]...)
+	v1Prefix = append(v1Prefix, versionBytes...)
+
+	// ellswiftTheirs should be at least 16 bytes if receive succeeded, but
+	// just in case, check the size.
+	if len(ellswiftTheirs) < 16 {
+		return errInsufficientBytes
+	}
+
+	if !initiating && bytes.Equal(ellswiftTheirs[4:16], v1Prefix[4:16]) {
```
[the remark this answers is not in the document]
Crypt-iQ: I think this is fine as-is? Maybe I'm misunderstanding, but I think we'll disconnect the node if they have the wrong network magic? Ideally this would also have a test.

v2transport/transport.go
```diff
@@ -0,0 +1,570 @@
...
+	}
+
+	// Send garbage terminator.
+	p.Send(p.sendGarbageTerm)
+
+	// Optionally send decoy packets after garbage terminator.
+	aad := p.sentGarbage
+	for i := 0; i < len(decoyContentLens); i++ {
+		decoyContent := make([]byte, decoyContentLens[i])
+		encPacket, _, err := p.V2EncPacket(decoyContent, aad, false)
```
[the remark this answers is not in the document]
Crypt-iQ: Yup, nice catch. This code is currently unused but should use a test.

v2transport/transport.go
```diff
@@ -0,0 +1,570 @@
...
+		}
+
+		p.Send(encPacket)
+
+		// TODO: Does this cause anything to leak?
+		aad = nil
+	}
+
+	// Send version packet.
+	_, _, err = p.V2EncPacket(transportVersion, aad, false)
```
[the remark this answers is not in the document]
Crypt-iQ: This needs to be `aad` in case we haven't sent decoy packets which we don't currently

peer/peer.go
```diff
@@ -1067,8 +1074,32 @@ func (p *Peer) handlePongMsg(msg *wire.MsgPong) {
...
+		if err != nil {
+			fmt.Printf("failed to read packet: %v\n", err)
+			return nil, nil, err
+		}
+
+		msg, buf, err = wire.ReadV2MessageN(
+			plaintext, p.ProtocolVersion(), encoding,
+		)
+		n = len(plaintext)
+		atomic.AddUint64(&p.bytesReceived, uint64(n))
```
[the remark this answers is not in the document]
Crypt-iQ: good catch!

v2transport/transport.go
```diff
@@ -0,0 +1,570 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Skip garbage until encountering garbage terminator.
+	recvGarbage, err := p.Receive(16)
+	if err != nil {
+		return err
+	}
+	for i := 0; i < 4096; i++ {
```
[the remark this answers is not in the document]
Crypt-iQ: Added a comment, I think this should actually be `4095`

v2transport/transport.go
```diff
@@ -0,0 +1,570 @@
...
+	}
+
+	// Construct the packet's header based on whether or not the peer should
+	// ignore this packet (i.e. if this is a decoy packet).
+	ignoreNum := 0
+	if ignore {
+		ignoreNum = 1
+	}
+
+	ignoreNum <<= ignoreBitPos
```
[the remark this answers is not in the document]
Crypt-iQ: Done, let me know if this is what you had in mind

v2transport/transport.go
```diff
@@ -0,0 +1,570 @@
...
+// initiateV2Handshake generates our private key and sends our public key as
+// well as garbage data to our peer.
+func (p *Peer) InitiateV2Handshake(garbageLen int) error {
+	var err error
+	p.privkeyOurs, p.ellswiftOurs, err = btcec.EllswiftCreate()
+	if err != nil {
+		return err
+	}
+
+	p.sentGarbage = make([]byte, garbageLen)
```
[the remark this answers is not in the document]
Crypt-iQ: Yup, now asserting that it's <= 4095 bytes

v2transport/chacha.go
```diff
@@ -0,0 +1,173 @@
...
+	key      []byte
+	chunkCtr uint64
+	cipher   *chacha20.Cipher
+}
+
+// NewFSChaCha20 ...
+func NewFSChaCha20(initialKey []byte) (*FSChaCha20, error) {
+	var initialNonce [12]byte
+	numRekeys := 0
+	binary.LittleEndian.PutUint64(initialNonce[4:12], uint64(numRekeys))
```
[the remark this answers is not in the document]
Crypt-iQ: I think this was used in an earlier iteration of the code, removed.

v2transport/transport.go
```diff
@@ -0,0 +1,570 @@
...
+
+	// We cut off a byte when feeding to Crypt.
+	contentsLE := make([]byte, 4)
+	binary.LittleEndian.PutUint32(contentsLE, uint32(contentLen))
+	encContentsLen, err := p.sendL.Crypt(contentsLE[:lengthFieldLen])
+	if err != nil {
+		return nil, 0, err
+	}
+
+	encPacket := make([]byte, 0)
```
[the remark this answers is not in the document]
Crypt-iQ: Done

v2transport/chacha.go
```diff
@@ -0,0 +1,173 @@
...
+		cipher:   cipher,
+	}, nil
+}
+
+// Crypt ...
+func (f *FSChaCha20) Crypt(text []byte) ([]byte, error) {
+	// XOR the text with the keystream to get either the cipher or plaintext.
+	textLen := len(text)
+	dst := make([]byte, textLen)
+	f.cipher.XORKeyStream(dst, text)
```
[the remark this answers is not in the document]
Crypt-iQ: Yeah, we don't actually need `block_counter` here. I implemented it with `block_counter` and found that it actually served no use since the `XORKeyStream` handles the keystream bytes for us.

v2transport/chacha.go
```diff
@@ -0,0 +1,172 @@
...
+			fmt.Println(err)
+			return nil, err
+		}
+		f.cipher = cipher
+	}
+
+	return result, nil
+}
+
+// FSChaCha20 ...
```
Roasbeef, line 110: Missing `godoc` comments.

Roasbeef: Should mention the purpose of these wrappers. Eg: this one is used to encrypt the lengths of all packets.

v2transport/chacha.go
```diff
@@ -0,0 +1,172 @@
...
+		// chacha20 library doesn't supply us with the keystream's bytes
+		// directly.
+		var dummyXor [32]byte
+		var newKey [32]byte
+		f.cipher.XORKeyStream(newKey[:], dummyXor[:])
+		f.key = newKey[:]
+
+		var nonce [12]byte
+		numRekeys := f.chunkCtr / rekeyInterval
+		binary.LittleEndian.PutUint64(nonce[4:12], numRekeys)
```
Roasbeef, line 159: Hmm, I don't see a section of the BIP that matches this.

v2transport/transport.go
```diff
@@ -0,0 +1,625 @@
...
+	receivedPrefix []byte
+
+	// sendL is the cipher used to send encrypted packet lengths.
+	sendL *FSChaCha20
+
+	// sendP is the cipher used to send encrypted packets.
+	sendP *FSChaCha20Poly1305
+
+	// sendGarbageTerm is the garbage terminator that we send.
+	sendGarbageTerm []byte
```
Roasbeef, line 109: IIUC this is always 16 bytes? So can make this a fixed size array.

v2transport/transport.go
```diff
@@ -0,0 +1,625 @@
...
+		initiatorL:     make([]byte, 32),
+		initiatorP:     make([]byte, 32),
+		responderL:     make([]byte, 32),
+		responderP:     make([]byte, 32),
+		sessionID:      make([]byte, 32),
+	}
+}
+
+// createV2Ciphers constructs the packet-length and packet encryption ciphers.
+func (p *Peer) createV2Ciphers(ecdh_secret []byte, initiating bool,
```
Roasbeef, line 156: ecdh_secret -> ecdhSecret

v2transport/transport.go
```diff
@@ -0,0 +1,625 @@
...
+		p.recvP, err = NewFSChaCha20Poly1305(p.initiatorP)
+		if err != nil {
+			return err
+		}
+
+		p.recvGarbageTerm = initiatorGarbageTerminator
+	}
+
+	// TODO:
+	//     To achieve forward secrecy we must wipe the key material used to initialize the ciphers:
```
Roasbeef, line 272: I think in go we can mainly w/ best effort basis:
  * use something like `mlock` to prevent the secret data from swapping to disk 
  * make functions to zero out bytes either by doing each byte, or setting the pointer to an empty value
      * optionally force a GC at this point 

Can be wroth checking if the crypto stuff in the stdlib does anything fancy here.

Roasbeef: Looks like there's this issue: https://github.com/golang/go/issues/21865

v2transport/transport.go
```diff
@@ -0,0 +1,625 @@
...
+	// - golang analogue?
+
+	return nil
+}
+
+// initiateV2Handshake generates our private key and sends our public key as
+// well as garbage data to our peer.
+func (p *Peer) InitiateV2Handshake(garbageLen int) error {
+	var err error
+	p.privkeyOurs, p.ellswiftOurs, err = btcec.EllswiftCreate()
```
Roasbeef, line 283: Thought this was in a sub-package? If not, we should move it to one.

v2transport/transport.go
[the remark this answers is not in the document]
Roasbeef: AFAICT, it's a bit hidden in the pseudo code, it makes V1 prrefix which includes the network magic, see `complete_handshake`: 
> NETWORK_MAGIC = b'\xf9\xbe\xb4\xd9' # Mainnet network magic; differs on other networks.
> V1_PREFIX = NETWORK_MAGIC + b'version\x00\x00\x00\x00\x00'

>     if not initiating and ellswift_theirs[4:16] == V1_PREFIX[4:16]:
>         # Looks like a v1 peer from the wrong network.
>         disconnect(peer)

v2transport/transport.go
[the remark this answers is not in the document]
Roasbeef: This matches the logic in `complete_handshake` for the psuedo code in the BIP.

v2transport/transport.go
```diff
@@ -0,0 +1,625 @@
...
+		}
+		recvGarbage = append(recvGarbage, recvData...)
+	}
+
+	// Return an error if the garbage terminator was not received after 4 KiB.
+	return errGarbageTermNotRecv
+}
+
+// V2EncPacket takes the contents and aad and returns a ciphertext.
+func (p *Peer) V2EncPacket(contents []byte, aad []byte, ignore bool) ([]byte,
```
Roasbeef, line 498: Would be good to create some basic benchmarks for:
  * handshake creation
  * handshake complete 
  * v2 encrypt 
  * v2 decrypt 

Just so we can get an ideal re the current cpu/memory utilization.

peer/peer.go
```diff
@@ -1065,10 +1072,42 @@ func (p *Peer) handlePongMsg(msg *wire.MsgPong) {
...
+		if err != nil {
+			fmt.Printf("failed to read packet: %v\n", err)
+			return nil, nil, err
+		}
+
+		msg, buf, err = wire.ReadV2MessageN(
+			plaintext, p.ProtocolVersion(), encoding,
+		)
+		n = len(plaintext)
+	} else if partial {
```
Roasbeef, line 1100: ?

peer/peer.go
```diff
@@ -2215,7 +2270,39 @@ func (p *Peer) waitToFinishNegotiation(pver uint32) error {
...
 //  6. If remote peer sent sendaddrv2 above, wait until receipt of verack.
 func (p *Peer) negotiateInboundProtocol() error {
-	if err := p.readRemoteVersionMsg(); err != nil {
+	// We may be anticipating a v2 connection, but if the initiating peer sends
+	// us a v1 version message, we need to note down that we've downgraded the
+	// connection internally.
+	downgradedConn := false
+
+	if p.cfg.UsingV2Conn {
+		// TODO: Change garbageLen to be random.
```
Roasbeef, line 2279: So just a random 12 byte int?

btcec/ellswift.go
```diff
@@ -0,0 +1,401 @@
...
+func init() {
+	c.SetByteSlice(cBytes[:])
+}
+
+// XSwiftEC() takes two field elements (u, t) and gives us an x-coordinate that
+// is on the secp256k1 curve. This is used to take an ElligatorSwift-encoded
+// public key (u, t) and return the point on the curve it maps to. This
+// function returns an error if there is no valid x-coordinate.
+//
+// TODO: Rewrite these to avoid new(FieldVal).Add(...) usage?
```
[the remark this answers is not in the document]
Crypt-iQ: I forget what I meant with this comment, I think I wanted to see if it was possible to initialize the `FieldVal` with the value instead of calling `new` followed by `Add`?

btcec/ellswift.go
```diff
@@ -0,0 +1,401 @@
...
+// NOTE: u, t MUST be normalized. The result x is normalized.
+func XSwiftEC(u, t *FieldVal) (*FieldVal, error) {
```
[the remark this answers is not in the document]
Crypt-iQ: I double-checked and it looks weird, but this is the proper code for `XSwiftEC` from the BIP. Just curious, what made you think that they were combined here?

btcec/ellswift.go
```diff
@@ -0,0 +1,401 @@
...
+// TODO: Rewrite these to avoid new(FieldVal).Add(...) usage?
+// NOTE: u, x MUST be normalized. The result `t` is normalized.
+func XSwiftECInv(u, x *FieldVal, caseNum int) *FieldVal {
+	v := new(FieldVal)
+	s := new(FieldVal)
+	twoInv := new(FieldVal).AddInt(2).Inverse()
+
+	if caseNum&2 == 0 {
+		// If lift_x(-x-u) succeeds, return None
+		if _, found := liftX(new(FieldVal).Add(x).Add(u).Negate(2)); found {
```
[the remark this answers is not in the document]
Crypt-iQ: The `2` is the magnitude of the value we are negating -- I think this is correct based on my reading of the `Negate` godoc.

btcec/ellswift.go
```diff
@@ -0,0 +1,401 @@
...
+	}
+
+	// Should have found a square above.
+	return nil, fmt.Errorf("no calculated x-values were square")
+}
+
+// isXOnCurve returns true if there is a corresponding y-value for the passed
+// x-coordinate.
+func isXOnCurve(x *secp256k1.FieldVal) bool {
+	y := new(FieldVal).Add(x).Square().Mul(x).AddInt(7)
```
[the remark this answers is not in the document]
Crypt-iQ: I thought so initially, but I don't think so in this case? From my reading of the godocs on magnitude in the various `FieldVal` functions, `isOnCurve` needs to normalize because `Equals` is being called which requires both inputs to be normalized. With this function `isXOnCurve`, I don't think we need to normalize since none of the used `FieldVal` functions require us to normalize unless the input magnitude is too large.

btcec/ellswift.go
```diff
@@ -0,0 +1,401 @@
...
+		// If we've reached here, the point does not exist on the curve.
+		return nil, false
+	}
+
+	if !y.Normalize().IsOdd() {
+		return NewPublicKey(x, y), true
+	}
+
+	// Negate y if it's odd.
+	if !y.Negate(1).Normalize().IsOdd() {
```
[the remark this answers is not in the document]
Crypt-iQ: Ah yeah, this does rely on mutation, @kcalvinalvin pointed out that we can just use `DecompressY` instead. I'll change the code to reflect that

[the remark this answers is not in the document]
Crypt-iQ: Hmm, on second thought. IIUC `DecompressY` also mutates `resultY` and the reason we want `liftX` is that it returns `PublicKey`. Should I still change this code to use `DecompressY`?

v2transport/chacha.go
Crypt-iQ: Done

v2transport/chacha.go
Crypt-iQ: This is an implementation detail specific to the golang `chacha20` library. We don't get the keystream bytes directly and instead have to use `XORKeyStream` and xor it with empty `[N]byte` to output the keystream bytes. The only time we need to change the nonce here is during each re-keying interval as o/w it stays the same.

v2transport/transport.go
Crypt-iQ: Yup

v2transport/transport.go
Crypt-iQ: Will move

peer/peer.go
Crypt-iQ: This is used for when we expect a v2 inbound connection but instead receive a v1 version message. In this case, we've already read some of the v1 version header and need to process the rest of it. Without this, the v2->v1 inbound downgrade doesn't work properly.

peer/peer.go
Crypt-iQ: I'm not sure why I chose it to be `1` here, it can instead be `0` I think. The garbage here is limited to `4095` bytes.

Roasbeef:
We should likely have a config to disabled the new behavior as well.

peer/peer.go
Roasbeef: Hmm, ok will examine this relevant area of the BIP further. 

Can we add a comment here explaining the rationale here a bit?

server.go
```diff
@@ -2209,6 +2209,19 @@ func (s *server) outboundPeerConnected(c *connmgr.ConnReq, conn net.Conn) {
 // done along with other performing other desirable cleanup.
 func (s *server) peerDoneHandler(sp *serverPeer) {
 	sp.WaitForDisconnect()
+
+	// If this is an outbound peer and the shouldDowngradeToV1 bool is set on
+	// the underlying Peer, trigger a reconnect using the OG v1 connection
+	// scheme.
+	if !sp.Inbound() && sp.Peer.ShouldDowngradeToV1() {
+		// TODO: Determine _how_ to trigger a reconnect that does not use v2
```
Roasbeef, line 2217: I think we can annotate `serverPeer` or wrap it in a struct that adds extra context. 

Then when we go to handle the done peer: https://github.com/btcsuite/btcd/blob/85de89db5b5dd6399e7ff8571ebbf6baece7606b/server.go#L1832-L1842

We annotate a new map (one route) that tracks downgrade attempts that need to be achieved. When we connect to the peer, and go to make a new peer struct, we check that map, and based on that value we set the v2 option in the config.

Crypt-iQ:
Pushed up a patch that implements the reconnection logic. It is not currently working on mainnet, I believe it is because I am catching `EOF` instead of some of the other read errors like "use of closed network connection".

Crypt-iQ:
Latest set of changes allow the outbound v2->v1 connection downgrade to work properly. Have run this on mainnet and btcd is able to reconnect when an outbound v2 fails.

peer/peer.go
```diff
@@ -293,6 +293,7 @@ type Config struct {
 
 	// UsingV2Conn is defined if and only if we accept and attempt to make
 	// v2 connections.
+	// TODO: Modify this so that we can downgrade certain peers.
```
Roasbeef: WIP TODO?

peer/peer.go
```diff
@@ -2345,13 +2347,19 @@ func (p *Peer) negotiateOutboundProtocol() error {
 		// and reconnect using a v1 connection. This is the logic that
 		// bitcoind uses.
 		// TODO: random value for garbage len?
-		if err := p.V2Transport.InitiateV2Handshake(1); err != nil {
+		if err := p.V2Transport.InitiateV2Handshake(0); err != nil {
```
Roasbeef, line 2349: Let's make this a constant somewhere so this section is more readable.

Roasbeef: Also why this change?

server.go
```diff
@@ -255,6 +255,12 @@ type server struct {
 	// agentWhitelist is a list of whitelisted user agent substrings, no
 	// whitelisting will be applied if the list is empty or nil.
 	agentWhitelist []string
+
+	// pendingReconnects is a map of peer addresses that have failed when
+	// connecting outbound.
+	// TODO: Memory leak if outbound target reached and non-persistent peer.
```
Roasbeef, line 261: Let's make this into an issue instead? Also what stops of from periodically garbage collecting this? IIUC this map should never have more than outbound target entries

Roasbeef: Can also prob just use `sync.Mutex` to not have to worry as much re critical areas.

server.go
```diff
@@ -281,6 +287,9 @@ type serverPeer struct {
 	// The following chans are used to sync blockmanager and server.
 	txProcessed    chan struct{}
 	blockProcessed chan struct{}
+
+	reconnectMtx    sync.RWMutex
+	shouldReconnect bool
```
Roasbeef, line 292: Why do we need this _and_ the field we added in the prior commit? `ShouldDowngradeToV1`

server.go
```diff
@@ -2187,7 +2218,24 @@ func (s *server) inboundPeerConnected(conn net.Conn) {
 // manager of the attempt.
 func (s *server) outboundPeerConnected(c *connmgr.ConnReq, conn net.Conn) {
 	sp := newServerPeer(s, c.Permanent)
-	p, err := peer.NewOutboundPeer(newPeerConfig(sp), c.Addr.String())
+
+	// Just an alias.
+	peerAddr := c.Addr.String()
```
Roasbeef, line 2224: Perhaps we should normalize a bit here and _only_ track the IP? IIUC the port would be included here.

server.go
```diff
@@ -2214,11 +2262,17 @@ func (s *server) peerDoneHandler(sp *serverPeer) {
...
-		//       connect to.
+		// Store the peer's address in a reconnect map that will store the
+		// set of connections to downgrade to v1.
+		s.pendingReconnectsMtx.Lock()
+		s.pendingReconnects[sp.Addr()] = struct{}{}
+		s.pendingReconnectsMtx.Unlock()
+
+		// Flip a bool in serverPeer to mark that we should reconnect to
+		// this peer.
+		sp.SetShouldReconnect()
```
Roasbeef, line 2274: See comment above, re just exposing the internal peer method directly.

connmgr/connmanager.go
```diff
@@ -463,14 +470,15 @@ func (cm *ConnManager) Connect(c *ConnReq) {
 
 // Disconnect disconnects the connection corresponding to the given connection
 // id. If permanent, the connection will be retried with an increasing backoff
-// duration.
-func (cm *ConnManager) Disconnect(id uint64) {
+// duration. If the reconnect bool is passed, we will attempt to reconnect
+// after the disconnect.
+func (cm *ConnManager) Disconnect(id uint64, reconnect bool) {
```
Roasbeef, line 475: Is this needed? I think we just actually need to call `Remove`, then we can manually queue the reconnection attempt ourselves.

Roasbeef (changes requested):

peer/peer.go
Roasbeef: Swapped in to use a proper random garbage length.

Roasbeef (approved):
LGTM 🎄

Pushed up a few commits to help take this across the finish line.

peer/peer.go
```diff
@@ -2282,16 +2281,19 @@ func (p *Peer) negotiateInboundProtocol() error {
...
+		case errors.Is(err, v2transport.ErrUseV1Protocol):
 			log.Infof("Inbound v2 connection attempt from %s "+
 				"downgraded to v1 (peer sent v1 version "+
 				"message)", p.addr)
 
 			p.cfg.UsingV2Conn = false
 			downgradedConn = true
 
-		case false:
+		case err != nil:
```
Crypt-iQ, line 2293: 👍

Crypt-iQ (commented):
Reviewed + tested and it works, ACK 5ab25f3f2bd418fba09325c0f496f56aca03e9a3

I think deployment and cross impl fuzzing will shake out any lingering bugs

Roasbeef:
Have been running this without issue for the past week. Confirmed the downgrade logic works for both inbound and outbound peers. Interop with the only other implementation that supports this BIP (`bitcoind`) also is operational. 

Will merge, then create follow up PRs to update the module tags, before a final PR that'll remove the replace directives from the top level module.

PR #2269, opened 2024-10-27
Fix non-root hosts failing on resolving DNS

same-id:
A fix for a bug introduced by #2168

Previously, config.Host worked in the following way:
1. Documented as supporting ip addresses only
2. In fact supported "host/path" syntax
3. Did not support "scheme" prefixes, i.e. https://

Not sure this is the desired approach, probably the best thing would have been to extend config to contain "Scheme" and "Path" fields as well.

However, this was the way it worked.

1. Now Host can contain scheme prefixes "unix://..."
2. Host can no longer contain ".../path"

This PR solves this behavior while maintaining support of the "unix://" flow as well.

For some reason, "scheme" is named "network" in #2168 - I did not change that.


same-id:
Fixed

rpcclient/infrastructure.go
```diff
@@ -1733,48 +1733,35 @@ func (c *Client) Send() error {
...
+		return net.ResolveUnixAddr("unixpacket", after)
+	}
+	if strings.Contains(strAddress, "://") { // Not supporting :// anywhere in the host or path
+		return nil, fmt.Errorf("unsupported protocol in address: %s", strAddress)
+	}
+	u, err := url.Parse("dummy://" + strAddress)
+	if err != nil {
+		return nil, err
 	}
+	return net.ResolveTCPAddr("tcp", verifyPort(u.Host))
```
guggero, line 1764: I don't actually understand how this fixes your issue. If you say you're using `somehost/path`, then `/path` would be removed here. Which makes me think that using `.../path` is wrong anyway?
What's the use case here? Is the RPC server behind a reverse proxy that routes traffic according to paths?

guggero (commented):
See inline comments, I'm not sure this actually fixes your issue?

rpcclient/infrastructure.go
same-id: When we previously had path here - net.ResolveTCPAddr failed, now it succeeds, since it only needs the host.

The path is then later actually used in:

```
var url string
	switch parsedAddr.Network() {
	case "unix", "unixpacket":
		// Using a placeholder URL because a non-empty URL is required.
		// The Unix domain socket is specified in the DialContext.
		url = protocol + "://unix"
	default:
		url = protocol + "://" + c.config.Host
	}
...
	for i := 0; i < tries; i++ {
...
		httpReq, err = http.NewRequest("POST", url, bodyReader)
...
```

same-id: We could have restricted path anyway in `config.Host`, and perhaps requested that it would be used in `config.Endpoint` which is actually used as path in websocket flows - but that would be breaking behavior.

Plus now `Host` starts to contain prefixes such as `unix://` where previously (and still now) `http://` was not allowed.

The cleanest way to do it obviously IMO is to have:
```
type ConnConfig struct {

	Scheme string

	Host string

	Endpoint string

}
`

And use it:

```
Scheme: "unix"
Host: "host-no-path:123"
Endpoint: "/path/to/call"
```

And throw all the URL parsing logic away.

File: rpcclient/infrastructure.go [hunk 2]
```diff
@@ -1733,48 +1733,35 @@ func (c *Client) Send() error {
...
-		// We'll now possibly use the local host short circuit
-		// or parse out an all interfaces listen.
-		addrWithPort := verifyPort(strAddress)
-
-		// Otherwise, we'll attempt to resolve the host.
-		return net.ResolveTCPAddr("tcp", addrWithPort)
+	if after, ok := cutPrefix(strAddress, "unixpacket://"); ok {
+		return net.ResolveUnixAddr("unixpacket", after)
+	}
+	if strings.Contains(strAddress, "://") { // Not supporting :// anywhere in the host or path
```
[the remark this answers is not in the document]
same-id: Fixed

rpcclient/infrastructure.go
```diff
@@ -1733,48 +1733,35 @@ func (c *Client) Send() error {
...
-	if strings.Contains(strAddress, "://") {
-		parts := strings.Split(strAddress, "://")
-		parsedNetwork, parsedAddr = parts[0], parts[1]
-	} else if strings.Contains(strAddress, ":") {
-		parts := strings.Split(strAddress, ":")
-		parsedNetwork = parts[0]
-		parsedAddr = strings.Join(parts[1:], ":")
-	} else {
-		parsedAddr = strAddress
+	// Addresses can either be in unix://address, unixpacket://address URL format
```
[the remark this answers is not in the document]
same-id: Fixed

rpcclient/infrastructure.go
```diff
@@ -1733,48 +1733,35 @@ func (c *Client) Send() error {
 	return nil
 }
 
+// Copied from newer Go version
```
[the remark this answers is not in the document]
same-id: Fixed

rpcclient/infrastructure.go
guggero: Yeah, I guess using the endpoint would've been the cleanest option. But that would be breaking to users that previously specified the path in the `Host`.

guggero (commented):
I think we should clean this up a bit more and add some basic tests.

Here's my proposal as a diff (feel free to just use in your commit, no attribution required):
[proposal.diff.txt](https://github.com/user-attachments/files/17666449/proposal.diff.txt)

rpcclient/infrastructure.go
same-id: However, we now also added the usage of `unix://` in the `Host` part - which will also break in the future if we introduce `Scheme`.

Might consider breaking the existing *new* unix socket users in favor of not breaking them later on once this gets more popular by adding a `Scheme`.

Regarding breaking existing `Host` users - that's why I wrote my PR this way, however it's tempting to breaking them as well since the documentation of the module clearly says host only - but I don't mind

guggero: Let's not do a breaking change now but instead fix the issue introduced by #2168. See my diff/proposal below.

same-id: Ok, so I added your cleanups and added some tests (hopefully they do not need extra cleanups).

guggero (approved):
Thanks, LGTM :tada:

same-id:
> Thanks, LGTM 🎉

Thanks for the review, still pending approval from another reviewer: @Crypt-iQ

same-id:
@guggero - just making sure we are really waiting for an additional reviewer here

sam-utila:
Hi, @guggero 

Gentle ping

This is still stuck on review

same-id:
Hi, @guggero

Gentle ping

This is still stuck on review

guggero:
cc @yyforyongyu

yyforyongyu (approved):
Thanks for the cleanup!

rpcclient/infrastructure.go
```diff
@@ -1733,53 +1749,48 @@ func (c *Client) Send() error {
 	return nil
 }
 
+// cutPrefix returns s without the provided leading prefix string
+// and reports whether it found the prefix.
+// If s doesn't start with prefix, CutPrefix returns s, false.
```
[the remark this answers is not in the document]
same-id: done

rpcclient/infrastructure.go
```diff
@@ -1733,53 +1749,48 @@ func (c *Client) Send() error {
...
+// If prefix is the empty string, CutPrefix returns s, true.
```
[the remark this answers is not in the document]
same-id: done

same-id:
Still blocked on merge

yyforyongyu (approved):

yyforyongyu (approved):
LGTM👍

same-id:
Thanks!

PR #2273, opened 2024-11-08
rpcclient: safe read and write to batch

ynewmann:
This PR tries to make batchList concurrent-safe.

The `Send()` could panic at the line 1730 if c.removeRequest  returned `nil`. It could return `nil` if we try to call `Send()` concurrently.
https://github.com/btcsuite/btcd/blob/684d64ad74fed203fb846c032f2b55b3e3c36734/rpcclient/infrastructure.go#L1708-L1730

After the quick look I've noticed that there are places where the batchList is update unsafely. So, the simplest fix is just to add locks, but I'm afraid if it could break something else, because I'm not very familiar with the library.

### DIFF
--- a/rpcclient/infrastructure.go
+++ b/rpcclient/infrastructure.go
@@ -74,6 +74,9 @@ var (
 	// client having already connected to the RPC server.
 	ErrClientAlreadyConnected = errors.New("websocket client has already " +
 		"connected")
+
+	// ErrEmptyBatch is an error to describe that there is nothing to send.
+	ErrEmptyBatch = errors.New("batch is empty")
 )
 
 const (
@@ -151,6 +154,7 @@ type Client struct {
 
 	// whether or not to batch requests, false unless changed by Batch()
 	batch     bool
+	batchLock sync.Mutex
 	batchList *list.List
 
 	// retryCount holds the number of times the client has tried to
@@ -214,7 +218,10 @@ func (c *Client) addRequest(jReq *jsonRequest) error {
 		element := c.requestList.PushBack(jReq)
 		c.requestMap[jReq.id] = element
 	} else {
+		c.batchLock.Lock()
 		element := c.batchList.PushBack(jReq)
+		c.batchLock.Unlock()
+
 		c.requestMap[jReq.id] = element
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


Roasbeef:
Approved CI.

rpcclient/infrastructure.go
```diff
@@ -1656,7 +1665,15 @@ func (c *Client) BackendVersion() (BackendVersion, error) {
 	return c.backendVersion, nil
 }
 
-func (c *Client) sendAsync() FutureGetBulkResult {
+func (c *Client) sendAsync() (FutureGetBulkResult, error) {
+	c.batchLock.Lock()
+	defer c.batchLock.Unlock()
+
+	// if batchlist is empty, there's nothing to send
```
guggero, line 1672: nit (pre-existing from copy/paste below): Please use full sentences for comments (with initial capitalization and punctuation). Also, it should either say "batch list" or "batchList".

rpcclient/infrastructure.go
```diff
@@ -1678,25 +1695,28 @@ func (c *Client) sendAsync() FutureGetBulkResult {
...
-	batchResp, err := c.sendAsync().Receive()
+	batchResp, err := future.Receive()
 	if err != nil {
 		// Clear batchlist in case of an error.
 		//
 		// TODO(yy): need to double check to make sure there's no
 		// concurrent access to this batch list, otherwise we may miss
 		// some batched requests.
+
+		c.batchLock.Lock()
```
guggero, line 1717: nit: can now remove the TODO above, as that's what the PR does.

guggero (commented):
Thanks for the fix! Just two minor nits.

ynewmann:
> Thanks for the fix! Just two minor nits.

One more nit here [b3d9947](https://github.com/btcsuite/btcd/pull/2273/commits/b3d9947a3706c83372ca94007b69c13eec20439e)

guggero (approved):
Can you squash the commits into one please?
Then this is good to go.

guggero (approved):

yyforyongyu:
Could you do a rebase on master? I'm trying to test it in `lnd` but the current branch doesn't include some recent fixes that were pushed to master.

ynewmann:
> Could you do a rebase on master? I'm trying to test it in `lnd` but the current branch doesn't include some recent fixes that were pushed to master.

done

yyforyongyu (approved):
LGTM 🚢

PR #2274, opened 2024-11-10
feat: Add testnet4 chain support

bullet-tooth:
This PR adds support of the Bitcoin testnet version 4 chain (testnet3 is in a way of deprecation).

Reference: https://bips.dev/94/

Testnet4 genesis block: https://mempool.space/testnet4/block/00000000da84f2bafbbc53dee25a72ae507ff4914b867c565be350b0da8bf043

_Also, I have intention to add testnet4 support to btcwallet after this PR_


Roasbeef:
Looks like we have another here: https://github.com/btcsuite/btcd/pull/2275

Neither that one, nor this one, have any of the new difficulty adjustment algorithm changes.

Roasbeef:
@bullet-tooth see the BIP: https://bips.dev/94/

guggero:
If you clean up the PR into a review friendly commit structure, I'll take a look.
Here are some docs to go by: https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md
and https://github.com/lightningnetwork/lnd/blob/master/docs/code_formatting_rules.md.
Those aren't strictly enforced in this project but definitely nice to have if you want review from the lnd team.

blockchain/difficulty.go
```diff
@@ -191,12 +191,22 @@ func calcNextRequiredDifficulty(lastNode HeaderCtx, newBlockTime time.Time,
 		adjustedTimespan = c.MaxRetargetTimespan()
 	}
 
+	var oldTarget *big.Int
+	// Special difficulty rule for Testnet4
```
guggero, line 195: Please use full sentences, including punctuation, for comments. Also, if a comment isn't at the beginning of a block, it's nice to have an empty line before it to have some vertical breathing room.
Finally, `else` is kind of discouraged in Go if it can be written in a different way, as it often leads to less readable code.
So I'd write this the following way:

```go

	// Special difficulty rule for Testnet4
	oldTarget := CompactToBig(lastNode.Bits())
	if c.ChainParams().EnforceBIP94 {
		// Here we use the first block of the difficulty period. This way
		// the real difficulty is always preserved in the first block as
		// it is not allowed to use the min-difficulty exception.
		oldTarget = CompactToBig(firstNode.Bits())
	}
```

Of course the assumption here is that `CompactToBig(lastNode.Bits())` isn't computationally expensive to wanting to avoid calling it twice in the worst case.

blockchain/error.go
```diff
@@ -220,6 +220,8 @@ const (
 	// current chain tip. This is not a block validation rule, but is required
 	// for block proposals submitted via getblocktemplate RPC.
 	ErrPrevBlockNotBest
+
+	ErrTimewarpAttack
```
guggero, line 224: nit: missing Godoc comment.

blockchain/validate.go
```diff
@@ -85,7 +90,7 @@ func ShouldHaveSerializedBlockHeight(header *wire.BlockHeader) bool {
 
 // IsCoinBaseTx determines whether or not a transaction is a coinbase.  A coinbase
 // is a special transaction created by miners that has no inputs.  This is
-// represented in the block chain by a transaction with a single input that has
+// represented in the blockchain by a transaction with a single input that has
```
guggero, line 93: Please revert these. This specifically refer to the chain of blocks, not the package name. IMO "Blockchain" is a marketing term used by people trying to sell you something that isn't Bitcoin.

blockchain/validate.go
```diff
@@ -710,11 +719,22 @@ func CheckBlockHeaderContext(header *wire.BlockHeader, prevNode HeaderCtx,
...
+			if blockHeight%c.BlocksPerRetarget() == 0 {
+				prevBlockTimestamp := time.Unix(prevNode.Timestamp(), 0)
+				if header.Timestamp.Before(prevBlockTimestamp.Add(-maxTimeWarp)) {
+					str := "block's timestamp %v is too early on diff adjustment block %v"
+					str = fmt.Sprintf(str, header.Timestamp, prevBlockTimestamp)
+					return ruleError(ErrTimewarpAttack, str)
+				}
+			}
+		}
+	}
```
guggero, line 737: nit: line length. Also, sounds like this whole block could be extracted into a standalone function. Something like `checkBip94()`.

btcd.go
```diff
@@ -64,6 +64,11 @@ func btcdMain(serverChan chan<- *server) error {
 	// Show version at startup.
 	btcdLog.Infof("Version %s", version())
 
+	if cfg.TestNet3 {
```
guggero, line 67: Not sure I agree with that whole block.

chaincfg/genesis_test.go
```diff
@@ -6,6 +6,7 @@ package chaincfg
 
 import (
 	"bytes"
+	"github.com/stretchr/testify/require"
```
guggero, line 9: nit: please sort this with the block below where non-std-lib imports go.

chaincfg/genesis_test.go
```diff
@@ -91,6 +92,34 @@ func TestTestNet3GenesisBlock(t *testing.T) {
 	}
 }
 
+// TestTestNet4GenesisBlock tests the genesis block of the test network (version
+// 4) for validity by checking the encoded bytes and hashes.
+func TestTestNet4GenesisBlock(t *testing.T) {
+	// Encode the genesis block to raw bytes.
+	var buf bytes.Buffer
+	err := TestNet4Params.GenesisBlock.Serialize(&buf)
+	if err != nil {
```
guggero, line 101: Should use `require.NoError()` and `require.Equal()` for new test code.

chaincfg/params.go
```diff
@@ -189,6 +189,8 @@ type Params struct {
 	// regtest like networks.
 	PoWNoRetargeting bool
 
+	EnforceBIP94 bool
```
guggero, line 192: nit: Godoc missing.

chaincfg/params.go
```diff
@@ -297,6 +299,7 @@ var MainNetParams = Params{
 	GenesisHash:              &genesisHash,
 	PowLimit:                 mainPowLimit,
 	PowLimitBits:             0x1d00ffff,
+	EnforceBIP94:             false,
```
guggero, line 302: IMO we don't need to add `false` to the existing params as that's the default value for a boolean.

chaincfg/params.go
```diff
@@ -673,6 +678,87 @@ var TestNet3Params = Params{
 	HDCoinType: 1,
 }
 
+var TestNet4Params = Params{
```
guggero, line 681: nit: missing Godoc params.

chaincfg/params.go
```diff
@@ -673,6 +678,87 @@ var TestNet3Params = Params{
...
+	Name:        "testnet4",
+	Net:         wire.TestNet4,
+	DefaultPort: "48333",
+	DNSSeeds: []DNSSeed{
+		{"seed.testnet4.bitcoin.sprovoost.nl", true},
+		{"seed.testnet4.wiz.biz", true},
+	},
+
+	//// Chain parameters
```
guggero, line 690: nit: double comment.

cmd/addblock/config.go
```diff
@@ -41,7 +41,8 @@ type config struct {
 	Progress       int    `short:"p" long:"progress" description:"Show a progress message each time this number of seconds have passed -- Use 0 to disable progress announcements"`
 	RegressionTest bool   `long:"regtest" description:"Use the regression test network"`
 	SimNet         bool   `long:"simnet" description:"Use the simulation test network"`
-	TestNet3       bool   `long:"testnet" description:"Use the test network"`
+	TestNet3       bool   `long:"testnet" description:"Use the test network (version 3). Support for testnet3 is deprecated. Consider moving to testnet4 now by using -testnet4"`
```
guggero, line 44: nit: remove deprecation comment. Should be out of scope for this PR.

cmd/addblock/config.go
```diff
@@ -79,6 +80,7 @@ func netName(chainParams *chaincfg.Params) string {
 	switch chainParams.Net {
 	case wire.TestNet3:
 		return "testnet"
+	// TODO do we need to use "testnet" dir for testnet4 as well?
```
guggero, line 83: We should use a distinct directory, otherwise you couldn't run both on the same computer.

cmd/btcctl/config.go
```diff
@@ -106,7 +106,8 @@ type config struct {
 	RPCUser        string `short:"u" long:"rpcuser" description:"RPC username"`
 	SimNet         bool   `long:"simnet" description:"Connect to the simulation test network"`
 	TLSSkipVerify  bool   `long:"skipverify" description:"Do not verify tls certificates (not recommended!)"`
-	TestNet3       bool   `long:"testnet" description:"Connect to testnet"`
+	TestNet3       bool   `long:"testnet" description:"Connect to testnet (version 3). Support for testnet3 is deprecated. Consider moving to testnet4 now by using -testnet4"`
```
guggero, line 109: Remove deprecation comment.

cmd/findcheckpoint/config.go
```diff
@@ -41,7 +41,8 @@ type config struct {
 	NumCandidates  int    `short:"n" long:"numcandidates" description:"Max num of checkpoint candidates to show {1-20}"`
 	RegressionTest bool   `long:"regtest" description:"Use the regression test network"`
 	SimNet         bool   `long:"simnet" description:"Use the simulation test network"`
-	TestNet3       bool   `long:"testnet" description:"Use the test network"`
+	TestNet3       bool   `long:"testnet" description:"Use the test network (version 3). Support for testnet3 is deprecated. Consider moving to testnet4 now by using -testnet4"`
```
guggero, line 44: Same here re deprecation.

config.go
```diff
@@ -169,7 +169,8 @@ type config struct {
 	SigNet               bool          `long:"signet" description:"Use the signet test network"`
 	SigNetChallenge      string        `long:"signetchallenge" description:"Connect to a custom signet network defined by this challenge instead of using the global default signet test network -- Can be specified multiple times"`
 	SigNetSeedNode       []string      `long:"signetseednode" description:"Specify a seed node for the signet network instead of using the global default signet network seed nodes"`
-	TestNet3             bool          `long:"testnet" description:"Use the test network"`
+	TestNet3             bool          `long:"testnet" description:"Use the test network (version 3). Support for testnet3 is deprecated. Consider moving to testnet4 now by using -testnet4"`
```
guggero, line 172: And here.

peer/peer.go
```diff
@@ -2377,6 +2377,7 @@ func newPeerBase(origCfg *Config, inbound bool) *Peer {
 
 	// Set the chain parameters to testnet if the caller did not specify any.
 	if cfg.ChainParams == nil {
+		// TODO: change to testnet4 because testnet3 will be deprecated soon
```
guggero, line 2380: Disagree.

guggero (changes requested):
Thanks for the update.
Still requires quite a bit of formal cleanup before more review time should be put in.

With "review friendly commit structure" I didn't mean squashing everything into a single commit.
Suggestion for commit order:
 - Any comment/typo fixes.
 - Any changes to non-Go files (e.g. `.gitignore`)
 - Add new parameter to `chaincfg` including unit tests.
 - Add `EnforceBip94` flag to parameters.
 - Add changes to `blockchain` package.
 - Add changes to `cmd` package.

humb1t:
Great work @bullet-tooth - can't wait to see this PR being merged.

PR #2276, opened 2024-11-21
wire: add bitcoin network magic for default SigNet

starius:
Added constant wire.SigNet, added it to stringer for BitcoinNet type.

Added a test in chaincfg to compare calculated magic value with the constant.


Roasbeef:
Approved CI.

wire/protocol.go
```diff
@@ -183,6 +183,9 @@ const (
 	// TestNet3 represents the test network (version 3).
 	TestNet3 BitcoinNet = 0x0709110b
 
+	// SigNet represents the signet (default).
+	SigNet BitcoinNet = 0x40CF030A
```
Roasbeef, line 187: Is this documented anywhere? Doesn't seem to be in the OG BIP.

Roasbeef (changes requested):

wire/protocol.go
starius: I copied the constant from https://en.bitcoin.it/wiki/Protocol_documentation#Common_structures
The constant is equal to `chaincfg.SigNetParams.Net`, this is checked in new test TestSigNetMagic.

guggero: It should also be equal to `CustomSignetParams(DefaultSignetChallenge, nil).Net`, or:

```
	challengeLength := byte(len(DefaultSignetChallenge))
	hashDouble := chainhash.DoubleHashB(
		append([]byte{challengeLength}, challenge...),
	)
	net := binary.LittleEndian.Uint32(hashDouble[0:4])
```

Perhaps we can add a simple test case for it?

wire/protocol.go
```diff
@@ -183,6 +183,9 @@ const (
 	// TestNet3 represents the test network (version 3).
 	TestNet3 BitcoinNet = 0x0709110b
 
+	// SigNet represents the signet (default).
```
guggero, line 186: I know this is probably the way it's called in most places (`signet (default)`), though I think we could make it a bit more obvious that this is only one of many possible signets, we could add something like: `// SigNet represents the public default SigNet. For custom signets, see `CustomSignetParams`.

wire/protocol.go
guggero: Ah, I guess that's what we're doing with `TestSigNetMagic`, so never mind.

guggero (approved):
Just one clarification request, otherwise LGTM :tada:

wire/protocol.go
starius: Thanks! Fixed.

Roasbeef (approved):
LGTM 🦀

PR #2284, opened 2024-12-03
txscript: add more detail to invalid tapscript merkle proof error

Roasbeef:
In this commit, we add more detail to the invalid tapscript merkle proof error. Before this commit, the error was blank, making such a case hard to debug. We'll now log the expected witness program, what we derived, and also the passed in tapscript root.

### DIFF
--- a/txscript/taproot_test.go
+++ b/txscript/taproot_test.go
@@ -293,13 +293,15 @@ func TestTapscriptCommitmentVerification(t *testing.T) {
 	// make from 0 to 1 leaf
 	// ensure verifies properly
 	testCases := []struct {
+		treeMutateFunc func(*IndexedTapScriptTree)
+
+		ctrlBlockMutateFunc func(*ControlBlock)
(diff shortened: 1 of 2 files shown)


kcalvinalvin:
ACK a155d466fc4fced5d34d257cb6bc6709a0676520

nit: Go is telling me that prand.Read on line 360 is deprecated. Maybe just use `crypto/rand`?

txscript/taproot_test.go
```diff
@@ -391,6 +396,12 @@ func TestTapscriptCommitmentVerification(t *testing.T) {
 						"valid=%v, got valid=%v", testCase.valid,
 						valid)
 				}
+
+				if !valid {
+					if !IsErrorCode(err, testCase.expectedErr) {
+						t.Fatalf("expected error code %v, got %v", testCase.expectedErr, err)
```
guggero: nit: lll

PR #2287, opened 2024-12-10
peer: make peer meet query.Peer interface

kcalvinalvin:
query.Peer is used for downloading blocks out of order during headers first download.  Methods SubscribeRecvMsg() and OnDisconnect() are added to abide by the interface.

This PR is part of #2226 and is split here so that it's easier to review.

### DIFF
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -494,6 +494,11 @@ type Peer struct {
 	queueQuit     chan struct{}
 	outQuit       chan struct{}
 	quit          chan struct{}
+
+	// subscribers is a channel for relaying all messages that were received
+	// to this peer.
+	subscribers    map[recvMsgsubscription]struct{}
+	subscriberLock sync.Mutex
 }
 
 // String returns the peer's address and directionality as a human-readable
@@ -1098,6 +1103,35 @@ func (p *Peer) readMessage(encoding wire.MessageEncoding) (wire.Message, []byte,
 	return msg, buf, nil
 }
 
+// recvMsgsubscription is two channels for a subscriber of recevied messages.
+// msgChan for sending the messages and quit for cancelling the subscription.
+type recvMsgsubscription struct {
+	msgChan chan wire.Message
+	quit    chan struct{}
+}
+
+// SubscribeRecvMsg adds a OnRead subscription to the peer.  All bitcoin
+// messages received from this peer will be sent on the returned
+// channel.  A closure is also returned, that should be called to cancel
+// the subscription.
+func (p *Peer) SubscribeRecvMsg() (<-chan wire.Message, func()) {
+	p.subscriberLock.Lock()
+	defer p.subscriberLock.Unlock()
+
+	// No need to buffer this channel as we'll spin up a new goroutine for
+	// every send.
+	msgChan := make(chan wire.Message)
+	quit := make(chan struct{})
+	sub := recvMsgsubscription{
+		msgChan,
+		quit,
+	}
+
+	p.subscribers[sub] = struct{}{}
+
+	return msgChan, func() { close(quit) }
+}
+
 // writeMessage sends a bitcoin message to the peer with logging.
 func (p *Peer) writeMessage(msg wire.Message, enc wire.MessageEncoding) error {
 	// Don't do anything if we're disconnecting.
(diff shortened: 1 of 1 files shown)


peer/peer.go
```diff
@@ -1098,6 +1102,24 @@ func (p *Peer) readMessage(encoding wire.MessageEncoding) (wire.Message, []byte,
 	return msg, buf, nil
 }
 
+// SubscribeRecvMsg adds a OnRead subscription to the peer.  All bitcoin
+// messages received from this peer will be sent on the returned
+// channel.  A closure is also returned, that should be called to cancel
+// the subscription.
+func (p *Peer) SubscribeRecvMsg() (<-chan wire.Message, func()) {
```
Roasbeef, line 1109: Does this need to be thread safe?

peer/peer.go
```diff
@@ -1098,6 +1102,24 @@ func (p *Peer) readMessage(encoding wire.MessageEncoding) (wire.Message, []byte,
...
+	msgChan := make(chan wire.Message, 1)
+	p.subscribers = append(p.subscribers, msgChan)
+
+	// Cancellation is just removing the channel from the subscribers list.
+	idx := len(p.subscribers) - 1
+	cancel := func() {
+		p.subscribers = append(p.subscribers[:idx],
```
Roasbeef, line 1116: Can use `slice.Delete` here: https://pkg.go.dev/slices#Delete

peer/peer.go
```diff
@@ -1098,6 +1102,24 @@ func (p *Peer) readMessage(encoding wire.MessageEncoding) (wire.Message, []byte,
 	return msg, buf, nil
 }
 
+// SubscribeRecvMsg adds a OnRead subscription to the peer.  All bitcoin
+// messages received from this peer will be sent on the returned
+// channel.  A closure is also returned, that should be called to cancel
+// the subscription.
+func (p *Peer) SubscribeRecvMsg() (<-chan wire.Message, func()) {
+	msgChan := make(chan wire.Message, 1)
```
Roasbeef, line 1110: Hmm, so a buffer of 1 will mean that `writeMessage` will block anytime a subscriber fails to immediately read out a message. If that goroutine blocks, then all outbound message writing will also halt.

peer/peer.go
```diff
@@ -1402,6 +1424,10 @@ out:
 		// needed.
 		rmsg, buf, err := p.readMessage(p.wireEncoding)
 		idleTimer.Stop()
+		// Send the received message to all the subscribers.
```
Roasbeef, line 1427: Style nit: lack of newline above.

peer/peer.go
```diff
@@ -1402,6 +1424,10 @@ out:
...
+		for _, sub := range p.subscribers {
+			sub <- rmsg
```
Roasbeef, line 1429: Shouldn't we only do this in the case of a non-error case?

Roasbeef (changes requested):
Straight forward diff. 

One thought I had is if we should actually overhaul the the `query.Peer` interface to better suit our needs, based on the new set of requirements. Given that you've started to use it to implement the out of order block download, do you find it lacking in any notable ways?

kcalvinalvin:
> Straight forward diff.
> 
> One thought I had is if we should actually overhaul the the `query.Peer` interface to better suit our needs, based on the new set of requirements. Given that you've started to use it to implement the out of order block download, do you find it lacking in any notable ways?

Biggest thing I'm noticing is:

1: Hard to punish slow peers on the btcd side since there's no really good way of figuring out which peer is slow or which peer timed out on a block. That information is kept on the `query` package.

2: The `query` package can't disconnect off of slow peers since the ability to disconnect off of a peer is kept on the btcd side. The best it can do is just not ask for blocks from that peer.

Also this is not exactly about the Peer interface but:

3: Peer string passed through `HandleResponse` https://github.com/lightninglabs/neutrino/blob/81d6cd2d2da5b7ce01664ae71d8fce027cd27de4/query/interface.go#L141 is lacking as it makes it annoying on the btcd side. You have to search through the `peerStates` map to get access to the methods on the peer struct like `Disconnect()`.

peer/peer.go
kcalvinalvin: Good point

peer/peer.go
kcalvinalvin: Should we update the go.mod to 1.18? go build fails because go.mod specifies 1.17

peer/peer.go
kcalvinalvin: It's thread safe now :)

peer/peer.go
kcalvinalvin: Actually no need since I changed this to a map

peer/peer.go
kcalvinalvin: Changed so that it's no longer buffered but has each msg send as a new goroutine

peer/peer.go
kcalvinalvin: Addressed

peer/peer.go
kcalvinalvin: Fixed

kcalvinalvin:
Addressed everything in the new push

kcalvinalvin:

PR #2288, opened 2024-12-10
btcec: add deep copy method for public keys

ziggie1984:


btcec/pubkey.go
```diff
@@ -52,6 +52,13 @@ func NewPublicKey(x, y *FieldVal) *PublicKey {
 	return secp.NewPublicKey(x, y)
 }
 
+// CopyPublicKey returns a deep copy of the public key.
+func CopyPublicKey(p *PublicKey) *PublicKey {
+	pk := secp.JacobianPoint{}
```
guggero, line 57: nit: use `var pk secp.JacobianPoint` instead?

guggero (approved):
LGTM :tada:

btcec/pubkey.go
Roasbeef: I don't think we want to copy things this way. You need to call [`ToAffine`](https://pkg.go.dev/github.com/decred/dcrd/dcrec/secp256k1/v4#JacobianPoint.ToAffine) when coming back from the Jacobian coords. 

Here's some code examples:
  * https://github.com/decred/dcrd/blob/692ec68023fb27eca7d5d32801202eed2d86a5c9/hdkeychain/extendedkey.go#L333-L340
  * https://github.com/decred/dcrd/blob/692ec68023fb27eca7d5d32801202eed2d86a5c9/dcrec/secp256k1/privkey.go#L87-L93

After computations, we always transport/handle in affine form.

Roasbeef (changes requested):

btcec/pubkey.go
Roasbeef: I think we should add a property based test here. I wager it would've caught this issue. 

The property we care about here is: `pub == copy(pub)`. 

If you look at what `ToAffine` does, it actually will arrive at new values for x and y, as it's a distinct coordinate system: https://github.com/decred/dcrd/blob/08d8572807872f2b9737f8a118b16c320a04b077/dcrec/secp256k1/curve.go#L136-L149

guggero: I thought the same about `ToAffine()`, but since we use `AsJacobian()` right before that, those should be affine coordinates already or does that no hold?

ziggie1984: it would still work because we are not doing any operations on it so z is set to 1 anyways and Affine and Jacobian are the same coordinates but technically it's wrong but it just happens to work for this particular cpy method. Will correct it.

btcec/pubkey_test.go
```diff
@@ -292,3 +294,61 @@ func TestIsCompressed(t *testing.T) {
...
+// TestCopyPublicKeyProperties tests that the CopyPublicKey function copies the
+// original public key and that the copy is equal to the original.
+func TestCopyPublicKeyProperties(t *testing.T) {
+	f := func(x, y [32]byte) bool {
+
+		// Convert byte slices to FieldVals
+		xf := new(FieldVal)
+		yf := new(FieldVal)
+		if overflow := xf.SetByteSlice(x[:]); overflow {
+			// Only check the function for valid coordinates
```
guggero, line 307: nit: incomplete comments throughout this test (capitalization, punctuation).

guggero (approved):
Nice, LGTM :tada:

PR #2292, opened 2024-12-26
netsync: add handleBlockMsgHeadersFirst

kcalvinalvin:
This is a re-worked part of #2226 only containing the part for adding `handleBlockMsgHeadersFirst`.

The main purpose of this PR is to have code separated out from handling blocks during IBD vs out of IBD.
Since handling of the blocks can be very different in both of the cases, it's useful to separate out the block
handling code as well.


kcalvinalvin:

PR #2294, opened 2024-12-26
netsync: add peerSubscription

kcalvinalvin:
peerSubscription is added to Manager which will allow it subscribers to
receive peers through the channel whenever the Manager is aware of a new
peer that it's been connected to.  This is useful to alert
query.Workmanager that a new peer that's been connected to is eligible
to download blocks from.

This is a part of #2226 and depends on #2287


kcalvinalvin:

PR #2295, opened 2024-12-26
main: add ConnectedPeers() to server

kcalvinalvin:
ConnectedPeers returns all the currently connected peers.  This is used to provide the query.WorkManager with all the currently connected peers from the netsync package.

This PR is part of #2226 and is split here so that it's easier to review.


kcalvinalvin:

Issue #2306, opened 2025-01-20
Bump go.mod version


rabbitprincess:
The current go.mod version is 1.17, which is already 3 years old.
To support new dependencies such as [pebbledb](https://github.com/cockroachdb/pebble), it is good to update to Go 1.22+.

Roasbeef:
Makes sense. In the past we held off, as a big user of the packages also uses `gopherjs` for transpilation, which doesn't yet support any of the new generics features: https://github.com/btcsuite/btcd/issues/1986#issuecomment-1563629309. 

Based on this latest comment, it looks like they now support generics via an experimental build flag: https://github.com/gopherjs/gopherjs/issues/1013#issuecomment-2002033648. Given that, I think we should consider updating to Go 1.22, as we usually lag behind the latest by one release version.

rabbitprincess:
@Roasbeef Can I submit a PR? Updating only go.mod without any dependency seems quite simple.

Issue #2024, opened 2023-08-22
database: investigate replacing leveldb with pebble


Roasbeef:
Today we use a pure Go implementation of `leveldb` for the main database and relevant indexes. The DB has been able to keep up pretty well over the years, but it's maintained by more or less a single person, and [hasn't received any significant commits over the past year or so](https://github.com/syndtr/goleveldb/commits/master). As a result, we should consider moving to a better maintained database that can be a drop in replacement for our current usage. 

Enter [pebble](https://github.com/cockroachdb/pebble) as pure Go KV database inspired by `rocksdb` (which was inspired by `leveldb`). One can view this as the 3rd generation of LSM (Log-Structured-Merge) Tree based KV stores. Compared to `goleveldb` pebble is actively maintained, and has been in used by Cockroachdb in production for ~3 years. The [API is very similar to](https://pkg.go.dev/github.com/cockroachdb/pebble?utm_source=godoc) `goleveldb`, which means we won't need a significant overhaul. 

Rather than replace it all at once, we should first add it as a new database option, with no migration path. After enough time, we can add a migration path, t hen eventually deprecate the old database type. At a glance, [`pebble` also appears to be more performant](https://www.fantom.foundation/blog/narrow-gateway-hurdles-pebbledb-vs-leveldb/#:~:text=Pebble%20was%20able%20to%20sync,sync%20in%20a%20fixed%20time) than `goleveldb`.

kcalvinalvin:
FWIW leveldb itself hasn't had that much development [either](https://github.com/google/leveldb/commits/main). I don't think it necessarily means lack of maintenance.

That being said, I'm not opposed to having pebble as an option. If it ends up being more performant, that's a win. Though maybe have a big warning that pebble itself may not be consensus compatible with leveldb.

0xEclair:
If no enough performance improvement, i think there is no need to replace leveldb with pebble. To replace, we must do a lot of test for pebble to check whether there are internal bugs like missing data or incorrect values when insert or get.

Another point, maybe keepping building `pebble` means that there are bugs. `leveldb` is enough stable and safe.

rabbitprincess:
@Roasbeef Can I proceed with this task? using Pebble enabled synchronization more than 5x faster in my benchmark.

kcalvinalvin:
> [@Roasbeef](https://github.com/Roasbeef) Can I proceed with this task? using Pebble enabled synchronization more than 3~5x faster in my benchmark.

[...]

PR #2307, opened 2025-01-23
database: add pebbledb support

rabbitprincess:
#### Overview
This PR introduces PebbleDB as a new database engine, an alternative to leveldb. https://github.com/btcsuite/btcd/issues/2024
To support multiple database engines (PebbleDB and LevelDB), db interface has been implemented in ffldb.

#### How to run
``
./btcd --dbtype=pebbledb
``

#### Caution
This update requires Go 1.22+. https://github.com/btcsuite/btcd/issues/2306


Roasbeef:
Approved CI run.

kcalvinalvin:
This is a big change and I see that you're still making changes here. Do ping me once things are ready for review!

rabbitprincess:
Thank you @kcalvinalvin, It is ready for review!
By the way, bump go.mod version is also big change. it might be better to work on other pull request.

yyforyongyu:
Approved CI run.

yyforyongyu (changes requested):
Thanks for the PR. Please check the contribution guidelines regarding [Ideal Git Commit Structure](https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md#ideal-git-commit-structure).

rabbitprincess:
@Roasbeef @kcalvinalvin Is there anything I should change or improve on the PR? It seems to be stuck.

kcalvinalvin:
Just found out that batch writing in pebbledb only supports up to 4GB. Shown in this test here:

```
func TestPebbleDBBatchWriteFail(t *testing.T) {
        tmpDir := t.TempDir()
        db, err := pebble.Open(tmpDir, nil)
        if err != nil {
                t.Fatal(err)
        }
        
        k := make([]byte, math.MaxUint32)
        v := [1]byte{}
        batch := db.NewBatch()
        err = batch.Set(k[:], v[:], nil)
        if err != nil {
                t.Fatal(err)
        }
        batch.Commit(nil)
```

This does make things tricky to have everything written atomically.

EDIT:

Seems like we should be bypassing the internal memtable for pebbledb and just write straight to the tables on disk.
https://github.com/cockroachdb/pebble/issues/702#issuecomment-632055748

gosunuts:
> Just found out that batch writing in pebbledb only supports up to 4GB. Shown in this test here:

Indeed. It must be a problem if a single batch exceeds 4GB. But I’m not sure this actually happens in practice. The batches I’ve looked at typically had ~10MB of keys and ~100MB of values (see [this](https://github.com/rabbitprincess/btcd/blob/619b77926ec950cc47b87e4ec988771ff885d62f/database/ffldb/dbcache.go#L472)).
Maybe in some cases, a utxoCache flush could exceed the limit. ( with --utxocachemaxsize=4000 )

> Seems like we should be bypassing the internal memtable for pebbledb and just write straight to the tables on disk.

It can be a great alternative for large flushes. Writing directly via SSTable ingest avoids the 4GB batch limit and keeps memory usage low.
A hybrid approach might work: use batch for regular commits and use Ingest for utxoCache flushes.

gosunuts:
@kcalvinalvin It seems that no single batch exceeds 1GB of key/values while syncing the bitcoin mainnet, even with `--utxocachemaxsize=2000`.
Could you share the circumstances under which you encountered a 4GB flush?
( In fact, flushing 4GB of data from memory to disk at once is something to avoid. )

kcalvinalvin:
> @kcalvinalvin It seems that no single batch exceeds 1GB of key/values, even with `--utxocachemaxsize=2000`. Could you share the circumstances under which you encountered a 4GB flush? ( In fact, flushing 4GB of data from memory to disk at once is something to avoid. )

The current UTXO set size is ~12GB. Since on disk we use varints, it's gonna be bigger in memory. Because of this, it's reasonable for someone doing ibd to give the `--utxocachemaxsize=16000`.

Bitcoin Core also dropped the limit.
https://github.com/bitcoin/bitcoin/issues/28249

gosunuts:
@kcalvinalvin It seems reasonable to take a two-track approach. It might be a good idea to call ingest() in PebbleDB when processing a batch larger than 4GB.

We could also start pebbledb support with a limit on the utxoCacheSize ( < 4000 ), and then remove the limit once ingest() is supported.

gosunuts:
As we all know, Pebble won't fully replace LevelDB. It's a good way to start with experimental usage and gradually improve it.

kcalvinalvin:
> As we all know, Pebble won't fully replace LevelDB. It's a good way to start with experimental usage and gradually improve it.

Some of the code here may be of interest.
https://github.com/utreexo/utreexod/pull/325

gosunuts:
@kcalvinalvin What do you think about setting a 4GB limit for maxUtxoCacheSize when using PebbleDB? We could consider this as an experimental approach for PebbleDB.

```
// Don`t allow utxoCacheMaxSize greater than 4GB (4096 MiB) when using PebbleDB.
// This is due to the batch size limitation of PebbleDB.
if cfg.DbType == ffldb.PebbleDB && cfg.UtxoCacheMaxSizeMiB > 4096 {
    str := "%s: The utxocachemaxsize option may not be greater " +
        "than 4096 MiB for the PebbleDB database backend -- " +
        "parsed [%d]"
    err := fmt.Errorf(str, funcName, cfg.UtxoCacheMaxSizeMiB)
    fmt.Fprintln(os.Stderr, err)
    return nil, nil, err
}
```

gosunuts:
@yyforyongyu Is there any progress?

PR #2318, opened 2025-02-10
feat: Add testnet4 chain support (recreated)

bullet-tooth:
_This PR is a recreation of the [previous PR](https://github.com/btcsuite/btcd/pull/2274) because I incidentally dropped a brunch and the PR was closed automatically._ 

# Description

This PR adds support of the Bitcoin testnet version 4 chain (testnet3 is in a way of deprecation).

Reference: https://bips.dev/94/

Testnet4 genesis block: https://mempool.space/testnet4/block/00000000da84f2bafbbc53dee25a72ae507ff4914b867c565be350b0da8bf043


wire/protocol.go
```diff
@@ -179,6 +179,9 @@ const (
 	// TestNet3 represents the test network (version 3).
 	TestNet3 BitcoinNet = 0x0709110b
 
+	// TestNet4 represents the test network (version 4).
+	TestNet4 BitcoinNet = 0x1c163f28
```
guggero, line 183: This has incorrect endianness and should be `0x283f161c`.

chaincfg/params.go
```diff
@@ -673,6 +677,89 @@ var TestNet3Params = Params{
...
+	Deployments: [DefinedDeployments]ConsensusDeployment{
+		DeploymentTestDummy: {
+			BitNumber: 28,
+			DeploymentStarter: NewMedianTimeDeploymentStarter(
+				time.Unix(1199145601, 0), // January 1, 2008 UTC
+			),
+			DeploymentEnder: NewMedianTimeDeploymentEnder(
+				time.Unix(1230767999, 0), // December 31, 2008 UTC
+			),
+		},
```
guggero, line 727: Seems like we need the CSV and SegWit deployments here as well, otherwise things panicked on my machine:
```
			DeploymentTestDummyMinActivation: {
				BitNumber:                 22,
				CustomActivationThreshold: 1815,    // Only needs 90% hash rate.
				MinActivationHeight:       10_0000, // Can only activate after height 10k.
				DeploymentStarter: NewMedianTimeDeploymentStarter(
					time.Time{}, // Always available for vote
				),
				DeploymentEnder: NewMedianTimeDeploymentEnder(
					time.Time{}, // Never expires
				),
			},
			DeploymentCSV: {
				BitNumber: 29,
				DeploymentStarter: NewMedianTimeDeploymentStarter(
					time.Time{}, // Always available for vote
				),
				DeploymentEnder: NewMedianTimeDeploymentEnder(
					time.Time{}, // Never expires
				),
			},
			DeploymentSegwit: {
				BitNumber: 29,
				DeploymentStarter: NewMedianTimeDeploymentStarter(
					time.Time{}, // Always available for vote
				),
				DeploymentEnder: NewMedianTimeDeploymentEnder(
					time.Time{}, // Never expires
				),
			},
```

guggero (commented):
Thanks for the updates. Definitely looks better now.

Have a couple of fixes that are required, otherwise things crash quite immediately (please test things before requesting review in the future).

After those fixes I was able to sync up to block height 70513

saubyk:
hey @bullet-tooth are you planning to work on this pr at a high priority? We are interested in moving this along quickly. Thanks.

config.go
```diff
@@ -170,7 +170,8 @@ type config struct {
 	SigNet               bool          `long:"signet" description:"Use the signet test network"`
 	SigNetChallenge      string        `long:"signetchallenge" description:"Connect to a custom signet network defined by this challenge instead of using the global default signet test network -- Can be specified multiple times"`
 	SigNetSeedNode       []string      `long:"signetseednode" description:"Specify a seed node for the signet network instead of using the global default signet network seed nodes"`
-	TestNet3             bool          `long:"testnet" description:"Use the test network"`
+	TestNet3             bool          `long:"testnet" description:"Use the test network (version 3)"`
+	TestNet4             bool          `long:"testnet4" description:"Use the test network (version 4)"`
```
Roasbeef, line 174: We should add this to the sample btcd conf: https://github.com/btcsuite/btcd/blob/master/sample-btcd.conf

Roasbeef (changes requested):

Roasbeef:
I added a commit to add some property based tests for `assertNoTimewrap`. Please take care to ensure those commits aren't dropped during a rebase.

chaincfg/params.go
```diff
@@ -673,6 +677,118 @@ var TestNet3Params = Params{
...
+			BitNumber: 29,
+			DeploymentStarter: NewMedianTimeDeploymentStarter(
+				time.Time{}, // Always available for vote
+			),
+			DeploymentEnder: NewMedianTimeDeploymentEnder(
+				time.Time{}, // Never expires
+			),
+		},
+		DeploymentTaproot: {
+			BitNumber: 2,
```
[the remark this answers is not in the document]
Roasbeef: > Or, will it be better to have hard coded heights for the activation similar to bitcoind?

Perhaps let's go with the hard coded activation heights? I think the code diff for both versions would be somewhat similar. 

The target area for this change would be the bip 9/8 activation logic: https://github.com/btcsuite/btcd/blob/2a64716b1c16394f716101e915cf248e25a6e119/blockchain/thresholdstate.go#L358-L446

We can update the `deploymentChecker` to read that activation height, and just go directly to active once that height has been passed. The `thresholdState` method already has access to the current block height, as it can derive that from the `prevNode`. 

I think we should do this in another PR so the diff is isolated, and we can add unit + itests for the change.

Roasbeef:
I added two commits here that implement the BIP 9 changes, can you add them to this branch? https://github.com/btcsuite/btcd/compare/master...Roasbeef:btcd:bip-9-always-active?expand=1

Issue #2319, opened 2025-02-12
Fix ci / makefile


gosunuts:
### Overview

Related to [btcsuite/btcd#2240](https://github.com/btcsuite/btcd/pull/2240)
Makefile is outdated, and several processes do not work correctly.
- `DEPGET` and `GOFILES_NOVENDOR` are unnecessary.
- `golint` and `goimports` are broken.
- [goacc](https://github.com/ory/go-acc) is deprecated. ( replace to builtin )
- Some commands and scripts are unnecessary and do not work on Windows.

Roasbeef:
Fixed by https://github.com/btcsuite/btcd/pull/2320

PR #2316, opened 2025-02-10
deps: bump go version

rabbitprincess:
#### Overview
Bump `go.mod` version from 1.17 to 1.22, related to https://github.com/btcsuite/btcd/issues/2306.

#### Notes
The current GitHub workflows and Makefile depend on `GOPATH`. It would be better to update them to use `Go Modules`.

PR #2320, opened 2025-02-12
feat: fix ci / makefile

gosunuts:
### Overview
fix / cleanup makefile and correcting several commands.
Related to https://github.com/btcsuite/btcd/issues/2319 https://github.com/btcsuite/btcd/pull/2316

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,15 @@
 PKG := github.com/btcsuite/btcd
 
-LINT_PKG := github.com/golangci/golangci-lint/cmd/golangci-lint
+LINT_PKG := github.com/golangci/golangci-lint/v2/cmd/golangci-lint
 GOIMPORTS_PKG := golang.org/x/tools/cmd/goimports
 
-GO_BIN := ${GOPATH}/bin
+GO_BIN := ${shell go env GOBIN}
 LINT_BIN := $(GO_BIN)/golangci-lint
+GOIMPORTS_BIN := $(GO_BIN)/goimports
 
-LINT_COMMIT := v1.18.0
+LINT_COMMIT := v2.1.6
+GOIMPORTS_COMMIT := a24facf9e5586c95743d2f4ad15d148c7a8cf00b
 
-DEPGET := cd /tmp && go install -v
 GOBUILD := go build -v
 GOINSTALL := go install -v 
 DEV_TAGS := rpctest
@@ -16,20 +17,14 @@ GOTEST_DEV = go test -v -tags=$(DEV_TAGS)
 GOTEST := go test -v
 COVER_FLAGS = -coverprofile=coverage.txt -covermode=atomic -coverpkg=$(PKG)/...
 
-GOFILES_NOVENDOR = $(shell find . -type f -name '*.go' -not -path "./vendor/*")
-
-RM := rm -f
-CP := cp
-MAKE := make
-XARGS := xargs -L 1
-
 # Linting uses a lot of memory, so keep it under control by limiting the number
 # of workers if requested.
 ifneq ($(workers),)
 LINT_WORKERS = --concurrency=$(workers)
 endif
+LINT_TIMEOUT := 5m
 
-LINT = $(LINT_BIN) run -v $(LINT_WORKERS)
+LINT = $(LINT_BIN) run -v $(LINT_WORKERS) --timeout=$(LINT_TIMEOUT)
 
 GREEN := "\\033[0;32m"
 NC := "\\033[0m"
@@ -49,12 +44,12 @@ all: build check
 
 $(LINT_BIN):
 	@$(call print, "Fetching linter")
-	$(DEPGET) $(LINT_PKG)@$(LINT_COMMIT)
+	$(GOINSTALL) $(LINT_PKG)@$(LINT_COMMIT)
 
 #? goimports: Install goimports
 goimports:
 	@$(call print, "Installing goimports.")
-	$(DEPGET) $(GOIMPORTS_PKG)
+	$(GOINSTALL) $(GOIMPORTS_PKG)@$(GOIMPORTS_COMMIT)
 
 # ============
 # INSTALLATION
@@ -95,30 +90,28 @@ check: unit
 unit:
 	@$(call print, "Running unit tests.")
 	$(GOTEST_DEV) ./... -test.timeout=20m
-	cd btcec; $(GOTEST_DEV) ./... -test.timeout=20m
-	cd btcutil; $(GOTEST_DEV) ./... -test.timeout=20m
-	cd btcutil/psbt; $(GOTEST_DEV) ./... -test.timeout=20m
+	cd btcec && $(GOTEST_DEV) ./... -test.timeout=20m
+	cd btcutil && $(GOTEST_DEV) ./... -test.timeout=20m
+	cd btcutil/psbt && $(GOTEST_DEV) ./... -test.timeout=20m
 
 #? unit-cover: Run unit coverage tests
 unit-cover:
 	@$(call print, "Running unit coverage tests.")
 	$(GOTEST) $(COVER_FLAGS) ./...
+
 	# We need to remove the /v2 pathing from the module to have it work
-	# nicely with the CI tool we use to render live code coverage.	
-	cd btcec; $(GOTEST) $(COVER_FLAGS) ./...; \
-		sed -i.bak 's/v2\///g' coverage.txt
-
-	cd btcutil; $(GOTEST) $(COVER_FLAGS) ./...
-
-	cd btcutil/psbt; $(GOTEST) $(COVER_FLAGS) ./...
+	# nicely with the CI tool we use to render live code coverage.
+	cd btcec && $(GOTEST) $(COVER_FLAGS) ./... && sed -i.bak 's/v2\///g' coverage.txt
+	cd btcutil && $(GOTEST) $(COVER_FLAGS) ./...
+	cd btcutil/psbt && $(GOTEST) $(COVER_FLAGS) ./...
 
 #? unit-race: Run unit race tests
 unit-race:
 	@$(call print, "Running unit race tests.")
 	env CGO_ENABLED=1 GORACE="history_size=7 halt_on_errors=1" $(GOTEST) -race -test.timeout=20m ./...
-	cd btcec; env CGO_ENABLED=1 GORACE="history_size=7 halt_on_errors=1" $(GOTEST) -race -test.timeout=20m ./...
-	cd btcutil; env CGO_ENABLED=1 GORACE="history_size=7 halt_on_errors=1" $(GOTEST) -race -test.timeout=20m ./...
-	cd btcutil/psbt; env CGO_ENABLED=1 GORACE="history_size=7 halt_on_errors=1" $(GOTEST) -race -test.timeout=20m ./...
+	cd btcec && env CGO_ENABLED=1 GORACE="history_size=7 halt_on_errors=1" $(GOTEST) -race -test.timeout=20m ./...
+	cd btcutil && env CGO_ENABLED=1 GORACE="history_size=7 halt_on_errors=1" $(GOTEST) -race -test.timeout=20m ./...
+	cd btcutil/psbt && env CGO_ENABLED=1 GORACE="history_size=7 halt_on_errors=1" $(GOTEST) -race -test.timeout=20m ./...
 
 # =========
 # UTILITIES
@@ -127,9 +120,9 @@ unit-race:
 #? fmt: Fix imports and formatting source
 fmt: goimports
 	@$(call print, "Fixing imports.")
-	goimports -w $(GOFILES_NOVENDOR)
+	$(GOIMPORTS_BIN) -w .
 	@$(call print, "Formatting source.")
-	gofmt -l -w -s $(GOFILES_NOVENDOR)
+	gofmt -l -w -s .
 
 #? lint: Lint source
 lint: $(LINT_BIN)
@@ -139,8 +132,8 @@ lint: $(LINT_BIN)
 #? clean: Clean source
 clean:
 	@$(call print, "Cleaning source.$(NC)")
-	$(RM) coverage.txt btcec/coverage.txt btcutil/coverage.txt btcutil/psbt/coverage.txt
-	
+	rm -f coverage.txt btcec/coverage.txt btcutil/coverage.txt btcutil/psbt/coverage.txt
+
 #? tidy-module: Run 'go mod tidy' for all modules
 tidy-module:
 	echo "Running 'go mod tidy' for all modules"
@@ -155,7 +148,8 @@ tidy-module:
 	unit-race \
 	fmt \
 	lint \
-	clean
+	clean \
+	tidy-module
 
 #? help: Get more info on make commands
 help: Makefile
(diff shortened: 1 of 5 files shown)


.github/workflows/go.yml
```diff
@@ -76,12 +76,12 @@ jobs:
...
     steps:
       - name: Set up Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v5
         with:
           go-version: ${{ env.GO_VERSION }}
 
       - name: Check out source
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
```
Roasbeef, line 84: 👍

Makefile
```diff
@@ -1,15 +1,12 @@
...
 LINT_PKG := github.com/golangci/golangci-lint/cmd/golangci-lint
-GOACC_PKG := github.com/ory/go-acc
 GOIMPORTS_PKG := golang.org/x/tools/cmd/goimports
 
 GO_BIN := ${GOPATH}/bin
 LINT_BIN := $(GO_BIN)/golangci-lint
-GOACC_BIN := $(GO_BIN)/go-acc
 
 LINT_COMMIT := v1.18.0
-GOACC_COMMIT := 80342ae2e0fcf265e99e76bcc4efd022c7c3811b
```
Roasbeef, line 12: So this is important, as it absolves any linter violations _before_ that commit. Otherwise, we'll have to have a PR which fixes _every_ linter violation. 

In `lnd`, we use `golangci` which still supports this feature: https://github.com/lightningnetwork/lnd/pull/9382/files

scripts/tidy_modules.sh
```diff
@@ -1,17 +0,0 @@
-#!/bin/bash
-
-SUBMODULES=$(find . -mindepth 2 -name "go.mod" | cut -d'/' -f2)
```
Roasbeef, line 3: Why remove? Given that it's able to locate sub-modules automatically? 

In the new diff, we need to enumerate them all by hand.

Makefile
```diff
@@ -27,8 +25,9 @@ XARGS := xargs -L 1
 ifneq ($(workers),)
 LINT_WORKERS = --concurrency=$(workers)
 endif
+LINT_TIMEOUT := 5m
 
-LINT = $(LINT_BIN) run -v $(LINT_WORKERS)
+LINT = $(LINT_BIN) run -v $(LINT_WORKERS) --timeout=$(LINT_TIMEOUT) --issues-exit-code=0
```
Roasbeef, line 24: Why's `issues-exit-code` needed now?

Makefile
```diff
@@ -14,8 +14,6 @@ GOTEST_DEV = go test -v -tags=$(DEV_TAGS)
 GOTEST := go test -v
 GOMODTIDY := go mod tidy
 
-GOFILES_NOVENDOR = $(shell find . -type f -name '*.go' -not -path "./vendor/*")
```
Roasbeef, line 21: 👍 it's been years since we vendor'd deps

rpcclient/chain_test.go
```diff
@@ -146,7 +146,7 @@ func TestClientConnectedToWSServerRunner(t *testing.T) {
 	}
 
 	testTable := []TestTableItem{
-		TestTableItem{
+		{
```
Roasbeef, line 149: Were these the only instances in violation? If so then, perhaps we don't need that lint commit hash after all. 

Was pretty sure we had some PRs merged in recent times that did things like violate the 80 char column limit.

scripts/tidy_modules.sh
rabbitprincess: I just thought it was simple approach. Other commands, such as unit / unit-cover, are still done entirely by hand. It works perfectly as long as nested go.mod is not added.

Makefile
rabbitprincess: It just exits with code 0 when a lint error occurs. It's not needed, so better to remove it.

rpcclient/chain_test.go
rabbitprincess: I do not know about commit history, but it is only change about `make fmt`.

rabbitprincess: Not `make lint` - As far as I know, Go lint only gives warnings and doesn't change the code automatically.

Makefile
rabbitprincess: resolved in [bdf7318](https://github.com/btcsuite/btcd/pull/2320/commits/bdf7318d0ea9b1648ba1a2d460bd88dcb5e4f5d6)

yyforyongyu:
Is this still under development?

rabbitprincess:
@yyforyongyu It is still in development, but it would be good to recreate the PR to follow [Ideal Git Commit Structure](https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md#ideal-git-commit-structure).

rabbitprincess:
@yyforyongyu this pr and https://github.com/btcsuite/btcd/pull/2307 are force-pushed. can you review again?

GustavoStingelin (changes requested):
Hey @rabbitprincess, I tested your PR and ran into an issue during the lint step. It looks like the linter version used is very outdated (v1.18.0, released 6 years ago):

```
➜  btcd git:(feature/fix-ci) ✗ make lint
Fetching linter
go install -v github.com/golangci/golangci-lint/cmd/golangci-lint@v1.18.0
go: github.com/golangci/golangci-lint/cmd/golangci-lint@v1.18.0:
  The go.mod file for the module providing named packages contains one or
  more replace directives. It must not contain directives that would cause
  it to be interpreted differently than if it were the main module.
make: *** [Makefile:46: /home/head/go/bin/golangci-lint] Error 1
```

System info:
```
Linux 6.11.0-26-generic x86_64
Go version go1.24.2 linux/amd64
```

Other steps like `make fmt` and `make unit-cover` are working fine. I think we could consider bumping the linter version to something more recent.

gosunuts:
@GustavoStingelin Sure! It's been updated to v2.1.6.

GustavoStingelin:
LGTM!
  
@Roasbeef @yyforyongyu with the updated `golangci-lint`, we're now seeing 162 issues across several linters. Since the linter isn’t part of the pipeline yet, would you prefer merging this PR as-is and handling the fixes in follow-up PRs? Or should we resolve all the issues before merging?
  
  Breakdown of issues:
  * errcheck: 50
  * ineffassign: 16
  * staticcheck: 50
  * unused: 46

GustavoStingelin:
or maybe copy the `.golangci.yml` from `lnd` as mentioned previously

GustavoStingelin (approved):

GustavoStingelin:
@kcalvinalvin any thoughts?

Roasbeef:
>  Since the linter isn’t part of the pipeline yet, would you prefer merging this PR as-is and handling the fixes in follow-up PRs?

Usually what we do is set a commit hash that grandfathers in any violations. So then only _new_ PRs that introduce _new_ linting errors get triggered.

Makefile
```diff
@@ -95,30 +90,29 @@ check: unit
...
 
 #? unit-cover: Run unit coverage tests
 unit-cover:
 	@$(call print, "Running unit coverage tests.")
-	$(GOTEST) $(COVER_FLAGS) ./...
-	# We need to remove the /v2 pathing from the module to have it work
-	# nicely with the CI tool we use to render live code coverage.	
-	cd btcec; $(GOTEST) $(COVER_FLAGS) ./...; \
-		sed -i.bak 's/v2\///g' coverage.txt
+	$(GOTEST) -coverprofile=coverage.txt ./...
```
Roasbeef: Looks like we're no longer using `COVER_FLAGS` at all here. 

We can resolve this by going back to
```
$(GOTEST) $(COVER_FLAGS)
```

IIUC, `covermode=atomic` is needed as we run tests in parallel.

gosunuts: @Roasbeef Great! but using -coverpkg=$(PKG)/... makes unit test extremely slow.
It's better to remove PKG and use a relative path instead (as recommended by Go modules).

```
COVER_FLAGS = -coverprofile=coverage.txt -covermode=atomic -coverpkg=./...

#? build: Build all binaries, place them in project directory
build:
	@$(call print, "Building all binaries")
	$(GOBUILD) .
	$(GOBUILD) ./cmd/btcctl
	$(GOBUILD) ./cmd/gencerts
	$(GOBUILD) ./cmd/findcheckpoint
	$(GOBUILD) ./cmd/addblock

#? install: Install all binaries, place them in $GOPATH/bin
install:
	@$(call print, "Installing all binaries")
	$(GOINSTALL) .
	$(GOINSTALL) ./cmd/btcctl
	$(GOINSTALL) ./cmd/gencerts
	$(GOINSTALL) ./cmd/findcheckpoint
	$(GOINSTALL) ./cmd/addblock

#? release-install: Install btcd and btcctl release binaries, place them in $GOPATH/bin
release-install:
	@$(call print, "Installing btcd and btcctl release binaries")
	env CGO_ENABLED=0 $(GOINSTALL) -trimpath -ldflags="-s -w -buildid=" .
	env CGO_ENABLED=0 $(GOINSTALL) -trimpath -ldflags="-s -w -buildid=" ./cmd/btcctl
	

#? unit-cover: Run unit coverage tests
unit-cover:
	@$(call print, "Running unit coverage tests.")
	$(GOTEST) $(COVER_FLAGS) ./...

	# We need to remove the /v2 pathing from the module to have it work
	# nicely with the CI tool we use to render live code coverage.
	cd btcec && $(GOTEST) $(COVER_FLAGS) ./... && sed -i.bak 's/v2\///g' coverage.txt
	cd btcutil && $(GOTEST) $(COVER_FLAGS) ./...
	cd btcutil/psbt && $(GOTEST) $(COVER_FLAGS) ./...
```

GustavoStingelin (changes requested):
@rabbitprincess please, add the file `.golangci.yml` to the root path with the value 

```
version: "2"
issues:
  # Only show newly introduced problems.
  new-from-rev: 80b74d6c5a0088a66dc96df6777d21e15c04849c

```

gosunuts:
> @rabbitprincess please, add the file `.golangci.yml` to the root path with the value
> 
> ```
> version: "2"
> issues:
>   # Only show newly introduced problems.
>   new-from-rev: 80b74d6c5a0088a66dc96df6777d21e15c04849c
> ```

Sure! It's been applied.

GustavoStingelin (approved):
tACK

Makefile
Roasbeef: > Great! but using -coverpkg=$(PKG)/... makes unit test extremely slow.

Sure, but this gives a more accurate coverage rating IIUC. See this repo for an example of the difference: https://github.com/yudai/go_cover. 

We used a combination of those flags to permit running tests in parallel across packages, ensuring that cross package test contribute to the final coverage report.

gosunuts: @Roasbeef Understood. I'll update to use COVER_FLAGS as you suggested.

GustavoStingelin:
I think you will need to rebase your first commit with the changes from the last

gosunuts:
@GustavoStingelin Sure. It's been applied.

gosunuts:
> If I run `make lint`, I get this error:
> 
> ```
> ⛰   make lint
>  Linting source.
> /Users/roasbeef/gocode/bin/golangci-lint run -v
> panic: load embedded ruleguard rules: rules/rules.go:13: can't load fmt
> 
> goroutine 1 [running]:
> github.com/go-critic/go-critic/checkers.init.22()
>         /Users/roasbeef/gocode/pkg/mod/github.com/go-critic/go-critic@v0.6.3/checkers/embedded_rules.go:47 +0x494
> make: *** [lint] Error 2
> ```

Can you check the output of `go env`?
It looks like go module might not be enabled in your environment (e.g., `GO111MODULE=off` )

GustavoStingelin:
> If I run make lint, I get this error:

FYI, `make lint` still works fine on my env (Go 1.24.4, Linux amd64, GO111MODULE='').

Roasbeef:
Ah I'm on Go 1.23.6 locally. So seems that we'll need to update the top level `go.mod` to use Go 1.24. 

[Go 1.25 comes out in August](https://tip.golang.org/doc/go1.25), so that should line up with our policy to support the past two go versions.

GustavoStingelin:
> Ah I'm on Go 1.23.6 locally. So seems that we'll need to update the top level `go.mod` to use Go 1.24.
> 
> [Go 1.25 comes out in August](https://tip.golang.org/doc/go1.25), so that should line up with our policy to support the past two go versions.

works well on version `1.23.6` here.

```
➜  btcd git:(feature/fix-ci) ✗ mise use go@1.23.6
mise ~/code/btcd/mise.toml tools: go@1.23.6

➜  btcd git:(feature/fix-ci) ✗ go version
go version go1.23.6 linux/amd64

➜  btcd git:(feature/fix-ci) ✗ rm $GOBIN/golangci-lint

➜  btcd git:(feature/fix-ci) ✗ make lint 
 Fetching linter
go install -v  github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1.6
 Linting source.
/home/head/.local/share/mise/installs/go/1.23.6/bin/golangci-lint run -v  --timeout=5m
INFO golangci-lint has version v2.1.6 built with go1.23.6 from (unknown, modified: ?, mod sum: "h1:LXqShFfAGM5BDzEOWD2SL1IzJAgUOqES/HRBsfKjI+w=") on (unknown) 
... omited logs
INFO [runner] linters took 116.4495ms with stages: goanalysis_metalinter: 38.478262ms 
0 issues.
INFO File cache stats: 0 entries of total size 0B 
INFO Memory: 5 samples, avg is 59.6MB, max is 94.0MB 
INFO Execution took 360.073736ms 
```

Makefile
```diff
@@ -1,35 +1,30 @@
 PKG := github.com/btcsuite/btcd
 
-LINT_PKG := github.com/golangci/golangci-lint/cmd/golangci-lint
+LINT_PKG := github.com/golangci/golangci-lint/v2/cmd/golangci-lint
 GOIMPORTS_PKG := golang.org/x/tools/cmd/goimports
 
-GO_BIN := ${GOPATH}/bin
+GO_BIN := ${shell go env GOPATH}/bin
```
GustavoStingelin:
```suggestion
GO_BIN := ${shell go env GOBIN}
```
When testing with different Go versions, I found that `go env GOBIN` is more reliable than assuming `$GOPATH/bin`. In my case, using [mise](https://mise.jdx.dev/demo.html) as a tool manager, GOBIN doesn't follow the traditional path structure.

gosunuts: Can I just commit this, or should I rebase it?

GustavoStingelin: rebase bc your previous commit modifies this line and is in this context

gosunuts: @GustavoStingelin It's been applied.

GustavoStingelin: cool, tACK in `go1.24.5` and `go1.23.1` `linux/amd64`

Roasbeef (approved):
LGTM 🚀

Makefile
guggero: `GOBIN` doesn't seem to be set by default on all systems. So we need a fallback here.

This is what works on my system:
```
# If GOBIN is not set, default to GOPATH/bin.
ifeq ($(GO_BIN),)
GO_BIN := $(shell go env GOPATH)/bin
endif
```

moawnallah: > `GOBIN` doesn't seem to be set by default on all systems. So we need a fallback here.
> 
> This is what works on my system:
> 
> ```
> # If GOBIN is not set, default to GOPATH/bin.
> ifeq ($(GO_BIN),)
> GO_BIN := $(shell go env GOPATH)/bin
> endif
> ```

Followed-up with this PR https://github.com/btcsuite/btcd/pull/2410, addressing this comment

PR #2322, opened 2025-02-17
rpcclient: clarify GetBlockCount description

starius:
`GetBlockCount` returns the height of the latest block, not the total number of blocks. The genesis block has a height of 0.

The previous description could lead to an off-by-one error for anyone relying on it.


Roasbeef (approved):
LGTM 🦋

PR #2325, opened 2025-02-24
musig2: Add adaptor signature support to MuSig2 protocol

martonp:
This commit adds comprehensive support for adaptor signatures in the MuSig2 protocol, including:

- New methods for generating and setting adaptor points
- Support for creating partial adaptor signatures
- Methods to adapt final signatures and recover adaptor secrets
- Updated signature verification and combination logic
- Added test cases for adaptor signature functionality across different signer configurations

### DIFF
--- a/btcec/schnorr/musig2/context.go
+++ b/btcec/schnorr/musig2/context.go
@@ -620,7 +620,7 @@ func (s *Session) SetAdaptorSecret(msg [32]byte,
 		return ErrAlreadySigned
 	}
 
-	nonce, _, err := computeSigningNonce(
+	nonce, _, err := ComputeSigningNonce(
 		*s.combinedNonce, s.ctx.combinedKey.FinalKey, msg,
 	)
 	if err != nil {
@@ -663,7 +663,7 @@ func (s *Session) SetAdaptorPoint(msg [32]byte,
 
 	// Verify that the adaptor point added to the combined nonce will result in
 	// a point with even y-coordinate.
-	nonce, _, err := computeSigningNonce(
+	nonce, _, err := ComputeSigningNonce(
 		*s.combinedNonce, s.ctx.combinedKey.FinalKey, msg,
 	)
 	if err != nil {
@@ -811,7 +811,7 @@ func (s *Session) FinalSig() *schnorr.Signature {
 
 // AdaptFinalSig adapts the final signature with the provided adaptor secret
 // key, and returns a valid signature.
-func (s *Session) AdaptFinalSig(adaptorSecret *btcec.ModNScalar) (
+func (s *Session) AdaptFinalSig() (
 	*schnorr.Signature, error) {
 
 	if s.finalSig == nil {
@@ -827,7 +827,7 @@ func (s *Session) AdaptFinalSig(adaptorSecret *btcec.ModNScalar) (
 	sigS := new(btcec.ModNScalar)
 	sigS.SetByteSlice(sigB[32:64])
 
-	sigS.Add(adaptorSecret)
+	sigS.Add(s.adaptorSecret)
 	adaptedSig := schnorr.NewSignature(r, sigS)
 
 	if !adaptedSig.Verify(s.msg[:], s.ctx.combinedKey.FinalKey) {
--- a/btcec/schnorr/musig2/sign.go
+++ b/btcec/schnorr/musig2/sign.go
@@ -207,9 +207,9 @@ func WithBip86SignTweak() SignOption {
 	}
 }
 
-// computeSigningNonce calculates the final nonce used for signing. This will
+// ComputeSigningNonce calculates the final nonce used for signing. This will
 // be the R value used in the final signature.
-func computeSigningNonce(combinedNonce [PubNonceSize]byte,
+func ComputeSigningNonce(combinedNonce [PubNonceSize]byte,
 	combinedKey *btcec.PublicKey, msg [32]byte) (
 	*btcec.JacobianPoint, *btcec.ModNScalar, error) {
 
@@ -325,7 +325,7 @@ func Sign(secNonce [SecNonceSize]byte, privKey *btcec.PrivateKey,
 	// We'll now combine both the public nonces, using the blinding factor
 	// to tweak the second nonce:
 	//  * R = R_1 + b*R_2
-	nonce, nonceBlinder, err := computeSigningNonce(
+	nonce, nonceBlinder, err := ComputeSigningNonce(
 		combinedNonce, combinedKey.FinalKey, msg,
 	)
 	if err != nil {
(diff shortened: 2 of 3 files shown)


btcec/schnorr/musig2/context.go
```diff
@@ -548,6 +573,129 @@ func (s *Session) RegisterPubNonce(nonce [PubNonceSize]byte) (bool, error) {
...
+
+	nonce, _, err := computeSigningNonce(
+		*s.combinedNonce, s.ctx.combinedKey.FinalKey, msg,
+	)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// The adaptor point added to the combined nonce must be even.
+	// We keep generating random secrets until we find one that works.
```
guggero, line 598: Do we need to do this in a loop? Can't we just take the inverse of the secret to get an even y coordinate if the first draw produces an odd one?

btcec/schnorr/musig2/context.go
```diff
@@ -548,6 +573,129 @@ func (s *Session) RegisterPubNonce(nonce [PubNonceSize]byte) (bool, error) {
 	return haveAllNonces, nil
 }
 
+// GenerateAdaptorSecret generates an adaptor secret key. This must be called
+// after the public nonces have been registered for all signers, because
+// the validity of the adaptor secret key depends on the combined nonce. It
+// must also be called before any partial signatures are generated or provided,
+// because the validation of the signatures depends on the adaptor point.
+func (s *Session) GenerateAdaptor(msg [32]byte) (*btcec.ModNScalar, *btcec.JacobianPoint, error) {
```
guggero, line 581: nit: doc was added after this PR was opened, but we try to wrap before 80 characters to help with review and readability in narrow/multi-column environments: https://github.com/btcsuite/btcd/blob/master/docs/code_formatting_rules.md#80-character-line-length

btcec/schnorr/musig2/context.go
```diff
@@ -548,6 +573,129 @@ func (s *Session) RegisterPubNonce(nonce [PubNonceSize]byte) (bool, error) {
 	return haveAllNonces, nil
 }
 
+// GenerateAdaptorSecret generates an adaptor secret key. This must be called
```
guggero, line 576: nit: Godoc comment/method name mismatch.

btcec/schnorr/musig2/context.go
```diff
@@ -548,6 +573,129 @@ func (s *Session) RegisterPubNonce(nonce [PubNonceSize]byte) (bool, error) {
...
+	// The adaptor point added to the combined nonce must be even.
+	// We keep generating random secrets until we find one that works.
+	for {
+		var secretB [32]byte
+		_, err := rand.Read(secretB[:])
+		if err != nil {
+			return nil, nil, err
+		}
+
+		var secret btcec.ModNScalar
```
guggero, line 606: This seems duplicated with `SetAdaptor`. Couldn't we instead just call into that method until we don't get `ErrInvalidAdaptorPoint`? Should reduce the size of the diff quite a bit.

btcec/schnorr/musig2/sign.go
```diff
@@ -320,6 +332,19 @@ func Sign(secNonce [SecNonceSize]byte, privKey *btcec.PrivateKey,
 		return nil, err
 	}
 
+	nonce.ToAffine()
+
+	var adaptedNonce *btcec.JacobianPoint
```
guggero, line 337: How about `adaptedNonce := nonce` to avoid `else` case? Same for `verifyPartialSig`.

btcec/schnorr/musig2/sign_test.go
```diff
@@ -389,3 +390,119 @@ func TestMusig2SignCombine(t *testing.T) {
 		})
 	}
 }
+
+func TestMusig2SignCombineAdaptor(t *testing.T) {
+	// Pre-generate 10 deterministic private keys
```
guggero, line 395: nit: comments should always be full sentences, including punctuation.

btcec/schnorr/musig2/sign.go
```diff
@@ -606,6 +635,8 @@ type combineOptions struct {
 	combinedKey *btcec.PublicKey
 
 	tweakAcc *btcec.ModNScalar
+
+	adaptorPub *btcec.JacobianPoint
```
guggero, line 639: Where is this actually used? It seems like you can set it with `WithAdaptorCombine`, but don't see it being read anywhere. Which indicates to me that we probably want a unit test case for this as well.

guggero (commented):
Thank you for the PR!
Did a first pass, mostly to load some context. Didn't do a deep dive into the cryptography part of it yet, need to refresh my mental model on that first.

btcec/schnorr/musig2/sign.go
martonp: You're right this is unnecessary. I've removed it along with `WithAdaptorCombine`.

btcec/schnorr/musig2/context.go
guggero: Yeah, I understand that. But the evenness should be deterministic. So wouldn't it be possible to just invert the secret if we find out that (after adding the adaptor point to the nonce) it's not even, we can invert it instead of drawing a new value?
It's just that a for loop with no defined exit condition feels suboptimal.

martonp: Sorry, I completely missed your message.

Inverting the secret will not work, you can try this test function here:

``` golang

func TestGenerateAdaptor(t *testing.T) {
	var nonceB [32]byte
	_, err := rand.Read(nonceB[:])
	if err != nil {
		t.Fatalf("failed to generate random nonce: %v", err)
	}

	nonceScalar := btcec.ModNScalar{}
	nonceScalar.SetBytes(&nonceB)
	nonce := btcec.JacobianPoint{}
	btcec.ScalarBaseMultNonConst(&nonceScalar, &nonce)

	createAdaptedNonce := func(secret btcec.ModNScalar) btcec.JacobianPoint {
		var adaptorPoint btcec.JacobianPoint
		btcec.ScalarBaseMultNonConst(&secret, &adaptorPoint)

		combinedNonce := btcec.JacobianPoint{}
		btcec.AddNonConst(&adaptorPoint, &nonce, &combinedNonce)
		combinedNonce.ToAffine()

		return combinedNonce
	}

	unchanged := 0

	for i := 0; i < 1000; i++ {
		var secretB [32]byte
		_, err := rand.Read(secretB[:])
		if err != nil {
			t.Fatalf("failed to generate random secret: %v", err)
		}
		secret := btcec.ModNScalar{}
		secret.SetBytes(&secretB)

		combinedNonceBeforeNegation := createAdaptedNonce(secret)
		secret.Negate()
		combinedNonceAfterNegation := createAdaptedNonce(secret)

		if combinedNonceBeforeNegation.Y.IsOdd() == combinedNonceAfterNegation.Y.IsOdd() {
			unchanged++
		}
	}

	t.Logf("Y-coordinate evenness unchanged %d times", unchanged)
}
```

There's no deterministic effect of the evenness of one of the points to the combined point. Maybe just incrementing the secret until we find a valid one, rather than choosing a new random number, would be more performant, but I'm not aware of a technique to avoid trying different points until we encounter one that works.

PR #2346, opened 2025-03-25
utxocache: replace mapslice with concurrent-swiss-map

rabbitprincess:
## Change Description
Currently, `utxocache` uses `mapslice` instead of a single large `map`. This seems to be necessary because Go's built-in `map` can consume a significant amount of memory when scaled.
However, I suggest using [ConcurrentSwissMap](https://github.com/mhmtszr/concurrent-swiss-map) as an alternative, which offers simple implementation and better performance in concurrent reads.

## Benchmark

**Environment**  
`GOOS`: linux  
`GOARCH`: amd64  
`CPU`: AMD EPYC 7571  
`PKG`: github.com/btcsuite/btcd/blockchain  

| Type              | Put                                  | Get                              | Delete                             |
|-------------------|------------------------|-------------------------|-----------------------------|
| MapSlice          | 209.5 ns/op      | 153.8 ns/op      | 99.15 ns/op      |
| Csmap             | 197.7 ns/op      | 252.5 ns/op      | 249.4 ns/op     |
| MapSlice (8 Core) | 539.7 ns/op      | 428.0 ns/op   | 287.6 ns/op     |
| Csmap (8 Core)    | 430.9 ns/op      | 70.30 ns/op  | 482.8 ns/op       |

Csmap outperforms the lock-based mapslice by over 6x in concurrent reads.
You can verify the performance using the [mapslice](https://github.com/rabbitprincess/btcd/blob/45313e138363c200938be118890bde4d79cf9c39/blockchain/utxocache_test.go#L198) and [concurrentSwissMap](https://github.com/rabbitprincess/btcd/blob/12c43509f38da3d41662a0da4cf5801beff5a7c2/blockchain/utxocache_test.go#L184) benchmarks.

## Steps to Test

### Testing

- [ ] The change is not [insubstantial](https://github.com/btcsuite/btcd/master/docs/code_contribution_guidelines.md#substantial-contributions-only). Typo fixes are not accepted to fight bot spam.

### DIFF
--- a/blockchain/utxocache.go
+++ b/blockchain/utxocache.go
@@ -7,7 +7,6 @@ package blockchain
 import (
 	"container/list"
 	"fmt"
-	"sync"
 	"time"
 
 	"github.com/btcsuite/btcd/btcutil"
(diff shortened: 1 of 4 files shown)


guggero:
Mostly curious: How does the concurrent-swiss-map in the library you used differ from the native Swiss Tables that are now implemented in Go 1.24?

Roasbeef:
> Csmap outperforms the lock-based mapslice by over 6x in concurrent reads.

What about allocs/op?

blockchain/utxocache.go
```diff
@@ -7,175 +7,64 @@ package blockchain
...
-		return // Return as we were successful in adding the entry.
-	}
-
-	// We only reach this code if we've failed to insert into the map above as
-	// all the current maps were full.  We thus make a new map and insert into
-	// it.
-	m := ms.makeNewMap(totalEntryMemory)
-	m[op] = entry
+func (ms *csMap) size() int {
+	return calculateRoughMapSize(ms.length(), bucketSize)
```
Roasbeef, line 44: I think we also need to modify `calculateRoughMapSize`, as it makes assumptions w.r.t the current runtime implementation. With the introduction of Swiss maps in Go 1.24, this is likely out of date anyway (results in incorrect estimates).

rabbitprincess:
> Mostly curious: How does the concurrent-swiss-map in the library you used differ from the native Swiss Tables that are now implemented in Go 1.24?

It is basically same because both project based by [dolthub-swissmap](https://github.com/dolthub/swiss).
However, concurrent-swiss-map supports concurrency by sharding the map, with each shard having its own lock. This allows fast parallel reads as they often avoid locks entirely.

blockchain/utxocache.go
kcalvinalvin: This right? 

https://go.dev/blog/swisstable

Maybe the map calculation might be easier now based on how the blog described it. We'll see...

kcalvinalvin:
Concurrent reads/writes are not much of a problem for performance since the block validation
doesn't really have concurrent reads/writes. You start having that once you're synced to the tip
and need to fetch utxos for tx validation.

I think getting rid of mapslice is good since it's ugly. But I don't think the external dep makes
sense for consensus critical code.

Since the builtin map has multiple swiss tables, maybe we can just replace the entire mapslice
with just a single map and update the map size calculation code?

rabbitprincess:
@kcalvinalvin To use swissmap, it need to update go.mod version to 1.24. Also, using concurrentSwissMap is a better option than default swissmap since it avoids global locks.

kcalvinalvin:
> @kcalvinalvin To use swissmap, it need to update go.mod version to 1.24.

@Roasbeef?

> Also, using concurrentSwissMap is a better option than default swissmap since it avoids global locks.

Fair point but I think avoiding an external dependency is better than avoiding global locks.

kmk142789 (approved):

PR #2358, opened 2025-04-14
database: shut the program down immediately if we run out of disk space

kcalvinalvin:
## Change Description
Currently btcd keeps downloading blocks and fails to verify them if the disk is out of space. We exit immediately if we detect that the disk is out of space to ensure the database is at a recoverable state later on.

## Steps to Test
Tested by using a macbook with full disk and checking if the program shuts down as expected. Verified that it's also able to recover when given enough storage again.

### Testing

### DIFF
--- a/database/ffldb/blockio.go
+++ b/database/ffldb/blockio.go
@@ -10,6 +10,7 @@ package ffldb
 import (
 	"container/list"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"hash/crc32"
 	"io"
@@ -19,6 +20,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/database"
@@ -402,6 +404,12 @@ func (s *blockStore) writeData(data []byte, fieldName string) error {
 	n, err := wc.curFile.file.WriteAt(data, int64(wc.curOffset))
 	wc.curOffset += uint32(n)
 	if err != nil {
+		if errors.Is(err, syscall.ENOSPC) {
+			log.Errorf("%v. Cannot save any more blocks "+
+				"due to the disk being full "+
+				"-- exiting", err)
+			os.Exit(1)
+		}
 		str := fmt.Sprintf("failed to write %s to file %d at "+
 			"offset %d: %v", fieldName, wc.curFileNum,
 			wc.curOffset-uint32(n), err)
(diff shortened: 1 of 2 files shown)


database/ffldb/blockio.go
```diff
@@ -402,6 +403,15 @@ func (s *blockStore) writeData(data []byte, fieldName string) error {
 	n, err := wc.curFile.file.WriteAt(data, int64(wc.curOffset))
 	wc.curOffset += uint32(n)
 	if err != nil {
+		if pathErr, ok := err.(*os.PathError); ok {
+			if errno, ok := pathErr.Err.(syscall.Errno); ok &&
+				errno == syscall.ENOSPC {
+
+				log.Errorf("%v. Cannot save any more blocks "+
```
yyforyongyu, line 410: can we also mention it's due to the disk being full so the operator can act accordingly?

kcalvinalvin: Addressed in the latest commit

yyforyongyu (approved):
LGTM🌺

database/ffldb/blockio.go
```diff
@@ -402,6 +403,16 @@ func (s *blockStore) writeData(data []byte, fieldName string) error {
 	n, err := wc.curFile.file.WriteAt(data, int64(wc.curOffset))
 	wc.curOffset += uint32(n)
 	if err != nil {
+		if pathErr, ok := err.(*os.PathError); ok {
+			if errno, ok := pathErr.Err.(syscall.Errno); ok &&
+				errno == syscall.ENOSPC {
```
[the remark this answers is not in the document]
kcalvinalvin: Updated to use errors.Is

guggero:
Just a side note here: I'm not super familiar with the internals of `btcd`, so not sure if the same applies here. But in `lnd` we avoid using `os.Exit()` because that prevents any `defer` cleanup statements to run (e.g. closing network or database connections, flushing stuff to disk and so on).

That's why we're using a [`ShutdownLogger`](https://github.com/lightningnetwork/lnd/blob/852a8d8746c371151b215e9e9f79825ff93a5b85/build/sub_logger.go#L53) instead, which causes the main Goroutine to initiate a clean shutdown whenever a message with the level `Critical` is logged.
Might be worth considering here.

kcalvinalvin:
> Just a side note here: I'm not super familiar with the internals of `btcd`, so not sure if the same applies here. But in `lnd` we avoid using `os.Exit()` because that prevents any `defer` cleanup statements to run (e.g. closing network or database connections, flushing stuff to disk, closing files and so on).
> 
> That's why we're using a [`ShutdownLogger`](https://github.com/lightningnetwork/lnd/blob/852a8d8746c371151b215e9e9f79825ff93a5b85/build/sub_logger.go#L53) instead, which causes the main Goroutine to initiate a clean shutdown whenever a message with the level `Critical` is logged. Might be worth considering here.

`btcd` has one database that handles everything and the flush happens here:
https://github.com/btcsuite/btcd/blob/cd05d9ad3d0597368adf95c54bdc530700393aed/btcd.go#L133-L137

That's a fair point and I did think about returning an error and letting the caller handle the out of disk error but ultimately thought this was the best way to handle things. The database only ever writes to the disk at the places this code change takes place and if that's full, there's nothing else to flush for the caller because anything that the caller tries to flush follows the same code path.

Roasbeef (approved):
LGTM 📮

PR #2362, opened 2025-04-19
Update validate.go

ta2ozg:
## Change Description
simplified previous long control structure to a single return statement and code became shorter and clearer (go).


saubyk:
Closing as this is not a substantial change.
https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md#substantial-contributions-only

PR #2368, opened 2025-05-10
Increase dust threshold multiplier

1uVU:
## Change Description
The purpose of this PR is to filter out more dust / nonmonetary transactions. The dust would be defined for outputs for which the cost to the network to spend coins is more than 1/9 of the minimum transaction relay fee.

### DIFF
--- a/mempool/policy.go
+++ b/mempool/policy.go
@@ -173,8 +173,8 @@ func checkPkScriptStandard(pkScript []byte, scriptClass txscript.ScriptClass) er
 }
 
 // GetDustThreshold calculates the dust limit for a *wire.TxOut by taking the
-// size of a typical spending transaction and multiplying it by 3 to account
-// for the minimum dust relay fee of 3000sat/kvb.
+// size of a typical spending transaction and multiplying it by 9 to account
+// for the minimum dust relay fee of 9000sat/kvb.
 func GetDustThreshold(txOut *wire.TxOut) int64 {
 	// The total serialized size consists of the output and the associated
 	// input script to redeem it.  Since there is no input script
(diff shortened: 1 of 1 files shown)


Roasbeef:
Modifying the existing value isn't reasonable as many higher level protocols such as LN rely on the ability to propagate transactions at the current dust level. If you were running an LN node against your node, and ran this patch, you'd find that you weren't able to force close at all. If you can't force close, then that means your funds my be lost in a breach event. 

I think the only thing that would make sense here is adding a config option to tweak this value. The public functions should be modified though, as callers expect stability from the return value of those functions (it only due to the witness scaling factor). 

As you as a user can end up with dust in your wallet even if you're only doing "monetary" transactions. If you have a small amount of change left over after a send, then that can be considered dust.

whileunless:
> Modifying the existing value isn't reasonable as many higher level protocols such as LN rely on the ability to propagate transactions at the current dust level. If you were running an LN node against your node, and ran this patch, you'd find that you weren't able to force close at all. If you can't force close, then that means your funds my be lost in a breach event.
> 
> I think the only thing that would make sense here is adding a config option to tweak this value. The public functions should be modified though, as callers expect stability from the return value of those functions (it only due to the witness scaling factor).

I have a nonrouting node on top of Bitcoin Knots. Bitcoin Core and Knots they have `dustrelayfee` config option. I am thinking about running a second lightning node with LND on top of BTCD, and such config option to tweak this value would be most welcome.

Roasbeef:
> I am thinking about running a second lightning node with LND on top of BTCD, and such config option to tweak this value would be most welcome.

Sure, want to update this PR to add a config option like that? As mentioned, all the other public functions in the sub-packages should remain unchanged. 

Also it's important that you understand this interaction: 
> Modifying the existing value isn't reasonable as many higher level protocols such as LN rely on the ability to propagate transactions at the current dust level. If you were running an LN node against your node, and ran this patch, you'd find that you weren't able to force close at all. If you can't force close, then that means your funds my be lost in a breach event.

If you set this value (once you add it to the code) to basically anything other than the current default, then you will willingly expose yourself to events that may cause **irrevocable loss of funds** on LN.

1uVU:
>  such config option to tweak this value would be most welcome.

Such config would be most welcome by me too. If it existed he would be no need for this PR. 

> Sure, want to update this PR to add a config option like that?

As of now I'm not quite versed in Go to code up such idea. I hope someone else will.

kmk142789 (approved):

PR #2370, opened 2025-05-14
btcjson: prevent infinite recursion in StringOrArray.MarshalJSON and …

### LINKED ISSUE #2369: [bug]: StringOrArray.MarshalJSON triggers infinite recursion → stack-overflow panic
### Background

`btcjson.StringOrArray` declares its own `MarshalJSON` method with a **value** receiver that simply calls `json.Marshal(h)`.
Because the value still satisfies `json.Marshaler`, `json.Marshal` re-invokes `MarshalJSON`, triggering infinite recursion and eventually a **stack‐overflow panic** at runtime.

This manifests anywhere a `StringOrArray` is marshalled (e.g. `GetNetworkInfoResult.Warnings`), crashing downstream apps that depend on `btcd/btcjson`.

### Your environment

* **btcd / btcjson version:** `v0.24.3-0.20250506233109-1eb974aab6ef`
* **Go version:** `go1.24.2`
* **OS:** `Darwin 23.4.0 arm64` (Apple Silicon macOS 14.4)
* Reproducible on Linux/amd64 as well.

### Steps to reproduce

```go
package main

import (
	"encoding/json"
)

func main() {
	val := btcjson.StringOrArray{"boom"}
	_, _ = json.Marshal(val) // panic: runtime: stack overflow
}
```

Or run any program that marshals `btcjson.GetNetworkInfoResult` when the
`warnings` field is non-nil (see attached stack trace snippet).

### Expected behaviour

`StringOrArray` should serialize to:

* `"foo"` when length == 1
* `["foo","bar"]` when length > 1

without panicking.

### Actual behaviour

`json.Marshal` enters an infinite loop inside `StringOrArray.MarshalJSON`, eventually crashing with:

```
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow
...
encoding/json.marshalerEncoder → btcjson.StringOrArray.MarshalJSON → json.Marshal → (repeat)
```

### Proposed fix (PR incoming)

Change the receiver to a **pointer** (or remove the method entirely) and add unit tests:

```go
func (h *StringOrArray) MarshalJSON() ([]byte, error) {
    return json.Marshal([]string(*h))
}
```

Tests confirming safe round-trips for single and multiple values are included.

cvieira-tng:
Fix https://github.com/btcsuite/btcd/issues/2369

## Change Description  
Fix an infinite-recursion bug in `btcjson.StringOrArray.MarshalJSON` that
caused a stack-overflow panic at runtime.  
* Switch receiver from value to pointer so we no longer call the same method
  on a **copy** of the slice (which re-enters indefinitely).  
* Add unit tests (`TestGetNetworkInfoWarnings`) to cover single-string and
  slice cases as well as malformed input.  
* No external API changes.  

Closes #<ISSUE-NUMBER>.

---

## Steps to Test  
1. **Build & test**  
   ```bash
   go test ./...          # all packages pass  
````

2. **Reproduce the old panic (optional)**

   ```bash
   git checkout HEAD^            # commit before this PR  
   go test -run TestGetNetworkInfoWarnings ./btcjson  # panic  
   ```

   Returning to this PR’s head makes the test suite pass again.

---

### Testing

* [x] Your PR passes all CI checks.
* [x] Tests covering the positive and negative (error paths) are included.
* [x] Bug fixes contain tests triggering the bug to prevent regressions.

* [x] The change is not
  [[insubstantial](https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md#substantial-contributions-only)](https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md#substantial-contributions-only).
* [x] The change obeys the
  [[Code Documentation and Commenting](https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md#code-documentation-and-commenting)](https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md#code-documentation-and-commenting)
  guidelines, and lines wrap at 80.
* [x] Commits follow the
  [[Ideal Git Commit Structure](https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md#model-git-commit-messages)](https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md#model-git-commit-messages).
* [x] Any new logging statements use an appropriate subsystem and logging
  level.


ja7ad (approved):

allocz (commented):
#2476 already handled this. Thanks for the PR, but I think it can be closed now.

PR #2371, opened 2025-05-15
mempool: check p2wsh standard

ChrisCho-H:
## Change Description

AFAIK, btcd curretly doesn't check standard rule for witness unlike [bitcoin core](https://github.com/bitcoin/bitcoin/blob/52ede28a8adb2c2d44d7f800bbfbef8aed86070e/src/policy/policy.cpp#L221). I add method `CheckWitnessStandard` to check standard rule for `p2wsh`, and execute it inside the method `validateStandardness` for mempool validation.

## Steps to Test
If this PR makes sense, I will add test vectors following the reviews.

### Testing

### DIFF
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -1639,6 +1639,22 @@ func (mp *TxPool) validateStandardness(tx *btcutil.Tx, nextBlockHeight int32,
 		return txRuleError(rejectCode, str)
 	}
 
+	// Check the witness standard.
+	err = checkWitnessStandard(tx, utxoView)
+	if err != nil {
+		// Attempt to extract a reject code from the error so it can be
+		// retained. When not possible, fall back to a non-standard
+		// error.
+		rejectCode, found := extractRejectCode(err)
+		if !found {
+			rejectCode = wire.RejectNonstandard
+		}
+		str := fmt.Sprintf("transaction %v has a non-standard "+
+			"witness: %v", tx.Hash(), err)
+
+		return txRuleError(rejectCode, str)
+	}
+
 	return nil
 }
 
--- a/mempool/policy.go
+++ b/mempool/policy.go
@@ -53,6 +53,15 @@ const (
 	// in a multi-signature transaction output script for it to be
 	// considered standard.
 	maxStandardMultiSigKeys = 3
+
+	// The maximum number of witness stack items in a standard P2WSH script.
+	maxStandardP2WSHStackItems = 100
+
+	// The maximum size in bytes of each witness stack item in a standard P2WSH script.
+	maxStandardP2WSHStackItemSize = 80
+
+	// The maximum size in bytes of a standard witnessScript.
+	maxStandardP2WSHScriptSize = 3600
 )
 
 // calcMinRequiredTxRelayFee returns the minimum transaction fee required for a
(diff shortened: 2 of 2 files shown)


mempool/policy.go
```diff
@@ -387,3 +396,51 @@ func GetTxVirtualSize(tx *btcutil.Tx) int64 {
...
+			}
+			// Check the size of witness stack, which excludes witness script.
+			if len(txIn.Witness)-1 > maxStandardP2WSHStackItems {
+				str := fmt.Sprintf("P2WSH witness stack items exceed max allowed: "+
+					"%d > %d items", len(txIn.Witness)-1, maxStandardP2WSHStackItems)
+				return txRuleError(wire.RejectNonstandard, str)
+			}
+			// Check the size of each of stack item except witness script.
+			for j, wit := range txIn.Witness {
+				if len(wit) > maxStandardP2WSHStackItemSize && j != len(txIn.Witness)-1 {
```
kcalvinalvin, line 436: No need for the j here or the branch checking for `j != len(txIn.Witness)-1`.

kcalvinalvin:
Verified that the added logic implements just the P2WSH standardness check in Core.

https://github.com/bitcoin/bitcoin/blob/725c9f7780e0def3d79be151c08a36fe7a9dc59c/src/policy/policy.cpp#L264-L275

It does seem like the original commit is very old. Not quite sure about adding it but am indifferent if it does get added. https://github.com/bitcoin/bitcoin/pull/8499

mempool/policy.go
ChrisCho-H: It's to exclude the last element of witness stack, which is witness script.
Core doesn't check the last element either.
https://github.com/bitcoin/bitcoin/blob/52ede28a8adb2c2d44d7f800bbfbef8aed86070e/src/policy/policy.cpp#L268-L271

mempool/policy.go
```diff
@@ -387,3 +396,51 @@ func GetTxVirtualSize(tx *btcutil.Tx) int64 {
 	return (blockchain.GetTransactionWeight(tx) + (blockchain.WitnessScaleFactor - 1)) /
 		blockchain.WitnessScaleFactor
 }
+
+// checkWitnessStandard follows the core policy.
+//
```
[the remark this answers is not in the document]
ChrisCho-H: Thanks! I added the comment.

Roasbeef:
>  Not quite sure about adding it but am indifferent if it does get added.

FWIW, there're quite a few newer standardness rules that `bitcoind` has added over the years that we never adopted. The new rules are generally a moving target, and don't have BIP implementations so all we have is the code to go for. 

IMO, the only standardness/policy rules that are actually important, are those that directly impact the relay of  time-sensitive transactions across the network. One such example is: TRUC and co.

ChrisCho-H:
[Original PR](https://github.com/bitcoin/bitcoin/pull/8499#issuecomment-252420342) of `bitcoind` explains the purpose of standardness/policy rules of p2wsh like following

- 3600 bytes witnessScript and 100 stack items are adequate for a n-of-100 multisig using OP_CHECKSIG, OP_ADD, and OP_EQUAL.
- The max size for ECDSA signature is 73 bytes and nothing should use more than that with the current scripting language.
- This is to prevent abuse of witness space, and reduce the risks of DoS attack with some unknown special and big scripts.

I agree that absence of this doesn't directly impact, but what explained in `bitcoind` also makes sense to mitigate the possible risks.

mempool/mempool.go
```diff
@@ -1639,6 +1639,22 @@ func (mp *TxPool) validateStandardness(tx *btcutil.Tx, nextBlockHeight int32,
 		return txRuleError(rejectCode, str)
 	}
 
+	// Check the witness standard.
+	err = checkWitnessStandard(tx, utxoView)
```
allocz, line 1643: Since this is policy and not consensus divergence, I don't see a valid reason for merging. Note that if we start rejecting arbitrary transactions, we may reach a point where we can't estimate fees using mempool data. I think that the best way to estimate if a transaction enters or not in the node mempool is the fee being paid by the transaction, and not arbitrary rules.

Issue #2372, opened 2025-05-16
[bug]: panic on `ReadTaprootBip32Derivation`


brunoerg:
When parsing a PSBT, just got a panic due to `makeslice: len out of range` in `ReadTaprootBip32Derivation`. Perhaps due to `LeafHashes: make([][]byte, int(numHashes))`: https://github.com/btcsuite/btcd/blob/1eb974aab6ef11097571f9517b4b6b0c639ab63b/btcutil/psbt/taproot.go#L118-L133

Test to reproduce:

```go
func TestPSBTCrash(t *testing.T) {
	hexString := "70736274ff01007374ff01030100000000002f0000002e2873007374ff01070100000000000000000000000000000000000000060680050000736274ff01000a0000000060c70006060000736274ff01000a000001000001002407010000000000000000000000000000000000000006060000736274ff01000a0000000000010024c760002a707362c760000b0500000000000000060605000073626274ff01000a00000000000100242121212121212121212121212121212121212121212121212121212121212121212121212107010000000000000000000000000000000000000006060000736274ff01000a000eff000001000a0a040404040404040400"
	// Convert hex string to byte slice
	buffer, err := hex.DecodeString(hexString)
	if err != nil {
		log.Fatalf("Failed to decode hex: %v", err)
	}
	psbt, err := NewFromRawBytes(bytes.NewBuffer(buffer), false)
	t.Logf("got transaction: %v", spew.Sdump(psbt.UnsignedTx))
}
```

PR #2374, opened 2025-05-18
psbt: overflow checks when computing Taproot BIP32 derivation min size

ffranr:
Closes https://github.com/btcsuite/btcd/issues/2372

Thank you to @brunoerg for opening the issue.

---

Protect against overflows when parsing malformed Taproot BIP32 derivation fields. This ensures that deserialization fails safely if the declared number of leaf hashes would otherwise cause an integer overflow.

### DIFF
--- a/btcutil/psbt/taproot.go
+++ b/btcutil/psbt/taproot.go
@@ -143,6 +143,13 @@ func minTaprootBip32DerivationByteSize(numHashes uint64) (uint64, error) {
 func ReadTaprootBip32Derivation(xOnlyPubKey,
 	value []byte) (*TaprootBip32Derivation, error) {
 
+	// This function allocates additional memory while parsing the serialized
+	// data. To prevent potential out-of-memory (OOM) issues, we must validate
+	// the length of the value slice before proceeding.
+	if len(value) > MaxPsbtValueLength {
+		return nil, ErrInvalidPsbtFormat
+	}
+
 	// The taproot key BIP 32 derivation path is defined as:
 	//   <hashes len> <leaf hash>* <4 byte fingerprint> <32-bit uint>*
 	// So we get at least 5 bytes for the length and the 4 byte fingerprint.
--- a/btcutil/psbt/taproot.go
+++ b/btcutil/psbt/taproot.go
@@ -2,6 +2,8 @@ package psbt
 
 import (
 	"bytes"
+	"math"
+	"math/bits"
 
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/btcsuite/btcd/txscript"
@@ -93,6 +95,48 @@ func (s *TaprootBip32Derivation) SortBefore(other *TaprootBip32Derivation) bool 
 	return bytes.Compare(s.XOnlyPubKey, other.XOnlyPubKey) < 0
 }
 
+// minTaprootBip32DerivationByteSize returns the minimum number of bytes
+// required to encode a Taproot BIP32 derivation field, given the number of
+// leaf hashes.
+//
+// NOTE: This function does not account for the size of the BIP32 child indexes,
+// as we are only computing the minimum size (which occurs when the path is
+// empty). The bits package is used to safely detect and handle overflows.
+func minTaprootBip32DerivationByteSize(numHashes uint64) (uint64, error) {
+	// The Taproot BIP32 derivation field is encoded as:
+	//   [compact size uint: number of leaf hashes]
+	//   [N × 32 bytes: leaf hashes]
+	//   [4 bytes: master key fingerprint]
+	//   [M × 4 bytes: BIP32 child indexes]
+	//
+	// To compute the minimum size given the number of hashes only, we assume:
+	// - N = numHashes (provided)
+	// - M = 0 (no child indexes)
+	//
+	// So the base byte size is:
+	//   1 (leaf hash count) + (N × 32) + 4 (fingerprint)
+	//
+	// First, we calculate the total number of bytes for the leaf hashes.
+	mulCarry, totalHashesBytes := bits.Mul64(numHashes, 32)
+	if mulCarry != 0 {
+		return 0, ErrInvalidPsbtFormat
+	}
+
+	// Since we're computing the minimum possible size, we add a constant that
+	// accounts for the fixed size fields:
+	// * 1 byte for the compact size leaf hash count (assumes numHashes < 0xfd)
+	// * 4 bytes for the master key fingerprint
+	// Total: 5 bytes.
+	// All other fields (e.g., BIP32 path) are assumed absent for minimum size
+	// calculation.
+	result, addCarry := bits.Add64(5, totalHashesBytes, 0)
+	if addCarry != 0 {
+		return 0, ErrInvalidPsbtFormat
+	}
+
+	return result, nil
+}
+
 // ReadTaprootBip32Derivation deserializes a byte slice containing the Taproot
 // BIP32 derivation info that consists of a list of leaf hashes as well as the
 // normal BIP32 derivation info.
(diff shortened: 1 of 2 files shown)


ffranr:
An alternative safeguard would be to explicitly cap `numHashes` at some chosen maximum. However, I couldn't find a BIP-defined upper bound for the number of leaf hashes.

Given that, I think it's more appropriate to rely on the natural limit imposed by `uint64` overflow rather than introduce an arbitrary cap. This keeps the implementation simple and avoids enforcing assumptions not defined by the spec.

btcutil/psbt/taproot.go
```diff
@@ -93,6 +94,43 @@ func (s *TaprootBip32Derivation) SortBefore(other *TaprootBip32Derivation) bool
...
+	//
+	// To compute the minimum size, we assume:
+	// - N = numHashes (provided)
+	// - M = 0 (no child indexes)
+	//
+	// So the base size is:
+	//   1 (leaf hash count) + (N × 32) + 4 (fingerprint) + 1 (empty path count)
+	//
+	// First, we calculate the total number of bytes for the leaf hashes.
+	totalHashesBytes, mulCarry := bits.Mul64(numHashes, 32)
```
yyforyongyu, line 120: Q: when will the `mulCarry` or the `addCarry` be non-zero? Why is this method preferred than, say, just calculating `32 * numHashes + 6`?

btcutil/psbt/taproot.go
```diff
@@ -113,9 +151,15 @@ func ReadTaprootBip32Derivation(xOnlyPubKey,
 		return nil, ErrInvalidPsbtFormat
 	}
 
-	// A hash is 32 bytes in size, so we need at least numHashes*32 + 5
-	// bytes to be present.
-	if len(value) < (int(numHashes)*32)+5 {
+	// Given the number of hashes, we can calculate the minimum byte size
+	// of the taproot BIP32 derivation.
+	minByteSize, err := minTaprootBip32DerivationByteSize(numHashes)
```
guggero, line 156: It looks to me like all we need to do is make sure `numHashes` isn't negative and isn't [greater than 128](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#cite_note-6).

btcutil/psbt/taproot.go
```diff
@@ -93,6 +94,43 @@ func (s *TaprootBip32Derivation) SortBefore(other *TaprootBip32Derivation) bool
...
+	//   1 (leaf hash count) + (N × 32) + 4 (fingerprint) + 1 (empty path count)
+	//
+	// First, we calculate the total number of bytes for the leaf hashes.
+	totalHashesBytes, mulCarry := bits.Mul64(numHashes, 32)
+	if mulCarry != 0 {
+		return 0, ErrInvalidPsbtFormat
+	}
+
+	// Add 6 bytes: 1 (leaf hash count) + 4 (fingerprint) + 1 (path length)
+	result, addCarry := bits.Add64(totalHashesBytes, 6, 0)
```
guggero, line 126: The path length is defined as `<32-bit little endian uint path element>*`, meaning zero paths is acceptable. That's why `6` is wrong here.

guggero (commented):
Thanks for the fix. Though I think this can be simplified quite a bit.

btcutil/psbt/taproot.go
ffranr: Since `numHashes` is `uint64`, `numHashes * 32` fits in 64 bits unless `numHashes > 2^64 / 32`. `mulCarry` will be non-zero only in that overflow case. Using `bits.Mul64` is safer and explicit about overflow handling.

btcutil/psbt/taproot.go
ffranr: I don't think a hash count limit of 128 would be correct here.

The 128-figure in BIP-341 is a limit on path depth, not on how many different leaves a key (and therefore a `TaprootBip32Derivation`) may reference.

As far as I understand, in a script-path spend the control-block length is `33 + 32 m` with `0 <= m <= 128`; this caps the number of sibling hashes on any single proof-of-inclusion to 128.

However, the `TaprootBip32Derivation` spec gives no upper bound on count. Therefore, the derivation could include 10,000 or more distinct leaves, each at depth <= 128; a PSBT may list all of them. Therefore imposing a consensus-style 128 cap on the number of hashes would incorrectly reject valid PSBTs.

We've seen elsewhere that hardware wallets, for example, often require all spending paths to be explicitly described, not just a single one. Ultimately, a valid `TaprootBip32Derivation` could reference more than 128 leaves.

guggero: Ah, yes, I indeed looked at the wrong value. You are correct that the number of possible leaves is $2^127$ and not 127 (I was thinking of sibling hashes in the merkle proof of the control block).

But I think we can still limit the number here to avoid an `int64` overflow. For example, if we allow up to `math.MaxUint32` (4.29 billion) leaf hashes, that alone would be 137 GiB of data, just for the leaf hashes. Do we really expect that to ever be processed using a PSBT?

ffranr: Yes I see that we could limit to`math.MaxUint32` or something like that. Here I wrote that we could just let the overflow protection on `uint64` set the limit: https://github.com/btcsuite/btcd/pull/2374#issuecomment-2889232620

ffranr: In the latest commit, `numHashes` is now constrained to `< math.MaxUint32`.

I’ve retained the overflow guard in `minTaprootBip32DerivationByteSize`, as IMO arithmetic should generally be overflow‑checked, barring rare exceptions.

guggero: Actually, we have `const MaxPsbtValueLength = 4000000`, so we can constrain the num hashes to a much lower value. Which we should definitely do, now that I thought about it a bit more. Since otherwise I can OOM your machine by preparing a specially crafted PSBT.
So I think we should limit to `MaxPsbtValueLength / 32` which is still more than will be possible to use.

If you feel so strongly about using safe arithmetics here, then I think we should create a `SafeAdd[T constraints.Integer](a, b T) (T, error)` and `SafeMultiply` so we can re-use those.

ffranr: > I can OOM your machine

In `ReadTaprootBip32Derivation`, the serialized `value` is already in memory. The function copies out the leaf hashes and BIP32 path indexes into new slices, but all allocations are bounded by `len(value)` — the total memory usage is still `O(len(value))`.

If we're concerned about OOM, the key thing to constrain is `len(value)`, not `numHashes`, since:

* The number of hashes is encoded *within* `value`, and
* The allocation for `LeafHashes` is directly proportional to that.

Is it possible to construct a specially crafted PSBT that causes `ReadTaprootBip32Derivation` to run with worse than `O(len(value))` memory?

Most of our code operates with linear memory complexity at best, so I don’t think linear allocation in `ReadTaprootBip32Derivation` is particularly concerning.

> If you feel so strongly about using safe arithmetics here, then I think we should create a SafeAdd[T constraints.Integer](a, b T) (T, error) and SafeMultiply so we can re-use those.

Yes, I was thinking of adding that sort functionality somewhere general. We use `bits.Add64` in tap also for the overflow check. I would like to do that in a separate PR. I think It's important that `minTaprootBip32DerivationByteSize` is safe on its own and doesn't rely on the caller to use it correctly.

btcutil/psbt/taproot.go
```diff
@@ -113,9 +157,22 @@ func ReadTaprootBip32Derivation(xOnlyPubKey,
...
+
+	// Given the number of hashes, we can calculate the minimum byte size
+	// of the taproot BIP32 derivation.
+	minByteSize, err := minTaprootBip32DerivationByteSize(numHashes)
+	if err != nil {
+		return nil, err
+	}
+
+	// Ensure that value is at least the minimum size.
+	if uint64(len(value)) < minByteSize {
```
guggero, line 175: Okay, I guess this check prevents us from OOMing if the encoded length is larger from the number of bytes we actually have available.

guggero (approved):
LGTM :tada:

btcutil/psbt/taproot.go
ffranr: I think you reviewed before I managed to push my latest change. I've added:
```
	// This function allocates additional memory while parsing the serialized
	// data. To prevent potential out-of-memory (OOM) issues, we must validate
	// the length of the value slice before proceeding.
	if len(value) > MaxPsbtValueLength {
	}
```

let me know if that works for you

btcutil/psbt/taproot.go
```diff
@@ -143,6 +143,13 @@ func minTaprootBip32DerivationByteSize(numHashes uint64) (uint64, error) {
 func ReadTaprootBip32Derivation(xOnlyPubKey,
 	value []byte) (*TaprootBip32Derivation, error) {
 
+	// This function allocates additional memory while parsing the serialized
```
yyforyongyu, line 146: 👍

PR #2380, opened 2025-05-22
mempool: Correctly identify all OP_RETURN outputs as NulldataTy

benthecarman:
## Change Description

Previously OP_RETURN outputs could be identified as `NonStandardTy` if they were over the `MaxDataCarrierSize`. This would cause in other parts of the code where OP_RETURN outputs would be handled weirdly (as dust and other things).

This changes it so it is correctly identified but still marks a transaction non-standard if the output violates `MaxDataCarrierSize`

## Steps to Test

### Testing

### DIFF
--- a/txscript/standard.go
+++ b/txscript/standard.go
@@ -505,7 +505,7 @@ func isNullDataScript(scriptVersion uint16, script []byte) bool {
 	//  OP_RETURN <optional data>
 	//
 	// Thus, it can either be a single OP_RETURN or an OP_RETURN followed by a
-	// data push up to MaxDataCarrierSize bytes.
+	// data push.
 
 	// The script can't possibly be a null data script if it doesn't start
 	// with OP_RETURN.  Fail fast to avoid more work below.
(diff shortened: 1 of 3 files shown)


txscript/standard.go
```diff
@@ -521,8 +521,7 @@ func isNullDataScript(scriptVersion uint16, script []byte) bool {
 	// OP_RETURN followed by data push up to MaxDataCarrierSize bytes.
 	tokenizer := MakeScriptTokenizer(scriptVersion, script[1:])
 	return tokenizer.Next() && tokenizer.Done() &&
-		(IsSmallInt(tokenizer.Opcode()) || tokenizer.Opcode() <= OP_PUSHDATA4) &&
```
Roasbeef, line 524: So this has some implications w/ the rbf coop close spec in LN, and the spec uses a strict interpretation that it's exactly no more than 80 bytes (not even multiple pushes are allowed as bitcoind does today). 

We already have a flag that removes all the runtime policy checks, so perhaps that's sufficient for your use case? 

https://github.com/btcsuite/btcd/blob/1eb974aab6ef11097571f9517b4b6b0c639ab63b/config.go#L157

benthecarman: My main motivation was a way around https://github.com/btcsuite/btcwallet/pull/907 because in btcwallet it considers op_returns > 80 bytes to be dust rather than op_returns

PR #2385, opened 2025-06-08
Remove datacarriersize and op_return output count limit

niteshbalusu11:
## Change Description
This PR pretty much does the same changes as the instagibbs PR for bitcoin core.

https://github.com/bitcoin/bitcoin/pull/32406

- Removed datacarriersize limit
- Removed the restriction for max op_return output count

Though bitcoin core is marking datacarriersize as depracated, btcd never supported increasing the size with a flag hence I am getting rid of it altogether.

## Steps to Test

### Testing

### DIFF
--- a/txscript/standard.go
+++ b/txscript/standard.go
@@ -13,10 +13,6 @@ import (
 )
 
 const (
-	// MaxDataCarrierSize is the maximum number of bytes allowed in pushed
-	// data to be considered a nulldata transaction
-	MaxDataCarrierSize = 80
-
 	// StandardVerifyFlags are the script flags which are used when
 	// executing transaction scripts to enforce additional checks which
 	// are required for the script to be considered standard.  These checks
@@ -518,11 +514,10 @@ func isNullDataScript(scriptVersion uint16, script []byte) bool {
 		return true
 	}
 
-	// OP_RETURN followed by data push up to MaxDataCarrierSize bytes.
+	// OP_RETURN followed by a data push.
 	tokenizer := MakeScriptTokenizer(scriptVersion, script[1:])
 	return tokenizer.Next() && tokenizer.Done() &&
-		(IsSmallInt(tokenizer.Opcode()) || tokenizer.Opcode() <= OP_PUSHDATA4) &&
-		len(tokenizer.Data()) <= MaxDataCarrierSize
+		(IsSmallInt(tokenizer.Opcode()) || tokenizer.Opcode() <= OP_PUSHDATA4)
 }
 
 // scriptType returns the type of the script being inspected from the known
@@ -883,15 +878,8 @@ func PayToAddrScript(addr btcutil.Address) ([]byte, error) {
 }
 
 // NullDataScript creates a provably-prunable script containing OP_RETURN
-// followed by the passed data.  An Error with the error code ErrTooMuchNullData
-// will be returned if the length of the passed data exceeds MaxDataCarrierSize.
+// followed by the passed data.
 func NullDataScript(data []byte) ([]byte, error) {
-	if len(data) > MaxDataCarrierSize {
-		str := fmt.Sprintf("data size %d is larger than max "+
-			"allowed size %d", len(data), MaxDataCarrierSize)
-		return nil, scriptError(ErrTooMuchNullData, str)
-	}
-
 	return NewScriptBuilder().AddOp(OP_RETURN).AddData(data).Script()
 }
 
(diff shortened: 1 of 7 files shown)


txscript/standard.go
```diff
@@ -518,11 +514,10 @@ func isNullDataScript(scriptVersion uint16, script []byte) bool {
 		return true
 	}
 
-	// OP_RETURN followed by data push up to MaxDataCarrierSize bytes.
+	// OP_RETURN followed by a data push.
 	tokenizer := MakeScriptTokenizer(scriptVersion, script[1:])
 	return tokenizer.Next() && tokenizer.Done() &&
-		(IsSmallInt(tokenizer.Opcode()) || tokenizer.Opcode() <= OP_PUSHDATA4) &&
-		len(tokenizer.Data()) <= MaxDataCarrierSize
+		(IsSmallInt(tokenizer.Opcode()) || tokenizer.Opcode() <= OP_PUSHDATA4)
```
darosior, line 520: Not familiar with btcd's code but this appears to mandate a single data push to follow the `OP_RETURN` for it to be standard. This is not Bitcoin Core's behaviour. If you want to match the behaviour from https://github.com/bitcoin/bitcoin/pull/32406 then i think just unconditionally `return`ing `len(script) >= 1 && script[0] == OP_RETURN` on line 508 above is all you need.

saubyk:
Don't mean to rehash the whole debate here, but why should this implementation follow the policy recommendations of Bitcoin core?

txscript/standard.go
niteshbalusu11: Could you please point me to the place in bitcoin core's PR where a push is not required after op_return?

niteshbalusu11:
> Don't mean to rehash the whole debate here, but why should this implementation follow the policy recommendations of Bitcoin core?

It doesn't have to. But it'll be nice to have the same standard.

txscript/standard.go
ajtowns: The requirement is that the script is push only apart from the OP_RETURN; that's any number of pushes, possibly zero. It's been this way since https://github.com/bitcoin/bitcoin/pull/6424 

https://github.com/bitcoin/bitcoin/blob/master/src/script/solver.cpp#L185-L187

yuvicc:
Concept ACK on this!

djkazic:
Concept ACK. Core just merged the PR (https://github.com/bitcoin/bitcoin/commit/f3bbc746647d1fd23bf5cfe357e32f38c5f6319c).

Having policy that aligns with what is mined will reduce the number of round trips fetching transactions during block sync

txscript/standard.go
Roasbeef: FWIW, we haven't kept up with `bitcoind` discretionary policy for quite some time now (many years). There're no BIPS for policies like this, so it makes it hard to follow, and if it doesn't need interop, then there isn't a strong reason to copy behavior. There's a PR open to update it to allow multiple pushes. But in LN, with the new rbf close flow, `OP_RETURN` is defined as it is here: the more restrictive version.

Roasbeef:
As I mentioned in [this comment](https://github.com/btcsuite/btcd/pull/2385#discussion_r2136152656) we haven't kept up with bitcoind's discretionary relay policy for [many years now](https://github.com/btcsuite/btcd/pull/2371#issuecomment-2896034480) (same goes for the dozens of JSON-RPC changes, just look at the issue/PR tracker). This type of policy doesn't necessarily desire wide spread interop. The fact that there aren't BIPs for such policies makes this clear. Without clear BIPs, we need to closely monitor all the releases for relevant changes to adopt, which would further strain the small set of contributors that actually reviews and tests PRs. 

The only relay policies that we endeavor to keep up with are those that require interop, and tangibly affect higher layer solutions such as: RBF, TRUC, etc. These policies all have BIPs, with a clear target scope and motivation. 

We already have a flag that enables users to [accept and relay all non-standard transactions](https://github.com/btcsuite/btcd/blob/ac58f6984cec305069c399d2e4e27ae742078aaa/sample-btcd.conf#L272-L276). Users that want to shed themselves of all discretionary relay policy can run with those flags.

For any prospective contributors that are interested in contributing to the project, please check out the set of open PRs and issues if you wish to productively contribute to the project. [Here's an impactful PR](https://github.com/btcsuite/btcd/pull/2226) that could use more review and testing.

niteshbalusu11:
Thank you for the reply, closing this PR.

PR #2389, opened 2025-06-23
Minor refactor to improve code health

yyforyongyu:
This PR introduces some refactoring to methods like `OnGetData` and `BtcEncode`, aiming to improve readability and code health.

### DIFF
--- a/server.go
+++ b/server.go
@@ -694,21 +694,24 @@ func (sp *serverPeer) OnGetData(_ *peer.Peer, msg *wire.MsgGetData) {
 	notFound := wire.NewMsgNotFound()
 
 	length := len(msg.InvList)
-	// A decaying ban score increase is applied to prevent exhausting resources
-	// with unusually large inventory queries.
-	// Requesting more than the maximum inventory vector length within a short
-	// period of time yields a score above the default ban threshold. Sustained
-	// bursts of small requests are not penalized as that would potentially ban
-	// peers performing IBD.
+
+	// A decaying ban score increase is applied to prevent exhausting
+	// resources with unusually large inventory queries.
+	//
+	// Requesting more than the maximum inventory vector length within a
+	// short period of time yields a score above the default ban threshold.
+	// Sustained bursts of small requests are not penalized as that would
+	// potentially ban peers performing IBD.
+	//
 	// This incremental score decays each minute to half of its value.
 	if sp.addBanScore(0, uint32(length)*99/wire.MaxInvPerMsg, "getdata") {
 		return
 	}
 
-	// We wait on this wait channel periodically to prevent queuing
-	// far more data than we can send in a reasonable time, wasting memory.
-	// The waiting occurs after the database fetch for the next one to
-	// provide a little pipelining.
+	// We wait on this wait channel periodically to prevent queuing far
+	// more data than we can send in a reasonable time, wasting memory. The
+	// waiting occurs after the database fetch for the next one to provide
+	// a little pipelining.
 	var waitChan chan struct{}
 	doneChan := make(chan struct{}, 1)
 
@@ -743,11 +746,11 @@ func (sp *serverPeer) OnGetData(_ *peer.Peer, msg *wire.MsgGetData) {
 		if err != nil {
 			notFound.AddInvVect(iv)
 
-			// When there is a failure fetching the final entry
-			// and the done channel was sent in due to there
-			// being no outstanding not found inventory, consume
-			// it here because there is now not found inventory
-			// that will use the channel momentarily.
+			// When there is a failure fetching the final entry and
+			// the done channel was sent in due to there being no
+			// outstanding not found inventory, consume it here
+			// because there is now not found inventory that will
+			// use the channel momentarily.
 			if i == len(msg.InvList)-1 && c != nil {
 				<-c
 			}
@@ -759,9 +762,9 @@ func (sp *serverPeer) OnGetData(_ *peer.Peer, msg *wire.MsgGetData) {
 		sp.QueueMessage(notFound, doneChan)
 	}
 
-	// Wait for messages to be sent. We can send quite a lot of data at this
-	// point and this will keep the peer busy for a decent amount of time.
-	// We don't process anything else by them in this time so that we
+	// Wait for messages to be sent. We can send quite a lot of data at
+	// this point and this will keep the peer busy for a decent amount of
+	// time. We don't process anything else by them in this time so that we
 	// have an idea of when we should hear back from them - else the idle
 	// timeout could fire when we were only half done sending the blocks.
 	if numAdded > 0 {
--- a/server.go
+++ b/server.go
@@ -687,8 +687,8 @@ func (sp *serverPeer) OnHeaders(_ *peer.Peer, msg *wire.MsgHeaders) {
 	sp.server.syncManager.QueueHeaders(msg, sp.Peer)
 }
 
-// handleGetData is invoked when a peer receives a getdata bitcoin message and
-// is used to deliver block and transaction information.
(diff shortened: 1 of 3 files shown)


wire/msgaddrv2.go
```diff
@@ -38,15 +38,14 @@ func (m *MsgAddrV2) BtcDecode(r io.Reader, pver uint32,
...
-		switch err {
-		case ErrSkippedNetworkID:
-			// This may be a network ID we don't know of, but is
-			// still valid. We can safely skip those.
-			continue
-
-		case ErrInvalidAddressSize:
-			// The encoding used by the peer does not follow
-			// BIP-155 and we should stop processing this message.
+		if err != nil {
```
ellemouton, line 41: slight change in behaviour here right? since before if the `err` didnt match one of the two cases, we'd fall through/carry on & now we dont. Def think the new behaviour is  better - just want to double check this is intended

server.go
```diff
@@ -718,33 +721,17 @@ func (sp *serverPeer) OnGetData(_ *peer.Peer, msg *wire.MsgGetData) {
...
-			err = sp.server.pushMerkleBlockMsg(sp, &iv.Hash, c, waitChan, wire.WitnessEncoding)
-		case wire.InvTypeFilteredBlock:
-			err = sp.server.pushMerkleBlockMsg(sp, &iv.Hash, c, waitChan, wire.BaseEncoding)
-		default:
-			peerLog.Warnf("Unknown type in inventory request %d",
-				iv.Type)
-			continue
-		}
+
+		err := sp.server.pushInventory(sp, iv, c, waitChan)
```
ellemouton: similar to previous commit, this is a slight behaviour change since before if we got an unknown type, we'd continue but now we'll error out.

server.go
```diff
@@ -690,8 +690,6 @@ func (sp *serverPeer) OnHeaders(_ *peer.Peer, msg *wire.MsgHeaders) {
 // handleGetData is invoked when a peer receives a getdata bitcoin message and
```
ellemouton: nit in commit message: s/prune/prone

server.go
```diff
@@ -690,8 +690,6 @@ func (sp *serverPeer) OnHeaders(_ *peer.Peer, msg *wire.MsgHeaders) {
...
 // is used to deliver block and transaction information.
 func (sp *serverPeer) OnGetData(_ *peer.Peer, msg *wire.MsgGetData) {
```
ellemouton, line 692: pre-existing but the method name is incorrect

server.go
```diff
@@ -715,38 +713,59 @@ func (sp *serverPeer) OnGetData(_ *peer.Peer, msg *wire.MsgGetData) {
 	// more data than we can send in a reasonable time, wasting memory. The
 	// waiting occurs after the database fetch for the next one to provide
 	// a little pipelining.
-	var waitChan chan struct{}
-	doneChan := make(chan struct{}, 1)
+
+	// We now create a doneChans with a size of 5, which essentially
```
ellemouton, line 717: very cool!

although, i wonder if we can just re-use the done channels?
something like this:

```
	const numBuffered = 5
	doneChans := make([]chan struct{}, numBuffered)
	for i := range doneChans {
		doneChans[i] = make(chan struct{}, 1)
	}

	for i, iv := range msg.InvList {
		// Wait for a doneChan to be available before
		// processing the next item.
		r := i % numBuffered
		doneChan := doneChans[r]

		err := sp.server.pushInventory(sp, iv, doneChan)
		if err != nil {
			failedMsg.AddInvVect(iv)
		}

		select {
		case <-doneChan:
		case <-sp.quit:
			return
		}

		// Move to the next item if we haven't processed 5 times yet.
		if (r + 1) == numBuffered {
			continue
		}

		// Empty all the slots.
		for _, dc := range doneChans {
			select {
			// NOTE: We always expect am empty struct to be sent to
			// this doneChan, even when `pushInventory` failed.
			case <-dc:

			// Exit if the server is shutting down.
			case <-sp.quit:
				peerLog.Debug("Server shutting down in " +
					"OnGetData")

				return
			}
		}
	}
```

server.go
```diff
@@ -690,82 +690,139 @@ func (sp *serverPeer) OnHeaders(_ *peer.Peer, msg *wire.MsgHeaders) {
...
+	// We wait on this wait channel periodically to prevent queuing far
+	// more data than we can send in a reasonable time, wasting memory. The
+	// waiting occurs after the database fetch for the next one to provide
+	// a little pipelining.
+
+	// We now create a doneChans with a size of 5, which essentially
+	// behaves like a semaphore that allows 5 goroutines to be running at
+	// the same time.
+	const numBuffered = 5
+	doneChans := make([]chan struct{}, 0, numBuffered)
```
GustavoStingelin, line 721: Hey @yyforyongyu, this change introduces a behavior change, not just a refactor. It also introduces a bug: we can no longer push more than 4 inventory items per message. I added a unit test to help expose this:

https://github.com/btcsuite/btcd/pull/2390/commits/af3c574445fdade5e7203dcdf5d2c7e55ac0ec98#diff-53170cac6baf0bf02f3aaa3994259ea4cffaaf2498e768fb7e03cf9f752af96c

GustavoStingelin: success from 0 to 4
error from 5 to 5000
```
➜  btcd git:(pr-2389-review) go test -run TestOnGetData_FinalBatchDrained_Table
--- FAIL: TestOnGetData_FinalBatchDrained_Table (0.00s)
    --- FAIL: TestOnGetData_FinalBatchDrained_Table/invSize_5 (1.01s)
        server_test.go:70: OnGetData did not return, possible deadlock
    --- FAIL: TestOnGetData_FinalBatchDrained_Table/invSize_6 (1.01s)
        server_test.go:70: OnGetData did not return, possible deadlock
    --- FAIL: TestOnGetData_FinalBatchDrained_Table/invSize_7 (1.01s)
        server_test.go:70: OnGetData did not return, possible deadlock
    --- FAIL: TestOnGetData_FinalBatchDrained_Table/invSize_10 (1.01s)
        server_test.go:70: OnGetData did not return, possible deadlock
    --- FAIL: TestOnGetData_FinalBatchDrained_Table/invSize_11 (1.01s)
        server_test.go:70: OnGetData did not return, possible deadlock
    --- FAIL: TestOnGetData_FinalBatchDrained_Table/invSize_15 (1.02s)
        server_test.go:70: OnGetData did not return, possible deadlock
    --- FAIL: TestOnGetData_FinalBatchDrained_Table/invSize_150 (1.15s)
        server_test.go:70: OnGetData did not return, possible deadlock
    --- FAIL: TestOnGetData_FinalBatchDrained_Table/invSize_1500 (2.50s)
        server_test.go:70: OnGetData did not return, possible deadlock
    --- FAIL: TestOnGetData_FinalBatchDrained_Table/invSize_5000 (6.00s)
        server_test.go:70: OnGetData did not return, possible deadlock
FAIL
exit status 1
FAIL    github.com/btcsuite/btcd        6.008s

```

server.go
yyforyongyu: fixed!

server.go
yyforyongyu: cool now fixed

server.go
yyforyongyu: turns out this line `doneChans = make([]chan struct{}, numBuffered)` is wrong, now fixed!

server.go
yyforyongyu: Thanks for the snippet! Yeah I like it - tho it somehow fails the test in #2390, will see what went wrong

yyforyongyu: Actually i don't think this approach works as we need to wait the done chans in the end.

GustavoStingelin (approved):
Nice! tACK!

Roasbeef (approved):
LGTM 💃🏾

PR #2396, opened 2025-07-10
p2p: remove alert msg

brunoerg:
This PR removes the `alert` message. This one was used to send general notification message throughout the network as an emergency communication mechanism. However, it's has not been used anymore and, for reference, was removed from bitcoind in [2016](https://github.com/bitcoin/bitcoin/pull/7692/files).


brunoerg:
cc @Roasbeef

Roasbeef:
Makes sense to remove it, as we don't even process it at all (see the comment in `newPeerConfig`) removed in this PR.

Roasbeef (approved):
LGTM 🌂

yyforyongyu:
When running on master, this PR breaks `lnd` and is now fixed in https://github.com/lightninglabs/neutrino/pull/329

PR #2405, opened 2025-08-15
rpcclient: add bitcoind version dependent error matching

### LINKED ISSUE #2404: string-matching will break in bitcoind v30.0
"non-mandatory-script-verify-flag" is replaced with "mempool-script-verify-flag-failed"

See: https://github.com/bitcoin/bitcoin/pull/33050/
Should change here: https://github.com/btcsuite/btcd/blob/baebb836c2d4692da3de3b0d437f4da6ce915546/rpcclient/errors.go#L334-L335

guggero:
Fixes #2404.
If different versions of bitcoind return different error strings, we need a way to match those as well.

### DIFF
--- a/rpcclient/errors.go
+++ b/rpcclient/errors.go
@@ -338,6 +338,23 @@ func (r BitcoindRPCErr) Error() string {
 	return "unknown error"
 }
 
+// BitcoindErrMap is a map of additional errors bitcoind can throw that are
+// version dependent (e.g. versions up to v29 return the error as specified in
+// `Error()` above, while versions v30 and beyond return the error as mapped
+// here. We add a new map for errors that were simply renamed but have the same
+// semantic meaning. New errors should be added above as new error constants.
+var BitcoindErrMap = map[string]error{
+	// The error message was changed in
+	// https://github.com/bitcoin/bitcoin/pull/33050 which will be included
+	// in bitcoind v30.0 and beyond.
+	"mempool script verify flag failed": ErrNonMandatoryScriptVerifyFlag,
+
+	// The error message was changed in
+	// https://github.com/bitcoin/bitcoin/pull/33183 which will also be
+	// included in bitcoind v30.0 and beyond.
+	"block script verify flag failed": ErrScriptVerifyFlag,
+}
+
 // BtcdErrMap takes the errors returned from btcd's `testmempoolaccept` and
 // `sendrawtransaction` RPCs and map them to the errors defined above, which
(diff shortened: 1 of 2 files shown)


yyforyongyu:
@ziggie1984 maybe you could take a look?

rpcclient/errors.go
```diff
@@ -338,6 +338,17 @@ func (r BitcoindRPCErr) Error() string {
...
 
+// BitcoindErrMap is a map of additional errors bitcoind can throw that are
+// version dependent (e.g. versions up to v29 return the error as specified in
+// `Error()` above, while versions v30 and beyond return the error as mapped
+// here.
+var BitcoindErrMap = map[string]error{
+	// The error message was changed in
+	// https://github.com/bitcoin/bitcoin/pull/33050 which will be included
+	// in bitcoind v30.0 and beyond.
+	"mempool script verify flag failed": ErrNonMandatoryScriptVerifyFlag,
```
ziggie1984, line 350: can you also add: `block-script-verify-flag-failed` see https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes-33183.md ?

rpcclient/errors.go
```diff
@@ -338,6 +338,17 @@ func (r BitcoindRPCErr) Error() string {
 	return "unknown error"
 }
 
+// BitcoindErrMap is a map of additional errors bitcoind can throw that are
+// version dependent (e.g. versions up to v29 return the error as specified in
+// `Error()` above, while versions v30 and beyond return the error as mapped
+// here.
+var BitcoindErrMap = map[string]error{
```
ziggie1984, line 346: Why don't we just introduce 2 new errors instead of mapping it to the old one ?

ziggie1984 (commented):
Thank you for monitoring the core lib, wondering if we just define these 2 new errors instead of mapping ?

rpcclient/errors.go
guggero: Ah, yes. Added.

rpcclient/errors.go
guggero: Because the errors were simply renamed, I wanted to keep the same semantic meaning (e.g. the same error constant) in case we matched on them anywhere. We currently don't, so we could've added new errors. But I still find it slightly better in terms of understanding what's going on. Added some additional context to the comment of the new map, why we use it and when we should create new constants.
I hope that makes sense to you?

guggero:
@Roasbeef requesting merge override.

PR #2423, opened 2025-09-16
btcec: add IsEquivalentPubKeys helper method

ffranr:
Add IsEquivalentPubKeys, a helper method that reports whether two public keys are equivalent in their BIP340-serialized form. This ensures a canonical comparison by avoiding discrepancies caused by multiple encodings of the same elliptic curve point, since BIP340 serialization provides a unique byte representation.

### DIFF
--- a/btcec/pubkey.go
+++ b/btcec/pubkey.go
@@ -5,6 +5,9 @@
 package btcec
 
 import (
+	"bytes"
+
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	secp "github.com/decred/dcrd/dcrec/secp256k1/v4"
 )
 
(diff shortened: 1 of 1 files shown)


btcec/pubkey.go
```diff
@@ -52,6 +55,17 @@ func NewPublicKey(x, y *FieldVal) *PublicKey {
 	return secp.NewPublicKey(x, y)
 }
 
+// IsEquivalentPubKeys reports whether two public keys are equivalent when
+// compared in their BIP340-serialized form. This avoids issues with multiple
+// encodings of the same elliptic curve point, since BIP340 serialization
+// provides a unique, canonical byte representation.
+func IsEquivalentPubKeys(a, b *PublicKey) bool {
```
allocz, line 62: Thanks for the human time spent writing the patch. But I don't agree with adding helpers without concrete uses of them, when needing something like this, we can write it. At least for now, my vote is NACK.

PR #2425, opened 2025-09-20
treap, ffldb: add treapNodePool

kcalvinalvin:
## Change Description
During the initial-block download, one of the biggest culprits of excessive memory allocation was the [immutable treap](https://github.com/btcsuite/btcd/blob/e8097a1b044c3c0b3843b2aba377e87eb5fc4f9f/database/internal/treap/immutable.go#L23-L46).
The immutable treap guarantees immutability by cloning all shared nodes that'll be modified inside the treap during deletion and put operations.

For the insertions of 3 keys:

key 1: 50
key 2: 10
key 3: 4
     
The insertion is like so:
```   
    1: allocate 50.
     
            50
     
    2: clone 50, allocate 10.
     
            50
           /  \
          10
     
    3: clone 50, clone 10, allocate 4
     
            50
           /  \
          10
         /
        4
```

In the above example, only the nodes allocated during insertion of (3) is going to be used. 
The rest is going to be garbage collected immediately which results in memory being allocated too much.

The memory allocation savings in this PR comes from putting the nodes that'll be garbage collected in the sync.Pool.

We're able to safely recycle the intermediary nodes as we now require the `Put()` operation to take in multiple key-value pairs.
Since there's a guarantee that the previous copies are not being accessed as they're *only* accessible within the `Put()` operation, it's safe to recycle them.

I tested both the `master` and `b3bf17cd0f924c82634271babb2e312f44b7a314` with the following command.
I used the `connect` flag to reduce the differences that would happen a random selection of peers.
I added a checkpoint just to speed up things on signet since there aren't checkpoints by default for signet.

```
btcd --datadir=. --logdir=. --signet --addcheckpoint=260000:0000000b2d5533cb361cab243eb2034986c579efceb2fe266baa057104537bf5 --memprofile=memprof-with-master --connect=100.115.246.54
```

On master, the pprof flamegraph for syncing on signet looks like so:

We can see that the `cloneTreapNode` function is taking up a lot of space in the flamegraph.
The total memory allocation is also 511GB.
IMG_0418

With this PR, the flamegraph looks like so:

We can see a dramatic reduction in the total memory allocated as it's now just 359GB.
The problematic `cloneTreapNode` is also significantly reduced in the flamegraph.
IMG_0419

I added the profiles that I took in this zip file:
[2025-09-20-treappool-pprof.zip](https://github.com/user-attachments/files/22441189/2025-09-20-treappool-pprof.zip)

## Steps to Test

To test that there is a memory reduction, run the following command on `master` and on this PR.
Compare that there's less memory allocation with this PR.

```
# Make sure to edit the output name for the flag `memprofile`!
btcd --datadir=. --logdir=. --signet --addcheckpoint=260000:0000000b2d5533cb361cab243eb2034986c579efceb2fe266baa057104537bf5 --memprofile=memprof-with-master
```

To test that the `Put()` function still guarantees immutability, check that the tests in `immutable_test.go` are correctly
modified to account for the function prototype change.

### Testing

### DIFF
--- a/database/ffldb/dbcache.go
+++ b/database/ffldb/dbcache.go
@@ -611,19 +611,27 @@ func (c *dbCache) commitTx(tx *transaction) error {
 	c.cacheLock.RUnlock()
 
 	// Apply every key to add in the database transaction to the cache.
+	pendingKeys := make([][]byte, 0, tx.pendingKeys.Len())
+	pendingKeyValues := make([][]byte, 0, tx.pendingKeys.Len())
 	tx.pendingKeys.ForEach(func(k, v []byte) bool {
+		pendingKeys = append(pendingKeys, k)
+		pendingKeyValues = append(pendingKeyValues, v)
+
 		newCachedRemove = newCachedRemove.Delete(k)
-		newCachedKeys = newCachedKeys.Put(k, v)
 		return true
 	})
+	newCachedKeys = newCachedKeys.Put(pendingKeys, pendingKeyValues)
 	tx.pendingKeys = nil
 
 	// Apply every key to remove in the database transaction to the cache.
+	pendingRemoveKeys := make([][]byte, 0, tx.pendingRemove.Len())
 	tx.pendingRemove.ForEach(func(k, v []byte) bool {
+		pendingRemoveKeys = append(pendingRemoveKeys, k)
+
 		newCachedKeys = newCachedKeys.Delete(k)
-		newCachedRemove = newCachedRemove.Put(k, nil)
 		return true
 	})
+	newCachedRemove = newCachedRemove.Put(pendingRemoveKeys, nil)
 	tx.pendingRemove = nil
 
 	// Atomically replace the immutable treaps which hold the cached keys to
--- a/database/internal/treap/common.go
+++ b/database/internal/treap/common.go
@@ -6,6 +6,7 @@ package treap
 
 import (
 	"math/rand"
+	"sync"
 	"time"
 )
 
@@ -33,6 +34,14 @@ var (
 	emptySlice = make([]byte, 0)
 )
 
+// treapNodePool defines a concurrent safe free list of treapNode used to
+// provide temporary buffers.
+var treapNodePool = sync.Pool{
+	New: func() any {
+		return &treapNode{}
+	},
+}
+
 // treapNode represents a node in the treap.
 type treapNode struct {
 	key      []byte
@@ -51,7 +60,14 @@ func nodeSize(node *treapNode) uint64 {
 // newTreapNode returns a new node from the given key, value, and priority.  The
 // node is not initially linked to any others.
 func newTreapNode(key, value []byte, priority int) *treapNode {
-	return &treapNode{key: key, value: value, priority: priority}
+	n := treapNodePool.Get().(*treapNode)
+	n.key = key
+	n.value = value
+	n.priority = priority
+	n.left = nil
+	n.right = nil
+
+	return n
 }
 
 // parentStack represents a stack of parent treap nodes that are used during
--- a/database/internal/treap/immutable.go
+++ b/database/internal/treap/immutable.go
@@ -106,6 +106,11 @@ func (t *Immutable) Get(key []byte) []byte {
 
 // Put inserts the passed key/value pair.
 func (t *Immutable) Put(key, value []byte) *Immutable {
+	return t.put(key, value)
+}
+
+// put inserts the passed key/value pair.
+func (t *Immutable) put(key, value []byte) *Immutable {
 	// Use an empty byte slice for the value when none was provided.  This
 	// ultimately allows key existence to be determined from the value since
 	// an empty byte slice is distinguishable from nil.
--- a/database/internal/treap/immutable.go
+++ b/database/internal/treap/immutable.go
@@ -11,13 +11,14 @@ import (
 
 // cloneTreapNode returns a shallow copy of the passed node.
 func cloneTreapNode(node *treapNode) *treapNode {
-	return &treapNode{
-		key:      node.key,
-		value:    node.value,
-		priority: node.priority,
-		left:     node.left,
-		right:    node.right,
-	}
+	n := treapNodePool.Get().(*treapNode)
+	n.key = node.key
+	n.value = node.value
+	n.priority = node.priority
+	n.left = node.left
+	n.right = node.right
+
+	return n
 }
 
 // Immutable represents a treap data structure which is used to hold ordered
--- a/database/internal/treap/immutable.go
+++ b/database/internal/treap/immutable.go
@@ -110,8 +110,11 @@ func (t *Immutable) Put(key, value []byte) *Immutable {
 	return t.put(key, value)
 }
 
-// put inserts the passed key/value pair.
-func (t *Immutable) put(key, value []byte) *Immutable {
+// put inserts the passed key/value pair and returns all the newly created
+// treapNodes that were created during this put operation.  The returned
+// treapNodes can then be put into data structures like sync.Pool to reduce the
+// memory overhead of allocating new treapNodes during multiple put calls.
+func (t *Immutable) put(key, value []byte) (*Immutable, [staticDepth]*treapNode) {
 	// Use an empty byte slice for the value when none was provided.  This
 	// ultimately allows key existence to be determined from the value since
 	// an empty byte slice is distinguishable from nil.
@@ -119,10 +122,17 @@ func (t *Immutable) put(key, value []byte) *Immutable {
 		value = emptySlice
 	}
 
+	// recycle is the treapNodes that are created during this put operation.
+	// We keep track of the nodes as the caller may be choose to recycle
+	// them to keep memory allocation low.
+	var recycle [staticDepth]*treapNode
+	var rIdx int
+
 	// The node is the root of the tree if there isn't already one.
 	if t.root == nil {
 		root := newTreapNode(key, value, rand.Int())
-		return newImmutable(root, 1, nodeSize(root))
+		recycle[rIdx] = root
+		return newImmutable(root, 1, nodeSize(root)), recycle
 	}
 
 	// Find the binary tree insertion point and construct a replaced list of
@@ -138,6 +148,16 @@ func (t *Immutable) put(key, value []byte) *Immutable {
 	for node := t.root; node != nil; {
 		// Clone the node and link its parent to it if needed.
 		nodeCopy := cloneTreapNode(node)
+
+		// Check if we still have space in the recycle for this node.
+		// It's ok if we don't put every single new node to be recycled
+		// as there's no guarantee in the sync.Pool that every recycled
+		// treapNode will be re-utilized.
+		if rIdx < staticDepth {
+			recycle[rIdx] = nodeCopy
+			rIdx++
+		}
+
 		if oldParent := parents.At(0); oldParent != nil {
 			if oldParent.left == node {
 				oldParent.left = nodeCopy
@@ -167,11 +187,20 @@ func (t *Immutable) put(key, value []byte) *Immutable {
 		newRoot := parents.At(parents.Len() - 1)
 		newTotalSize := t.totalSize - uint64(len(node.value)) +
 			uint64(len(value))
-		return newImmutable(newRoot, t.count, newTotalSize)
+		return newImmutable(newRoot, t.count, newTotalSize), recycle
+	}
+
+	// Check if we still have space in the recycle for this node.
+	// It's ok if we don't put every single new node to be recycled
+	// as there's no guarantee in the sync.Pool that every recycled
+	// treapNode will be re-utilized.
+	node := newTreapNode(key, value, rand.Int())
+	if rIdx < staticDepth {
+		recycle[rIdx] = node
+		rIdx++
 	}
 
 	// Link the new node into the binary tree in the correct position.
-	node := newTreapNode(key, value, rand.Int())
 	parent := parents.At(0)
 	if compareResult < 0 {
 		parent.left = node
@@ -211,7 +240,7 @@ func (t *Immutable) put(key, value []byte) *Immutable {
 		}
 	}
 
-	return newImmutable(newRoot, t.count+1, t.totalSize+nodeSize(node))
+	return newImmutable(newRoot, t.count+1, t.totalSize+nodeSize(node)), recycle
 }
 
 // Delete removes the passed key from the treap and returns the resulting treap
--- a/database/internal/treap/immutable.go
+++ b/database/internal/treap/immutable.go
@@ -105,9 +105,62 @@ func (t *Immutable) Get(key []byte) []byte {
 	return nil
 }
 
-// Put inserts the passed key/value pair.
-func (t *Immutable) Put(key, value []byte) *Immutable {
-	return t.put(key, value)
+// Put puts the passed in key/value pairs into the treap.  For operations
+// requiring many insertions at once, Put is memory efficient as the
+// intermediary treap nodes created between each put operation is recycled
+// through an internal sync.Pool, reducing overall memory allocation.
+//
+// If the passed in length of values is less than keys, the keys with no
+// matching values will have nil values.  If the passed in values are nil, all
+// keys will be saved will nil values.
+func (t *Immutable) Put(keys, values [][]byte) *Immutable {
+	treap := t
+	var prevTreapNodes [staticDepth]*treapNode
+
+	for i, key := range keys {
+		// Use an empty byte slice for the value when none was provided.
+		// This ultimately allows key existence to be determined from
+		// the value since an empty byte slice is distinguishable from nil.
+		value := emptySlice
+		if values != nil && i < len(values) {
+			value = values[i]
+		}
+
+		newTreap, newTreapNodes := treap.put(key, value)
+
+		// Loop through the prevTreapNodes and check for treapNodes that
+		// are no longer being utilized.  These will be garbaged collected
+		// and they're better off being recycled in the treapNodePool.
+		for _, node := range prevTreapNodes {
+			if node == nil {
+				break
+			}
+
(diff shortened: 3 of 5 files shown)


saubyk:

saubyk:
cc: @bhandras for review

database/internal/treap/immutable.go
```diff
@@ -110,19 +110,29 @@ func (t *Immutable) Put(key, value []byte) *Immutable {
...
 	// ultimately allows key existence to be determined from the value since
 	// an empty byte slice is distinguishable from nil.
 	if value == nil {
 		value = emptySlice
 	}
 
+	// recycle is the treapNodes that are created during this put operation.
+	// We keep track of the nodes as the caller may be choose to recycle
+	// them to keep memory allocation low.
+	var recycle [staticDepth]*treapNode
```
Roasbeef: Style nit: can use a single `var` statement here.

database/internal/treap/immutable.go
```diff
@@ -110,19 +110,29 @@ func (t *Immutable) Put(key, value []byte) *Immutable {
...
+	var rIdx int
```
Roasbeef: `currentRecycleIndex`?

database/internal/treap/immutable.go
```diff
@@ -110,19 +110,29 @@ func (t *Immutable) Put(key, value []byte) *Immutable {
 	return t.put(key, value)
 }
 
-// put inserts the passed key/value pair.
-func (t *Immutable) put(key, value []byte) *Immutable {
+// put inserts the passed key/value pair and returns all the newly created
+// treapNodes that were created during this put operation.  The returned
+// treapNodes can then be put into data structures like sync.Pool to reduce the
+// memory overhead of allocating new treapNodes during multiple put calls.
+func (t *Immutable) put(key, value []byte) (*Immutable, [staticDepth]*treapNode) {
```
Roasbeef, line 160: IIUC compilation is broken in this commit, we an extra arg is returned, but no callers are updated.

database/internal/treap/common.go
```diff
@@ -33,6 +34,14 @@ var (
 	emptySlice = make([]byte, 0)
 )
 
+// treapNodePool defines a concurrent safe free list of treapNode used to
+// provide temporary buffers.
+var treapNodePool = sync.Pool{
+	New: func() any {
+		return &treapNode{}
+	},
+}
```
Roasbeef, line 43: What if we added a `recycle` method to the `treapNode`? It would be what calls `put` to return things back into the buffer pool.

database/internal/treap/immutable.go
Roasbeef: We can also wrap this array into a helps struct, which can help avoid pointer derefs. As is, the caller would need to know what the current `rIdx` actually is to avoid panics.

database/internal/treap/immutable.go
```diff
@@ -105,9 +105,62 @@ func (t *Immutable) Get(key []byte) []byte {
...
-	return t.put(key, value)
+// Put puts the passed in key/value pairs into the treap.  For operations
+// requiring many insertions at once, Put is memory efficient as the
+// intermediary treap nodes created between each put operation is recycled
+// through an internal sync.Pool, reducing overall memory allocation.
+//
+// If the passed in length of values is less than keys, the keys with no
+// matching values will have nil values.  If the passed in values are nil, all
+// keys will be saved will nil values.
+func (t *Immutable) Put(keys, values [][]byte) *Immutable {
```
Roasbeef: We should assert that `len(keys) == len(values)`.

database/internal/treap/immutable.go
```diff
@@ -105,9 +105,62 @@ func (t *Immutable) Get(key []byte) []byte {
...
+			// if the pointer value of the node is the same.
+			got := newTreap.get(node.key)
+			if got == node {
+				continue
+			}
+
+			// This node is only being used by the previous immutable
+			// copy and can safely be put into the treapNodePool to be
+			// recycled.
+			node.key = nil
```
Roasbeef: We can wrap this into a `recycle` method for the treap node, which nils out the values, then returns them to the pool.

database/internal/treap/immutable.go
```diff
@@ -105,9 +105,62 @@ func (t *Immutable) Get(key []byte) []byte {
...
+		// and they're better off being recycled in the treapNodePool.
+		for _, node := range prevTreapNodes {
+			if node == nil {
+				break
+			}
+
+			// Make sure that the node we're going to recycle isn't
+			// being used by the latest immutable treap by checking
+			// if the pointer value of the node is the same.
+			got := newTreap.get(node.key)
```
Roasbeef, line 137: 👍

database/internal/treap/immutable.go
Roasbeef: Perhaps we should leave the old `Put` in place, and call this `BatchPut`? We can then make a wrapper struct around the k/v pair. Then we don't need to try to line up each key to the value, as they're coupled together, and the value is `nil` if there is none.

database/ffldb/dbcache.go
```diff
@@ -611,19 +611,27 @@ func (c *dbCache) commitTx(tx *transaction) error {
...
-		newCachedKeys = newCachedKeys.Put(k, v)
 		return true
 	})
+	newCachedKeys = newCachedKeys.Put(pendingKeys, pendingKeyValues)
 	tx.pendingKeys = nil
 
 	// Apply every key to remove in the database transaction to the cache.
+	pendingRemoveKeys := make([][]byte, 0, tx.pendingRemove.Len())
 	tx.pendingRemove.ForEach(func(k, v []byte) bool {
+		pendingRemoveKeys = append(pendingRemoveKeys, k)
```
Roasbeef: Re the comment below, we can also have a bit/bool to indicate delete vs insert.

database/internal/treap/immutable.go
```diff
@@ -104,19 +105,87 @@ func (t *Immutable) Get(key []byte) []byte {
...
-func (t *Immutable) Put(key, value []byte) *Immutable {
+// Put puts the passed in key/value pairs into the treap.  For operations
+// requiring many insertions at once, Put is memory efficient as the
+// intermediary treap nodes created between each put operation is recycled
+// through an internal sync.Pool, reducing overall memory allocation.
+//
+// If the passed in length of values is less than keys, the keys with no
+// matching values will have nil values.  If the passed in values are nil, all
+// keys will be saved will nil values.
+func (t *Immutable) Put(keys, values [][]byte) *Immutable {
```
Roasbeef: What about also applying this to `Delete`?

database/ffldb/dbcache.go
```diff
@@ -611,19 +611,27 @@ func (c *dbCache) commitTx(tx *transaction) error {
...
+
```
Roasbeef, line 628: Can also do batch delete.

Roasbeef:
Here's an attempt to make the test diff a bit easier to follow: 
```diff
diff --git a/database/internal/treap/common_test.go b/database/internal/treap/common_test.go
index c43e678d..526c60cc 100644
--- a/database/internal/treap/common_test.go
+++ b/database/internal/treap/common_test.go
@@ -5,6 +5,7 @@
 package treap
 
 import (
+	"crypto/sha256"
 	"encoding/binary"
 	"encoding/hex"
 	"math/rand"
@@ -31,6 +32,65 @@ func serializeUint32(ui uint32) []byte {
 	return ret[:]
 }
 
+// makeSeqKeysWithStep returns serialized uint32 keys in ascending order
+// for the half-open range [start, end), advancing by step.  A step of 0
+// is treated as 1.
+func makeSeqKeysWithStep(start, end, step int) [][]byte {
+	if step == 0 {
+		step = 1
+	}
+	if end <= start {
+		return nil
+	}
+	// Preallocate best-effort capacity.
+	n := (end - start + (step - 1)) / step
+	keys := make([][]byte, 0, n)
+	for i := start; i < end; i += step {
+		keys = append(keys, serializeUint32(uint32(i)))
+	}
+	return keys
+}
+
+// makeSeqKeys returns serialized uint32 keys in ascending order for the
+// half-open range [start, end).
+func makeSeqKeys(start, end int) [][]byte {
+	return makeSeqKeysWithStep(start, end, 1)
+}
+
+// makeRevKeys returns serialized uint32 keys in descending order for the
+// inclusive range [start, end].  It is a no-op when start < end.
+func makeRevKeys(start, end int) [][]byte {
+	if start < end {
+		return nil
+	}
+	n := start - end + 1
+	keys := make([][]byte, 0, n)
+	for v := start; v >= end; v-- {
+		keys = append(keys, serializeUint32(uint32(v)))
+	}
+	return keys
+}
+
+// makeHashedSeqKeys returns sha256(serialized uint32) keys in ascending order
+// for the half-open range [start, end).
+func makeHashedSeqKeys(start, end int) [][]byte {
+	if end <= start {
+		return nil
+	}
+	keys := make([][]byte, 0, end-start)
+	for i := start; i < end; i++ {
+		sum := sha256.Sum256(serializeUint32(uint32(i)))
+		keys = append(keys, sum[:])
+	}
+	return keys
+}
+
+// putKeysAsValues is a small helper to put the provided keys with values equal
+// to the keys, returning the updated immutable treap.
+func putKeysAsValues(t *Immutable, keys [][]byte) *Immutable {
+	return t.Put(keys, keys)
+}
+
 // TestParentStack ensures the treapParentStack functionality works as intended.
 func TestParentStack(t *testing.T) {
 	t.Parallel()
diff --git a/database/internal/treap/immutable_test.go b/database/internal/treap/immutable_test.go
index f24d406f..4ef2dad9 100644
--- a/database/internal/treap/immutable_test.go
+++ b/database/internal/treap/immutable_test.go
@@ -64,14 +64,8 @@ func TestImmutableSequential(t *testing.T) {
 	keyCount := 100
 	testTreap := NewImmutable()
 	for i := 0; i < numItems/keyCount; i++ {
-		keys := make([][]byte, 0, keyCount)
-		for j := 0; j < keyCount; j++ {
-			n := i*keyCount + j
-			key := serializeUint32(uint32(n))
-			keys = append(keys, key)
-		}
-
-		testTreap = testTreap.Put(keys, keys)
+		keys := makeSeqKeys(i*keyCount, (i+1)*keyCount)
+		testTreap = putKeysAsValues(testTreap, keys)
 
 		// Ensure the treap length is the expected value.
 		if gotLen := testTreap.Len(); gotLen != (i+1)*keyCount {
@@ -174,14 +168,10 @@ func TestImmutableReverseSequential(t *testing.T) {
 	keyCount := 100
 	testTreap := NewImmutable()
 	for i := 0; i < numItems/keyCount; i++ {
-		keys := make([][]byte, 0, keyCount)
-		for j := 0; j < keyCount; j++ {
-			n := numItems - (i * keyCount) - j - 1
-			key := serializeUint32(uint32(n))
-			keys = append(keys, key)
-		}
-
-		testTreap = testTreap.Put(keys, keys)
+		start := numItems - (i * keyCount) - 1
+		end := numItems - (i+1)*keyCount
+		keys := makeRevKeys(start, end)
+		testTreap = putKeysAsValues(testTreap, keys)
 
 		// Ensure the treap length is the expected value.
 		if gotLen := testTreap.Len(); gotLen != (i+1)*keyCount {
@@ -286,15 +276,8 @@ func TestImmutableUnordered(t *testing.T) {
 	testTreap := NewImmutable()
 	for i := 0; i < numItems/keyCount; i++ {
 		// Hash the serialized int to generate out-of-order keys.
-		keys := make([][]byte, 0, keyCount)
-		for j := 0; j < keyCount; j++ {
-			n := i*keyCount + j
-			hash := sha256.Sum256(serializeUint32(uint32(n)))
-			key := hash[:]
-			keys = append(keys, key)
-		}
-
-		testTreap = testTreap.Put(keys, keys)
+		keys := makeHashedSeqKeys(i*keyCount, (i+1)*keyCount)
+		testTreap = putKeysAsValues(testTreap, keys)
 
 		// Ensure the treap length is the expected value.
 		if gotLen := testTreap.Len(); gotLen != (i+1)*keyCount {
@@ -448,12 +431,8 @@ func TestImmutableForEachStopIterator(t *testing.T) {
 	// Insert a few keys.
 	numItems := 10
 	testTreap := NewImmutable()
-	keys := make([][]byte, 0, numItems)
-	for i := 0; i < numItems; i++ {
-		key := serializeUint32(uint32(i))
-		keys = append(keys, key)
-	}
-	testTreap = testTreap.Put(keys, keys)
+	keys := makeSeqKeys(0, numItems)
+	testTreap = putKeysAsValues(testTreap, keys)
 
 	// Ensure ForEach exits early on false return by caller.
 	var numIterated int
@@ -482,14 +461,8 @@ func TestImmutableSnapshot(t *testing.T) {
 	for i := 0; i < numItems/keyCount; i++ {
 		treapSnap := testTreap
 
-		keys := make([][]byte, 0, keyCount)
-		for j := 0; j < keyCount; j++ {
-			n := i*keyCount + j
-			key := serializeUint32(uint32(n))
-			keys = append(keys, key)
-		}
-
-		testTreap = testTreap.Put(keys, keys)
+		keys := makeSeqKeys(i*keyCount, (i+1)*keyCount)
+		testTreap = putKeysAsValues(testTreap, keys)
 
 		// Ensure the length of the treap snapshot is the expected
 		// value.
diff --git a/database/internal/treap/treapiter_test.go b/database/internal/treap/treapiter_test.go
index 73a53d27..2063de18 100644
--- a/database/internal/treap/treapiter_test.go
+++ b/database/internal/treap/treapiter_test.go
@@ -496,12 +496,8 @@ testLoop:
 	for i, test := range tests {
 		// Insert a bunch of keys.
 		testTreap := NewImmutable()
-		keys := make([][]byte, 0, test.numKeys)
-		for i := 0; i < test.numKeys; i += test.step {
-			key := serializeUint32(uint32(i))
-			keys = append(keys, key)
-		}
-		testTreap = testTreap.Put(keys, keys)
+		keys := makeSeqKeysWithStep(0, test.numKeys, test.step)
+		testTreap = putKeysAsValues(testTreap, keys)
 
 		// Create new iterator limited by the test params.
 		iter := testTreap.Iterator(test.startKey, test.limitKey)
```

database/internal/treap/immutable.go
kcalvinalvin: I guess if we do wrap this into a different struct, we can just use https://github.com/btcsuite/btcd/blob/442ef28bcf03797e845c8e957e5cd6d4bffb5764/database/internal/treap/common.go#L64-L68

But the caller would never actually need to know the current `rIdx` is since the caller wouldn't be allocating new treapNodes.

database/internal/treap/immutable.go
kcalvinalvin: We can. In terms of performance it'd be negligible so I didn't bother but I will

kmk142789 (approved):

database/internal/treap/immutable.go
kcalvinalvin: We can but having this in here is nice because all the tests are based around `Put()`. If we have a separate `BatchPut()` we need to write separate tests and check that the tests are equals to `Put()`.

database/ffldb/dbcache.go
kcalvinalvin: I'll do this in a follow up PR since it's an extra feature.

database/internal/treap/immutable.go
kcalvinalvin: I'll do so in a follow up PR.

database/ffldb/dbcache.go
kcalvinalvin: I'll do so in a follow up PR.

database/internal/treap/immutable.go
kcalvinalvin: Changed the function prototype in the latest push to:

```
type KVPair struct {
       Key   []byte
       Value []byte
}
func (t *Immutable) Put(kvs ...KVPair) *Immutable {}
```

kcalvinalvin: No longer needed in the latest push

database/internal/treap/immutable.go
kcalvinalvin: Changed to a single var statement and the name of rIdx to currentRecycleIndex

database/internal/treap/immutable.go
kcalvinalvin: Fixed in the latest push

database/internal/treap/immutable.go
kcalvinalvin: Applied

database/internal/treap/common.go
kcalvinalvin: Applied in the latest push

database/internal/treap/immutable.go
kcalvinalvin: Done in the latest push

kcalvinalvin:
Addressed the fixes requested here besides doing the batch deletions.

guibressan:
## Tracing/profiling comparison between master and pr/2425

Config: 

container memory limit: 4096MB; container cpu's: 4; GOMEMLIMIT=1024M; GOGC=50; 
btcd_utxocachemaxsize=250M

master(1c55c7c18179f9638f5a481151692516c9db45d6)
pr/2425(e00912474828233f524d24678335220c702b7506)

### During normal IBD:

Allocations:

Allocation flamegraph is mosly the same, ImmutableTreap operations are not
doing significant allocations

Garbage Collection:

On master: ~2 seconds between each cycle, <200ms per GC cycle
On pr/2425: ~3 seconds between each cycle, <200ms per GC cycle

### During IBD on cache flush:

Allocations:

On master: treap node cloning doing ~70% of total memory allocation
On pr/2425: treap node cloning doing 4% of total memory allocations

We can see considerable improvements here.

Garbage Collection:

On master: GC running for 41 of 60 seconds, 68% of the time
On pr/2425: GC running for 6 of 60 seconds, ~10% of the time

### Final considerations

Performance improvement of PR/2425 is considerable, and I think it should be
merged. 

Good job @kcalvinalvin!

PR #2428, opened 2025-09-22
blockchain, netsync: implement a complete headers-first download during ibd

kcalvinalvin:
## Change Description
Right now the headers-first download is only based off of the checkpoints and is thus limited to the last checkpoint.
The newly implemented headers-first download will always download headers-first and will validate them to see if the headers connect and have proper proof of work.

Then the block download will be based off of the verified headers. This now eliminates any potential downloading of txs or orphan blocks during ibd. It also makes future parallel block download much better as a parallel block download can only happen for blocks we already have headers for.

~~It's not yet put into the code yet but this allows the node to also receive block headers instead of invs during block propagation.~~ I'll do this in a follow up later.

## Steps to Test

- Perform ibd on any of the networks and see that the node syncs up fine.
- Check that the newly added function tests are correct.

### Testing

### DIFF
--- a/netsync/manager_test.go
+++ b/netsync/manager_test.go
@@ -1169,3 +1169,47 @@ func TestStallNoDisconnectAtSameHeight(t *testing.T) {
 	require.Nil(t, sm.syncPeer,
 		"we should have nil syncPeer after handleStallSample")
 }
+
+// TestStartSyncChainCurrent verifies that startSync does not set syncPeer
+// or ibdMode when the chain is current and no peer is strictly higher.
+// isInIBDMode sees IsCurrent()==true with no higher peers, returns false,
+// and startSync exits immediately.
+func TestStartSyncChainCurrent(t *testing.T) {
+	t.Parallel()
+
+	params := chaincfg.RegressionNetParams
+	params.Checkpoints = nil
+
+	sm, tearDown := makeMockSyncManager(t, &params)
+	defer tearDown()
+
+	// Mine a single block with a recent timestamp so
+	// IsCurrent() returns true.
+	cb := createTestCoinbase(1, &params)
+	header := wire.BlockHeader{
+		Version:    1,
+		PrevBlock:  *params.GenesisHash,
+		MerkleRoot: cb.TxHash(),
+		Timestamp:  time.Now().Truncate(time.Second),
+		Bits:       params.PowLimitBits,
+	}
+	require.True(t, solveTestBlock(&header, &params))
+
+	block := btcutil.NewBlock(&wire.MsgBlock{
+		Header:       header,
+		Transactions: []*wire.MsgTx{cb},
+	})
+	_, _, err := sm.chain.ProcessBlock(block, blockchain.BFNone)
+	require.NoError(t, err)
+	require.True(t, sm.chain.IsCurrent())
+
+	// Peer at our height — not higher.
+	newSyncCandidate(t, sm, 1)
+
+	sm.startSync()
+
+	require.Nil(t, sm.syncPeer,
+		"syncPeer should not be set when chain is already current")
+	require.False(t, sm.ibdMode,
+		"ibdMode should not be activated when chain is already current")
+}


saubyk:
cc: @gijswijs for review

saubyk:

blockchain/blockindex.go
```diff
@@ -46,6 +49,11 @@ func (status blockStatus) HaveData() bool {
 	return status&statusDataStored != 0
 }
 
+// HaveHeader returns whether the header data is stored in the database.
+func (status blockStatus) HaveHeader() bool {
```
Roasbeef, line 53: Isn't this just the inverse of `HaveData`?

blockchain/process_test.go
```diff
@@ -0,0 +1,231 @@
...
+		rand.Read(randBytes[:])
+		merkle := chainhash.HashH(randBytes[:])
+
+		header := wire.BlockHeader{
+			Version:    1,
+			PrevBlock:  tip.BlockHash(),
+			MerkleRoot: merkle,
+			Bits:       chainParams.PowLimitBits,
+			Timestamp:  ts,
+			Nonce:      0, // To be solved.
```
Roasbeef: Superfluous comment.

blockchain/process_test.go
```diff
@@ -0,0 +1,231 @@
...
+			t.Fatalf("expected header %v to be in the main chain",
+				header.BlockHash())
+		}
+	}
+
+	// Check that the tip is correct.
+	lastHeader := headers[len(headers)-1]
+	lastHeaderHash := lastHeader.BlockHash()
+	tipNode := chain.bestHeader.Tip()
+	if !tipNode.hash.IsEqual(&lastHeaderHash) {
```
Roasbeef: You can use the `require` package for all these assertions.

blockchain/process_test.go
```diff
@@ -0,0 +1,231 @@
...
+		t.Fatalf("expected statusHeaderStored but got %v",
+			tipNode.status)
+	}
+	if tipNode.height != int32(len(headers)) {
+		t.Fatalf("expected height of %v but got %v",
+			len(headers), tipNode.height)
+	}
+
+	// Create invalid header at the checkpoint.
+	//
```
Roasbeef: Style nit: blank line, same below.

blockchain/process_test.go
```diff
@@ -0,0 +1,231 @@
...
+
+	// Create more sidechain block headers so that it becomes the active chain.
+	//
+	// 	genesis -> 1  -> 2  -> 3  -> 4  -> 5  -> ... -> 10 (valid-fork)
+	//                                            \-> 6  -> ... -> 18 (active)
+	lastSideChainHeaderIdx := len(sidechainExtendingHeaders) - 1
+	lastSidechainHeader := sidechainExtendingHeaders[lastSideChainHeaderIdx]
+	lastSidechainHeaderHash := lastSidechainHeader.BlockHash()
+
+	// Check that the tip is now different.
```
Roasbeef, line 163: 👍

blockchain/chain.go
```diff
@@ -1535,6 +1544,52 @@ func (b *BlockChain) BlockHashByHeight(blockHeight int32) (*chainhash.Hash, erro
...
 
+// IsValidHeader checks that we've already checked that this header connects to the
+// chain of best headers and did not receive an invalid state.
+func (b *BlockChain) IsValidHeader(blockHash *chainhash.Hash) bool {
+	node := b.index.LookupNode(blockHash)
+	if node == nil || !b.bestHeader.Contains(node) {
+		return false
+	}
+
+	if node.status.KnownInvalid() {
```
Roasbeef:
```go
return !node.status.KnownInvalid()
```

?

blockchain/chain.go
```diff
@@ -1535,6 +1544,52 @@ func (b *BlockChain) BlockHashByHeight(blockHeight int32) (*chainhash.Hash, erro
...
+// This function is safe for concurrent access.
+func (b *BlockChain) LatestBlockLocatorByHeader() (BlockLocator, error) {
+	b.chainLock.RLock()
+	locator := b.bestHeader.BlockLocator(nil)
+	b.chainLock.RUnlock()
+	return locator, nil
+}
+
+// HeaderHashByHeight returns the block header's hash given its height.
+func (b *BlockChain) HeaderHashByHeight(blockHeight int32) (*chainhash.Hash, error) {
```
Roasbeef: We should add to the `godoc` here and below that this'll return an error if the height/hash isn't in the best chain.

netsync/manager_test.go
```diff
@@ -0,0 +1,242 @@
...
+
+	file, err := os.Open(filename)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Create a scanner
+	scanner := bufio.NewScanner(file)
+
+	// Read line by line
```
Roasbeef: Comments should end in a period, but this here and above doesn't seem all that useful.

netsync/manager.go
```diff
@@ -1075,99 +1075,40 @@ func (sm *SyncManager) handleHeadersMsg(hmsg *headersMsg) {
...
-			e := sm.headerList.PushBack(&node)
-			if sm.startHeader == nil {
-				sm.startHeader = e
-			}
-		} else {
-			log.Warnf("Received block header that does not "+
-				"properly connect to the chain from peer %s "+
-				"-- disconnecting", peer.Addr())
+		_, err := sm.chain.ProcessBlockHeader(
+			blockHeader, blockchain.BFNone, false)
```
Roasbeef: Style nit: close paren should go on the line below: 
```go
		_, err := sm.chain.ProcessBlockHeader(
			blockHeader, blockchain.BFNone, false,
          )
```

netsync/manager.go
```diff
@@ -1075,99 +1075,40 @@ func (sm *SyncManager) handleHeadersMsg(hmsg *headersMsg) {
...
-	}
-
-	// When this header is a checkpoint, switch to fetching the blocks for
-	// all of the headers since the last checkpoint.
-	if receivedCheckpoint {
-		// Since the first entry of the list is always the final block
-		// that is already in the database and is only used to ensure
-		// the next header links properly, it must be removed before
-		// fetching the blocks.
-		sm.headerList.Remove(sm.headerList.Front())
```
Roasbeef, line 1047: Can we remove the header list all together now? Or is it still needed for re-orgs?

netsync/manager.go
```diff
@@ -327,127 +327,52 @@ func (sm *SyncManager) startSync() {
...
-		// select.
-		higherPeers = append(higherPeers, peer)
-	}
-
-	if sm.chain.IsCurrent() && len(higherPeers) == 0 {
-		log.Infof("Caught up to block %s(%d)", best.Hash.String(), best.Height)
+	// If we're in the initial block download mode, check if we have
+	// peers that we can download headers from.
+	_, bestHeaderHeight := sm.chain.BestHeader()
+	higherHeaderPeers := sm.fetchHigherPeers(bestHeaderHeight)
```
Roasbeef, line 313: Even though many years have passed at this point, we should retain the logic that we removed to filter out non-segwit peers once segwit is active.

netsync/manager.go
```diff
@@ -876,60 +857,41 @@ func (sm *SyncManager) handleBlockMsg(bmsg *blockMsg) {
...
-	// request more blocks using the header list when the request queue is
-	// getting short.
-	if !isCheckpointBlock {
-		if sm.startHeader != nil &&
-			len(state.requestedBlocks) < minInFlightBlocks {
-			sm.fetchHeaderBlocks()
+	// If we're on a checkpointed block, check if we still have checkpoints
+	// to let the user know if we're switching to normal mode.
+	if isCheckpointBlock {
+		log.Infof("on checkpoint block %v(%v)",
```
Roasbeef: Log suggestion: 
> Continuing IBD, on check point block %v(%v)

netsync/manager.go
```diff
@@ -200,30 +199,11 @@ type SyncManager struct {
 
 	// The following fields are used for headers-first mode.
 	headersFirstMode bool
-	headerList       *list.List
-	startHeader      *list.Element
-	nextCheckpoint   *chaincfg.Checkpoint
```
Roasbeef, line 205: 👍

kmk142789 (approved):

blockchain/blockindex.go
kcalvinalvin: Then you can't communicate not having headers for a given blockhash so I think having this is better.

You can use `!HaveData()` to check if you have the headers but then what do you do when you want to check if you have a block header.

blockchain/process_test.go
kcalvinalvin: Will remove

blockchain/process_test.go
kcalvinalvin: Will change

blockchain/process_test.go
kcalvinalvin: Will remove

blockchain/chain.go
kcalvinalvin: Good catch

blockchain/chain.go
kcalvinalvin: Will add

netsync/manager_test.go
kcalvinalvin: Will remove

netsync/manager.go
kcalvinalvin: Yes we can remove it. Done so in a later commit in this PR

netsync/manager.go
kcalvinalvin: Will change

netsync/manager.go
kcalvinalvin: In `sm.fetchHigherPeers()`, we filter out peers that aren't sync candidates. The peers can only be sync candidates if they're a segwit activated peer.

This check is done in `isSyncCandidate()`

netsync/manager.go
kcalvinalvin: Will apply

netsync/manager.go
kcalvinalvin: Ye

kcalvinalvin:
Addressed all the review comments by roasbeef

Roasbeef:
Tested in the wild, and does what it says on the tin: 
<details>
```
2025-12-17 18:32:10.885 [INF] SYNC: Downloading headers for blocks 136001 to 2401568 from peer PEER_A:18333
2025-12-17 18:32:10.885 [INF] SYNC: Lost peer PEER_B:18333 (outbound)
2025-12-17 18:32:16.309 [INF] SYNC: New valid peer PEER_C:18333 (outbound) (/Satoshi:27.1.0/)
2025-12-17 18:32:40.885 [INF] SYNC: Downloading headers for blocks 136001 to 4807558 from peer PEER_D:18333
2025-12-17 18:32:40.885 [INF] SYNC: Lost peer PEER_A:18333 (outbound)
2025-12-17 18:32:41.044 [INF] SYNC: New valid peer PEER_E:18333 (outbound) (/Satoshi:28.1.0/)
2025-12-17 18:32:41.135 [INF] SYNC: New valid peer PEER_F:18333 (outbound) (/Satoshi:27.0.0/)
2025-12-17 18:32:46.438 [INF] CHAN: Verified checkpoint at height 200000/block 0000000000287bffd321963ef05feab753ebe274e1d78b2fd4e2bfe9ad3aa6f2
2025-12-17 18:32:51.245 [INF] SYNC: New valid peer PEER_G:18333 (outbound) (/Satoshi:28.1.0/)
2025-12-17 18:32:55.171 [INF] CHAN: Verified checkpoint at height 300001/block 0000000000004829474748f3d1bc8fcf893c88be255e6d7f571c548aff57abf4
2025-12-17 18:33:03.103 [INF] CHAN: Verified checkpoint at height 400002/block 0000000005e2c73b8ecb82ae2dbc2e8274614ebad7172b53528aba7501f5a089
2025-12-17 18:33:10.885 [INF] SYNC: Downloading headers for blocks 490001 to 4807558 from peer PEER_G:18333
2025-12-17 18:33:10.885 [INF] SYNC: Lost peer PEER_D:18333 (outbound)
2025-12-17 18:33:13.135 [INF] CHAN: Verified checkpoint at height 500011/block 00000000000929f63977fbac92ff570a9bd9e7715401ee96f2848f7b07750b02
2025-12-17 18:33:25.990 [INF] CHAN: Verified checkpoint at height 600002/block 000000000001f471389afd6ee94dcace5ccc44adc18e8bff402443f034b07240
2025-12-17 18:33:38.331 [INF] CHAN: Verified checkpoint at height 700000/block 000000000000406178b12a4dea3b27e13b3c4fe4510994fd667d7c1e6a3f4dc1
2025-12-17 18:33:40.884 [INF] SYNC: Downloading headers for blocks 718001 to 4807558 from peer PEER_E:18333
2025-12-17 18:33:40.888 [INF] SYNC: Lost peer PEER_G:18333 (outbound)
2025-12-17 18:33:46.254 [INF] SYNC: New valid peer PEER_H:18333 (outbound) (/Satoshi:25.1.0/)
2025-12-17 18:33:50.353 [INF] CHAN: Verified checkpoint at height 800010/block 000000000017ed35296433190b6829db01e657d80631d43f5983fa403bfdb4c1
2025-12-17 18:34:01.433 [INF] CHAN: Verified checkpoint at height 900000/block 0000000000356f8d8924556e765b7a94aaebc6b5c8685dcfa2b1ee8b41acd89b
2025-12-17 18:34:10.884 [INF] SYNC: Downloading headers for blocks 986001 to 4807558 from peer PEER_E:18333
2025-12-17 18:34:10.884 [INF] SYNC: Lost peer PEER_E:18333 (outbound)
2025-12-17 18:34:10.884 [INF] SYNC: Downloading headers for blocks 986001 to 4807558 from peer PEER_H:18333
2025-12-17 18:34:11.269 [INF] SYNC: New valid peer PEER_I:18333 (outbound) (/Satoshi:25.1.0/)
2025-12-17 18:34:11.270 [INF] SYNC: New valid peer PEER_J:18333 (outbound) (/Satoshi:0.20.1/)
2025-12-17 18:34:13.213 [INF] CHAN: Verified checkpoint at height 1000007/block 00000000001ccb893d8a1f25b70ad173ce955e5f50124261bbbc50379a612ddf
2025-12-17 18:34:16.312 [INF] SYNC: New valid peer PEER_K:18333 (outbound) (/Satoshi:27.0.0/)
2025-12-17 18:34:25.081 [INF] CHAN: Verified checkpoint at height 1100007/block 00000000000abc7b2cd18768ab3dee20857326a818d1946ed6796f42d66dd1e8
```
</details> 



One thing I noticed though is that the peers may D/C us along the way, so we rotate through a few peers to fetch all the headers. This is testnet3 , there's quite a lot of headers (over 4 million). Are we hitting something like default per-peer upload limit for `bitcoind`? 

IIUC this is a stepping stone for multi-peer header download, so perhaps that'll be sorted out as we'll fetch blocks of headers from many peers in parallel.

netsync/manager.go
```diff
@@ -252,137 +232,127 @@ func (sm *SyncManager) findNextHeaderCheckpoint(height int32) *chaincfg.Checkpoi
...
+}
+
+// fetchHeaders randomly picks a peer that has a higher advertised header
+// and pushes a get headers message to it.
+func (sm *SyncManager) fetchHeaders() {
+	_, height := sm.chain.BestHeader()
+	higherPeers := sm.fetchHigherPeers(height)
+	var bestPeer *peerpkg.Peer
+	switch {
+	case len(higherPeers) > 0:
```
Roasbeef: I think we can simplify this a bit as: 
```go
if len(higherPeers) == 0 {"
     log 
    return
}

bestPeer = higherPeers[rand.Intn(len(higherPeers))]
```

blockchain/process_test.go
```diff
@@ -0,0 +1,180 @@
...
+				blockHeight))
+		}
+		headers = append(headers, &header)
+		tip = &header
+	}
+
+	return headers
+}
+
+func TestProcessBlockHeader(t *testing.T) {
```
Abdulkbk, line 57: nit: godoc

netsync/manager.go
```diff
@@ -690,6 +690,47 @@ func (sm *SyncManager) current() bool {
...
+		return false, blockchain.BFNone
+	}
+
+	isCheckpointBlock := false
+	behaviorFlags := blockchain.BFNone
+
+	// If we don't already know this is a valid header, return false and
+	// BFNone.
+	if !sm.chain.IsValidHeader(blockHash) {
+		return false, blockchain.BFNone
```
Abdulkbk, line 678: do we want to be explicit? else we can just return the decalred variables.

netsync/manager_test.go
```diff
@@ -0,0 +1,239 @@
...
+	return chain, teardown, nil
+}
+
+// loadHeaders loads headers from mainnet from 1 to 11.
+func loadHeaders(t *testing.T) []*wire.BlockHeader {
+	testFile := "blockheaders-mainnet-1-11.txt"
+	filename := filepath.Join("testdata/", testFile)
+
+	file, err := os.Open(filename)
+	if err != nil {
```
Abdulkbk, line 110: nit: use `require.NoError`

netsync/manager_test.go
```diff
@@ -0,0 +1,239 @@
...
+		t.Fatal(err)
+	}
+
+	headers := make([]*wire.BlockHeader, 0, 10)
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := scanner.Text()
+		b, err := hex.DecodeString(line)
+		if err != nil {
```
Abdulkbk, line 119: nit: use the `require` package, and other places where `t.Fatal` was used.

netsync/manager_test.go
```diff
@@ -0,0 +1,239 @@
...
+	// Setup SyncManager with headers processed.
+	headers := loadHeaders(t)
+	for _, header := range headers {
+		isMainChain, err := sm.chain.ProcessBlockHeader(
+			header, blockchain.BFNone, false)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if !isMainChain {
```
Abdulkbk, line 180: yeah we can use `require.True`

netsync/manager_test.go
```diff
@@ -0,0 +1,239 @@
...
+			t.Fatalf("expected block header %v to be in the main chain",
+				header.BlockHash())
+		}
+	}
+	if err != nil {
```
Abdulkbk: I think this check is redundant.

netsync/manager_test.go
```diff
@@ -237,3 +238,62 @@ func TestCheckHeadersList(t *testing.T) {
 		require.Equal(t, test.behaviorFlags, gotFlags)
 	}
 }
+
```
Abdulkbk, line 237: missing godoc

netsync/manager_test.go
```diff
@@ -297,3 +298,162 @@ func TestFetchHigherPeers(t *testing.T) {
...
+}
+
+// Offset isn't relevant so we just return 0.
+//
+// Part of the MedianTimeSource interface implementation.
+func (m *mockTimeSource) Offset() time.Duration {
+	return 0
+}
+
+func TestIsInIBDMode(t *testing.T) {
```
Abdulkbk, line 424: godoc

blockchain/accept.go
```diff
@@ -92,3 +93,141 @@ func (b *BlockChain) maybeAcceptBlock(block *btcutil.Block, flags BehaviorFlags)
...
+	err := CheckBlockHeaderSanity(
+		header, b.chainParams.PowLimit, b.timeSource, flags)
+	if err != nil {
+		return false, err
+	}
+
+	// Orphan headers are not allowed and this function should never be called
+	// with the genesis block.
+	prevHash := &header.PrevBlock
+	prevNode := b.index.LookupNode(prevHash)
```
Abdulkbk: seems like a duplicate prevnode lookup, we just did it above in the `if node != nil` block. Maybe we should consolidate it?.

netsync/manager.go
```diff
@@ -284,6 +284,39 @@ func (sm *SyncManager) isInIBDMode() bool {
 	return true
 }
 
+// fetchHeaders randomly picks a peer that has a higher advertised header
+// and pushes a get headers message to it.
+func (sm *SyncManager) fetchHeaders() {
+	_, height := sm.chain.BestHeader()
+	higherPeers := sm.fetchHigherPeers(height)
+	var bestPeer *peerpkg.Peer
+	switch {
```
Abdulkbk: do we expect other cases in the future else we can just use `if`?

yyforyongyu (commented):
Cool stuff - left a few comments that i think we need to fix before merging.

blockchain/accept.go
```diff
@@ -92,3 +93,141 @@ func (b *BlockChain) maybeAcceptBlock(block *btcutil.Block, flags BehaviorFlags)
...
+
+	// Create a new block node for the block and add it to the block index.
+	//
+	// Note that the additional information for the actual transactions and
+	// witnesses in the block can't be populated until the full block data is
+	// known since that information is not available in the header.
+	if node == nil {
+		node = newBlockNode(header, prevNode)
+		node.status = statusHeaderStored
+		b.index.AddNode(node)
```
yyforyongyu, line 186: The `if node == nil` guard here correctly avoids overwriting an existing node. But `maybeAcceptBlock` at line 67 has no such guard — it unconditionally creates a fresh `*blockNode` and calls `AddNode`, overwriting the index entry. If this header-only node already exists in `bestHeader`'s chainView, the old pointer becomes orphaned while the chainView still holds it.

After that, `bestHeader.Contains(b.index.LookupNode(&hash))` returns false because `LookupNode` returns the new pointer while `bestHeader` still has the old one. This breaks `IsValidHeader`, `HeaderHashByHeight`, and `checkHeadersList` in netsync.

Could `maybeAcceptBlock` check if a node already exists and upgrade its status to `statusDataStored` rather than creating a new one? Similar to the `if node == nil` pattern here.

blockchain/accept.go
```diff
@@ -92,3 +93,141 @@ func (b *BlockChain) maybeAcceptBlock(block *btcutil.Block, flags BehaviorFlags)
...
+		prevHash := &header.PrevBlock
+		prevNode := b.index.LookupNode(prevHash)
+		if prevNode == nil {
+			str := fmt.Sprintf("previous block %s is unknown", prevHash)
+			return false, ruleError(ErrPreviousBlockUnknown, str)
+		} else if b.index.NodeStatus(prevNode).KnownInvalid() {
+			str := fmt.Sprintf(
+				"previous block %s is known to be invalid", prevHash)
+			return false, ruleError(ErrInvalidAncestorBlock, str)
+		}
```
yyforyongyu, line 155: Small gap here -- when the header is already known (`node != nil`), the code checks `prevNode.KnownInvalid()` but not `node.status.KnownInvalid()` itself. The godoc at line 109-111 promises we return an error for known-invalid headers, so we should probably add that check before the `bestHeader.Contains` early return.

blockchain/blockindex.go
```diff
@@ -377,14 +385,15 @@ func newBlockIndex(db database.DB, chainParams *chaincfg.Params) *blockIndex {
...
+// and if the data exists on disk.
 //
 // This function is safe for concurrent access.
 func (bi *blockIndex) HaveBlock(hash *chainhash.Hash) bool {
 	bi.RLock()
-	_, hasBlock := bi.index[*hash]
+	node, hasBlock := bi.index[*hash]
 	bi.RUnlock()
-	return hasBlock
+	return hasBlock && node.status.HaveData()
```
yyforyongyu: Small concurrency note -- `HaveBlock` now reads `node.status` after releasing `RLock` at line 395. The old code only returned a captured bool. Since `SetStatusFlags`/`UnsetStatusFlags` mutate `status` under the write lock, this is technically a data race under Go's memory model, even though byte reads are usually atomic in practice. Moving the `HaveData()` check inside the lock would keep it clean.

blockchain/blockindex.go
```diff
@@ -499,6 +508,17 @@ func (bi *blockIndex) flushToDB() error {
...
 	err := bi.db.Update(func(dbTx database.Tx) error {
 		for node := range bi.dirty {
+			// NOTE: we specifically don't flush the block indexes that
+			// we don't have the data for backwards compatibility.
+			// While flushing would save us the work of re-downloading
+			// the block headers upon restart, if the user were to start
+			// up a btcd node with an older version, it would result in
+			// an unrecoverable error as older versions would consider a
+			// blockNode being present as having the block data as well.
+			if node.status.HaveHeader() &&
```
yyforyongyu, line 537: `flushToDB` opens a DB write transaction whenever dirty is non-empty, but during header processing every dirty node is header-only and gets skipped here. That means each `ProcessBlockHeader` call ends up doing a write txn that writes nothing. During long header sync that's a lot of no-op transactions. Maybe we could skip the flush entirely in `maybeAcceptBlockHeader`, or have `flushToDB` bail early if all dirty nodes would be skipped?

blockchain/chainio.go
```diff
@@ -1262,6 +1263,9 @@ func (b *BlockChain) initChainState() error {
 			i++
 		}
 
+		// Set the lastNode in the blockindex as the bestheader.
+		b.bestHeader.SetTip(lastNode)
```
yyforyongyu: `initChainState` sets `bestHeader` to `lastNode` here, which is the last entry in the block index bucket's key order (height+hash). If side-chain blocks exist at a height above the best chain tip, this picks the wrong branch. A few lines below, the best *block* chain tip is resolved from `state.hash` at line 1270 — should `bestHeader` be set to the same `tip` node? Since header-only nodes aren't persisted, after restart the header and block views should be identical.

blockchain/chain.go
```diff
@@ -1530,6 +1544,54 @@ func (b *BlockChain) BlockHashByHeight(blockHeight int32) (*chainhash.Hash, erro
...
 
+// IsValidHeader checks that we've already checked that this header connects to the
+// chain of best headers and did not receive an invalid state.
+func (b *BlockChain) IsValidHeader(blockHash *chainhash.Hash) bool {
+	node := b.index.LookupNode(blockHash)
+	if node == nil || !b.bestHeader.Contains(node) {
+		return false
+	}
+
+	return !node.status.KnownInvalid()
```
yyforyongyu: `IsValidHeader` reads `node.status` directly here after all locks are released. Since `SetStatusFlags`/`UnsetStatusFlags` mutate status under the index write lock, this is a data race. Using `b.index.NodeStatus(node).KnownInvalid()` instead of `node.status.KnownInvalid()` would fix it — `NodeStatus` reads under the lock.

netsync/manager.go
```diff
@@ -252,137 +232,127 @@ func (sm *SyncManager) findNextHeaderCheckpoint(height int32) *chaincfg.Checkpoi
...
+// given height.  The peers that are not sync candidates are omitted from the
+// returned list.
+func (sm *SyncManager) fetchHigherPeers(height int32) []*peerpkg.Peer {
+	higherPeers := make([]*peerpkg.Peer, 0, len(sm.peerStates))
+	for peer, state := range sm.peerStates {
+		if !state.syncCandidate {
+			continue
+		}
+
+		if peer.LastBlock() <= height {
```
yyforyongyu, line 245: `fetchHigherPeers` uses strictly-greater comparison here, so peers at the same height are excluded. If all connected peers are at the same height as the local tip, we never pick a sync peer. Combined with the inv gating at line 1088, this can stall progress when we actually need blocks from equal-height peers. An `>=` comparison with a tiebreaker (or fallback) might be worth considering.

netsync/manager.go
```diff
@@ -845,85 +831,64 @@ func (sm *SyncManager) handleBlockMsg(bmsg *blockMsg) {
...
-				"peer %s: %v", peer.Addr(), err)
-			return
-		}
-		log.Infof("Downloading headers for blocks %d to %d from "+
-			"peer %s", prevHeight+1, sm.nextCheckpoint.Height,
-			sm.syncPeer.Addr())
+	// Fetch more blocks if we're still not caught up to the best header and
+	// don't have any more requestedBlocks.
+	_, lastHeight := sm.chain.BestHeader()
+	if bmsg.block.Height() < lastHeight && len(state.requestedBlocks) == 0 {
```
yyforyongyu: This waits for `len(state.requestedBlocks) == 0` before requesting more blocks. The old code refilled at `< minInFlightBlocks`, keeping the pipeline warm. Waiting for full drain means we're single-threading block downloads — there's always a gap between the last block arriving and the next batch being requested. This could be a meaningful IBD throughput regression.

netsync/manager.go
```diff
@@ -845,85 +831,64 @@ func (sm *SyncManager) handleBlockMsg(bmsg *blockMsg) {
...
-		return
-	}
-
 	// Build up a getdata request for the list of blocks the headers
 	// describe.  The size hint will be limited to wire.MaxInvPerMsg by
 	// the function, so no need to double check it here.
-	gdmsg := wire.NewMsgGetDataSizeHint(uint(sm.headerList.Len()))
+	bestState := sm.chain.BestSnapshot()
+	_, bestHeaderHeight := sm.chain.BestHeader()
+	length := bestHeaderHeight - bestState.Height
```
yyforyongyu: If `bestHeaderHeight` is somehow less than `bestState.Height`, the subtraction goes negative and the uint cast wraps to a huge value. The clamp in `NewMsgGetDataSizeHint` prevents a crash, but you'd end up with a silent stall. A quick guard or `max(0, ...)` would make the intent clearer.

netsync/manager.go
```diff
@@ -969,99 +934,41 @@ func (sm *SyncManager) handleHeadersMsg(hmsg *headersMsg) {
 	// The remote peer is misbehaving if we didn't request headers.
 	msg := hmsg.headers
 	numHeaders := len(msg.Headers)
-	if !sm.headersFirstMode {
-		log.Warnf("Got %d unrequested headers from %s -- "+
-			"disconnecting", numHeaders, peer.Addr())
-		peer.Disconnect()
-		return
-	}
```
yyforyongyu: The old `headersFirstMode` guard got removed but there's no replacement check for unsolicited headers. In non-IBD with `syncPeer == nil`, a peer sending valid unsolicited headers can drive `handleHeadersMsg` into `fetchHeaderBlocks` (line 971), which dereferences `sm.syncPeer` at lines 899/907/920 — that's a nil pointer panic. Might need either restoring a request/response guard or adding a nil check on `syncPeer` before entering `fetchHeaderBlocks`.

netsync/manager.go
```diff
@@ -969,99 +934,41 @@ func (sm *SyncManager) handleHeadersMsg(hmsg *headersMsg) {
...
-	// all of the headers since the last checkpoint.
-	if receivedCheckpoint {
-		// Since the first entry of the list is always the final block
-		// that is already in the database and is only used to ensure
-		// the next header links properly, it must be removed before
-		// fetching the blocks.
-		sm.headerList.Remove(sm.headerList.Front())
-		log.Infof("Received %v block headers: Fetching blocks",
-			sm.headerList.Len())
 		sm.progressLogger.SetLastLogTime(time.Now())
```
yyforyongyu, line 993: `fetchHeaders` at line 294 sends getheaders but never refreshes `lastProgressTime`, and `handleHeadersMsg` doesn't update it on successful header processing either. With ~870k mainnet headers and a 3-minute stall timeout, the stall handler at line 470 will fire mid-header-download and disconnect a healthy sync peer. This would hit almost immediately on mainnet IBD. We probably need a `lastProgressTime` refresh in the header processing path, similar to how block processing refreshes it.

netsync/manager.go
```diff
@@ -969,99 +934,41 @@ func (sm *SyncManager) handleHeadersMsg(hmsg *headersMsg) {
...
-	// next checkpoint.
-	locator := blockchain.BlockLocator([]*chainhash.Hash{finalHash})
-	err := peer.PushGetHeadersMsg(locator, sm.nextCheckpoint.Hash)
-	if err != nil {
-		log.Warnf("Failed to send getheaders message to "+
-			"peer %s: %v", peer.Addr(), err)
-		return
+	bestHash, bestHeight := sm.chain.BestHeader()
+	if sm.ibdMode {
+		if bestHeight < sm.syncPeer.LastBlock() {
```
yyforyongyu, line 1009: `sm.syncPeer.LastBlock()` here has no nil guard. `syncPeer` can already be nil when we enter `handleHeadersMsg` (e.g., if a peer disconnect was processed first), so this is a reachable panic. A quick nil check before this call would make it safe.

netsync/manager_test.go
```diff
@@ -0,0 +1,459 @@
...
+		Chain:       chain,
+		ChainParams: params,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	return sm, tearDown
+}
+
+func TestCheckHeadersList(t *testing.T) {
```
yyforyongyu, line 151: The new tests cover the helper predicates well, but the risky control flow in `handleHeadersMsg`, `fetchHeaderBlocks`, `startSync`, and stall handling isn't exercised. Those are where the trickiest behavior changes live — would be great to have at least a basic end-to-end test that drives headers through to block download, so regressions in the sync state machine get caught.

blockchain/process_test.go
```diff
@@ -0,0 +1,180 @@
...
+	for _, header := range headers {
+		isMainChain, err := chain.ProcessBlockHeader(header, BFNone, false)
+		require.NoError(t, err)
+		require.True(t, isMainChain)
+	}
+
+	// Check that the tip is correct.
+	lastHeader := headers[len(headers)-1]
+	lastHeaderHash := lastHeader.BlockHash()
+	tipNode := chain.bestHeader.Tip()
```
yyforyongyu, line 86: Small thing — `tipNode` is captured once here and reused in assertions at lines 134 and 179. If `bestHeader` moved unexpectedly, these checks would still pass against the stale pointer. Re-fetching `chain.bestHeader.Tip()` for each assertion would make the "tip didn't move" checks more robust.

yyforyongyu (changes requested):

netsync/manager.go
kcalvinalvin: Yes I think it's better to be explicit. Easier to review

blockchain/process_test.go
kcalvinalvin: I guess it's better to have comments on the tests too but test files are exempt from go documentation unless if they're example functions.

I'll skip them in this round to focus on other parts of the review.

netsync/manager_test.go
kcalvinalvin: I'll skip the require package changes so it's easier to diff with other parts of the code change from the outstanding reviews.

netsync/manager_test.go
kcalvinalvin: I'll skip the require package changes so it's easier to diff with other parts of the code change from the outstanding reviews.

netsync/manager_test.go
kcalvinalvin: I'll skip the require package changes so it's easier to diff with other parts of the code change from the outstanding reviews.

netsync/manager_test.go
kcalvinalvin: Same as above. I'll skip on these on this round

netsync/manager_test.go
kcalvinalvin: Same as above. Since godoc skips tests I'll skip on these for now

netsync/manager_test.go
kcalvinalvin: Removed the redundant if branch in the latest push

blockchain/accept.go
kcalvinalvin: Simplified the logic and removed the lookup in the branching in the latest push

netsync/manager.go
kcalvinalvin: We can just use if. Applied in the latest push

blockchain/accept.go
kcalvinalvin: Addressed and added a test for this. The fix will be included in my next push

blockchain/accept.go
kcalvinalvin: Addressed. The fix will be included in my next push

blockchain/blockindex.go
kcalvinalvin: Addressed. The fix will be included in my next push

blockchain/blockindex.go
kcalvinalvin: Ended up going with the fix where we bail early. Will be included in my next push

yyforyongyu:
Cool lmk when this is ready for another look!

blockchain/chainio.go
kcalvinalvin: Good catch. Addressed and will be included in my next push.

blockchain/chain.go
kcalvinalvin: Will be addressed in my next push

netsync/manager.go
kcalvinalvin: ~~It should be fine if we don't have a syncPeer because once we're caught up to the tip, we get block headers from our peers instead of invs. And these headers aren't requested, they're just relayed to us by any connected peer.~~

The syncPeer is only used during ibd and isn't when we're all caught up so I think this is the correct behavior.

EDIT: ok I checked again and we're not yet getting headers instead of invs from our peers. We could/should be getting headers instead of invs though. However, the comment still stands because we request blocks from the peer that gave us the inv for the block, not the syncPeer.

netsync/manager.go
kcalvinalvin: Addressed. Will be included in my next push

netsync/manager.go
kcalvinalvin: It should never happen and if bestHeaderHeight is less than bestState.Height (now forkHeight), we won't be downloading any blocks.

I'll guard with a check to make sure bestHeaderHeight is greater than forkHeight and return early if it isn't.

netsync/manager.go
kcalvinalvin: Oh hmm good point. We'd want to switch over to only receiving headers (I checked and I haven't done it in this PR) and in that case, we'd get unsolicited headers.

I'll fix this by making the caller for `fetchHeaderBlocks` pass in the peer they wanna fetch from.

netsync/manager.go
kcalvinalvin: Will add a nil guard

netsync/manager.go
kcalvinalvin: Good point. Address and the fix will be included in my next push

blockchain/process_test.go
kcalvinalvin: Addressed. Is included in the latest push

kcalvinalvin:
All the outstanding review comments from both reviewers have been addressed.

~~One thing I'm still working on is https://github.com/btcsuite/btcd/pull/2428#discussion_r2791188069 from YY but that's for adding new tests and isn't a blocker for re-addressing the reviews.~~
Completed now.

netsync/manager_test.go
kcalvinalvin: Addressed as well now in the latest push. Covered mainly by `TestSyncStateMachine`

netsync/manager.go
```diff
@@ -262,127 +302,60 @@ func (sm *SyncManager) startSync() {
...
 
-	// Pick randomly from the set of peers greater than our block height,
-	// falling back to a random peer of the same height if none are greater.
+	// We don't have any more headers to download at this
+	// point so start asking for blocks.
+	best := sm.chain.BestSnapshot()
+	higherPeers := sm.fetchHigherPeers(best.Height)
+
+	// Pick randomly from the set of peers greater than our
+	// block height, falling back to a random peer of the same
```
yyforyongyu, line 332: We don't do the fallback (tho think we should)?

netsync/manager.go
yyforyongyu: > request blocks from the peer that gave us the inv for the block, not the syncPeer.

Correct - but we do have this check before we process the `inv`, at L1137
```go
if peer != sm.syncPeer && !sm.current() {
```
If `syncPeer` is nil, the `inv` processing will be aborted.

kcalvinalvin: Yes but it's && if the sync manager is not current.

If the sync manager is current, it's ok if syncPeer == nil, we'll still process the inv.

I wrote up this test that test that the inv is processed even if the syncPeer is nil:

```
// TestHandleInvNilSyncPeer tests the inv-processing guard that ignores
// invs from non-sync peers when the chain is not current.  With a nil
// syncPeer, every peer is "not the sync peer".
//
// When the chain is current, current() returns true, the guard does NOT
// fire, and the inv is processed.  When the chain is not current,         
// current() returns false, the guard fires, and the inv is ignored.
func TestHandleInvNilSyncPeer(t *testing.T) {                                   
        t.Parallel()
                                                                                                                
        tests := []struct {
                name             string
                setup            func(*testing.T, *SyncManager)
                wantInvProcessed bool
        }{
                {
                        name: "current chain processes inv",
                        setup: func(t *testing.T, sm *SyncManager) {
                                // Mine a block with a recent timestamp
                                // so IsCurrent() returns true.
                                params := sm.chainParams
                                cb := createTestCoinbase(1, params)
                                header := wire.BlockHeader{
                                        Version:    1,
                                        PrevBlock:  *params.GenesisHash,
                                        MerkleRoot: cb.TxHash(),
                                        Timestamp:  time.Now().Truncate(
                                                time.Second),
                                        Bits: params.PowLimitBits,
                                }
                                require.True(t,
                                        solveTestBlock(&header, params))

                                block := btcutil.NewBlock(&wire.MsgBlock{
                                        Header:       header,
                                        Transactions: []*wire.MsgTx{cb},
                                })
                                _, _, err := sm.chain.ProcessBlock(
                                        block, blockchain.BFNone)
                                require.NoError(t, err)
                                require.True(t, sm.chain.IsCurrent())
                        },
                        wantInvProcessed: true,
                },
                {
                        name: "non-current chain ignores inv",
                        setup: func(t *testing.T, sm *SyncManager) {
                                // Genesis-only regtest chain with old
                                // timestamps is not current.
                                require.False(t, sm.chain.IsCurrent())
                        },
                        wantInvProcessed: false,
                },
        }

        for _, tc := range tests {
                t.Run(tc.name, func(t *testing.T) {
                        t.Parallel()

                        params := chaincfg.RegressionNetParams
                        params.Checkpoints = nil

                        sm, tearDown := makeMockSyncManager(t, &params)
                        defer tearDown()

                        tc.setup(t, sm)

                        require.Nil(t, sm.syncPeer)

                        invPeer := newSyncCandidate(t, sm, 1)
                        peerState := sm.peerStates[invPeer]

                        unknownHash := chainhash.Hash{0xab}
                        inv := wire.NewMsgInv()
                        err := inv.AddInvVect(wire.NewInvVect(
                                wire.InvTypeWitnessBlock, &unknownHash,
                        ))
                        require.NoError(t, err)

                        sm.handleInvMsg(&invMsg{
                                inv: inv, peer: invPeer,
                        })

                        _, requested := peerState.requestedBlocks[unknownHash]
                        require.Equal(t, tc.wantInvProcessed, requested)

                        _, globalRequested := sm.requestedBlocks[unknownHash]
                        require.Equal(t, tc.wantInvProcessed, globalRequested)
                })
        }
}
```

netsync/manager.go
kcalvinalvin: I'll update the comment. I still think it's ok to fetch just higher peers.

kcalvinalvin: Oh shoot no you're right. We have headers at this point but not blocks so we can fetch blocks from peers at the same height.

~~Ok I'll change this.~~

Changed in the latest push

netsync/manager.go
kcalvinalvin: Ok so this should really be a non-issue now since:

1: We'd still process the inv when syncPeer == nil as long as sm.isCurrent() == true.
2: We fetch same height peers during block downloads as well now.

netsync/manager.go
```diff
@@ -1685,19 +1668,11 @@ func New(config *Config) (*SyncManager, error) {
...
-	if !config.DisableCheckpoints {
-		// Initialize the next checkpoint based on the current height.
-		sm.nextCheckpoint = sm.findNextHeaderCheckpoint(best.Height)
-		if sm.nextCheckpoint != nil {
-			sm.resetHeaderState(&best.Hash, best.Height)
-		}
-	} else {
+	if config.DisableCheckpoints {
 		log.Info("Checkpoints are disabled")
 	}
```
yyforyongyu, line 1651: There is an edge case, which I believe will probably only happen in regtest/testnet. Say `sm.chain.IsCurrent()` is false if the tip block timestamp is older than 24 hours, even if the node's height matches peers.

1. `startSync()` sees `!IsCurrent()` and enters IBD logic at L322.
2. Since there are no higher-header peers, it falls back to same-height block peers at L359.
3. It sets `sm.syncPeer` and `sm.ibdMode` = true at L375.
4. But if `bestHeader == bestBlock`, `buildBlockRequest()` has nothing to ask for, so `fetchHeaderBlocks()` sends no getdata at L911.

and if that sync peer later announces a new block in the normal way `inv`, we can still be stuck because this line.

yyforyongyu (approved):
LGTM🥇

The only comment I have is this [one](https://github.com/btcsuite/btcd/pull/2428/changes#r2920007215), it's non-blocking as I believe it's extremely unlikely to happen on mainnet, tho plausible on testnet or regtest, while I still consider it an edge case, I think it's good to flag it here for future ref.

netsync/manager.go
kcalvinalvin: Actually no, we've already downloaded all the headers and we're using BestState() so we should strictly be using only higher peers here. I'll revert my last push.

netsync/manager.go
kcalvinalvin: Wouldn't this be fixed if we only fetched from higher peers? I'll revert the last change.

PR #2432, opened 2025-09-25
multi: make v3 transaction standard (no TRUC impl yet)

Roasbeef:
In this commit, we make a series of changes to make v3 transactions standard. This means that we'll relay, then and accept them into our mempool. This PR doesn't yet implement the new TRUC semantics yet.


saubyk:
cc: @bhandras for review

integration/v3_transaction_test.go
```diff
@@ -0,0 +1,319 @@
...
+	t.Cleanup(func() {
+		require.NoError(t, r.TearDown())
+	})
+
+	// Make sure that normal v1 transactions are still accepted.
+	t.Run("v1 transaction accepted", func(t *testing.T) {
+		testVersionedTransaction(t, r, 1, true)
+	})
+
+	// MAke sure that v2 transactions are still accepted.
```
kcalvinalvin: s/MAke/Make

integration/v3_transaction_test.go
```diff
@@ -0,0 +1,319 @@
...
+	script, err := txscript.PayToAddrScript(addr)
+	require.NoError(t, err)
+
+	output := &wire.TxOut{
+		PkScript: script,
+		Value:    50000000,
+	}
+
+	var tx *wire.MsgTx
+	if version == 1 {
```
kcalvinalvin: Super nitpicking but can't you just replace the entire if branching with:

`r.CreateTransaction([]*wire.TxOut{output}, 10, true, rpctest.WithTxVersion(version))`.

Not sure why there needs to be a check for `version == 1`.

integration/v3_transaction_test.go
```diff
@@ -0,0 +1,319 @@
...
+		t, found, "v3 transaction should be included in the mined "+
+			"block",
+	)
+
+	// Verify the transaction is no longer in the mempool.
+	mempool, err = r.Client.GetRawMempool()
+	require.NoError(t, err)
+	require.NotContains(t, mempool, v3TxHash)
+
+	// Verify by checking the block contains our transaction. The
```
kcalvinalvin: Eh? The check is the same as the ones in lines 178-180. And the check being described in the comment is also already being done in lines 161-171 & 178-180.

integration/v3_transaction_test.go
```diff
@@ -0,0 +1,319 @@
...
+
+		mempool, err := r.Client.GetRawMempool()
+		require.NoError(t, err)
+		require.NotContains(t, mempool, txHash)
+	})
+}
+
+// TestV3TransactionChaining tests that v3 transactions can be chained
+// (spending outputs from other v3 transactions).
+func TestV3TransactionChaining(t *testing.T) {
```
kcalvinalvin: The test isn't doing what it's claiming to do.

There's no code guaranteeing that the inputs for `tx2` will be the outputs from `tx1`.

And doing a simple println shows that the transactions are not being chained.

```
[I] u0_a306@localhost ~/b/integration (2025-12-04-tmp-test)> go test -tags=rpctest -run=TestV3TransactionChaining
2025-12-04 13:10:52.187 [INF] ITEST: Established connection to RPC server 127.0.0.1:8335
2025-12-04 13:10:52.502 [INF] ITEST: Established connection to RPC server 127.0.0.1:8335
2025-12-04 13:10:52.670 [INF] ITEST: Established connection to RPC server 127.0.0.1:8337
2025-12-04 13:10:52.981 [INF] ITEST: Established connection to RPC server 127.0.0.1:8337
2025-12-04 13:10:53.192 [DBG] ITEST: Detected btcd version: 250000
tx1: 04dd06c83b204a6bc3c6034df8a11306eb7c460e1ac535090bac98c355d2ae01
04dd06c83b204a6bc3c6034df8a11306eb7c460e1ac535090bac98c355d2ae01 0
04dd06c83b204a6bc3c6034df8a11306eb7c460e1ac535090bac98c355d2ae01 1
tx2 in a6ad4800142e85e8a7ffe289eb8f97acefcd5b9cda1d5d1e799a15cbc7d2a5f3:0
tx2: 337115644e3a8195a7ff1ef240c1205a2cf11ce3250a60d8024f5c9d46bbd392
```

Added code:

```
  293 ▎ │       fmt.Println("tx1:", tx1.TxHash().String())                                                                                                                                                                                                             
    1 ▎ │       for i := range tx1.TxOut {                                                                                                                                                                                                                             
    2 ▎ │       │       fmt.Println(tx1.TxHash().String(), i)                                                                                                                                                                                                          
    3 ▎ │       }                                                                                                                                                                                                                                                      
    4 ▎ │       for _, txIn := range tx2.TxIn {                                                                                                                                                                                                                        
    5 ▎ │       │       fmt.Println("tx2 in", txIn.PreviousOutPoint.String())                                                                                                                                                                                          
    6 ▎ │       }                                                                                                                                                                                                                                                      
    7 ▎ │       fmt.Println("tx2:", tx2.TxHash().String())                                                                                                                            
```

integration/v3_transaction_test.go
```diff
@@ -0,0 +1,319 @@
...
+
+	tx1Hash, err := r.Client.SendRawTransaction(tx1, true)
+	require.NoError(t, err)
+	require.NotNil(t, tx1Hash)
+
+	// Mine a block to confirm the first transaction.
+	blocks, err := r.Client.Generate(1)
+	require.NoError(t, err)
+	require.Len(t, blocks, 1)
+
```
kcalvinalvin: Could check that `tx1` is no longer in the mempool & is included in the block at this point.

kcalvinalvin:
Most of it looks good. `TestV3TransactionChaining` needs work though

server.go
```diff
@@ -2973,7 +2973,7 @@ func newServer(listenAddrs, agentBlacklist, agentWhitelist []string,
 			MaxOrphanTxSize:      defaultMaxOrphanTxSize,
 			MaxSigOpCostPerTx:    blockchain.MaxBlockSigOpsCost / 4,
 			MinRelayTxFee:        cfg.minRelayTxFee,
-			MaxTxVersion:         2,
+			MaxTxVersion:         mempool.MaxStandardTxVersion,
```
yyforyongyu, line 2976: 👍

yyforyongyu: a minor point: given that TRUC is not implemented, should we move this to a side branch instead?

integration/rpctest/memwallet.go
```diff
@@ -481,12 +503,18 @@ func (m *memWallet) SendOutputsWithoutChange(outputs []*wire.TxOut,
...
 // This function is safe for concurrent access.
 func (m *memWallet) CreateTransaction(outputs []*wire.TxOut,
-	feeRate btcutil.Amount, change bool) (*wire.MsgTx, error) {
+	feeRate btcutil.Amount, change bool, options ...CreateTxOption) (*wire.MsgTx, error) {
 
 	m.Lock()
 	defer m.Unlock()
 
-	tx := wire.NewMsgTx(wire.TxVersion)
+	// Apply functional options.
```
yyforyongyu, line 511: Q: should this be applied before the `m.Lock`? in case there is another lock in `options`

integration/v3_transaction_test.go
yyforyongyu: yeah it should be removed

integration/v3_transaction_test.go
```diff
@@ -0,0 +1,319 @@
...
+	// Mine another block.
+	blocks, err = r.Client.Generate(1)
+	require.NoError(t, err)
+	require.Len(t, blocks, 1)
+
+	// Verify the second v3 transaction is no longer in mempool (was mined).
+	mempool, err = r.Client.GetRawMempool()
+	require.NoError(t, err)
+	require.NotContains(t, mempool, tx2Hash)
+}
```
yyforyongyu: cool - tho I think we should also add the case that when `tx1` is unconfirmed, we assert that `tx2` can spend it, given that TRUC is about enforcing new policy in the mempool.

server.go
Roasbeef: Even though we haven't implemented it, right now we won't relay any of the transactions even though they're valid. Also us not supporting it at the package level here prevents devs from easily creating and interacting with these transactions. 

Relay is a bit more flexible, than say consensus in this case.

integration/rpctest/memwallet.go
Roasbeef: Yeah makes sense, will update.

integration/v3_transaction_test.go
Roasbeef: Totally right, simplified.

integration/v3_transaction_test.go
Roasbeef: Done.

integration/v3_transaction_test.go
Roasbeef: I decided to just remove this test, it doesn't really assert anything relevant. It was part of a scratch pad before I started to implement the other TRUC logic.

integration/v3_transaction_test.go
Roasbeef: I decided to just remove this test, it doesn't really assert anything relevant. It was part of a scratch pad before I started to implement the other TRUC logic.

server.go
Roasbeef: Without this, if a dev was using `btcd` for first hop relay, and wanted to use P2A or TRUC or w/e (relying on the rest of the network to implement the policy rules), btcd would reject it as non-standard.

integration/rpctest/memwallet.go
kcalvinalvin: Just sanity checking here but was this applied by removing `m.Lock` from `options`?

kcalvinalvin:
Ok seems like things are good. Just want to sanity check on this though https://github.com/btcsuite/btcd/pull/2432#discussion_r2598132775

PR #2433, opened 2025-09-27
multi: recognize new standard P2A output type 

Roasbeef:
In this PR, we add initial recognition of the new P2A output type. It actually overloads the existing Taproot witness version, but with a distinct witness program. We adopt the proposed standardness rules to make only spends without any witness data standard. Utilities to parse the script, make addresses, etc - have also been added. 

Note that this doesn't include the logic to support "ephemeral dust", which enables zero fee and zero value outputs under certain conditions. 

A replace dir is used as we modified `btcutil` in this PR.


txscript/pkscript.go
```diff
@@ -155,6 +162,11 @@ func (s PkScript) Address(chainParams *chaincfg.Params) (btcutil.Address, error)
 		return nil, fmt.Errorf("unable to parse address: %v", err)
 	}
 
+	// P2A scripts don't have addresses.
```
sputn1ck: there is an address for p2a outputs `bc1pfeessrawgf`
https://mempool.space/address/bc1pfeessrawgf
https://github.com/bitcoin/bitcoin/blob/25212dfdb4cd7291392b6a94130f658c5bfa0a48/test/functional/rpc_validateaddress.py#L171

sputn1ck (approved):
LGTM!

txscript/pkscript.go
Roasbeef: Looks like a stray line, but `ExtractPkScriptAddrs` should handle that. Will add a test for this.

txscript/standard.go
```diff
@@ -875,6 +909,12 @@ func PayToAddrScript(addr btcutil.Address) ([]byte, error) {
 				nilAddrErrStr)
 		}
 		return payToWitnessTaprootScript(addr.ScriptAddress())
+	case *btcutil.AddressPayToAnchor:
```
kcalvinalvin, line 974: nit

`btcutil.AddressPayToAnchor` isn't introduced until commit `3f99cfe0a9b0ed9a33e0c11136127e8b6e2906ea`.

So this commit doesn't build.

txscript/p2a_test.go
```diff
@@ -0,0 +1,147 @@
...
+			want:   false,
+		},
+		{
+			name:   "invalid - wrong length",
+			script: []byte{OP_1, OP_DATA_2, 0x4e},
+			want:   false,
+		},
+		{
+			name:   "invalid - wrong version",
+			script: []byte{OP_0, OP_DATA_2, 0x4e, 0x73},
```
kcalvinalvin, line 34: ~~Just a question: is `OP_1` a version? Curious as to the test's name~~

nvm figured it out

go.mod
```diff
@@ -66,3 +66,5 @@ retract (
 )
 
 go 1.23.2
+
+replace github.com/btcsuite/btcd/btcutil => ./btcutil
```
kcalvinalvin, line 70: Are we ok with this or is this temporary?

btcutil/p2a_address_test.go
```diff
@@ -0,0 +1,155 @@
...
+			wantErr: false,
+		},
+		{
+			name:    "regtest P2A",
+			address: "bcrt1pfeesnyr2tx",
+			net:     &chaincfg.RegressionNetParams,
+			wantErr: false,
+		},
+		{
+			// Address will decode but won't be for the right
```
kcalvinalvin: This isn't testing for anything..?

btcutil/p2a_address_test.go
```diff
@@ -0,0 +1,155 @@
...
+			if !addr.IsForNet(tt.net) {
+				t.Errorf("IsForNet() = false, want true")
+			}
+
+			otherNet := &chaincfg.MainNetParams
+			if tt.net == &chaincfg.MainNetParams {
+				otherNet = &chaincfg.TestNet3Params
+			}
+			if addr.IsForNet(otherNet) {
+				t.Errorf("IsForNet(other) = true, want false")
```
kcalvinalvin: What's this testing for? Why do we want to test "othernet-ness"

btcutil/p2a_address_test.go
```diff
@@ -0,0 +1,155 @@
...
+			}
+
+			expectedScript := []byte{0x51, 0x02, 0x4e, 0x73}
```
kcalvinalvin: I think it's better if I we declare something like the declaration in txscript/standard.go

```
var PayToAnchorScript = []byte{OP_1, OP_DATA_2, 0x4e, 0x73}
```

btcutil/p2a_address_test.go
```diff
@@ -0,0 +1,155 @@
...
+			address: "bc1pfeessrawgf",
+			net:     &chaincfg.TestNet3Params,
+			wantErr: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			addr, err := btcutil.DecodeAddress(tt.address, tt.net)
+			if (err != nil) != tt.wantErr {
```
kcalvinalvin: Wild `if` statement

btcutil/address.go
```diff
@@ -709,3 +723,61 @@ func newAddressTaproot(hrp string,
...
+	}
+	return addr
+}
+
+// ScriptAddress returns the raw bytes of the P2A script to be used when
+// inserting the address into a txout's script. For P2A, this returns the fixed
+// script bytes: OP_1 <0x4e73>.
+func (a *AddressPayToAnchor) ScriptAddress() []byte {
+	// Return the fixed P2A script bytes.
+	return []byte{0x51, 0x02, 0x4e, 0x73}
```
kcalvinalvin: Yeah here too. Would be better if we declared some variable and referred to it

go.mod
Roasbeef: This is temp. It's needed as we ref the newly added addr type in `txscript/standard.go`. 

https://github.com/btcsuite/btcd/pull/1825 will resolve this circualr dep module issue. After this is merged, I'll  be ale to update the module dep, then remove this.

btcutil/p2a_address_test.go
Roasbeef: Just that `IsForNet` properly detects if an address isn't for the target network. 

I agree it's a trivial test, tbh I did it just so the test coverage reported is higher.

btcutil/p2a_address_test.go
Roasbeef: Yeah the issue with that again is the module issue I brought up which will be addressed by #1825. This is defined, but the `btcutil` module wasn't updated (as this was only added in this PR). So I had to add yet another replace to amek things work.

btcutil/p2a_address_test.go
Roasbeef: LOL

Simplified, using `require` now as well.

btcutil/address.go
Roasbeef: Done.

Roasbeef: Actually this won't work, as it causes an import cycle. We can't import `txscript` from `btcutil`. The other changed worked at it was in `btcutil_test`.

Roasbeef:
@kcalvinalvin PTAL

btcutil/address.go
```diff
@@ -709,3 +723,65 @@ func newAddressTaproot(hrp string,
...
+// EncodeAddress returns the bech32m string encoding of the pay-to-anchor
+// address. P2A addresses are encoded using witness version 1 with the program
+// bytes 0x4e73, resulting in these addresses per network:
+//
+//   - Mainnet: bc1pfeessrawgf
+//   - Testnet: tb1pfees9rn5nz
+//   - Regtest: bcrt1pfeesnyr2tx
+//   - Simnet:  sb1pfeesxv0pfa
+func (a *AddressPayToAnchor) EncodeAddress() string {
+	// For unknown networks, generate the address from the anchor data.
```
kcalvinalvin, line 779: Weird comment. Should remove this.

btcutil/p2a_address_test.go
```diff
@@ -0,0 +1,115 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			addr, err := btcutil.NewAddressPayToAnchor(tt.net)
+			require.NoError(t, err)
+
+			require.Equal(t, tt.wantAddress, addr.EncodeAddress())
+			require.Equal(t, tt.wantAddress, addr.String())
+			require.True(t, addr.IsForNet(tt.net))
+
+			// Verify it's not for a different network.
```
kcalvinalvin, line 49: The `require.True()` does cover this and we don't need this extra check but it's ok.

Claude could be better but heh.

txscript/standard.go
```diff
@@ -1011,6 +1051,13 @@ func ExtractPkScriptAddrs(pkScript []byte,
 		return NullDataTy, nil, 0, nil
 	}
 
+	// Check for pay-to-anchor script.
+	if IsPayToAnchorScript(pkScript) {
+		// P2A scripts don't have traditional addresses and require no
+		// signatures (anyone can spend them).
+		return PayToAnchorTy, nil, 0, nil
```
kcalvinalvin: We should be returning the address here because of this. https://github.com/btcsuite/btcd/pull/2433/changes/2eadd59bd9ff45da7e8a45e777d32b009849f640#r2390989822

txscript/p2a_test.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		t.Fatalf("ExtractPkScriptAddrs() error = %v", err)
+	}
+
+	if class != PayToAnchorTy {
+		t.Errorf("ExtractPkScriptAddrs() class = %v, want %v",
+			class, PayToAnchorTy)
+	}
+
+	if len(addrs) != 0 {
```
kcalvinalvin: We'd want 1 address since they P2A has an address

txscript/p2a_test.go
```diff
@@ -0,0 +1,147 @@
...
+	if class != PayToAnchorTy {
+		t.Errorf("GetScriptClass() = %v, want %v", class, PayToAnchorTy)
+	}
+
+	if class.String() != "anchor" {
+		t.Errorf("PayToAnchorTy.String() = %v, want 'anchor'",
+			class.String())
+	}
+
+	// Test that regular taproot scripts are still recognized as
```
kcalvinalvin, line 80: Weird thing to test for here but ok

mempool/policy_test.go
```diff
@@ -508,3 +659,118 @@ func TestCheckTransactionStandard(t *testing.T) {
...
+	return args.Get(0).([]byte)
+}
+
+// mockUtxoView mocks the utxoView interface using testify/mock.
+type mockUtxoView struct {
+	mock.Mock
+}
+
+// LookupEntry returns the entry for a given outpoint.
+func (m *mockUtxoView) LookupEntry(op wire.OutPoint) utxoEntry {
```
kcalvinalvin, line 687: Seems like these mock utxo entry and mock utxoview is only needed for the tests but it really doesn't need this abstraction. Not a blocker but yeah I think simpler is better.

```diff
diff --git a/mempool/policy.go b/mempool/policy.go
index 798d6686..46339d33 100644
--- a/mempool/policy.go
+++ b/mempool/policy.go
@@ -79,29 +79,6 @@ func calcMinRequiredTxRelayFee(serializedSize int64, minRelayTxFee btcutil.Amoun
        return minFee
 }
 
-// utxoEntry is an interface that provides access to UTXO data needed for
-// input standardness checks.
-type utxoEntry interface {
-       PkScript() []byte
-}
-
-// utxoView is an interface that provides access to UTXOs needed for
-// input standardness checks.
-type utxoView interface {
-       LookupEntry(wire.OutPoint) utxoEntry
-}      
-       
-// utxoViewpointAdapter wraps a blockchain.UtxoViewpoint to implement the
-// utxoView interface.
-type utxoViewpointAdapter struct {
-       view *blockchain.UtxoViewpoint
-}
-       
-// LookupEntry returns the entry for a given outpoint.
-func (u *utxoViewpointAdapter) LookupEntry(op wire.OutPoint) utxoEntry {
-       return u.view.LookupEntry(op)
-}      
-               
 // checkInputsStandard performs a series of checks on a transaction's inputs
 // to ensure they are "standard".  A standard transaction input within the
 // context of this function is one whose referenced public key script is of a
@@ -113,12 +90,6 @@ func (u *utxoViewpointAdapter) LookupEntry(op wire.OutPoint) utxoEntry {
 // accurately and concisely via the txscript.ScriptVerifyCleanStack and
 // txscript.ScriptVerifySigPushOnly flags.
 func checkInputsStandard(tx *btcutil.Tx, utxoView *blockchain.UtxoViewpoint) error {
-       return checkInputsStandardWithView(tx, &utxoViewpointAdapter{view: utxoView})
-}      
-
-// checkInputsStandardWithView performs input standardness checks using the
-// utxoView interface.
-func checkInputsStandardWithView(tx *btcutil.Tx, utxoView utxoView) error {
        // NOTE: The reference implementation also does a coinbase check here,
        // but coinbases have already been rejected prior to calling this
        // function so no need to recheck.
diff --git a/mempool/policy_test.go b/mempool/policy_test.go
index c3259fd0..3411fa08 100644
--- a/mempool/policy_test.go
+++ b/mempool/policy_test.go
@@ -9,13 +9,13 @@ import (
        "testing"
        "time"  
 
+       "github.com/btcsuite/btcd/blockchain"
        "github.com/btcsuite/btcd/btcec/v2"
        "github.com/btcsuite/btcd/btcutil"
        "github.com/btcsuite/btcd/chaincfg"
        "github.com/btcsuite/btcd/chaincfg/chainhash"
        "github.com/btcsuite/btcd/txscript"
        "github.com/btcsuite/btcd/wire"
-       "github.com/stretchr/testify/mock"
 )
                                        
 // TestCalcMinRequiredTxRelayFee tests the calcMinRequiredTxRelayFee API.
@@ -660,44 +660,28 @@ func TestCheckTransactionStandard(t *testing.T) {
        }
 }                      
                        
-// mockUtxoEntry mocks the utxoEntry interface using testify/mock.
-type mockUtxoEntry struct {
-       mock.Mock
-}
-
-// PkScript returns the public key script.
-func (m *mockUtxoEntry) PkScript() []byte {
-       args := m.Called()
-       return args.Get(0).([]byte)
-}
-
-// mockUtxoView mocks the utxoView interface using testify/mock.
-type mockUtxoView struct {
-       mock.Mock
-}
-
-// LookupEntry returns the entry for a given outpoint.
-func (m *mockUtxoView) LookupEntry(op wire.OutPoint) utxoEntry {
-       args := m.Called(op)
-       if args.Get(0) == nil {
-               return nil
-       }
-       return args.Get(0).(utxoEntry)
-}
-
 // TestP2ASpendingStandardness tests that P2A outputs require empty witness
 // and empty signature script to be considered standard.
 func TestP2ASpendingStandardness(t *testing.T) {
        // Create a previous transaction with a P2A output.
-       prevTxHash, _ := chainhash.NewHashFromStr("0101010101010101010101010101010101010101010101010101010101010101")
-       prevOut := wire.OutPoint{Hash: *prevTxHash, Index: 0}
+       prevTx := wire.NewMsgTx(2)
+       prevTx.AddTxIn(&wire.TxIn{
+               PreviousOutPoint: wire.OutPoint{
+                       Hash:  chainhash.Hash{},
+                       Index: 0,
+               },
+       })
+       prevTx.AddTxOut(&wire.TxOut{
+               Value:    1000,
+               PkScript: txscript.PayToAnchorScript,
+       })
+       prevBtcTx := btcutil.NewTx(prevTx)
 
-       // Create mocked UTXO entry and view.
-       mockEntry := new(mockUtxoEntry)
-       mockEntry.On("PkScript").Return(txscript.PayToAnchorScript)
+       // Create a UTXO view with the P2A output.
+       utxoView := blockchain.NewUtxoViewpoint()
+       utxoView.AddTxOuts(prevBtcTx, 100)
 
-       mockView := new(mockUtxoView)
-       mockView.On("LookupEntry", prevOut).Return(mockEntry)
+       prevOut := wire.OutPoint{Hash: *prevBtcTx.Hash(), Index: 0}
 
        tests := []struct {
                name       string
@@ -756,7 +740,7 @@ func TestP2ASpendingStandardness(t *testing.T) {
                        })
 
                        btcTx := btcutil.NewTx(tx)
-                       err := checkInputsStandardWithView(btcTx, mockView)
+                       err := checkInputsStandard(btcTx, utxoView)
 
                        if test.shouldFail {
                                if err == nil {
@@ -767,10 +751,6 @@ func TestP2ASpendingStandardness(t *testing.T) {
                                        t.Errorf("Unexpected error for valid P2A spend: %v", err)
                                }
                        }
-
-                       // Verify that the mock was called.
-                       mockView.AssertExpectations(t)
-                       mockEntry.AssertExpectations(t)
                })
        }
 }
```

kcalvinalvin:
Most things look good. But `ExtractPkScriptAddrs` not returning an address for P2A should be fixed before this gets merged.

Things done:
1. Went through each of the commits again to sanity check the changes.
2. Checked that the implementation here aligns with BIP-433.
3. Checked that the tests make sense in each of the commits.

go.mod
yyforyongyu, line 70: This local replace makes the PR build in this checkout, but downstream modules will not inherit it. Since `txscript` now type-switches on `btcutil.AddressPayToAnchor`, `go get github.com/btcsuite/btcd@this-commit` will resolve `btcutil v1.1.5` and fail to compile because that type does not exist there. We need a real `btcutil` version or pseudo-version with the new address type, or avoid the cross-module API dependency before merge.

mempool/policy.go
```diff
@@ -260,6 +316,13 @@ func IsDust(txOut *wire.TxOut, minRelayTxFee btcutil.Amount) bool {
 		return true
 	}
 
+	// P2A outputs have a fixed dust threshold of 240 satoshis regardless
+	// of the relay fee. This optimization avoids the more complex
+	// calculation below.
+	if txscript.IsPayToAnchorScript(txOut.PkScript) {
+		return txOut.Value < 240
```
yyforyongyu: This hard-codes P2A dust at 240 sats regardless of `minRelayTxFee`. That matches the default P2A dust threshold because the existing witness dust formula already evaluates to 240 sats at the default rate, but it changes configured relay policy: with a zero relay fee, 239-sat P2A is still dust, and with a high relay fee, 240-sat P2A is still non-dust. Bitcoin Core still derives the threshold from the supplied dust relay feerate, so I think this should use the existing fee-sensitive dust calculation rather than special-casing `IsDust`.

btcutil/address.go
```diff
@@ -709,3 +723,65 @@ func newAddressTaproot(hrp string,
...
+	if err != nil {
+		return ""
+	}
+	return addr
+}
+
+// ScriptAddress returns the raw bytes of the P2A script to be used when
+// inserting the address into a txout's script. For P2A, this returns the fixed
+// script bytes: OP_1 <0x4e73>.
+func (a *AddressPayToAnchor) ScriptAddress() []byte {
```
yyforyongyu: `AddressPayToAnchor.ScriptAddress` returns the full scriptPubKey, unlike the other segwit address types where `ScriptAddress` returns the witness program and `PayToAddrScript` wraps it. That exception makes generic address code easy to misuse. Can we either return just `0x4e73` and route output creation through `PayToAddrScript`, or document the exception clearly and add coverage for the generic `PayToAddrScript` path?

txscript/engine.go
```diff
@@ -760,11 +767,22 @@ func (vm *Engine) verifyWitnessProgram(witness wire.TxWitness) error {
 			vm.SetStack(witness[:len(witness)-2])
 		}
 
+	// Pay-to-anchor (P2A) outputs are special anyone-can-spend outputs.
+	// They only work as native witness programs, not wrapped in P2SH.
+	case IsPayToAnchorWitnessProgram(
+		vm.witnessVersion, vm.witnessProgram,
+	) && !vm.bip16:
+		// P2A spending always succeeds regardless of witness content.
+		vm.dstack.stk = vm.dstack.stk[:1]
```
yyforyongyu: This direct slice truncation relies on the current engine having already executed the native witness program opcodes, leaving `OP_1` as the first stack item. It would be more robust and clearer to set the intended success stack explicitly with `vm.SetStack([][]byte{{0x01}})`, avoiding a possible panic or silent false result if witness execution flow changes later.

go.mod
Roasbeef: Yeah we'll fix this before the release.

mempool/policy.go
Roasbeef: Added a comment, this is from the BIP: https://github.com/bitcoin/bips/blob/master/bip-0433.mediawiki#:~:text=As%20an%20implementation%20detail%2C%20dust%20threshold%20for%20P2A%20is%20set%20at%20240%20satoshis%20by%20default

btcutil/address.go
Roasbeef, line 793: Yeah modified to just return the push.

txscript/standard.go
```diff
@@ -50,6 +51,10 @@ const (
 		ScriptVerifyConstScriptCode
 )
 
+// PayToAnchorScript is the fixed script bytes for a pay-to-anchor output.
+// This is OP_1 OP_DATA_2 0x4e73 (witness v1 with program 0x4e73).
+var PayToAnchorScript = []byte{OP_1, OP_DATA_2, 0x4e, 0x73}
```
Roasbeef, line 61: Yeah have this avoid an import cycle. Added the comment.

integration/p2a_test.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	// Deduct a small fee from the P2A output value.
+	spendOut := wire.NewTxOut(int64(amount-100), spendScript)
+	spendP2ATx.AddTxOut(spendOut)
+
+	// Broadcast the spend transaction to verify network acceptance. P2A
+	// outputs are witness programs and are validated through the normal
+	// transaction validation path in the mempool and consensus.
+	spendTxHash, err := harness.Client.SendRawTransaction(spendP2ATx, true)
```
Roasbeef, line 106: Added the extra cases.

mempool/policy.go
yyforyongyu: Then what if people decide to increase/decrease it? Also just realized in `btcd` we are missing the config `-dustrelayfee`, similar to what bitcoind has https://github.com/bitcoin/bitcoin/blob/859215218667ca9f35d5adae0289e4a125798087/src/policy/policy.h#L63-L68

btcutil/address.go
```diff
@@ -171,8 +171,22 @@ func DecodeAddress(addr string, defaultNet *chaincfg.Params) (Address, error) {
...
 
 			switch len(witnessProg) {
+			case 2:
+				// Check if it's a P2A address (witness version
+				// 1, program 0x4e73).
+				if witnessVer == 1 && bytes.Equal(
+					witnessProg, []byte{0x4e, 0x73},
+				) {
+					return &AddressPayToAnchor{
+						hrp: hrp,
```
yyforyongyu: This stores the HRP directly from the input prefix, unlike the other segwit address constructors that normalize decoded HRPs to lowercase. A valid all-uppercase P2A address such as `BC1PFEESSRAWGF` will decode, and `EncodeAddress()` will re-emit lowercase, but `IsForNet(&chaincfg.MainNetParams)` returns false because the stored HRP is `BC` instead of `bc`.

Can we normalize the HRP here, or use the lowercase HRP returned by bech32 decoding, and add uppercase decode coverage?

integration/p2a_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/integration/rpctest"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+)
+
+// TestPayToAnchorSimple tests creating and spending P2A outputs.
+func TestPayToAnchorSimple(t *testing.T) {
```
yyforyongyu, line 17: this test failed tho,
```
GOWORK=off go test -tags rpctest ./integration -run TestPayToAnchorSimple -count=1
```

Got,
```
2026-05-11 14:12:53.606 [INF] ITEST: Established connection to RPC server 127.0.0.1:8335
2026-05-11 14:12:53.913 [INF] ITEST: Established connection to RPC server 127.0.0.1:8335
2026-05-11 14:12:54.033 [INF] ITEST: Established connection to RPC server 127.0.0.1:8337
2026-05-11 14:12:54.188 [INF] ITEST: Established connection to RPC server 127.0.0.1:8337
2026-05-11 14:12:54.304 [DBG] ITEST: Detected btcd version: 250000
--- FAIL: TestPayToAnchorSimple (0.33s)
    p2a_test.go:134: Successfully created P2A output in tx 9ca4cff5dc5f850f1b767247c7a70832085673908a50b3aba0465622e4c8113c and spent it in tx f4abe270a4fe372d7ee4bcf1b108f2c1c33f95bb19711e9071d8065d4a0cea9c
    --- FAIL: TestPayToAnchorSimple/sub-dust_output_rejected (0.00s)
        p2a_test.go:190: sub-dust P2A output was accepted; expected mempool dust rejection
FAIL
FAIL    github.com/btcsuite/btcd/integration    1.376s
FAIL
```

I think the simnet accepts non-standard txs by default. It should start the harness with `--rejectnonstd` or adjust the test.

mempool/policy.go
Roasbeef: Hmm, yeah I see what you're getting at. Will remove this to instead just have it calculate it directly removing this cut out.

Roasbeef: Remove this hard coding in the latest version. Also reached out to instagibbs to see if we should update the BIP.

btcutil/address.go
Roasbeef: Yeah moved to normalize

integration/p2a_test.go
Roasbeef: Yep, added that on

yyforyongyu (approved):
LGTM🚢

PR #2441, opened 2025-10-04
txscript: increase OP_RETURN data limit to 100KB

mikekelly:
## Change Description

This PR implements Bitcoin Core v30 OP_RETURN policy changes by:
1. Increasing `MaxDataCarrierSize` from 80 bytes to 100,000 bytes
2. Removing the restriction limiting transactions to a single OP_RETURN output

These changes align btcd with Bitcoin Core v30, which removed both the 80-byte data limit and the single OP_RETURN restriction.

### Motivation

Bitcoin Core v30 updated OP_RETURN policies by removing the specific `-datacarriersize` default of 80 bytes (replacing it with 100,000 bytes) and eliminating the single OP_RETURN per transaction restriction. These changes bring btcd into alignment with Bitcoin Core v30 for several critical reasons:

**1. Policy Alignment**  
Maintains consistency with Bitcoin Core v30, ensuring btcd nodes accept the same set of standard transactions as the Bitcoin reference implementation. Both the data size limit and multiple OP_RETURN restrictions were arbitrary policy limitations with no consensus basis.

**2. Accurate Fee Estimation**  
Filtering large OP_RETURN transactions or transactions with multiple OP_RETURNs from the mempool hides transactions competing for blockspace, leading to inaccurate fee estimation. Full mempool visibility is essential for informed fee rate calculation.

**3. Validation Performance & Block Propagation**  
Rejecting transactions that will be mined by other nodes means they must be validated "cold" when processing blocks, without cached results. This causes slower block validation, delayed block propagation, and increased orphan risk for miners.

**4. Network Decentralization**  
Slow block propagation disproportionately advantages large mining operations with preferentially peered nodes, giving them a head start on the next block. Pre-validating all likely-to-be-mined transactions helps maintain network decentralization.

#### Data Carrier Size Increase
- **txscript/standard.go**: Increased `MaxDataCarrierSize` from 80 to 100,000 bytes
- **txscript/standard.go**: Modified `NullDataScript()` to use `AddFullData()` to bypass `MaxScriptElementSize` (520 bytes) since OP_RETURN outputs are never executed
- **txscript/scriptbuilder.go**: Updated `AddFullData()` documentation to reflect its legitimate use for large OP_RETURN outputs
- **txscript/standard_test.go**: Updated tests for new size limits (81 bytes, 9,995 bytes, 100,001 bytes)

#### Multiple OP_RETURN Support
- **mempool/policy.go**: Removed the check restricting transactions to one OP_RETURN output
- **mempool/policy_test.go**: Added tests for multiple OP_RETURNs with varying sizes
  - Two small OP_RETURNs
  - Three 25KB OP_RETURNs (~75KB total)
  - Two 37KB OP_RETURNs + payment output
  - Three 60KB OP_RETURNs (180KB total - correctly rejected by transaction weight)

#### Integration Tests
- **integration/rawtx_test.go**: Comprehensive end-to-end tests
  - 75KB single OP_RETURN: Accepted
  - 100KB single OP_RETURN: Rejected (exceeds transaction weight limit)
  - 101KB single OP_RETURN: Rejected (exceeds MaxDataCarrierSize)
  - Multiple OP_RETURNs (3 × 25KB): Accepted into mempool

### Technical Notes

**MaxDataCarrierSize**  
Increased to 100,000 bytes, effectively removing the specific data carrier size limit. The existing transaction size limit (maxStandardTxWeight of 400,000 weight units ≈ 100KB total transaction) will apply before the data carrier limit is reached due to transaction overhead (inputs, outputs, signatures, etc.).

**Multiple OP_RETURNs**  
Each OP_RETURN output is independently validated against MaxDataCarrierSize (100KB). The transaction as a whole remains subject to maxStandardTxWeight (400,000 weight units ≈ 100KB total transaction).

**Script Limits**  
OP_RETURN outputs are exempt from `MaxScriptElementSize` (520 bytes) execution limits because they are provably unspendable and never executed by the script engine.

### Compatibility

- ✅ Backward compatible: All existing OP_RETURN transactions remain valid
- ✅ Forward compatible: Aligns with Bitcoin Core v30 behavior  
- ✅ No consensus changes: These are policy-only changes

### 1. Run Unit Tests
```bash
go test ./txscript/... -v
```
Expected: All tests pass, including new tests for:
- 81-byte OP_RETURN (now standard, previously rejected)
- 9,995-byte OP_RETURN (under MaxScriptSize)
- 100,001-byte OP_RETURN (exceeds MaxDataCarrierSize, rejected)

### 2. Run Policy Tests
```bash
go test ./mempool/... -v -run TestCheckTransactionStandard
```
Expected: All tests pass, including:
- 75KB single OP_RETURN (accepted)
- Multiple OP_RETURNs totaling ~75KB (accepted)
- Multiple OP_RETURNs totaling 180KB (rejected by transaction weight)

### 3. Run Integration Tests
```bash
go test -tags=rpctest ./integration/... -v -run TestLargeOPReturnMempool
```
Expected: All test cases pass:
- 75KB OP_RETURN: Accepted into mempool (weight: 300,840 < 400,000 limit)
- 100KB OP_RETURN: Rejected (weight: 400,840 > 400,000 limit)
- 101KB OP_RETURN: Rejected (exceeds MaxDataCarrierSize)
- Multiple OP_RETURNs (3 × 25KB): Accepted into mempool

### 4. Run Full Test Suite
```bash
go test ./...
```
Expected: All tests pass across the entire codebase.

### 5. Code Quality Checks
```bash
go fmt ./...
go vet ./...
```
Expected: No issues found.

### Testing

### DIFF
--- a/integration/rawtx_test.go
+++ b/integration/rawtx_test.go
@@ -201,3 +201,88 @@ func decodeHex(t *testing.T, txHex string) *wire.MsgTx {
 
 	return tx.MsgTx()
 }
+
+// TestLargeOPReturnMempool tests that large OP_RETURN transactions (up to 100KB)
+// are accepted into the mempool. This validates the Bitcoin Core v30 change that
+// removes the 80-byte OP_RETURN limit.
+func TestLargeOPReturnMempool(t *testing.T) {
+	t.Parallel()
+
+	// Create a single node for testing mempool acceptance.
+	// Use --rejectnonstd to enforce standardness rules (including maxStandardTxWeight).
+	btcdCfg := []string{"--rejectnonstd", "--debuglevel=debug"}
+	node, err := rpctest.New(&chaincfg.SimNetParams, nil, btcdCfg, "")
+	require.NoError(t, err)
+
+	// Setup the node with mature coinbase outputs.
+	require.NoError(t, node.SetUp(true, 100))
+	t.Cleanup(func() {
+		require.NoError(t, node.TearDown())
+	})
+
+	testCases := []struct {
+		name          string
+		dataSize      int
+		shouldSucceed bool
+		failureReason string
+	}{
+		{
+			name:          "75KB OP_RETURN (well over old 80 byte/520 byte limits)",
+			dataSize:      75000,
+			shouldSucceed: true,
+		},
+		{
+			name:          "100KB OP_RETURN (at MaxDataCarrierSize limit)",
+			dataSize:      100000,
+			shouldSucceed: false,
+			failureReason: "transaction size exceeds maxStandardTxWeight (100,000 vbytes)",
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			// Create large OP_RETURN script.
+			largeData := make([]byte, tc.dataSize)
+			for i := range largeData {
+				largeData[i] = byte(i % 256)
+			}
+
+			opReturnScript, err := txscript.NullDataScript(largeData)
+			require.NoError(t, err)
+
+			// Create a transaction with OP_RETURN output using the harness wallet.
+			// The wallet will handle signing and input selection.
+			opReturnOutput := &wire.TxOut{
+				Value:    0,
+				PkScript: opReturnScript,
+			}
+
+			// Create transaction with the OP_RETURN output.
+			// The wallet automatically adds inputs and a change output if needed.
+			tx, err := node.CreateTransaction([]*wire.TxOut{opReturnOutput}, 10, true)
+			require.NoError(t, err)
+
+			// Log the actual transaction size.
+			txSize := tx.SerializeSize()
+			txWeight := txSize * 4 // Non-segwit: weight = size * 4
+			t.Logf("Transaction size: %d bytes, weight: %d (limit: 400000)", txSize, txWeight)
+
+			// Submit to the mempool.
+			txHash, err := node.Client.SendRawTransaction(tx, true)
+
+			if tc.shouldSucceed {
+				require.NoError(t, err, "Large OP_RETURN tx should be accepted")
+				t.Logf("✓ Large OP_RETURN tx (%d bytes data) accepted: %s", tc.dataSize, txHash)
+
+				// Verify it's in the mempool.
+				mempool, err := node.Client.GetRawMempool()
+				require.NoError(t, err)
+				require.Contains(t, mempool, txHash, "Transaction should be in mempool")
+			} else {
+				require.Error(t, err, "Transaction should be rejected: %s", tc.failureReason)
+				t.Logf("✓ Transaction correctly rejected (%s): %v", tc.failureReason, err)
+			}
+		})
+	}
+}
--- a/integration/rawtx_test.go
+++ b/integration/rawtx_test.go
@@ -221,10 +221,11 @@ func TestLargeOPReturnMempool(t *testing.T) {
 	})
 
 	testCases := []struct {
-		name          string
-		dataSize      int
-		shouldSucceed bool
-		failureReason string
+		name                string
+		dataSize            int
+		shouldSucceed       bool
+		failureReason       string
+		failsScriptCreation bool
 	}{
 		{
 			name:          "75KB OP_RETURN (well over old 80 byte/520 byte limits)",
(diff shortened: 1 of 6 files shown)


integration/rawtx_test.go
```diff
@@ -201,3 +201,88 @@ func decodeHex(t *testing.T, txHex string) *wire.MsgTx {
 
 	return tx.MsgTx()
 }
+
+// TestLargeOPReturnMempool tests that large OP_RETURN transactions (up to 100KB)
+// are accepted into the mempool. This validates the Bitcoin Core v30 change that
+// removes the 80-byte OP_RETURN limit.
+func TestLargeOPReturnMempool(t *testing.T) {
```
murchandamus, line 208: Bitcoin Core 30.0 will also allow multiple OP_RETURN outputs on transactions. I’m not sure whether BTCD ever restricted transactions to a single one, but another test that adds two small OP_RETURN outputs would demonstrate that BTCD is compatible, or reveal that additional changes would need to be made to make it compatible.

mikekelly: @murchandamus It's enforced atm see https://github.com/btcsuite/btcd/blob/b7d070601def0f94e3713879b19f1678949a8d68/mempool/policy.go#L368-L373

Looking into it now

mikekelly: @murchandamus done

saubyk:
How's this pr different from https://github.com/btcsuite/btcd/pull/2385 which was closed after [this comment](https://github.com/btcsuite/btcd/pull/2385#issuecomment-2956481722) from @Roasbeef ?

mikekelly:
@saubyk it's more thorough implementation and tests, and the PR does a better job of explaining why this is a good idea for users

saubyk:
Regardless of a better rationale I believe Laolu's comment is very clear on why the changes proposed here are not acceptable for this project. And that status quo has not been changed so far.

"_The only relay policies that we endeavor to keep up with are those that require interop, and tangibly affect higher layer solutions such as: RBF, TRUC, etc. These policies all have BIPs, with a clear target scope and motivation._"

mikekelly:
@saubyk  I read it - my point was that motives 2 3 and 4 all "tangibly affect higher layer solutions". iow they're good for users

mikekelly:
@saubyk Genuine question; what would btcd users gain from _not_ incorporating this change given that it's proposed with additional testing to mitigate regression? How does that weigh up against the benefits listed in the motivation section?

mikekelly:
If the goal of mempool policies and standardness is to make nodes as efficient as possible, whilst not filtering out likely-to-get-mined transactions (because it harms fee estimation and block validation), then this PR should be attractive.

If that's not the goal of mempool policies and standardness in btcd, then what is the goal of them?

saubyk:
> If the goal of mempool policies and standardness is to make nodes as efficient as possible, whilst not filtering out likely-to-get-mined transactions (because it harms fee estimation and block validation), then this PR should be attractive.
> 
> If that's not the goal of mempool policies and standardness in btcd, then what is the goal of them?

Let me point out again to the [same comment](https://github.com/btcsuite/btcd/pull/2385#issuecomment-2956481722) from @Roasbeef ...

"_We already have a flag that enables users to [accept and relay all non-standard transactions](https://github.com/btcsuite/btcd/blob/ac58f6984cec305069c399d2e4e27ae742078aaa/sample-btcd.conf#L272-L276). Users that want to shed themselves of all discretionary relay policy can run with those flags._"

It's unfair for us to reject one pr from a contributor, while accepting the same change in a different pr just because it has better tests. There is no fundamental change in status quo after the previous pr has been rejected.

mikekelly:
A flag that enables users to accept and relay all non-standard transactions is not a reasonable alternative; that forces btcd users to pick either the status quo of over-filtering (which harms fee estimation and block validation) or a very inefficient node which is vulnerable to DoS and continually wastes resources processing and relaying transactions that even miners want nothing to do with.

There was no discussion on the other thread, the author simply accepted the response and closed it. I'm asking, constructively, for more clarification on the basis for rejecting such a change.

Having additional tests is a material difference.

I'm getting the impression that you're reaching for reasons to reject this out of hand rather than discuss the merits. If that is the case, then this probably isn't the PR thread for you and I'll wait for someone else who is willing to discuss it properly.

mikekelly:
thanks @kmk142789

allocz (commented):
Some questions:

1. For what use cases we need more than 80 bytes of OP_RETURN?
2. Since we are already bounded by the max block size, why 100KB of OP_RETURN limit, why not just remove the size limit?

mikekelly:
@allocz thanks for this 

1. The goal isn't enabling use cases, it's bringing the configuration in line with the relay policies of Bitcoin's reference implementation. 

2. This was covered in the PR description:


> MaxDataCarrierSize
> Increased to 100,000 bytes, effectively removing the specific data carrier size limit. The existing transaction size limit (maxStandardTxWeight of 400,000 weight units ≈ 100KB total transaction) will apply before the data carrier limit is reached due to transaction overhead (inputs, outputs, signatures, etc.).

PR #2443, opened 2025-10-07
musig2: add WithExternalCombinedNonce option to Sign

sputn1ck:
This PR adds a new functional option to the Session.Sign function which allows specifying an external combined Nonce. This is useful when a central coordinator aggregates all nonces.

Externally Aggregated Nonces are part of the spec:
```
Third-party nonce and partial signature aggregation: Instead of every signer sending their nonce and partial signature to every 
other signer, it is possible to use an untrusted third-party aggregator in order to reduce the communication complexity from 
quadratic to linear in the number of signers. In each of the two rounds, the aggregator collects all signers' contributions 
(nonces or partial signatures), aggregates them, and broadcasts the aggregate back to the signers. A malicious aggregator can 
force the signing session to fail to produce a valid Schnorr signature but cannot negatively affect the unforgeability of the
 scheme.
```
https://github.com/bitcoin/bips/blob/master/bip-0327.mediawiki#user-content-Nonce_Aggregation

### DIFF
--- a/btcec/schnorr/musig2/context.go
+++ b/btcec/schnorr/musig2/context.go
@@ -553,6 +553,20 @@ func (s *Session) RegisterPubNonce(nonce [PubNonceSize]byte) (bool, error) {
 	return haveAllNonces, nil
 }
 
+// CombinedNonce returns the combined public nonce for the signing session.
+// This will be available after either:
+//   - All individual nonces have been registered via RegisterPubNonce, or
+//   - A combined nonce has been registered via RegisterCombinedNonce
+//
+// If the combined nonce is not yet available, this method returns an error.
+func (s *Session) CombinedNonce() ([PubNonceSize]byte, error) {
+	if s.combinedNonce == nil {
+		return [PubNonceSize]byte{}, ErrCombinedNonceUnavailable
+	}
+
+	return *s.combinedNonce, nil
+}
+
 // RegisterCombinedNonce allows a caller to directly register a combined nonce
 // that was generated externally. This is useful in coordinator-based
 // protocols where the coordinator aggregates all nonces and distributes the
--- a/btcec/schnorr/musig2/musig2_test.go
+++ b/btcec/schnorr/musig2/musig2_test.go
@@ -746,6 +746,86 @@ func TestSigningWithAggregatedNonce(t *testing.T) {
 			t.Fatalf("final signature is invalid")
 		}
 	})
+
+	t.Run("get combined nonce after RegisterCombinedNonce", func(t *testing.T) {
+		privKey, _ := btcec.NewPrivateKey()
+		privKey2, _ := btcec.NewPrivateKey()
+		signSet := []*btcec.PublicKey{privKey.PubKey(), privKey2.PubKey()}
+
+		signCtx, _ := NewContext(privKey, false, WithKnownSigners(signSet))
+		session, _ := signCtx.NewSession()
+
+		// Should fail before registering combined nonce.
+		_, err := session.CombinedNonce()
+		if err != ErrCombinedNonceUnavailable {
+			t.Fatalf("expected ErrCombinedNonceUnavailable, got: %v", err)
+		}
+
+		// Register combined nonce.
+		expectedNonce := getValidNonce(t)
+		err = session.RegisterCombinedNonce(expectedNonce)
+		if err != nil {
+			t.Fatalf("RegisterCombinedNonce failed: %v", err)
+		}
+
+		// Should succeed after registering.
+		gotNonce, err := session.CombinedNonce()
+		if err != nil {
+			t.Fatalf("CombinedNonce failed: %v", err)
+		}
+
+		if gotNonce != expectedNonce {
+			t.Fatalf("expected nonce %x, got %x", expectedNonce, gotNonce)
+		}
+	})
+
+	t.Run("get combined nonce after RegisterPubNonce", func(t *testing.T) {
+		const numSigners = 3
+
+		signerKeys := make([]*btcec.PrivateKey, numSigners)
+		signSet := make([]*btcec.PublicKey, numSigners)
+		for i := 0; i < numSigners; i++ {
+			privKey, _ := btcec.NewPrivateKey()
+			signerKeys[i] = privKey
+			signSet[i] = privKey.PubKey()
+		}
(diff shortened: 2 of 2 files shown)


btcec/schnorr/musig2/context.go
```diff
@@ -554,15 +554,31 @@ func (s *Session) RegisterPubNonce(nonce [PubNonceSize]byte) (bool, error) {
...
 
-	switch {
 	// If no local nonce is present, then this means we already signed, so
 	// we'll return an error to prevent nonce re-use.
-	case s.localNonces == nil:
+	if s.localNonces == nil {
 		return nil, ErrSigningContextReuse
+	}
+
+	opts := defaultSignOptions()
```
Roasbeef: One alternative would be to add a method to the session to add the combined nonce. 

Then we can leave this error path as is (failed if no combined nonce is present).

sputn1ck: Pushed the changes

bhandras (approved):
LGTM, thank you! 🎉

btcec/schnorr/musig2/musig2_test.go
```diff
@@ -439,3 +439,83 @@ func (mr *memsetRandReader) Read(buf []byte) (n int, err error) {
 	}
 	return len(buf), nil
 }
+
+// TestSigningWithAggregatedNonce tests the aggregated nonce signing flow where
+// a coordinator aggregates nonces and distributes the combined nonce to
+// participants, rather than each participant aggregating nonces themselves.
+func TestSigningWithAggregatedNonce(t *testing.T) {
```
Roasbeef, line 447: We should also add tests for the input validation edge cases:
  * Call to `RegisterCombinedNonce` fails if we have a combined nonce, or already partially registered some of the 3rd party nonces.

Roasbeef (changes requested):

btcec/schnorr/musig2/musig2_test.go
sputn1ck: added test cases for all errors paths. Also added a check in `RegisterPubNonce` to fail if we already have a combined nonce

btcec/schnorr/musig2/context.go
```diff
@@ -548,6 +548,20 @@ func (s *Session) RegisterPubNonce(nonce [PubNonceSize]byte) (bool, error) {
 	return haveAllNonces, nil
 }
 
+// CombinedNonce returns the combined public nonce for the signing session.
+// This will be available after either:
+//  - All individual nonces have been registered via RegisterPubNonce, or
+//  - A combined nonce has been registered via RegisterCombinedNonce
+//
+// If the combined nonce is not yet available, this method returns an error.
+func (s *Session) CombinedNonce() ([PubNonceSize]byte, error) {
```
sputn1ck, line 562: Not sure if this approach is good(allows for more missuse of the nonces i guess) or if we should just return the combined nonce _one time only_ if we register the last pubnonce? (changing RegisterPubNonce sig to return the aggregated nonce, instead of a havAllNonces bool)

btcec/schnorr/musig2/context.go
```diff
@@ -548,6 +548,33 @@ func (s *Session) RegisterPubNonce(nonce [PubNonceSize]byte) (bool, error) {
 	return haveAllNonces, nil
 }
 
+// RegisterCombinedNonce allows a caller to directly register a combined nonce
+// that was generated externally. This is useful in coordinator-based
+// protocols where the coordinator aggregates all nonces and distributes the
+// combined nonce to participants, rather than each participant aggregating
+// nonces themselves.
+func (s *Session) RegisterCombinedNonce(
+	combinedNonce [PubNonceSize]byte) error {
```
Roasbeef, line 576: To project against invalid user input, let's also attempt to parse the combined nonce into the two expected compressed points here.

Roasbeef: We can then add a test case where we pass in an obviously invalid nonce.

btcec/schnorr/musig2/context.go
```diff
@@ -548,6 +548,33 @@ func (s *Session) RegisterPubNonce(nonce [PubNonceSize]byte) (bool, error) {
...
+	// If we already have a combined nonce, then this method was called too
+	// many times.
+	if s.combinedNonce != nil {
+		return ErrAlredyHaveAllNonces
+	}
+
+	// We also don't allow this method to be called if we already registered
+	// some public nonces.
+	if len(s.pubNonces) != 1 {
+		return fmt.Errorf("cannot register combined nonce after " +
```
Roasbeef: Let's make this a pre-defined error, then we can use `errors.Is` within the test cases.

btcec/schnorr/musig2/context.go
```diff
@@ -548,6 +548,33 @@ func (s *Session) RegisterPubNonce(nonce [PubNonceSize]byte) (bool, error) {
...
+
+	// If we already have a combined nonce, then this method was called too
+	// many times.
+	if s.combinedNonce != nil {
+		return ErrAlredyHaveAllNonces
+	}
+
+	// We also don't allow this method to be called if we already registered
+	// some public nonces.
+	if len(s.pubNonces) != 1 {
```
Roasbeef: Perhaps this should be `len(s.pubNonces) > 1`? Or I guess we want to ensure that the user has already generated their own nonce, before trying to register the final combined nonces.

btcec/schnorr/musig2/context.go
sputn1ck: done

btcec/schnorr/musig2/context.go
sputn1ck: done

btcec/schnorr/musig2/context.go
sputn1ck: changed it to `len(s.pubNonces) > 1`. I think the context _always_ generates (or expects the user to pass) a nonce, so `!=1` would have been fine, but now it's more readable

btcec/schnorr/musig2/context.go
Roasbeef: Each time we sign, we blank out `localNonces`, which prevents us from signing again directly after unless a new set of local nonces are generated.

PR #2450, opened 2025-10-29
rpcclient: ensure http dial respects timeout

wydengyre:
## Change Description
Ensure that the dial phase of rpcclient's HTTP posts respects the http.Client's Timeout. It was instead falling back to the OS settings.

## Steps to Test
This is a bit challenging to test, as it requires simulating a very laggy dial. Not sure the juice is worth the squeeze.

### Testing
- [x] Your PR passes all CI checks.
- [ ] Tests covering the positive and negative (error paths) are included.
- [ ] Bug fixes contain tests triggering the bug to prevent regressions.

### Code Style and Documentation
- [x] The change is not [insubstantial](https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md#substantial-contributions-only). Typo fixes are not accepted to fight bot spam.
- [x] The change obeys the [Code Documentation and Commenting](https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md#code-documentation-and-commenting) guidelines, and lines wrap at 80.
- [x] Commits follow the [Ideal Git Commit Structure](https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md#model-git-commit-messages). 
- [x] Any new logging statements use an appropriate subsystem and logging level.

📝 Please see our [Contribution Guidelines](https://github.com/btcsuite/btcd/blob/master/docs/code_contribution_guidelines.md) for further guidance.

Issue #1323, opened 2018-10-11
rpcclient.Client doesn't reuse network connections


weirdgiraffe:
Establishing a connection is one of the most long running operation in HTTP and enabling connection reuse will increase the speed drastically. I propose to remove this line of code:

https://github.com/btcsuite/btcd/blob/2a560b2036bee5e3679ec2133eb6520b2f195213/rpcclient/infrastructure.go#L819

I've tracked this line up to initial commit and haven't found any explanation why Client needs to close the connection explicitly, but performance gain is huge. Here is an example test:

```
package main

import (
	"log"
	"testing"

	"github.com/btcsuite/btcd/rpcclient"
)

func TestSpeed(t *testing.T) {
	// Connect to local bitcoin core RPC server using HTTP POST mode.
	connCfg := &rpcclient.ConnConfig{
		Host:         "localhost:8332",
		User:         "username",
		Pass:         "password",
		HTTPPostMode: true,    // Bitcoin core only supports HTTP POST mode
		DisableTLS:   true,    // Bitcoin core does not provide TLS by default
	}

	client, err := rpcclient.New(connCfg, nil)
	if err != nil {
		log.Fatal(err)
	}
	defer client.Shutdown()

	// Get the current block count.
	blockCount, err := client.GetBlockCount()
	if err != nil {
		log.Fatal(err)
	}
	log.Printf("Block count: %d", blockCount)

	for i := blockCount - 100; i < blockCount; i++ {
		hash, err := client.GetBlockHash(i)
		if err != nil {
			log.Fatal(err)
		}
		log.Printf("Block hash: %v", hash)
	}
}
```
This example test runs on my machine with the following numbers:

with `Close == false`:
```
❯ go test example_test.go
ok      command-line-arguments  3.204s
```

with `Close == true`:
```
❯ go test example_test.go
ok      command-line-arguments  10.252s
```

totaloutput:
I think for infrastructure layer of code, it's safer to close and reconnect for each command. For the specific usecase like you showed above to retrieve 100 blocks, it's perfectly fine for the client to optimize the performance, keep the connection open until the batch of the operations finish.
Keeping the connection always on will have some other complications, for example:
How long the keep alive will be? (this might be an easy answer, we just need to be clear about it).
Will there be possibility for someone else to hijack the connection that other process previously established? (I don't have a specific scenario, but it's always good to consider it at infrastructure layer).

weirdgiraffe:
[...]

PR #2451, opened 2025-10-30
rpcclient: make shutdown interrupt in-flight POSTs

wydengyre:
## Change Description
Modify the rpcclient http POST call to ensure that a shutdown immediately interrupts in-flight requests, which otherwise would have to wait until timeout.

This PR includes the change in https://github.com/btcsuite/btcd/pull/2450 as it is necessary to ensure interruption during Dial.

## Steps to Test
The test suite has an added test to ensure this works.

### Testing

### DIFF
--- a/rpcclient/infrastructure.go
+++ b/rpcclient/infrastructure.go
@@ -789,10 +789,6 @@ retryloop:
 		bodyReader := bytes.NewReader(jReq.marshalledJSON)
 		httpReq, err = http.NewRequestWithContext(ctx, "POST", httpURL, bodyReader)
 		if err != nil {
-			// We must observe the contract that shutdown returns ErrClientShutdown.
-			if errors.Is(err, context.Canceled) && errors.Is(context.Cause(ctx), ErrClientShutdown) {
-				err = ErrClientShutdown
-			}
 			jReq.responseChan <- &Response{result: nil, err: err}
 			return
 		}
--- a/rpcclient/infrastructure_test.go
+++ b/rpcclient/infrastructure_test.go
@@ -161,12 +161,14 @@ func TestHTTPPostShutdownInterruptsPendingRequest(t *testing.T) {
 
 	future := client.GetBlockCountAsync()
 
+	// Ensure the server sees the request before we initiate shutdown.
 	select {
 	case <-requestAccepted:
 	case <-time.After(2 * time.Second):
 		t.Fatalf("server did not accept client connection")
 	}
 
+	// The request should remain pending until shutdown is requested.
 	select {
 	case <-future:
 		t.Fatalf("expected request to remain pending until shutdown")
@@ -181,6 +183,7 @@ func TestHTTPPostShutdownInterruptsPendingRequest(t *testing.T) {
 		close(waitDone)
 	}()
 
+	// Wait for shutdown to complete before asserting the final error.
 	select {
 	case <-waitDone:
 	case <-time.After(5 * time.Second):


kmk142789 (approved):

rpcclient/infrastructure.go
```diff
@@ -782,12 +782,17 @@ func (c *Client) handleSendPostMessage(jReq *jsonRequest) {
...
+retryloop:
 	for i := 0; i < tries; i++ {
 		var httpReq *http.Request
 
 		bodyReader := bytes.NewReader(jReq.marshalledJSON)
-		httpReq, err = http.NewRequest("POST", httpURL, bodyReader)
+		httpReq, err = http.NewRequestWithContext(ctx, "POST", httpURL, bodyReader)
 		if err != nil {
+			// We must observe the contract that shutdown returns ErrClientShutdown.
+			if errors.Is(err, context.Canceled) && errors.Is(context.Cause(ctx), ErrClientShutdown) {
```
kcalvinalvin: This check is unneeded since `http.NewRequestWithContext` will never return a error that's of type `context.Canceled`.

kcalvinalvin:
Checked:

1: The added `TestHTTPPostShutdownInterruptsPendingRequest` test is correct.
2: Sanity checked the changes in `infrastructure.go`.

One nitpicky gripe is maybe the test could be commented better.

Looks good overall except for that one unneeded check.

rpcclient/infrastructure.go
wydengyre: Removed in https://github.com/btcsuite/btcd/pull/2451/commits/f87d91434ee7912a128a4606a9401427a70ad96a

wydengyre:
> Checked:
> 
> 1: The added `TestHTTPPostShutdownInterruptsPendingRequest` test is correct. 2: Sanity checked the changes in `infrastructure.go`.
> 
> One nitpicky gripe is maybe the test could be commented better.
> 
> Looks good overall except for that one unneeded check.

@kcalvinalvin thanks for your review.

I've addressed your comments in https://github.com/btcsuite/btcd/pull/2451/commits/f87d91434ee7912a128a4606a9401427a70ad96a

Can you re-review?

seeforschauer:
Hey @wydengyre, nice work on this — we hit the same pain point in production.

We run a HAProxy health checker that polls BTC nodes via `GetBlockCount()`. Each probe spawns a goroutine; when a node hangs, `GetBlockCount` blocks for ~600s (retry loop in `handleSendPostMessage` — 10 attempts x 60s HTTP timeout) while new probes keep arriving. We observed **117 leaked goroutines** on staging.

We searched the issue tracker and PRs for existing context support work — found #2450 (dial timeout, merged), this PR (shutdown interrupt), and #1323 (connection reuse), but nothing addressing per-request `context.Context` for RPC methods. The gap is clear: callers currently have no way to cancel or timeout an individual `GetBlockCount`/`SendRawTransaction`/etc. call.

Your PR fixes shutdown propagation, which is great. The transport plumbing is almost there with `http.NewRequestWithContext`. Would you or maintainers be open to extending this to per-request context support? Something like:

```go
func (c *Client) GetBlockCountWithContext(ctx context.Context) (int64, error)
```

Minimal changes: `ctx` field on `jsonRequest`, `ctx.Done()` check in the retry loop, `SendCmdCtx` variant. Non-breaking — existing methods keep working. ~80-100 lines.

Happy to open an issue + PR for this if there's interest. We worked around it with a singleflight pattern at our layer, but the proper fix belongs here.

wydengyre:
> Would you or maintainers be open to extending this to per-request context support?

@seeforschauer I had the same problem as you, and actually have a private library that fixes it in a backward-compatible way. But I'm not affiliated with this project in any official way, so sadly the decision to support such a feature is outside my influence. I would certainly vote in favor of it, though, as it's both pragmatic/necessary and more "Go-like".

seeforschauer:
Thanks for the support, and good to know you've solved it independently — validates the approach.

I went ahead and opened #2499 to formalize the request for per-request `context.Context` support. If you'd be open to sharing your backward-compatible approach (or collaborating on a PR), happy to coordinate there.

In the meantime, pinging @kcalvinalvin — any thoughts on merging this PR as-is and tracking context support separately in #2499?

rpcclient/infrastructure.go
starius, line 812: If `errors.Is(err, context.Canceled) && errors.Is(context.Cause(ctx), ErrClientShutdown) && i == tries-1`, the added error remapping logic is skipped. The code returns context.Canceled in this case, but it is expected to return ErrClientShutdown instead.

starius, line 865: If `ctx` is cancelled in the middle of `io.ReadAll` running, we return `fmt.Errorf("error reading json reply: ...`, but I think we should also return ErrClientShutdown in this case, right?

rpcclient/infrastructure_test.go
```diff
@@ -108,3 +112,85 @@ func TestParseAddressString(t *testing.T) {
...
+	select {
+	case <-waitDone:
+	case <-time.After(5 * time.Second):
+		t.Fatalf("client shutdown did not complete")
+	}
+
+	result, err := future.Receive()
+	require.Zero(t, result)
+	require.ErrorContains(t, err, ErrClientShutdown.Error())
+}
```
starius, line 196: I propose to add more test coverage of the edge cases:

- Cancellation while waiting in retry backoff path (`ctx.Done()` select branch).
- Cancellation on final retry attempt (`i == tries-1`).
- Cancellation during response body read after a successful `Do`.

starius:
I'm going to copy the commits from this PR on top of https://github.com/btcsuite/btcd/pull/2500 and resolve the remaining comments.

@wydengyre If you already addressed some of them, please push, so I have the latest version of your commits.

wydengyre:
> @wydengyre If you already addressed some of them, please push, so I have the latest version of your commits.

I have not, sorry. Please go ahead.

starius:
This PR was integrated into https://github.com/btcsuite/btcd/pull/2500

PR #2476, opened 2026-01-14
btcjson: fix StringOrArray JSON marshaling and add warnings to GetBlockChainInfoResult

### LINKED ISSUE #2369: [bug]: StringOrArray.MarshalJSON triggers infinite recursion → stack-overflow panic

### LINKED ISSUE #2444: [bug]: type mismatch at  response of 'GetBlockChainInfo'
### Background

Describe your issue here.

### Your environment

* bitcore core@latest 
* rpcclient 

### Actual behavior

getting this error as `GetBlockChainInfo` response 

`{
  "error": "unable to detect bitcoind version: json: cannot unmarshal array into Go struct field GetNetworkInfoResult.warnings of type string"
}`

EricGrill:
## Summary
- Fixes infinite recursion in `StringOrArray.MarshalJSON` by converting to underlying `[]string` type before marshaling
- Adds `Warnings` field to `GetBlockChainInfoResult` using `StringOrArray` type

## Background
The `StringOrArray.MarshalJSON` method was calling `json.Marshal(h)` which caused infinite recursion because Go sees that `h` implements `json.Marshaler` and calls `MarshalJSON()` again.

Bitcoin Core added a `warnings` field to `getblockchaininfo` in PR #10858, and recently changed it from a string to an array in PR #29845. Using `StringOrArray` handles both formats.

## Test plan
- [x] `StringOrArray.MarshalJSON` no longer causes stack overflow
- [x] `GetBlockChainInfoResult` can unmarshal warnings as string or array

Closes #2369
Closes #2444

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/btcjson/chainsvrresults.go
+++ b/btcjson/chainsvrresults.go
@@ -371,7 +371,9 @@ type StringOrArray []string
 
 // MarshalJSON implements the json.Marshaler interface.
 func (h StringOrArray) MarshalJSON() ([]byte, error) {
-	return json.Marshal(h)
+	// Convert to []string to avoid infinite recursion since calling
+	// json.Marshal on StringOrArray would invoke MarshalJSON again.
+	return json.Marshal([]string(h))
 }
 
 // UnmarshalJSON implements the json.Unmarshaler interface.
--- a/btcjson/chainsvrresults.go
+++ b/btcjson/chainsvrresults.go
@@ -226,18 +226,19 @@ type UnifiedSoftForks struct {
 // GetBlockChainInfoResult models the data returned from the getblockchaininfo
 // command.
 type GetBlockChainInfoResult struct {
-	Chain                string  `json:"chain"`
-	Blocks               int32   `json:"blocks"`
-	Headers              int32   `json:"headers"`
-	BestBlockHash        string  `json:"bestblockhash"`
-	Difficulty           float64 `json:"difficulty"`
-	MedianTime           int64   `json:"mediantime"`
-	VerificationProgress float64 `json:"verificationprogress,omitempty"`
-	InitialBlockDownload bool    `json:"initialblockdownload,omitempty"`
-	Pruned               bool    `json:"pruned"`
-	PruneHeight          int32   `json:"pruneheight,omitempty"`
-	ChainWork            string  `json:"chainwork,omitempty"`
-	SizeOnDisk           int64   `json:"size_on_disk,omitempty"`
+	Chain                string        `json:"chain"`
+	Blocks               int32         `json:"blocks"`
+	Headers              int32         `json:"headers"`
+	BestBlockHash        string        `json:"bestblockhash"`
+	Difficulty           float64       `json:"difficulty"`
+	MedianTime           int64         `json:"mediantime"`
+	VerificationProgress float64       `json:"verificationprogress,omitempty"`
+	InitialBlockDownload bool          `json:"initialblockdownload,omitempty"`
+	Pruned               bool          `json:"pruned"`
+	PruneHeight          int32         `json:"pruneheight,omitempty"`
+	ChainWork            string        `json:"chainwork,omitempty"`
+	SizeOnDisk           int64         `json:"size_on_disk,omitempty"`
+	Warnings             StringOrArray `json:"warnings"`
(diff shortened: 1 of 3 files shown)


guibressan:
I've written a test that reproduces the bug. Here's the diff:

```diff
diff --git a/btcjson/chainsvrresults_test.go b/btcjson/chainsvrresults_test.go
index 122af3dc..715de4a9 100644
--- a/btcjson/chainsvrresults_test.go
+++ b/btcjson/chainsvrresults_test.go
@@ -5,6 +5,7 @@
 package btcjson_test

 import (
+       "bytes"
        "encoding/json"
        "net/url"
        "reflect"
@@ -263,3 +264,33 @@ func TestGetNetworkInfoWarnings(t *testing.T) {
                }
        }
 }
+
+func TestStringOrArrayMarshalJson(t *testing.T) {
+       t.Parallel()
+
+       tests := []struct{
+               input []string
+               expected []byte
+       }{
+               {
+                       input: []string{"test"},
+                       expected: []byte(`["test"]`),
+               },
+               {
+                       input: []string{"test", "test1"},
+                       expected: []byte(`["test","test1"]`),
+               },
+       }
+
+       t.Logf("Running %d tests", len(tests))
+       for _, test := range tests {
+               s := btcjson.StringOrArray(test.input)
+               jsonBytes, err := s.MarshalJSON()
+               if err != nil {
+                       t.Errorf("unexpected error %s", err)
+               }
+               if !bytes.Equal(test.expected, jsonBytes) {
+                       t.Errorf("expected %s got %s", test.expected, jsonBytes)
+               }
+       }
+}
```

Before the patch - master branch (dd78e092763785ac00506ee03ddb268454d985d7):
Screenshot 2026-04-01 at 17 39 57

After the patch:
Screenshot 2026-04-01 at 18 10 57

Also verified Bitcoin Core documentation for the RPC method [getblockchaininfo](https://bitcoincore.org/en/doc/30.0.0/rpc/blockchain/getblockchaininfo) and in fact, there's a `"warnings":["string"]` field, which was a single string in older bitcoind versions.

My only recommendation is to add a new commit with the test.

LGTM

Good job @EricGrill.

Roasbeef (approved):
LGTM 🦊

Added a regression test directly to the PR

PR #2478, opened 2026-01-19
genswiftsyncbitmap, chaincfg, blockchain: implement swiftsync

kcalvinalvin:
## Change Description

Adds  [SwiftSync](https://gist.github.com/RubenSomsen/a61a37d14182ccd78760e477c78133cd), a way to
build the UTXO set with minimal disk access. A few things are cleaned up from the last draft:

- hintsfiles are now committed by sha256 hashes and are gzipped
- recovery when the node has crashed and is resuming during the swiftsync part of the ibd
- aggregator is now implemented (critical for verification)

### Hintsfile format

We number every output in blockchain order, then list the indices of the unspent ones in increasing order (OP_RETURN outputs, outputs with pkscript size that's greater than 10,000 bytes, and BIP30 duplicate coinbase outputs are excluded).

The file's format is a single gzip file where the header is a `"UTXO"` magic, a version byte, and the height as a little endian uint32. Then we have one entry per block: a CompactSize byte length followed by gap encoded CompactSize indices of that block's still unspent spendable outputs.

The length prefix lets entries be located without decoding, so they decode lazily during sync. The result stays
small and compresses well (mainnet about 88 MB, testnet3 about 19 MB). It is generated by `cmd/genswiftsynchints`.

### Boundaries

A network supports swift sync by declaring a boundary: the checkpoint height a hintsfile must cover for that
network, plus the hintsfile's expected hash. This is a new field on the per network parameters (`chaincfg.Params`):
`SwiftSync *SwiftSyncCheckpoint{Height, HintsHash}`, left nil where swift sync is not supported.

`HintsHash` is the SHA256 of the gzipped hintsfile. The loader checks it before inflating the file, so a wrong or
corrupt hintsfile is rejected at startup instead of after a full sync.

Three networks are configured:

* signet: height 285205, hintsfile sha256 `df2a00fe29905bbc1897d23a04f169ea2b5e909b53e55dacc41eb1ab9d04edde`
* mainnet: height 933038, hintsfile sha256 `baf5875e7b62173514d9063687e7f422d168ffb7057190dcdeebcc6f9edbc85c`
* testnet3: height 4950000, hintsfile sha256 `3eb03d76b0a88299283c07cb97a5cc6608cf5f146aab9aa10ade7d5e99f21976`

A swift sync boundary must be a checkpoint, so the boundary block is pinned by checkpoint enforcement. mainnet and testnet3 add a new `Checkpoints` entry at their boundary.

At startup btcd looks for `sshints.dat` via `--swiftsynchints`, then `<datadir>/<network>/`, then next to the
binary. A missing or hash mismatched file is nonfatal. Swift sync only engages on a fresh database.

### Not in this PR: shipping the hintsfile
The code and the boundary/hash config are here. The hintsfiles themselves and the release are left for a
follow up.

I propose committing the files under `hintsfile/<network>/sshints.dat` via Git LFS. `release.sh` would then copy the file next to the binaries. I'm open to other distribution methods as well.

## Steps to Test

1. Build the generator and produce a hintsfile (signet is quickest), or obtain one:
   `go run ./cmd/genswiftsynchints --signet` writes `sshints.dat` (gzipped) into the signet data directory.
2. Start a node on a fresh data directory with the file in place:
   `./btcd --signet --datadir=<fresh> --logdir=<fresh>`.
3. Watch the logs. You should see `Loaded swift sync hintsfile ... covering blocks 1..285205`, then
   `Swift sync enabled: target height=285205`, then `Swift sync completed at height 285205`. Block processing
   should not report per block transaction counts, since transactions are not replayed.
4. Negative path: corrupt or swap the file. The node logs `Swift sync disabled: ... hash ... want ...`
   (or a height mismatch) and falls back to a normal sync.
5. Crash the node with sigkill and restart to see if it actually restarts properly.
6. Run the unit tests: `go test ./blockchain/... ./chaincfg/... .` and `go test ./cmd/genswiftsynchints/...`.
7. Of course, read the tests to see they're doing what they're claiming. Read over the code to check for any dumb things that might be happening.

### Testing

<details>
<summary>Original description (outdated)</summary>
## Change Description

This is a first implementation of [SwiftSync](https://gist.github.com/RubenSomsen/a61a37d14182ccd78760e477c78133cd) for btcd. SwiftSync enables fast UTXO set bootstrapping during initial block download by using a precomputed bitmap that indicates which transaction outputs have been spent.

Instead of replaying and validating all historical transactions from genesis, nodes can use the bitmap to directly construct the UTXO set:
  - Bit 0 = output is unspent (add to UTXO cache)
  - Bit 1 = output was spent (skip)

The main benefit is performance improvement during ibd. On this PR when testing on signet, it was 3.3x faster than on master.

Without SwiftSync on master to block height 285,205 took 22m 59s.

Ran with command:
```
./btcd --datadir=. --logdir=. --signet --cpuprofile=cpuprof-without-ss --memprofile=memprof-without-ss --connect=127.0.0.1 --addcheckpoint=285205:00000004b1cc694c48295fc56c2d78b88abdd648ee60a21548feba259e6dedf1
```

pprof:
IMG_0571

With SwiftSync on this PR to block height 285,205 took 6m 56s.

Ran with command:

```
./btcd --datadir=. --logdir=. --signet --cpuprofile=cpuprof-with-ss --memprofile=memprof-with-ss --connect=127.0.0.1
```

IMG_0572

One caveat is that we can't generate the SpendJournals for the blocks that are processed with SwiftSync.
Note that the aggregator part of SwiftSync is not yet implemented which makes this not safe to run for production uses.

This is a first pass at the implementation and is not final. Feedback welcome.

## Steps to Test

### Testing

  <details>


GeorgeTsagk:
Will review this

GeorgeTsagk:
Took a look at this alongside [2140-dev/swiftsync](https://github.com/2140-dev/swiftsync) and https://gist.github.com/RubenSomsen/a61a37d14182ccd78760e477c78133cd.

Some notes on compatibility, since the reference implementation has changed a bit since this PR was opened:

#### Output filtering
The reference impl skips OP_RETURN and scripts >10,000 bytes when building the bitmap — they're provably unspendable so they're not counted. This PR indexes every output. This means the bit indices won't line up between the two implementations even if the encoding matched. Worth aligning early since it affects any cross-implementation tooling.

#### Bitmap format
The reference recently switched to Elias-Fano encoding per block. It is quite different, and more complicated, than the flat bit field. It does trim a good amount of MBs though.

------

The 3.3x speedup from just skipping validation/lookups is impressive.

One thing worth thinking about: the spec's main selling point is that the accumulator makes block processing fully order-independent, which opens the door to downloading and processing blocks from multiple peers in parallel. Would be great to see that as a follow-up once the accumulator is in place. That's where the really big IBD gains come from.

rustaceanrob:
We've formalized the recommended approach more with the [following specifications](https://github.com/2140-dev/swiftsync-bips), which detail the new hintsfile format.

allocz:
I ran the IBD with 128MB utxcache until signet height 285205 and master processed average of 17 blocks/s while SwiftSync processed average of 36 blocks/s.

Nice job.

edilmedeiros:
SwiftSync [hint file specification](https://github.com/2140-dev/swiftsync-bips/blob/master/hintsfile.md) was changed based on Elias-Fano encoding instead of a plain bitset.

sipa is working on an [alternative encoding](https://github.com/bitcoin/bitcoin/compare/master...sipa:bitcoin:wip_swifthints) based on range Asymmetric Numeral Systems (rANS), which can potentially squeeze even more bits from the hints file.

I went through this PR (not tested) and I think this is a good starting point. It implements the main infrastructure required for using SwiftSync in IBD. The hint file deserialization can be added later.

It doesn't feel too hard too change this when the file format settles (maybe initially decoding the file upfront and loading the data into the `SwiftSyncData` data structure and call it a day at the cost of memory and setup time - some hundreds of miliseconds?)

Both Rob's and sipa's proposals allow for some random access. The Elias-Fano scheme encodes block by block, sipa's encodes groups of blocks, no more than 32768 blocks together. The `SwiftSyncData` data structure could be redesigned to hold data for only a block, taking information from the decoder when it's needed.

kcalvinalvin:
> SwiftSync [hint file specification](https://github.com/2140-dev/swiftsync-bips/blob/master/hintsfile.md) was changed based on Elias-Fano encoding instead of a plain bitset.

I already lost trust in the process of the "specification". Too much wasted time there. Just encoding with varints and then gzipping the thing gets us better numbers than Elias-Fano anyways. If there's something better than good-ole varints and gzip, I'd take a stab. But right now, I see no reason to follow the "specification" since it's worse.

rustaceanrob:
> Just encoding with varints and then gzipping the thing gets us better numbers than Elias-Fano anyways

Can you be precise with these findings? Elias-fano encoding at block 930,000 is 119mb, with `zstd` at `-22` it is reduced to 104mb. What is the side of the varint and gzip encoding before and after compression?

saubyk:
cc: @GeorgeTsagk this pr is now ready for review if you want to take a look
cc: @guggero and @TechLateef would be cool to have your review here as well

kcalvinalvin:
> > Just encoding with varints and then gzipping the thing gets us better numbers than Elias-Fano anyways
> 
> Can you be precise with these findings? Elias-fano encoding at block 930,000 is 119mb, with `zstd` at `-22` it is reduced to 104mb. What is the side of the varint and gzip encoding before and after compression?

88MB with varint&gzip at height 933038. Github won't let me upload beyond 25MB but I can post elsewhere.

image

blockchain/chain.go
```diff
@@ -1179,6 +1216,68 @@ func (b *BlockChain) connectBestChain(node *blockNode, block *btcutil.Block, fla
...
 		fastAdd = fastAdd || b.index.NodeStatus(node).KnownValid()
 
+		// Swift sync: use the precomputed hintsfile for fast UTXO
+		// bootstrapping.  This skips all validation and uses the hints
+		// to directly construct the UTXO set.
+		if b.isSwiftSyncActive(node.height) {
+			// Add only unspent outputs to the UTXO cache using the
+			// per-block hints, and update the aggregate that validates
+			// them at the boundary.
+			err := b.utxoCache.swiftSyncConnectTransactions(
```
TechLateef, line 1226: The existing comment above the `swiftSyncConnectTransactions` call explains what the function does but doesn't document that the third argument controls aggregate updating and must be `false` on resume. Since passing the wrong value would silently corrupt the aggregate, it would be worth noting that explicitly in the comment  e.g. "pass true on first sync, false on resume to avoid double-counting the aggregate."

mk-lateef (commented):
LGTM!

TechLateef (commented):
LGTM! Although it would be worth adding tests for mainnet and testnet3 similar to `TestSigNetSwiftSyncCheckpoint` to catch any hash or height typos at CI time rather than at runtime.

cmd/genswiftsynchints/genswiftsynchints.go
```diff
@@ -0,0 +1,1364 @@
...
+
+	swiftSyncHintsVersion = 0x01
+
+	// swiftSyncHintsHeaderSize is the byte length of the hintsfile header:
+	// magic(4) + version(1) + height(4).
+	swiftSyncHintsHeaderSize = 9
+)
+
+var (
+	cfg *config
```
guggero, line 73: Having a global variable is bad practice IMO. And it looks like for most functions just the `cfg.DataDir` is used, which can easily be passed into these functions to make them "pure".

cmd/genswiftsynchints/genswiftsynchints.go
```diff
@@ -0,0 +1,1364 @@
...
+		return startIndex + uint64(numOutputs), true, nil
+	}
+
+	return 0, true, nil
+}
+
+// genOPMapping generates the outpoint -> bit index mapping by iterating through
+// all blocks. It is resumable: progress is stored in Pebble and the function
+// can be stopped via context cancellation.
+func genOPMapping(ctx context.Context, pdb *pebble.DB, chain *blockchain.BlockChain, lastHash *chainhash.Hash) error {
```
guggero, line 347: Please [wrap lines at 80 character](https://github.com/btcsuite/btcd/blob/master/docs/code_formatting_rules.md#80-character-line-length) to keep things readable.
You can use my [`goformat`](https://github.com/guggero/goformat) tool to help your agent out with keeping to the style guide. It's not perfect yet but should avoid a lot of manual re-formatting.

go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/btcsuite/go-socks v0.0.0-20170105172521-4720035b7bfd
 	github.com/btcsuite/websocket v0.0.0-20150119174127-31079b680792
 	github.com/btcsuite/winsvc v1.0.0
+	github.com/cockroachdb/pebble/v2 v2.1.6
```
guggero, line 18: Is there any reason for _not_ just using the `goleveldb` package we already use? Looks like PebbleDB pulls in a lot of dependencies... If it's just performance, can you roughly outline what the difference would be?

guggero: Or, I guess since this DB is only used in the `cmd/genswiftsynchints` package, that should instead be turned into a Go module to avoid all these dependencies being pulled in for the main `btcd` package.

chaincfg/swiftsync.go
```diff
@@ -0,0 +1,216 @@
...
+	return &SwiftSyncData{
+		Height:  int32(fileHeight),
+		body:    body,
+		offsets: offsets,
+	}, nil
+}
+
+// readCompactSize parses a Bitcoin CompactSize (varint) from buf, returning
+// the parsed value and the number of bytes consumed.
+func readCompactSize(buf []byte) (uint64, int, error) {
```
guggero, line 131: We already import the `wire/v2` submodule, why not just use `wire.ReadVarIntBuf()` here instead of duplicating the code?

cmd/genswiftsynchints/genswiftsynchints.go
```diff
@@ -0,0 +1,1364 @@
...
+	swiftSyncHintsHeaderSize = 9
+)
+
+var (
+	cfg *config
+
+	syncWriteOptions = pebble.Sync
+
+	// swiftSyncHintsMagic is the hintsfile magic, the ASCII bytes "UTXO".
+	swiftSyncHintsMagic = [4]byte{0x55, 0x54, 0x58, 0x4f}
```
guggero, line 78: This is re-defined in `chaincfg` in the next commit. Perhaps we should move all code to _produce_ a hints file into `chaincfg` as well? Then we'd also have a bit more context on the file format, at least for that file. Otherwise a reviewer is presented with a 1.1k line file to review in this commit with minimal context about either the database nor the file structure of the involved components.

cmd/genswiftsynchints/genswiftsynchints.go
```diff
@@ -0,0 +1,1364 @@
...
+// metaKey returns the Pebble key for a metadata entry.
+func metaKey(name string) []byte {
+	key := make([]byte, 1+len(name))
+	key[0] = metaPrefix
+	copy(key[1:], name)
+	return key
+}
+
+// getMetaUint64 retrieves a uint64 metadata value from Pebble.
+func getMetaUint64(db *pebble.DB, name string) (uint64, error) {
```
guggero, line 124: Similar comment about DB related code. Perhaps that should be moved into its own file, so it's easier to grasp what the "business" logic of the hint file creation tool is.

blockchain/chain.go
```diff
@@ -1179,6 +1216,68 @@ func (b *BlockChain) connectBestChain(node *blockNode, block *btcutil.Block, fla
...
+			)
+			if err != nil {
+				return false, err
+			}
+
+			b.swiftSyncHeight = node.height
+
+			// Connect the block to the main chain. Pass nil for stxos
+			// since we don't need spend journal during swift sync.
+			err = b.connectBlock(node, block, nil)
```
guggero, line 1237: The STXOs aren't only used for journal entries but also any active index. For example, the CfIndex and address index both use the STXOs and would silently corrupt if this is passed as `nil`.
So I think we need to enforce disabling these indexes when SwiftSync is enabled.
Or find a way to assemble the STXOs before calling this.

blockchain/chain.go
```diff
@@ -1179,6 +1216,68 @@ func (b *BlockChain) connectBestChain(node *blockNode, block *btcutil.Block, fla
...
+			b.index.SetStatusFlags(node, statusValid)
+			flushIndexState()
+
+			// At the swift sync boundary, verify the aggregate is zero,
+			// which proves the hints are correct and the UTXO set we
+			// built is valid.  The boundary block is the swift sync
+			// checkpoint, so checkpoint enforcement already pins its
+			// identity.
+			if node.height == b.swiftSync.Height {
+				if !b.swiftSyncAgg.IsZero() {
```
guggero, line 1251: Should this check come before committing the best state, height and so on in `b.connectBlock()`?

blockchain/chainio.go
```diff
@@ -1162,6 +1232,18 @@ func (b *BlockChain) createChainState() error {
 			return err
 		}
 
+		// When the database is created with swift sync enabled, generate and
+		// store its secret salt.  The salt is constant for the whole sync and
+		// is what lets the aggregate detect an invalid or malicious hintsfile.
+		if b.swiftSyncEnabled {
+			if _, err := rand.Read(b.swiftSyncSalt[:]); err != nil {
```
guggero, line 1239: What if we're resuming a previous sync that was interrupted? Don't we need to re-use the same salt as before?
Also, if we started syncing with a hints file (and wrote a salt), but now for some reason the hints file is no longer present or disabled, then we would fall back to normal syncing but the UTXO set would be incomplete and fail with a cryptic (to the user) error.
So we should also check the presence of a salt independent of `swiftSyncEnabled`.

server.go
```diff
@@ -2859,6 +2864,157 @@ func setupRPCListeners() ([]net.Listener, error) {
 	return listeners, nil
 }
 
+// swiftSyncHintsFilename is the default name of the swift sync hintsfile.  It is
+// searched for in the network's data directory and next to the btcd binary, and
+// matches the name cmd/genswiftsynchints writes, so a freshly generated or
+// bundled file is found without any extra configuration.
+const swiftSyncHintsFilename = "sshints.dat"
```
guggero, line 2871: nit: collect constants in a single `const` statements at the top of the file.

server.go
```diff
@@ -2978,13 +3134,23 @@ func newServer(listenAddrs, agentBlacklist, agentWhitelist []string,
...
 	}
 
+	// Load a swift sync hintsfile if one is available so the UTXO set can be
+	// bootstrapped without replaying historical spends.  A missing file or a
+	// load error just means a normal sync.
+	swiftSync, err := loadSwiftSyncHints(
+		s.chainParams, cfg.DataDir, cfg.SwiftSyncHints,
+	)
+	if err != nil {
+		btcdLog.Warnf("Swift sync disabled: %v", err)
```
guggero, line 3144: If the user set a specific `--swiftsynchints` path (not auto-detected), then just logging an error and continuing normal sync might not be the right mode. Should we fail hard if `--swiftsynchints` is not empty?

server.go
```diff
@@ -2859,6 +2864,157 @@ func setupRPCListeners() ([]net.Listener, error) {
...
+	}
+	if int64(len(compressed)) > maxSwiftSyncHintsFileSize {
+		return nil, fmt.Errorf("swift sync hintsfile %q is too large "+
+			"(limit %d bytes)", path, maxSwiftSyncHintsFileSize)
+	}
+
+	// Reject a hintsfile whose hash does not match the one pinned for the
+	// network, catching a wrong or corrupt file here rather than at the
+	// aggregate check after a full sync.
+	if cfg.HintsHash != "" {
```
guggero, line 2946: Should there be an error or at least warning if the `HintsHash` is not set?

guggero (commented):
Did a first pass. The actual diff in production code is reasonably small.
The main issue I see is the interaction with any potentially enabled index that might stop working correctly due to the missing STXO information.

cmd/genswiftsynchints/genswiftsynchints.go
kcalvinalvin: Ok I will edit this to wrap at 80 but I'm very much against 80 character wraps.

btcd code *had not* had a 80 char rule historically. There are many parts of the code that doesn't do this.
80 char rule is *outdated* and even linux kernel that's been picky about this has a 100 char rule.

The example in the doc is terrible. The wrapped version is so much more unreadable and ugly.
```
myKey := "0214cd678a565041d00e6cf8d62ef8add33b4af4786fb2beb87b366a2e1" +
	"51fcee7"
```

This is *much* superior.
```
myKey := "0214cd678a565041d00e6cf8d62ef8add33b4af4786fb2beb87b366a2e151fcee7"
```

Yes to line wrapping somewhere. We don't wanna be Bitcoin Core with function signatures that are 1000 char long.

But 80? Cmon. 120 is fine too imo. I'd be happy with 100 too. Most people here have modern monitors right? If someone's programming on 80×25 CRT monitor I'll happily buy them something new.

go.mod
kcalvinalvin: > Is there any reason for _not_ just using the `goleveldb` package we already use? Looks like PebbleDB pulls in a lot of dependencies... If it's just performance, can you roughly outline what the difference would be?

I thought this was a good place to sorta test out pebbledb because we've been toying with the idea of replacing goleveldb with it https://github.com/btcsuite/btcd/issues/2024. It's an isolated package so nothing wrong is gonna happen to btcd.

> Or, I guess since this DB is only used in the `cmd/genswiftsynchints` package, that should instead be turned into a Go module to avoid all these dependencies being pulled in for the main `btcd` package.

I'll do that.

blockchain/chain.go
kcalvinalvin: Oh. That's a problem then.

For swiftsync you just don't have that information at hand. You do know which txos are spent so I guess you can still make them but this would need some thinking..

> So I think we need to enforce disabling these indexes when SwiftSync is enabled.

I'll make this for now since what we have now is catastrophic.

blockchain/chainio.go
kcalvinalvin: > What if we're resuming a previous sync that was interrupted? Don't we need to re-use the same salt as before?

Yes we need to use the same salt as before. But since this is under `createChainState()` that only gets called for brand new databases, we're fine.

> Also, if we started syncing with a hints file (and wrote a salt), but now for some reason the hints file is no longer present or disabled, then we would fall back to normal syncing but the UTXO set would be incomplete and fail with a cryptic (to the user) error. So we should also check the presence of a salt independent of `swiftSyncEnabled`.

This is handled in `initChainState()` at the bottom from lines 1394-1412.

server.go
kcalvinalvin: That would be more user friendly. I'll do that.

blockchain/chain.go
kcalvinalvin: We've already haved all the best state for all the blocks prior to the swiftsync block so I think it's ok if we still save an extra block.

But yeah, I can still make the change.

blockchain/swiftsync.go
```diff
@@ -0,0 +1,144 @@
...
+// to the accumulator width.  The salt is unknown to whoever produced the
+// hintsfile, so the value cannot be predicted and a crafted hintsfile cannot
+// pick offsetting wrong hints that cancel the aggregate.
+func saltedOutpoint(salt *[32]byte, op *wire.OutPoint) [agg512Size]byte {
+	var v [agg512Size]byte
+	copy(v[0:32], op.Hash[:])
+	binary.LittleEndian.PutUint32(v[32:36], op.Index)
+	copy(v[36:68], salt[:])
+	return v
+}
```
pzafonte, line 144: Hi new to this repo, but I have a background in Go and am really interested in SwiftSync. I was wondering if this need to be hashed, or if this implementation is choosing not to for some reason that I'm not understanding. The [reference design](https://gist.github.com/RubenSomsen/a61a37d14182ccd78760e477c78133cd#regarding-the-hash-aggregate) hashes outpoint‖salt before summing.

```suggestion
func saltedOutpoint(salt *[32]byte, op *wire.OutPoint) [agg512Size]byte {
	var idx [4]byte
	binary.LittleEndian.PutUint32(idx[:], op.Index)

	h := sha256.New()
	h.Write(op.Hash[:])
	h.Write(idx[:])
	h.Write(salt[:])
	sum := h.Sum(nil)

	var v [agg512Size]byte
	copy(v[0:32], sum)
	return v
}
```

allocz:
@kcalvinalvin There's a requirement of minimum amount of RAM? I'm trying to generate the hints on mainnet, but I'm getting high memory usage.

scr

blockchain/chain.go
GeorgeTsagk, line 1251: The `IsZero()` check runs after `connectBlock` already committed the boundary block, height, and non-zero aggregate. On restart `initChainState` reloads that state without re-checking. This is the only `IsZero()` call site and it never re-runs once the boundary is the tip, so the node syncs on a UTXO set built from unverified hints. Suggest checking the aggregate before committing the boundary block, and persisting a verified flag startup won't proceed without.

<details>
<summary>Repro test, <code>go test -run TestSwiftSyncBadAggSurvivesRestart</code></summary>

```go
func TestSwiftSyncBadAggSurvivesRestart(t *testing.T) {
	const boundary = int32(3)

	dbPath := filepath.Join(t.TempDir(), "ssrepro")
	db, err := database.Create(testDbType, dbPath, blockDataNet)
	require.NoError(t, err)
	defer db.Close()

	// A wrong or malicious hintsfile leaves the boundary aggregate non-zero;
	// TestSwiftSyncAggregateCancels proves bad hints produce exactly this.
	var badAgg Agg512
	badAgg[0] = 1
	require.False(t, badAgg.IsZero())

	var salt [32]byte
	salt[0] = 0x42

	// First run: mimic what connectBlock commits while connecting the boundary
	// block. This happens BEFORE connectBestChain evaluates IsZero(), so the
	// invalid state is already durable when the check fails.
	require.NoError(t, db.Update(func(dbTx database.Tx) error {
		if err := dbPutSwiftSyncSalt(dbTx, &salt); err != nil {
			return err
		}
		if err := dbPutSwiftSyncHeight(dbTx, boundary); err != nil {
			return err
		}
		return dbPutSwiftSyncAgg(dbTx, &badAgg)
	}))

	// Second run: a fresh chain loads the persisted swift sync state exactly as
	// initChainState does. This is the whole resume path for swift sync state.
	b := &BlockChain{
		swiftSync:        &chaincfg.SwiftSyncData{Height: boundary},
		swiftSyncEnabled: true,
	}
	require.NoError(t, db.View(func(dbTx database.Tx) error {
		b.swiftSyncHeight = dbFetchSwiftSyncHeight(dbTx)
		b.swiftSyncAgg = dbFetchSwiftSyncAgg(dbTx)
		s, ok := dbFetchSwiftSyncSalt(dbTx)
		require.True(t, ok)
		b.swiftSyncSalt = s
		return nil
	}))

	// The invalid state survived the restart untouched, and nothing re-checks it.
	require.Equal(t, boundary, b.swiftSyncHeight)
	require.False(t, b.swiftSyncAgg.IsZero(),
		"non-zero aggregate reloaded verbatim, no validation on load")
	require.True(t, b.swiftSyncEnabled,
		"node keeps running; no fail-closed on a bad persisted aggregate")

	// The boundary block is already the tip, so IsZero() never runs again and
	// the next block synced takes the normal path over the unverified UTXO set.
	require.False(t, b.isSwiftSyncActive(boundary+1),
		"sync resumes normally over the unverified UTXO set")
}
```

</details>

server.go
GeorgeTsagk, line 2946: Empty `HintsHash` skips the only pre-trust integrity check and accepts any file. Shipped networks all set it, so latent, but since the aggregate doesn't fail closed per the chain.go comment, the hash pin is the sole backstop. Suggest making an empty `HintsHash` while `SwiftSync != nil` a hard config error.

GeorgeTsagk (changes requested):
looks good, will run some mainnet testing soon

server.go
```diff
@@ -2859,6 +2864,157 @@ func setupRPCListeners() ([]net.Listener, error) {
...
+		got := fmt.Sprintf("%x", sha256.Sum256(compressed))
+		if !strings.EqualFold(got, cfg.HintsHash) {
+			return nil, fmt.Errorf("swift sync hintsfile %q has hash %s, "+
```
allocz, line 2949: I had a problem here because we are just logging this instead of stopping btcd. I've generated the hint for height 935000 on mainnet, but obviously the hash is not what we have hardcoded in `chainconfig/params.go`, and the height is not before the latest checkpoint. I've not seen the log message in the first IBD and thought that the hint was being used for IBD. If the node stopped instead, I would have figured this out earlier.

What do you think about stopping the node if the hint hash does not match the hardcoded one?

allocz:
Executed an IBD on mainnet here, from height 0...930000. Pretty good results as the benchmark machine is a 3rd gen Core i5 with 8GB of DDR3 ram and a SATA SSD, and was also running two btcd instances, the full node with the blocks in an HDD and the swiftsync enabled node performing the IBD. I created this setup to avoid network bottlenecks in the results.

The following are the stats of the IBD.
```
$ pstat -h /tmp/btcdswiftsync-perfstat --stopatheight=930000 --prune=2048 \
    --datadir=${HOME}/.local/share/btcdswiftsync_930000_prune2048 \
    --connect=127.0.0.1:8333 \
    --rpclisten=127.0.0.1:8335 \
    --swiftsynchints=/mnt/btcd-pool/.btcd/mainnet/sshints.dat \
    | tee /tmp/btcdswiftsync-perfstat.log

Resident Mem Max 1.494GiB
Virtual Mem Max  3.583GiB
Elapsed          12H 58m 33s 5ms
User Time        19H 45m 8s 330ms
System Time      55m 36s 710ms 0μs
Read Char        1.194TiB
Write Char       1.219TiB
Syscall Read     111.279M
Syscall Write    117.416M
Read Bytes       502.840GiB
Write Bytes      1.220TiB
```

About hint generation, currently, we create a new database of significant size when generating the hints, but to avoid increasing the storage requirements would be nice to have an option to generate the hints without creating the database, even if the process can not be stopped before completed, maybe enabling this mode with a flag `--nointermediarydb`, or something like this. Also, as I've generated the hints in an HDD, the process took several days because of I/O bottlenecks, I believe.

allocz (commented):
SwiftSync IBD height 0..930000 with data on HDD:
```
$ pstat -h /tmp/btcdswiftsync-perfstat --stopatheight=930000 --prune=2048 \                                                                                                                                            --datadir=/mnt/pool600/btcdswiftsync_930000_prune2048 \                                                                                                                                                            --connect=127.0.0.1:8333 --rpclisten=127.0.0.1:8335 \
    --swiftsynchints=/mnt/btcd-pool/.btcd/mainnet/sshints.dat \
    | tee /tmp/btcdswiftsync-perfstat-0-930000-hdd.log

Resident Mem Max 1.486GiB
Virtual Mem Max  3.583GiB
Elapsed          13H 17m 8s 6ms
User Time        19H 24m 10s 900ms
System Time      58m 30s 480ms 0μs
Read Char        976.983GiB
Write Char       998.472GiB
Syscall Read     61.657M
Syscall Write    67.378M
Read Bytes       429.772GiB
Write Bytes      998.446GiB
```

Nice results for HDD initial block download, less than 14H for IBD from height 0 to 930000. Less than
500GiB of disk read and less than 1TiB of disk writes.

PR #2480, opened 2026-02-12
server: fix peer add/done race between peerHandler and syncManager

Aharonee:
## Summary

Fix a race condition where the sync manager can permanently get stuck with a dead sync peer after rapid peer connect/disconnect cycles.

## The Race Condition

`peerDoneHandler` ran as a **separate goroutine per peer** and independently notified two event loops about a disconnect:

1. It sent to the `donePeers` channel (consumed by `peerHandler`).
2. It called `syncManager.DonePeer()` directly (sends to `sm.msgChan`, consumed by `blockHandler`).

Meanwhile, `peerHandler` only called `syncManager.NewPeer()` when it processed the `newPeers` channel. Because these two paths were unsynchronized, `blockHandler` could observe `DonePeer` before `NewPeer` for the same peer.

A second vector existed even if `DonePeer` were moved into `peerHandler`: two separate buffered channels (`newPeers`/`donePeers`) let Go's `select` randomly pick the done case before the add case when both were ready simultaneously.

A third vector existed due to `negotiateTimeout`: if the 30s timeout in `peer.Peer.start()` fired between `verAckReceived = true` and the `OnVerAck` callback completing, `peerDoneHandler` could observe `VerAckReceived() == true` and send `peerDone` before the `OnVerAck` callback sent `peerAdd`.

**Consequences:** The sync manager receives `DonePeer` for an unknown peer (logged as a warning, no cleanup). Then `NewPeer` arrives for the already-dead peer -- the sync manager registers it as a candidate and potentially selects it as `syncPeer`. Since it is already disconnected, no subsequent `DonePeer` arrives to clear it. The node is stuck: it believes it has a sync peer, ignores new candidates, and never makes chain progress.

## What Triggers It

Any scenario that produces rapid connect/disconnect cycles:
- Attacker traffic (connections that complete the version/verack handshake then immediately drop)
- Flaky network conditions with many short-lived peers
- High peer churn under load (e.g., maxpeers limit causing immediate disconnects)

## The Fix

Three structural changes eliminate all race vectors:

1. **Merge `newPeers` and `donePeers` into a single `peerLifecycle` channel.** A single FIFO channel eliminates the select-ambiguity vector where Go's `select` could pick done before add.

2. **Move `syncManager.DonePeer()` and orphan eviction into `handleDonePeerMsg`.** All sync manager notifications now flow through the `peerHandler` goroutine.

3. **Make `peerLifecycleHandler` (renamed from `peerDoneHandler`) the sole sender of both `peerAdd` and `peerDone` for each peer.** `OnVerAck` no longer sends to the channel directly; it closes a signal channel (`verAckCh`). `peerLifecycleHandler` selects on `verAckCh` vs `peer.Done()` (new method exposing the peer's quit channel), sends `peerAdd` if verack was received, then waits for disconnect and sends `peerDone`. Because both sends originate from the same goroutine, `peerAdd` is never enqueued *after* `peerDone`. If both channels are ready simultaneously, Go's `select` is nondeterministic so `peerAdd` may be skipped -- this is harmless (the peer is already disconnected) and documented in the `peerLifecycleEvent` comment.

Additionally, `OnVerAck` is guarded against double-close: if called more than once on the same peer, it logs an error instead of panicking.

## Reproducing on `master` (without the fix)

The included integration test can demonstrate the corruption on an unpatched `master` branch:

```bash
git checkout master
git checkout bugfix/peer_race_condition -- integration/sync_race_test.go
go test -tags=rpctest -v -run TestSyncManagerRaceCorruption ./integration/ -count=10 -timeout 900s
```

## Test Plan

- [x] `go build ./...` compiles cleanly
- [x] `go test -tags=rpctest -v -run TestSyncManagerRaceCorruption ./integration/ -count=10 -timeout 900s` passes
- [x] `TestPreVerackDisconnect` passes (disconnect before verack)
- [x] Existing integration tests unaffected
- [x] `TestOnVerAckDoubleCall` -- unit test calling `OnVerAck` twice on the same `serverPeer`, asserting no panic and `verAckCh` remains closed
- [x] `TestPeerLifecycleOrdering` -- unit test asserting verack-then-disconnect emits `peerAdd` followed by `peerDone`
- [x] `TestPeerLifecycleSimultaneousReady` -- unit test asserting stable behavior when both `verAckCh` and `Peer.Done()` are ready simultaneously (100 iterations, verifying ordering invariant holds regardless of which `select` case wins)

### DIFF
--- a/integration/sync_race_test.go
+++ b/integration/sync_race_test.go
@@ -23,10 +23,12 @@ const (
 	syncRaceProofWait   = 8 * time.Second
 )
 
-// fakePeerConn connects to the node at nodeAddr, performs the minimum version/verack
-// handshake so the node registers a peer (NewPeer) and then disconnects so the node
-// runs DonePeer. This simulates attacker traffic: many connections that complete
-// handshake then drop, stressing the sync manager's ordering of NewPeer/DonePeer.
+// fakePeerConn connects to the node at nodeAddr, performs the
+// minimum version/verack handshake so the node registers a peer
+// (NewPeer) and then disconnects so the node runs DonePeer. This
+// simulates attacker traffic: many connections that complete the
+// handshake then drop, stressing the sync manager's ordering of
+// NewPeer/DonePeer.
 func fakePeerConn(nodeAddr string) error {
 	conn, err := net.DialTimeout("tcp", nodeAddr, 5*time.Second)
 	if err != nil {
@@ -54,7 +56,10 @@ func fakePeerConn(nodeAddr string) error {
 	msgVersion := wire.NewMsgVersion(me, you, nonce, 0)
 	msgVersion.Services = wire.SFNodeNetwork | wire.SFNodeWitness
 
-	if err := wire.WriteMessage(conn, msgVersion, wire.ProtocolVersion, wire.SimNet); err != nil {
+	err = wire.WriteMessage(
+		conn, msgVersion, wire.ProtocolVersion, wire.SimNet,
+	)
+	if err != nil {
 		return err
 	}
 
@@ -66,7 +71,11 @@ func fakePeerConn(nodeAddr string) error {
 		switch msg.(type) {
 		case *wire.MsgVersion:
 			// Node's version; send verack.
-			if err := wire.WriteMessage(conn, wire.NewMsgVerAck(), wire.ProtocolVersion, wire.SimNet); err != nil {
+			err := wire.WriteMessage(
+				conn, wire.NewMsgVerAck(),
+				wire.ProtocolVersion, wire.SimNet,
+			)
+			if err != nil {
 				return err
 			}
 
@@ -83,12 +92,12 @@ func fakePeerConn(nodeAddr string) error {
 	}
 }
 
-// TestSyncManagerRaceCorruption stresses a single simnet node with many inbound
-// connections that complete the version/verack handshake then disconnect. It then
-// proves corruption without a dedicated RPC: connect a fresh node that generates
-// blocks; if the stressed node does not sync, it was stuck with a dead sync peer
-// (getpeerinfo returns 0 peers when all disconnected; in the corrupted state the
-// sync manager still has a dead peer as sync peer, so it ignores the new live one).
+// TestSyncManagerRaceCorruption stresses a single simnet node
+// with many inbound connections that complete the version/verack
+// handshake then disconnect. It then proves corruption: connect a
+// fresh node that generates blocks; if the stressed node does not
+// sync, it was stuck with a dead sync peer (the sync manager still
+// has a dead peer as sync peer, so it ignores the new live one).
 func TestSyncManagerRaceCorruption(t *testing.T) {
 	stressedHarness, err := rpctest.New(&chaincfg.SimNetParams, nil, nil, "")
 	require.NoError(t, err)
@@ -117,8 +126,9 @@ func TestSyncManagerRaceCorruption(t *testing.T) {
 		iter += syncRaceConcurrency
 	}
 
-	// Prove corruption: connect a live node and generate blocks.
-	// If the stressed node was corrupted (dead sync peer, 0 connected peers per getpeerinfo), it will not sync from the new one.
+	// Prove corruption: connect a live node and generate blocks. If
+	// the stressed node was corrupted (dead sync peer, 0 connected
+	// peers), it will not sync from the new one.
 	newHarness, err := rpctest.New(&chaincfg.SimNetParams, nil, nil, "")
 	require.NoError(t, err)
 	require.NoError(t, newHarness.SetUp(true, 0))
@@ -138,18 +148,23 @@ func TestSyncManagerRaceCorruption(t *testing.T) {
 	_, heightAfter, err := stressedHarness.Client.GetBestBlock()
 	require.NoError(t, err)
 
-	if heightAfter < heightBefore+int32(syncRaceProofBlocks) {
-		t.Fatalf("proved sync manager corruption after %d fake peer cycles: stressed node did not sync from new peer (height %d -> %d); it was stuck with a dead sync peer instead of the new live one",
-			done, heightBefore, heightAfter)
-	}
+	expected := heightBefore + int32(syncRaceProofBlocks)
+	require.GreaterOrEqualf(t, heightAfter, expected,
+		"sync manager corruption after %d fake "+
+			"peer cycles: node stuck with dead "+
+			"sync peer (height %d -> %d)",
+		done, heightBefore, heightAfter)
 
-	t.Logf("completed %d fake peer cycles; stressed node synced from new peer (height %d -> %d), no corruption observed", done, heightBefore, heightAfter)
+	t.Logf("completed %d fake peer cycles; "+
+		"node synced (height %d -> %d)",
+		done, heightBefore, heightAfter)
 }
 
-// TestPreVerackDisconnect verifies that a peer disconnecting before completing
-// the version/verack handshake does not corrupt the sync manager state. In this
-// case only a peerDone event is produced (no peerAdd), since AddPeer is only
-// called from OnVerAck. The node must remain healthy and able to sync afterward.
+// TestPreVerackDisconnect verifies that a peer disconnecting
+// before completing the version/verack handshake does not corrupt
+// the sync manager state. In this case only a peerDone event is
+// produced (no peerAdd), since peerLifecycleHandler only sends
+// peerAdd when verAckCh is closed. The node must remain healthy.
 func TestPreVerackDisconnect(t *testing.T) {
 	harness, err := rpctest.New(&chaincfg.SimNetParams, nil, nil, "")
 	require.NoError(t, err)
--- a/integration/sync_race_test.go
+++ b/integration/sync_race_test.go
@@ -110,17 +110,16 @@ func TestSyncManagerRaceCorruption(t *testing.T) {
 	deadline := time.Now().Add(syncRaceRunDuration)
 	iter := 0
 	var done int
-	doneCh := make(chan struct{}, syncRaceConcurrency*2)
+	errCh := make(chan error, syncRaceConcurrency*2)
 
 	for time.Now().Before(deadline) && iter < syncRaceIterations {
 		for i := 0; i < syncRaceConcurrency; i++ {
 			go func() {
-				_ = fakePeerConn(nodeAddr)
-				doneCh <- struct{}{}
+				errCh <- fakePeerConn(nodeAddr)
 			}()
 		}
 		for i := 0; i < syncRaceConcurrency; i++ {
-			<-doneCh
+			require.NoError(t, <-errCh)
 			done++
 		}
 		iter += syncRaceConcurrency
@@ -160,6 +159,45 @@ func TestSyncManagerRaceCorruption(t *testing.T) {
 		done, heightBefore, heightAfter)
 }
 
+// dialAndSendVersion connects to nodeAddr and sends a version
+// message, returning the open connection. The caller is
+// responsible for closing it.
+func dialAndSendVersion(
+	t *testing.T, nodeAddr string,
+) net.Conn {
+
+	t.Helper()
+
+	conn, err := net.DialTimeout("tcp", nodeAddr, 5*time.Second)
+	require.NoError(t, err)
+
+	_ = conn.SetDeadline(time.Now().Add(5 * time.Second))
+
+	nodeTCP, err := net.ResolveTCPAddr("tcp", nodeAddr)
+	require.NoError(t, err)
+
+	you := wire.NewNetAddress(
+		nodeTCP, wire.SFNodeNetwork|wire.SFNodeWitness,
+	)
+	me := wire.NewNetAddress(
+		&net.TCPAddr{IP: net.IPv4(127, 0, 0, 1), Port: 0},
+		wire.SFNodeNetwork|wire.SFNodeWitness,
+	)
+	you.Timestamp = time.Time{}
+	me.Timestamp = time.Time{}
+
+	nonce := uint64(rand.Int63())
+	msgVersion := wire.NewMsgVersion(me, you, nonce, 0)
+	msgVersion.Services = wire.SFNodeNetwork | wire.SFNodeWitness
+
+	err = wire.WriteMessage(
+		conn, msgVersion, wire.ProtocolVersion, wire.SimNet,
+	)
+	require.NoError(t, err)
+
+	return conn
+}
+
 // TestPreVerackDisconnect verifies that a peer disconnecting
 // before completing the version/verack handshake does not corrupt
 // the sync manager state. In this case only a peerDone event is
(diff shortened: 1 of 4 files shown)


naorye:
I experience the same issue.. Wow, really need this.

server.go
```diff
@@ -2348,13 +2365,14 @@ func (s *server) peerHandler() {
...
-			s.handleAddPeerMsg(state, p)
-
-		// Disconnected peers.
-		case p := <-s.donePeers:
-			s.handleDonePeerMsg(state, p)
+		// Peer connected or disconnected.
+		case event := <-s.peerLifecycle:
+			switch event.action {
+			case peerAdd:
+				s.handleAddPeerMsg(state, event.sp)
```
starius, line 2417: I propose to add an empty line between `case`s for readability.

integration/sync_race_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/integration/rpctest"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	syncRaceIterations  = 1000
+	syncRaceConcurrency = 50
```
starius: I propose to increase `syncRaceConcurrency` to 300 so it could overflow the buffer of `peerLifecycle` (at least in theory) and we could test how the fix works under backpressure.

The default MaxPeers is 125 and capacity of peerLifecycle is `cfg.MaxPeers*2 = 250`.

integration/sync_race_test.go
```diff
@@ -0,0 +1,139 @@
...
+				return err
+			}
+		case *wire.MsgSendAddrV2:
+			// Optional; keep reading.
+		case *wire.MsgVerAck:
+			// Handshake complete; close to trigger DonePeer.
+			return nil
+		default:
+			// Ignore other messages (e.g. wtxidrelay) and keep reading.
+		}
```
starius, line 91: Please add empty lines between `case`s.

server.go
```diff
@@ -151,6 +151,23 @@ type updatePeerHeightsMsg struct {
...
+// peerLifecycleAction describes the type of peer lifecycle event.
+type peerLifecycleAction uint8
+
+const (
+	peerAdd  peerLifecycleAction = iota
+	peerDone
+)
+
+// peerLifecycleEvent represents a peer connection or disconnection event.
+// Using a single channel for both event types guarantees FIFO ordering:
```
starius: Do we have the "first-in" part? Can OnVerAck be delayed and send its part after "done" event is sent? E.g. if `OnVerAck` runs longer than `negotiateTimeout`.

integration/sync_race_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+	_, err = newHarness.Client.Generate(syncRaceProofBlocks)
+	require.NoError(t, err)
+
+	time.Sleep(syncRaceProofWait)
+
+	_, heightAfter, err := stressedHarness.Client.GetBestBlock()
+	require.NoError(t, err)
+
+	if heightAfter < heightBefore+int32(syncRaceProofBlocks) {
```
starius: Use [require.Lessf](https://pkg.go.dev/github.com/stretchr/testify/require#Lessf)

integration/sync_race_test.go
```diff
@@ -0,0 +1,139 @@
...
+	syncRaceConcurrency = 50
+	syncRaceRunDuration = 90 * time.Second
+	syncRaceProofBlocks = 5
+	syncRaceProofWait   = 8 * time.Second
+)
+
+// fakePeerConn connects to the node at nodeAddr, performs the minimum version/verack
+// handshake so the node registers a peer (NewPeer) and then disconnects so the node
+// runs DonePeer. This simulates attacker traffic: many connections that complete
+// handshake then drop, stressing the sync manager's ordering of NewPeer/DonePeer.
```
starius: There are many lines in this files longer than 80 chars. Could you format the file to fit into 80 chars limit, please? A tab is counted as 8 spaces.

integration/sync_race_test.go
```diff
@@ -0,0 +1,139 @@
...
+	defer conn.Close()
+
+	_ = conn.SetDeadline(time.Now().Add(15 * time.Second))
+
+	nodeTCP, err := net.ResolveTCPAddr("tcp", nodeAddr)
+	if err != nil {
+		return err
+	}
+	you := wire.NewNetAddress(nodeTCP, wire.SFNodeNetwork|wire.SFNodeWitness)
+	me := wire.NewNetAddress(&net.TCPAddr{IP: net.IPv4(127, 0, 0, 1), Port: 0}, wire.SFNodeNetwork|wire.SFNodeWitness)
```
starius: Please fix the formatting of long function calls:

```go
me := wire.NewNetAddress(
	&net.TCPAddr{IP: net.IPv4(127, 0, 0, 1), Port: 0},
	wire.SFNodeNetwork|wire.SFNodeWitness,
)
```

here and below in the file.

integration/sync_race_test.go
```diff
@@ -0,0 +1,139 @@
...
+	_, heightAfter, err := stressedHarness.Client.GetBestBlock()
+	require.NoError(t, err)
+
+	if heightAfter < heightBefore+int32(syncRaceProofBlocks) {
+		t.Fatalf("proved sync manager corruption after %d fake peer cycles: stressed node did not sync from new peer (height %d -> %d); it was stuck with a dead sync peer instead of the new live one",
+			done, heightBefore, heightAfter)
+	}
+
+	t.Logf("completed %d fake peer cycles; stressed node synced from new peer (height %d -> %d), no corruption observed", done, heightBefore, heightAfter)
+}
```
starius, line 160: Can we add another test: disconnect prior to verack? It should result in just "done" event, without corresponding "add", which is fine, but it is worth testing it.

server.go
Aharonee: Good catch, there seems to still be a potential race in that scenario.
I've pushed a commit which changes the `peerDoneHandler` into `peerLifecycleHandler`, and delegates responsibility for both add peer and done peer events to it. 
That way a single goroutine will manage synchronization and correct ordering of the peer lifecycle events.

Does that make sense?

server.go
```diff
@@ -2304,25 +2304,45 @@ func (s *server) outboundPeerConnected(c *connmgr.ConnReq, conn net.Conn) {
...
+	// disconnect, whichever comes first.
+	select {
+	case <-sp.verAckCh:
+		s.peerLifecycle <- peerLifecycleEvent{
+			action: peerAdd, sp: sp,
+		}
+
+	case <-sp.Peer.Done():
+		// Disconnected before verack; no peerAdd needed.
+	}
```
starius, line 2352: If both `sp.verAckCh` and `sp.Peer.Done()` have messages to receive, `select` chooses pseudorandomly among them. So `peerAdd` can be skipped even if `VerAckReceived` is true, and `handleDonePeerMsg` will call DonePeer for an unknown peer.

Does it make sense to prioritize receiving from `sp.verAckCh` or check `VerAckReceived` if `sp.Peer.Done()` fired?

server.go
```diff
@@ -155,14 +155,14 @@ type updatePeerHeightsMsg struct {
 type peerLifecycleAction uint8
 
 const (
-	peerAdd  peerLifecycleAction = iota
+	peerAdd peerLifecycleAction = iota
```
starius, line 158: this formatting change should belong to the first commit

server.go
```diff
@@ -551,10 +553,11 @@ func (sp *serverPeer) OnVersion(_ *peer.Peer, msg *wire.MsgVersion) *wire.MsgRej
...
 
-// OnVerAck is invoked when a peer receives a verack bitcoin message and is used
-// to kick start communication with them.
+// OnVerAck is invoked when a peer receives a verack bitcoin message.
+// It signals the peer's lifecycle handler that the handshake is
+// complete so it can register the peer with the server.
 func (sp *serverPeer) OnVerAck(_ *peer.Peer, _ *wire.MsgVerAck) {
-	sp.server.AddPeer(sp)
+	close(sp.verAckCh)
 }
```
starius: In the current code version allows calling `OnVerAck` only once. Should we safeguard for the future using sync.Once?

server.go
Aharonee: I think it should be fine to skip add peer if done peer event has already occurred.
After all, the peer has disconnected so we can avoid notifying the server of a new peer just to notify it right away after to remove it.

My main concern was done peer being processed before add peer, but done peer processing for an unknown peer that has already disconnected seems harmless.

server.go
Aharonee: I can squash both commits and force push if you prefer, but wouldn't it be more convenient for you to review the diff each time and only squash merge at the end?

server.go
Aharonee: I think safeguarding this could potentially hide a bug, and if it is called twice we would prefer a loud panic.
This pattern is also a consistent pattern used in the codebase, for example: `Peer.quit` channel is not safeguarded and is closed by `Peer.Disconnect()`.

starius: Hmm, maybe we can produce an error instead, if it is closed already?

```go
select {
case <-sp.verAckCh:
 log Error

default:
 close(sp.verAckCh)
}
```

The error won't let it pass unnoticed, but at least it won't panic and crash. What do you think?

server.go
starius: Sure, let's keep them separate for now.

server.go
starius: You are right. My proposal would only improve log message clarity (avoiding "unknown peer" being logged), not the correctness of the code itself. It is optional.

server.go
```diff
@@ -151,6 +151,23 @@ type updatePeerHeightsMsg struct {
...
+
+const (
+	peerAdd peerLifecycleAction = iota
+	peerDone
+)
+
+// peerLifecycleEvent represents a peer connection or disconnection
+// event. Both event types for a given peer are sent by a single
+// goroutine (peerLifecycleHandler), guaranteeing that peerAdd is
+// always enqueued before peerDone.
```
starius: I propose to adjust the comment to reflect that peerAdd may be skipped.

server.go
```diff
@@ -279,6 +295,7 @@ type serverPeer struct {
 	knownAddresses lru.Cache
 	banScore       connmgr.DynamicBanScore
 	quit           chan struct{}
+	verAckCh       chan struct{} // closed when OnVerAck fires
```
starius: Formatting:

```go
// Closed when OnVerAck fires.
verAckCh chan struct{}
```

server.go
starius: This change looks good to me!

server.go
Aharonee: Sure, makes sense

server.go
```diff
@@ -557,7 +560,13 @@ func (sp *serverPeer) OnVersion(_ *peer.Peer, msg *wire.MsgVersion) *wire.MsgRej
...
 // complete so it can register the peer with the server.
 func (sp *serverPeer) OnVerAck(_ *peer.Peer, _ *wire.MsgVerAck) {
-	close(sp.verAckCh)
+	select {
+	case <-sp.verAckCh:
+		peerLog.Errorf("OnVerAck called more than once "+
+			"for peer %v", sp)
+	default:
+		close(sp.verAckCh)
+	}
```
starius: I propose to cover this with a direct test (not rpctest) to make sure that it works as expected (logs, not panics by closing the channel again). To be clear, the code is correct, but it is better to cover such fragile things explicitly, IMHO. Specifically: call OnVerAck twice on the same serverPeer and assert no panic + verAckCh remains closed.

server.go
starius: Note that the code no longer matches the commit message. The commit says: "Prioritize verAckCh in peerLifecycleHandler select to avoid nondeterministic peerAdd skipping when both channels are ready." However, there are no changes in `peerLifecycleHandler`'s `select` logic.

I suggest implementing prioritization exactly as described in the commit message: first attempt a non-blocking receive from `sp.verAckCh`, and only in the `default` case fall back to the two-channel `select`. That would bring the code in line with the commit message.

server.go
```diff
@@ -2273,38 +2313,44 @@ func (s *server) outboundPeerConnected(c *connmgr.ConnReq, conn net.Conn) {
...
-// peerDoneHandler handles peer disconnects by notifying the server that it's
-// done along with other performing other desirable cleanup.
-func (s *server) peerDoneHandler(sp *serverPeer) {
+// peerLifecycleHandler is the sole sender of lifecycle events for a
+// given peer. It waits for either verack (handshake complete) or
+// disconnect (handshake failed/timed out), sends peerAdd if verack
+// was received, then waits for disconnect and sends peerDone.
+// Because both sends originate from this single goroutine,
+// peerAdd is always enqueued before peerDone.
+func (s *server) peerLifecycleHandler(sp *serverPeer) {
```
starius, line 2340: Please add a deterministic unit test for `peerLifecycleHandler` where verack arrives before disconnect, and assert peerLifecycle emits peerAdd first and peerDone second.

We don't have internal observability in integration/sync_race_test.go since it runs against a separate btcd process, so we can add a direct regression test `server` to cover the fix.

The test has to build before the fix is applied, but fail because of that race that we're fixing.

integration/sync_race_test.go
```diff
@@ -0,0 +1,243 @@
...
+	nodeAddr := stressedHarness.P2PAddress()
+	deadline := time.Now().Add(syncRaceRunDuration)
+	iter := 0
+	var done int
+	doneCh := make(chan struct{}, syncRaceConcurrency*2)
+
+	for time.Now().Before(deadline) && iter < syncRaceIterations {
+		for i := 0; i < syncRaceConcurrency; i++ {
+			go func() {
+				_ = fakePeerConn(nodeAddr)
```
starius: We should check this error and fail the test if it fails.

`fakePeerConn` errors are ignored and the test increments `done` regardless. On constrained/slow environments, many connection attempts can fail and the test may not actually execute enough handshake+disconnect cycles to stress the race as intended.

integration/sync_race_test.go
```diff
@@ -0,0 +1,243 @@
...
+
+	nodeAddr := harness.P2PAddress()
+
+	// Connect and send version, then disconnect before receiving or
+	// sending verack. This produces a peerDone without a preceding
+	// peerAdd in the lifecycle channel.
+	for i := 0; i < 50; i++ {
+		conn, err := net.DialTimeout("tcp", nodeAddr, 5*time.Second)
+		if err != nil {
+			continue
```
starius: This error also should not be ignored. This may result in the test passing without meaningful stress.

integration/sync_race_test.go
```diff
@@ -0,0 +1,243 @@
...
+		}
+
+		_ = conn.SetDeadline(time.Now().Add(5 * time.Second))
+
+		nodeTCP, err := net.ResolveTCPAddr("tcp", nodeAddr)
+		if err != nil {
+			conn.Close()
+			continue
```
starius: This error also should not be ignored. This may result in the test passing without meaningful stress.

integration/sync_race_test.go
```diff
@@ -0,0 +1,243 @@
...
+			wire.SFNodeNetwork|wire.SFNodeWitness,
+		)
+		you.Timestamp = time.Time{}
+		me.Timestamp = time.Time{}
+
+		nonce := uint64(rand.Int63())
+		msgVersion := wire.NewMsgVersion(me, you, nonce, 0)
+		msgVersion.Services = wire.SFNodeNetwork | wire.SFNodeWitness
+
+		_ = wire.WriteMessage(
```
starius: This error also should not be ignored. This may result in the test passing without meaningful stress.

integration/sync_race_test.go
```diff
@@ -0,0 +1,243 @@
...
+	harness, err := rpctest.New(&chaincfg.SimNetParams, nil, nil, "")
+	require.NoError(t, err)
+	require.NoError(t, harness.SetUp(true, 0))
+	t.Cleanup(func() { _ = harness.TearDown() })
+
+	nodeAddr := harness.P2PAddress()
+
+	// Connect and send version, then disconnect before receiving or
+	// sending verack. This produces a peerDone without a preceding
+	// peerAdd in the lifecycle channel.
```
starius: s/produces/is expected to produce/

We do not validate the internal state of the running btcd process, so this is only our assumption, not an actual verification.

starius (approved):
**LGTM!** 💾

The fix is currently validated only by integration tests. They cover the main failure mode well, but there are still unit-level coverage gaps that should be addressed with direct regression tests:

- Add an `OnVerAck` idempotency test for the double-call guard (`case <-sp.verAckCh`): call `OnVerAck` twice on the same `serverPeer`, assert no panic on the second call, and assert `verAckCh` remains closed.
- Add a deterministic `peerLifecycleHandler` test for the simultaneous-ready edge case: make both `verAckCh` and `Peer.Done()` ready at the same time, then assert the intended behavior is stable and documented.
- Add a direct per-peer lifecycle ordering test: in the verack-then-disconnect path, assert the emitted events for the same peer are exactly `peerAdd` followed by `peerDone`.

I suggest adding these as regression tests that pass with the fix and fail if any part of the fix is rolled back. That ensures the core invariants are verified directly.

allocz:
Diff looks nice, ran the entire test suite, all green. Nice work.

Roasbeef (approved):
LGTM 🫐

PR #2483, opened 2026-02-14
docs: add PSBT workflow documentation and examples

### LINKED ISSUE #1988: Signing PSBT
Hi, thank you for your work. Due to the lack of examples, I have a question, how can I add FinalScriptWitness to PSBT? The thing is, I have two base64 encoded PSBT: signed and unsinged. The only difference in them is that the one signed by the wallet (manually) has a FinalScriptWitness parameter. However, I have not been able to reproduce it. It is also worth noting that the signature was made with a Taproot address.

studentyashsrivastava:
Document the PSBT workflow in btcd and clarify how FinalScriptSig and
FinalScriptWitness are populated during finalization.

This adds:
- Expanded package documentation describing the lifecycle of a PSBT
- Runnable examples for creation, decoding and finalized inputs

Fixes #1988.

### DIFF
--- a/btcutil/psbt/example_test.go
+++ b/btcutil/psbt/example_test.go
@@ -1,0 +1,164 @@
+// Copyright (c) 2018 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package psbt_test
+
+import (
+	"bytes"
+	"encoding/hex"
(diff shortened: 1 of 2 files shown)


studentyashsrivastava:
please tell if anything is wrong or any suggestions i am new in this and wanted to contribute

PR #2485, opened 2026-02-17
txscript: reject OP_CODESEPARATOR in unexecuted branches for non-segwit

Roasbeef:
In this commit, we fix a policy-level divergence with Bitcoin Core when handling OP_CODESEPARATOR inside unexecuted OP_IF branches in non-segwit scripts with the ScriptVerifyConstScriptCode flag.

Bitcoin Core's EvalScript (interpreter.cpp:474-476) places the SCRIPT_VERIFY_CONST_SCRIPTCODE check for OP_CODESEPARATOR before the fExec branch-execution gate, causing it to fire unconditionally on every OP_CODESEPARATOR encountered during script iteration -- even inside OP_FALSE OP_IF ... OP_ENDIF envelopes.

Previously, btcd's equivalent check lived inside the opcodeCodeSeparator handler, which was never reached for opcodes in unexecuted branches due to the early return in executeOpcode that skips non-conditional opcodes when isBranchExecuting() is false. This meant a script like:

  OP_FALSE OP_IF OP_CODESEPARATOR OP_ENDIF <validation>

would be rejected by Bitcoin Core's mempool but accepted by btcd's.

The fix moves the check before the branch-execution gate in executeOpcode, matching Bitcoin Core's structure. This follows the existing pattern in btcd where isOpcodeDisabled and isOpcodeAlwaysIllegal checks already fire regardless of branch execution state.

Note: SCRIPT_VERIFY_CONST_SCRIPTCODE is purely a policy flag (included in STANDARD_SCRIPT_VERIFY_FLAGS but not MANDATORY_SCRIPT_VERIFY_FLAGS), so this was not a consensus divergence. Both implementations would accept such transactions if mined in a block.

Found via differential fuzzing by Bruno from bitcoinfuzz.


brunoerg (approved):
ACK 3e17ea43c80e6a3f7b49a2512e039886fc6bd4c0

saubyk:
cc: @gijswijs for review

txscript/engine.go
```diff
@@ -485,6 +485,18 @@ func (vm *Engine) executeOpcode(op *opcode, data []byte) error {
 		return scriptError(ErrElementTooBig, str)
 	}
 
+	// With ScriptVerifyConstScriptCode, OP_CODESEPARATOR in non-segwit
+	// script is rejected even in an unexecuted branch. This mirrors
+	// Bitcoin Core's behavior where the check fires unconditionally
+	// before the branch execution gate.
+	if op.value == OP_CODESEPARATOR && vm.taprootCtx == nil &&
```
kcalvinalvin, line 500: Bit weird that we check for non-segwitness by asserting witnessProgram and taprootCtx is nil but this works.

kcalvinalvin:
Checked against Core at commit `ff7cdf633e375f151cccbcc78c7add161b3d29b8` and verified that the behavior matches.

But can we add a segwit case? It's not tested here and would be good to have. Something like this maybe:

```diff
diff --git a/txscript/engine_test.go b/txscript/engine_test.go
index dea9f2f7..311a9038 100644
--- a/txscript/engine_test.go
+++ b/txscript/engine_test.go
@@ -6,6 +6,7 @@
 package txscript
 
 import (
+       "crypto/sha256"
        "testing"
 
        "github.com/btcsuite/btcd/chaincfg/chainhash"
@@ -466,6 +467,7 @@ func TestCodeSepUnexecutedBranch(t *testing.T) {
                name    string
                script  string
                flags   ScriptFlags
+               segwit  bool
                wantErr bool
                errCode ErrorCode
        }{
@@ -503,13 +505,37 @@ func TestCodeSepUnexecutedBranch(t *testing.T) {
                        wantErr: true,
                        errCode: ErrCodeSeparator,
                },
+               {
+                       // OP_CODESEPARATOR in a segwit P2WSH witness script
+                       // should succeed even with const scriptcode, since
+                       // the flag only applies to non-segwit scripts.
+                       name:    "codesep in segwit P2WSH with const scriptcode",
+                       script:  "CODESEPARATOR TRUE",
+                       flags: ScriptVerifyConstScriptCode | ScriptVerifyWitness | ScriptBip16,
+                       segwit:  true,
+                       wantErr: false,
+               },
        }
 
        for _, test := range tests {
                t.Run(test.name, func(t *testing.T) {
-                       pkScript := mustParseShortForm(test.script)
+                       scriptBytes := mustParseShortForm(test.script)
+
+                       pkScript := scriptBytes
+                       testTx := tx
+                       if test.segwit {
+                               hash := sha256.Sum256(scriptBytes)
+                               pkScript, _ = NewScriptBuilder().
+                                       AddOp(OP_0).AddData(hash[:]).Script()
+                               testTx = tx.Copy()
+                               testTx.TxIn[0].SignatureScript = nil
+                               testTx.TxIn[0].Witness = wire.TxWitness{
+                                       scriptBytes,
+                               }
+                       }
+
                        vm, err := NewEngine(
-                               pkScript, tx, 0, test.flags, nil, nil,
+                               pkScript, testTx, 0, test.flags, nil, nil,
                                -1, nil,
                        )
                        if err != nil {
```

Roasbeef:
@kcalvinalvin  PTAL. 

One other thing we can do is remove the old check. Since we know check it each time, we no longer need that logic within the op code invocation itself.

kcalvinalvin:
Nothing changed since I reviewed it..? Not sure what I should look at.

I mean the code change itself is still ok. I just think it'd be good to have the non segwit case

allocz:
@kcalvinalvin 

I agree with the concept of adding a test to cover `OP_CODESEPARATOR` on segwit, but maybe instead of adding custom behavior in the test execution loop just for the segwit variant of this test wouldn't be cleaner to add another test `TestCodeSepExecutedBranchSegwit`?

txscript/engine.go
```diff
@@ -485,6 +485,18 @@ func (vm *Engine) executeOpcode(op *opcode, data []byte) error {
 		return scriptError(ErrElementTooBig, str)
 	}
```
gijswijs, line 487: The duplicate check in `opcodeCodeSeparator` (line 1956-1961) is now dead code.

You could argue it's defensive programming, but either remove it for clarity, or add a comment explaining it's intentionally kept as a belt-and-suspenders measure.

txscript/engine_test.go
```diff
@@ -426,3 +426,113 @@ func TestCheckSignatureEncoding(t *testing.T) {
...
+			errCode: ErrCodeSeparator,
+		},
+		{
+			// Nested unexecuted branches should still be caught.
+			name:    "codesep in nested unexecuted IF with const scriptcode",
+			script:  "0 IF 1 IF CODESEPARATOR ENDIF ENDIF TRUE",
+			flags:   ScriptVerifyConstScriptCode,
+			wantErr: true,
+			errCode: ErrCodeSeparator,
+		},
```
gijswijs, line 505: Maybe also test that a segwit script with `OP_CODESEPARATOR` and `ScriptVerifyConstScriptCode` is still **accepted**.

txscript/engine_test.go
```diff
@@ -426,3 +426,113 @@ func TestCheckSignatureEncoding(t *testing.T) {
...
+					0xfa, 0x20, 0x9c, 0x6a,
+					0x85, 0x2d, 0xd9, 0x06,
+					0x60, 0xa2, 0x0b, 0x2d,
+					0x9c, 0x35, 0x24, 0x23,
+					0xed, 0xce, 0x25, 0x85,
+					0x7f, 0xcd, 0x37, 0x04,
+				}),
+				Index: 0,
+			},
+			SignatureScript: mustParseShortForm("TRUE"),
```
gijswijs, line 455: Is it meaningful to also test `sigScript` for rejection of `OP_CODESEPARATOR` in non-segwit scripts? 

`SCRIPT_VERIFY_SIGPUSHONLY` enforces that sigScripts must be push-only. Since that flag is part of `STANDARD_SCRIPT_VERIFY_FLAGS` (just like `SCRIPT_VERIFY_CONST_SCRIPTCODE`), any `sigScript` containing `OP_CODESEPARATOR` would already be rejected by the push-only check before the const-scriptcode check fires.

So a test with `OP_CODESEPARATOR` in `sigScript` would only be meaningful if you set `ScriptVerifyConstScriptCode` without `ScriptVerifySigPushOnly` which is a flag combination that doesn't happen with standard policy.

Just putting it out here, so that @Roasbeef can decide.

gijswijs (changes requested):
A few minor comments.

yyforyongyu (approved):
LGTM 🔨

txscript/engine_test.go
Roasbeef: Good food for thought, skipping as it isn't a standard policy flag combo

gijswijs (approved):
LGTM 🎉

PR #2490, opened 2026-03-05
`rpcclient`: clarify `ConnConfig` `Host` field description

JeremyRand:
## Change Description

The previous description incorrectly implied that Unix sockets were not supported (support was added in #2168).

I think the description is still a bit more vague than would be ideal (i.e. it doesn't say how the IP should be separated from the port, nor does it say what prefix is used to signify a Unix socket). This PR is still an improvement, but if you prefer that I fix the vagueness too, let me know and I'm happy to update this PR to do so.

## Steps to Test

No testing applicable; it's just a comment change.

### Testing
- [x] Your PR passes all CI checks. (I don't think there should be any PR check that apply to comments, but maybe I'm mistaken?)
- [x] Tests covering the positive and negative (error paths) are included. (No new code.)
- [x] Bug fixes contain tests triggering the bug to prevent regressions. (No code fixes.)

- [x] Any new logging statements use an appropriate subsystem and logging level. (No new code.)

### DIFF
--- a/rpcclient/infrastructure.go
+++ b/rpcclient/infrastructure.go
@@ -1204,8 +1204,8 @@ func (c *Client) WaitForShutdown() {
 // ConnConfig describes the connection configuration parameters for the client.
 // This
 type ConnConfig struct {
-	// Host is the IP address and port of the RPC server you want to connect
-	// to.
+	// Host is the IP address and port, or Unix socket path, of the RPC server
+	// you want to connect to.
 	Host string
 
 	// Endpoint is the websocket endpoint on the RPC server.  This is


JeremyRand:
Been 6 months since I submitted this. Some review would be appreciated if it's not too much bother.

odingaval (commented):
Good catch! 

You're right that after #2168 added Unix socket support, the godoc for `ConnConfig.Host` was never updated. It still said `IP address and port of the RPC server`, which makes it sound like only TCP is supported.

In reality `newHTTPClient` -> `ParseAddressString` already handles a path starting with `/` as a Unix socket and dials `unix` instead of `tcp` (avoids the `net.ResolveTCPAddr` DNS lookup). So `Host: "/tmp/btcd.sock"` works fine, but a reader of the current doc would never know to try it.

Your change to `IP address and port, or Unix socket path` makes the documentation match what the code already does. 
This small clarification will actually help prevent duplicate questions and confusion.

PR #2491, opened 2026-03-06
wire: reject messages with unconsumed trailing payload bytes

Roasbeef:
In this PR, we add a strictness check to `readMessageWithEncodingNInternal` in
the `wire` package that rejects any message whose payload isn't fully consumed
by `BtcDecode`. After decoding completes, we now check `pr.Len()` and return a
`MessageError` if any bytes remain in the buffer.

Previously, the payload buffer was handed to `BtcDecode` without any
post-decode validation. If `BtcDecode` stopped reading before exhausting the
buffer (which is the correct behavior for all message types, they read exactly
what they need), any extra bytes at the tail were silently ignored. The full
raw payload, trailing bytes included, would then flow through to application
code unchecked. We now reject such messages at the wire layer before they reach
any higher-level processing.

The check is intentionally placed in the shared `readMessageWithEncodingNInternal`
path so it applies uniformly to all message types, not just blocks or transactions.

See each commit message for a detailed description w.r.t the incremental changes.

### DIFF
--- a/wire/message.go
+++ b/wire/message.go
@@ -669,6 +669,16 @@ func readMessageWithEncodingNInternal(r io.Reader, pver uint32,
 		return totalBytes, nil, nil, err
 	}
 
+	// Reject messages where the payload was not fully consumed by
+	// BtcDecode. A peer could otherwise append arbitrary trailing bytes
+	// to an otherwise valid message, which would be silently accepted
+	// and persisted (e.g., in the block database).
+	if pr.Len() > 0 {
+		str := fmt.Sprintf("message payload has %d extra bytes "+
+			"after decode", pr.Len())
+		return totalBytes, nil, nil, messageError("ReadMessage", str)
+	}
+
 	return totalBytes, msg, payload, nil
 }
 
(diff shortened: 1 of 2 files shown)


wire/message.go
```diff
@@ -669,6 +669,16 @@ func readMessageWithEncodingNInternal(r io.Reader, pver uint32,
 		return totalBytes, nil, nil, err
 	}
 
+	// Reject messages where the payload was not fully consumed by
+	// BtcDecode. A peer could otherwise append arbitrary trailing bytes
+	// to an otherwise valid message, which would be silently accepted
+	// and persisted (e.g., in the block database).
+	if pr.Len() > 0 {
```
allocz, line 676: Code looks good, confirmed that reverting the patch causes failure of the test `TestReadMessageTrailingBytes`.

Nice work @Roasbeef

allocz (approved):

kcalvinalvin:
ACK on c64768b2822a4915c438d69f9092b569b4665639

Still feel that I could've got a CVE. Meh. Next time I'll crash some nodes beforehand

kcalvinalvin (approved):

PR #2494, opened 2026-03-10
rpcclient: fix HTTP POST mode with scheme-prefixed Host

hieblmi:
Fixes #2464, a regression introduced by PR #2168  which added unix socket support.

When `ConnConfig.Host` contains a scheme prefix (e.g. `https://go.getblock.io/xxxx`), the rpcclient fails with a dial error because `ParseAddressString` rejects addresses containing `://` (only `unix://` is supported).

### DIFF
--- a/rpcclient/infrastructure.go
+++ b/rpcclient/infrastructure.go
@@ -1348,7 +1348,13 @@ func newHTTPClient(config *ConnConfig) (*http.Client, error) {
 		}
 	}
 
-	parsedDialAddr, err := ParseAddressString(config.Host)
+	// Strip any scheme prefix from the host since the scheme is determined
+	// by DisableTLS. Then extract just the host:port for dialing (ignoring
+	// any path component).
+	host := stripScheme(config.Host)
+	dialHost := extractHostPort(host)
+
+	parsedDialAddr, err := ParseAddressString(dialHost)
 	if err != nil {
 		return nil, err
 	}
@@ -1379,7 +1385,13 @@ func (config *ConnConfig) httpURL() (string, error) {
 		protocol = "https"
 	}
 
-	parsedAddr, err := ParseAddressString(config.Host)
+	// Strip any scheme prefix from the host since the scheme is determined
+	// by DisableTLS. This allows users to pass full URLs like
+	// "https://host.example/path" in the Host field.
+	host := stripScheme(config.Host)
+	dialHost := extractHostPort(host)
+
+	parsedAddr, err := ParseAddressString(dialHost)
 	if err != nil {
 		return "", fmt.Errorf("error parsing host '%v': %v",
 			config.Host, err)
@@ -1392,7 +1404,7 @@ func (config *ConnConfig) httpURL() (string, error) {
 		// The Unix domain socket is specified in the DialContext.
 		httpURL = protocol + "://unix"
 	default:
-		httpURL = protocol + "://" + config.Host
+		httpURL = protocol + "://" + host
 	}
 
 	return httpURL, nil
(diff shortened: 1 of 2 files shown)


seeforschauer (commented):
Nice fix — this also resolves a separate issue we hit with Alchemy's BTC endpoint where `Host = "bitcoin-mainnet.g.alchemy.com/v2/<key>"` (path but no scheme prefix) caused `dial tcp <ip>:0: connect: can't assign requested address`. The `extractHostPort` helper correctly strips the path for dialing while preserving it in the URL.

One suggestion: `TestIssue2464_SchemeInHostHTTPPost` currently uses `srv.URL + "/myapikey"` which always includes a scheme prefix. Could be worth adding a case for path-without-scheme, since that's a common usage pattern with third-party providers:

```go
// Host with path but no scheme (e.g. API key in URL path).
connCfg := &ConnConfig{
    Host:         strings.TrimPrefix(srv.URL, "http://") + "/myapikey",
    User:         "user",
    Pass:         "pass",
    HTTPPostMode: true,
    DisableTLS:   true,
}
```

rpcclient/infrastructure.go
```diff
@@ -1783,6 +1795,34 @@ func cutPrefix(s, prefix string) (after string, found bool) {
...
+	// contain a path. We prepend a dummy scheme so that url.Parse treats
+	// the input as a hierarchical URL.
+	u, err := url.Parse("dummy://" + host)
+	if err != nil {
+		return host
+	}
+	if u.Host != "" {
+		return u.Host
+	}
+	return host
```
seeforschauer, line 1823: bug: `extractHostPort` breaks unix socket paths.

When `Host = "unix:///var/run/btcd.sock"`, `stripScheme` correctly leaves it alone, but then `url.Parse("dummy://unix:///var/run/btcd.sock")` parses `Host` as `"unix:"` and `Path` as `"///var/run/btcd.sock"` — the socket path is lost.

Verified locally:
```
url.Parse("dummy://unix:///var/run/btcd.sock")
  → Host="unix:"  Path="///var/run/btcd.sock"
```

Then `ParseAddressString("unix:")` fails because it expects `unix:///path`.

This affects the unix socket + HTTP POST mode path that `httpURL()` already supports (the `parsedAddr.Network() == "unix"` branch).

Fix: bail out early if the input still contains a scheme separator. After `stripScheme`, the only remaining `://` would be non-HTTP schemes like `unix://` or `unixpacket://` that `ParseAddressString` handles natively.

```suggestion
	// If the host still contains a scheme (e.g. unix://, unixpacket://),
	// return it as-is for ParseAddressString to handle directly.
	// The dummy-scheme url.Parse trick below would strip the path.
	if strings.Contains(host, "://") {
		return host
	}

	// Use url.Parse to reliably extract the host from a string that may
	// contain a path. We prepend a dummy scheme so that url.Parse treats
	// the input as a hierarchical URL.
	u, err := url.Parse("dummy://" + host)
	if err != nil {
		return host
	}
	if u.Host != "" {
		return u.Host
	}
	return host
```

seeforschauer (approved):
Fix is correct for the main use case (scheme-prefixed hosts like `https://go.getblock.io/xxxx`). Helper functions are clean and well-tested.

One finding: `extractHostPort` breaks unix socket paths in HTTP POST mode (see inline). The fix is small — a `://` guard before the `url.Parse` trick.

TechLateef (commented):
`TestExtractHostPort` has no test case for unix socket paths. Given the `url.Parse("dummy://"+host)` trick, passing `unix:///var/run/btcd.sock` would return `"unix:"` and break `ParseAddressString`.

The early `strings.Contains(host, "://")` guard suggested by @seeforschauer  fixes this  worth also adding `{"unix:///var/run/btcd.sock", "unix:///var/run/btcd.sock"}` to `TestExtractHostPort` to prevent regressions.

PR #2500, opened 2026-03-20
rpcclient: fix several synchronization bugs

starius:
## Change Description

Incorporated https://github.com/btcsuite/btcd/pull/2451 :

Modify the rpcclient http POST call to ensure that a shutdown immediately interrupts in-flight requests, which otherwise would have to wait until timeout.

Fix 3 other problems in rpcclient:

 - HTTP POST shutdown can deadlock WaitForShutdown due to double-response race
 - Batch-mode Send() error path leaves queued per-request futures unresolved
 - NewBatch starts HTTP POST handlers twice, increasing concurrency surface

## Steps to Test

```
go test ./rpcclient -count=1
```

Each test is a regression test. It fails if the patch is reverted.

### Testing

### DIFF
--- a/rpcclient/infrastructure.go
+++ b/rpcclient/infrastructure.go
@@ -956,19 +956,32 @@ cleanup:
 // HTTP client associated with the client.  It is backed by a buffered channel,
 // so it will not block until the send channel is full.
 func (c *Client) sendPostRequest(jReq *jsonRequest) {
-	// Don't send the message if shutting down.
+	// Prefer shutdown when it is already closed so this path is
+	// deterministic. This mirrors addRequest and avoids post-shutdown
+	// enqueueing.
 	select {
 	case <-c.shutdown:
-		jReq.responseChan <- &Response{result: nil, err: ErrClientShutdown}
+		jReq.responseChan <- &Response{
+			result: nil,
+			err:    ErrClientShutdown,
+		}
+
+		return
+
 	default:
 	}
 
+	// Normal path: either enqueue, or fail if shutdown closes in the race
+	// window after the guard above.
 	select {
 	case c.sendPostChan <- jReq:
 		log.Tracef("Sent command [%s] with id %d", jReq.method, jReq.id)
 
 	case <-c.shutdown:
-		return
+		jReq.responseChan <- &Response{
+			result: nil,
+			err:    ErrClientShutdown,
+		}
 	}
 }
 
--- a/rpcclient/infrastructure.go
+++ b/rpcclient/infrastructure.go
@@ -1762,6 +1762,38 @@ func (c *Client) sendAsync() (FutureGetBulkResult, error) {
 	return responseChan, nil
 }
 
+// failBatchRequests resolves every queued batch request with the provided error
+// and clears all internal request tracking.
+//
+// This function is safe for concurrent access.
+func (c *Client) failBatchRequests(err error) {
+	c.requestLock.Lock()
+	defer c.requestLock.Unlock()
+
+	c.batchLock.Lock()
+	defer c.batchLock.Unlock()
+
+	for e := c.batchList.Front(); e != nil; e = e.Next() {
+		req := e.Value.(*jsonRequest)
+
+		// Resolve all pending futures on the first batch-level failure
+		// so callers waiting on Receive don't block indefinitely.
+		// Safe: batch-mode responseChan buffers are unwritten here,
+		// so this send won't block while locks are held. Batch-mode
+		// requests only use addRequest (not sendPostRequest), so each
+		// responseChan buffer is still empty.
+		req.responseChan <- &Response{err: err}
+	}
+
+	c.requestMap = make(map[uint64]*list.Element)
+	c.batchList = list.New()
+
+	// Batch-mode requests are tracked in batchList, so requestList should
+	// already be empty. Keep this defensive reset for invariants and future
+	// call paths.
+	c.requestList.Init()
+}
+
 // Marshall's bulk requests and sends to the server
 // creates a response channel to receive the response
 func (c *Client) Send() error {
@@ -1772,12 +1804,7 @@ func (c *Client) Send() error {
 
 	batchResp, err := future.Receive()
 	if err != nil {
-		// Clear batchlist in case of an error.
-
-		c.batchLock.Lock()
-		c.batchList = list.New()
-		c.batchLock.Unlock()
-
+		c.failBatchRequests(err)
 		return err
 	}
 
--- a/rpcclient/infrastructure.go
+++ b/rpcclient/infrastructure.go
@@ -98,6 +98,10 @@ const (
 	// defaultHTTPTimeout is the default timeout for an http request, so the
 	// request does not block indefinitely.
 	defaultHTTPTimeout = time.Minute
+
+	// sendPostRequestTries is the number of times to retry failed HTTP POST
+	// requests before giving up.
+	sendPostRequestTries = 10
 )
 
 // jsonRequest holds information about a json request that is used to properly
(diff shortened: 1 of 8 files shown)


saubyk:
hello @wydengyre @seeforschauer @jcvernaleo would you consider reviewing this one?

rpcclient/infrastructure.go
```diff
@@ -928,19 +928,17 @@ cleanup:
...
-	}
+		return
 
-	select {
 	case c.sendPostChan <- jReq:
 		log.Tracef("Sent command [%s] with id %d", jReq.method, jReq.id)
-
-	case <-c.shutdown:
-		return
 	}
```
seeforschauer: suggestion (non-blocking): Consider the priority-select pattern here.

When `shutdown` is already closed (its permanent state after `Shutdown()`), the single `select` randomly picks between responding and enqueueing (~50/50 per Go spec). If `sendPostHandler` has already exited cleanup, the enqueued request's future is never resolved.

A non-blocking shutdown guard first makes the common post-shutdown path deterministic — `addRequest` (line 213) already uses this exact pattern for the same reason. The remaining race (shutdown closing *between* the two selects) has a much narrower window.

```suggestion
	// Prefer shutdown: if already closed, fail the request immediately.
	// This avoids a random race between shutdown and enqueue when both
	// channels are ready, consistent with the guard in addRequest.
	select {
	case <-c.shutdown:
		jReq.responseChan <- &Response{result: nil, err: ErrClientShutdown}
		return
	default:
	}

	// Normal path: enqueue or fail on shutdown. Exactly one outcome.
	select {
	case c.sendPostChan <- jReq:
		log.Tracef("Sent command [%s] with id %d", jReq.method, jReq.id)
	case <-c.shutdown:
		jReq.responseChan <- &Response{result: nil, err: ErrClientShutdown}
	}
```

rpcclient/infrastructure.go
```diff
@@ -1716,6 +1719,30 @@ func (c *Client) sendAsync() (FutureGetBulkResult, error) {
...
+
+	c.batchLock.Lock()
+	defer c.batchLock.Unlock()
+
+	for e := c.batchList.Front(); e != nil; e = e.Next() {
+		req := e.Value.(*jsonRequest)
+
+		// Resolve all pending futures on the first batch-level failure so
+		// callers waiting on Receive don't block indefinitely.
+		req.responseChan <- &Response{err: err}
```
seeforschauer, line 1790: nit: This send is safe because in batch mode `sendRequest` only calls `addRequest` (never `sendPostRequest`), so individual `responseChan` buffers (size 1) are guaranteed unwritten at this point. A brief comment documenting this invariant would help future readers — e.g.:

```go
// Safe: batch-mode responseChan buffers are unwritten here,
// so this send won't block while locks are held.
req.responseChan <- &Response{err: err}
```

rpcclient/infrastructure.go
```diff
@@ -1716,6 +1719,30 @@ func (c *Client) sendAsync() (FutureGetBulkResult, error) {
...
+	}
+
+	c.requestMap = make(map[uint64]*list.Element)
+	c.requestList.Init()
```
seeforschauer: question: In batch mode, `addRequest` pushes to `batchList`, never `requestList`, so this should already be empty. Intentional defensive reset, or leftover? If intentional, a quick comment would clarify.

seeforschauer (approved):
LGTM — all three bugs are real and the fixes are correct. I've been working in `rpcclient/infrastructure.go` recently (#2506, #2505), so I have direct context on these paths.

Commit structure is clean (one bug, one test, one commit). Each test fails when the fix is reverted — solid regression coverage.

One suggestion on `sendPostRequest` (inline) for tighter shutdown determinism using the priority-select pattern already in `addRequest`. Two minor notes on `failBatchRequests`.

rpcclient/infrastructure.go
Roasbeef: This is effectively how the code already was. 

@starius isn't it better to prioritize the shutdown path?

Roasbeef (changes requested):
Change looks good, only comment is why we'd move away from the pattern that prioritizes a cancel path.

rpcclient/infrastructure.go
seeforschauer: Right, the structure is two selects like the original — the critical difference is the `return` after the first shutdown case. The original fell through into the second select even after sending `ErrClientShutdown`, which allowed the double-resolve.

With the `return`, it becomes the standard priority-select: if shutdown is already closed, respond deterministically and exit. The second select is only reached when shutdown wasn't closed at check time.

starius: Fixed! Now it prioritizes the shutdown path:

```go
// sendPostRequest sends the passed HTTP request to the RPC server using the
// HTTP client associated with the client.  It is backed by a buffered channel,
// so it will not block until the send channel is full.
func (c *Client) sendPostRequest(jReq *jsonRequest) {
	// Prefer shutdown when it is already closed so this path is
	// deterministic. This mirrors addRequest and avoids post-shutdown
	// enqueueing.
	select {
	case <-c.shutdown:
		jReq.responseChan <- &Response{
			result: nil,
			err:    ErrClientShutdown,
		}

		return

	default:
	}

	// Normal path: either enqueue, or fail if shutdown closes in the race
	// window after the guard above.
	select {
	case c.sendPostChan <- jReq:
		log.Tracef("Sent command [%s] with id %d", jReq.method, jReq.id)

	case <-c.shutdown:
		jReq.responseChan <- &Response{
			result: nil,
			err:    ErrClientShutdown,
		}
	}
}

```

rpcclient/infrastructure.go
starius: Added a comment:
```go
		// Resolve all pending futures on the first batch-level failure
		// so callers waiting on Receive don't block indefinitely.
		// Safe: batch-mode responseChan buffers are unwritten here,
		// so this send won't block while locks are held. Batch-mode
		// requests only use addRequest (not sendPostRequest), so each
		// responseChan buffer is still empty.
		req.responseChan <- &Response{err: err}
```

rpcclient/infrastructure.go
starius: It is a defensive reset. Added a comment:
```go
	// Batch-mode requests are tracked in batchList, so requestList should
	// already be empty. Keep this defensive reset for invariants and future
	// call paths.
	c.requestList.Init()
```

starius:
I addressed the remaining comments in https://github.com/btcsuite/btcd/pull/2451 and added the updated version of it here as commit "rpcclient: support canceling in-flight http requests". Functionally it is the same, but refactored. I simplified the code by making a function that returns `([]byte, error)`; error remapping and channel sending done is done on the call site.

CC @Roasbeef @seeforschauer  @wydengyre

guibressan:
I read the diff and ran the tests, LGTM. Nice work @starius.

seeforschauer (approved):
LGTM, thanks for the clean fixes!

rpcclient/infrastructure.go
```diff
@@ -871,50 +877,70 @@ func (c *Client) handleSendPostMessage(jReq *jsonRequest) {
...
+		// cannot "error out" other than through the status code error
+		// handled above.
+		return batchResponse, nil
+	}
+
+	return resp.result()
+}
+
+// handleSendPostMessageWithRetry runs handleSendPostMessage using the provided
+// retry count. It exists so tests can exercise retry edge cases quickly.
```
kcalvinalvin: nit: this might read as `handleSendPostMessageWithRetry` is for internal tests.

rpcclient/infrastructure_test.go
```diff
@@ -93,18 +169,333 @@ func TestParseAddressString(t *testing.T) {
...
+	}
+
+	result, err := future.Receive()
+	require.Zero(t, result)
+	require.ErrorContains(t, err, ErrClientShutdown.Error())
+}
+
+// TestHandleSendPostMessageShutdownDuringRetryBackoff ensures shutdown
+// cancellation interrupts retry backoff and remaps to ErrClientShutdown.
+func TestHandleSendPostMessageShutdownDuringRetryBackoff(t *testing.T) {
```
kcalvinalvin: These three tests:

`TestHandleSendPostMessageShutdownDuringRetryBackoff`
`TestHandleSendPostMessageShutdownOnFinalRetry`
`TestHandleSendPostMessageShutdownDuringBodyRead`

iiuc, these are all just trying to assert that the err returns as `ErrClientShutdown`? They exist to test like this one branch and is ridiculously verbose. imo these need to be cleaned up into one table driven test for future maintainability.

rpcclient/infrastructure_test.go
```diff
@@ -93,18 +169,333 @@ func TestParseAddressString(t *testing.T) {
...
+		false,
+	)
+	require.NoError(t, err)
+	require.Equal(t, []byte("1"), result)
+}
+
+// TestHandleSendPostMessageWithRetryShutdownDuringRetryBackoff ensures
+// that handleSendPostMessageWithRetry returns context cancellation from the
+// retry-backoff path.
+func TestHandleSendPostMessageWithRetryShutdownDuringRetryBackoff(
```
kcalvinalvin: The tests `TestHandleSendPostMessageWithRetryShutdownDuringRetryBackoff`, `TestHandleSendPostMessageWithRetryShutdownOnFinalRetry`, `TestHandleSendPostMessageWithRetryShutdownOnBodyRead` should also really be a single table driven test.

kcalvinalvin:
The actual code looks good but the tests... they're rough.

Like it looks like completely claude generated in a single go. I left some reviews but even the other tests don't look great.

rpcclient/infrastructure.go
starius: I renamed it to `sendPostRequestWithRetry` so it does not sound like test-specific.

rpcclient/infrastructure_test.go
starius: I replaced them with a single table-driven test that reuses the same shutdown scenarios as the helper-level test.

> these are all just trying to assert that the err returns as ErrClientShutdown?

Yes, every shutdown-triggered path must receive a `ErrClientShutdown` on `responseChan`. The tests check this.

rpcclient/infrastructure_test.go
starius: I collapsed those three helper-level shutdown tests into one table-driven test and kept the success case separate:
 - retry-backoff cancellation
 - final-retry cancellation
 - cancellation during body read.

I also tightened the final-retry case so it depends on request-context propagation rather than a transport stub returning `context.Canceled` directly, which keeps it a stronger regression test.

starius:
> The actual code looks good but the tests... they're rough.
> 
> Like it looks like completely claude generated in a single go. I left some reviews but even the other tests don't look great.

I cleaned up the shutdown-path tests to reduce copy-paste and converted them into table tests. I also did other small changes, described in new commits.

I kept the changes in fixup commits which can be applied automatically by `git rebase --autosquash`. I'm keeping them separate for now to make it clear what changed.

starius:
Rebased and formatted code (pre-existing not formatted code).


<details>
  <summary>CI is failing because of pre-existing issues in other packages:</summary>
  
  - `blockchain/TestMaybeAcceptBlockReusesHeaderNode` fails with coinbase height enforcement.
  - `blockchain/TestProcessBlockHeader` fails because it still builds regtest headers with Version: 1.
  - `netsync/TestStartSyncChainCurrent`, `TestStartSyncBlockFallback`, and the `TestSyncStateMachine` cases fail for the same reason, and the expected 20, actual 0 assertions are downstream from those headers being rejected.

Why this is happening:

  - Regtest now enforces BIP34/BIP65/BIP66 from height 1 and rejects old block versions.
  - But these older tests still construct invalid regtest blocks/headers:
      - blockchain/process_test.go:39 uses Version: 1
      - netsync/manager_test.go:1190 uses Version: 1
      - blockchain/accept_test.go:29 goes through addBlocks(...), which now trips the coinbase-height rule in this setup

Local reproduction:

  - `go test ./blockchain -run 'TestMaybeAcceptBlockReusesHeaderNode|TestProcessBlockHeader|TestRegtestBlockVersions|TestRegtestRejectsCoinbaseMissingHeight' -count=1`
  - `go test ./netsync -run 'TestStartSyncChainCurrent|TestStartSyncBlockFallback|TestSyncStateMachine|TestIsSyncCandidateRegtest' -count=1`
  
</details>

starius:
Fixed the existing tests failures in a separate commit.

Roasbeef (approved):
LGTM 🐴

PR #2506, opened 2026-03-23
rpcclient: add context.Context support for HTTP POST mode

seeforschauer:
## Change Description

Add `context.Context` support to rpcclient's HTTP POST mode, allowing callers to cancel or timeout individual RPC calls. Fixes #2499.

Currently, rpcclient methods like `GetBlockCount()` don't accept a context, which means callers have no way to cancel or set deadlines on individual RPC calls. In HTTP POST mode, `handleSendPostMessage` uses `http.NewRequest` without context, and the 10-retry loop doesn't respect external cancellation.

### Changes

- Add `ctx` field to `jsonRequest` struct
- Use `http.NewRequestWithContext` in `handleSendPostMessage` when ctx is set
- Respect context cancellation during retry backoff (`select` on `reqCtx.Done()`)
- Add `SendCmdWithContext(ctx, cmd)` to `Client`
- Add `GetBlockCountWithContext(ctx)` as the first context-aware method variant
- Existing methods are unchanged — fully backwards compatible

### Design decisions

- `SendCmdWithContext` mirrors `SendCmd` with an added ctx parameter
- When `ctx` is nil (e.g., from existing `SendCmd` callers), `context.Background()` is used as fallback
- Context cancellation during retry backoff returns immediately instead of waiting for next attempt
- Additional `*WithContext` methods for other RPC calls can be added incrementally in follow-up PRs

## Steps to Test

```
go test ./rpcclient/ -run TestGetBlockCountWithContext -v
go test ./rpcclient/ -run TestSendCmdWithContext -v
go test ./rpcclient/ -run TestGetBlockCount_Backwards -v
```

### Testing

- [x] The change is not insubstantial.


rpcclient/chain.go
```diff
@@ -302,12 +303,25 @@ func (c *Client) GetBlockCountAsync() FutureGetBlockCountResult {
...
 
-// GetBlockCount returns the height of the most-work fully-validated chain.
-// The genesis block has height 0.
+// GetBlockCount returns the height of the most-work fully-validated
+// chain. The genesis block has height 0.
 func (c *Client) GetBlockCount() (int64, error) {
 	return c.GetBlockCountAsync().Receive()
 }
 
+// GetBlockCountWithContext is the context-aware variant of
```
Roasbeef, line 295: Alternatively, we could update _all_ the methods to accept a context param, but via a functional option. This would break any interfaces created off the client, be direct callers would still compile.

Roasbeef: That would be a larger diff tho, so perhaps we'll land this one extension for now, then return to this later.

rpcclient/context_test.go
```diff
@@ -0,0 +1,212 @@
...
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestGetBlockCountWithContext_Success verifies that the context-
+// aware GetBlockCountWithContext returns the correct block height.
+func TestGetBlockCountWithContext_Success(t *testing.T) {
```
Roasbeef: We could camel case everywhere, including tests.

rpcclient/infrastructure.go
```diff
@@ -1045,6 +1060,47 @@ func (c *Client) SendCmd(cmd interface{}) chan *Response {
 	return responseChan
 }
 
+// SendCmdWithContext sends the passed command with the given context.
+// The context is used to cancel the underlying HTTP request in
+// HTTP POST mode. In websocket mode the context is currently
+// ignored. The returned channel delivers the response.
+func (c *Client) SendCmdWithContext(
```
Roasbeef: This duplicates SendCmd, let's refactor such that `SendCmdWithContext` can just call `SendCmd, but pass in the context from the user.

rpcclient/infrastructure.go
```diff
@@ -832,6 +841,12 @@ func (c *Client) handleSendPostMessage(jReq *jsonRequest) {
 
 		case <-c.shutdown:
 			return
+
```
Roasbeef: Above when we call `c.httpClient.Do(httpReq)`, we should check the context error, as perhaps we should bail right there instead of retrying again.

Roasbeef (changes requested):

rpcclient/infrastructure.go
seeforschauer: Done — `SendCmd` now delegates to `SendCmdWithContext(context.Background(), cmd)`. Single implementation.

rpcclient/infrastructure.go
seeforschauer: Done — added `reqCtx.Err()` check right after `httpClient.Do`. If context is cancelled, we bail out immediately instead of entering the backoff/retry path.

rpcclient/context_test.go
seeforschauer: Done — all test names use camelCase now.

rpcclient/chain.go
seeforschauer: Like this direction a lot — `CallOption` would avoid the `WithContext` method explosion across 100+ methods and is extensible for future options. Happy to do it as a follow-up PR once this one lands.

guibressan:
Diff looks good, tests passing.

Nice work @seeforschauer!

seeforschauer:
@Roasbeef all four items addressed — mind taking another look when you get a chance?

rpcclient/context_test.go
```diff
@@ -0,0 +1,212 @@
...
+	)
+}
+
+// TestSendCmdWithContextRetryRespectsCancel verifies that context
+// cancellation during the retry backoff terminates immediately
+// rather than waiting for all retry attempts.
+func TestSendCmdWithContextRetryRespectsCancel(t *testing.T) {
+	t.Parallel()
+
+	var attempts int
```
kcalvinalvin: nit: this doesn't do anything right?

rpcclient/context_test.go
```diff
@@ -0,0 +1,212 @@
...
+	// SendCmd (no context) should still work — ctx field is nil,
+	// handleSendPostMessage falls back to context.Background.
+	count, err := client.GetBlockCount()
+	require.NoError(t, err)
+	require.Equal(t, int64(500), count)
+}
+
+// TestGetBlockCountWithContextRPCError verifies that JSON-RPC
+// errors are properly propagated through the context path.
+func TestGetBlockCountWithContextRPCError(t *testing.T) {
```
kcalvinalvin, line 186: This test passes even if we use the non context getblockcount.

It isn't really testing the context path.

kcalvinalvin (commented):
Looks good overall. Some minor test related issues that I saw

rpcclient/context_test.go
seeforschauer: Good catch. The old test used an HTTP 500 response, which doesn't actually exercise the retry/backoff path, and the `attempts` counter wasn't asserting anything meaningful. Updated in 3c0fd18 to force a transport error via a custom `RoundTripper`, then assert we exit with `context.DeadlineExceeded` after exactly one attempt.

rpcclient/context_test.go
seeforschauer: Yep, agreed. That test wasn't exercising the nil-ctx branch in `handleSendPostMessage` at all — `client.GetBlockCount()` would have succeeded via the context path too. Replaced it in 3c0fd18 with a direct `jsonRequest{ctx: nil}` call into `handleSendPostMessage`, which actually forces the `context.Background()` fallback on line 791–794.

seeforschauer:
@kcalvinalvin pls review

seeforschauer:
@Roasbeef would you please review once more?

address/base58/README.md
```diff
@@ -19,13 +19,13 @@ $ go get -u github.com/btcsuite/btcd/address/v2/base58
 
 ## Examples
 
-* [Decode Example](http://godoc.org/github.com/btcsuite/btcd/address/v2/base58#example-Decode)  
+* [Decode Example](http://godoc.org/github.com/btcsuite/btcd/address/v2/base58#example-Decode)
```
kcalvinalvin: I get that these are easy (probably automatic) fixes but it does make reviewing a bit more annoying since I believe all the reviewers read git diffs commit by commit

kcalvinalvin:
The code looks good. Though the tests aren't doing much.

These test all pass even when the context is dropped in GetBlockCountWithConext:
  - TestGetBlockCountWithContextSuccess
  - TestGetBlockCountBackwardsCompatible
  - TestGetBlockCountWithContextRPCError
  - TestHandleSendPostMessageNilCtxFallsBackToBackground

Ran the tests with this code:

```
[I] calvin@nixos ~/b/b/b/rpcclient ((6e5bb34c)) [1]> git diff
diff --git a/rpcclient/chain.go b/rpcclient/chain.go
index a5cedc64..6381534d 100644
--- a/rpcclient/chain.go
+++ b/rpcclient/chain.go
@@ -301,7 +301,8 @@ func (c *Client) GetBlockCountWithContext(

        cmd := btcjson.NewGetBlockCountCmd()
        return FutureGetBlockCountResult(
-               c.SendCmdWithContext(ctx, cmd),
+               //c.SendCmdWithContext(ctx, cmd),
+               c.SendCmdWithContext(context.Background(), cmd),
        ).Receive()
 }
```

Like the tests are implementing the handler in the test and is just checking if that works. I don't think that's useful really.

I feel that the actual code is mergeable. The tests... eh.

Another nit is the git commit format. I'd prefer the follow up commits to be amended into the originals. But these aren't a blocker.

seeforschauer:
updated

saubyk:
Hi @seeforschauer why is this pr in draft status?

kcalvinalvin:
The latest commit ccbcd30 fixes the test problems I pointed out. LGTM.

Still not a fan of the commit order but that's not a stopper

kcalvinalvin (approved):
Looks good

saubyk:
@seeforschauer does this pr still need to be in draft?

seeforschauer:
No, it does not need to stay in draft. Marked ready for review.

seeforschauer:
Also pushed 920a81ed to remove unrelated README/doc/newline cleanup, so the PR diff is now limited to the rpcclient context changes. Local `go test ./rpcclient` passes. GitHub Actions is currently `action_required` for the new fork commit.

odingaval (approved):
Approach ACK

Checked the postRequestContext (L783-802) - merging of clientCtx and requestCtx with   WithCancelCause + AfterFunc is correct. Handles nil cases, propagates cause, and cleanup func avoids leak.
This correctly preserves shutdown cancellation from #2451 while adding user cancellation.

The nil-ctx fix in 3c0fd18 also looks good

LGTM

seeforschauer:
oops.

seeforschauer:
New PR: https://github.com/btcsuite/btcd/pull/2608

seeforschauer:
cannot reopen, 
https://github.com/btcsuite/btcd/pull/2608

odingaval:
> oops.

What happened?

PR #2508, opened 2026-03-24
blockchain+netsync: follow-up fixes for headers-first IBD review

Roasbeef:
**NOTE:** This PR is stacked on top of #2428 and should be reviewed/merged
after that PR lands. The diff shown here includes #2428's changes; the new
commits are the last four on this branch.

In this PR, we address a set of issues found during code review of #2428
(the headers-first IBD rework). The changes here are correctness fixes,
thread safety improvements, and performance/cleanup items that sit on top
of the new IBD pipeline. See each commit message for a detailed description
w.r.t the incremental changes.

## Correctness Fixes

The new `fetchHigherPeers` helper dropped the segwit peer filtering that
the old `startSync` had. Post-segwit activation, this meant btcd could
pick a non-witness peer for sync and fail to fully validate the chain. We
restore the `IsDeploymentActive` + `IsWitnessEnabled` check inside
`fetchHigherPeers` (and the new `fetchEqualPeers`) so witness-only sync
is enforced whenever segwit is active.

The old `startSync` also had an `equalPeers` fallback for peers at the
same height, which is critical for regtest where both nodes start at
genesis (height 0) and neither is strictly higher. We add
`fetchEqualPeers` and wire it into the block-download path of `startSync`
to restore this behavior.

We also bring back `syncCandidate` demotion: peers whose last advertised
block has fallen below our height get `syncCandidate = false` so they
aren't repeatedly considered in future sync rounds. The demotion uses `<`
(not `<=`) to preserve the existing behavior where equal-height peers
remain candidates.

## Thread Safety

`IsValidHeader`, `HeaderHashByHeight`, and `HeaderHeightByHash` all read
from `bestHeader` without holding `chainLock`, while
`maybeAcceptBlockHeader` modifies `bestHeader` under `chainLock.Lock()`.
We acquire `chainLock.RLock()` in each accessor. The redundant `Contains`
check in `HeaderHashByHeight` is also removed since `NodeByHeight` already
guarantees membership in the chain view.

## Performance

`buildBlockRequest` previously iterated from `forkHeight+1` to
`bestHeaderHeight` on every refill call, i.e. O(n) per invocation. We
introduce a `lastBlockRequested` cursor that tracks the highest scanned
height, so subsequent calls skip already-requested ranges. The cursor
resets on peer disconnect and IBD completion.

In `maybeAcceptBlockHeader`, known non-invalid side-chain headers were
falling through to re-run `CheckBlockHeaderSanity` and
`CheckBlockHeaderContext`. Since these headers passed validation when first
added, the re-validation is pure overhead. We return early for known
side-chain headers.

`checkHeadersList` was calling `IsValidHeader` then `HeaderHeightByHash`,
each doing their own `LookupNode`. We add `ValidHeaderHeight` that
combines both into a single index lookup.

## Cleanup

The unused `headerNode` type (left over from the old `headerList`-based
sync) is removed. The duplicate nil guard in `buildBlockRequest` is
removed since the only caller (`fetchHeaderBlocks`) already checks. A
BIP130 design note is added to `handleHeadersMsg` explaining why
unsolicited headers are now accepted (peers can proactively push headers
via `sendheaders`). The crash-recovery behavior of `flushToDB`'s
header-only skip is documented: a restart during header sync loses header
progress, which is an acceptable trade-off given the small size and fast
re-validation of headers.

### DIFF
--- a/blockchain/accept.go
+++ b/blockchain/accept.go
@@ -154,43 +154,53 @@ func (b *BlockChain) maybeAcceptBlockHeader(header *wire.BlockHeader,
 			return false, ruleError(ErrInvalidAncestorBlock, str)
 		}
 
-		// If the node is in the bestHeaders chainview, it's in the main chain.
-		// If it isn't, then we'll go through the verification process below.
+		// The header already exists in the block index and is not
+		// invalid.  If it is in the best header chain, report it as
+		// main chain.
 		if b.bestHeader.Contains(node) {
 			return true, nil
 		}
-	}
 
-	// Perform context-free sanity checks on the block header.
-	err := CheckBlockHeaderSanity(
-		header, b.chainParams.PowLimit, b.timeSource, flags)
-	if err != nil {
-		return false, err
-	}
+		// This is a known, valid side-chain header, so re-validation
+		// can be skipped.  However, we still fall through to the tip
+		// promotion logic below: after a restart the best header chain
+		// is reset to the best block tip, so a previously known side
+		// chain with more cumulative work must be able to become the
+		// best header chain again when its headers are re-announced.
+	} else {
+		// Perform context-free sanity checks on the block header.
+		err := CheckBlockHeaderSanity(
+			header, b.chainParams.PowLimit, b.timeSource, flags)
+		if err != nil {
+			return false, err
+		}
 
-	// The block must pass all of the validation rules which depend on the
-	// position of the block within the block chain.
-	err = CheckBlockHeaderContext(header, prevNode, flags, b, skipCheckpoint)
-	if err != nil {
-		return false, err
-	}
+		// The block must pass all of the validation rules which depend
+		// on the position of the block within the block chain.
+		err = CheckBlockHeaderContext(
+			header, prevNode, flags, b, skipCheckpoint,
+		)
+		if err != nil {
+			return false, err
+		}
 
-	// Create a new block node for the block and add it to the block index.
-	//
-	// Note that the additional information for the actual transactions and
-	// witnesses in the block can't be populated until the full block data is
-	// known since that information is not available in the header.
-	if node == nil {
+		// Create a new block node for the block and add it to the
+		// block index.
+		//
+		// Note that the additional information for the actual
+		// transactions and witnesses in the block can't be populated
+		// until the full block data is known since that information is
+		// not available in the header.
 		node = newBlockNode(header, prevNode)
 		node.status = statusHeaderStored
 		b.index.AddNode(node)
-	}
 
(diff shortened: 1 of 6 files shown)


Roasbeef:
Pushed up a new version. First, we've rebased on top of master to pick up the v2 module migration and the recent netsync changes. The rebase surfaced a semantic conflict: on the v2 `chaincfg`, regtest/simnet mark segwit as always active, so the witness peer filter in `fetchHigherPeers`/`fetchEqualPeers` started rejecting every non-witness peer on regtest, which broke local sync (and a pile of tests). We now skip the witness requirement on local test networks via a new `requireWitnessPeers` helper, mirroring the existing carve out in `isSyncCandidate`.

Beyond the rebase, this version fixes a few issues found in a correctness review pass, folded into the relevant commits.

The block request cursor (`lastBlockRequested`) is now invalidated when the best header chain reorgs beneath it. The cursor is only meaningful for the header chain it was built against: we now remember the header hash at the cursor height, and if that hash no longer matches on the next refill, we reset and re-scan from the fork point. Without this, blocks on the new branch below the cursor would never be requested, and IBD could stall until the stall handler fired ~3 minutes later. A `notfound` for a requested block also rewinds the cursor now, so the block gets re-requested on the next refill rather than waiting out the stall timer. New coverage in `TestBuildBlockRequestReorgResetsCursor` and `TestNotFoundResetsBlockRequestCursor`.

Peer demotion has been narrowed to the best _block_ height. Previously the demotion lived inside `fetchHigherPeers`, which `startSync` also calls with the best _header_ height. During IBD the header chain races ahead of every peer's advertised block, so any peer even one block behind our header tip would get permanently demoted, eventually draining the whole sync candidate pool. Demotion now lives in a dedicated `demoteStalePeers` that `startSync` invokes against the block height only, and `fetchHigherPeers` is a pure filter again.

`maybeAcceptBlockHeader` still skips re-validation for known side-chain headers, but no longer returns early: it falls through to the tip promotion logic instead. This matters after a restart, since `initChainState` resets the best header chain to the best block tip, so a previously known heavier side chain must be able to win back the best header tip when re-announced (otherwise we'd loop on getheaders without making progress until the remote chain grew a new header). `TestProcessBlockHeaderRePromoteSideChain` covers the restart scenario.

Lastly, `TestStartSyncEqualPeersFallback` now actually exercises the equal-height fallback: the old version had the peer at height 5 with our block height at 0, so the peer was selected via the ordinary higher-peers path and the assertion passed by accident. The test now pins everything at genesis, matching the two-fresh-regtest-nodes case the fallback exists for.

PR #2513, opened 2026-03-25
treap, ffldb: improve pool recycling, add batched Delete

Roasbeef:
## Change Description

In this PR, we improve the `sync.Pool`-based node recycling introduced in #2425 and extend the same optimization to `Delete` operations.

### Faster Recycling Check in `Put()`

The original `Put()` recycling loop called `newTreap.get(node.key)` for each intermediate node to check if it was still live in the latest treap version. This worked correctly but cost O(depth × log n) key comparisons per batch iteration. We replace this with a pointer-set approach: `put()` now returns both the set of newly created nodes _and_ the set of original source nodes that were cloned (the "replaced" set). The recycling loop checks pointer identity via `nodeInSet()`, which is just a linear scan of a `[staticDepth]*treapNode` array. This brings the cost down to O(depth²) pointer comparisons, i.e. ~400 pointer equality checks for a typical depth of 20, versus ~8000 key comparisons before.

We also document the MVCC snapshot safety argument directly in the `Put()` godoc. The core insight is that intermediate clones are always fresh pool allocations, never shared with any pre-existing snapshot. Only clones that were subsequently re-cloned (and thus replaced in the new treap) are candidates for recycling. Clones still structurally shared with the new treap are left alone.

To gain confidence in this safety argument, we built a [Quint formal model](https://gist.github.com/Roasbeef/8e8afd1f24a02e775809688d5ec96b8a) (TLA+-inspired) that models the treap as a set of node IDs, non-deterministically applies batched puts, deletes, snapshot captures, and snapshot releases, then checks three invariants across 30,000 randomized traces: no recycled node appears in any snapshot, in the current treap, or anywhere live. All three invariants passed. The model confirmed that fresh clone IDs are always >= `nextId` at allocation time, so they can never collide with any pre-existing snapshot, which is exactly what allows us to use the simpler pointer-set check instead of a full treap traversal.

### Batched `Delete()` with Pool Recycling

Previously, `Delete` accepted a single key and allocated cloned nodes from the pool via `cloneTreapNode` without ever returning them. During IBD, `commitTx` issued many individual `Delete` calls inside `ForEach` loops, creating a one-way pool drain where nodes were pulled from the pool but never recycled back.

We extend `Delete` to accept variadic keys (`Delete(keys ...[]byte)`), mirroring `Put`'s batched API. The internal `delete()` method returns the same `(created, replaced)` arrays, enabling the same inter-step recycling. The `commitTx` callsite in `dbcache.go` now collects deletion keys into slices and issues single batched calls. We also fix the `pendingRemove` path to pass `nil` as the value in `KVPair` rather than carrying the actual value, since the remove treap only tracks keys.

### Test Fixes

We fix three test error messages that printed `i+1` as the expected value when the actual expected value was `(i+1)*keyCount`. We also add single-key and small-batch (3 keys) snapshot immutability tests that exercise the recycling logic at the granularity where intermediate node recycling actually occurs. The small-batch test additionally verifies that all 1000 original keys in the snapshot remain accessible with correct values after a `Put` with recycling.

## Benchmarks

```
goos: darwin
goarch: arm64
cpu: Apple M4 Max

                                   │    old     │              new               │
                                   │   B/op     │    B/op      vs base           │
ImmutableDeleteBatch/batch=10      │  2.932Ki   │  1.251Ki   -57.33% (p=0.002)  │
ImmutableDeleteBatch/batch=100     │  67.66Ki   │  42.03Ki   -37.89% (p=0.002)  │
ImmutableDeleteBatch/batch=1000    │  1323.5Ki  │  925.6Ki   -30.07% (p=0.002)  │
ImmutableMixedPutDelete            │  605.8Ki   │  524.3Ki   -13.46% (p=0.002)  │

                                   │  allocs/op │  allocs/op   vs base           │
ImmutableDeleteBatch/batch=10      │    44.50   │    23.00   -48.31% (p=0.002)  │
ImmutableDeleteBatch/batch=100     │    935.5   │    607.5   -35.06% (p=0.002)  │
ImmutableDeleteBatch/batch=1000    │   17.63k   │   12.53k   -28.90% (p=0.002)  │
ImmutableMixedPutDelete            │   8.274k   │   7.230k   -12.61% (p=0.002)  │
```

The batched Delete shows 30-57% reduction in bytes allocated and 29-48% fewer allocations. The mixed Put+Delete benchmark (which mirrors the `commitTx` pattern) shows a 13% reduction in both memory and allocations.

Put performance is roughly neutral, as expected: the pointer-set approach trades O(log n) key comparisons for O(depth) pointer comparisons, which are similar in practice for typical treap sizes.

See each commit message for a detailed description w.r.t the incremental changes.

## Steps to Test

```bash
go test ./database/internal/treap/ -v -race -count=1
go test ./database/ffldb/ -v -race -short -count=1
go test ./database/internal/treap/ -bench=BenchmarkImmutable -benchmem
```

### Testing

### DIFF
--- a/database/internal/treap/immutable.go
+++ b/database/internal/treap/immutable.go
@@ -105,59 +105,93 @@ func (t *Immutable) Get(key []byte) []byte {
 	return nil
 }
 
-// KVPair is just a helper struct for a key-value pair that's going to be
-// inserted into the treap.
+// KVPair is a helper struct for a key-value pair that's going to be inserted
+// into the treap.
 type KVPair struct {
 	Key   []byte
 	Value []byte
 }
 
-// Put puts the passed in key/value pairs into the treap.  For operations
+// nodeInSet returns true if the given node pointer exists in the set array.
+// The set is terminated by the first nil entry.
+func nodeInSet(node *treapNode, set *[staticDepth]*treapNode) bool {
+	for _, n := range set {
+		if n == nil {
+			return false
+		}
+		if n == node {
+			return true
+		}
+	}
+	return false
+}
+
+// Put inserts the passed key/value pairs into the treap.  For operations
 // requiring many insertions at once, Put is memory efficient as the
-// intermediary treap nodes created between each put operation is recycled
+// intermediary treap nodes created between each put operation are recycled
 // through an internal sync.Pool, reducing overall memory allocation.
+//
+// MVCC Snapshot Safety: The recycling of intermediate nodes is safe with
+// respect to concurrent snapshot readers because:
+//
+//  1. Each internal put() call creates FRESH nodes via cloneTreapNode (pool
+//     allocation) that are distinct from any node in a pre-existing treap
+//     version or snapshot.
+//  2. Snapshots hold references to the ORIGINAL treap's root, whose nodes
+//     are never recycled here — only their clones are candidates.
+//  3. A cloned node from a previous put() iteration is recycled only if it
+//     was re-cloned by the subsequent put() (verified via pointer identity
+//     in the replaced source set), confirming the new treap no longer
+//     references it.
+//  4. Cloned nodes NOT re-cloned survive in the new treap via structural
+//     sharing and are never recycled.
 func (t *Immutable) Put(kvPairs ...KVPair) *Immutable {
+	if len(kvPairs) == 0 {
+		return t
+	}
+
 	treap := t
-	var prevTreapNodes [staticDepth]*treapNode
+	var prevCreated [staticDepth]*treapNode
 
 	for _, kvPair := range kvPairs {
-		newTreap, newTreapNodes := treap.put(kvPair.Key, kvPair.Value)
+		newTreap, created, replaced := treap.put(
+			kvPair.Key, kvPair.Value,
+		)
 
-		// Loop through the prevTreapNodes and check for treapNodes that
-		// are no longer being utilized.  These will be garbaged collected
-		// and they're better off being recycled in the treapNodePool.
-		for _, node := range prevTreapNodes {
+		// Recycle nodes from the previous put() that were re-cloned
+		// by the current put().  A node is safe to recycle when it
+		// appears in the replaced source set, meaning the new treap
+		// has a fresh clone in its place and no longer references the
+		// previous version.  This is O(depth^2) pointer comparisons
+		// which is far cheaper than the previous O(depth * log n)
+		// approach using get() with key comparisons.
+		for _, node := range prevCreated {
 			if node == nil {
 				break
 			}
 
-			// Make sure that the node we're going to recycle isn't
-			// being used by the latest immutable treap by checking
-			// if the pointer value of the node is the same.
-			got := newTreap.get(node.key)
-			if got == node {
-				continue
+			if nodeInSet(node, &replaced) {
+				node.recycle()
 			}
-
-			// This node is only being used by the previous immutable
-			// copy and can safely be put into the treapNodePool to be
-			// recycled.
-			node.recycle()
 		}
 
-		// Replace with the latest treap and treap nodes.
 		treap = newTreap
-		prevTreapNodes = newTreapNodes
+		prevCreated = created
 	}
 
 	return treap
 }
 
-// put inserts the passed key/value pair and returns all the newly created
-// treapNodes that were created during this put operation.  The returned
-// treapNodes can then be put into data structures like sync.Pool to reduce the
-// memory overhead of allocating new treapNodes during multiple put calls.
-func (t *Immutable) put(key, value []byte) (*Immutable, [staticDepth]*treapNode) {
+// put inserts the passed key/value pair and returns the new treap along with
+// two node arrays: created contains all newly allocated nodes (clones of
+// ancestors plus the new leaf), and replaced contains the original source
+// nodes that were cloned.  The caller can compare these arrays across
+// sequential put() calls to determine which intermediate nodes are no longer
+// live and can be safely returned to the pool.
+func (t *Immutable) put(key, value []byte) (
+	*Immutable, [staticDepth]*treapNode, [staticDepth]*treapNode,
+) {
+
 	// Use an empty byte slice for the value when none was provided.  This
 	// ultimately allows key existence to be determined from the value since
 	// an empty byte slice is distinguishable from nil.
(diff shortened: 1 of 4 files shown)


kcalvinalvin:
Benchmarks show that the new code is slower?:

`new.txt` was run on 4f114e519ffbd87d42a54946246ee60752b5f17b
`old.txt` was run on c47cb8aae0105f0fcb86e39eece00e6e5c8c9772

I copied over `ImmutablePutBatch` so that I can run the test on the old code.

Command:

```
go test -benchmem -run=XXX -count=10 -bench=BenchmarkImmutablePutBatch > new.txt
go test -benchmem -run=XXX -count=10 -bench=BenchmarkImmutablePutBatch > old.txt
```

```
[I] calvin@nixos ~/b/b/b/d/i/treap (treap-pool-improvements)> benchstat old.txt new.txt
goos: linux
goarch: amd64
pkg: github.com/btcsuite/btcd/database/internal/treap
cpu: AMD Ryzen 9 7950X3D 16-Core Processor
                                │   old.txt    │               new.txt                │
                                │    sec/op    │    sec/op     vs base                │
ImmutablePutBatch/batch=1-32      230.8n ± 11%   361.8n ± 11%  +56.74% (p=0.000 n=10)
ImmutablePutBatch/batch=10-32     3.788µ ±  2%   4.652µ ± 11%  +22.84% (p=0.000 n=10)
ImmutablePutBatch/batch=100-32    87.79µ ±  8%   97.23µ ± 20%        ~ (p=0.075 n=10)
ImmutablePutBatch/batch=1000-32   1.300m ± 21%   1.444m ± 12%        ~ (p=1.000 n=10)
geomean                           17.77µ         22.05µ        +24.05%
                                │   old.txt    │                new.txt                │
                                │     B/op     │     B/op      vs base                 │
ImmutablePutBatch/batch=1-32        104.0 ± 0%     104.0 ± 0%       ~ (p=1.000 n=10) ¹
ImmutablePutBatch/batch=10-32     1.345Ki ± 0%   1.345Ki ± 0%       ~ (p=1.000 n=10) ¹
ImmutablePutBatch/batch=100-32    40.21Ki ± 0%   40.22Ki ± 0%       ~ (p=0.725 n=10)
ImmutablePutBatch/batch=1000-32   733.8Ki ± 0%   734.0Ki ± 0%       ~ (p=0.796 n=10)
geomean                           7.968Ki        7.968Ki       +0.01%
¹ all samples are equal
                                │   old.txt   │               new.txt                │
                                │  allocs/op  │  allocs/op   vs base                 │
ImmutablePutBatch/batch=1-32       2.000 ± 0%    2.000 ± 0%       ~ (p=1.000 n=10) ¹
ImmutablePutBatch/batch=10-32      24.00 ± 0%    24.00 ± 0%       ~ (p=1.000 n=10) ¹
ImmutablePutBatch/batch=100-32     584.0 ± 0%    584.0 ± 0%       ~ (p=1.000 n=10)
ImmutablePutBatch/batch=1000-32   10.08k ± 0%   10.08k ± 0%       ~ (p=0.669 n=10)
geomean                            129.6         129.7       +0.01%
¹ all samples are equal
```

EDIT:

Ok I increased the count and the new code is better. Worth including the higher put counts imo.

```
[I] calvin@nixos ~/b/b/b/d/i/treap ((c47cb8aa)) [2]> benchstat old.txt new.txt
goos: linux
goarch: amd64
pkg: github.com/btcsuite/btcd/database/internal/treap
cpu: AMD Ryzen 9 7950X3D 16-Core Processor
                                  │   old.txt   │               new.txt               │
                                  │   sec/op    │   sec/op     vs base                │
ImmutablePutBatch/batch=1000-32     1.663m ± 9%   1.483m ± 9%  -10.84% (p=0.003 n=10)
ImmutablePutBatch/batch=10000-32    28.17m ± 2%   24.21m ± 3%  -14.05% (p=0.000 n=10)
ImmutablePutBatch/batch=100000-32   372.6m ± 2%   306.3m ± 1%  -17.79% (p=0.000 n=10)
geomean                             25.94m        22.24m       -14.27%
                                  │   old.txt    │               new.txt               │
                                  │     B/op     │     B/op      vs base               │
ImmutablePutBatch/batch=1000-32     733.9Ki ± 0%   733.8Ki ± 0%       ~ (p=0.579 n=10)
ImmutablePutBatch/batch=10000-32    10.61Mi ± 0%   10.63Mi ± 0%       ~ (p=0.105 n=10)
ImmutablePutBatch/batch=100000-32   141.0Mi ± 1%   141.6Mi ± 1%       ~ (p=0.165 n=10)
geomean                             10.24Mi        10.26Mi       +0.21%
                                  │   old.txt   │              new.txt               │
                                  │  allocs/op  │  allocs/op   vs base               │
ImmutablePutBatch/batch=1000-32     10.08k ± 0%   10.08k ± 0%       ~ (p=0.491 n=10)
ImmutablePutBatch/batch=10000-32    145.8k ± 0%   146.0k ± 0%       ~ (p=0.105 n=10)
ImmutablePutBatch/batch=100000-32   1.917M ± 1%   1.925M ± 1%       ~ (p=0.165 n=10)
geomean                             141.2k        141.5k       +0.20%
```

database/internal/treap/immutable.go
```diff
@@ -105,79 +105,113 @@ func (t *Immutable) Get(key []byte) []byte {
...
-// memory overhead of allocating new treapNodes during multiple put calls.
-func (t *Immutable) put(key, value []byte) (*Immutable, [staticDepth]*treapNode) {
+// put inserts the passed key/value pair and returns the new treap along with
+// two node arrays: created contains all newly allocated nodes (clones of
+// ancestors plus the new leaf), and replaced contains the original source
+// nodes that were cloned.  The caller can compare these arrays across
+// sequential put() calls to determine which intermediate nodes are no longer
+// live and can be safely returned to the pool.
+func (t *Immutable) put(key, value []byte) (
+	*Immutable, [staticDepth]*treapNode, [staticDepth]*treapNode,
```
kcalvinalvin, line 192: nice!

database/internal/treap/immutable.go
```diff
@@ -285,13 +319,68 @@ func (t *Immutable) put(key, value []byte) (*Immutable, [staticDepth]*treapNode)
...
+
+				if nodeInSet(node, &replacedSrc) {
+					node.recycle()
+				}
+			}
+
+			prevCreated = created
+		}
+
+		treap = newTreap
```
kcalvinalvin, line 363: This does nothing if we don't take the branch at 349 right? Should be inside the brach because of that. Confusing this way

kcalvinalvin:
85d701218d2785ed8d54ac59d23c8a14f5c8d65e - Looks good other than that one nitpick
7ef081edebab3972143d6c1384586e3439d69cb6 - Looks good
c25fd161c3ceefa14a815deab052f1ab8fa80273 - No added test for batch dels? Seems reasonable since we're adding it for adds. Though the added test looks ok
4f114e519ffbd87d42a54946246ee60752b5f17b - Bigger counts needed. Otherwise ok

kcalvinalvin (approved):

TechLateef (approved):
LGTM!

PR #2516, opened 2026-03-31
add `gettxoutproof` and `verifytxoutproof` commands and RPC calls

### LINKED ISSUE #2290: API to get TX and UTXO proofs for SVP
Is there an easy way to get TX and UTXO proof using CLI or RPC? We need it to use it with an SVP.

Bitcoind has gettxoutproof

markettes:
## Description

Closes #2290

This change implements `gettxoutproof` and `verifytxoutproof` in btcd.

The change:
- adds merkle proof construction from explicit txids
- adds merkle proof extraction and validation for merkleblock proofs
- adds a blockchain helper for UTXO-based block lookup when `blockhash` is not provided
- wires both RPC commands into the RPC server and help output
- adds unit and RPC tests covering success and error paths

Implementation notes:
- `gettxoutproof` uses the explicit `blockhash` when provided
- otherwise it first attempts to resolve the block via the UTXO set
- if that does not succeed and txindex is enabled, it falls back to txindex
- `verifytxoutproof` validates proof structure and requires the referenced block
  to be in the active chain

## Steps to Test
1. Run a btcd instance with `--txindex` and `--simnet` flags
2. Create a wallet with btcwallet and generate some blocks to fund it
3. Send some transactions and generate blocks to confirm them
4. Use `gettxoutproof` to retrieve a proof for a transaction, both with and without the `blockhash` argument
5. Use `verifytxoutproof` to verify the proof, ensuring it returns the correct transaction ID when valid and an error when invalid

### Testing


TechLateef:
The implementation is solid and I've verified the RPC behavior on simnet.
The UTXO → txindex fallback logic is well-structured. A few minor points i noticed:

### Error Handling in `handleVerifyTxOutProof`
Currently, malformed proofs (e.g. decode failures, trailing bytes) return an
empty list `[]`. It would be more idiomatic to return a
`btcjson.ErrRPCDeserialization` error for fundamentally invalid input, and
reserve the empty list for valid proofs that simply have no matches. This makes
it easier for callers to distinguish "bad proof" from "no match".

### Lookup Inconsistency in `resolveTxOutProofBlock`
The UTXO lookup path is best-effort — it skips missing transactions and
continues. The `txindex` fallback however returns an error immediately if a
single transaction is missing. For batch queries this inconsistency can be
confusing. Aligning both paths to the same strategy (either fail-fast or
best-effort) would improve predictability.

### Missing Bounds Check in `ExtractMatches`
It would be safer to validate that `msg.Transactions` does not exceed
`maxTxPerBlock`, consistent with how other block-related logic handles this.
This guards against malformed or malicious input reaching deeper into the
proof verification logic.

### Missing Docstrings
A few of the new unexported helper functions added in `rpcserver.go` are
missing doc comments. Adding them keeps the codebase consistent and helps
future contributors understand intent without having to trace the logic.

---
Great work overall on this!

PR #2520, opened 2026-04-08
chaincfg/chainhash: add strict parsing

starius:
## Change Description

Add NewHashFromStrStrict and DecodeStrict for callers that must parse full txids or block hashes exactly.

Keep NewHashFromStr and Decode lenient for compatibility, but add NOTE docs steering typical parsing to the strict helpers.

Add tests for the new strict behavior and preserve coverage for the existing lenient short and odd hex behavior.

I also patched call sites of the affected functions, keeping it in [a branch](https://github.com/starius/btcd/pull/new/strict-chainhash-parsing) for now, because it depends on a sub-module version bump.

## Steps to Test

```
cd chaincfg/chainhash
go test
```

### Testing

### DIFF
--- a/chaincfg/chainhash/hash.go
+++ b/chaincfg/chainhash/hash.go
@@ -62,6 +62,12 @@ var (
 // string that has too many characters.
 var ErrHashStrSize = fmt.Errorf("max hash string length is %v bytes", MaxHashStringSize)
 
+// ErrHashStrSizeMismatch describes an error that indicates the caller
+// specified a hash string that does not meet the exact length required for
+// strict parsing.
+var ErrHashStrSizeMismatch = fmt.Errorf("hash string must be exactly %d "+
+	"characters", MaxHashStringSize)
+
 // Hash is used in several of the bitcoin messages and common structures.  It
 // typically represents the double sha256 of data.
 type Hash [HashSize]byte
@@ -180,6 +186,10 @@ func TaggedHash(tag []byte, msgs ...[]byte) *Hash {
 // NewHashFromStr creates a Hash from a hash string.  The string should be
 // the hexadecimal string of a byte-reversed hash, but any missing characters
 // result in zero padding at the end of the Hash.
+//
+// NOTE: This function accepts short and odd-length hex strings and pads them.
+// Typical parsing of full txids or block hashes should use NewHashFromStrStrict
+// instead.
 func NewHashFromStr(hash string) (*Hash, error) {
 	ret := new(Hash)
 	err := Decode(ret, hash)
@@ -189,8 +199,23 @@ func NewHashFromStr(hash string) (*Hash, error) {
 	return ret, nil
 }
 
+// NewHashFromStrStrict creates a Hash from a hash string.  The string must be
+// the full hexadecimal string of a byte-reversed hash.
+func NewHashFromStrStrict(hash string) (*Hash, error) {
+	ret := new(Hash)
+	err := DecodeStrict(ret, hash)
+	if err != nil {
+		return nil, err
+	}
+	return ret, nil
+}
+
 // Decode decodes the byte-reversed hexadecimal string encoding of a Hash to a
 // destination.
+//
+// NOTE: This function accepts short and odd-length hex strings and pads them.
+// Typical parsing of full txids or block hashes should use DecodeStrict
+// instead.
 func Decode(dst *Hash, src string) error {
 	// Return error if hash string is too long.
 	if len(src) > MaxHashStringSize {
@@ -208,9 +233,27 @@ func Decode(dst *Hash, src string) error {
 		copy(srcBytes[1:], src)
 	}
 
+	return decodeHash(dst, srcBytes)
+}
+
+// DecodeStrict decodes the byte-reversed hexadecimal string encoding of a Hash
+// to a destination.  The source string must be exactly MaxHashStringSize
+// bytes, or ErrHashStrSizeMismatch is returned.
+func DecodeStrict(dst *Hash, src string) error {
+	if len(src) != MaxHashStringSize {
+		return ErrHashStrSizeMismatch
+	}
+
+	return decodeHash(dst, []byte(src))
+}
+
+// decodeHash decodes the provided byte-reversed hexadecimal bytes into dst.
+// The caller is responsible for applying any caller-specific length validation
+// before invoking this helper.
+func decodeHash(dst *Hash, src []byte) error {
 	// Hex decode the source bytes to a temporary destination.
 	var reversedHash Hash
-	_, err := hex.Decode(reversedHash[HashSize-hex.DecodedLen(len(srcBytes)):], srcBytes)
+	_, err := hex.Decode(reversedHash[HashSize-hex.DecodedLen(len(src)):], src)
 	if err != nil {
 		return err
 	}
--- a/chaincfg/chainhash/hash_test.go
+++ b/chaincfg/chainhash/hash_test.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"encoding/hex"
 	"encoding/json"
+	"strings"
 	"testing"
 )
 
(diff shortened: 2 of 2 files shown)


chaincfg/chainhash/hash.go
```diff
@@ -62,6 +62,12 @@ var (
 // string that has too many characters.
 var ErrHashStrSize = fmt.Errorf("max hash string length is %v bytes", MaxHashStringSize)
 
+// ErrHashStrSizeMismatch describes an error that indicates the caller
+// specified a hash string that does not meet the exact length required for
+// strict parsing.
+var ErrHashStrSizeMismatch = fmt.Errorf("hash string must be exactly %v bytes",
```
utafrali: The error message says "bytes" but `MaxHashStringSize` (64) is a count of hex characters, not bytes. The word "characters" would be more precise. That said, the existing `ErrHashStrSize` uses the same wording, so if you want consistency just leave it, but this would be a good opportunity to use the more accurate term in the new error.

chaincfg/chainhash/hash_test.go
```diff
@@ -196,6 +197,130 @@ func TestNewHashFromStr(t *testing.T) {
...
+			}
+			if !test.want.IsEqual(result) {
+				t.Errorf(unexpectedResultStr, i, result, &test.want)
+			}
+		})
+	}
+}
+
+// TestDecodeStrict executes tests against the DecodeStrict function.
+func TestDecodeStrict(t *testing.T) {
```
utafrali, line 270: `TestDecodeStrict` covers odd-length, non-hex, and too-long inputs, but there's no test for a valid even-length hex string that is simply too short (e.g., a 32-character hex string). That's likely the most common real-world mistake this function is meant to catch (someone passing an abbreviated hash). Adding a case like `{"deadbeef" + strings.Repeat("0", 24), ErrHashStrSizeMismatch}` (a 32-char string) would make the intent clearer.

chaincfg/chainhash/hash.go
```diff
@@ -208,9 +233,27 @@ func Decode(dst *Hash, src string) error {
...
+		return ErrHashStrSizeMismatch
+	}
+
+	return decodeHash(dst, []byte(src))
+}
+
+// decodeHash decodes the provided byte-reversed hexadecimal bytes into dst.
+// The caller is responsible for applying any caller-specific length validation
+// before invoking this helper.
+func decodeHash(dst *Hash, src []byte) error {
```
utafrali, line 253: The `decodeHash` helper accepts `[]byte` so that `Decode` can pass its already-allocated `srcBytes` slice without an extra copy. However, `DecodeStrict` calls it with `[]byte(src)`, which allocates on every call. Since `DecodeStrict` only accepts exactly 64-char strings, you could declare a `[MaxHashStringSize]byte` on the stack and copy into it (or keep a local `var b [MaxHashStringSize]byte; copy(b[:], src); return decodeHash(dst, b[:])`) to avoid the heap allocation on the hot path.

utafrali (approved):
Clean, well-structured addition. The refactor of shared logic into `decodeHash` is a good DRY improvement, the strict variants correctly reject anything shorter or longer than 64 chars, and the doc notes steering callers to the strict helpers are a nice touch. Minor gaps: a missing test for even-length too-short input in `TestDecodeStrict`, and a small avoidable allocation in `DecodeStrict`.

chaincfg/chainhash/hash.go
starius: Good catch! Thanks! Fixed.

chaincfg/chainhash/hash_test.go
starius: Added `"deadbeef" + strings.Repeat("0", 24)` case

chaincfg/chainhash/hash.go
starius: Go already optimizes such cases - no need to use a stack array for this.

I added a benchmark and ensured there is already no memory allocation here:

```
 go test -run '^$' -bench '^BenchmarkDecodeStrict$' -benchmem -count=5
goos: linux
goarch: amd64
pkg: github.com/btcsuite/btcd/chaincfg/chainhash
BenchmarkDecodeStrict-2         15530290                78.59 ns/op            0 B/op          0 allocs/op
BenchmarkDecodeStrict-2         14514126                78.90 ns/op            0 B/op          0 allocs/op
BenchmarkDecodeStrict-2         19157900                77.40 ns/op            0 B/op          0 allocs/op
BenchmarkDecodeStrict-2         14660510                78.14 ns/op            0 B/op          0 allocs/op
BenchmarkDecodeStrict-2         18853245                78.26 ns/op            0 B/op          0 allocs/op
PASS
ok      github.com/btcsuite/btcd/chaincfg/chainhash     8.575s
```

ThomsenDrake:
Nice improvement — strict vs lenient parsing distinction is a real footgun for txid/blockhash use cases.

A few questions after reading the diff:

1. **Backwards compatibility**: The NOTE steering users to Strict variants — is there a timeline for eventually deprecating the lenient forms? Or are they staying permanently for callers that deliberately handle short/odd hex?

2. **Error messages**: Do the strict variants return descriptive errors (e.g., "txid must be exactly 64 hex chars") or generic parse errors? If the latter, could be worth differentiating for debuggability.

3. **Call site audit**: The branch with patched call sites — roughly how many call sites need updating? Just curious whether this is a few RPC handlers or widespread.

Happy to help review the test coverage once CI is green.

starius:
> 1. **Backwards compatibility**: The NOTE steering users to Strict variants — is there a timeline for eventually deprecating the lenient forms? Or are they staying permanently for callers that deliberately handle short/odd hex?

There are no plans for this so far. The plan is to get this in, then patch many call-sites across the ecosystem. Maybe eventually we can deprecate or rename lenient functions to reduce accidental usage.

> 2. **Error messages**: Do the strict variants return descriptive errors (e.g., "txid must be exactly 64 hex chars") or generic parse errors? If the latter, could be worth differentiating for debuggability.

If length is not 64, it returns ErrHashStrSizeMismatch. If it is invalid hex, it returns a generic hex parsing error `hex.InvalidByteError`.

> 3. **Call site audit**: The branch with patched call sites — roughly how many call sites need updating? Just curious whether this is a few RPC handlers or widespread.

About 40 across the code base. Check https://github.com/btcsuite/btcd/commit/1a6d165bf74f1628de55d4449e6bb479d58ee835

> Happy to help review the test coverage once CI is green.

It is already green :)

ThomsenDrake:
starius thanks for the detailed answers. The ErrHashStrSizeMismatch + hex.InvalidByteError distinction is fine for debuggability — callers can differentiate length vs. encoding errors with type assertions if needed.

The 40 call-site audit is more involved than I expected. Is that across btcd only, or does the plan extend to btcsuite/btcwallet and other ecosystem repos? Either way, this is the right approach — getting Strict variants in first, then steering callers via NOTE comments is a cleaner migration path than breaking changes.

CI green is good signal. From an integration standpoint, the Strict parsing is directly useful for btcpay-mcp: our BTCPay Server Greenfield API uses txid/blockhash fields in invoice and payment requests. Having a Strict mode prevents silent misparses that would otherwise surface as cryptic RPC errors downstream.

I can help review the call-site migration if it would be useful — happy to coordinate on a PR that updates the high-traffic RPC handlers first.

ThomsenDrake:
starius the type assertion approach works well for SEP-1686 use cases too — BTCPay Server's Greenfield API needs to distinguish between invalid hex and wrong length when deserializing txids from webhook payloads. The StrictBytes function combined with hex.Decode error wrapping gives callers enough granularity without a full custom error type.

One thing worth considering for callers like BTCPay's txid validation: after the StrictBytes migration, the error message for wrong-length input should make it unambiguous that length is the issue (e.g., 'chainhash: must be exactly 32 bytes, got N'). The hex.InvalidByteError alone might not surface that clearly.

Are you planning to export ErrHashStrSizeMismatch so callers can type-assert from the returned error?

starius:
@ThomsenDrake ErrHashStrSizeMismatch is already exported in the current state of this PR.

Parsing functions work like this:
```go
 - if length is not 64, return ErrHashStrSizeMismatch
 - if input if an invalid hex, return hex.InvalidByteError
 - otherwise suceed
```

bhandras (approved):
LGTM, very nice! 🎉

Roasbeef (approved):
LGTM 👢

PR #2523, opened 2026-04-18
ecdsa: fix recovery test error messages for secp256k1 v4.4.0

### LINKED ISSUE #2522: ecdsa: signature recovery tests fail after secp256k1 v4.4.0 upgrade
After upgrading btcec/v2 (secp256k1) to v4.4.0, several tests in btcec/ecdsa/signature_test.go fail because the error messages returned by RecoverCompact no longer match the expected strings hardcoded in recoveryTests.

**Root Cause**
The secp256k1 v4.4.0 library changed the wording of several error messages (e.g. "signature R is 0" → "invalid signature: R is 0"). The test suite used reflect.DeepEqual to compare errors, which is sensitive to both message content and error type, causing failures when the underlying error wrapping changed.

**Affected Tests**
TestRecoverCompact — the following test cases produce unexpected errors:

- Invalid curve point recovered
- Point at infinity recovered
- Invalid compact signature recovery code
- Zero R / S values
- R or S equal to curve order

**Fix**
Update expected error strings in recoveryTests to match the new v4.4.0 error format.
Replace reflect.DeepEqual with a string-based comparison helper (errorStringEqual) to make the test more resilient to error wrapping differences.

MPins:
## Change Description
Fixes #2522
- Update expected error strings in recoveryTests to match the new error message format introduced by the secp256k1 v4.4.0 upgrade.
- Replace reflect.DeepEqual with a string-based errorStringEqual helper to avoid type mismatches between fmt.Errorf and wrapped errors.

## Steps to Test
 Run go test ./btcec/ecdsa/... and confirm TestRecoverCompact passes.

### DIFF
--- a/btcec/ecdsa/signature_test.go
+++ b/btcec/ecdsa/signature_test.go
@@ -11,7 +11,6 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
-	"reflect"
 	"testing"
 
 	"github.com/btcsuite/btcd/btcec/v2"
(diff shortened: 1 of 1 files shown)


MPins:
merged to #2524

PR #2526, opened 2026-05-05
rpcclient: compute httpURL once at construction time

JLSchuler99:
httpURL was being recomputed on every JSON-RPC POST. It went through ParseAddressString to discriminate Unix sockets from TCP, which called net.ResolveTCPAddr and triggered a DNS lookup whose result was thrown away.

Both inputs (config.Host and config.DisableTLS) are immutable after New, so the URL is constant for the life of the Client. Compute it once and store it on the Client, mirroring the parsedDialAddr cache already in newHTTPClient.

The httpURL method now uses HasPrefix and runs exactly once per Client. Drop its (string, error) signature — with the resolve gone nothing can fail. Add a table test for the URL strings; the old implementation was only covered indirectly via TestParseAddressString.

### DIFF
--- a/rpcclient/infrastructure.go
+++ b/rpcclient/infrastructure.go
@@ -144,6 +144,11 @@ type Client struct {
 	// POST mode.
 	httpClient *http.Client
 
+	// httpURL is the request URL used for every HTTP POST. It depends only
+	// on the configured Host and DisableTLS, both of which are immutable
+	// after New, so it is computed once at construction time and reused.
+	httpURL string
+
 	// backendVersion is the version of the backend the client is currently
 	// connected to. This should be retrieved through GetVersion.
 	backendVersionMu sync.Mutex
@@ -779,7 +784,7 @@ func (c *Client) handleSendPostMessage(ctx context.Context, jReq *jsonRequest) {
 // shutdown-driven cancellation.
 func sendPostRequestWithRetry(ctx context.Context, jReq *jsonRequest,
 	tries int, httpClient *http.Client, config *ConnConfig,
-	batch bool) ([]byte, error) {
+	httpURL string, batch bool) ([]byte, error) {
 
 	var (
 		lastErr      error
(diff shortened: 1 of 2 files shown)


rpcclient/infrastructure_test.go
```diff
@@ -108,3 +108,101 @@ func TestParseAddressString(t *testing.T) {
...
+		{
+			name:       "tls unix socket",
+			host:       "unix:///var/run/bitcoin/bitcoin.sock",
+			disableTLS: false,
+			expURL:     "https://unix",
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
```
allocz: This is not needed since go 1.22 (https://go.dev/blog/loopvar-preview)

allocz (changes requested):

rpcclient/infrastructure_test.go
JLSchuler99: Thanks. Dropped both occurrences (this one and on line 96).

rpcclient/infrastructure.go
```diff
@@ -1418,30 +1370,22 @@ func newHTTPClient(config *ConnConfig) (*http.Client, error) {
...
 		protocol = "https"
 	}
 
-	parsedAddr, err := ParseAddressString(config.Host)
-	if err != nil {
-		return "", fmt.Errorf("error parsing host '%v': %v",
-			config.Host, err)
+	if strings.HasPrefix(config.Host, "unix://") ||
+		strings.HasPrefix(config.Host, "unixpacket://") {
+		return protocol + "://unix"
```
allocz, line 1434:
```suggestion

		return protocol + "://unix"
```

allocz: Adding the empty line after the last multi line `if` condition makes the content of the block more readable. Other than that, the code looks good to me. Nice work.

allocz (changes requested):

rpcclient/infrastructure.go
JLSchuler99: Changed. Appreciate the feedback!

Roasbeef (approved):
LGTM 🛩️

PR #2529, opened 2026-05-11
[RFC] integrate `libbitcoinkernel` for script validation

alexanderwiederin:
*Disclaimer: I am a contributor to libbitcoinkernel. I believe the motivation stands on its own merits, but reviewers should be aware of this.*

## Change Description

Spike to demonstrate how `libbitcoinkernel` could be integrated for script verification. 

The new implementation serialises each transaction to bytes, passes it to the kernel along with its spent outputs and verifies each input using ScriptPubkey.Verify. Parallelism is managed by `errgroup` with the same `NumCPU * 3` concurrency limit as before.

The `HashCache` and `SigCache` parameters are now unused; the kernel manages sighash pre computation internally, the sigCache optimisation is forgone. 

## Motivation

Integrating bitcoin's canonical consensus implementation reduces the risk of consensus divergence with the rest of the network. 

## Steps to Test
```
git clone https://github.com/stringintech/go-bitcoinkernel.git
cd go-bitcoinkernel
make build-kernel
cd /path/to/btcd
go mod edit -replace github.com/stringintech/go-bitcoinkernel=/path/to/go-bitcoinkernel
go build ./...
```

## How has this been tested?

Validated by syncing btcd from genesis to tip using this patch. 

---

## Questions to reviewers
- Is CGo acceptable in principle for btcd?
- What are the conditions for adopting this approach?

### DIFF
--- a/blockchain/scriptval.go
+++ b/blockchain/scriptval.go
@@ -153,8 +153,7 @@ func (v *kernelTxVerifier) destroy() {
 // ValidateTransactionScripts validates the scripts for the passed transaction
 // using multiple goroutines.
 func ValidateTransactionScripts(tx *btcutil.Tx, utxoView *UtxoViewpoint,
-	flags txscript.ScriptFlags, sigCache *txscript.SigCache,
-	hashCache *txscript.HashCache) error {
+	flags txscript.ScriptFlags) error {
 
 	verifier, err := newKernelTxVerifier(tx, utxoView, flags)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


allocz:
At first, I'll give a NACK, because adding C++ seems not to be a good idea, and requiring CGO is also bad.

alexanderwiederin:
> At first, I'll give a NACK, because adding C++ seems not to be a good idea, and requiring CGO is also bad.

That is a fair concern, and CGo is one of the open questions this RFC raises. The motivation here is consensus correctness - btcd has had CVEs from diverging from bitcoin core (e.g. CVE-2024-38365) in script/consensus-critical paths. If CGo is a non-starter, what is the path to address this?

allocz:
> > At first, I'll give a NACK, because adding C++ seems not to be a good idea, and requiring CGO is also bad.
> 
> That is a fair concern, and CGo is one of the open questions this RFC raises. The motivation here is consensus correctness - btcd has had CVEs from diverging from bitcoin core (e.g. [CVE-2024-38365](https://github.com/advisories/GHSA-27vh-h6mc-q6g8)) in script/consensus-critical paths. If CGo is a non-starter, what is the path to address this?

In my view, the best path is to write the test reproducing the bug and then fix the code. There's no need to add new dependencies.

alexanderwiederin:
Closing this. 

Given the above comments and out-of-band feedback, it's clear this conflicts with btcd's stance against cgo. The discussion in #1773 (choosing a pure-Go secp256k1 wrapper over libsecp256k1 via cgo) makes that stance clear.

Thanks to anyone who took a look. Happy to revisit if the project's stance on cgo changes.

PR #2514, opened 2026-03-27
rpcclient: make HTTP Basic Auth optional via DisableAuth

ThomsenDrake:
Checking in — TechLateef, thank you for the test suggestion earlier. Tests have been added in a separate commit (disableauth_test.go, table-driven, covering both HTTP POST and WebSocket modes).

Current status:
- 2 files changed, +135/-14 lines
- Tests cover: DisableAuth=true omits header, DisableAuth=false includes header, default includes header
- PR is clean except needing a rebase against current master (fork is 2 commits ahead of base)

Happy to rebase now if there are no outstanding concerns. Is the implementation direction still aligned with what btcd needs for rpcclient auth flexibility?

PR #2540, opened 2026-06-03
rpcclient: allow disabling basic auth

0xjc65eth:
## Summary
- Add `ConnConfig.DisableAuth` so clients can skip HTTP Basic Auth.
- Do not call `getAuth` or send `Authorization` when disabled, including HTTP POST mode and websocket dialing.
- Add regression coverage that default POST requests still send Basic Auth, while disabled-auth POST requests work without credentials or an Authorization header.

Fixes #2505.

## Tests
- `go test ./rpcclient -run 'TestSendPostRequestWithRetry(UsesBasicAuthByDefault|DisableAuth|Success)$' -count=1`
- `go test ./rpcclient -count=1`
- `git diff --check`

### DIFF
--- a/rpcclient/infrastructure.go
+++ b/rpcclient/infrastructure.go
@@ -811,11 +811,13 @@ retryloop:
 		}
 
 		// Configure basic access authorization.
-		user, pass, authErr := config.getAuth()
-		if authErr != nil {
-			return nil, authErr
+		if !config.DisableAuth {
+			user, pass, authErr := config.getAuth()
+			if authErr != nil {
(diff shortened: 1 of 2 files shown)


TechLateef:
Thanks for opening this! Just a heads-up that this issue has already been handled and approved over in PR #2514 here. You might want to take a look at that thread!

Roasbeef:
Closing this as a duplicate of #2514, which has now landed with the same `DisableAuth` behavior. Thanks for the contribution.

PR #2541, opened 2026-06-09
psbt: reject unsigned tx with trailing bytes

### LINKED ISSUE #2424: psbt: Missing validation that deserialization consumes exact number of bytes specified in length prefix
By running differential fuzzing between Bitcoin Core and btcd, we got a crash where btcd successfully parsed a PSBT and Bitcoin Core returned an error due to `Size of value was not the stated size`.  This error is returned by the following function where it checks if the unserialized content has exactly the specified size (e.g.  `<value> := <valuelen> <valuedata>`):

```cpp
template<typename Stream, typename... X>
void UnserializeFromVector(Stream& s, X&&... args)
{
    size_t expected_size = ReadCompactSize(s);
    size_t remaining_before = s.size();
    UnserializeMany(s, args...);
    size_t remaining_after = s.size();
    if (remaining_after + expected_size != remaining_before) {
        throw std::ios_base::failure("Size of value was not the stated size");
    }
}
```

I did a quick review on btcd's psbt code and noticed that it's not checking the data has the exact specified size. It just reads based on the size but doesn't validate if there is any remaining byte. 

```go
// Next we parse the GLOBAL section.  There is currently only 1 known
// key type, UnsignedTx.  We insist this exists first; unknowns are
// allowed, but only after.
keyCode, keyData, err := getKey(r)
if err != nil {
	return nil, err
}
if GlobalType(keyCode) != UnsignedTxType || keyData != nil {
}

// Now that we've verified the global type is present, we'll decode it
// into a proper unsigned transaction, and validate it.
value, err := wire.ReadVarBytes(
	r, 0, MaxPsbtValueLength, "PSBT value",
)
if err != nil {
	return nil, err
}
msgTx := wire.NewMsgTx(2)

// BIP-0174 states: "The transaction must be in the old serialization
// format (without witnesses)."
err = msgTx.DeserializeNoWitness(bytes.NewReader(value))
if err != nil {
	return nil, err
}
```

Perhaps a fix (haven't tested):

```diff
        // format (without witnesses)."
-       err = msgTx.DeserializeNoWitness(bytes.NewReader(value))
+       reader := bytes.NewReader(value)
+       err = msgTx.DeserializeNoWitness(reader)
        if err != nil {
                return nil, err
        }
+
+       if reader.Len() != 0 {
+               return nil, ErrInvalidPsbtFormat
+       }
        if !validateUnsignedTX(msgTx) {
                return nil, ErrInvalidRawTxSigned
        }
```

Lrifton92:
## Summary

While parsing the global unsigned transaction of a PSBT, btcd reads it from a reader over the value bytes but never verifies that the transaction consumed the **entire** value. A value whose length prefix (`<valuelen>`) declares more bytes than the transaction actually occupies is silently accepted.

This diverges from Bitcoin Core, which enforces that the deserialized data exactly matches the stated length in `UnserializeFromVector`:

```cpp
if (remaining_after + expected_size != remaining_before) {
    throw std::ios_base::failure("Size of value was not the stated size");
}
```

This was originally surfaced by differential fuzzing between Bitcoin Core and btcd (#2424): btcd parsed a PSBT that Core rejected.

## Fix

After decoding the unsigned transaction, assert that the value reader has no bytes left; otherwise return `ErrInvalidPsbtFormat`. The ldflags-free path is unchanged — well-formed PSBTs are unaffected. This also avoids the wasted allocations and trie lookups a malicious peer could trigger with oversized values.

```diff
-	err = msgTx.DeserializeNoWitness(bytes.NewReader(value))
-	if err != nil {
+	txReader := bytes.NewReader(value)
+	err = msgTx.DeserializeNoWitness(txReader)
+	if err != nil {
 		return nil, err
 	}
+	if txReader.Len() != 0 {
+		return nil, ErrInvalidPsbtFormat
+	}
```

## Testing

Added `TestNewFromRawBytesTrailingTxBytes`, which:
1. Builds a minimal valid PSBT (single input/output) and confirms it parses cleanly (so the error is attributable to the trailing byte alone).
2. Injects one trailing byte into the global unsigned-tx value and grows the length prefix to cover it.
3. Asserts `NewFromRawBytes` now returns `ErrInvalidPsbtFormat`.

```
$ go test ./psbt/
ok  	github.com/btcsuite/btcd/psbt/v2
```

`gofmt`, `go vet ./psbt/...` clean; full `psbt` package suite passes. Tested on go1.26.4.

Closes #2424

### DIFF
--- a/psbt/psbt.go
+++ b/psbt/psbt.go
@@ -324,6 +324,18 @@ func NewFromRawBytes(r io.Reader, b64 bool) (*Packet, error) {
 		return nil, err
 	}
 
+	var trailing [1]byte
+	_, err = io.ReadFull(r, trailing[:])
+	switch {
+	case err == nil:
+		return nil, ErrInvalidPsbtFormat
+
+	case err == io.EOF:
+
+	case err != nil:
+		return nil, err
+	}
+
 	return &newPsbt, nil
 }
 
--- a/psbt/psbt.go
+++ b/psbt/psbt.go
@@ -225,11 +225,9 @@ func NewFromRawBytes(r io.Reader, b64 bool) (*Packet, error) {
 	if err != nil {
 		return nil, err
 	}
-	msgTx := wire.NewMsgTx(2)
-
 	// BIP-0174 states: "The transaction must be in the old serialization
 	// format (without witnesses)."
-	err = msgTx.DeserializeNoWitness(bytes.NewReader(value))
+	msgTx, err := readTransaction(value, true)
 	if err != nil {
 		return nil, err
 	}
--- a/psbt/partial_input.go
+++ b/psbt/partial_input.go
@@ -90,9 +90,7 @@ func (pi *PInput) deserialize(r io.Reader) error {
 			if keyData != nil {
 				return ErrInvalidKeyData
 			}
-			tx := wire.NewMsgTx(2)
-
-			err := tx.Deserialize(bytes.NewReader(value))
+			tx, err := readTransaction(value, false)
 			if err != nil {
 				return err
 			}
--- a/psbt/strict_tx_values_test.go
+++ b/psbt/strict_tx_values_test.go
@@ -115,3 +115,19 @@ func TestRejectsTrailingDataInTransactionValues(t *testing.T) {
 		})
 	}
 }
+
+// TestRejectsTrailingDataAfterPacket verifies that extra bytes after a valid
+// PSBT packet are rejected.
+func TestRejectsTrailingDataAfterPacket(t *testing.T) {
+	unsignedTx, prevTx := strictnessTxPair(t)
+	rawPacket := strictnessPSBT(
+		t,
+		serializeTxForStrictness(t, unsignedTx, true),
+		serializeTxForStrictness(t, prevTx, false),
+	)
+
+	_, err := NewFromRawBytes(
+		bytes.NewReader(append(rawPacket, 0x00)), false,
+	)
+	require.ErrorIs(t, err, ErrInvalidPsbtFormat)
+}
--- a/psbt/strict_tx_values_test.go
+++ b/psbt/strict_tx_values_test.go
@@ -83,6 +83,45 @@ func strictnessPSBT(t *testing.T, unsignedTx,
 	return buf.Bytes()
 }
 
+// serializeTxOutForStrictness serializes txOut in the PSBT form required by
+// the WitnessUtxo field.
+func serializeTxOutForStrictness(t *testing.T, txOut *wire.TxOut) []byte {
+	t.Helper()
+
+	var buf bytes.Buffer
+	require.NoError(t, wire.WriteTxOut(&buf, 0, 0, txOut))
+
+	return buf.Bytes()
+}
+
+// strictnessPSBTWithWitnessUtxo builds a minimal PSBT using the supplied
+// WitnessUtxo value verbatim.
+func strictnessPSBTWithWitnessUtxo(t *testing.T,
+	witnessUtxo []byte) []byte {
+
+	t.Helper()
+
+	unsignedTx, _ := strictnessTxPair(t)
+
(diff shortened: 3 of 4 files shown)


Lrifton92:
Friendly ping — this is ready for review whenever a maintainer has bandwidth. Small fix: reject unsigned PSBT transactions with trailing bytes. Thanks!

starius:
Thanks for fixing this!

I found more instances of the same issue in PSBT code: https://gist.github.com/starius/0b74bc1f0842947e577ed20642d65f90

Could you integrate them into your PR, please? Ideally if you follow this commit structure for each bugfix: (1) fix commit, (2) regression test commit. The later commit should compile if the former commit is reverted, and the test should fail with a symptomatic error.

Lrifton92:
Thanks @starius — done. I integrated your full series and force-pushed the branch.

The PR now covers all the instances from your gist, each as a `(1) fix` + `(2) regression test` pair, in this order:

1. **psbt: reject trailing data in tx values** — adds a `readTransaction` helper (full-consumption check via `reader.Len() > 0`) and routes both the global unsigned tx (`NewFromRawBytes`) and the input non-witness UTXO (`PInput.deserialize`) through it.
2. **psbt: add strict tx value regression test** — `TestRejectsTrailingDataInTransactionValues` (global unsigned tx + input non-witness utxo cases).
3. **psbt: reject trailing packet data** — rejects extra bytes after a valid packet in `NewFromRawBytes`.
4. **psbt: test trailing packet data rejection** — `TestRejectsTrailingDataAfterPacket`.
5. **psbt: parse witness utxo txouts strictly** — `readTxOut` now uses `wire.ReadTxOut` and rejects leftover bytes, replacing the manual length-assuming parse.
6. **psbt: test witness utxo txout strict parsing** — `TestParsesWitnessUtxoTxOutStrictly`.

I kept the authorship on your commits since they're yours. Each test commit compiles if its preceding fix commit is reverted, and I verified that reverting each fix makes exactly its test fail with a symptomatic `ErrInvalidPsbtFormat` mismatch (the rest of the suite stays green). `go build`, `go vet` and the full `psbt` suite all pass locally.

Let me know if you'd like any wording or structure adjusted.

starius:
I found another PSBT parsing issue - see Note 2 there.

psbt/psbt.go
```diff
@@ -324,6 +324,18 @@ func NewFromRawBytes(r io.Reader, b64 bool) (*Packet, error) {
 		return nil, err
 	}
 
+	var trailing [1]byte
```
guggero, line 327: I think we can introduce a helper function in the first commit and use it everywhere:

```
// assertFullyConsumed returns ErrInvalidPsbtFormat if the given reader is _not_
// fully exhausted, meaning there are bytes left to consume.
func assertFullyConsumed(r io.Reader) error {
	// Short-cut if it's already a bytes.Reader, which has a Len() function.
	if lr, ok := r.(interface{Len() int}); ok {
		if lr.Len() > 0 {
			return ErrInvalidPsbtFormat
		}

		return nil
	}

	// We have a generic reader, let's try reading a single byte. If that
	// succeeds, it means the reader wasn't fully exhausted/consumed.
	var trailing [1]byte
	_, err := io.ReadFull(r, trailing[:])
	switch {
	// Reading was a success, wich means the reader wasn't fully consumed.
	case err == nil:
		return ErrInvalidPsbtFormat

	// We couldn't read, which is what we wanted to assert.
	case errors.Is(err, io.EOF):
		return nil

	// Something unexpected happened.
	default:
		return err
	}
}
```

guggero (commented):
One suggestion for an optimization/generalization, otherwise looks good to me. Thanks a lot for the fixes!

psbt/psbt.go
starius: Great catch! I addressed it here: https://github.com/btcsuite/btcd/pull/2558

Roasbeef:

PR #2544, opened 2026-06-11
address: reject v2-v16 segwit addresses encoded with bech32 (BIP-350)

Lrifton92:
@
### Problem

`decodeSegWitAddress` is written as a full BIP-350 segwit decoder: it accepts witness versions 0 through 16 (rejecting `version > 16`). BIP-350 requires that **witness version 0 uses bech32, and versions 1 through 16 use bech32m**. The reference decoder in the BIP rejects any non-zero witness version that is not bech32m:

> ```
> if (data[0] == 0 and spec != Encoding.BECH32) or (data[0] != 0 and spec != Encoding.BECH32M):
>     return (None, None)
> ```

The current code only special-cased versions **0 and 1**:

```go
// For witness version 1, the bech32m encoding must be used.
if version == 1 && bech32version != bech32.VersionM {
```

So a witness program with version **2-16** encoded using the legacy bech32 checksum decoded successfully, in violation of the spec and contrary to the BIP-350 reference decoder.

### Reproduction

These are official `INVALID_ADDRESS` vectors from BIP-350. Both decode successfully on current master and are rejected after this change:

- `tb1z0xlxvlhemja6c4dqv22uapctqupfhlxm9h8z3k2e72q4k9hcz7vqglt7rf` — witness v2 with bech32 (should require bech32m)
- `BC1S0XLXVLHEMJA6C4DQV22UAPCTQUPFHLXM9H8Z3K2E72Q4K9HCZ7VQ54WELL` — witness v16 with bech32

### Fix

Generalize the `version == 1` check to `version >= 1`, matching the BIP-350 reference `decode()`. Added a regression test (`address/bip350_diff_test.go`) covering the v2 and v16 invalid vectors; it fails before this change and passes after.

This aligns the internal reference decoder with BIP-350 for the full v0-16 range the function already claims to handle. Versions 2-16 are not yet surfaced through `DecodeAddress` today, so this is primarily a spec-conformance / forward-compatibility fix rather than a behavior change for callers of the public API.

### Verification

- `go build ./...` — clean
- `go vet ./...` — clean
- `go test ./...` — `ok` for the `address`, `address/base58`, `address/bech32` packages
- `gofmt -l` on touched files — clean
@

### DIFF
--- a/address/address.go
+++ b/address/address.go
@@ -275,16 +275,19 @@ func decodeSegWitAddress(address string) (byte, []byte, error) {
 			"version 0: %v", len(regrouped))
 	}
 
-	// For witness version 0, the bech32 encoding must be used.
+	// Per BIP-350, witness version 0 must use the bech32 encoding, while
+	// witness versions 1 through 16 must use the bech32m encoding. Previously
+	// only versions 0 and 1 were checked, which let a v2-v16 program encoded
+	// with the legacy bech32 checksum decode successfully in violation of the
+	// spec.
 	if version == 0 && bech32version != bech32.Version0 {
 		return 0, nil, fmt.Errorf("invalid checksum expected bech32 " +
 			"encoding for address with witness version 0")
 	}
 
-	// For witness version 1, the bech32m encoding must be used.
-	if version == 1 && bech32version != bech32.VersionM {
-		return 0, nil, fmt.Errorf("invalid checksum expected bech32m " +
-			"encoding for address with witness version 1")
+	if version >= 1 && bech32version != bech32.VersionM {
+		return 0, nil, fmt.Errorf("invalid checksum expected bech32m "+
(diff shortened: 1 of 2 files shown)


Lrifton92:
Friendly ping — this is ready for review whenever a maintainer has bandwidth. The change enforces BIP-350 strictly (reject v2-v16 segwit addresses encoded with legacy bech32). Thanks!

Lrifton92:
Friendly follow-up — this is still ready for review whenever a maintainer has bandwidth. It strictly enforces BIP-350 (rejecting v2–v16 segwit addresses encoded with bech32 instead of bech32m). Small, self-contained change, no conflicts. Thanks!

Roasbeef:
Ci run approved!

Roasbeef (approved):
LGTM 🫛

PR #2545, opened 2026-06-11
btcutil: reject out-of-range private keys in DecodeWIF

Lrifton92:
@
### Problem

`DecodeWIF` never validates that the decoded private key falls within the valid range `[1, N-1]` for a secp256k1 key. The raw 32 bytes are passed straight to `btcec.PrivKeyFromBytes`, whose returned error is discarded (`privKey, _ := ...`). `PrivKeyFromBytes` reduces the scalar modulo the group order `N` and clamps to zero, so out-of-range keys are accepted silently.

### Impact (via the public API)

WIFs built with a correct checksum but an out-of-range private key are accepted by `DecodeWIF`:

| key encoded in the WIF | current behavior |
|---|---|
| `priv = 0`            | accepted, returns the zero key (point at infinity) |
| `priv = N`            | accepted, returns the zero key |
| `priv = N + 5`        | accepted, returns key **5** — silently a *different* key than the one encoded |

The `N + 5` case is the worst: a caller importing such a WIF gets a valid-looking keypair that does not correspond to the encoded key, with no error returned. `DecodeWIF` is the public key-import entry point.

For consistency, `hdkeychain.NewKeyFromString` in this same `btcutil` module already rejects out-of-range keys (`ErrUnusableSeed`).

### Fix

Validate the scalar before constructing the key, using a constant-time `btcec.ModNScalar`: `SetByteSlice` reports `overflow` when the value is `>= N`, and `IsZero()` covers the zero key. Out-of-range keys now return the existing `ErrMalformedPrivateKey`.

### Test

Added `zero`, `order_N`, and `N+5` cases to `TestEncodeDecodeWIF`. Verified red-before / green-after by stashing the fix (3 FAIL without it, all PASS with it). The two existing valid-WIF round-trip cases are unaffected.

### Verification

- `go build ./...` — clean
- `go vet ./...` — clean
- `go test ./...` (whole `btcutil` module) — ok
- `gofmt -l` on touched files — clean
@

### DIFF
--- a/btcutil/wif.go
+++ b/btcutil/wif.go
@@ -127,6 +127,7 @@ func DecodeWIF(wif string) (*WIF, error) {
 	// the one actually encoded in the WIF (or the all-zero key) without
 	// reporting an error.
 	var keyScalar btcec.ModNScalar
+	defer keyScalar.Zero()
 	if overflow := keyScalar.SetByteSlice(privKeyBytes); overflow ||
 		keyScalar.IsZero() {
 
--- a/btcutil/wif.go
+++ b/btcutil/wif.go
@@ -118,6 +118,21 @@ func DecodeWIF(wif string) (*WIF, error) {
 
 	netID := decoded[0]
 	privKeyBytes := decoded[1 : 1+btcec.PrivKeyBytesLen]
+
+	// Ensure the private key is within the valid range for a secp256k1
+	// private key, that is [1, N-1].  Without this check, a WIF encoding a
+	// key of zero or one greater than or equal to the group order N is
+	// silently accepted: btcec.PrivKeyFromBytes reduces the scalar modulo
+	// N, so DecodeWIF would otherwise return a private key that differs from
+	// the one actually encoded in the WIF (or the all-zero key) without
+	// reporting an error.
+	var keyScalar btcec.ModNScalar
+	if overflow := keyScalar.SetByteSlice(privKeyBytes); overflow ||
+		keyScalar.IsZero() {
+
+		return nil, ErrMalformedPrivateKey
+	}
+
 	privKey, _ := btcec.PrivKeyFromBytes(privKeyBytes)
 	return &WIF{privKey, compress, netID}, nil
 }
--- a/btcutil/wif_test.go
+++ b/btcutil/wif_test.go
@@ -122,6 +122,30 @@ func TestEncodeDecodeWIF(t *testing.T) {
 			wif:  "5HueCGU8rMjxEXxiPuD5BDku4MkFqeZyd4dZ1jvhTVqvbTLvyTj",
 			err:  address.ErrChecksumMismatch,
 		},
+		{
+			// A WIF encoding a private key of zero, which is
+			// outside the valid range [1, N-1] for a secp256k1
+			// private key.
+			name: "decodeZeroPrivKeyWif",
+			wif:  "5HpHagT65TZzG1PH3CSu63k8DbpvD8s5ip4nEB3kEsreAbuatmU",
+			err:  ErrMalformedPrivateKey,
+		},
+		{
+			// A WIF encoding a private key equal to the group order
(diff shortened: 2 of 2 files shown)


TechLateef (approved):
LGTM. Thanks for the fix!

btcutil/wif.go
```diff
@@ -118,6 +118,21 @@ func DecodeWIF(wif string) (*WIF, error) {
...
 	privKeyBytes := decoded[1 : 1+btcec.PrivKeyBytesLen]
+
+	// Ensure the private key is within the valid range for a secp256k1
+	// private key, that is [1, N-1].  Without this check, a WIF encoding a
+	// key of zero or one greater than or equal to the group order N is
+	// silently accepted: btcec.PrivKeyFromBytes reduces the scalar modulo
+	// N, so DecodeWIF would otherwise return a private key that differs from
+	// the one actually encoded in the WIF (or the all-zero key) without
+	// reporting an error.
+	var keyScalar btcec.ModNScalar
```
allocz, line 129: This private key is being leaked. Can be fixed by zeroing the value before returning from the function.

allocz (changes requested):

btcutil/wif.go
allocz: Reading through all the code inside `wif.go`, there's a lot of places where memory containing private keys is not being zeroed after use, I just don't know if this is ok for the project security assumptions.

Lrifton92: Good catch, thanks. I've pushed f10224dd which adds `defer keyScalar.Zero()` so the scalar this PR introduces for the range check is wiped before `DecodeWIF` returns.

On the wider point: the other private-key material in `wif.go` (the `privKeyBytes` slice into `decoded`, and the `*btcec.PrivateKey` returned to the caller) isn't zeroed today, but that predates this PR — its scope is strictly rejecting out-of-range keys. A project-wide zeroing policy for `wif.go` touches several functions and depends on btcd's security assumptions (which, as you note, aren't settled here), so I'd rather not expand this PR to cover it. Happy to open a separate issue to track that if maintainers think it's worth doing.

allocz (approved):
Nice work. LGTM.

Lrifton92:
Two approvals landed here on June 24 (@TechLateef, @allocz) and it has been quiet since, so I went digging for why — I think this PR never had CI run on it at all.

The head commit shows zero check-runs, whereas merged PRs on this repo show the usual four (Build, Unit race, Unit rpctest, Unit coverage) — #2558 for example. My #2544 and #2561 are at zero too, so it looks systemic to outside-contributor PRs rather than specific to this change. Since `main.yml` triggers on `[push, pull_request]`, my guess is these are sitting behind the "approve and run workflows" gate and nobody has had a reason to click it. If that is what is going on, this is not a review problem at all.

Would a maintainer mind kicking the workflows off? Happy to do whatever is needed on my side if it turns out to be something else.

For reference, I re-checked locally against master (`6cfd7172`) today: the branch rebases cleanly (34 behind, no conflicts, still mergeable) and `btcutil` builds, vets and passes every suite. I have deliberately not force-pushed, so the existing approvals stay intact.

TechLateef:
Hi @Lrifton92 

Quick one on the CI front since we're outside contributors, only someone with write access can actually trigger those workflows. Hopefully, a maintainer swings by soon to kick it off!

Also note: If you're looking to get your PRs moving faster, a great hack in this repo is to jump in and review other people's PRs. It's a huge help to the maintainers and definitely increases the chances of getting your own stuff reviewed sooner.

Lrifton92:
Gentle follow-up: this now has two approvals (@TechLateef, @allocz) with all threads resolved and no conflicts. As @TechLateef pointed out, the CI workflows can't be triggered by an outside contributor — would a maintainer with write access be able to kick them off so this can move toward merge? Happy to rebase if anything has drifted in the meantime. Thanks!

Roasbeef (approved):

Issue #2543, opened 2026-06-11
[bug]: schnorr signature does not allow messages with length different than 32 bytes


allocz:
The current implementation fails the bip340 vector tests because the signature verification fails when the message length is not 32 bytes. Both bip340 test vectors and libsecp256k1 allows messages with variable length.

The fix for this bug is to remove the restrictions of the message length in both schnorr signing and verifying procedures.

I can provide the PR with the fix and tests upon ACK from maintainers.

allocz:
This also closes #2501, I could not find an issue linked to it.

Issue #2542, opened 2026-06-11
[bug]: schnorr signature does not match behavior specified by BIP340


allocz:
When signing without the auxRandom data, the signature differ from the signature generated by libsecp256k1, this happens because the current implementation uses RFC6979 when auxRandom is nil, while libsecp256k1 just uses an empty 32byte array to XOR with the privkey in this case.

The fix: instead of falling back to RFC6979, just use the empty 32 byte array when auxRandom is nil. Matching the behavior specified by BIP340 and libsecp256k1 implementation.

I can open the PR with the fix and new entries in the test vectors after acknowledment from maintainers.

yyforyongyu:
Yeah feel free to open a PR.

PR #2546, opened 2026-06-11
btcec/schnorr: document signing and verifying divergences against BIP340 and libsecp256k1 behavior

allocz:
## Change Description

Sign and Verify: document the requirement of 32 byte messages, as BIP340 requires and libsecp256k1 supports variable length messages.

Sign: since RFC6979 is used when `CustomNonce` is not passed, which is a deviation from BIP340 and secp256k1, a not was added explaining that passing an empty 32 byte array with CustomNonce is the same as the default behavior of libsecp256k1 when producing signatures.

Several leakages of sensitive material were fixed.

I would like to make the default behavior BIP340 compliant, matching libsecp256k1, but as this would introduce different behavior, even with the same API, the recommendation was to keep the current behavior but document it better.

Closes #2501 #2542 #2543

## Steps to Test

```
cd ./btcec
go test . -count=1
```
  
### Testing

### DIFF
--- a/btcec/schnorr/signature.go
+++ b/btcec/schnorr/signature.go
@@ -112,19 +112,20 @@ func (sig Signature) IsEqual(otherSig *Signature) bool {
 // error to support better testing while the exported method simply returns a
 // bool indicating success or failure.
 func schnorrVerify(sig *Signature, hash []byte, pubKeyBytes []byte) error {
-	// The algorithm for producing a BIP-340 signature is described in
-	// README.md and is reproduced here for reference:
+	// The reference algorithm for verifying a BIP-340 signature:
 	//
-	// 1. Fail if m is not 32 bytes
-	// 2. P = lift_x(int(pk)).
-	// 3. r = int(sig[0:32]); fail is r >= p.
-	// 4. s = int(sig[32:64]); fail if s >= n.
-	// 5. e = int(tagged_hash("BIP0340/challenge", bytes(r) || bytes(P) || M)) mod n.
-	// 6. R = s*G - e*P
-	// 7. Fail if is_infinite(R)
-	// 8. Fail if not hash_even_y(R)
-	// 9. Fail is x(R) != r.
-	// 10. Return success iff failure did not occur before reaching this point.
+	// 01. Fail if m is not 32 bytes
+	// 02. P = lift_x(int(pk)); fail if if that fails.
+	// 03. r = int(sig[0:32]); fail if r >= p.
+	// 04. s = int(sig[32:64]); fail if s >= n.
+	// 05. e = int(tagged_hash(
+	//     "BIP0340/challenge", bytes(r) || bytes(P) || m)) mod n.
+	// 06. R = s*G - e*P
+	// 07. Fail if is_infinite(R)
+	// 08. Fail if not hash_even_y(R)
+	// 09. Fail is x(R) != r.
+	// 10. Return success if failure did not occur before reaching this
+	//     point.
 
 	// Step 1.
 	//
@@ -163,7 +164,9 @@ func schnorrVerify(sig *Signature, hash []byte, pubKeyBytes []byte) error {
 
 	// Step 5.
 	//
-	// e = int(tagged_hash("BIP0340/challenge", bytes(r) || bytes(P) || M)) mod n.
+	// e = int(tagged_hash("BIP0340/challenge", bytes(r) || bytes(P) || M))
+	// mod n.
+
 	var rBytes [32]byte
 	sig.r.PutBytesUnchecked(rBytes[:])
 	pBytes := SerializePubKey(pubKey)
@@ -219,7 +222,7 @@ func schnorrVerify(sig *Signature, hash []byte, pubKeyBytes []byte) error {
 
 	// Step 10.
 	//
-	// Return success iff failure did not occur before reaching this point.
+	// Return success if failure did not occur before reaching this point.
 	return nil
 }
 
@@ -237,49 +240,53 @@ func zeroArray(a *[scalarSize]byte) {
 	}
 }
 
-// schnorrSign generates a BIP-340 signature over the secp256k1 curve for the
-// provided hash (which should be the result of hashing a larger message) using
-// the given nonce and private key.  The produced signature is deterministic
-// (same message, nonce, and key yield the same signature) and canonical.
+// schnorrSign generates a signature over the secp256k1 curve for the provided
+// hash (which should be the result of hashing a larger message) using the given
+// nonce and private key.  The produced signature is deterministic (same
+// message, nonce, and key yield the same signature) and canonical.
+//
+// NOTE: To generate signatures matching libsecp256k1 and BIP-340 when auxiliary
+// randomness is not provided, the CustomNonce option should be used with an
+// empty 32 byte array (32 zeroes).
 //
 // WARNING: The hash MUST be 32 bytes and both the nonce and private keys must
 // NOT be 0.  Since this is an internal use function, these preconditions MUST
 // be satisfied by the caller.
-func schnorrSign(privKey, nonce *btcec.ModNScalar, pubKey *btcec.PublicKey, hash []byte,
-	opts *signOptions) (*Signature, error) {
+func schnorrSign(privKey, nonce *btcec.ModNScalar, pubKey *btcec.PublicKey,
+	hash []byte, opts *signOptions) (*Signature, error) {
 
-	// The algorithm for producing a BIP-340 signature is described in
-	// README.md and is reproduced here for reference:
+	// The reference algorithm for producing a BIP-340 signature:
 	//
 	// G = curve generator
 	// n = curve order
 	// d = private key
 	// m = message
-	// a = input randomness
+	// a = auxiliary random data
 	// r, s = signature
 	//
-	// 1. d' = int(d)
-	// 2. Fail if m is not 32 bytes
-	// 3. Fail if d = 0 or d >= n
-	// 4. P = d'*G
-	// 5. Negate d if P.y is odd
-	// 6. t = bytes(d) xor tagged_hash("BIP0340/aux", t || bytes(P) || m)
-	// 7. rand = tagged_hash("BIP0340/nonce", a)
-	// 8. k' = int(rand) mod n
-	// 9. Fail if k' = 0
+	// 01. d' = int(d)
+	// 02. Fail if m is not 32 bytes
+	// 03. Fail if d = 0 or d >= n
+	// 04. P = d'*G
+	// 05. Negate d if P.y is odd
+	// 06. t = bytes(d) xor tagged_hash("BIP0340/aux", a)
+	// 07. rand = tagged_hash("BIP0340/nonce", t || bytes(P) || m)
+	// 08. k' = int(rand) mod n
+	// 09. Fail if k' = 0
 	// 10. R = 'k*G
 	// 11. Negate k if R.y id odd
-	// 12. e = tagged_hash("BIP0340/challenge", bytes(R) || bytes(P) || m) mod n
+	// 12. e = tagged_hash("BIP0340/challenge", bytes(R) || bytes(P) || m)
+	//     mod n
 	// 13. sig = bytes(R) || bytes((k + e*d)) mod n
 	// 14. If Verify(bytes(P), m, sig) fails, abort.
 	// 15. return sig.
 	//
 	// Note that the set of functional options passed in may modify the
 	// above algorithm. If CustomNonce is used, steps 6-8 follow BIP-340's
-	// nonce derivation using the provided auxiliary randomness. If CustomNonce
-	// is NOT used (the default), steps 6-8 are replaced with RFC6979 nonce
-	// generation for deterministic signing. If FastSign is passed, we skip
-	// step 14 (signature verification).
+	// nonce derivation using the provided auxiliary randomness. If
+	// CustomNonce is NOT used (the default), steps 6-8 are replaced with
+	// RFC6979 nonce generation for deterministic signing. If FastSign is
+	// passed, we skip step 14 (signature verification).
 
 	// NOTE: Steps 1-9 are performed by the caller.
 
@@ -350,8 +357,8 @@ type signOptions struct {
 	// where we attempt to verify the produced signature.
 	fastSign bool
 
-	// authNonce allows the user to pass in their own nonce information, which
-	// is useful for schemes like mu-sig.
+	// authNonce allows the user to pass in their own nonce information,
+	// which is useful for schemes like mu-sig.
 	authNonce *[32]byte
 }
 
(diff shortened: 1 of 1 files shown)


btcec/schnorr/signature.go
```diff
@@ -19,21 +19,6 @@ const (
...
-	// ensures the same nonce is not generated for the same message and key
-	// as for other signing algorithms such as ECDSA.
-	//
-	// It is equal to SHA-256([]byte("BIP-340")).
-	rfc6979ExtraDataV0 = [32]uint8{
-		0xa3, 0xeb, 0x4c, 0x18, 0x2f, 0xae, 0x7e, 0xf4,
-		0xe8, 0x10, 0xc6, 0xee, 0x13, 0xb0, 0xe9, 0x26,
-		0x68, 0x6d, 0x71, 0xe8, 0x7f, 0x39, 0x4f, 0x79,
-		0x9c, 0x00, 0xa5, 0x21, 0x03, 0xcb, 0x4e, 0x17,
-	}
```
GustavoStingelin, line 34: I guess the `BenchmarkSignRfc6979` should be renamed?

btcec/schnorr/signature.go
```diff
@@ -440,103 +400,73 @@ func Sign(privKey *btcec.PrivateKey, hash []byte,
...
-		// t = bytes(d) xor tagged_hash("BIP0340/aux", a)
-		privBytes := privKeyScalar.Bytes()
-		t := chainhash.TaggedHash(
-			chainhash.TagBIP0340Aux, (*opts.authNonce)[:],
-		)
-		for i := 0; i < len(t); i++ {
-			t[i] ^= privBytes[i]
-		}
+		copy(authNonce[:], opts.authNonce[:])
+	}
```
GustavoStingelin: While this converges toward the secp approach, does it also introduce any nonce reuse risk?

If a bug causes signatures to be produced with a zeroed nonce, the private key could be extracted. Given that, I think falling back to RFC6979 would provide an extra safety layer.

Rejecting a zeroed nonce outright is another option, but that seems to conflict with the goal of this PR, which is to match secp signing behavior.

At minimum, I think it would be helpful to document why secp considers this safe with a zeroed nonce, and why this code historically used RFC6979 instead.

GustavoStingelin: We should also zero `t`. With a zero nonce, the XOR operation leaves a recoverable private key in that variable.

btcec/schnorr/signature.go
```diff
@@ -440,103 +400,73 @@ func Sign(privKey *btcec.PrivateKey, hash []byte,
...
-	}
+	// Step 7.
+	//
+	// k'= int(rand) mod n
+	var kPrime btcec.ModNScalar
+	kPrime.SetBytes((*[32]byte)(rand))
 
-	var privKeyBytes [scalarSize]byte
-	privKeyScalar.PutBytes(&privKeyBytes)
-	defer zeroArray(&privKeyBytes)
```
GustavoStingelin, line 521: why no more privKeyBytes zeroing?

btcec/schnorr/signature.go
```diff
@@ -313,7 +281,8 @@ func schnorrSign(privKey, nonce *btcec.ModNScalar, pubKey *btcec.PublicKey, hash
 	if overflow := e.SetBytes((*[32]byte)(commitment)); overflow != 0 {
 		k.Zero()
 		str := "hash of (r || P || m) too big"
-		return nil, signatureError(ecdsa_schnorr.ErrSchnorrHashValue, str)
+		return nil, signatureError(ecdsa_schnorr.ErrSchnorrHashValue,
+			str)
 	}
```
GustavoStingelin, line 324: Not introduced here, but I think this should be impossible to overflow, since the value should be reduced mod `n`.

I saw that `SetBytes` already performs one reduction, so I believe the code should never reach this path. Still, it would be good to have evidence for that later.

GustavoStingelin (changes requested):

btcec/schnorr/signature.go
allocz: The data derived from `auxRand` (`t`) is hashed with the public key and the message to generate `rand`, so, there's no risk of nonce reusage by using a zeroed `auxRand`. Also, the name `CustomNonce` is misleading, the correct name should be `AuxRand`, for this reason I marked `CustomNonce` as deprecated and added `AuxRand` in replacement.

Now, zeroing both `t` and `privBytes`.

btcec/schnorr/signature.go
allocz: Note that before, when CustomNonce was passed, there was no private key zeroing, fixed now, thanks. https://github.com/btcsuite/btcd/blob/1966c384533d08539685d4ef4f1cd07bb11a549b/btcec/schnorr/signature.go#L471-L517

btcec/schnorr/signature.go
```diff
@@ -410,27 +378,19 @@ func Sign(privKey *btcec.PrivateKey, hash []byte,
...
+	// 5. t = bytes(d) xor tagged_hash("BIP0340/aux", t || bytes(P) || m)
+	// 6. rand = tagged_hash("BIP0340/nonce", a)
+	// 7. k' = int(rand) mod n
+	// 8. Fail if k' = 0
+	// 9. R = 'k*G
+	// 10. Negate k if R.y id odd
+	// 11. e = tagged_hash("BIP0340/challenge", bytes(R) || bytes(P) || mod) mod n
+	// 12. sig = bytes(R) || bytes((k + e*d)) mod n
+	// 13. If Verify(bytes(P), m, sig) fails, abort.
+	// 14. return sig.
```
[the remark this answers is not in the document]
allocz: Comment fixed, in fact steps 5 and 6 were wrong (ref: https://github.com/bitcoin/bips/blob/40cd7b4dd6dae2d9d38162b019ba588839331ba1/bip-0340.mediawiki?plain=1#L150-L151)

btcec/schnorr/signature.go
```diff
@@ -440,103 +400,73 @@ func Sign(privKey *btcec.PrivateKey, hash []byte,
...
-		if err != nil {
-			return nil, err
-		}
+	// Step 6.
+	//
+	// rand = tagged_hash("BIP0340/nonce", t || bytes(P) || m)
+	//
+	// We snip off the first byte of the serialized pubkey, as we
+	// only need the x coordinate and not the market byte.
+	rand := chainhash.TaggedHash(chainhash.TagBIP0340Nonce, t[:],
```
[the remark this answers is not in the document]
allocz: Typo fixed.

btcec/schnorr/signature.go
```diff
@@ -276,9 +243,10 @@ func schnorrSign(privKey, nonce *btcec.ModNScalar, pubKey *btcec.PublicKey, hash
...
 	// Note that the set of functional options passed in may modify the
 	// above algorithm. If CustomNonce is used, steps 6-8 follow BIP-340's
-	// nonce derivation using the provided auxiliary randomness. If CustomNonce
-	// is NOT used (the default), steps 6-8 are replaced with RFC6979 nonce
-	// generation for deterministic signing. If FastSign is passed, we skip
+	// nonce derivation using the provided auxiliary randomness. If
+	// CustomNonce is NOT used (the default), we match libsecp256k1 behavior
+	// by using the same algorithm but with auxRand being a zeroed 32 byte
+	// array, which is allowed by bip340. If FastSign is passed, we skip
 	// step 14 (signature verification).
```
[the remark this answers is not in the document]
allocz: The comment is describing what is happening, the same flow is used in both cases, the difference is that now instead of falling back to RFC6979 when auxRand is nil, we just use a 32 byte zeroed array, which is allowed by BIP-340 and implemented in libsecp256k1.

btcec/schnorr/signature.go
allocz: Renamed, as requested.

allocz:
`SignDefault`, which is the path that previously used RFC6979 is now ~5% faster, with a drop of 50% in the total allocations.
```
goos: linux
goarch: amd64
pkg: github.com/btcsuite/btcd/btcec/v2/schnorr
              │ /tmp/before │             /tmp/after             │
SigVerify-8     169.4µ ± 0%   160.8µ ± 0%  -5.09% (n=60)
Sign-8          70.07µ ± 0%   69.88µ ± 0%  -0.27% (p=0.004 n=60)
SignDefault-8   74.23µ ± 0%   69.90µ ± 0%  -5.84% (n=60)
geomean         95.86µ        92.25µ       -3.76%

              │ /tmp/before │              /tmp/after              │
              │    B/op     │    B/op     vs base                  │
SigVerify-8      240.0 ± 0%   240.0 ± 0%        ~ (p=1.000 n=60) ¹
Sign-8           496.0 ± 0%   496.0 ± 0%        ~ (p=1.000 n=60) ¹
SignDefault-8   1216.0 ± 0%   464.0 ± 0%  -61.84% (n=60)
geomean          525.1        380.8       -27.47%
¹ all samples are equal

              │ /tmp/before │              /tmp/after              │
              │  allocs/op  │ allocs/op   vs base                  │
SigVerify-8      5.000 ± 0%   5.000 ± 0%        ~ (p=1.000 n=60) ¹
Sign-8           12.00 ± 0%   12.00 ± 0%        ~ (p=1.000 n=60) ¹
SignDefault-8    22.00 ± 0%   11.00 ± 0%  -50.00% (n=60)
geomean          10.97        8.707       -20.63%
¹ all samples are equal
```

btcec/schnorr/signature.go
allocz: This implementation was written in the commit d6d38ad4a, which states: 

```
The signing implementation by default, deviates from BIP-340 as it opts
to use rfc6979 deterministic signatures by default, which means callers
don't need to always pass in their own `auxNonce` randomness. A set of
functional arguments allows callers to pass in their own value, which is
the way all the included test vectors function.
```

Refreshing it again, BIP-340 allows the `auxRand` to be all zeroes, which is the case for multiple of the test vectors. The generated `rand` is always safe to use, different messages will result in different nonces, even with `auxRand` being all zeroes, so there's no safety added by using RFC6979 here, only additional compute cost.

libsecp256k1 states that passing null (which in this case is not passing `AuxRand`) is the same as passing the zeroed 32 byte array. Adopting this enables faster signature generation and a behavior that matches both libsecp256k1 and BIP-340 without compromising security. 

Would be nice to see @Roasbeef take on this.

btcec/schnorr/signature.go
allocz: This additional check is not causing measurable differences in terms of performance, so I think we may keep it.

```
goos: linux
goarch: amd64
pkg: github.com/btcsuite/btcd/btcec/v2/schnorr
              │ /tmp/after  │            /tmp/after2             │
SigVerify-8     160.8µ ± 0%   160.5µ ± 0%  -0.17% (p=0.011 n=60)
Sign-8          69.88µ ± 0%   69.82µ ± 0%       ~ (p=0.524 n=60)
SignDefault-8   69.90µ ± 0%   69.97µ ± 0%  +0.10% (p=0.043 n=60)
geomean         92.25µ        92.21µ       -0.05%

              │ /tmp/after │             /tmp/after2             │
              │    B/op    │    B/op     vs base                 │
SigVerify-8     240.0 ± 0%   240.0 ± 0%       ~ (p=1.000 n=60) ¹
Sign-8          496.0 ± 0%   496.0 ± 0%       ~ (p=1.000 n=60) ¹
SignDefault-8   464.0 ± 0%   464.0 ± 0%       ~ (p=1.000 n=60) ¹
geomean         380.8        380.8       +0.00%
¹ all samples are equal

              │ /tmp/after │             /tmp/after2             │
              │ allocs/op  │ allocs/op   vs base                 │
SigVerify-8     5.000 ± 0%   5.000 ± 0%       ~ (p=1.000 n=60) ¹
Sign-8          12.00 ± 0%   12.00 ± 0%       ~ (p=1.000 n=60) ¹
SignDefault-8   11.00 ± 0%   11.00 ± 0%       ~ (p=1.000 n=60) ¹
geomean         8.707        8.707       +0.00%
¹ all samples are equal
```

GustavoStingelin: I was talking in terms of correctness instead of performance

GustavoStingelin (approved):
Cool, LGTM!

I noticed a few linting issues, but it looks like linting is not running in CI yet. I opened an issue to track that separately: https://github.com/btcsuite/btcd/issues/2549

Roasbeef:
> Sign, and Verify: accept variable length messages, as required by BIP340 and implemented by libsecp256k1.

So this was a purposeful decision w.r.t the API. Do you have a concrete use case of signing arbitrary length messages?

>  Sign: don't use RFC6979 when auxRand bytes are not passed by the caller, instead, match BIP340 spec and libsecp256k1 implementation by making not passing CustomNonce equivalent to passing a zeroed 32 byte array.

Again a divergence from the original design goals of the API. RFC6979 is a safer default, as it makes nonce re-use harder for callers.

allocz:
> So this was a purposeful decision w.r.t the API. Do you have a concrete use case of signing arbitrary length messages?

For now, the objective is to pass the BIP-340 test vectors.
 
> Again a divergence from the original design goals of the API. RFC6979 is a safer default, as it makes nonce re-use harder for callers.

The generated rand is always safe to use, different messages will result in different nonces, even with auxRand being all zeroes, so there's no safety added by using RFC6979 here, only additional compute cost.

TechLateef (commented):
Since the nonce derivation in BIP340 includes the private key and message, zeroed `auxRand` doesn't introduce nonce reuse risk. That said, @Roasbeef  concern about API safety for callers who don't understand the internals is worth considering  maybe a doc comment explicitly warning that passing no `AuxRand` is equivalent to zeroed bytes (not random) would help callers make an informed decision.

kcalvinalvin:
Thanks for the work! Some thoughts:

1: I think we should stick to the original design with the 32 byte check since we're changing the public API and the behavior. If passing the BIP340 tests are the goal, testing the unexported functions `schnorrSign` and `schnorrVerify` is better imo. Best not to sliently change what's happening under the hood. Later down the road if someone wants to sign arbitrary length messages, we can have a separate function for that instead of modifying `Sign`.

2: I went over RFC6979 vs 32 zeros and came to the conclusion that zeroing out doesn't really affect safety. For single sig it doesn't matter anyways since the r value is different for each message. For multi-sig, it doesn't matter if you're using 0s or RFC6979 since they're both deterministic. But again, since this will cause the API to break, I'm leaning towards keeping it the same.

allocz:
As the deviation from BIP340 does not introduce vulnerabilities, I've changed the commit to keep the current behavior (as @kcalvinalvin suggested), but also added more documentation covering that BIP340 signatures can be generated by using a zero 32 byte array with CustomNonce and also about arbitrary length messages not being supported.

Also fixed the docs and some private key and nonce leaks.

allocz:
I think this one is good for review. CC @kcalvinalvin.

PR #2547, opened 2026-06-12
scripts: add script to upgrade to v2 modules automatically

guggero:
Will allow downstream consumer projects to more easily switch to the new `v2` modules by using an upgrade script.


guggero:
I think a lot of the consumers have already used the script and it was never really meant to be added to the repo anyway (the PR was just a place for the script to live until it's not needed anymore). So I'll close this but the script can still be used of course (it's not perfect but does about 90% of the work).

PR #2548, opened 2026-06-12
netsync: require block-serving services on regtest/simnet sync peers

guggero:
## Change Description

Fixes a sync livelock on regtest/simnet introduced by #2487
(26124d275, `netsync: allow sync with non-localhost peers on
regtest/simnet`).

That commit's goal was to drop the localhost-only *address* restriction
so nodes on Docker bridge networks (e.g. `172.18.0.x`) can be used as
sync peers. However, the rewritten check returns `true` unconditionally
for regtest and simnet, so the *service-flag* requirement was dropped as
well — any connected peer became a valid sync candidate, including light
clients.

### The livelock

A light client such as neutrino advertises a recent best height in its
version message, but signals neither `SFNodeNetwork` nor
`SFNodeNetworkLimited` and cannot serve `getheaders`/`getblocks`. When a
simnet/regtest node has a light client connected:

1. `startSync` elects the light client as the sync peer (it advertises a
   height above our tip), and the header/block download stalls because
   the peer never responds.
2. After the stall timeout (~90s) the sync manager disconnects it.
3. The light client's persistent connection logic reconnects, and the
   light client is elected again — repeat indefinitely.

The damage isn't limited to the stalled download: while a sync peer is
set and its advertised height is above our tip, `sm.current()` returns
false, so `handleInvMsg` also discards block announcements from the
*other* (full-node) peers. The node can therefore neither pull from the
light client nor accept pushes from honest peers, and never syncs.

Log signature of the livelock (here `neutrino:0.17.1` is the light
client being elected over and over):

```
16:26:06 [INF] SYNC: Downloading headers for blocks 1 to 1225 from peer 127.0.0.1:36004
16:27:36 [INF] SYNC: Lost peer 127.0.0.1:36004 (inbound)
16:27:41 [INF] SYNC: Downloading headers for blocks 1 to 1225 from peer 127.0.0.1:60982
16:29:11 [INF] SYNC: Lost peer 127.0.0.1:60982 (inbound)
16:29:16 [INF] SYNC: Downloading headers for blocks 1 to 1225 from peer 127.0.0.1:56826
...
```

This surfaced in lightninglabs/neutrino's sync tests
(`TestNeutrinoSyncWithoutHeadersImport` and friends), which spin up
several simnet btcd nodes via `rpctest` while a neutrino `ChainService`
is connected to all of them: `rpctest.JoinNodes` never converges and the
test suite hits its 10 minute timeout.

### The fix

Keep accepting any peer *address* on regtest/simnet (the intent of
#2487), but require the peer to signal `SFNodeNetwork` or
`SFNodeNetworkLimited`, the same block-serving capability expected on
every other network. The further mainnet-only refinements (segwit
upgrade check, pruned-peer depth window) remain skipped on test
networks, matching the pre-#2487 behavior.

### DIFF
--- a/netsync/manager.go
+++ b/netsync/manager.go
@@ -360,38 +360,13 @@ func (sm *SyncManager) startSync() {
 // isSyncCandidate returns whether or not the peer is a candidate to consider
 // syncing from.
 func (sm *SyncManager) isSyncCandidate(peer *peerpkg.Peer) bool {
-	// Typically a peer is not a candidate for sync if it's not a full node,
-	// however regression test is special in that the regression tool is
-	// not a full node and still needs to be considered a sync candidate.
-	switch sm.chainParams.Name {
-	case chaincfg.RegressionNetParams.Name, chaincfg.SimNetParams.Name:
-		// In regtest/simnet mode, any peer is a valid sync candidate
-		// regardless of its address or service flags. This allows
-		// syncing from peers on non-localhost networks such as Docker
-		// bridge networks.
-		return true
-	}
-
-	// If the segwit soft-fork package has activated, then the peer must
-	// also be upgraded.
-	segwitActive, err := sm.chain.IsDeploymentActive(
-		chaincfg.DeploymentSegwit,
-	)
-	if err != nil {
-		log.Errorf("Unable to query for segwit soft-fork state: %v",
-			err)
-	}
-
-	if segwitActive && !peer.IsWitnessEnabled() {
-		return false
-	}
-
 	var (
 		nodeServices = peer.Services()
 		fullNode     = nodeServices.HasFlag(wire.SFNodeNetwork)
 		prunedNode   = nodeServices.HasFlag(wire.SFNodeNetworkLimited)
 	)
 
+	// We check the node's ability to serve blocks first.
 	switch {
 	case fullNode:
 		// Node is a sync candidate if it has all the blocks.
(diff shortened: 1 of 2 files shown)


Roasbeef (commented):
Thx for the fix! 

Can you amend the commit to drop the Claude attribution (there's also a config setting to disable that)? Looks good after that

guggero:
Ah yes, thanks. I disabled the setting and updated the commit.

netsync/manager.go
```diff
@@ -365,11 +365,16 @@ func (sm *SyncManager) isSyncCandidate(peer *peerpkg.Peer) bool {
...
-		return true
+		// In regtest/simnet mode, any peer address is acceptable (this
+		// allows syncing from peers on non-localhost networks such as
+		// Docker bridge networks), but the peer must still be able to
+		// serve us blocks. Light clients (e.g. neutrino) advertise
+		// neither flag; electing one as the sync peer would stall the
+		// sync until the stall handler disconnects it, only for the
+		// next light client to be elected and stall again.
+		services := peer.Services()
+		return services.HasFlag(wire.SFNodeNetwork) ||
```
yyforyongyu: Optional follow-up: this regtest/simnet branch now requires a block-serving service flag, but `SFNodeNetworkLimited` is accepted without the depth-window check used for other networks below. A limited peer that is more than `wire.NodeNetworkLimitedBlockThreshold` blocks ahead may still be unable to serve our next block and could stall sync. Consider sharing the limited-peer height check here while still skipping address/segwit restrictions.

netsync/manager_test.go
```diff
@@ -1231,36 +1230,37 @@ func TestIsSyncCandidateRegtest(t *testing.T) {
...
+			name:  "just node network",
+			flags: wire.SFNodeNetwork,
+			want:  true,
 		},
 		{
-			name: "docker bridge ip",
-			addr: "172.18.0.2:18444",
-			want: true,
+			name:  "node network and limited node network",
+			flags: wire.SFNodeNetwork | wire.SFNodeNetworkLimited,
```
yyforyongyu, line 1256: Optional follow-up: this case does not cover `SFNodeNetworkLimited` by itself because `SFNodeNetwork` is also set. Since the implementation accepts either flag, it would be useful to add a limited-only case and a simnet case to cover both networks mentioned by the fix.

yyforyongyu (approved):
LGTM❤️

netsync/manager.go
guggero: Hmm, yeah, you're right. I moved the full vs. pruned node check to the beginning. Thanks!

yyforyongyu (approved):
LGTM🥇

Roasbeef (approved):
LGTM 🐛

PR #2550, opened 2026-06-12
rpc: reject trailing bytes in decoderawtransaction

### LINKED ISSUE #2195: Check if input was entirely consumed in `decoderawtransaction`/`btcDecode`
Using the `decoderawtransaction` command (`btcDecode` function?) we noticed that it does not check if the entire input was successfully consumed when decoding which is important to keep it consistent. 

To reproduce:
```
decoderawtransaction "0a9a9e0101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010100000000010001010101a09a"
```

Obs.: Running same command on Bitcoin Core fails.

Lrifton92:
### Summary

`handleDecodeRawTransaction` deserialized the supplied hex into a `wire.MsgTx` but never checked that the entire input was consumed. `MsgTx.Deserialize` reads exactly the bytes that make up a transaction and stops, so any extra bytes appended after a fully-formed transaction were silently discarded and the RPC still returned a successful decode result.

This diverges from bitcoind, which rejects such input with a deserialization error.

### Root cause

The handler passed the raw bytes to `Deserialize` via a `*bytes.Reader` and inspected only the returned error, ignoring whether unread bytes remained in the reader.

### Fix

After a successful `Deserialize`, check `txReader.Len() != 0`. If any bytes remain, the input is not a valid serialized transaction and an `ErrRPCDeserialization` error is returned, matching Core's behavior.

### Testing

Added `TestHandleDecodeRawTransactionTrailingBytes`, which appends a spurious byte to an otherwise valid transaction and asserts the handler now returns `ErrRPCDeserialization`. Before the fix the trailing byte was silently dropped (the handler continued past deserialization); after the fix it is rejected.

Verified locally: `go build ./...`, `go test .`, `go vet .`, and `gofmt` are all clean.

### Repro from the issue

```
decoderawtransaction "0a9a9e01010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010100000000010001010101a09a"
```

This decoded successfully on btcd while Bitcoin Core errors out; with this change btcd now rejects the trailing bytes.

Closes #2195

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -792,11 +792,24 @@ func handleDecodeRawTransaction(s *rpcServer, cmd interface{}, closeChan <-chan 
 		return nil, rpcDecodeHexError(hexStr)
 	}
 	var mtx wire.MsgTx
-	err = mtx.Deserialize(bytes.NewReader(serializedTx))
+	txReader := bytes.NewReader(serializedTx)
+	err = mtx.Deserialize(txReader)
 	if err != nil {
 		return nil, &btcjson.RPCError{
 			Code:    btcjson.ErrRPCDeserialization,
 			Message: "TX decode failed: " + err.Error(),
+		}
+	}
+
+	// Ensure the entire input was consumed during deserialization. Any
+	// trailing bytes mean the input is not a valid serialized transaction,
+	// so reject it to match bitcoind's behavior instead of silently
+	// discarding the extra data.
+	if txReader.Len() != 0 {
+		return nil, &btcjson.RPCError{
+			Code: btcjson.ErrRPCDeserialization,
+			Message: "TX decode failed: transaction has " +
+				"unexpected trailing bytes",
 		}
 	}
 
--- a/rpcserver_test.go
+++ b/rpcserver_test.go
@@ -67,6 +67,36 @@ func TestHandleTestMempoolAcceptFailDecode(t *testing.T) {
 	}
 }
 
+// TestHandleDecodeRawTransactionTrailingBytes checks that
+// decoderawtransaction rejects an input that has trailing bytes left over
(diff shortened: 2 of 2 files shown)


Lrifton92:
Friendly ping — this is ready for review whenever a maintainer has bandwidth. CI is green and the change is small (reject trailing bytes in decoderawtransaction). Thanks!

starius:

kcalvinalvin:
> Replacing this PR with #2558

This PR is much smaller in scope and a lot easier to review. I'd say land this and the others in the #2558 instead of replacing it?

LGTM at 6966e6964f497907539fc21b90d4b24560ee579f

starius:
@kcalvinalvin Thanks for review! This PR does fix the bug in `decoderawtransaction`, but we need to fix all such cases eventually, that is why I opened https://github.com/btcsuite/btcd/pull/2558 .

Note that https://github.com/btcsuite/btcd/pull/2558 deduplicates the trailing tail checks by reusing one fixed functions from others, not to have this logic spread across the repo more than needed. In particular `decoderawtransaction` is fixed in this commit: https://github.com/btcsuite/btcd/pull/2558/changes/0d7259ee159abec827436e1083230439f08f1ab5 . It builds upon `btcutil.NewTxFromBytes` which is fixed in another commit https://github.com/btcsuite/btcd/pull/2558/changes/41d537d85e7cdde325fbac61f34701e0d8ae3dc1 .

Could you go through https://github.com/btcsuite/btcd/pull/2558 please? Most commits are small. The number of commits is huge, because each fix has a separate regression test commit, to make it more convenient to validate the test by reverting the main commit and running the test.

kcalvinalvin (approved):

Roasbeef:

PR #2551, opened 2026-06-16
rpcclient: add typed SubmitPackage method

ellemouton:
## Summary

Adds a typed `SubmitPackage` to `rpcclient`, wrapping the `submitpackage` RPC the same way `TestMempoolAccept` wraps `testmempoolaccept`.

The `btcjson` `submitpackage` command + result types already exist; this adds the client-side method so callers don't have to drop down to `RawRequest`:

- `FutureSubmitPackageResult` + `Receive()` — decodes the response into `btcjson.SubmitPackageResult` (which already maps the raw fields to higher-level types via its `UnmarshalJSON`).
- `SubmitPackageAsync(txns, maxFeeRate, maxBurnAmount)` — serializes the topologically-sorted package (parents first, child last) to hex and issues the command.
- `SubmitPackage(...)` — the blocking wrapper.

This keeps the clean layering where `rpcclient` owns the multi-backend RPCs. `submitpackage` is a Bitcoin Core RPC (v24+); btcd has no server handler for it, so this is a client-only addition.

## Test plan

`go build ./...`, `go vet ./rpcclient/...`, `gofmt` clean. The method mirrors the existing `TestMempoolAccept` plumbing; it's exercised end-to-end by a downstream `btcwallet` change that relays v3/TRUC CPFP packages through it.

### DIFF
--- a/rpcclient/rawtransactions.go
+++ b/rpcclient/rawtransactions.go
@@ -1015,6 +1015,110 @@ func (c *Client) TestMempoolAccept(txns []*wire.MsgTx,
 	return c.TestMempoolAcceptAsync(txns, maxFeeRate).Receive()
 }
 
+// FutureSubmitPackageResult is a future promise to deliver the result of a
+// SubmitPackage RPC invocation (or an applicable error).
+type FutureSubmitPackageResult chan *Response
+
+// Receive waits for the Response promised by the future and returns the result
+// of submitting the package.
+func (r FutureSubmitPackageResult) Receive() (*btcjson.SubmitPackageResult,
+	error) {
+
+	response, err := ReceiveFuture(r)
+	if err != nil {
+		return nil, err
+	}
+
+	// btcjson.SubmitPackageResult implements a custom UnmarshalJSON that
+	// maps the raw submitpackage response into higher-level types.
+	var result btcjson.SubmitPackageResult
+	if err := json.Unmarshal(response, &result); err != nil {
+		return nil, err
+	}
+
+	return &result, nil
+}
+
+// maxPackageTxns is the maximum number of transactions a single submitpackage
+// call may contain. It mirrors bitcoind's MAX_PACKAGE_COUNT (see
+// src/policy/packages.h), which caps a package at 25 transactions.
+const maxPackageTxns = 25
+
+// SubmitPackageAsync returns an instance of a type that can be used to get the
+// result of the RPC at some future time by invoking the Receive function on
+// the returned instance.
+//
+// See SubmitPackage for the blocking version and more details.
+func (c *Client) SubmitPackageAsync(txns []*wire.MsgTx,
+	maxFeeRate, maxBurnAmount *float64) FutureSubmitPackageResult {
+
+	// Gate on the backend version so an unsupported backend (btcd, or
+	// Bitcoin Core before v24) returns a typed ErrBackendVersion rather
+	// than a raw method-not-found error the caller would have to
+	// string-match. This mirrors TestMempoolAccept/GetTxSpendingPrevOut.
+	version, err := c.BackendVersion()
+	if err != nil {
+		return newFutureError(err)
+	}
+
+	if !version.SupportSubmitPackage() {
+		err := fmt.Errorf("%w: %v", ErrBackendVersion, version)
+
+		return newFutureError(err)
+	}
+
+	// A package must contain at least one transaction (the child) and at
+	// most maxPackageTxns.
+	if len(txns) == 0 {
+		err := fmt.Errorf("%w: no transactions provided",
+			ErrInvalidParam)
+
+		return newFutureError(err)
+	}
+
+	if len(txns) > maxPackageTxns {
+		err := fmt.Errorf("%w: too many transactions provided",
+			ErrInvalidParam)
+
+		return newFutureError(err)
+	}
+
+	// Serialize each transaction to a hex string, preserving the
+	// topological order (parents first, child last) the caller provides.
+	rawTxns := make([]string, 0, len(txns))
+	for _, tx := range txns {
+		buf := bytes.NewBuffer(make([]byte, 0, tx.SerializeSize()))
+		if err := tx.Serialize(buf); err != nil {
+			err = fmt.Errorf("%w: %v", ErrInvalidParam, err)
+
+			return newFutureError(err)
+		}
+
+		rawTxns = append(rawTxns, hex.EncodeToString(buf.Bytes()))
+	}
+
+	cmd := btcjson.NewJsonSubmitPackageCmd(
+		rawTxns, maxFeeRate, maxBurnAmount,
+	)
+
+	return c.SendCmd(cmd)
+}
+
+// SubmitPackage submits a package of related, topologically-sorted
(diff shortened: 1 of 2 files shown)


ellemouton:
cc @qustavo & @yyforyongyu

qustavo:
I believe you wanted to mention @GustavoStingelin

rpcclient/rawtransactions.go
```diff
@@ -1022,6 +1015,93 @@ func (c *Client) TestMempoolAccept(txns []*wire.MsgTx,
...
+	// A package must contain at least one transaction (the child) and at
+	// most 25.
+	if len(txns) == 0 {
+		err := fmt.Errorf("%w: no transactions provided",
+			ErrInvalidParam)
+
+		return newFutureError(err)
+	}
+
+	if len(txns) > 25 {
```
qustavo: where does this number (25) come from?
would it be better to inform that the right number of txns if between 1 and 25?

rpcclient/rawtransactions.go
```diff
@@ -1022,6 +1015,93 @@ func (c *Client) TestMempoolAccept(txns []*wire.MsgTx,
...
+// SubmitPackageAsync returns an instance of a type that can be used to get the
+// result of the RPC at some future time by invoking the Receive function on
+// the returned instance.
+//
+// See SubmitPackage for the blocking version and more details.
+func (c *Client) SubmitPackageAsync(txns []*wire.MsgTx,
+	maxFeeRate, maxBurnAmount *float64) FutureSubmitPackageResult {
+
+	// A package must contain at least one transaction (the child) and at
+	// most 25.
```
GustavoStingelin: Given the TRUC context, what is the rationale for keeping this 25-child constraint?

rpcclient/rawtransactions.go
```diff
@@ -1022,6 +1015,93 @@ func (c *Client) TestMempoolAccept(txns []*wire.MsgTx,
...
+// transactions (unconfirmed parents first, child last) to the backend's
+// mempool for atomic validation and acceptance via the submitpackage RPC. It
+// lets a zero-fee v3/TRUC parent be accepted via its fee-paying CPFP child,
+// which a standalone broadcast would reject.
+//
+// maxFeeRate (BTC/kvB) and maxBurnAmount (BTC) are optional limits; a nil
+// value uses the backend's RPC default.
+//
+// NOTE: submitpackage is a Bitcoin Core RPC (v24+); btcd has no equivalent
+// handler.
```
GustavoStingelin: I guess this one was released in v26?
https://github.com/bitcoin/bitcoin/blob/a30ef6b91f73a6669e5aeedf460e0c9be2745a49/doc/release-notes/release-notes-26.0.md?plain=1#L114

rpcclient/rawtransactions.go
Roasbeef: IIUC this is to align with bitcoind's package submission limit: https://github.com/bitcoin/bitcoin/blob/a30ef6b91f73a6669e5aeedf460e0c9be2745a49/src/policy/packages.h#L18-L19

rpcclient/rawtransactions.go
Roasbeef: I think can ditch this portion of the comment, as it'll become obsolete with the stacked PRs in the tracker.

rpcclient/rawtransactions.go
Roasbeef: We can define it as a top level constant here, see my comment above.

ellemouton: yeah will add the constant

rpcclient/rawtransactions.go
```diff
@@ -1022,6 +1015,95 @@ func (c *Client) TestMempoolAccept(txns []*wire.MsgTx,
...
+	// A package must contain at least one transaction (the child) and at
+	// most maxPackageTxns.
+	if len(txns) == 0 {
+		err := fmt.Errorf("%w: no transactions provided",
+			ErrInvalidParam)
+
+		return newFutureError(err)
+	}
+
+	if len(txns) > maxPackageTxns {
```
qustavo, line 1079: If we skip this check, I assume the bitcoind will return an error and the request will fail. If that's the case, wouldn't be best just to let the server to return such an error? this way there's no need to keep the rpcclient in sync with bitcoind's constants?

rpcclient/rawtransactions.go
yyforyongyu: this is the default mempool policy from bitcoind, which is also used in `testmempoolaccept`

rpcclient/rawtransactions.go
```diff
@@ -1015,6 +1015,95 @@ func (c *Client) TestMempoolAccept(txns []*wire.MsgTx,
...
+// call may contain. It mirrors bitcoind's MAX_PACKAGE_COUNT (see
+// src/policy/packages.h), which caps a package at 25 transactions.
+const maxPackageTxns = 25
+
+// SubmitPackageAsync returns an instance of a type that can be used to get the
+// result of the RPC at some future time by invoking the Receive function on
+// the returned instance.
+//
+// See SubmitPackage for the blocking version and more details.
+func (c *Client) SubmitPackageAsync(txns []*wire.MsgTx,
```
yyforyongyu, line 1052: Should this mirror TestMempoolAccept/GetTxSpendingPrevOut and gate on BackendVersion? As written, this sends submitpackage to every backend/version, so Core <26 or btcd returns a raw method-not-found error instead of rpcclient.ErrBackendVersion or another typed unsupported result. A SupportSubmitPackage predicate would keep callers from string-matching RPC errors.

rpcclient/rawtransactions.go
```diff
@@ -1015,6 +1015,95 @@ func (c *Client) TestMempoolAccept(txns []*wire.MsgTx,
...
+		if err := tx.Serialize(buf); err != nil {
+			err = fmt.Errorf("%w: %v", ErrInvalidParam, err)
+
+			return newFutureError(err)
+		}
+
+		rawTxns = append(rawTxns, hex.EncodeToString(buf.Bytes()))
+	}
+
+	cmd := btcjson.NewJsonSubmitPackageCmd(
```
yyforyongyu, line 1100: This forwards maxFeeRate/maxBurnAmount whenever they are non-nil, but Core v26/v27 submitpackage only accepts the package argument; maxfeerate/maxburnamount were added in v28. For example, passing &0 to disable the fee limit will fail on v26/v27 with a parameter error. The wrapper likely needs a version split: v26/v27 package-only, v28+ optional args, or an explicit ErrBackendVersion when optional args are requested on older Core.

yyforyongyu (approved):
Post some comments as non-blocking, as I think we can merge this PR first to unblock upstream PRs. I will create PRs to address them later.

GustavoStingelin (approved):
LGTM

rpcclient/rawtransactions.go
ellemouton: added this :+1:

ellemouton:
Thanks for the review yall!
@yyforyongyu  - addressed the `SupportSubmitPackage` but will leave the package arguments comment for follow up 👍

Roasbeef (approved):
LGTM 🦫

PR #2552, opened 2026-06-18
version: bump to v0.26.0-beta for final release

Roasbeef:
In this PR, we drop the `rc1` pre-release suffix, bumping the version from
`v0.26.0-beta.rc1` to `v0.26.0-beta` for the final release.


saubyk (approved):
Ack

PR #2554, opened 2026-06-18
btcutil: vendor anet instead of depending on a personal fork

Roasbeef:
In this PR, we vendor the Android networking shim that backs `InterfaceAddrs` on android into the `btcutil` module, so we no longer pull it from `github.com/kcalvinalvin/anet`. The shim exists to work around [golang/go#40569](https://github.com/golang/go/issues/40569), where `net.InterfaceAddrs` fails with `route ip+net: netlinkrib: permission denied` on Android 11 and later because of the tightened restrictions on `NETLINK` sockets. `btcutil` only ever calls `anet.InterfaceAddrs`, wired in from `net_android.go` under the `android` build tag.

The motivation here is dependency hygiene. The old import pointed at a fork hosted under an individual contributor's account, which isn't a home we want a release dependency to live in. By copying the source into `btcutil/internal/anet` we drop the external module entirely: the package is now unimportable from outside `btcutil`, and `btcutil/go.mod` no longer references the fork.

## What's in the vendored copy

The source originates from [`github.com/wlynxg/anet`](https://github.com/wlynxg/anet), and its BSD 3-Clause `LICENSE` travels with it for attribution. The `mobile/` demo subpackage from upstream is dropped since we don't need it.

One nice property of this particular fork: unlike upstream, it doesn't reach into the standard library's `net` package via `//go:linkname`. It carries its own `ipv6ZoneCache` instead, so there's no need for the `-ldflags "-checklinkname=0"` linker flag that upstream's README calls for on Go 1.23+. I swapped the upstream README for a short provenance note so that misleading build instruction doesn't linger.

## Scope

The real change lives in the `btcutil` module, which is the only consumer of `anet`. The root and `psbt` modules still carry `anet` as a transitive `// indirect` requirement because they depend on the published `btcutil/v2 v2.0.0` (there's no `replace` or `go.work` wiring them to the local tree). Those entries will fall away naturally on the next `btcutil` tag that ships this commit.

Verified that `internal/anet` and `btcutil` both build for `GOOS=android GOARCH=arm64`, the normal host build and `go vet` pass, and no `kcalvinalvin` references remain in the module.

### DIFF
--- a/btcutil/go.mod
+++ b/btcutil/go.mod
@@ -12,7 +12,6 @@ require (
 	github.com/btcsuite/btcd/wire/v2 v2.0.0
 	github.com/davecgh/go-spew v1.1.1
 	github.com/decred/dcrd/dcrec/secp256k1/v4 v4.4.0
-	github.com/kcalvinalvin/anet v0.0.0-20251112173137-d8ddc1f6dbee
 	github.com/kkdai/bstream v1.0.0
 )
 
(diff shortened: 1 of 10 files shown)


allocz (approved):
Compared the diff against https://github.com/kcalvinalvin/anet, LGTM.

kcalvinalvin (approved):
LGTM ce4e0f53af7e1b146ea7d202bfe109eadaffb707

Verified that the contents of anet/ are the same.

PR #2556, opened 2026-06-20
rpc: reject trailing bytes in sendrawtransaction

Lrifton92:
### Problem

`handleSendRawTransaction` deserializes the supplied hex into a `wire.MsgTx` but never verifies that the entire input was consumed:

```go
var msgTx wire.MsgTx
err = msgTx.Deserialize(bytes.NewReader(serializedTx))
```

`MsgTx.Deserialize` reads exactly the bytes that make up a transaction and stops, so any extra bytes appended after a fully-formed transaction are silently discarded — and the transaction is still accepted and relayed.

This diverges from `bitcoind`, whose `sendrawtransaction` rejects such input with a deserialization error (`TX decode failed`) instead of broadcasting a transaction parsed from a malformed payload.

### Fix

Read the serialized transaction through a `*bytes.Reader` and, after a successful `Deserialize`, check that no unread bytes remain. If any are left over, return `ErrRPCDeserialization`, matching Core's behavior.

This mirrors the same trailing-bytes hardening applied to `decoderawtransaction` in #2550 and to the PSBT global unsigned tx in #2541 — completing the set across the RPC tx-decode paths.

### Test

`TestHandleSendRawTransactionTrailingBytes` appends a spurious byte to an otherwise valid transaction and asserts the handler now returns `ErrRPCDeserialization`. Verified red before the fix (the handler proceeded past deserialization) and green after.

`go build ./...`, `go vet .`, `gofmt -l` and the RPC handler tests all pass.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -3442,11 +3442,24 @@ func handleSendRawTransaction(s *rpcServer, cmd interface{}, closeChan <-chan st
 		return nil, rpcDecodeHexError(hexStr)
 	}
 	var msgTx wire.MsgTx
-	err = msgTx.Deserialize(bytes.NewReader(serializedTx))
+	txReader := bytes.NewReader(serializedTx)
+	err = msgTx.Deserialize(txReader)
 	if err != nil {
 		return nil, &btcjson.RPCError{
 			Code:    btcjson.ErrRPCDeserialization,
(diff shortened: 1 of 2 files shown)


allocz (approved):
LGTM.

starius:

Roasbeef:

PR #2558, opened 2026-06-26
multi: reject trailing characters when parsing inputs

### LINKED ISSUE #2195: Check if input was entirely consumed in `decoderawtransaction`/`btcDecode`

### LINKED ISSUE #2424: psbt: Missing validation that deserialization consumes exact number of bytes specified in length prefix

starius:
## Change Description

Some functions accept `[]byte` inputs with trailing characters. Many such cases were fixed and covered with regression tests.

This PR replaces there PRs:
  * https://github.com/btcsuite/btcd/pull/2541
  * https://github.com/btcsuite/btcd/pull/2550
  * https://github.com/btcsuite/btcd/pull/2556

Fixes https://github.com/btcsuite/btcd/issues/2424
Fixes https://github.com/btcsuite/btcd/issues/2195

**Note 1**. In commit "multi: use local submodules in root" I replaced `btcutil/v2` and `wire/v2` with local versions to make the fixed changes available in `rpc: *` commits.

**Note 2**. PSBT parsing accepted `\n` and `\r\n` in the end and in the middle, if base64 parsing is enabled. This is not aligned with [BIP 174](https://bips.dev/174/)/[RFC4648](https://datatracker.ietf.org/doc/html/rfc4648#section-4) and [Bitcoin Core](https://github.com/bitcoin/bitcoin/blob/011ad6ea3c422100210392686af1bf4022172b37/src/util/strencodings.cpp#L108-L140), which reject such inputs. I added this validation in a separate commit. Now base64 PSBT string must have only base64 commot charset + `=` characters in the end, but no whitespace, new line or tabs.

## Steps to Test

Run the tests. Each fix has a regression test which can be tried with the fix commit reverted. A regression test compiles, but fails with a symptomatic error.

### Testing


Lrifton92 (commented):
Went through the full series (27 commits) @starius — happy to see all the trailing-byte cases consolidated in one place, and thanks for extending it well beyond the original three (`wire` v2 payloads, `musig2` partial sigs, `btcutil` byte constructors, block proposals, DB block loading).

I built the root module + every touched submodule and ran their suites locally (go1.26.4):

- `psbt`, `wire`, `btcutil` (+ `bloom`), `btcec/schnorr/musig2`, `blockchain`, and root `.` (rpcserver) — all green.

I also validated the "revert the fix, keep the test → red" property on a few representative pairs, since that's the whole point of splitting each fix from its regression test:

- **rpc**: restoring `rpcserver.go` to pre-`59db835` → `TestHandleSendRawTransactionRejectsTrailingBytes` and `TestHandleDecodeRawTransactionRejectsTrailingBytes` fail. ✅
- **btcutil**: restoring `tx.go`/`block.go` to pre-`41d537d` → `TestNewTxFromBytesRejectsTrailingData` and `TestNewBlockFromBytesRejectsTrailingData` fail ("expected error for … with trailing data"). ✅
- **psbt**: restoring `psbt.go` to pre-`70e8ceb` → `TestRejectsNonCanonicalBase64Packet` fails (raw `illegal base64 data` instead of `ErrInvalidPsbtFormat`). ✅

One heads-up so it doesn't get blamed on this PR: `netsync` `TestSyncStateMachine` (`handleHeadersMsg should update lastProgressTime`) fails for me locally — but it fails identically on a clean `master` (280d4d2d) and this PR doesn't touch `netsync`, so it looks like a pre-existing timing flake, unrelated here.

LGTM. This correctly supersedes #2541 / #2550 / #2556 — glad to close those in favor of this consolidated version whenever you'd like.

psbt/utils.go
```diff
@@ -278,17 +277,69 @@ func getKey(r io.Reader) (int, []byte, error) {
 	return int(keyType), keyData, nil
 }
 
-// readTxOut is a limited version of wire.ReadTxOut, because the latter is not
-// exported.
+// assertFullyConsumed returns ErrInvalidPsbtFormat if r still has bytes
+// available after parsing.
+func assertFullyConsumed(r io.Reader) error {
```
Roasbeef, line 282: 👍 

Was originally thinking we could just do it unilaterally at the `wire` level, but this is better as we can't assume too much about the context the reader/writer impls are used within.

go.mod
```diff
@@ -39,6 +39,11 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
+replace (
+	github.com/btcsuite/btcd/btcutil/v2 => ./btcutil
+	github.com/btcsuite/btcd/wire/v2 => ./wire
```
Roasbeef, line 44: Reminder to push a tag and then make a PR to update after.

Roasbeef (commented):
LGTM 🐧

psbt/utils.go
```diff
@@ -278,17 +277,69 @@ func getKey(r io.Reader) (int, []byte, error) {
...
+
+// readTxOut parses a transaction output value and requires the full value to
+// be consumed.
 func readTxOut(txout []byte) (*wire.TxOut, error) {
-	if len(txout) < 10 {
-		return nil, ErrInvalidPsbtFormat
+	txOut := &wire.TxOut{}
+	reader := bytes.NewReader(txout)
+
+	if err := wire.ReadTxOut(reader, 0, 0, txOut); err != nil {
```
yyforyongyu, line 311: This introduces a memory-retention regression for parsed `WitnessUtxo` values. `wire.ReadTxOut` returns `TxOut.PkScript` as a slice into a 4 MiB `wire.scriptSlab`; `psbt/partial_input.go` then stores the returned `TxOut` directly as `pi.WitnessUtxo`, so each witness input can retain the whole slab even when the script is tiny.

Suggested fix: compact/copy the script after `wire.ReadTxOut` before returning it from `readTxOut`:

```go
script := make([]byte, len(txOut.PkScript))
copy(script, txOut.PkScript)
txOut.PkScript = script
```

A regression test can assert that parsed witness scripts do not retain the slab, e.g. `cap(packet.Inputs[0].WitnessUtxo.PkScript) == len(packet.Inputs[0].WitnessUtxo.PkScript)`.

starius: Great catch! Fixed in https://github.com/btcsuite/btcd/pull/2563

starius:
> One heads-up so it doesn't get blamed on this PR: `netsync` `TestSyncStateMachine` (`handleHeadersMsg should update lastProgressTime`) fails for me locally — but it fails identically on a clean `master` ([280d4d2](https://github.com/btcsuite/btcd/commit/280d4d2db6400136f284f121189d9c2942425a78)) and this PR doesn't touch `netsync`, so it looks like a pre-existing timing flake, unrelated here.

Fixed in https://github.com/btcsuite/btcd/pull/2564

Lrifton92:
Went back through the merged series once more — no concerns on correctness. Two non-blocking follow-ups worth tracking:

1. The `replace` directives in root `go.mod` (`btcutil/v2 => ./btcutil`, `wire/v2 => ./wire`) are fine for wiring the submodule fixes into the `rpc:*` commits, but they'll need to come back out once `btcutil`/`wire` are tagged, otherwise `go get` of the root module breaks for downstream. Might be worth a checklist note on the release so it doesn't get missed.
2. `NewTxFromBytes`/`NewBlockFromBytes` are now strict about trailing bytes — that's exactly the fix, and every in-tree caller passes exact-length input, but since `btcutil` is imported pretty broadly it might deserve a line in the release notes for anyone who was (accidentally) relying on the old lenient behavior.

Also nice bonus catch on the `readTxOut` rewrite — switching to `wire.ReadTxOut` fixes the multi-byte CompactSize script-length case the old `txout[9:]` slice got wrong, not just trailing bytes. The `0xfd` 300-byte-script test makes that explicit. 👍

PR #2561, opened 2026-07-01
btcec/schnorr: reject s >= group order in ParseSignature

Lrifton92:
### Change Description

`schnorr.ParseSignature` documents that it enforces the BIP-340 requirement that the `s` component lie in the range `[0, n-1]`, alongside the range check it already performs on `r`:

```go
// - The r component must be in the valid range for secp256k1 field elements
// - The s component must be in the valid range for secp256k1 scalars
```

The `r` check is implemented correctly (overflow → `ErrSigRTooBig`), but for `s` the overflow return value of `ModNScalar.SetByteSlice` was discarded:

```go
var s btcec.ModNScalar
s.SetByteSlice(sig[32:64])
```

Because `SetByteSlice` reduces its input modulo `n` and reports the overflow via its return value, an encoding with `s >= n` was **silently reduced modulo n and accepted** instead of being rejected. This diverges from:

- the function's own documented contract (and the inline comment stating `s` is enforced to be in `[0, n-1]`);
- **BIP-340**, whose verification algorithm requires *"fail if s >= n"*;
- the reference implementation in `decred/dcrd/dcrec/secp256k1/schnorr` — of which this file is a port — which rejects the same input with `ErrSigSTooBig`. The `ErrSigSTooBig` error kind is already defined in the imported `schnorr` package but was unused here, which is what tipped me off that the check was dropped during the port.

The fix mirrors the existing `r` check exactly:

```go
var s btcec.ModNScalar
if overflow := s.SetByteSlice(sig[32:64]); overflow {
	str := "invalid signature: s >= group order"
	return nil, signatureError(ecdsa_schnorr.ErrSigSTooBig, str)
}
```

`ParseSignature` is on the taproot signature-validation path (`txscript` key-path and script-path spends), so this also brings btcd's strictness in line with Bitcoin Core, which rejects non-canonical (`s >= n`) Schnorr encodings.

### Test

Adds `TestParseSignatureComponentRange`, reusing the reference `s == n` and `s > n` vectors. The test fails before this change (both encodings are accepted, returning `nil`) and passes after. Verified by reverting the one-line fix while keeping the test:

```
--- FAIL: TestParseSignatureComponentRange
    signature_test.go: s == n: mismatched err -- got <nil>, want ErrSigSTooBig
    signature_test.go: s > n:  mismatched err -- got <nil>, want ErrSigSTooBig
```

`go build ./... && go test ./...` and `go vet ./...` are green across the whole `btcec` module.

### Notes

This is intentionally minimal and independent of the other open schnorr PRs (#2546, #2501) — neither touches the `s`-range check in `ParseSignature` (I checked their diffs). Happy to rebase or fold this into #2546 if a maintainer prefers.

### DIFF
--- a/btcec/schnorr/signature_test.go
+++ b/btcec/schnorr/signature_test.go
@@ -291,3 +291,41 @@ func TestSchnorrSignNoMutate(t *testing.T) {
 		t.Fatalf("private key modified: %v", err)
 	}
 }
+
+// TestParseSignatureComponentRange ensures ParseSignature enforces the BIP-340
+// range restrictions on both the r and s components, in particular that an s
+// value greater than or equal to the group order n is rejected rather than
+// silently reduced modulo n.
+func TestParseSignatureComponentRange(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name string
+		sig  string
+		err  error
+	}{
+		{
+			name: "r == p",
+			sig:  "fffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2f181522ec8eca07de4860a4acdd12909d831cc56cbbac4622082221a8768d1d09",
+			err:  ecdsa_schnorr.ErrSigRTooBig,
+		},
+		{
+			name: "s == n",
+			sig:  "4e45e16932b8af514961a1d3a1a25fdf3f4f7732e9d624c6c61548ab5fb8cd41fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141",
+			err:  ecdsa_schnorr.ErrSigSTooBig,
+		},
+		{
+			name: "s > n",
+			sig:  "4e45e16932b8af514961a1d3a1a25fdf3f4f7732e9d624c6c61548ab5fb8cd41fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364142",
+			err:  ecdsa_schnorr.ErrSigSTooBig,
+		},
+	}
+
+	for _, test := range tests {
+		_, err := ParseSignature(decodeHex(test.sig))
+		if !errors.Is(err, test.err) {
+			t.Errorf("%s: mismatched err -- got %v, want %v",
+				test.name, err, test.err)
+		}
+	}
+}
--- a/btcec/schnorr/signature.go
+++ b/btcec/schnorr/signature.go
@@ -91,7 +91,10 @@ func ParseSignature(sig []byte) (*Signature, error) {
 		return nil, signatureError(ecdsa_schnorr.ErrSigRTooBig, str)
 	}
 	var s btcec.ModNScalar
-	s.SetByteSlice(sig[32:64])
+	if overflow := s.SetByteSlice(sig[32:64]); overflow {
+		str := "invalid signature: s >= group order"
+		return nil, signatureError(ecdsa_schnorr.ErrSigSTooBig, str)
+	}
 
 	// Return the signature.
 	return NewSignature(&r, &s), nil


btcec/schnorr/signature_test.go
```diff
@@ -291,3 +291,40 @@ func TestSchnorrSignNoMutate(t *testing.T) {
...
+		name: "s > n",
+		sig: "4e45e16932b8af514961a1d3a1a25fdf3f4f7732e9d624c6c61548ab5fb8cd41" +
+			"fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364142",
+		err: ecdsa_schnorr.ErrSigSTooBig,
+	}}
+
+	for _, test := range tests {
+		_, err := ParseSignature(decodeHex(test.sig))
+		if !errors.Is(err, test.err) {
+			t.Errorf("%s: mismatched err -- got %v, want %v", test.name,
```
allocz: If the objective is to wrap at 80 columns, the wrapping is at the wrong position. Maybe your editor tab size is less than 8. The brackets are also not in the right place, use:
```
tests := []struct{
}{
        {
        },
        {
        },
}
```

Image

?

allocz (changes requested):
Nice catch. In fact BIP-340 fails the verification if `s` is >= `N`, as we can see [here](https://github.com/bitcoin/bips/blob/f078d4f5ff70fe2921fd005d799a3d4a8ff7d55c/bip-0340.mediawiki?plain=1#L184).

Confirmed that the tests are passing and reverting `signature.go` causes the test to fail.

Lrifton92:
Thanks for the review @allocz — and for confirming the fix against BIP-340.

Addressed both formatting notes in the test:
- Re-wrapped the `t.Errorf` call so no line exceeds 80 columns at tab width 8 (my editor was rendering tabs narrower — good catch).
- Switched the table to the expanded struct-literal layout you showed (each case in its own `{ ... },` block).

`gofmt`/`go vet` clean and the full `btcec/schnorr/...` suite passes. The `signature.go` fix itself is unchanged.

btcec/schnorr/signature_test.go
```diff
@@ -291,3 +291,44 @@ func TestSchnorrSignNoMutate(t *testing.T) {
...
+		{
+			name: "s == n",
+			sig: "4e45e16932b8af514961a1d3a1a25fdf3f4f7732e9d624c6c61548ab5fb8cd41" +
+				"fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141",
+			err: ecdsa_schnorr.ErrSigSTooBig,
+		},
+		{
+			name: "s > n",
+			sig: "4e45e16932b8af514961a1d3a1a25fdf3f4f7732e9d624c6c61548ab5fb8cd41" +
+				"fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364142",
```
allocz: The line wrap still after the 80 column limit. The formatting rules of the project states that the line should not have more than 80 columns. I personally don't like wrapping strings, but as you choose to wrap them, I recommend keeping max of 80 chars per line. If you don't wrap the strings at all, I will also approve the PR.

Image

allocz (changes requested):

Lrifton92:
Thanks @allocz — you're right, the wrapped continuation still exceeded 80 columns. I've now unwrapped the sig strings entirely (single line each, as you suggested you'd approve), and re-ran `gofmt` so the struct fields realign cleanly. Test still passes. Ready for another look.

allocz (approved):

Lrifton92:
Thanks again @allocz for the review and approval. Just flagging that this is approved, all threads resolved, no conflicts — is there anything else needed on my side before it can be merged? Happy to rebase if required.

Roasbeef:
CI run approved!

Roasbeef (approved):
LGTM 🙊

PR #2564, opened 2026-07-02
netsync: avoid clock tick race in sync state test

starius:
## Change Description
`TestSyncStateMachine` checked that `handleHeadersMsg` advanced `lastProgressTime` by comparing it against the timestamp written by `startSync`. Those two writes can happen within the same clock tick, so the handler can update the field while `time.After` still reports false.

Reset `lastProgressTime` to the zero value before delivering headers and assert that the handler writes a non-zero value. This keeps the test focused on the behavior under test without depending on adjacent time.Now calls producing distinct timestamps.

## Steps to Test
```
go test ./netsync -run TestSyncStateMachine -count=50
```

### Testing

### DIFF
--- a/netsync/manager_test.go
+++ b/netsync/manager_test.go
@@ -869,9 +869,10 @@ func syncSendHeaders(t *testing.T, sm *SyncManager,
 
 	t.Helper()
 
-	// Record the progress time set by startIBD so we can verify
-	// that handleHeadersMsg advances it.
-	progressBefore := sm.lastProgressTime
+	// Reset the progress time to a zero sentinel so the assertion below
+	// verifies that handleHeadersMsg writes it without depending on the
+	// system clock advancing between calls to time.Now.
+	sm.lastProgressTime = time.Time{}
 
 	headers := wire.NewMsgHeaders()
 	for _, block := range blocks {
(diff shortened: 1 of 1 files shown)


allocz:
I could not trigger the tick race here, but depending of the processor speed I agree that it may happen, so I'll give an ACK.

allocz (approved):

TechLateef (commented):
Good fix!

Lrifton92 (approved):
Verified locally on this branch: `go test ./netsync -run TestSyncStateMachine -count=50` passes (18s, Windows). For context, I hit this exact flake while testing #2558 on a clean master — it reproduced there too, so it's a genuine pre-existing timing issue, not related to any recent change.

The zero-sentinel is the right idiom here: the old assertion depended on two adjacent `time.Now()` calls landing on different clock ticks, which is not guaranteed on coarse-granularity clocks (Windows is ~15ms). And the relaxed assertion (non-zero instead of monotonic advance) still catches the regression under test — if `handleHeadersMsg` stops writing `lastProgressTime`, it stays at the zero value and the test fails.

Roasbeef (approved):
LGTM 🥝

PR #2565, opened 2026-07-02
wire: replace 4 MiB script slab with tiered arena allocator

Roasbeef:
In this PR, we replace the fixed 4 MiB script decode slab with a tiered arena allocator. Today every transaction decode borrows the full slab, even when the transaction contains only a few hundred bytes of script data. The channel-backed slab pool can also retain up to 125 slabs, or roughly 524 MB, outside the runtime's GC pressure heuristics. `ReadTxOut` goes a step further and returns a script backed by its staging slab, which keeps the full allocation live with the output.

The new arena draws from 16 KiB, 128 KiB, 1 MiB, and 4 MiB chunk classes. Standalone transactions and blocks both begin at 16 KiB. Block decoding shares one arena across the block and rewinds it after each transaction has copied its scripts into owned storage, so its live staging requirement is determined by the largest transaction rather than the aggregate block. Larger transactions promote through the chunk ladder on demand. The existing 4 MiB limit on staged bytes remains unchanged.

## Pooling and ownership

Each size class uses a `sync.Pool`, which keeps idle chunks visible to the runtime across GC cycles. The 1 MiB and 4 MiB classes also have bounded front caches of 16 and 8 chunks. Those caches cap deterministic retention at 48 MiB. Bursts beyond the front caches fall back to `sync.Pool` and remain reclaimable by the GC.

Arena storage has one lexical owner. A successful transaction decode copies every signature script, witness item, and output script into message-owned storage before the arena is rewound or released. `ReadTxOut` now makes the same ownership transfer and returns an exact-sized script. Release is idempotent, and use after release is rejected until the arena handle is borrowed again.

## Benchmarks

The following results compare public master at `ca0bb02cf` with this branch at `850492f46` on an Apple M4 Max using Go 1.26.1 on darwin/arm64.

The cold allocation measurements ran each benchmark once in a fresh process, repeated across ten processes:

```
                      │     master      │     arena      │
                      │      B/op       │      B/op      │
ReadTxOut-16             4096.20Ki ± 0%    20.85Ki ± 0%  -99.49% (p=0.000 n=10)
DeserializeTxSmall-16    4101.48Ki ± 0%    20.91Ki ± 0%  -99.49% (p=0.000 n=10)
DeserializeBlock-16        7.256Mi ± 0%     3.408Mi ± 0%  -53.03% (p=0.000 n=10)
```

Command, repeated ten times per benchmark and revision:

```
go test -run '^$' -bench '^Benchmark<Name>$' -benchmem -benchtime=1x -count=1 -cpu=16
```

The one-iteration timing samples are dominated by process and first-use noise, so the cold comparison is limited to allocation footprint.

Steady serial decode used eight one-second samples:

```
                      │    master     │    arena     │
                      │    sec/op     │    sec/op    │
ReadTxOut-16             66621.00n ± 21%   78.28n ± 3%  -99.88% (p=0.000 n=8)
DeserializeTxSmall-16       182.1n ±  1%   193.0n ± 5%    +6.01% (p=0.000 n=8)
DeserializeTxLarge-16       261.3µ ±  1%   260.3µ ± 1%         ~ (p=0.574 n=8)
DeserializeBlock-16         1.031m ±  2%   1.031m ± 2%         ~ (p=0.505 n=8)
```

```
go test -run '^$' -bench '^(BenchmarkReadTxOut|BenchmarkDeserializeTxSmall|BenchmarkDeserializeTxLarge|BenchmarkDeserializeBlock)$' -benchmem -benchtime=1s -count=8 -cpu=16
```

The parallel benchmark functions are new in this branch. For the base comparison, the same `bench_parallel_test.go` harness was applied unchanged to a detached master checkout. Ten base/head process pairs were run in alternating order:

```
                              │   master    │    arena    │
                              │   sec/op    │   sec/op    │
DeserializeTxLargeParallel-16    35.97µ ± 13%   48.87µ ± 9%  +35.85% (p=0.000 n=10)
DeserializeBlockParallel-16      220.2µ ± 14%   266.0µ ± 8%  +20.80% (p=0.002 n=10)
```

These parallel results expose the main throughput tradeoff for large transactions and blocks. The steady serial large-transaction and block paths are unchanged within noise, while parallel execution pays for allocator and pool coordination. The warm `B/op` figures do not include staging chunks retained by the pools, so they should not be read as peak or retained-memory measurements.

## Testing

The change adds table-driven allocator tests, Rapid property tests, concurrent decode coverage, and transaction/block fuzz targets. The ownership tests overwrite the exact chunks used by a decode before reserializing its result. The fuzz targets also reserialize the first decoded object after a second decode, checking that arena reuse leaves the earlier value stable.

The full repository unit suite, the wire suite under the race detector, `go vet`, and `git diff --check` pass on this branch.

### DIFF
--- a/wire/arena_rapid_test.go
+++ b/wire/arena_rapid_test.go
@@ -1,0 +1,295 @@
+// Copyright (c) 2026 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package wire
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"pgregory.net/rapid"
+)
+
+// poisonScriptChunkPools scribbles a poison pattern over chunks sitting in
+// the class pools.  Any decoded message that still aliases arena memory
+// (rather than owning an exact-size copy of its scripts) will have its
+// contents corrupted by this, so re-verifying a message after poisoning
+// proves that no arena memory escaped the decode.
+func poisonScriptChunkPools() {
+	for _, pool := range scriptChunkPools {
+		// Pull a handful of chunks out of each pool, poison them, and
+		// put them back.  This reaches the chunks released by recent
+		// decodes on this P.
+		var chunks []*[]byte
+		for i := 0; i < 4; i++ {
+			chunks = append(chunks, pool.get())
+		}
+		for _, chunk := range chunks {
+			for i := range *chunk {
+				(*chunk)[i] = 0xaa
+			}
+			pool.put(chunk)
+		}
+	}
+}
+
+// genScript returns a rapid generator for a non-nil script of up to maxLen
+// bytes.  Zero-length scripts are always non-nil so that decoded messages,
+// which never produce nil scripts, compare equal structurally.
+func genScript(rt *rapid.T, label string, maxLen int) []byte {
+	n := rapid.IntRange(0, maxLen).Draw(rt, label+"Len")
+	s := make([]byte, n)
+	for i := 0; i < n; i++ {
+		s[i] = byte(rapid.IntRange(0, 255).Draw(rt, label+"Byte"))
+	}
+	return s
+}
+
+// genMsgTx generates a structurally valid random transaction.  Script and
+// witness sizes are biased small like real traffic but occasionally exceed
+// the 16 KiB starting chunk class so the arena growth path is exercised.
+func genMsgTx(rt *rapid.T) *MsgTx {
+	tx := NewMsgTx(int32(rapid.IntRange(0, 2).Draw(rt, "version")))
+
+	// Most generated scripts are small, but roughly one in ten draws
+	// from a range that overflows the first chunk class.
+	scriptMax := 512
+	if rapid.IntRange(0, 9).Draw(rt, "bigScripts") == 0 {
+		scriptMax = 3 * scriptChunkClasses[0]
+	}
+
+	numIn := rapid.IntRange(1, 6).Draw(rt, "numIn")
+	hasWitness := rapid.Bool().Draw(rt, "hasWitness")
+	for i := 0; i < numIn; i++ {
+		ti := &TxIn{
+			PreviousOutPoint: OutPoint{
+				Index: uint32(rapid.IntRange(0, 1<<30).Draw(
+					rt, "prevIndex",
+				)),
+			},
+			SignatureScript: genScript(rt, "sigScript", scriptMax),
+			Sequence: uint32(
+				rapid.IntRange(0, 1<<30).Draw(rt, "sequence"),
+			),
+		}
+		for j := range ti.PreviousOutPoint.Hash {
+			ti.PreviousOutPoint.Hash[j] = byte(
+				rapid.IntRange(0, 255).Draw(rt, "prevHash"),
+			)
+		}
+
+		// When the transaction is witnessy every input carries a
+		// (possibly empty) non-nil witness stack, matching what the
+		// decoder produces.
+		if hasWitness {
+			numItems := rapid.IntRange(0, 3).Draw(rt, "numWitness")
+			ti.Witness = make(TxWitness, numItems)
+			for j := 0; j < numItems; j++ {
+				ti.Witness[j] = genScript(
+					rt, "witnessItem", scriptMax,
+				)
+			}
+		}
+
+		tx.AddTxIn(ti)
+	}
+
+	// A witness marker with zero total witness items is rejected by the
+	// decoder (and never produced by the encoder), so force at least one
+	// item when the transaction claims to be witnessy.
+	if hasWitness && !tx.HasWitness() {
+		tx.TxIn[0].Witness = TxWitness{
+			genScript(rt, "forcedWitness", scriptMax),
+		}
+	}
+
+	numOut := rapid.IntRange(0, 6).Draw(rt, "numOut")
+	for i := 0; i < numOut; i++ {
+		tx.AddTxOut(&TxOut{
+			Value: int64(
+				rapid.IntRange(0, 1<<40).Draw(rt, "value"),
+			),
+			PkScript: genScript(rt, "pkScript", scriptMax),
+		})
+	}
+
+	tx.LockTime = uint32(rapid.IntRange(0, 1<<30).Draw(rt, "lockTime"))
+
+	return tx
+}
+
+// TestScriptArenaPropertyTxRoundTrip checks two properties over random
+// transactions: serialize/deserialize is the identity, and the decoded
+// transaction owns all of its memory (poisoning the arena chunk pools after
+// the decode must not change it).
+func TestScriptArenaPropertyTxRoundTrip(t *testing.T) {
+	rapid.Check(t, func(rt *rapid.T) {
+		tx := genMsgTx(rt)
+
+		var wireBuf bytes.Buffer
+		require.NoError(rt, tx.Serialize(&wireBuf))
+
+		var decoded MsgTx
+		err := decoded.Deserialize(bytes.NewReader(wireBuf.Bytes()))
+		require.NoError(rt, err)
+
+		var reserialized bytes.Buffer
+		require.NoError(rt, decoded.Serialize(&reserialized))
+		require.Equal(rt, wireBuf.Bytes(), reserialized.Bytes())
+
+		// If the decoded transaction aliased arena memory, poisoning
+		// the pools would corrupt its scripts and the second
+		// serialization would differ.
+		poisonScriptChunkPools()
+
+		var afterPoison bytes.Buffer
+		require.NoError(rt, decoded.Serialize(&afterPoison))
+		require.Equal(rt, wireBuf.Bytes(), afterPoison.Bytes())
+	})
+}
+
+// TestScriptArenaPropertyBlockRoundTrip checks the same round-trip and
+// ownership properties for full blocks, which share a single arena across
+// all of their transactions.
+func TestScriptArenaPropertyBlockRoundTrip(t *testing.T) {
+	rapid.Check(t, func(rt *rapid.T) {
+		block := &MsgBlock{Header: blockOne.Header}
+		numTxs := rapid.IntRange(1, 8).Draw(rt, "numTxs")
+		for i := 0; i < numTxs; i++ {
+			block.AddTransaction(genMsgTx(rt))
+		}
+
+		var wireBuf bytes.Buffer
+		require.NoError(rt, block.Serialize(&wireBuf))
+
+		var decoded MsgBlock
+		err := decoded.Deserialize(bytes.NewReader(wireBuf.Bytes()))
+		require.NoError(rt, err)
+
+		var reserialized bytes.Buffer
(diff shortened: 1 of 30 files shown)


allocz:
Ran the benchmarks here.

```
goos: linux
goarch: amd64
pkg: github.com/btcsuite/btcd/wire/v2
                             │  /tmp/before   │             /tmp/after              │
                             │     sec/op     │   sec/op     vs base                │
DeserializeTxSmallParallel-8      654.4n ± 0%   597.3n ± 1%   -8.72% (p=0.000 n=12)
DeserializeTxLargeParallel-8      366.9µ ± 0%   387.1µ ± 0%   +5.50% (p=0.000 n=12)
DeserializeBlockParallel-8        1.319m ± 1%   1.529m ± 5%  +15.95% (p=0.000 n=12)
ReadTxOut-8                    254955.0n ± 1%   285.3n ± 1%  -99.89% (p=0.000 n=12)
ReadTxIn-8                        81.63n ± 0%   85.14n ± 0%   +4.29% (p=0.000 n=12)
DeserializeTxSmall-8              663.8n ± 0%   690.1n ± 1%   +3.95% (p=0.000 n=12)
DeserializeTxLarge-8              792.1µ ± 1%   876.5µ ± 1%  +10.65% (p=0.000 n=12)
DeserializeBlock-8                3.102m ± 1%   3.374m ± 1%   +8.77% (p=0.000 n=12)
geomean                           42.55µ        19.08µ       -55.17%

                             │    /tmp/before    │               /tmp/after                │
                             │       B/op        │     B/op      vs base                   │
DeserializeTxSmallParallel-8        225.0 ± 0%       227.0 ± 0%    +0.89% (p=0.000 n=12)
DeserializeTxLargeParallel-8      1.305Mi ± 0%     1.310Mi ± 0%    +0.38% (p=0.000 n=12)
DeserializeBlockParallel-8        3.256Mi ± 0%     3.256Mi ± 0%    +0.00% (p=0.000 n=12)
ReadTxOut-8                    4194305.00 ± 0%       80.00 ± 0%  -100.00% (p=0.000 n=12)
ReadTxIn-8                          0.000 ± 0%       0.000 ± 0%         ~ (p=1.000 n=12) ¹
DeserializeTxSmall-8                225.0 ± 0%       225.0 ± 0%         ~ (p=1.000 n=12) ¹
DeserializeTxLarge-8              1.305Mi ± 0%     1.312Mi ± 0%    +0.57% (p=0.000 n=12)
DeserializeBlock-8                3.256Mi ± 0%     3.256Mi ± 0%    +0.01% (p=0.000 n=12)
geomean                                        ²                  -74.23%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean

                             │  /tmp/before  │              /tmp/after              │
                             │   allocs/op   │  allocs/op   vs base                 │
DeserializeTxSmallParallel-8    6.000 ± 0%      6.000 ± 0%       ~ (p=1.000 n=12) ¹
DeserializeTxLargeParallel-8    6.000 ± 0%      6.000 ± 0%       ~ (p=1.000 n=12) ¹
DeserializeBlockParallel-8     8.323k ± 0%     8.323k ± 0%       ~ (p=1.000 n=12) ¹
ReadTxOut-8                     1.000 ± 0%      1.000 ± 0%       ~ (p=1.000 n=12) ¹
ReadTxIn-8                      0.000 ± 0%      0.000 ± 0%       ~ (p=1.000 n=12) ¹
DeserializeTxSmall-8            6.000 ± 0%      6.000 ± 0%       ~ (p=1.000 n=12) ¹
DeserializeTxLarge-8            6.000 ± 0%      6.000 ± 0%       ~ (p=1.000 n=12) ¹
DeserializeBlock-8             8.323k ± 0%     8.323k ± 0%       ~ (p=1.000 n=12) ¹
geomean                                    ²                +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean
```

To reduce variance I ran a stress test with N cores in a busy loop until discover the final frequency after the CPU heats up, then used `cpupower` to limit the frequency of all cores.

wire/common.go
```diff
@@ -22,8 +23,97 @@ const (
...
+	remaining, known := readerRemaining(r)
+	if known && count <= remaining {
+		result := make([]byte, count)
+		n, err := io.ReadFull(r, result)
+		return result[:n], err
+	}
+
+	var result bytes.Buffer
+	result.Grow(int(min(count, defaultReadBufferSize)))
+	n, err := io.CopyN(&result, r, int64(count))
```
GeorgeTsagk, line 55: `io.CopyN` invokes `bytes.Buffer.ReadFrom`, which grows even before its final EOF check. Complete framed reads increase allocations from 216 to 1,888 B/op for ping and 5.68 to 11.84 MB/op for the block fixture. The deserialize benchmarks miss this path.

Could we bound growth by the payload length and add complete-message streaming benchmarks?

wire/arena_rapid_test.go
```diff
@@ -0,0 +1,314 @@
...
+
+		var afterPoison bytes.Buffer
+		require.NoError(rt, decoded.Serialize(&afterPoison))
+		require.Equal(rt, wireBuf.Bytes(), afterPoison.Bytes())
+	})
+}
+
+// TestScriptArenaPropertyBlockRoundTrip checks round-trip stability for full
+// blocks, including the rewinds performed while transactions share an arena.
+func TestScriptArenaPropertyBlockRoundTrip(t *testing.T) {
```
GeorgeTsagk, line 148: Could we assert per-transaction staging reuse? Removing `ar.rewind()` still leaves the existing wire suite passing. This test covers both block decoders: it passes now and fails without the rewind because small transactions borrow a larger chunk.

<details>
<summary>Regression test for block arena reuse</summary>

Save as `wire/review_footprint_test.go`:

```go
package wire

import (
	"bytes"
	"sync"
	"testing"
)

// TestReviewBlockArenaFootprint requires staging memory to depend on the
// largest transaction, even when a block contains many small transactions.
func TestReviewBlockArenaFootprint(t *testing.T) {
	for _, locations := range []bool{false, true} {
		name := "deserialize"
		if locations {
			name = "transaction_locations"
		}
		t.Run(name, func(t *testing.T) {
			original := scriptChunkPools
			t.Cleanup(func() { scriptChunkPools = original })
			var allocated [len(scriptChunkClasses)]int
			for i, size := range scriptChunkClasses {
				scriptChunkPools[i] = &chunkClassPool{
					fixed: make(chan *[]byte, scriptChunkFixedCaps[i]),
					pool: sync.Pool{New: func() any {
						allocated[i]++
						chunk := make([]byte, size)
						return &chunk
					}},
				}
			}
			block := MsgBlock{Header: blockOne.Header}
			for i := 0; i < 1000; i++ {
				block.AddTransaction(blockOne.Transactions[0])
			}
			var encoded bytes.Buffer
			if err := block.Serialize(&encoded); err != nil {
				t.Fatal(err)
			}
			var decoded MsgBlock
			if locations {
				locs, err := decoded.DeserializeTxLoc(bytes.NewBuffer(encoded.Bytes()))
				if err != nil {
					t.Fatal(err)
				}
				if len(locs) != len(block.Transactions) {
					t.Fatal("wrong location count")
				}
			} else if err := decoded.Deserialize(bytes.NewReader(encoded.Bytes())); err != nil {
				t.Fatal(err)
			}
			var roundTrip bytes.Buffer
			if err := decoded.Serialize(&roundTrip); err != nil {
				t.Fatal(err)
			}
			if !bytes.Equal(encoded.Bytes(), roundTrip.Bytes()) {
				t.Fatal("block changed during decoding")
			}
			for i, count := range allocated {
				want := 0
				if i == txScriptChunkClass {
					want = 1
				}
				if count != want {
					t.Errorf("class %d allocated %d chunks, want %d", i, count, want)
				}
			}
		})
	}
}
```

From the `wire` directory:

```sh
go test -run '^TestReviewBlockArenaFootprint$' -count=1
```

Temporarily remove `ar.rewind()` from `MsgTx.btcDecode` to see the test fail; restore it afterward.

</details>

GeorgeTsagk (changes requested):
Please address the streaming allocation regression and add a block staging reuse assertion. Refresh the benchmarks for the current revision, including complete streaming reads and retained heap.

Roasbeef:
Thanks, both comments were valid. I addressed the streaming path in `34ad12219` and added the block-level reuse assertion in `133e4897d`.

The opaque-reader fallback previously used `io.CopyN` with a `bytes.Buffer`. `CopyN` dispatches to `Buffer.ReadFrom`, which reserves additional space before its final EOF check. The replacement reads geometrically sized chunks capped by the remaining payload, returns the first chunk directly for small messages, and joins larger messages into exact-sized storage after the complete read. The loop also preserves an error returned with a full internal chunk when the overall payload remains incomplete.

I reran the complete framed-message benchmark at the old PR head and the current head with Go 1.26.1 on darwin/arm64 (Apple M4 Max), using ten one-second samples. This is the direct `benchstat` comparison:

```text
                                             │ old PR head │ current head │
                                             │   sec/op    │ sec/op vs base
ReviewCompleteStreamingRead/message/ping-16     546.8n ± 3%   323.4n ± 2%  -40.86% (p=0.000 n=10)
ReviewCompleteStreamingRead/message/block-16    2.045m ± 1%   2.001m ± 2%   -2.20% (p=0.007 n=10)

                                             │ old PR head │ current head │
                                             │    B/op     │ B/op vs base
ReviewCompleteStreamingRead/message/ping-16      1960 ± 0%      288 ± 0%   -85.31% (p=0.000 n=10)
ReviewCompleteStreamingRead/message/block-16   11.292Mi ± 0%  7.630Mi ± 0% -32.43% (p=0.000 n=10)

                                             │ old PR head │ current head │
                                             │ allocs/op   │ allocs/op vs base
ReviewCompleteStreamingRead/message/ping-16      15.00 ± 0%    11.00 ± 0%  -26.67% (p=0.000 n=10)
ReviewCompleteStreamingRead/message/block-16     8.353k ± 0%   8.352k ± 0%  -0.01% (p=0.000 n=10)
```

The timing change is small for the block case, but the allocation reduction is consistent: 85% fewer bytes for ping and 32% fewer bytes for the block fixture.

I also reran the retained-heap harness against the current revision. It starts 16 block decoders together, waits until each has acquired staging storage, completes the decodes, then runs two collections plus `debug.FreeOSMemory`. Across five fresh processes, the median retained `HeapAlloc` delta was 67,167,576 bytes on current master and 58,968 bytes on this branch.

Finally, `TestBlockArenaTransactionReuse` now covers both `Deserialize` and `DeserializeTxLoc` with 1,000 small transactions and isolated chunk pools. It requires the whole block to use one 16 KiB staging chunk. Removing the per-transaction rewind promotes the decode into a larger class and fails the assertion even though the decoded block still round trips correctly.

wire/common.go
Roasbeef: Adopted a tiered decode/read here instead

GeorgeTsagk (approved):
LGTM, both comments are addressed.

Nit: please refresh the benchmark results in the description. Complete streamed block reads still allocate about 8.04 MB/op versus 5.68 MB/op on the base, down from 11.84 MB/op before the follow-up. Worth documenting the remaining staging/copying cost.

Issue #2560, opened 2026-06-30
[feature]: Improve test harness to enable more complex test cases


allocz:
The current Harness under `integration/rpctest/rpc_harness.go` would be more powerful if we add the following features:

1. Log assertion: an option accepting a func(string) which is called for each btcd log line, allowing matching on log messages.
2. Stop procedure: stops btcd returning the exit code, to test scenarios where btcd exits with error or scenarios where the node should be restarted.
3. Start procedure: allows the harness to be started again after `stop` is called.
4. Change `New()` signature: pass opts as single nullable pointer to struct instead of having to pass multiple zero values to get default initialization.

This would make possible write tests for complex scenarios, and create incentives to write more integration tests.

`1` and `4` already addressed on https://github.com/allocz/btcd/pull/7, which currently allow setting an alternative node directory to work around the missing `2` and `3`.

Roasbeef:
> Log assertion: an option accepting a func(string) which is called for each btcd log line, allowing matching on log messages.

Hmm, that seems sort of brittle: logging assumptions are directly tied to test infra. 

The rest of the suggestions make a lot of sense IMO. 

Instead of tying to logging, there should be some build tagged dev event stream infrastructure, which can be used to make lower level assertions. However stepping back the whole point of the `rpctest` framework is to exercise daemon capabilities as visible/driven-by the RPC system.

allocz:
Nice, something like `type event struct { code uint64; data []byte }` for the event format? Would be easy to implement the encode/decode procedures for a byte stream, code: (8 bytes); dataLen: (8 bytes); data: (arbitrary size).

PR #2571, opened 2026-07-07
integration/rpctest: improve rpctest harness

allocz:
## Change Description

The main motivation of this PR is to improve the process of writing integration
tests, currently there's no way to write complex tests in a simple way.

As discussed in #2560.

Changes
1. Debug event stream: connect to btcd and receive debug events.
2. Harness teardown: stops btcd returning the exit code, but allow stopping
   without cleaning up node data, enabling the test of scenarios where btcd
   exits with error or scenarios where the node should be restarted.
3. Harness start: allow the harness to be started again after teardown is
   called.
4. Change Harness related procedure signatures: pass opts as a struct instead
   of having to pass multiple zero values to get default initialization.

The debug event stream is replaced by a NOP implementation when the compile tag
debug is not defined, so that the compiler can optimize out all the debug event
broadcast calls. Also, the hidden flag --debugstream=<host:port> should be used
to start the event stream.

There's also bugfixes, the complete changelog is in the commit message.

## Steps to Test

```
make unit
make unit-cover
make unit-race
```

### Testing

### DIFF
--- a/debugstream/client.go
+++ b/debugstream/client.go
@@ -1,0 +1,113 @@
+package debugstream
+
+import (
+	"fmt"
+	"net"
+	"time"
+)
+
+// Client calls handler with events coming from the [StreamServer].
+type Client struct {
+	addr string
+
+	conn    *net.TCPConn
+	handler func(e Event)
+
+	stop chan struct{}
+	done chan struct{}
+}
+
+// NewClient initializes a [Client] instance.
+func NewClient(streamAddr string, handler func(ev Event)) *Client {
+	done := make(chan struct{})
+	close(done)
+	stop := make(chan struct{})
+	close(stop)
+	return &Client{
+		addr:    streamAddr,
+		handler: handler,
+		done:    done,
+		stop:    stop,
+	}
+}
+
+func (c *Client) connect() (*net.TCPConn, error) {
+	const maxConnAttempts = 7
+	var (
+		conn net.Conn
+		err  error
+	)
+
+	for i := range maxConnAttempts {
+		if i > 0 {
+			time.Sleep((time.Millisecond * 100) << (i - 1))
+		}
+		conn, err = net.Dial("tcp", c.addr)
+		if err != nil {
+			continue
+		}
+		break
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	return conn.(*net.TCPConn), nil
+}
+
+func (c *Client) loop() {
+	for {
+		var ev Event
+		err := ev.read(c.conn)
+		if err == nil {
+			c.handler(ev)
+			continue
+		}
+		select {
+		case <-c.stop:
+			return
+		default:
+			cliLog.Errorf("client loop %s", err)
+			return
+		}
+	}
+}
+
+func (c *Client) Start() error {
+	select {
+	case <-c.done:
+		c.done = make(chan struct{})
+	default:
+		return fmt.Errorf("client running")
+	}
+
+	conn, err := c.connect()
+	if err != nil {
+		close(c.done)
+		return err
+	}
+
+	c.conn = conn
+	c.stop = make(chan struct{})
+
+	go func() {
+		c.loop()
+		close(c.done)
+	}()
+
+	return nil
+}
+
+func (c *Client) Stop() {
+	select {
+	case <-c.stop:
+		return
+	default:
+	}
+
+	close(c.stop)
+	c.conn.Close()
+	<-c.done
+
+	c.conn = nil
+}
--- a/integration/rpctest/node.go
+++ b/integration/rpctest/node.go
@@ -5,6 +5,7 @@
 package rpctest
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"os"
@@ -30,6 +31,7 @@ type nodeConfig struct {
 	profile    string
 	debugLevel string
 	extra      []string
+	startArgs  []string
 	nodeDir    string
 
 	exe          string
@@ -131,6 +133,7 @@ func (n *nodeConfig) arguments() []string {
 		args = append(args, fmt.Sprintf("--debuglevel=%s", n.debugLevel))
 	}
 	args = append(args, n.extra...)
+	args = append(args, n.startArgs...)
 	return args
 }
 
@@ -166,6 +169,8 @@ type node struct {
 	pidFile string
 
 	dataDir string
+
+	started bool
 }
 
 // newNode creates a new node instance according to the passed config. dataDir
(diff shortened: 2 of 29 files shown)


integration/rpctest/rpc_harness.go
```diff
@@ -127,30 +127,49 @@ type Harness struct {
...
-// Optionally, websocket handlers and a specified configuration may be passed.
-// In the case that a nil config is passed, a default configuration will be
-// used. If a custom btcd executable is specified, it will be used to start the
-// harness node. Otherwise a new binary is built on demand.
 //
 // NOTE: This function is safe for concurrent access.
-func New(activeNet *chaincfg.Params, handlers *rpcclient.NotificationHandlers,
-	extraArgs []string, customExePath string) (*Harness, error) {
-
+func New(opts ...HarnessOpts) (*Harness, error) {
```
TechLateef: The `...HarnessOpts` parameter seems a little misleading since only the first element is ever used. Would it make sense to accept a `*HarnessOpts` instead, with `nil` representing the default configuration? That would make it clearer that only a single optional configuration is expected and avoid callers accidentally passing multiple option structs that are silently ignored.

integration/rpctest/node.go
```diff
@@ -242,13 +262,24 @@ func (n *node) cleanup() error {
 // shutdown terminates the running btcd process, and cleans up all
```
TechLateef, line 254: Here in `shutdown`, the `else if err != nil { return } `path skips `cleanup() `entirely when `stop()` returns a non-exit error.
This means the pid file and temp directory are left behind if the process fails unexpectedly.
Worth documenting whether this is intentional—callers who need to restart after an unexpected failure would need to handle cleanup themselves.

debugstream/client.go
```diff
@@ -0,0 +1,121 @@
...
+
+	const maxConnAttempts = 7
+
+	var (
+		conn net.Conn
+		err  error
+	)
+	for i := range maxConnAttempts {
+		if i > 0 {
+			time.Sleep((time.Millisecond*100)<<i - 1)
```
TechLateef: I think the backoff calculation here may not be doing what's intended. Since `<<` has higher precedence than `-`, the expression is evaluated as `((time.Millisecond * 100) << i) - 1`. That means when `i == 1`, the delay becomes `200ms - 1ns` rather than `100ms`. Would wrapping it as `(time.Millisecond * 100) << (i - 1)` better match the intended exponential backoff?

debugstream/client.go
```diff
@@ -0,0 +1,121 @@
...
+func NewClient(streamAddr string, handler func(ev Event)) *Client {
+	return &Client{
+		addr:              streamAddr,
+		eventCh:           make(chan Event),
+		handler:           handler,
+		shutdownRequested: make(chan struct{}),
+	}
+}
+
+func (c *Client) loop() {
```
TechLateef: if `loop()` exits unexpectedly, it looks like the client stays in the `csStarted` state. Would it make sense to transition back to `csStopped` (or otherwise notify the caller) so `Start()` can be called again?

integration/rpctest/rpc_harness.go
allocz: I did this before, but from the point of view of the caller, I think that is nice to not pass arguments at all to get the default configuration. I tried functional options too, but when you're passing multiple options the code becomes cluttered, for example:

Option 1 (your suggestion): 
```
h, err := rpctest.New(nil)
```

Option 2 (the way I think is cleaner):
```
h, err := rpctest.New()
```

When multiple options are passed in both pointer to struct or variadic struct (the only difference is that with variadics, I don't pass a pointer, but the struct itself):
```nctional options we need to pass the name of the package multiple times, and also in the case where we have multiple options it becomes harder to read mostly due to the project formatting rules, which does not allow passing one argument per li
h, err := rpctest.New(rpctest.HarnessOpts{
        Option1: value,
        Option2: value,
})
```

And now with functional options `func New(opts ...HarnessOptFunc) (*Harness, error)`:
```
h, err := rpctest.New(rpctest.Option1(value), rpctest.Option2(value))
```

Note that with functional options we need to pass the name of the package multiple times, and also in the case where we have multiple options it becomes harder to read mostly due to the project formatting rules, which as far as I know, does not allow passing one argument per line, like:
```
h, err := rpctest.New(
        rpctest.LongLongLongLongNameOption1(value),
        rpctest.LongLongLongLongNameOption2(value),
)
```

In the point of view of the caller, what do you think is better?

debugstream/client.go
allocz: If loop exits, nothing happens. [Client] should only be started again after calling [Client.Stop], which will set the state to `csStopped`.

integration/rpctest/rpc_harness.go
TechLateef: You have a point  `rpctest.New()` for defaults is cleaner than `rpctest.New(nil)`, and the functional options approach does get cluttered with repeated package names. The variadic struct is a reasonable tradeoff.

debugstream/client.go
TechLateef: The log message in the `default` branch helps, but would it also make sense to update `state` to `csStopped` when the loop exits unexpectedly? That way `Start()` could be called directly without needing `Stop()` first.

debugstream/common.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		return err
+	}
+	var dataLen uint64
+	err = binary.Read(r, binary.BigEndian, &dataLen)
+	if err != nil {
+		return err
+	}
+	e.Data = make([]byte, dataLen)
+	n, err := r.Read(e.Data)
```
TechLateef: Would `io.ReadFull` be more appropriate here? `io.Reader.Read` isn't guaranteed to fill the buffer in a single call, so a partial read could result in `n != dataLen` even though more data is available.

allocz:
For now I think that we achieved what was proposed. Ready for review again. Thanks to @TechLateef for providing several fixes and improvements.

TechLateef (changes requested):
nit: Since this PR touches a number of different areas, would it make sense to split it into a few logically grouped commits? I think it would make the review a bit easier to follow and align with the project's recommendation of keeping changes in small, reviewable commits.

allocz:
> nit: Since this PR touches a number of different areas, would it make sense to split it into a few logically grouped commits? I think it would make the review a bit easier to follow and align with the project's recommendation of keeping changes in small, reviewable commits.

Yes, it makes sense.

allocz:
@TechLateef, I've improved the commit history, as requested.

allocz:
I think that the current PR is doing too many things, causing difficulties to review the changes. So I'll close this PR and open smaller ones.

PR #2575, opened 2026-07-14
multi: relax trailing byte handling for DB blocks and PSBT readers

Roasbeef:
In this PR, we address a few follow-ups to the trailing byte hardening that landed in #2558.

## Tolerate trailing bytes in the node's own database

Databases written by older versions of btcd may have persisted blocks with trailing bytes: the old `submitblock` path and the pre-hardening p2p path both accepted them, and ffldb stores the raw bytes verbatim. With the strict loading added in #2558, such a block at the chain tip would prevent the node from ever starting, and a block deeper in the chain would fail `getblock`, reorg handling, and indexer catch-up, with no recovery path short of a full resync.

We now load database blocks via a new `blockchain.DBBlockFromBytes` helper (used by `initChainState` and `dbFetchBlockByNode`) that parses leniently: any trailing bytes are logged, ignored, and stripped from the serialization cached on the returned block, so downstream consumers of the raw bytes never observe them. The external ingress paths (RPC, p2p, PSBT) remain strict, this only relaxes reads from the node's own database.

The final commit extends the same treatment to the two remaining call sites that re-parse own-DB blocks strictly: the `getblock` RPC (which also served the dirty bytes verbatim at verbosity 0, and counted them in `Size`) and the index manager's init path, which loads orphaned blocks directly from the database when rolling an index tip back to the main chain. Both now go through the shared helper, so `getblock` serves the exact block serialization at every verbosity level. We can't route `getblock` through `Chain.BlockByHash` here since that rejects side chain blocks, which `getblock` serves today.

## Fix blocking reads and unbounded allocations in PSBT parsing

The trailing data check in `NewFromRawBytes` probed the caller supplied reader with a blocking one byte read, so a reader without a `Len` method (a `net.Conn`, an `io.Pipe`) that stays open after delivering a complete packet would hang the parser forever. We now only enforce the check when the reader can report its remaining length without an additional read, which covers in-memory readers along with the decoded base64 path, and we document the reader contract on `NewFromRawBytes`.

The base64 path also read the entire input into memory before any validation ran, so a very large input could force an arbitrarily large allocation before the first validity check. We now bound the read to `wire.MaxMessagePayload` expanded by the base64 encoding overhead.

## Steps to Test

Each fix ships with a regression test: startup with a trailing byte tip block now succeeds (with the trailing byte excluded from the recorded block size), `DBBlockFromBytes` strips trailing bytes from the cached serialization while still rejecting truncated blocks, a stream reader is never probed past the packet, and oversized base64 input is rejected (pinned with an erroring sentinel reader so the size bound itself is exercised). Run `go test ./blockchain/ ./blockchain/indexers/` in the root module and `go test ./...` in the psbt module.

### DIFF
--- a/blockchain/chainio.go
+++ b/blockchain/chainio.go
@@ -1277,7 +1277,7 @@ func (b *BlockChain) initChainState() error {
 		if err != nil {
 			return err
 		}
-		block, err := btcutil.NewBlockFromBytes(blockBytes)
+		block, err := DBBlockFromBytes(blockBytes, state.hash)
 		if err != nil {
 			return err
 		}
@@ -1301,8 +1301,15 @@ func (b *BlockChain) initChainState() error {
 			}
 		}
 
-		// Initialize the state related to the best block.
-		blockSize := uint64(len(blockBytes))
+		// Initialize the state related to the best block.  The block
+		// bytes are re-derived from the block itself so any trailing
+		// bytes ignored during deserialization are excluded from the
+		// recorded size.
+		serializedBlock, err := block.Bytes()
+		if err != nil {
+			return err
+		}
+		blockSize := uint64(len(serializedBlock))
 		blockWeight := uint64(GetBlockWeight(block))
 		numTxns := uint64(len(block.MsgBlock().Transactions))
 		b.stateSnapshot = newBestState(tip, blockSize, blockWeight,
@@ -1367,6 +1374,35 @@ func dbFetchHeaderByHeight(dbTx database.Tx, height int32) (*wire.BlockHeader, e
 	return dbFetchHeaderByHash(dbTx, hash)
 }
 
+// DBBlockFromBytes deserializes a block fetched from the local database,
+// tolerating trailing bytes rather than rejecting them outright.  Databases
+// written by older btcd versions may have persisted blocks with trailing
+// bytes, and failing here would make such blocks permanently unreadable.
+// Instead, any trailing bytes are logged, ignored, and excluded from the
+// serialization cached in the returned block.
+//
+// This lenient parsing is only appropriate for blocks read back from the
+// node's own database.  Blocks from external sources (p2p, RPC) should be
+// parsed with the strict btcutil.NewBlockFromBytes instead.
+func DBBlockFromBytes(blockBytes []byte, hash chainhash.Hash) (*btcutil.Block,
+	error) {
+
+	blockReader := bytes.NewReader(blockBytes)
+	var msgBlock wire.MsgBlock
+	if err := msgBlock.Deserialize(blockReader); err != nil {
+		return nil, err
+	}
+	if trailing := blockReader.Len(); trailing > 0 {
+		log.Warnf("Block %v has %d trailing bytes in the database; "+
+			"ignoring them", hash, trailing)
+		blockBytes = blockBytes[:len(blockBytes)-trailing]
+	}
+
+	// Cache the exact serialization on the block so downstream consumers
+	// of the raw bytes never observe the trailing bytes.
+	return btcutil.NewBlockFromBlockAndBytes(&msgBlock, blockBytes), nil
+}
+
 // dbFetchBlockByNode uses an existing database transaction to retrieve the
 // raw block for the provided node, deserialize it, and return a btcutil.Block
 // with the height set.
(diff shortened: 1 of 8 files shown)


Lrifton92 (approved):
Went through all three commits. The split makes sense: strict at ingestion (p2p/RPC/PSBT), lenient only when re-reading the node's own already-validated database — this composes cleanly with the trailing-byte rejection work on the external-facing paths, where strictness is the right default. Nice touch that `DBBlockFromBytes` strips the trailing bytes from the cached serialization via `NewBlockFromBlockAndBytes`, so `getblock` verbosity=0 and `Size` now serve the exact serialization instead of the dirty DB bytes. The PSBT stream fix is correct — the base64 path re-wraps in a `bytes.Reader`, so the `Len()` check still covers it, and the oversized-input test pinning the bound with an erroring sentinel is a good pattern.

One question: `DBBlockFromBytes` logs at Warn on every fetch — a single dirty block deep in the chain could get hammered by indexer catch-up or p2p serving and flood the logs. Worth demoting to Debug or warn-once per hash?

blockchain/chainio.go
```diff
@@ -1277,7 +1277,7 @@ func (b *BlockChain) initChainState() error {
 		if err != nil {
 			return err
 		}
-		block, err := btcutil.NewBlockFromBytes(blockBytes)
+		block, err := DBBlockFromBytes(blockBytes, state.hash)
```
allocz, line 1309: Wouldn't be better to put the `DBBlockFromBytes` definition before its usage?

blockchain/chainio.go
```diff
@@ -1367,6 +1374,35 @@ func dbFetchHeaderByHeight(dbTx database.Tx, height int32) (*wire.BlockHeader, e
...
+func DBBlockFromBytes(blockBytes []byte, hash chainhash.Hash) (*btcutil.Block,
+	error) {
+
+	blockReader := bytes.NewReader(blockBytes)
+	var msgBlock wire.MsgBlock
+	if err := msgBlock.Deserialize(blockReader); err != nil {
+		return nil, err
+	}
+	if trailing := blockReader.Len(); trailing > 0 {
+		log.Warnf("Block %v has %d trailing bytes in the database; "+
```
allocz: I'm not sure if this log is a good idea, I would not log this, but changing the level from `Warn` to `Debug` may be good enough.

allocz (changes requested):
Reverting the changes inside previously existing procedures causes the tests `TestInitChainStateToleratesTrailingBestBlockBytes` and `TestStreamReaderNotProbedPastPacket` to fail, as expected.

blockchain/chainio.go
Roasbeef: Will move it to warn instead

blockchain/chainio.go
Roasbeef: Will do

allocz (approved):
LGTM. Thanks for the fix!

psbt/psbt.go
```diff
@@ -324,34 +331,50 @@ func NewFromRawBytes(r io.Reader, b64 bool) (*Packet, error) {
...
 
 	return &newPsbt, nil
 }
 
+// maxBase64PsbtSize is the maximum number of base64 characters accepted when
+// decoding a PSBT. It is wire.MaxMessagePayload, the largest payload the
+// wire protocol will carry, expanded by the 4/3 base64 encoding overhead. It
+// bounds the memory allocated for a caller-supplied reader before any
+// validation runs.
+const maxBase64PsbtSize = 4 * ((wire.MaxMessagePayload + 2) / 3)
```
yyforyongyu: Using wire.MaxMessagePayload as the aggregate cap for base64 PSBT input rejects valid packets that the raw binary path still accepts. BIP-174 does not bind PSBT packet size to the p2p message payload limit, and the binary path has no equivalent total-size cap. I reproduced this with a packet containing nine individually valid 4,000,000-byte unknown values: NewFromRawBytes(..., false) succeeds, while the identical base64 packet returns ErrInvalidPsbtFormat here. Could we either decode base64 incrementally, or define a documented PSBT-specific aggregate limit and apply it consistently to both raw and base64 inputs?

yyforyongyu (approved):
LGTM🛡️ Left one unblocking comment, think it's a rare case.

psbt/psbt.go
Roasbeef: Hmm, that's a good point. Pushed a commit to unify the parsing across raw and base64 packets.

PR #2576, opened 2026-07-18
multi: tighten inbound connection admission

starius:
## Change Description
Inbound connection admission was tightened across the connection manager, peer lifecycle, server, and v2 transport handshake.

At the listener boundary, accepted inbound sockets now hold a bounded slot for their full lifetime. The server derives the inbound budget from the configured peer mode: connect-only and simnet nodes reserve their permanent peers, while automatic mode reserves its outbound target and configured addpeers, capped by the total peer limit.

The server applies a second layer of per-source accounting while inbound handshakes are incomplete. Loopback and whitelisted peers remain subject to these source limits, while whitelisted peers retain their existing no-ban behavior. Source leases are released after verack or disconnect, and peer registration and removal are ordered through a single lifecycle path.

The source and CPU admission policy lives in `internal/inbound`, which keeps normalization, token accounting, concurrency control, and their tests out of `package main`. The package follows btcd's existing `UseLogger` pattern, while `server.go` binds connection and handshake lifetimes to the policy. Global, per-source, and concurrent budgets remain independent.

For v2 responders, transport classification is separated from the CPU-bound key setup. After the first byte that diverges from the v1 prefix, the responder acquires a short-lived lease for key generation, releases it, and immediately sends its key so the handshake can make progress. It then reads the rest of the initiator key and acquires a second short-lived lease for key agreement and cipher setup. A handshake consumes its global and per-source rate budgets only once, while each CPU-bound phase acquires its own concurrency slot. Neither lease spans network I/O, and v1 fallback does not enter the v2 admission path.

Peer connection association is serialized with disconnection so a connection delivered after shutdown is closed instead of being published or leaked. The transport refactor also propagates handshake write errors, detects short writes, and removes the duplicate decoy write while preserving the existing wire transcript.

_Note_. The root module intentionally selects the modified nested v2 transport through its local replacement. A follow-up will publish and tag the v2 transport module, then update the root requirement before this series is merge-ready for downstream module consumers.

## Steps to Test

- `go test ./... -count=1`
- `go test -race ./connmgr ./peer . -count=1`
- `go test -race ./internal/inbound . -count=1`
- `cd v2transport && go test -race ./... -count=1`
- `go vet ./...`
- `cd v2transport && go vet ./...`
- `go test -tags=rpctest ./integration -run TestSyncManagerRaceCorruption -count=1`
- Manually tested in regtest: verification for v1/v2 coexistence, admission rejection, and recovery.

### Testing


Roasbeef (changes requested):
Overall the approach looks right to me. Moving the admission boundary down to
the accepted socket lifetime, keeping the expensive v2 work behind a narrow
interface, and serializing late connection association are all the right
pieces.

I did a full pass over the wire ordering, admission accounting, connection
lifecycle, config modes, and the tagged sync-manager stress test. There are a
few things we should address before this lands:

1. The responder now waits for the rest of the initiator's 64-byte ElligatorSwift
   key after the first v1-prefix mismatch, before it writes its own key. A
   sliced initiator is allowed to wait for responder progress before sending
   the remainder, so the two sides can deadlock here. Once the mismatch tells
   us this is a v2 candidate, we should generate and send the responder
   key+garbage, release admission across the network I/O, then reacquire a CPU
   slot for ECDH.

2. Loopback and whitelisted peers currently bypass the per-source pending and
   v2 rate limits. The whitelist option is a no-ban policy, not a resource
   reservation. In particular, forwarded local traffic can otherwise consume
   the global budget without paying the source budget. We should keep the
   existing whitelist bit for banning while applying ordinary resource
   admission to both classes.

3. The `MaxPeers` reservation infers automatic outbound mode from the peer
   lists and uses `max(targetOutbound, addPeers)`. That reserves nonexistent
   automatic peers on SimNet, and it under-reserves the opposite case when
   `ConnManager.Start` wins the race against the permanent `Connect`
   goroutines. The listener bound needs the actual automatic-outbound mode;
   when it is enabled, the conservative bound is `targetOutbound + addPeers`,
   capped by `MaxPeers`.

4. Raising `--maxpeers` in `TestSyncManagerRaceCorruption` isn't enough once
   the source admission is active. All 300 clients come from one loopback
   `/24`, so most are rejected while their handshakes are incomplete. We
   should stage at most eight incomplete handshakes, retain all 300 registered
   peers, then release one shared disconnect wave. The failure path also needs
   to cancel queued workers, drain every result, and bound its RPC count waits.

The good news is that these are localized follow-ups. The connmgr socket-slot
accounting and the late `AssociateConnection`/`Disconnect` fix held up under
the race tests.

Roasbeef:
I have fixes for the review items staged as fixup commits against the existing
transport, server-admission, and peer-lifecycle commits. I'll push those
directly to `starius:v2transport` now so CI can evaluate the exact amended
series. I won't force-push or autosquash in this step, so each follow-up stays
visible on the PR.

Once that is green, we can decide whether to autosquash the fixups before the
merge. There's also one release-ordering item outside the correctness review:
we need to tag the nested `v2transport` module and update the root requirement
before cutting the RC, since downstream modules don't inherit the local
`replace` directive.

Roasbeef:
Follow-up: I’ve consolidated the review follow-ups into three normal commits on top of the contributor’s original head:

* `server+inbound: correct inbound admission accounting`
* `v2transport: restore responder handshake progress`
* `integration: synchronize p2p lifecycle stress batches`

The resulting tree is byte-for-byte identical to the reviewed head that passed all four CI jobs. I updated the branch with `--force-with-lease`, so the original authorship/history remains intact and no intervening push was overwritten. CI is now rerunning on the consolidated head.

Roasbeef:
Pushed up a bug fix for one of my bug fixes 😅. 

With those commit landed, I think this is g2g. Will do some additional active/manual testing during the rc.

Roasbeef (approved):
LGTM 🧑‍🌾

PR #2580, opened 2026-07-24
build: bump version to v0.26.2-beta

Roasbeef:
In this PR, we move btcd past the v0.26.1 release candidate and set the
development version to v0.26.2-beta. This increments the patch series while
dropping the RC suffix, so the daemon and RPC server report the next
in-development version as work resumes on master.

## Testing

```text
GOWORK=off GOCACHE=/tmp/btcd-v0262-go-build go test .
```

A locally built binary reports `0.26.2-beta`.


saubyk (approved):
ack

Jason-123-cyber (commented):
Ok

Jason-123-cyber (commented):
Ok

PR #2594, opened 2026-08-19
rpctest,integration: options struct instead of multiple args

allocz:
## Change Description

In the context of the issue #2560 and pull request #2571, the current patch
changes the RPC Harness API to use options structs instead of several parameters
being passed to `rpctest.New()` and `Harness.SetUp()`. Options struct was also
added to `Harness.TearDown()`.

The changes make the API cleaner when the default config is desired, e.g.
`rpctest.New(nil)` instead of
`rpctest.New(&chaincfg.SimNetParams, nil, nil, "")`.

Also, by using options structs we can extend functionality by adding new fields
without breaking the tests.

Improving testability creates incentives to increase the amount of tests, which
allows catching regressions.

## Steps to Test
run `make check`

### Testing

### DIFF
--- a/integration/p2a_test.go
+++ b/integration/p2a_test.go
@@ -29,10 +29,9 @@ func TestPayToAnchorSimple(t *testing.T) {
 	// default, but the sub-dust and non-empty-witness cases below rely on
 	// standardness checks running, so we start the node with
 	// --rejectnonstd.
-	btcdCfg := []string{"--rejectnonstd"}
-	harness, err := rpctest.New(
-		&chaincfg.SimNetParams, nil, btcdCfg, "",
-	)
+	harness, err := rpctest.New(&rpctest.HarnessOpts{
+		ExtraArgs: []string{"--rejectnonstd"},
+	})
 	if err != nil {
 		t.Fatalf("unable to create test harness: %v", err)
 	}
@@ -40,7 +39,10 @@ func TestPayToAnchorSimple(t *testing.T) {
 
 	// Initialize the test harness with mining enabled to confirm
 	// transactions.
-	err = harness.SetUp(true, 25)
+	err = harness.SetUp(&rpctest.SetUpOpts{
+		CreateTestChain:  true,
+		NumMatureOutputs: 25,
+	})
 	if err != nil {
 		t.Fatalf("unable to setup test harness: %v", err)
 	}
@@ -197,4 +199,3 @@ func TestPayToAnchorSimple(t *testing.T) {
 		}
 	})
 }
-
--- a/integration/rpctest/rpc_harness.go
+++ b/integration/rpctest/rpc_harness.go
@@ -127,30 +127,67 @@ type Harness struct {
 	sync.Mutex
 }
 
-// New creates and initializes new instance of the rpc test harness.
-// Optionally, websocket handlers and a specified configuration may be passed.
-// In the case that a nil config is passed, a default configuration will be
-// used. If a custom btcd executable is specified, it will be used to start the
-// harness node. Otherwise a new binary is built on demand.
+// HarnessOpts are options that can be passed to New when initializing the
+// harness instance.
+type HarnessOpts struct {
+	// ActiveNet are the parameters of the network to be used, if nil,
+	// SimNet will be used.
+	ActiveNet *chaincfg.Params
+
+	// Handlers are the RPC client notification handlers than can optionally
+	// be passed in.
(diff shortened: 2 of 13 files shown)


TechLateef (changes requested):
Could we avoid introducing temporary `New2`, `SetUp2`, and `TearDown2` functions and instead update the existing functions and their callers in the same commit? This would keep each commit clean, self-contained, and easier to review without requiring follow-up commits to rename or fix the previous ones.

integration/rpctest/rpc_harness.go
```diff
@@ -127,30 +127,67 @@ type Harness struct {
...
-// Optionally, websocket handlers and a specified configuration may be passed.
-// In the case that a nil config is passed, a default configuration will be
-// used. If a custom btcd executable is specified, it will be used to start the
-// harness node. Otherwise a new binary is built on demand.
+// HarnessOpts are options that can be passed to New when initializing the
+// harness instance.
+type HarnessOpts struct {
+	// Params are the parameters of the network to be used, if nil, SimNet
+	// will be used.
+	Params *chaincfg.Params
```
TechLateef: Would it make sense to rename `Params` to `ActiveNet` (or `NetParams`)?

`Params` feels a bit generic at call sites (e.g. `HarnessOpts{Params: ...}`), whereas `ActiveNet` matches the `h.ActiveNet` field on the `Harness` struct itself and makes it immediately clear that it refers to the chain network parameters.

integration/p2a_test.go
```diff
@@ -29,18 +29,20 @@ func TestPayToAnchorSimple(t *testing.T) {
 	// default, but the sub-dust and non-empty-witness cases below rely on
 	// standardness checks running, so we start the node with
 	// --rejectnonstd.
-	btcdCfg := []string{"--rejectnonstd"}
-	harness, err := rpctest.New(
-		&chaincfg.SimNetParams, nil, btcdCfg, "",
-	)
+	opts := &rpctest.HarnessOpts{ExtraArgs: []string{"--rejectnonstd"}}
```
TechLateef: **nit:** In places like `p2a_test.go`, `prune_test.go`, and `bip0009_test.go`, we create a single-use `opts` variable just to pass it to `rpctest.New(...)` on the next line. We could pass `&rpctest.HarnessOpts{...}` inline directly to keep the setup concise and avoid declaring an unnecessary local variable:
```go
 r, err := rpctest.New(&rpctest.HarnessOpts{
     ExtraArgs: []string{"--prune=1536"},
 })
 
 ```

File: integration/rpctest/rpc_harness.go [hunk 3]
```diff
@@ -398,30 +398,21 @@ func (h *Harness) tearDown() error {
 	return nil
 }
 
-// TearDownOpts are options that can be passed to TearDown2.
+// TearDownOpts are options that can be passed to TearDown.
```
TechLateef: `TearDownOpts` currently has no fields, and `opts` is unused inside `TearDown`. Because of this, callers across the test suite are forced to pass `nil` (`r.TearDown(nil)`).

Could we keep `TearDown()` parameterless (`func (h *Harness) TearDown() error`) until there are actual options to support?

allocz: The idea is that a future PR will pass options to `TearDown`, to be able to change `TearDown` behavior, like skipping cleanup, which is necessary to be able to restart the node, and also waiting for the node process to stop without sending signal, which is necessary to be able to test scenarios where the node stops by reasons other than receiving interrupt signals.

To avoid changing the `TearDown` calls in the future, I chose to change the signature on this PR. But if you still think that is better to not change it now, I can revert this.

As I've written in #2571, I chose to break down the PR into several ones, to make the review easier by reducing the size of the diffs.

integration/rpctest/rpc_harness.go
allocz: Ok, I will keep ActiveNet.

integration/p2a_test.go
allocz: Suggestion applied.

integration/rpctest/rpc_harness.go
allocz: Anyway, in the end, I agree with you. Suggestion applied.

TechLateef:
LGTM! Just needs a rebase.

allocz:
> LGTM! Just needs a rebase.

Rebased.

TechLateef (approved):

PR #2596, opened 2026-08-21
btcctl: preserve semantic version separators

pucedoteth:
65db493 ("version: preserve semantic version separators", #2578) added periods to `semanticAlphabet` in the root `version.go`, because SemVer separates pre-release identifiers with periods and the old filter silently collapsed `beta.rc1` into `betarc1`.

`cmd/btcctl` carries its own near-identical copy of that file, and it was not updated. The two now disagree:

```
version.go:14            ...uvwxyz-."
cmd/btcctl/version.go:14 ...uvwxyz-"
```

So `btcctl` still drops the separator. The upstream test, run unchanged against the btcctl copy, fails:

```
--- FAIL: TestNormalizeVerString
    version_test.go:16: unexpected normalized version: got "betarc1", want "beta.rc1"
```

This is reachable rather than latent. `normalizeVerString` is applied to `appBuild`, which the comment above it documents as settable at build time:

```go
// appBuild is defined as a variable so it can be overridden during the build
// process with '-ldflags "-X main.appBuild foo' if needed.
```

SemVer build metadata is dot-separated (`exp.sha.5114f85` is an example from the spec itself), so a btcctl built with `-ldflags "-X main.appBuild=exp.sha.5114f85"` reports `0.23.1-beta+expsha5114f85` — the dots stripped and the metadata no longer round-trips. `btcd` built the same way reports it correctly.

The change is the one-character fix from 65db493 applied to the btcctl copy, plus that commit's test mirrored into `cmd/btcctl`, with a build metadata case added for the `appBuild` path.

`go build ./...`, `go test ./cmd/...`, `gofmt` and `go vet` are all clean.

I left the stale `appMinor`/`appPatch` in the btcctl copy alone (0.23.1 there vs 0.26.2 at the root) since bumping those is a release decision, not part of this fix.

---

*Disclosure: written with AI assistance; the analysis, fix and test were verified locally.*


pucedoteth:
Pushed a test-only follow-up.

The original table asserted `normalizeVerString(x) == x` for every case, which an identity function satisfies — so it could not have failed if the filter were dropped. (Thanks to @Lrifton92 for spotting this on the equivalent btcwallet PR.) The table is now in/want-based with a case whose input contains a character outside the semantic alphabet.

Verified by neutralising the alphabet check in place (`if true || strings.ContainsRune(...)`, which still compiles): the new case fails with got `"beta!rc1"`, want `"betarc1"`. The dotted cases still fail in the other direction if `.` is removed from the alphabet, so the test now pins both ends.

Roasbeef (approved):
LGTM

pucedoteth:
Rebased onto master (`9b323a7c`), which clears the coveralls drift warning above — its "85 coverage regressions" were in `rpcclient/infrastructure.go`, `btcec/v2/ellswift`, `peer/peer.go` and `wire/v2/message.go`, none of which this PR touches. The same report says `Patch coverage: No coverable lines changed in this PR`, which is expected: the change is a one-character const edit plus a new test file.

On the red `Unit rpctest`: it is not from this PR. The failure is

```
--- FAIL: TestSyncManagerRaceCorruption
    sync_race_test.go:349: "0" is not greater than or equal to "5"
    sync manager corruption after 1200 fake peer cycles: node stuck with dead sync peer (height 0 -> 0)
```

**master fails the same test with the same message.** Run `89c50d76` (2026-09-09T00:31, push to master) failed `Unit rpctest` on `TestSyncManagerRaceCorruption`, and `22d98417` two minutes later passed. My run was in that same window.

It also cannot be reached from this diff: the only production change is adding `.` to `semanticAlphabet` in `cmd/btcctl/version.go`, and nothing in `integration/` or the sync manager imports `cmd/btcctl`.

Worth flagging separately though — that test is named for the condition it is asserting, and "node stuck with dead sync peer" reproducing on master twice in one window looks more like an intermittent sync-manager issue than ordinary CI noise. I have not investigated it; it just is not this PR.

Roasbeef:
@pucedoteth I saw that failure after the latest merge myself, looking into it. Approved the CI run on this PR.

Roasbeef:
Can you just rebase on top of master?

pucedoteth:
Done — rebased onto `c5f2e2e` and force-pushed (`daf274f`). The merge commit is gone; it is now the two original commits linear on top of master.

Content is unchanged from what you approved — `git diff` between the pre-rebase tip and the rebased head over `cmd/btcctl` is empty, and `go test ./cmd/btcctl/... -run TestNormalizeVerString` passes.

pucedoteth:
Rebased on master (now `f29679a`). `go test ./cmd/btcctl/...` passes.

pucedoteth:
@Roasbeef the rebase you asked for landed on the 16th — `f29679a`, the two commits linear on master, content identical to what you approved on the 9th.

Two things stand in the way now, both mechanical rather than anything in the diff:

- the force-push cleared your approval, so this reads as `REVIEW_REQUIRED` again
- CI has not run on the new head — the PR currently shows no checks at all, which I think needs the same approve-and-run you did last time

`go test ./cmd/btcctl/...` passes locally on that head. Happy to rebase again if master has moved since.

PR #2600, opened 2026-09-08
btcd: fix getblock panic triggered by rpc call

### LINKED ISSUE #2597: [bug]: rpcserver getblock panics on explicit JSON null verbosity (nil pointer dereference)
### Background

`handleGetBlock` dereferences the optional `Verbosity *int` parameter without a nil check on one of its branches, so a `getblock` request that supplies verbosity as an explicit JSON `null` panics the handler with a nil pointer dereference instead of returning a JSON-RPC error.

`GetBlockCmd.Verbosity` defaults to `1`, but `btcjson` applies a `jsonrpcdefault` only when the parameter is **omitted** (`btcjson/cmdparse.go:157`, `if numParams < info.maxParams`). When the client sends both parameters and the second is JSON `null`, defaulting is skipped and `json.Unmarshal` of `null` into a pointer leaves it `nil`.

The handler guards the first read but not the second:

- `rpcserver.go:1119` — `if c.Verbosity != nil && *c.Verbosity == 0` (guarded)
- `rpcserver.go:1165` — `if *c.Verbosity == 1` (**unguarded dereference**)

`getblock` is in the `rpcLimited` allowlist (`rpcserver.go:271`), so any RPC credential holder, including a limited user, reaches it. On the HTTP path `net/http` recovers the panic per connection (the request dies with a dropped connection and a logged stacktrace); on the WebSocket path the request runs in its own goroutine with no recover (`rpcwebsocket.go` `serviceRequest`), so the panic kills that goroutine and leaks a concurrency slot.

This requires a valid RPC credential, so there is no unauthenticated-attacker angle — filing it as an ordinary robustness bug rather than through the security contact. Found during a code review.

### Your environment

* `btcd` v0.26.2, and present on `master` (`rpcserver.go:1119`/`:1165` unchanged)
* platform-independent

### Steps to reproduce

Against a running node with RPC enabled, any valid credential (a limited user suffices):

```bash
curl --user <user>:<pass> --data-binary \
  '{"jsonrpc":"1.0","id":"x","method":"getblock","params":["<valid block hash hex>",null]}' \
  http://127.0.0.1:<rpcport>/
```

The same nil dereference, with no network, as a unit-style check:

```go
package main

import (
	"encoding/json"

)

func main() {
	req := &btcjson.Request{
		Method: "getblock",
		Params: []json.RawMessage{
			json.RawMessage(`"0000000000000000000000000000000000000000000000000000000000000000"`),
			json.RawMessage(`null`),
		},
	}
	cmd, _ := btcjson.UnmarshalCmd(req) // no error: both params present, so the default is not applied
	gb := cmd.(*btcjson.GetBlockCmd)    // gb.Verbosity == nil
	_ = *gb.Verbosity                   // == rpcserver.go:1165 -> panic
}
```

### Expected behaviour

An explicit `null` verbosity is either treated as the default (`1`) or rejected with a JSON-RPC error. No panic.

### Actual behaviour

`runtime error: invalid memory address or nil pointer dereference` at `rpcserver.go:1165`; on the WebSocket path the serving goroutine dies and leaks a concurrency slot.

### Suggested fix

Guard the dereference, e.g. `if c.Verbosity != nil && *c.Verbosity == 1 { ... }`, or apply the default even when the parameter is supplied as JSON `null`. As defense in depth, `recover()` in the per-request WebSocket `serviceRequest` goroutine to avoid the slot leak.

allocz:
## Change Description
This patch Fixes #2597

As stated on issue #2597, a JSONRPC request containing `null` as value for `GetBlockCmd.Verbosity` causes a panic in the `handleGetBlock` procedure.

This commit fixes it by setting the default value of `1` when `Verbosity` is `nil` in the start of the `handleGetBlock` procedure.

## Steps to Test

1. Start the Bitcoin node
```
$ go build .
$ ./btcd --regtest --rpcuser=bitcoin --rpcpass=bitcoin --rpclisten=127.0.0.1:8332
```

2. Send the payload
```
$ curl -k --user bitcoin:bitcoin -d   '{"jsonrpc":"2.0","id":0,"method":"getblock","params":["0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206",null]}'  'https://127.0.0.1:8332'

{"jsonrpc":"2.0","result":{"hash":"0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206","confirmations":1,"strippedsize":285,"size":285,"weight":1140,"height":0,"version":1,"versionHex":"00000001","merkleroot":"4a5e1e4baab89f3a32518a88c31bc87f618f76673e2cc77ab2127b7afdeda33b","tx":["4a5e1e4baab89f3a32518a88c31bc87f618f76673e2cc77ab2127b7afdeda33b"],"time":1296688602,"nonce":2,"bits":"207fffff","difficulty":1,"previousblockhash":"0000000000000000000000000000000000000000000000000000000000000000"},"error":null,"id":0}
```

Before the patch, step `2` triggers panic and expected result is not returned.

### Testing

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -1074,6 +1074,10 @@ func getDifficultyRatio(bits uint32, params *chaincfg.Params) float64 {
 // handleGetBlock implements the getblock command.
 func handleGetBlock(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (interface{}, error) {
 	c := cmd.(*btcjson.GetBlockCmd)
+	if c.Verbosity == nil {
+		c.Verbosity = new(int)
+		*c.Verbosity = 1
+	}
 
 	// Load the raw block bytes from the database.
 	hash, err := chainhash.NewHashFromStr(c.Hash)
@@ -1116,7 +1120,7 @@ func handleGetBlock(s *rpcServer, cmd interface{}, closeChan <-chan struct{}) (i
 	}
 
 	// If verbosity is 0, return the serialized block as a hex encoded string.
-	if c.Verbosity != nil && *c.Verbosity == 0 {
+	if *c.Verbosity == 0 {
 		return hex.EncodeToString(blkBytes), nil
 	}
 


allocz:
CC @gijswijs @Roasbeef

Roasbeef:
Approved CI.

PR #2603, opened 2026-09-09
build: enforce format and lint checks in CI

### LINKED ISSUE #2549: [feature]: add linter on CI
**Is your feature request related to a problem? Please describe.**

The current Makefile already provides lint and fmt targets, but they are not enforced in CI.

**Describe the solution you'd like**

Add fmt and lint checks to CI as required steps. This will probably also require updating the tool packages used by the workflow.

**Describe alternatives you've considered**

No alternatives considered.

**Additional context**

The tools in the btcwallet have a good example of how this could be implemented:
https://github.com/btcsuite/btcwallet/blob/sql-wallet/tools/Dockerfile

vbrekher:
## Change Description

Run the existing `fmt` and `lint` Makefile targets in CI.

The old golangci-lint baseline is no longer in the current master history, so this also moves `new-from-rev` to the current master and gives the lint job full git history.

Closes #2549

## Steps to Test

`make fmt && git diff --exit-code`

`make lint`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,6 +9,28 @@ env:
   GO_VERSION: 1.26.3
 
 jobs:
+  lint:
+    name: Format and lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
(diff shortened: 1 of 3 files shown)


GustavoStingelin (approved):
Thanks for the PR! Think it is a good starting point.

PR #2604, opened 2026-09-09
rpcserver: omit witness commitment from coinbasetxn

### LINKED ISSUE #2572: [bug]: mining: getblocktemplate sends witness commitment inside coinbasetxn, violating BIP 145
### Background

BIP 145 explicitly states:

> "Servers MUST NOT include a commitment in the `coinbasetxn` key on the template. Clients MUST insert the commitment as an additional output at the end of the final generation (coinbase) transaction."

btcd currently violates this requirement. In `mining/mining.go` line 808, `AddWitnessCommitment` appends the witness commitment directly to the coinbase transaction. In `rpcserver.go` line 1814, `blockTemplateResult` serializes `msgBlock.Transactions[0]`  the same coinbase  into `CoinbaseTxn`, so the commitment is included in the response.

I noticed this while working on PR #2569 which enforces BIP 145 segwit rule requirements in `getblocktemplate` requests.

---

### Your environment

* Latest master branch of btcd
* Verified on Linux

---

### Steps to reproduce

1. Start btcd on a network where segwit is active
2. Call `getblocktemplate` with `coinbasetxn` capability and `segwit` in rules
3. Inspect the returned `coinbasetxn` field  it contains an OP_RETURN output with the witness commitment already included

Relevant code:
- `mining/mining.go` line 808 — `AddWitnessCommitment` appends commitment to coinbase
- `rpcserver.go` line 1814 — coinbase serialized into `coinbasetxn` with commitment already present

---

### Expected behaviour

The `coinbasetxn` field should contain a clean coinbase transaction without the witness commitment. The client is responsible for calculating and appending the commitment as the final output per BIP 145.

---

### Actual behaviour

The `coinbasetxn` field contains the coinbase transaction with the witness commitment already appended as an OP_RETURN output. A strict BIP 145 compliant client would append a second commitment, producing an invalid block rejected by the network.

**Note:** fixing this is non-trivial because `CheckConnectBlockTemplate` requires the commitment to be present internally for validation. A fix would need to strip the commitment before serializing into `coinbasetxn` while keeping it in the internal template.

vbrekher:
## Change Description

BIP 145 requires `coinbasetxn` to omit the witness commitment so miners can append a commitment for their final transaction set.

Keep the commitment in btcd's internal block template for validation, but strip the final commitment output from the copy serialized into `coinbasetxn`.

Adds regression coverage to ensure the internal coinbase remains unchanged.

Fixes #2572.

## Steps to Test

`go test .`

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -1699,6 +1699,32 @@ func (state *gbtWorkState) updateBlockTemplate(s *rpcServer, useCoinbaseValue bo
 	return nil
 }
 
+// coinbaseTxForTemplate returns the coinbase transaction to expose through the
+// getblocktemplate coinbasetxn field. BIP 145 requires the witness commitment
+// to be omitted so the client can append the commitment for its final
+// transaction set.
+func coinbaseTxForTemplate(tx *wire.MsgTx, witnessCommitment []byte) *wire.MsgTx {
+	if len(witnessCommitment) == 0 || len(tx.TxOut) == 0 {
+		return tx
+	}
+
+	expectedScript := make(
+		[]byte, 0, len(blockchain.WitnessMagicBytes)+len(witnessCommitment),
+	)
+	expectedScript = append(expectedScript, blockchain.WitnessMagicBytes...)
+	expectedScript = append(expectedScript, witnessCommitment...)
+
+	lastOutput := len(tx.TxOut) - 1
+	if !bytes.Equal(tx.TxOut[lastOutput].PkScript, expectedScript) {
+		return tx
+	}
+
+	tx = tx.Copy()
+	tx.TxOut = tx.TxOut[:lastOutput]
+
+	return tx
+}
+
 // blockTemplateResult returns the current block template associated with the
 // state as a btcjson.GetBlockTemplateResult that is ready to be encoded to JSON
 // and returned to the caller.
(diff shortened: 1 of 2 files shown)


Lrifton92 (commented):
Checked at 4bf47d2d.

The stripping itself looks right: `AddWitnessCommitment` returns the 32-byte commitment and appends `WitnessMagicBytes || commitment` as the last output, so `coinbaseTxForTemplate` rebuilds exactly that script, copies the transaction and drops only that output. The internal template keeps its commitment for validation, and the `hash` reported for `coinbasetxn` is now computed from the stripped transaction, which is consistent with the `data` field. The coinbase input still carries the 32-byte witness reserved value, which the client needs anyway once it adds its own commitment.

Two points:

1. The call site is not covered. `TestCoinbaseTxForTemplate` exercises the helper only: reverting `blockTemplateResult` to `tx := msgBlock.Transactions[0]` still passes `go test . -run 'CoinbaseTxForTemplate|GetBlockTemplate'`. A test that goes through `blockTemplateResult` with a segwit template and `coinbasetxn` requested, and asserts that the decoded `CoinbaseTxn.Data` has no commitment output while `msgBlock` still has one, would pin the actual fix for #2572.

2. Not introduced here, but it matters more now that clients must add the commitment themselves: `default_witness_commitment` is set to `hex(template.WitnessCommitment)`, i.e. only the 32-byte hash (the help text says "the witness commitment itself"). Bitcoin Core returns the full output script there (`6a24aa21a9ed` followed by the hash), and pools written against Core append that value as the commitment output's `pkScript`. With btcd's value they would build an output without the `OP_RETURN` and magic prefix, and the block would fail the witness commitment check. Might be worth aligning it in this PR or a follow-up, since #2572 is about making `coinbasetxn` usable per BIP 145.

PR #2605, opened 2026-09-11
rpcserver: test null getblock verbosity

Roasbeef:
In this PR, we add the missing end-to-end regression test for #2597 and the fix merged in #2600. The test sends `getblock` through `RawRequest` with an explicit JSON `null` verbosity, then verifies that btcd applies the default verbose response rather than disconnecting the WebSocket client.

The same test fails on #2600 parent with `websocket: close 1006 unexpected EOF` and `the client has been disconnected`, and passes on current `master`.

## Testing

`make unit`

`golangci-lint run --timeout=5m --new-from-rev=a3f203843`

### DIFF
--- a/integration/rpcserver_test.go
+++ b/integration/rpcserver_test.go
@@ -10,6 +10,7 @@ package integration
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"os"
 	"runtime/debug"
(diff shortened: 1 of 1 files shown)


allocz (approved):
Confirmed that reverting #2600 causes the test to fail. LGTM.

PR #2607, opened 2026-09-11
netsync: recover after losing sync peer near tip

### LINKED ISSUE #2606: [bug]: node stops fetching blocks after losing its sync peer near tip; recovers only when a new peer connects
### Background

We run a mainnet btcd node behind a small public block explorer (https://btcd.watch). Twice in three days the node silently stopped following the chain: it stayed connected to all 8 outbound peers, logged nothing, and simply never fetched another block until we intervened by hand. Each time it sat 7–8 hours behind the network.

Both times, `getpeerinfo` showed **no peer with `syncnode: true`**, and both times the fix was to disconnect any one peer:

```
btcctl node disconnect <addr>:8333
```

Within a minute btcd connected a replacement peer, downloaded the missing headers, and resumed. No restart needed.

Timeline of the second incident (2026-09-11, UTC):

```
08:47:10 [INF] SYNC: Lost peer [2600:...]:8333 (outbound)
08:47:10 [INF] SYNC: New valid peer 81.107.181.46:8333 (outbound) (/Satoshi:31.1.0/)
08:47:16 [INF] SYNC: Lost peer 81.107.181.46:8333 (outbound)
08:49:16 [INF] SYNC: New valid peer 50.72.102.138:8333 (outbound) (/Satoshi:31.1.0/)
   ... nothing at all for 7h50m; getblockcount frozen at 966481, network reached 966527 ...
16:39:23 [INF] SYNC: Lost peer [2001:...]:8333 (outbound)            <- our manual "node disconnect"
16:39:24 [INF] SYNC: New valid peer 193.201.82.36:8333 (outbound) (/Satoshi:30.2.0/)
16:39:24 [INF] SYNC: Downloading headers for blocks 966482 to 966527 from peer 193.201.82.36:8333
16:39:24 [INF] SYNC: downloaded headers to ...(966527) from peer 193.201.82.36:8333 -- now fetching blocks
16:39:25 [INF] SYNC: Lost peer 193.201.82.36:8333 (outbound)
16:39:55 [INF] SYNC: New valid peer 209.126.10.240:8333 (outbound) (/Satoshi:29.0.0/)
16:39:55 [INF] SYNC: Syncing to block height 966527 from peer 209.126.10.240:8333
16:40:57 [INF] SYNC: Processed 2 blocks in the last 1m33.03s (height 966482, ...)
```

The first incident (2026-09-08, 17:35 UTC, ~7.5 h) looked the same: sync peer lost, "New valid peer", then silence, no `syncnode` peer, fixed by the same disconnect command. It happened on v0.26.0; we upgraded to v0.26.2 and it recurred.

What it looks like from the outside: once the node loses its sync peer while it is still a block or two behind, it stops reacting to the new-block announcements its peers keep sending, and it only wakes up when a *new* peer connects that advertises a higher height. Our host is slow on disk and is often a block or two behind, which probably makes us hit this more than most. Our own write-up with the code paths we believe are involved (in `netsync/manager.go`, around how `ibdMode` and `syncPeer` interact) is here: https://github.com/saubyk/btcd.watch/issues/37 — happy to be corrected on that part.

### Your environment

* btcd v0.26.2 (stock, built from tag with Go 1.26.4); previously v0.26.0, same behaviour
* Ubuntu Linux x86_64 VPS
* mainnet, `txindex=1`, `addrindex=1`, `nolisten=1` (outbound peers only, default 8), `utxocachemaxsize=450`
* slow virtual disk: blocks often take 30–90 s to process at tip

### Steps to reproduce

We can't trigger it on demand. Conditions both times: node at or within a couple of blocks of the tip, sync peer disconnects, no other peer is known to be ahead. From then on the node ignores new blocks until a fresh outbound connection is made. `btcctl node disconnect` on any peer reliably recovers it, which may help when trying to reproduce.

### Expected behaviour

After losing its sync peer, btcd should keep accepting the new-block announcements it receives from its remaining peers, or re-elect a sync peer on its own. At minimum it should log that it is stuck.

### Actual behaviour

btcd stays connected to all peers, logs nothing, reports no `syncnode`, and does not fetch new blocks for hours, until a new peer connection happens to be made. `getblockchaininfo` reports `headers == blocks` at the stale height, so nothing on the node itself indicates a problem.

Roasbeef:
In this PR, we clear ibdMode when updateSyncPeer removes the active sync peer. Near the chain tip, startSync can find no strictly higher replacement and return with syncPeer nil. Leaving IBD enabled in that state causes handleInvMsg to discard every new block announcement, so the node remains stalled until a newly connected peer advertises a higher height.

Resetting both pieces of sync state together lets the remaining peers announce and request blocks normally, while a replacement peer still re-enters IBD through startSync. The regression test covers the near-tip disconnect and verifies that the next block announcement is requested.

Fixes #2606

Tests: go test ./...

The repository-wide make lint target still reports 13 existing findings outside this diff, including pre-existing unchecked calls in netsync.

### DIFF
--- a/netsync/manager.go
+++ b/netsync/manager.go
@@ -556,6 +556,10 @@ func (sm *SyncManager) updateSyncPeer(dcSyncPeer bool) {
 	}
 
 	sm.syncPeer = nil
+	// IBD requires an active sync peer. Clear it with the peer so a lost
+	// sync peer cannot leave the manager in a state where announcements are
+	// ignored indefinitely.
+	sm.ibdMode = false
 	sm.startSync()
 }
 
--- a/netsync/manager_test.go
+++ b/netsync/manager_test.go
@@ -1221,6 +1221,70 @@ func TestStartSyncChainCurrent(t *testing.T) {
 		"ibdMode should not be activated when chain is already current")
 }
 
+// TestLostSyncPeerNearTip verifies that losing the sync peer near the tip
+// leaves announcements from the remaining peers eligible for download.
+func TestLostSyncPeerNearTip(t *testing.T) {
+	t.Parallel()
+
+	params := chaincfg.RegressionNetParams
+	params.Checkpoints = nil
+
+	sm, tearDown := makeMockSyncManager(t, &params)
+	defer tearDown()
+
+	coinbase1 := createTestCoinbase(1, &params)
+	header1 := wire.BlockHeader{
+		Version:    4,
+		PrevBlock:  *params.GenesisHash,
+		MerkleRoot: coinbase1.TxHash(),
+		Timestamp:  time.Now().Truncate(time.Second),
+		Bits:       params.PowLimitBits,
+	}
+	require.True(t, solveTestBlock(&header1, &params))
+	block1 := btcutil.NewBlock(&wire.MsgBlock{
+		Header:       header1,
+		Transactions: []*wire.MsgTx{coinbase1},
+	})
+	_, _, err := sm.chain.ProcessBlock(block1, blockchain.BFNone)
+	require.NoError(t, err)
+	require.True(t, sm.chain.IsCurrent())
+
+	coinbase2 := createTestCoinbase(2, &params)
+	header2 := wire.BlockHeader{
+		Version:    4,
+		PrevBlock:  *block1.Hash(),
+		MerkleRoot: coinbase2.TxHash(),
+		Timestamp:  header1.Timestamp.Add(time.Minute),
+		Bits:       params.PowLimitBits,
+	}
+	require.True(t, solveTestBlock(&header2, &params))
+	block2 := btcutil.NewBlock(&wire.MsgBlock{
+		Header:       header2,
+		Transactions: []*wire.MsgTx{coinbase2},
+	})
+
+	disconnectedPeer := newSyncCandidate(t, sm, 1)
+	remainingPeer := newSyncCandidate(t, sm, 1)
+	sm.syncPeer = disconnectedPeer
+	sm.ibdMode = true
+
+	sm.handleDonePeerMsg(disconnectedPeer)
+
+	require.Nil(t, sm.syncPeer)
+	require.False(t, sm.ibdMode)
+
+	blockHash := block2.Hash()
+	inv := wire.NewMsgInvSizeHint(1)
+	err = inv.AddInvVect(wire.NewInvVect(
+		wire.InvTypeWitnessBlock, blockHash,
+	))
+	require.NoError(t, err)
+	sm.handleInvMsg(&invMsg{peer: remainingPeer, inv: inv})
+
+	require.Contains(t, sm.requestedBlocks, *blockHash,
+		"the remaining peer's announcement should be requested")
+}
+
 // TestIsSyncCandidateRegtest verifies that isSyncCandidate accepts peers
 // on regtest and simnet based on their service flags.
 func TestIsSyncCandidateRegtest(t *testing.T) {
--- a/netsync/manager_test.go
+++ b/netsync/manager_test.go
@@ -22,6 +22,7 @@ import (
 	"github.com/btcsuite/btcd/txscript/v2"
 	"github.com/btcsuite/btcd/wire/v2"
 	"github.com/stretchr/testify/require"
+	"pgregory.net/rapid"
 )
 
 // The package-level log variable is nil by default. Set it to the
@@ -602,7 +603,11 @@ func solveTestBlock(header *wire.BlockHeader, params *chaincfg.Params) bool {
 }
 
 // generateTestBlocks creates count valid blocks chaining from the genesis
-// block of the given params.  Each block contains only a coinbase transaction.
+// block of the given params. Each block contains only a coinbase transaction.
+//
+// The timestamps start from the current time so the chain is IsCurrent at
+// any height, which keeps the near-tip sync paths reachable (a chain built
+// from the 2011 regtest genesis timestamp is never current).
 func generateTestBlocks(
 	t *testing.T, params *chaincfg.Params, count int) []*btcutil.Block {
 
@@ -610,7 +615,7 @@ func generateTestBlocks(
 
 	blocks := make([]*btcutil.Block, 0, count)
 	prevHash := params.GenesisHash
-	prevTime := params.GenesisBlock.Header.Timestamp
+	prevTime := time.Now().Truncate(time.Minute)
 
 	for h := int32(1); h <= int32(count); h++ {
 		cb := createTestCoinbase(h, params)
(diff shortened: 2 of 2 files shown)


saubyk:
cc: @kcalvinalvin for review

Roasbeef:
Two property-based tests added on top of the regression test (rapid is already a go.mod dependency, used in `internal/inbound`):

**TestPropertyLostSyncPeerRecovers** generalizes the near-tip regression across arbitrary chain heights and remaining-peer height combinations. For any generated scenario (chain height 1–31, 1–4 remaining peers at tip or tip+1), losing the sync peer with IBD enabled must end in one of exactly two live states: a strictly higher peer was promoted and the header-first path recovers (verified end-to-end: header delivered, block requested), or the sync state was fully reset and a new-block announcement from any remaining peer is requested. The dead state `(syncPeer=nil, ibdMode=true)` — the state in which `handleInvMsg` line ~1165 drops every announcement — is asserted unreachable.

**TestPropertySyncStateTransition** is a stateful property over the whole sync state machine: a generated schedule of events (peer connects at arbitrary heights, header/block delivery from the sync peer, stall detection, sync-peer loss) is replayed against the manager with the same invariant asserted after *every* event.

One supporting change was needed and is worth flagging for reviewers: `generateTestBlocks` previously anchored timestamps to the regtest genesis (2011), so a generated chain was never `IsCurrent()` and the near-tip code paths the bug lives in were unreachable at generated heights. Timestamps now start from the current time, which keeps those paths reachable — this is what lets the properties actually reproduce the bug: **both fail on the parent commit and pass with the fix** (verified by reverting `manager.go` locally and re-running).

With the fix reverted, the properties fail with exactly the dead-state message; with the fix, 100 generated cases pass per property (`-count=3` also clean, full `go test ./...` green).

On issue coverage: the issue describes a node 1–2 blocks behind tip whose sync peer departs while no other peer is strictly higher — that exact scenario is now generated by both properties. The one behavioral gap the issue mentions that this PR does *not* address is the "at minimum it should log that it is stuck" ask — the fix removes the stuck state rather than logging it, which resolves the incident class outright.

netsync/manager.go
```diff
@@ -556,6 +556,10 @@ func (sm *SyncManager) updateSyncPeer(dcSyncPeer bool) {
 	}
 
 	sm.syncPeer = nil
+	// IBD requires an active sync peer. Clear it with the peer so a lost
+	// sync peer cannot leave the manager in a state where announcements are
+	// ignored indefinitely.
+	sm.ibdMode = false
```
saubyk, line 562: This clears IBD mode unconditionally, including when the node is far from the tip and no higher-height peer is available to take over. In that window `syncPeer` is nil and `ibdMode` is false, so a headers message from *any* peer now falls through to `fetchHeaderBlocks` (~line 1014) with no check that the sender is the sync peer or even a sync candidate.

As I understand it, that can reserve up to 50k block hashes in `requestedBlocks` from a peer that never serves them, and nothing clears those entries while there's no sync peer. A later honest sync peer then skips every reserved hash and stalls out. Pre-PR the `ibdMode` guard in `handleHeadersMsg` blocked this. Can you confirm whether that's reachable, and if so gate the fall-through on the peer being the sync peer or a sync candidate?

netsync/manager_test.go
```diff
@@ -1221,6 +1226,426 @@ func TestStartSyncChainCurrent(t *testing.T) {
...
+
+	sm.handleDonePeerMsg(disconnectedPeer)
+
+	require.Nil(t, sm.syncPeer)
+	require.False(t, sm.ibdMode)
+
+	blockHash := block2.Hash()
+	inv := wire.NewMsgInvSizeHint(1)
+	err = inv.AddInvVect(wire.NewInvVect(
+		wire.InvTypeWitnessBlock, blockHash,
```
saubyk, line 1284: The announcement here uses `InvTypeWitnessBlock`, but real peers announce with `InvTypeBlock`. When I swapped this to `InvTypeBlock` the `requestedBlocks` assertion fails even with the fix present, because the mock peer from `newSyncCandidate` isn't witness-enabled and `handleInvMsg` drops the inv. So this test currently proves recovery through the direct `ibdMode` field check, not through the announcement path users actually hit.

The two property tests below (lines 1366 and 1594) announce the same way. Could the test peer get the witness service flag so the tests can use `InvTypeBlock`? Ideally the announced block also goes through `handleBlockMsg` so we see it accepted end to end.

netsync/manager_test.go
```diff
@@ -1221,6 +1226,426 @@ func TestStartSyncChainCurrent(t *testing.T) {
 		"ibdMode should not be activated when chain is already current")
 }
 
+// TestLostSyncPeerNearTip verifies that losing the sync peer near the tip
```
saubyk, line 1229: Coverage gap worth flagging against the PR description ("remaining peers announce and request blocks normally"): all three new tests start with headers and blocks at the same height. The state from the issue report is headers ahead of blocks. In that state, if the sync peer is lost and no higher peer exists, the already-validated headers are never re-requested and recovery waits for the next block to be mined. That's pre-existing rather than introduced here, but the description reads as if it's covered. Could we either add a case that processes a header ahead of the tip before the disconnect, or narrow the description to what's actually fixed?

saubyk (commented):
Thanks for the quick turnaround on #2606. I had the change checked out and exercised locally: build, vet, and the netsync tests (including the new ones, with `-race`) all pass on this head.

From a product/risk angle the fix does what the issue asks, but it also opens a couple of side doors that I'd like a second look at before this ships. Details are inline; two more that don't sit on changed lines:

- **`handleInvMsg` (~line 1165) still trusts the `ibdMode` flag alone.** This PR fixes the flag at one write site. If any future code path clears the sync peer without clearing the flag, we're back to #2606 with no test catching it. `handleHeadersMsg` already double-checks `syncPeer != nil` locally. Is there a reason not to do the same here? It should be behavior-neutral today.
- **Full UTXO cache flush when the sync peer is lost mid-IBD (~line 834).** With `ibdMode` now cleared even when the node is far from the tip, the first block from a remaining peer can hit the periodic-flush branch and write the whole cache to disk, which IBD then refills. One-time cost, not a correctness issue, but it's a new user-visible slowdown that didn't exist before. Gating that flush on chain currency instead of the flag would avoid it.

Small one: the doc comment on `updateSyncPeer` still mentions resetting header prefetch state, which was removed a while back, and doesn't mention that the function now leaves IBD mode. Worth folding the inline note into the doc comment.

saubyk:
Tested this end-to-end on a real btcd process (not just the unit test), since this PR targets the stall we reported in #2606. Result: the PR fixes it; stock v0.26.2 and current master both reproduce the stall under the same script.

| build | commit | after 2 new blocks are announced by inv | `node disconnect` + reconnect remedy |
|---|---|---|---|
| v0.26.2 | 9d95119 | **stalled** at 10 (peer at 12), no `syncnode` | recovers to 12 |
| master | 05585e0 | **stalled** at 10 (peer at 12), no `syncnode` | recovers to 12 |
| this PR | 8ad3174 | **follows** to 11, then 12 | not needed |

Each build was run through the sequence twice with identical results.

### Setup

All on loopback, regtest, three processes: a miner btcd (`--miningaddr`, blocks via `generate`), the btcd under test, and a ~100-line "fake peer" built on btcd's own `peer` package. The fake peer completes the version handshake advertising a height above the node's tip, waits until the node elects it as sync peer (i.e. the first `getheaders` arrives), then drops the connection. That makes "sync peer lost mid-sync" deterministic instead of racing a real peer's disconnect.

Sequence, matching the production log in the issue:

1. Node under test syncs 10 blocks from the miner; miner is its sync peer.
2. Fake peer connects claiming height 15.
3. `btcctl node disconnect <miner>` → `Lost peer <miner>` → netsync elects the fake peer → `Downloading headers for blocks 11 to 15`.
4. Fake peer drops on that `getheaders` → `Lost peer <fakepeer>`. `syncPeer` is now nil.
5. Miner reconnects at **equal** height, so no peer is strictly ahead. `getpeerinfo` shows no `syncnode: true` — the signature from both production incidents.
6. Miner mines two blocks, 10 s apart, each announced by inv.
7. The remedy we used in production: disconnect and reconnect a peer, whose new version message carries the higher height.

### v0.26.2 — stalls

Node-under-test log from step 2 onward. Nothing happens between the reconnect at 11:56:26 and our manual remedy at 11:56:49, even though the miner announced blocks 11 and 12 in between:

```
11:56:25.942 [INF] SYNC: New valid peer 127.0.0.1:50356 (inbound) (/btcwire:0.5.0/fakepeer2606:0.0.1/)
11:56:25.989 [INF] SYNC: Lost peer 127.0.0.1:28801 (outbound)
11:56:25.989 [INF] SYNC: Downloading headers for blocks 11 to 15 from peer 127.0.0.1:50356
11:56:25.990 [INF] SYNC: Lost peer 127.0.0.1:50356 (inbound)
11:56:26.022 [INF] SYNC: New valid peer 127.0.0.1:28801 (outbound) (/btcwire:0.5.0/btcd:0.26.2/)
11:56:49.024 [INF] SYNC: Lost peer 127.0.0.1:28801 (outbound)
11:56:49.046 [INF] SYNC: New valid peer 127.0.0.1:28801 (outbound) (/btcwire:0.5.0/btcd:0.26.2/)
11:56:49.046 [INF] SYNC: Downloading headers for blocks 11 to 12 from peer 127.0.0.1:28801
11:56:49.046 [INF] SYNC: downloaded headers to 6230963cb336f4c72de887e88992b329e682c2784c693412d4e16175c2ca6fa7(12) from peer 127.0.0.1:28801 (outbound) -- now fetching blocks
11:56:49.047 [INF] SYNC: Finished the initial block download and caught up to block 6230963cb336f4c72de887e88992b329e682c2784c693412d4e16175c2ca6fa7(12) -- now listening to blocks.
```

### This PR — recovers

Same sequence. Blocks 11 and 12 are fetched via the normal inv path as soon as they are announced (no sync-peer lines because none is needed); the remedy step is a no-op because the node is already at 12:

```
11:56:23.231 [INF] SYNC: New valid peer 127.0.0.1:50320 (inbound) (/btcwire:0.5.0/fakepeer2606:0.0.1/)
11:56:23.265 [INF] SYNC: Lost peer 127.0.0.1:28801 (outbound)
11:56:23.265 [INF] SYNC: Downloading headers for blocks 11 to 15 from peer 127.0.0.1:50320
11:56:23.265 [INF] SYNC: Lost peer 127.0.0.1:50320 (inbound)
11:56:23.296 [INF] SYNC: New valid peer 127.0.0.1:28801 (outbound) (/btcwire:0.5.0/btcd:0.26.2/)
11:56:24.506 [INF] SYNC: Lost peer 127.0.0.1:28801 (outbound)
11:56:24.526 [INF] SYNC: New valid peer 127.0.0.1:28801 (outbound) (/btcwire:0.5.0/btcd:0.26.2/)
```

### Reproduction

<details>
<summary><code>fakepeer/main.go</code> — build inside the btcd module, e.g. as <code>cmd/fakepeer2606/main.go</code></summary>

```go
// fakepeer is a minimal Bitcoin P2P peer built on btcd's own peer package.
// It exists to reproduce btcsuite/btcd#2606 deterministically:
//
//   - it dials a btcd node and completes the version handshake advertising a
//     starting height above the node's tip (-height), which makes the node's
//     netsync manager treat it as a "higher" peer;
//   - as soon as the node elects it as sync peer and sends the first
//     getheaders, it drops the connection.
//
// That is exactly what a sync peer that disconnects mid-sync looks like
// from the node's point of view, minus the race of timing a real peer's
// disconnect.
//
// -mkaddr prints a valid regtest P2PKH address for --miningaddr and exits.
package main

import (
	"crypto/sha256"
	"flag"
	"fmt"
	"net"
	"os"
	"time"

	"github.com/btcsuite/btcd/address/v2"
	"github.com/btcsuite/btcd/chaincfg/v2"
	"github.com/btcsuite/btcd/chainhash/v2"
	"github.com/btcsuite/btcd/peer"
	"github.com/btcsuite/btcd/wire/v2"
)

func logf(format string, args ...interface{}) {
	fmt.Printf(time.Now().UTC().Format("15:04:05.000")+" fakepeer: "+format+"\n", args...)
}

func main() {
	connect := flag.String("connect", "", "host:port of the btcd node to dial")
	height := flag.Int("height", 0, "starting height to advertise in our version message")
	timeout := flag.Duration("timeout", 60*time.Second, "give up if no getheaders arrives in time")
	mkaddr := flag.Bool("mkaddr", false, "print a regtest mining address and exit")
	flag.Parse()

	params := &chaincfg.RegressionNetParams

	if *mkaddr {
		h := sha256.Sum256([]byte("btcd-2606-repro-miningaddr"))
		a, err := address.NewAddressPubKeyHash(h[:20], params)
		if err != nil {
			fmt.Fprintln(os.Stderr, err)
			os.Exit(1)
		}
		fmt.Println(a.EncodeAddress())
		return
	}
	if *connect == "" || *height <= 0 {
		fmt.Fprintln(os.Stderr, "usage: fakepeer -connect host:port -height N")
		os.Exit(2)
	}

	gotGetHeaders := make(chan struct{}, 1)
	cfg := &peer.Config{
		UserAgentName:    "fakepeer2606",
		UserAgentVersion: "0.0.1",
		ChainParams:      params,
		Services:         wire.SFNodeNetwork | wire.SFNodeWitness,
		NewestBlock: func() (*chainhash.Hash, int32, error) {
			return params.GenesisHash, int32(*height), nil
		},
		Listeners: peer.MessageListeners{
			OnVersion: func(p *peer.Peer, msg *wire.MsgVersion) *wire.MsgReject {
				logf("node's version: agent=%s height=%d", msg.UserAgent, msg.LastBlock)
				return nil
			},
			OnVerAck: func(p *peer.Peer, _ *wire.MsgVerAck) {
				logf("handshake complete; we advertised height %d", *height)
			},
			OnGetHeaders: func(p *peer.Peer, msg *wire.MsgGetHeaders) {
				logf("node sent getheaders (locator=%d hashes) -> we were elected sync peer; dropping connection now",
					len(msg.BlockLocatorHashes))
				select {
				case gotGetHeaders <- struct{}{}:
				default:
				}
				p.Disconnect()
			},
		},
	}

	p, err := peer.NewOutboundPeer(cfg, *connect)
	if err != nil {
		fmt.Fprintln(os.Stderr, err)
		os.Exit(1)
	}
	conn, err := net.Dial("tcp", *connect)
	if err != nil {
		fmt.Fprintln(os.Stderr, err)
		os.Exit(1)
	}
	logf("dialed %s from %s", *connect, conn.LocalAddr())
	p.AssociateConnection(conn)

	done := make(chan struct{})
	go func() {
		p.WaitForDisconnect()
		close(done)
	}()

	select {
	case <-done:
	case <-time.After(*timeout):
		logf("timeout: no getheaders within %s", *timeout)
		p.Disconnect()
		<-done
		os.Exit(1)
	}
	select {
	case <-gotGetHeaders:
		logf("disconnected after getheaders; exiting")
	default:
		logf("disconnected WITHOUT ever receiving getheaders")
		os.Exit(3)
	}
}
```
</details>

<details>
<summary><code>run.sh &lt;btcd-binary&gt; &lt;label&gt;</code> — expects <code>bin/btcctl</code> and <code>bin/fakepeer</code> next to it</summary>

```bash
#!/usr/bin/env bash
# End-to-end reproduction of btcsuite/btcd#2606 against one btcd binary.
#
#   ./run.sh <btcd-binary> <label>
#
# Topology (all on loopback, all regtest, no Docker):
#
#   miner  (btcd, --miningaddr, mines with `generate`)  <--p2p-->  SUT (btcd under test)
#   fakepeer (bin/fakepeer)  --p2p-->  SUT
#
# Sequence, mirroring the production log in the issue:
#   1. miner mines 10 blocks; SUT connects and syncs (miner is SUT's sync peer).
#   2. fakepeer connects to SUT advertising height 15 (a "higher" peer).
#   3. SUT is told to disconnect the miner  -> "Lost peer <miner>"; netsync
#      elects fakepeer and sends getheaders -> "Downloading headers 11..15".
#   4. fakepeer drops on that getheaders     -> "Lost peer <fakepeer>".
#      SUT now has syncPeer == nil. Stock btcd leaves ibdMode == true.
#   5. SUT reconnects to the miner at EQUAL height (so no strictly-higher
#      peer exists, exactly as in production after the sync peer was lost).
#   6. miner mines 2 blocks, one at a time. Each is announced via inv.
#      STALLED  = SUT stays at 10        (announcements discarded)
#      RECOVERS = SUT reaches 12         (announcements fetched)
#   7. Remedy from the issue: disconnect + reconnect a peer. The miner's new
#      version message now carries height 12, so startSync elects it.
set -euo pipefail

HERE=$(cd "$(dirname "$0")" && pwd)
BTCD=$1
LABEL=$2
BTCCTL=$HERE/bin/btcctl
FAKEPEER=$HERE/bin/fakepeer
OUT=$HERE/out/$LABEL
rm -rf "$OUT"
mkdir -p "$OUT/miner" "$OUT/sut"

PM=28801 RM=28802   # miner p2p / rpc
PS=28811 RS=28812   # SUT   p2p / rpc
U=repro P=repro
MINER_ADDR=127.0.0.1:$PM
MINING_ADDR=$("$FAKEPEER" -mkaddr)

log() { printf '%s | %s\n' "$(date -u +%H:%M:%S)" "$*" | tee -a "$OUT/run.log"; }
ctl_m() { "$BTCCTL" --configfile=/dev/null --regtest --notls --rpcserver=127.0.0.1:$RM --rpcuser=$U --rpcpass=$P "$@"; }
ctl_s() { "$BTCCTL" --configfile=/dev/null --regtest --notls --rpcserver=127.0.0.1:$RS --rpcuser=$U --rpcpass=$P "$@"; }

start_node() { # name p2p rpc [extra btcd flags...]
	local name=$1 p2p=$2 rpc=$3
	shift 3
	"$BTCD" --regtest --datadir="$OUT/$name" --logdir="$OUT/$name/logs" \
		--listen=127.0.0.1:"$p2p" --rpclisten=127.0.0.1:"$rpc" --notls \
		--rpcuser=$U --rpcpass=$P --nodnsseed --debuglevel=SYNC=debug \
		"$@" >"$OUT/$name/stdout.log" 2>&1 &
	echo $!
}

MPID=""; SPID=""
cleanup() {
	[ -n "$SPID" ] && kill "$SPID" 2>/dev/null || true
	[ -n "$MPID" ] && kill "$MPID" 2>/dev/null || true
	wait 2>/dev/null || true
}
trap cleanup EXIT

wait_rpc() { # ctl-fn
	for _ in $(seq 1 60); do
		if "$1" getblockcount >/dev/null 2>&1; then return 0; fi
		sleep 0.5
	done
	log "FATAL: rpc for $1 never came up"; return 1
}
wait_height() { # ctl-fn want timeout-seconds -> 0 if reached
	local ctl=$1 want=$2 secs=$3 h
	for _ in $(seq 1 $((secs * 2))); do
		h=$($ctl getblockcount 2>/dev/null || echo -1)
		[ "$h" = "$want" ] && return 0
		sleep 0.5
	done
	return 1
}
peer_count() { ctl_s getpeerinfo | grep -c '"addr"' || true; }
wait_peers() { # want timeout
	for _ in $(seq 1 $(($2 * 2))); do
		[ "$(peer_count)" = "$1" ] && return 0
		sleep 0.5
	done
	log "FATAL: SUT never reached $1 peer(s); has $(peer_count)"; return 1
}
syncnode_count() { ctl_s getpeerinfo | grep -c '"syncnode": true' || true; }
peers_brief() {
	ctl_s getpeerinfo | grep -E '"(addr|inbound|startingheight|currentheight|syncnode)"' \
		| tr -d ' ,"' | paste -d' ' - - - - - | sed 's/^/      /'
}

log "=== $LABEL: $("$BTCD" --version) ==="

# 1. miner up, 10 blocks; SUT up, sync from miner.
MPID=$(start_node miner $PM $RM --miningaddr="$MINING_ADDR")
wait_rpc ctl_m
ctl_m generate 10 >/dev/null
wait_height ctl_m 10 10 || { log "FATAL: miner did not reach 10"; exit 1; }
log "miner at height $(ctl_m getblockcount)"

SPID=$(start_node sut $PS $RS)
wait_rpc ctl_s
ctl_s node connect $MINER_ADDR temp
wait_height ctl_s 10 30 || { log "FATAL: SUT did not sync to 10"; exit 1; }
log "SUT synced to height $(ctl_s getblockcount); syncnode peers: $(syncnode_count)"

# 2. fakepeer connects advertising height 15.
"$FAKEPEER" -connect 127.0.0.1:$PS -height 15 -timeout 60s >"$OUT/fakepeer.log" 2>&1 &
FPID=$!
wait_peers 2 15
log "fakepeer connected (SUT peers: $(peer_count), syncnode peers: $(syncnode_count))"

# 3+4. lose the sync peer; netsync elects fakepeer; fakepeer drops on getheaders.
ctl_s node disconnect $MINER_ADDR
if wait "$FPID"; then
	log "fakepeer: $(tail -1 "$OUT/fakepeer.log")"
else
	log "FATAL: fakepeer exited $? -- $(tail -1 "$OUT/fakepeer.log")"; exit 1
fi
wait_peers 0 15
log "SUT has lost both peers; syncPeer is now nil"

# 5. reconnect miner at equal height.
ctl_s node connect $MINER_ADDR temp
wait_peers 1 15
sleep 1
log "miner reconnected at equal height; syncnode peers: $(syncnode_count)"
peers_brief | tee -a "$OUT/run.log"

# 6. mine two blocks, one at a time; watch whether SUT follows.
for want in 11 12; do
	ctl_m generate 1 >/dev/null
	if wait_height ctl_s $want 10; then
		log "miner mined block $want -> SUT followed to $want"
	else
		log "miner mined block $want -> SUT STILL AT $(ctl_s getblockcount) after 10s"
	fi
done
FINAL=$(ctl_s getblockcount)
if [ "$FINAL" = "12" ]; then VERDICT="RECOVERS"; else VERDICT="STALLED at $FINAL (miner at $(ctl_m getblockcount))"; fi
log "VERDICT [$LABEL]: $VERDICT; syncnode peers: $(syncnode_count)"
peers_brief | tee -a "$OUT/run.log"

# 7. production remedy: disconnect + reconnect a peer.
ctl_s node disconnect $MINER_ADDR
wait_peers 0 15
ctl_s node connect $MINER_ADDR temp
if wait_height ctl_s 12 20; then
	log "remedy (disconnect+reconnect peer): SUT now at $(ctl_s getblockcount)"
else
	log "remedy (disconnect+reconnect peer): SUT STILL AT $(ctl_s getblockcount) after 20s"
fi

log "--- SUT netsync log ---"
grep -E '\] (SYNC|PEER|SRVR):' "$OUT/sut/stdout.log" \
	| grep -vE 'Ignoring|Received done peer|Adding TX|new peer state' \
	| sed -E 's/^[0-9-]+ //' | sed 's/^/      /' | tee -a "$OUT/run.log"
echo "$VERDICT" >"$OUT/verdict"
```
</details>

Caveats: this forces the "replacement sync peer lost" event with a scripted peer, so it proves the mechanism rather than how often it happens in the wild. It also does not exercise the 3-minute stall-handler path, though that ends in the same `updateSyncPeer` call this PR changes.

netsync/manager.go
kcalvinalvin: This test covers that case. You're correct in that this does force a fetch from any peer that sends headers.

We could gate it again but the original reason why I left the node accepting headers from any peers is so that we can eventually do headers block propagation instead of the inv msg block propagation. But yeah right now we'd be screwed.

```
+// TestHeadersMsgRequestGating verifies that a headers message only triggers
+// block downloads when it comes from the sync peer or a peer eligible to
+// become the sync peer.  Otherwise a peer that never serves the blocks could
+// reserve hashes in the shared requested set, which the eventual sync peer
+// then skips and the download stalls.
+func TestHeadersMsgRequestGating(t *testing.T) {
+       t.Parallel()
+
+       const numBlocks = 8
+
+       tests := []struct {
+               name          string
+               candidate     bool
+               wantRequested bool
+       }{
+               {
+                       name:          "non-candidate peer",
+                       candidate:     false,
+                       wantRequested: false,
+               },
+               {
+                       name:          "sync candidate peer",
+                       candidate:     true,
+                       wantRequested: true,
+               },
+       }
+
+       for _, tc := range tests {
+               t.Run(tc.name, func(t *testing.T) {
+                       t.Parallel()
+
+                       params := chaincfg.RegressionNetParams
+                       params.Checkpoints = nil
+
+                       sm, tearDown := makeMockSyncManager(t, &params)
+                       defer tearDown()
+
+                       blocks := generateTestBlocks(t, &params, numBlocks)
+
+                       // Register the peer under test directly;
+                       // handleNewPeerMsg marks every regtest peer a
+                       // candidate, so it cannot produce a
+                       // non-candidate peer.
+                       p := peer.NewInboundPeer(&peer.Config{
+                               ChainParams: &params,
+                       })
+
+                       // Place it at our height so startSync never
+                       // treats it as a higher peer.
+                       p.UpdateLastBlockHeight(0)
+                       sm.peerStates[p] = &peerSyncState{
+                               syncCandidate: tc.candidate,
+                               requestedTxns: make(
+                                       map[chainhash.Hash]struct{},
+                               ),
+                               requestedBlocks: make(
+                                       map[chainhash.Hash]struct{},
+                               ),
+                       }
+
+                       // Reach the post-loss state through the real
+                       // transition: startSync promotes a strictly
+                       // higher peer, which is then dropped, leaving
+                       // no higher candidate to take over.
+                       syncPeer := newSyncCandidate(t, sm, 1)
+                       sm.startSync()
+                       sm.handleDonePeerMsg(syncPeer)
+
+                       // Feed solved headers from the peer under test.
+                       // Processing them is not gated, only the block
+                       // requests they produce are.
+                       headers := wire.NewMsgHeaders()
+                       for _, block := range blocks {
+                               err := headers.AddBlockHeader(
+                                       &block.MsgBlock().Header,
+                               )
+                               require.NoError(t, err)
+                       }
+                       sm.handleHeadersMsg(&headersMsg{
+                               headers: headers, peer: p,
+                       })
+
+                       peerState := sm.peerStates[p]
+
+                       // Headers are processed regardless of who sends them;
+                       // only the block download is gated.
+                       _, bestHeight := sm.chain.BestHeader()
+                       require.Equal(t, int32(numBlocks), bestHeight)
+
+                       if !tc.wantRequested {
+                               // A peer that will never sync must not
+                               // reserve blocks in the shared set.
+                               require.Empty(t, sm.requestedBlocks)
+                               require.Empty(t, peerState.requestedBlocks)
+                               return
+                       }
+
+                       // A sync candidate fetches the blocks, so the
+                       // shared set mirrors its in-flight request.
+                       require.NotEmpty(t, sm.requestedBlocks)
+                       require.Equal(t, sm.requestedBlocks,
+                               peerState.requestedBlocks)
+               })
+       }
 }
```

netsync/manager_test.go
kcalvinalvin: Good catch. This here allows one to make a witness enabled peer.

```
+// handshakeConn is a net.Conn that feeds a version message followed by a
+// verack to the peer, discards everything the peer writes, and then blocks
+// until closed. It lets tests bring up a fully connected, witness-enabled
+// peer without a real remote.
+type handshakeConn struct {
+       reader    *bytes.Reader
+       addr      net.Addr
+       closed    chan struct{}
+       closeOnce sync.Once
+}
+
+// newHandshakeConn returns a conn that completes the version handshake with
+// the given advertised service flags.
+func newHandshakeConn(btcnet wire.BitcoinNet,
+       services wire.ServiceFlag) *handshakeConn {
+
+       na := wire.NewNetAddressIPPort(net.IPv4(127, 0, 0, 1), 8333, services)
+       ver := wire.NewMsgVersion(na, na, 0, 0)
+       ver.Services = services
+
+       var buf bytes.Buffer
+       _ = wire.WriteMessage(
+               &buf, ver, wire.ProtocolVersion, btcnet,
+       )
+       _ = wire.WriteMessage(
+               &buf, wire.NewMsgVerAck(), wire.ProtocolVersion, btcnet,
+       )
+
+       return &handshakeConn{
+               reader: bytes.NewReader(buf.Bytes()),
+               addr: &net.TCPAddr{
+                       IP: net.IPv4(127, 0, 0, 1), Port: 8333,
+               },
+               closed: make(chan struct{}),
+       }
+}
+
+func (c *handshakeConn) Read(b []byte) (int, error) {
+       if c.reader.Len() > 0 {
+               return c.reader.Read(b)
+       }
+       <-c.closed
+       return 0, io.EOF
+}
+
+func (c *handshakeConn) Write(b []byte) (int, error) { return len(b), nil }
+
+func (c *handshakeConn) Close() error {
+       c.closeOnce.Do(func() { close(c.closed) })
+       return nil
+}
+
+func (c *handshakeConn) LocalAddr() net.Addr              { return c.addr }
+func (c *handshakeConn) RemoteAddr() net.Addr             { return c.addr }
+func (c *handshakeConn) SetDeadline(time.Time) error      { return nil }
+func (c *handshakeConn) SetReadDeadline(time.Time) error  { return nil }
+func (c *handshakeConn) SetWriteDeadline(time.Time) error { return nil }
+
+// newWitnessSyncCandidate creates and registers a connected, witness-enabled
+// sync candidate at the given height. The peer completes an emulated version
+// handshake so IsWitnessEnabled reports true, letting tests announce blocks
+// with InvTypeBlock as a real peer would.
+func newWitnessSyncCandidate(t *testing.T, sm *SyncManager,
+       height int32) *peer.Peer {
+
+       t.Helper()
+
+       verAck := make(chan struct{}, 1)
+       p := peer.NewInboundPeer(&peer.Config{
+               ChainParams:    sm.chainParams,
+               AllowSelfConns: true,
+               Listeners: peer.MessageListeners{
+                       OnVerAck: func(*peer.Peer, *wire.MsgVerAck) {
+                               select {
+                               case verAck <- struct{}{}:
+                               default:
+                       },
+               },
+       })
+       p.AssociateConnection(newHandshakeConn(
+               sm.chainParams.Net, wire.SFNodeNetwork|wire.SFNodeWitness,
+       ))
+
+       select {
+       case <-verAck:
+       case <-time.After(5 * time.Second):
+               t.Fatal("timed out waiting for peer handshake")
+       }
+       require.True(t, p.IsWitnessEnabled())
+
+       p.UpdateLastBlockHeight(height)
+       sm.peerStates[p] = &peerSyncState{
+               syncCandidate:   true,
+               requestedTxns:   make(map[chainhash.Hash]struct{}),
+               requestedBlocks: make(map[chainhash.Hash]struct{}),
+       }
+
+       // Release the peer's handler goroutines when the test ends.
+       t.Cleanup(p.Disconnect)
+
+       return p
+}
```

The modified test worked. Added code for the actual block acceptance too. 

```diff
 // assertIBDComplete verifies that IBD finished: chain height matches
 // totalBlocks, ibdMode is off, and no blocks remain in-flight.
 func assertIBDComplete(t *testing.T, sm *SyncManager,
@@ -1269,7 +1376,7 @@ func TestLostSyncPeerNearTip(t *testing.T) {
        })

        disconnectedPeer := newSyncCandidate(t, sm, 1)
-       remainingPeer := newSyncCandidate(t, sm, 1)
+       remainingPeer := newWitnessSyncCandidate(t, sm, 1)
        sm.syncPeer = disconnectedPeer
        sm.ibdMode = true

@@ -1278,16 +1385,191 @@ func TestLostSyncPeerNearTip(t *testing.T) {
        require.Nil(t, sm.syncPeer)
        require.False(t, sm.ibdMode)

+       // A real post-segwit peer announces blocks with InvTypeBlock, so the
+       // announcing peer must be witness-enabled for the inv to be accepted.
        blockHash := block2.Hash()
        inv := wire.NewMsgInvSizeHint(1)
        err = inv.AddInvVect(wire.NewInvVect(
-               wire.InvTypeWitnessBlock, blockHash,
+               wire.InvTypeBlock, blockHash,
        ))
        require.NoError(t, err)
        sm.handleInvMsg(&invMsg{peer: remainingPeer, inv: inv})

        require.Contains(t, sm.requestedBlocks, *blockHash,
                "the remaining peer's announcement should be requested")
+
+       // Deliver the announced block through the sync manager and verify it
+       // is accepted, completing the path a real peer's announcement takes.
+       sm.handleBlockMsg(&blockMsg{
+               block: block2, peer: remainingPeer,
+               reply: make(chan struct{}, 1),
+       })
+       require.Equal(t, *blockHash, sm.chain.BestSnapshot().Hash,
+               "the announced block should be accepted end to end")
+}
```

netsync/manager_test.go
kcalvinalvin: This test asserts that. It does look like we recover ok.

```
+// TestLostSyncPeerWithAheadHeaders covers the state from the issue report
+// where the header chain is ahead of the block chain when the sync peer is
+// lost, and the recovery paths that follow. The header-first path requests the
+// pending block while the sync peer is active. Losing it clears that request
+// with no higher peer to take over, so the pending block is not requested on
+// its own. A newly connected peer then recovers: at the header tip it has the
+// blocks requested directly, while above the header tip it is first asked for
+// headers.
+func TestLostSyncPeerWithAheadHeaders(t *testing.T) {
+       t.Parallel()
+
+       tests := []struct {
+               name          string
+               newPeerHeight int32
+       }{
+               {
+                       name:          "at header tip fetches blocks",
+                       newPeerHeight: 2,
+               },
+               {
+                       name:          "above header tip fetches headers",
+                       newPeerHeight: 3,
+               },
+       }
+
+       for _, tc := range tests {
+               t.Run(tc.name, func(t *testing.T) {
+                       t.Parallel()
+
+                       params := chaincfg.RegressionNetParams
+                       params.Checkpoints = nil
+
+                       sm, tearDown := makeMockSyncManager(t, &params)
+                       defer tearDown()
+
+                       blocks := generateTestBlocks(t, &params, 3)
+                       block1, block2 := blocks[0], blocks[1]
+
+                       // Block 1 is connected while block 2 is only a header.
+                       _, _, err := sm.chain.ProcessBlock(
+                               block1, blockchain.BFNone,
+                       )
+                       require.NoError(t, err)
+
+                       // The sync peer is promoted and delivers block 2's
+                       // header. Reaching its advertised height makes the
+                       // header-first path request the block.
+                       syncPeer := newSyncCandidate(t, sm, 2)
+                       sm.startSync()
+                       require.Equal(t, syncPeer, sm.syncPeer)
+                       require.True(t, sm.ibdMode)
+
+                       headers := wire.NewMsgHeaders()
+                       require.NoError(t, headers.AddBlockHeader(
+                               &block2.MsgBlock().Header,
+                       ))
+                       sm.handleHeadersMsg(&headersMsg{
+                               headers: headers, peer: syncPeer,
+                       })
+
+                       blockHash := block2.Hash()
+                       require.Contains(t, sm.requestedBlocks, *blockHash,
+                               "the header path should request the block")
+
+                       // Losing the sync peer clears the request and no
+                       // remaining peer is higher, so the pending block is
+                       // not requested on its own.
+                       sm.handleDonePeerMsg(syncPeer)
+
+                       require.Nil(t, sm.syncPeer)
+                       require.False(t, sm.ibdMode)
+                       require.NotContains(t, sm.requestedBlocks, *blockHash)
+
+                       // The header chain stays caught up to the announced
+                       // height.
+                       bestHash, bestHeight := sm.chain.BestHeader()
+                       require.Equal(t, int32(2), bestHeight)
+                       require.Equal(t, *blockHash, bestHash)
+
+                       // A newly connected peer recovers. At the header tip
+                       // it has the blocks requested directly; above the
+                       // header tip it is first asked for headers.
+                       newPeer := newWitnessSyncCandidate(
+                               t, sm, tc.newPeerHeight,
+                       )
+                       sm.handleNewPeerMsg(newPeer)
+
+                       require.Equal(t, newPeer, sm.syncPeer)
+                       require.True(t, sm.ibdMode)
+
+                       // A peer above our header tip is asked for headers
+                       // first; one at the tip has the pending block
+                       // requested directly.
+                       if tc.newPeerHeight > bestHeight {
+                               require.Empty(t, sm.requestedBlocks)
+
+                               nextHeaders := wire.NewMsgHeaders()
+                               require.NoError(t, nextHeaders.AddBlockHeader(
+                                       &blocks[2].MsgBlock().Header,
+                               ))
+                               sm.handleHeadersMsg(&headersMsg{
+                                       headers: nextHeaders, peer: newPeer,
+                               })
+                       }
+                       require.Contains(t, sm.requestedBlocks, *blockHash,
+                               "recovery should request the pending block")
+
+                       // Deliver the block and verify it is accepted.
+                       sm.handleBlockMsg(&blockMsg{
+                               block: block2, peer: newPeer,
+                               reply: make(chan struct{}, 1),
+                       })
+                       require.Equal(t, *blockHash,
+                               sm.chain.BestSnapshot().Hash)
+               })
+       }
+}
```

PR #2608, opened 2026-09-17
rpcclient: add context.Context support for HTTP POST mode

seeforschauer:
## Change Description

Add `context.Context` support to rpcclient's HTTP POST mode, allowing callers to cancel or timeout individual RPC calls. Fixes #2499.

Currently, rpcclient methods like `GetBlockCount()` don't accept a context, which means callers have no way to cancel or set deadlines on individual RPC calls. In HTTP POST mode, `handleSendPostMessage` uses `http.NewRequest` without context, and the 10-retry loop doesn't respect external cancellation.

### Changes

- Add `ctx` field to `jsonRequest` struct
- Use `http.NewRequestWithContext` in `handleSendPostMessage` when ctx is set
- Respect context cancellation during retry backoff (`select` on `reqCtx.Done()`)
- Add `SendCmdWithContext(ctx, cmd)` to `Client`
- Add `GetBlockCountWithContext(ctx)` as the first context-aware method variant
- Existing methods are unchanged — fully backwards compatible

### Design decisions

- `SendCmdWithContext` mirrors `SendCmd` with an added ctx parameter
- When `ctx` is nil (e.g., from existing `SendCmd` callers), `context.Background()` is used as fallback
- Context cancellation during retry backoff returns immediately instead of waiting for next attempt
- Additional `*WithContext` methods for other RPC calls can be added incrementally in follow-up PRs

## Steps to Test

```
go test ./rpcclient/ -run TestGetBlockCountWithContext -v
go test ./rpcclient/ -run TestSendCmdWithContext -v
go test ./rpcclient/ -run TestGetBlockCount_Backwards -v
```

### Testing

- [x] The change is not insubstantial.


odingaval (approved):
Re-review of #2506 which was closed.
 
 rpcclient/infrastructure.go L792-L802 `postRequestContext`:
 - Correct use of `WithCancelCause` + `AfterFunc` to merge `requestCtx` and `clientCtx`
 - `stopClientCancel()` + `cancel(nil)` in the returned cleanup func prevents leak - looks good
 *L817-L819 `sendPostRequestWithRetry`:*
 - `reqCtx, releaseReqCtx := postRequestContext(ctx, jReq.ctx)` + `defer releaseReqCtx()` - ensures merged context is always released, even on retry loop early return
 - This is the fix for the goroutine accumulation described in #2451 - HTTP requests will now respect `ctx.Done()` instead of blocking 600s in `GetBlockCount`
 Same code as before, still looks good
This fixes the leak at the library layer.

PR #2612, opened 2026-09-21
btcutil/gcs: drop the redundant remainder subtraction

kcalvinalvin:
## Summary

`BuildGCSFilter` computes the Golomb quotient of the gap between two sorted
values as

```go
value = (v - lastValue - remainder) >> f.p
```

Subtracting `remainder` first clears the low `p` bits, but the shift discards
those bits anyway, so the result is the same as

```go
value = (v - lastValue) >> f.p
```

This drops the subtraction. `remainder` is unchanged and is still written to
the bitstream after the unary quotient, so the encoded filter is identical.

## Why it is safe

A right shift by `p` discards exactly the low `p` bits that `remainder` holds,
so clearing them first cannot change the quotient. The added test states that
identity directly and checks it for every collision parameter the filter
accepts.

The serialized output was also confirmed byte-for-byte identical to the
previous code across several collision parameters.


saubyk (approved):
Pulled this locally and ran `go test ./...` in btcutil/gcs. Both the gcs and
gcs/builder packages pass with the change applied.

The reasoning makes sense to me: the shift already drops the low P bits, so
subtracting the remainder first was doing nothing. Encoded filters stay
byte-for-byte the same, so no light-client impact.

LGTM.

Lrifton92 (approved):
Verified at 291f6b53.

- The identity holds for every P the builder accepts: `BuildGCSFilter` rejects P > 32, so the shift stays below 64, and `remainder` is computed from the same `v - lastValue`, so it is exactly the low P bits the shift discards.
- Byte-for-byte check: I built 1,400 filters with fixed keys and data (P in {0, 1, 5, 19, 20, 31, 32}, 1 to 97 elements each) on master (b48125d) and on this branch. The SHA-256 of the concatenated `NBytes()` output is identical on both (`0cf4291c…`).
- `go test ./gcs/...` passes.

One non-blocking note on the test: `TestQuotientIgnoresRemainder` checks the arithmetic identity on its own and never calls `BuildGCSFilter`, so it would still pass if the encoder line regressed. As a quick mutation, `value = (v - lastValue + 1) >> f.p` (wrong only when the low P bits are all ones) passes the whole gcs and builder suites; only a blunter change such as `((v - lastValue) >> f.p) + 1` is caught, through the match tests. btcd has no BIP 158 test vectors today, so nothing pins the encoded bytes. A small golden test (fixed key and data, `NBytes()` compared to a hex string captured on master), or importing Bitcoin Core's `blockfilters.json` vectors, would cover the claim in the description. Happy to open a follow-up PR with the vectors if that helps.

